PR #2279, opened 2020-09-10
Add the scheme if the UrlGenerator returns a network path

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2171
| Docs PR or issue | -

Alternative approach to #2262 that does not require to make all paths absolute.

PR #2262, opened 2020-09-09
Always use absolute paths

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2171

In Contao 4.10, we switched to using the current page model in routing. If a page's host does not match the current host, Symfony automatically adds the host name by using `NETWORK_PATH` (see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L264-L269).

Our current attempt at making paths relative breaks the `NETWORK_PATH`, and I don't think it's necessary. I know we used to need that for some browser quirks regarding the base tag, but according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) the base tag should only be applied _for all relative URLs in a document_.

I think we should adhere to the Symfony standard and always use absolute paths (or absolute URLs).

PR #2323, opened 2020-09-22
Fix the routing SSL logic

### LINKED ISSUE #2171: {{link_url}} Insert-Tag broken on roots with different domains
**Affected version(s)**

Contao 4.10

**Description**

When creating a link to a page from another root with a different domain the link generated by e.g `{{link_url::1}}` should be an absolute but Contao generates a relative one.

**Steps to reproduce**

- Create two roots with domain `a.com` and `b.com`
- Create a link on `a.com` which links to a site on `b.com` using e.g `{{link_url::1}}`
- The generated link says `/b.com/site.html` which will be interpreted as `http://a.com/b.com/site.html` because of the base path

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2171
| Docs PR or issue | -

This PR replaces #2262 and #2279.

## As discussed at the developer's meeting:

- Only switch to using absolute paths in Contao 5
- Fix the routing in Contao 4.10:
  - Checkbox not checked -> HTTP (not also HTTPS anymore!)
  - Checkbox checked -> HTTPS
- Use radio buttons instead of a checkbox

## Why radio buttons do not work

In order to remain BC, the option values need to be `''` and `'1'`, however, our `RadioButton` class interprets `''` as "not checked" and therefore never renders the HTTP option as checked. Of course, we could adjust the `Widget::optionChecked()` method, however, it would most likely have side effects. I therefore went with a select menu:

## What the PR fixes

The UrlGenerator now prepends the `http://` protocol if the current website uses HTTPS and the target website does not. ✅ 

## What the PR still does not fix

If both the current and the target website use HTTPS, the UrlGenerator still generates a network path, therefore we still need #2279 to actually fix #2171. 🤷🏻‍♂️


fritzmg (dismissed):

aschempp (commented):
I wouldn't switch the palette field position, even though I understand the intention. 

Also, can you also add the fix for #2171 here so we have one place for it all? However, we should accept that Symfony generates a network path and simply accept that, and not append the protocol on our own.

ausi:
> However, we should accept that Symfony generates a network path and simply accept that, and not append the protocol on our own.

Would be OK for me too. IMO we should then change the if condition from

```php
if (0 === strncmp($strUrl, '/', 1))
```

to:

```php
if (0 === strncmp($strUrl, '/', 1) && 0 !== strncmp($strUrl, '//', 2))
```

leofeyer:
> Also, can you also add the fix for #2171 here so we have one place for it all?

Will do.

> we should accept that Symfony generates a network path and simply accept that

Ok, if you both think so. I would prefer to generate absolute URLs, but I will bow to the majority.

aschempp (approved):

m-vo:
The description of this PR says it does not fix #2171, but #2279 does.
But #2279 is closed and points back to this PR which would solve the issue.

 I'm confused.

ausi:
> The description of this PR says it does not fix #2171

This PR was later updated to actually fix #2171.

m-vo:
Ah, thanks, think I got it now.

PR #1058, opened 2019-12-04
Do not use Chosen in the meta wizard anymore

leofeyer:
Chosen causes [issues](https://github.com/contao/contao/issues/910) in the meta wizard, therefore this PR removes it. The original use case – namely being able to quickly find your language in a long list of languages – no longer exists now that the select menu contains only the existing root page languages, therefore this change should not be problematic.

PR #2329, opened 2020-09-22
Load all languages for metawizard options

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | https://contao.slack.com/archives/CK4J0KNDB/p1600785877095400
| Docs PR or issue | n/a

The metawizard select should display all languages, not just the one available in the backend:

**wrong** 
Bildschirmfoto 2020-09-22 um 17 05 42

**correct**
Bildschirmfoto 2020-09-22 um 17 05 59

Origin: https://github.com/contao/core-bundle/commit/740ede20eb1dbffe79204cd7f4ea8ff0ed4b62ed

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/MetaWizard.php
+++ b/core-bundle/src/Resources/contao/widgets/MetaWizard.php
@@ -128,7 +128,7 @@ class MetaWizard extends Widget
 			return '<p class="tl_info">' . $GLOBALS['TL_LANG']['MSC']['metaNoLanguages'] . '</p>';
 		}
 
-		$languages = $this->getLanguages(true);
+		$languages = $this->getLanguages();
 
 		// Add the existing entries
 		if (!empty($this->varValue))


aschempp:
does this consider the source of the flag? https://github.com/contao/core/issues/6254

bytehead:
As I can see it - yes.

leofeyer:
The drop-down menu shows all languages that are

* assigned to a root page or
* installed in the core bundle.

Do you have a website root page in French?

bytehead:
Yes - but it's only displayed as `fr` (not translated) and does not show up in the select.

bytehead:
Why this:

> - installed in the core bundle.

?

bytehead:
Isn't the file meta information more or less only related to the frontend?

leofeyer:
IDK. But we do:

https://github.com/contao/contao/blob/0e1e15247894d57266b5301c8e78e154a7cbe888/core-bundle/src/Resources/contao/widgets/MetaWizard.php#L170-L174

bytehead:
It should add the metafields for the available root languages, but the select should contain all available languages, even if there is no root page configured for them.

bytehead:
> IDK. But we do:
> 
> https://github.com/contao/contao/blob/0e1e15247894d57266b5301c8e78e154a7cbe888/core-bundle/src/Resources/contao/widgets/MetaWizard.php#L170-L174

Problem is: `$languages = $this->getLanguages(true);` - you won't get any other than the installed ones...

fritzmg:
I don't think we should add back all languages (see also https://github.com/contao/contao/pull/1058). However, it's definitely a bug that French is missing, if you have a website root with language `fr`.

m-vo:
> I don't think we should add back all languages

Why would you need to limit them? What if I've got a frontend in german only but want to interact with an API in a multi-language way for instance (e.g. search with english key words)?

bytehead:
In my case, the client wants to add the meta data for all languages that will be available in the future. But at this time the corresponding page root is not available yet.

leofeyer:
Wouldn't it be easier to just add the root page then? It does not even have to be published.

Toflar:
I kind of remember us having discussed this already...

bytehead:
it just doesn't make sense to me, even the argument from #1058, because it was buggy with chosen and a big list of values... 🤷‍♂️

leofeyer:
> I kind of remember us having discussed this already...

We have indeed. And as a result, we have limited the language menu to root page languages and installed languages instead of showing a list with hundreds of options (of which most will never be used).

bytehead:
Showing only root page languages can be discussed/makes sense, but I can't see any reason for _installed languages_ - this makes never sense or do I miss something here?

leofeyer:
@bytehead Another reason why you do want to add the french root page: If you add a new file, Contao will automatically enable all root page languages. If there is no french root page, you will have to manually add the language for every file.

leofeyer:
> but I can't see any reason for installed languages - this makes never sense or do I miss something here?

Probably not. As far as I am concerned, we can remove that. @contao/developers WDYT?

ausi:
> As far as I am concerned, we can remove that.

I think I agree. But should we do this in 4.11 instead of as a bugfix?

leofeyer:
See #2404.

PR #2337, opened 2020-09-24
Fix the asset URL handling in the image studio

### LINKED ISSUE #2330: Image size set to width=0 height=0
**Affected version(s)** 
Contao 4.10

**Description**

If a files/assets static URL is set in the root page, the image dimensions are missing.

I think the following line https://github.com/contao/contao/blob/b7e1aecd3ebe998b4189f530fe857eb8df1bf788/core-bundle/src/Image/Studio/Figure.php#L260 needs to be replaced with: 

```php
$fileInfoImageSize = (new File(StringUtil::stripRootDir($image->getPicture()->getImg()['src']->getPath())))->imageSize;
```

Also, if the size cannot be determined we should set an empty string here I think: https://github.com/contao/contao/blob/b7e1aecd3ebe998b4189f530fe857eb8df1bf788/core-bundle/src/Image/Studio/Figure.php#L276 this currently results in ` width="0" height="0"` which ist most probably not desired.

/cc @m-vo

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2330 
| Docs PR or issue | -

This PR addresses two issues that occur when using asset urls:
1. the `contao.assets.files_context` was used instead of `contao.assets.asset_context`
2. the `File#imageSize` calculation for the legacy templates wasn't always getting a relative path.

/cc @ausi


ausi:
> the `contao.assets.files_context` was used instead of `contao.assets.asset_context`

I think `files_context` was actually correct 🙈, see https://github.com/contao/contao/blob/67a2f2f4cab6b54ef1732cc8d9d0599e47ff7fac/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1608 and https://github.com/contao/core-bundle/pull/661#issuecomment-269697134

m-vo:
> I think `files_context` was actually correct , see
> ...

Oh I see, a Contao 2 legacy... :fearful: 

Reverted in 9e7afa6de4cb3bdcaea76748382b32c362f37ec2.

ausi (dismissed):

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -257,7 +257,7 @@ public function getLegacyTemplateData($margin = null, string $floating = null, b
 
         $image = $this->getImage();
         $originalSize = $image->getOriginalDimensions()->getSize();
-        $fileInfoImageSize = (new File(rawurldecode($image->getImageSrc())))->imageSize;
+        $fileInfoImageSize = (array) ((new File($image->getImageSrc(true)))->imageSize);
```
leofeyer:
```suggestion
        $fileInfoImageSize = (array) (new File($image->getImageSrc(true)))->imageSize;
```

I think the extra parentheses are not necessary: https://3v4l.org/imrP3

core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -100,10 +101,18 @@ public function getImg(): array
...
-     * Returns the "src" attribute of the image.
+     * Returns the "src" attribute of the image. This will return an URL by
+     * default. Set $asPath to true to get a relative file path instead.
      */
-    public function getImageSrc(): string
+    public function getImageSrc(bool $asPath = false): string
     {
+        if ($asPath) {
+            /** @var Image $image */
+            $image = $this->getPicture()->getImg()['src'];
```
leofeyer, line 111: In line 116, you return `$this->getImg()['src']` but here you return `$this->getPicture()->getImg()['src']`. Intended?

m-vo: Yes, `$this->getImg()` calls `PictureInterface#getImg()` with the current project dir and static url whereas in this line we're using no arguments (and are then getting `Image`s back instead of urls).

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2339, opened 2020-09-24
Optimize the check for inlined services

aschempp:
Our current implementation calls `$container->getRemovedIds()` for every class before checking if its a singleton. This results in a lot of overhead, because `$container->getRemovedIds()` loads the removed IDs from file on every access. I assume that's to not memory-cache the IDs, but in our case it results in a ton of unnecessary calls.

The behavior of `System::import()` and `System::importStatic()` have slightly changed:
In debug mode, the existing behavior will remain the same. In production, we use `getInstance()` or try to create a new instance of the class before checking for removed IDs.


core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -257,13 +276,40 @@ public static function importStatic($strClass, $strKey=null, $blnForce=false)
...
+	{
+		$container = static::getContainer();
+
+		if (!$container instanceof Container)
+		{
+			return false;
+		}
+
+		// Only cache the remove service IDs in debug mode
+		$removedIds = $container->getParameter('kernel.debug') ? (static::$removedServiceIds ?: $container->getRemovedIds()) : $container->getRemovedIds();
```
fritzmg: Is there any debugging benefit from not caching the service IDs in debug mode for the current request?

leofeyer:
It seems that you are never setting `static::$removedServiceIds`. 🤔

aschempp:
Good catch! 😆 
I updated to PR to "always cache", because in production one will always immediately get an exception when checking for the services ID, either because it was inlined or because the class does not exist/cannot be instantiated.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -257,13 +276,42 @@ public static function importStatic($strClass, $strKey=null, $blnForce=false)
...
+		{
+			return false;
+		}
+
+		if (null === static::$removedServiceIds)
+		{
+			static::$removedServiceIds = $container->getRemovedIds();
+		}
+
+		return isset(static::$removedServiceIds[$strClass]);
```
ausi: As `$removedServiceIds` is `private`, shouldn’t it be accessed via `self`?

```suggestion
		if (null === self::$removedServiceIds)
		{
			self::$removedServiceIds = $container->getRemovedIds();
		}

		return isset(self::$removedServiceIds[$strClass]);
```

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -188,12 +188,12 @@ protected function import($strClass, $strKey=null, $blnForce=false)
...
 				$this->arrObjects[$strKey] = $container->get($strClass);
 			}
-			elseif (($container->getParameter('kernel.debug') || !\class_exists($strClass)) && self::isServiceInlined($strClass))
+			elseif (($container->getParameter('kernel.debug') || !class_exists($strClass)) && self::isServiceInlined($strClass))
 			{
 				// In debug mode, we check for inlined services before trying to create a new instance of the class
 				throw new ServiceNotFoundException($strClass, null, null, array(), sprintf('The "%s" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.', $strClass));
 			}
-			elseif (!\class_exists($strClass))
+			elseif (!class_exists($strClass))
```
Toflar, line 196: Why are we not using the global functions here? We should use it everywhere, also in legacy code because it's a huge performance gain.

aschempp: because the CS fixer asked me not to … 😢

Toflar: @leofeyer any reason why we do not use it? It's wasted performance :(

m-vo: >  It's wasted performance

Is it really? This will only shave off a small amount of time (in the ns range) for the function call which will only add up to a significant amount (ms range) if it happens millions of times.

But prefixing also doesn't hurt, so +1 for a homogenous coding style (imo not necessary for the legacy code).

Toflar: Loads of this stuff could be replaced by OPCodes. If it was added to the whole code base, I'm pretty sure it would be measurable.

leofeyer: `class_exists()` is not one of the compiler optimizable functions: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L355-L400

leofeyer (changes requested):
I find the current logic a little confusing:

```php
if (\is_object($strClass)) {
	// Use object
}
elseif (isset(static::$arrSingletons[$strClass])) {
	// Use object
}
elseif ($container->has($strClass) && (strpos($strClass, '\\') !== false || !class_exists($strClass))) {
	// Get service
}
elseif (($container->getParameter('kernel.debug') || !class_exists($strClass)) && self::isServiceInlined($strClass)) {
	// throw ServiceNotFoundException()
}
elseif (!class_exists($strClass)) {
	// throw RuntimeException('class not found')
}
```

Why don't we make the "class not found" case more straight-forward?

```php
if (\is_object($strClass)) {
	// Use object
}
elseif (isset(static::$arrSingletons[$strClass])) {
	// Use object
}
elseif ($container->has($strClass) && (strpos($strClass, '\\') !== false || !class_exists($strClass))) {
	// Get service
}
elseif (!class_exists($strClass)) {
	if (self::isServiceInlined($strClass)) {
		// throw ServiceNotFoundException()
	}

	// throw RuntimeException('class not found')
}
```

aschempp:
This would not consider the case when a class exists but the service was inlined. If I define a service for a class, I would always assume the service is used. That's why we currently check for that case first, but for performance reasons this PR changes it to only check in dev mode.

leofeyer:
> but for performance reasons this PR changes it to only check in dev mode.

It apparently does more than that, because the prior condition was `elseif ($container instanceof Container && isset($container->getRemovedIds()[$strClass]))` and had no `class_exists()` check. I do not really understand what the code does now. 🤷🏻‍♂️

leofeyer (approved):

PR #2341, opened 2020-09-24
Make the SimpleTokenParser backwards compatible

### LINKED ISSUE #2315: Simple tokens evaluation behaviour different between Contao 4.10 and <4.10
**Affected version(s)**
Contao 4.10

**Description**
There is an incompatibility in simple token evaluation in Contao 4.10 and earlier versions. When a simple token value is non-existent in prior versions this evaluated to NULL however starting from Contao 4.10 this will throw an exception. The root cause seems to be the switch to using Symfony Expression Language for simple token evaluation.
The problem for me popped up because I'm using a template in notification center for different forms. Because some input fields are only available in certain forms I used the following statements for dealing with the problem:
```
{if form_foo!=NULL}
Bar: ##form_foo##
{endif}
```
However this does not work anymore with Contao 4.10.

For Symfony Expression Language the issue is known for a long time and was reported in https://github.com/symfony/symfony/issues/21691. However it seems that resolution is not planned in a sensible timeframe. Community user webstar came up with the idea of resolving the incompatibility within Contao (see https://community.contao.org/de/newreply.php?do=newreply&p=531754).

Any changes resolving the problem will be greatly appreciated. Thanks

### LINKED ISSUE #2387: "gender" is not valid around position 1 - kein Newsletter Testversand
**Affected version(s)**

Contao 4.10.2

**Description**
{if gender=="male"}Sehr geehrter Herr ##lastname##,{elseif gender=="female"}Sehr geehrte Frau ##lastname##,{else}Liebe Mitglieder,{endif}

Variable "gender" is not valid around position 1 for expression `gender=="male"`

Im Newsletter Testversand, der reguläre Versand funktioniert

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2315, #2387 
| Docs PR or issue | -

To regain BC there now is an additional step in which the variables used in the expression are identified. With this we can find out if any variable isn't provided in the array of data and log + exit gracefully (= return `false`) instead of throwing an exception. This also allows logging all unknown tokens not only the first occurrence.

I also added a `bool` cast to the evaluated result to support shorthand/non-boolean expressions like `{if foo}`, `{if 1}`, `{if a + b}`.


m-vo:
Here are some samples where the execution differs - please feel free to add others (via PR, see `SimpleTokenParserTest#testDoesNotBreakBackwardsCompatibility`).

The following output shows the diff between legacy (= Expected) and new implementation (= Actual) in the categories 

- result after evaluating the subject
- logged messages during execution (log level | message)
- thrown exceptions during execution (exception type | message).


```diff
{if foo}match{endif}
['foo' => true]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
+    'result' => 'match'
     'log' => Array &1 ()
     'exception' => null
 )

 
{if foo}match{endif}
['foo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
+    'result' => null
     'log' => Array &1 ()
-    'exception' => null
+    'exception' => 'TypeError|Return value of Contao\CoreBundle\Util\SimpleTokenParser::evaluateExpression() must be of the type bool, int returned'
 )

 
{if foo}match{endif}
['foo' => '']

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
+    'result' => null
     'log' => Array &1 ()
-    'exception' => null
+    'exception' => 'TypeError|Return value of Contao\CoreBundle\Util\SimpleTokenParser::evaluateExpression() must be of the type bool, string returned'
 )

 
{if foo}match{endif}
['foo' => null]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
+    'result' => null
     'log' => Array &1 ()
-    'exception' => null
+    'exception' => 'TypeError|Return value of Contao\CoreBundle\Util\SimpleTokenParser::evaluateExpression() must be of the type bool, null returned'
 )

 
{if unknown}match{endif}
[]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
+    'result' => null
     'log' => Array &1 ()
-    'exception' => null
+    'exception' => 'InvalidArgumentException|Variable "unknown" is not valid around position 1 for expression `unknown`.'
 )

 
{if unknown==foo}match{endif}
['foo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
-    'log' => Array &1 (
-        0 => 'info|Tried to evaluate unknown simple token "unknown".'
-    )
-    'exception' => null
+    'result' => null
+    'log' => Array &1 ()
+    'exception' => 'InvalidArgumentException|Variable "unknown" is not valid around position 1 for expression `unknown==foo`.'
 )

 
{if foo==unknown}match{endif}
['foo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
     'result' => null
     'log' => Array &1 ()
-    'exception' => 'InvalidArgumentException|Unknown data type of comparison value "unknown".'
+    'exception' => 'InvalidArgumentException|Variable "unknown" is not valid around position 6 for expression `foo==unknown`.'
 )

 
{if foo==null}match{endif}
[]

--- Expected
+++ Actual
@@ @@
 Array &0 (
-    'result' => ''
-    'log' => Array &1 (
-        0 => 'info|Tried to evaluate unknown simple token "foo".'
-    )
-    'exception' => null
+    'result' => null
+    'log' => Array &1 ()
+    'exception' => 'InvalidArgumentException|Variable "foo" is not valid around position 1 for expression `foo==null`.'
 )


{if foo<=>1}match{endif}
['foo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
     'result' => null
     'log' => Array &1 ()
-    'exception' => 'InvalidArgumentException|Unknown simple token comparison operator "<=>".'
+    'exception' => 'InvalidArgumentException|Unexpected token "operator" of value ">" around position 6 for expression `foo<=>1`.'
 )

 
{if foo==1}match
['foo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
     'result' => null
     'log' => Array &1 ()
-    'exception' => 'Exception|Error parsing simple tokens'
+    'exception' => 'RuntimeException|Error parsing simple tokens'
 )

 
{if}match{endif}
['fo' => 1]

--- Expected
+++ Actual
@@ @@
 Array &0 (
     'result' => null
     'log' => Array &1 ()
-    'exception' => 'Exception|Error parsing simple tokens'
+    'exception' => 'RuntimeException|Error parsing simple tokens'
 )

 ```

File: core-bundle/tests/Fixtures/Util/LegacySimpleTokenParser.php [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
+use Psr\Log\LogLevel;
+
+class LegacySimpleTokenParser implements LoggerAwareInterface
+{
+    use LoggerAwareTrait;
+
+    /**
+     * This is a copy of the StringUtil::parseSimpleTokens implementation prior
```
Toflar: Can just use `System::setContainer()` no? I find it quite confusing to have this class in our testsuite :D

m-vo: I did not intend to keep this anyways. But it was an easy way to compare both implementations to find out what should be done.

m-vo: Ah and maybe the docbloc is misleading. This is the code from the 4.9 codebase (no EL) that differs quite a bit from the current one.

m-vo:
This is imho ready to be merged. See the updated description.

btw.: Before you're asking "Why an additional lexing step - can't we catch the syntax error and look at the token?" &rarr; This unfortunately does not work if we do not want to rely on pattern matching the exception message. The information is not stored in the exception. Also we would only see the first unmatched variable.

core-bundle/src/Util/SimpleTokenParser.php
```diff
@@ -137,25 +140,36 @@ private function canUseExpressionLanguage(array $data): bool
...
-            try {
-                return $this->expressionLanguage->evaluate($expression, $data);
-            } catch (\Exception $e) {
-                throw new \InvalidArgumentException($e->getMessage(), 0, $e);
-            }
+        if (!$canUseExpressionLanguage) {
+            return $this->evaluateExpressionLegacy($expression, $data);
+        }
+
+        $unmatchedVariables = array_diff($this->getVariables($expression), array_keys($data));
```
m-vo, line 147: We could consider making `getVariables()` a method of our `SimpleTokenExpressionLanguage` and (feature testing the usage with `method_exists`, falling back to `[]`) so that people creating their own EL with custom parsing could handle this individually.

leofeyer: > (feature testing the usage with `method_exists`,

Not that big a fan of that.

m-vo: Could also be an interface oc (but maybe overkill for such a niche topic).

leofeyer: Is it necessary at all?

m-vo: Only if we'd like to factor the `getVariables` method out of our `SimpleTokenParser` service. It could get necessary if one uses their own expression language that parses things differently. But again, pretty niche. 

I'd vote for leaving it as it is in this PR for now and considering the option if someone ever shows up with a problem.

leofeyer (dismissed):

core-bundle/src/Util/SimpleTokenParser.php
```diff
@@ -137,25 +140,36 @@ private function canUseExpressionLanguage(array $data): bool
...
+
+        // If the expression contains variables that aren't provided, log and exit gracefully
+        if (!empty($unmatchedVariables)) {
+            $this->logUnmatchedVariables(...$unmatchedVariables);
+
+            return false;
```
ausi: If we always return `false` if the condition contains unknown variables wouldn’t that flip the logic of inverted comparisons like `{if foo!=""}`?

m-vo: Does this work right now? (or should I say: < 4.10)

ausi: I think so because unknown values are treated as `NULL` I guess. But I didn’t test it yet.

ausi: Instead of returning `false` we could set `$data[$unmatchedVariable] = null;` for every value of `$unmatchedVariables` and afterwards evaluate the expression. WDYT?

m-vo: If so, we could just preset all missing vars with `null` and continue (instead of returning), I guess.

m-vo: Changed in ec3d58dc3f6e2834c1ed7df3f9e363b9a5dcd907.

ausi: It seems that we have to fix that in Contao 4.4/4.9 too: https://github.com/contao/contao/blob/2afe13cdc425cd4da774df205d6df4758f5e47ca/core-bundle/src/Resources/contao/library/Contao/StringUtil.php#L603

We introduced that bug in Contao 4.2.3: https://github.com/contao/core-bundle/commit/b68d28266bda83da2cf0cdf03df9d37c809b9464

m-vo: see https://github.com/contao/contao/issues/2402

core-bundle/src/Util/SimpleTokenParser.php
```diff
@@ -137,25 +140,39 @@ private function canUseExpressionLanguage(array $data): bool
...
 
         [, $token, $operator, $value] = $matches;
 
         if (!\array_key_exists($token, $data)) {
-            if (null !== $this->logger) {
-                $this->logger->log(LogLevel::INFO, sprintf('Tried to evaluate unknown simple token "%s".', $token));
-            }
+            $this->logUnmatchedVariables($token);
 
             return false;
```
ausi: We need to fix it here too, right?

m-vo: Done in d7995b689f23d4520289fd1c046293ac3b356b3b.

m-vo:
Just for the record, something I realized while implementing the latest changes and I don't quite like: Our logic that decides if we can use the EL or must fallback to the legacy version (`canUseExpressionLanguage()`) decides based on the provided data not based on the expression. 

This means if I've got an expression like `This is my {if val&#ue==5}match{endif}` and I'm *not* providing the token `['val&ue' => 5]` this will throw an `InvalidArgumentException` (because: valid behavior in new code path) but it will output `This is my match` of `This is my` if I do provide any "weirdly-named" key in my data (old code path).

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2342, opened 2020-09-24
Fix entering 0 in the back end

### LINKED ISSUE #2347: #2294 breaks several form field generator fields using value '0'
**Affected version(s)**

Contao 4.9.6 (likely 4.4 and 4.10 branches as well)

**Description**

In versions up to 4.9.5, it was possible to enter the integer `0` (which gets converted to `"0"`) as value in a select menu field of the form generator, like so:

Bildschirmfoto von 2020-09-24 17-22-40

I believe this may very well be a common use case, with any form field widget that allows options, like select, radio or checkbox menus.

With the change in #2294, entering the value `0` will be serialized as `""` (not as `"0"` like before) because of [this line](https://github.com/contao/contao/blame/4.9/core-bundle/src/Resources/contao/library/Contao/Widget.php#L813) having changed from `$value == ''` (which was `false` for `$value = "0"`) to `!$value`, which unfortunately evaluates to `true` for `$value = "0"`.

This subtle change has actually wreaked havoc on our install and caused some data loss. I could imagine this having adverse effects in several other more or less expected places.

### LINKED ISSUE #2352: Main article module is not generate
**Affected version(s)**

Contao 4.9.6?

Description

After updating Contao from 4.9.5 to 4.9.6, when i save a layout (tl_layout). The main article module is save with an empty ID, not "0" and isn't generate.

I try that in 3 differents projects.

leofeyer:
As discussed in Mumble on September, 24th.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_module.php
+++ b/core-bundle/src/Resources/contao/dca/tl_module.php
@@ -786,7 +786,7 @@ class tl_module extends Contao\Backend
 	 */
 	public function getActivationDefault($varValue)
 	{
-		if (!trim($varValue))
+		if (trim($varValue) === '')
 		{
 			$varValue = (is_array($GLOBALS['TL_LANG']['tl_module']['emailText']) ? $GLOBALS['TL_LANG']['tl_module']['emailText'][1] : $GLOBALS['TL_LANG']['tl_module']['emailText']);
 		}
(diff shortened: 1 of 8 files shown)


m-vo (approved):

ausi (approved):

fritzmg:
For anyone stumbling over this: this also fixes the article module not being generated in the front end any more, after you have saved your existing page layout once (or created a new one).

markusmilkereit:
When will this one be released? I vote for "as soon as possible". The commit was already in when we wanted to post the issue - especially the broken article modules in the layout are a big problem since the update. We had to fix a couple of layouts by hand via database in the updated Contao and had to stop actively working on a couple of websites because of it :/

ausi:
> When will this one be released?

Tomorrow.

PR #2350, opened 2020-09-24
Fix includeBlankOption with value 0 - issue 1022

### LINKED ISSUE #1022: includeBlankOption does not work at select if first value 0
**Affected version(s)**
C 4.4.43

**Description**  
its not possible to select `-` - the selected jump to `0`

**How to reproduce**  
see https://github.com/menatwork/contao-multicolumnwizard-bundle/issues/63

zonky2:
Fix includeBlankOption with value 0

implement https://github.com/contao/contao/issues/1022#issuecomment-575210705 with cast option values to string if values int - see https://github.com/menatwork/contao-multicolumnwizard-bundle/issues/63

| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1022

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -1153,7 +1153,7 @@ abstract class Widget extends Controller
 			return '';
 		}
 
-		return (\is_array($varValues) ? \in_array($strOption, $varValues) : $strOption == $varValues) ? ' selected' : '';
+		return \in_array((string) $strOption, array_map('strval', (array) $varValues), true) ? ' selected' : '';
 	}
 
 	/**


ausi:
Did you check if the fix does not have any side effects?

Did you check if this is really the only place where we need to adjust something? What about line 1174 for example?

zonky2:
> What about line 1174 for example?

checkboxes have no blankOption

> Did you check if this is really the only place where we need to adjust something?

I think so, I have tested the option array with integer and strings

ausi:
This should be a bugfix for Contao 4.4 I think. Please (re)base your pull request on the 4.4 branch.
(see https://github.com/contao/contao/blob/master/.github/CONTRIBUTING.md#pull-requests)

ausi:
> > What about line 1174 for example?
> 
> checkboxes have no blankOption

But radio buttons can have a blank option.

zonky2:
> But radio buttons can have a blank option.

good point - but I do not see the same error with radio as with select

zonky2:
pls reopen this issue - the fault still exists

fritzmg:
@zonky2 you need to address above comments. (Also for some reason I cannot reopen this.)

zonky2:
reopen for Contao 4.9
https://github.com/contao/contao/pull/3411

PR #2300, opened 2020-09-15
Do not use the default player size

fritzmg:
In the media player content element, Contao currently _always_ outputs a `width` and `height` attribute, even if no size is defined within the content element itself.

However, this prevents the element to be sized and resized responsively via CSS according to its actual size and aspect ratio. 

Thus I think we should simply remove these attributes. May be even as a bug fix in earlier versions?

PR #2351, opened 2020-09-24
Override the size variable for the ce_player template

fritzmg:
In https://github.com/contao/contao/pull/2300 we removed the old default player size for the media content element. However, since the template is, by default, always filled with all the fields from `tl_content`, the `$this->size` variable might then contain image size data (for text, image and gallery content elements). 

**Reproduction:**

1. Create a new `text` content element.
2. Enable _Add image_.
3. Save the content element.
4. Change the content element type to `Video/audio player`.
5. Select a video and save.

The front end will then output the following invalid HTML:

```html
<div class="ce_player first block">
  <figure class="video_container">
    <videoa:3:{i:0;s:0:"";i:1;s:0:"";i:2;s:0:"";} controls>
      <source type="video/mp4" src="files/foo/bar.mp4" title="bar.mp4">
    </video>
  </figure>
</div>
```

This PR fixes this by setting the `size` template variable to an empty string, if no size is given.

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentMedia.php
+++ b/core-bundle/src/Resources/contao/elements/ContentMedia.php
@@ -157,6 +157,7 @@ class ContentMedia extends ContentElement
 		}
 		else
 		{
+			// $this->size might contain image size data, therefore unset it (see #2351)
 			$this->Template->size = '';
 		}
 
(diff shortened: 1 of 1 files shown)


ausi (dismissed):

core-bundle/src/Resources/contao/elements/ContentMedia.php
```diff
@@ -155,6 +155,10 @@ protected function compile()
 		{
 			$this->Template->size = ' width="' . $size[0] . '" height="' . $size[1] . '"';
 		}
+		else
+		{
+			$this->Template->size = '';
```
leofeyer, line 161:
```suggestion
			// $this->size might contain image size data, therefore unset it (see #2351)
			$this->Template->size = '';
```

leofeyer (changes requested):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2353, opened 2020-09-24
Fix picker for providers with insert tags

### LINKED ISSUE #2346: Link-Address is not replaced correctly
**Affected version(s)**

Contao 4.10.1

**Description**

When selecting a new link-address for a Hyperlink content element (after a link-address was previously already set), the insert-tag is not set correctly and the link is not working:

**Example:**
Insert Tag before selecting a new page for the link-address: {{link_url::21}}
Insert Tag after selecting a new page for the link-address: 21,{{link_url::25}}
Expected insert tag after selecting a new page for the link-address: {{link_url::25}}

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2346
| Docs PR or issue | -

This fixes the picker for picker providers that return insert tags (see #2346). It uses the picker callback to transform the DCA values back to the appropriate picker values and also checks whether it is actually necessary to load the previous values - which it only should be for the `checkbox` picker type.

I've tested this solution with:

* single `fileTree` and `pageTree` widgets
* `fileTree` and `pageTree` widgets that allow multiple selections
* pickers that allow the selection of files, articles, news, events, pages (e.g. the `url` field of `tl_content`)

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -969,24 +969,24 @@ abstract class DataContainer extends \Backend
 	}
 
 	/**
-	 * Returns a data-picker-value attribute with the currently selected picker values (#1816).
-	 * 
+	 * Returns a data-picker-value attribute with the currently selected picker values (#1816)
+	 *
 	 * @return string
 	 */
 	protected function getPickerValueAttribute()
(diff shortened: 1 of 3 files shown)


leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2360, opened 2020-09-25
Correctly detect unknown options

### LINKED ISSUE #2349: 'Unknown Option' in several checkbox widgets
**Affected version(s)**

Contao 4.10.1?

**Description**

After updating Contao from 4.10 to 4.10.1 there is an 'Unknown Option' in several checkbox widgets like 'Convert image formats' in the image sizes or the 'News archives' in newslist module. They appeared without making any changes. Maybe a effect of #920 and #2215?

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2349
| Docs PR or issue | -

This fixes all the issues mentioned in #2349. However, there is still a problem that I could not yet solve:

This is because the value of the `ImageSize` widget can be an array with three items (e.g. `[300, 300, 'crop']`) and all of the values are added as unknown options. The `Widget` class does not know that only the third value needs to be checked.

@contao/developers Any idea how to fix this without a protected method that can be overwritten in each widget class?

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -1277,6 +1277,7 @@ abstract class Widget extends \Controller
 		$arrAttributes['description'] = $arrData['label'][1];
 		$arrAttributes['type'] = $arrData['inputType'];
 		$arrAttributes['dataContainer'] = $objDca;
+		$arrAttributes['value'] = \StringUtil::deserialize($varValue);
 
 		// Internet Explorer does not support onchange for checkboxes and radio buttons
 		if ($arrData['eval']['submitOnChange'])
@@ -1354,7 +1355,7 @@ abstract class Widget extends \Controller
 				$arrAttributes['options'][] = array('value'=>'', 'label'=>$strLabel);
 			}
 
-			$isKnownOption = false;
+			$unknown = (array) $arrAttributes['value'];
 
 			foreach ($arrData['options'] as $k=>$v)
 			{
@@ -1362,9 +1363,9 @@ abstract class Widget extends \Controller
 				{
 					$value = $blnIsAssociative ? $k : $v;
 
-					if ($varValue && $varValue == $value)
+					if (($i = array_search($value, $unknown)) !== false)
 					{
-						$isKnownOption = true;
+						unset($unknown[$i]);
 					}
 
 					$arrAttributes['options'][] = array('value'=>$value, 'label'=>($blnUseReference ? ((($ref = (\is_array($arrData['reference'][$v]) ? $arrData['reference'][$v][0] : $arrData['reference'][$v])) != false) ? $ref : $v) : $v));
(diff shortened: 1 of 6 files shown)


ausi:
> This is because the value of the `ImageSize` widget can be an array with three items

This applies to all widgets that have multiple fields I guess, like `InputUnit`.

Bildschirmfoto 2020-09-25 um 16 49 06

> Any idea how to fix this without a protected method that can be overwritten in each widget class?

I think thats the way to go. Something like `getOptionFromValue()` or so.

leofeyer:
> I think thats the way to go. Something like `getOptionFromValue()` or so.

Ok. I will adjust the PR accordingly then.

aschempp:
Maybe we should just fix this in the known widgets like select and checkbox? Otherwise we'll affect any widget other people have developed.

leofeyer:
I have updated the PR in 0272778d8e9426a92db8a24e3c8589d5bb0b0e2d. Now the unknown options are passed as DCA attributes and only the `SelectMenu` and `ImageSize` widgets handle them. Not sure if there are checkbox wizards with unknown options as well?

> Otherwise we'll affect any widget other people have developed.

But that also implies that everyone has to adjust their widgets if they want to support unknown options.

aschempp (commented):
I'm not sure this will work. Here's my chain of thoughts (for an image size widget):

1. Store the current value as `$unknown`: `[0 => '100', 1 => '100', 2 => 'center_center']`
2. For every `options`, unset key of value if in `$unknown`
3. If `$unknown` is not empty, add an empty option

Because `100` is never in the valid options, it would always result in an *Unknown Option* `<option>`?

leofeyer:
This would fail as soon as there is an image size with the ID `100`. Then `100` suddenly is a valid option.

leofeyer:
This PR is ready to merge now. The following widgets support unknown options:

* `Checkbox`
* `CheckboxWizard`
* `ImageSize`
* `RadioButton`
* `SelectMenu`

If it turns out that more widgets need to support this, we can add them later.

core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -1378,22 +1379,16 @@ public static function getAttributesFromDca($arrData, $strName, $varValue=null,
...
 			}
 
-			// If the value is not in the options array, the current user most
-			// likely cannot access it. We add the value as unknown option, so
-			// it does not get lost when saving the record (see #920).
-			if ($varValue && !$isKnownOption)
-			{
-				$arrAttributes['options'][] = array('value'=>$varValue, 'label'=>$GLOBALS['TL_LANG']['MSC']['unknownOption']);
-			}
+			$arrAttributes['unknownOption'] = array_filter($unknown);
```
ausi, line 1391: What is the purpose of `array_filter()` here?

leofeyer: It filters empty entries, which can never be valid options.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Widget.php
dmolineus: This would also strips values like `int 0` and `string '0'`. Is it supposed to be stripped even for arrays not being associative? See https://3v4l.org/RquTv

ausi (approved):

core-bundle/src/Resources/contao/library/Contao/Widget.php
leofeyer: Yes. `0` and `'0'` cannot be valid options in Contao 4.9. There is a PR to change this behavior in a future version, however, we fist have to make sure that there are no side-effects (see #2350).

Kahmoon:
I just installed 4.10.3 and one module gets type "unbekannte Option". All options from the module are still there but its not working in the frontend anymore. I´m pretty sure it worked on 4.10.2

<a href="https://ibb.co/G7GK8wY">Bildschirmfoto-2020-10-07-um-14-08-42</a>

Module listings are OK. The correct name/type is shown

<a href="https://ibb.co/sKb0c0z">Bildschirmfoto-2020-10-07-um-15-00-42</a>

https://github.com/markocupic/resource-booking-bundle/issues/24

Does this have anything to do with this issue here?

Frank

fritzmg:
Hm, the front end module is implemented as a fragment. I'll check whether there are any new issues with that in 4.10.3.

fritzmg:
https://github.com/contao/contao/issues/2415

PR #2361, opened 2020-09-25
Prevent using page aliases that could be page IDs

### LINKED ISSUE #1015: Prevent entering numeric page aliases
**Affected version(s)**

4.8

**Description**  

While fixing contao/contao#744, I noticed that it is possible to enter numeric aliases such as `2` or `2/foo` (with folder URLs enabled), even though these aliases will not work because of contao/core-bundle#729. Even worse, when using the alias `2/foo`, `2` will be in the routing candidates, potentially causing unexpected results.

**How to fix**  

IMHO the most reasonable fix would be to trigger an error if someone enters a numeric alias. @contao/developers WDYT?

### LINKED ISSUE #2306: Numeric alias in Symfony routing
> I've just come across this issue in the wild and it sent me down a bit of a rabbit hole.
> 
> 4.4.50 and 4.9.5 are also affected by the bug described in the initial message.
>
> I've tried to follow the decisions in this thread, but I'm confused as to why the fix suggested by @leofeyer in the initial message ("trigger an error if someone enters a numeric alias") was not adopted until Contao can work with numeric aliases. Is it because the issue was never technically confirmed for an LTS?
> 
> Side note: Interestingly, an alias like `1234e0` also causes the problem because it `is_numeric`. The way Contao checks for IDs vs aliases was fixed in d3839c398db358d029ea8ca0f3ec7308d25cf043, but the Symfony CMF Router integration that was developed in parallel was not adjusted accordingly, and so the check there is still `is_numeric`, e. g.:  https://github.com/contao/contao/blob/f1530fc6b67f8a2d104943d9d0397a8a23dd5616/core-bundle/src/Routing/RouteProvider.php#L475

_Originally posted by @gmpf in https://github.com/contao/contao/issues/1015#issuecomment-692046902_

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2306
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -1150,6 +1150,10 @@ class tl_page extends Contao\Backend
 				$varValue = $objPage->folderUrl . $varValue;
 			}
 		}
+		elseif (preg_match('/^[1-9]\d*$/', $varValue))
+		{
+			throw new Exception(sprintf($GLOBALS['TL_LANG']['ERR']['aliasNumeric'], $varValue));
+		}
 		elseif ($aliasExists($varValue))
 		{
 			throw new Exception(sprintf($GLOBALS['TL_LANG']['ERR']['aliasExists'], $varValue));
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -672,6 +672,10 @@ class tl_calendar_events extends Contao\Backend
 		{
 			$varValue = Contao\System::getContainer()->get('contao.slug')->generate($dc->activeRecord->title, Contao\CalendarModel::findByPk($dc->activeRecord->pid)->jumpTo, $aliasExists);
 		}
+		elseif (preg_match('/^[1-9]\d*$/', $varValue))
+		{
+			throw new Exception(sprintf($GLOBALS['TL_LANG']['ERR']['aliasNumeric'], $varValue));
+		}
 		elseif ($aliasExists($varValue))
 		{
 			throw new Exception(sprintf($GLOBALS['TL_LANG']['ERR']['aliasExists'], $varValue));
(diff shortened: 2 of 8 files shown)


core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1150,6 +1150,10 @@ static function ($alias) use ($objPage, $aliasExists)
 				$varValue = $objPage->folderUrl . $varValue;
 			}
 		}
+		elseif (preg_match('/^[1-9]\d*$/', $varValue))
+		{
+			throw new Exception(sprintf($GLOBALS['TL_LANG']['ERR']['aliasNumeric'], $varValue));
```
ausi, line 1155: I think we need to add a translation for the `aliasNumeric` key.

leofeyer: Added in b5243def42157a0c7aa1bffa8fd669a19af7f888. Is the check only needed for page aliases? What about news, events and FAQs?

ausi: > What about news, events and FAQs?

Yes, we should problably add the check there too.

aschempp (changes requested):
This is not the correct fix, because it would prevent using numeric aliases. We explicitly agreed on allowing that in https://github.com/contao/contao/issues/1015#issuecomment-567546050

ausi:
> we want to allow numeric aliases in the future

In the current verions we don’t allow them. In a future version we want to change that (that’s why the linked issue is assigned to the 4.11 milestone).

ausi:
To clarify:

In the current version it is possible to enter a numeric alias in the backend, but this page will then not be accessible in the frontend. This is very confusing for editors.

With this pull request we want to fix the backend side to show an error if someone enters a numeric alias in the backend.

In a future (minor) version we want to change that behavior to allow numeric aliases. But this change will require more work/time.

core-bundle/src/Resources/contao/dca/tl_page.php
leofeyer: Added in 1f9dc566781750b6e46aee7c8d2bbc694604cb8a.

ausi (approved):

aschempp:
> In the current version it is possible to enter a numeric alias in the backend, but this page will then not be accessible in the frontend.

You mean it wouldn't be accessible because we only search for database ID and not for the alias?

ausi:
> You mean it wouldn't be accessible because we only search for database ID and not for the alias?

Correct.

PR #2363, opened 2020-09-25
Reset the KEY_BLOCK_SIZE when migrating the MySQL engine and row format

aschempp:
I have a Contao 4.6 setup on a server that did not support InnoDB. Therefore, I configured the DBAL like so which worked for about 2 years now.

```yaml
doctrine:
    dbal:
        connections:
            default:
                charset: utf8
                default_table_options:
                    charset: utf8
                    collate: utf8_unicode_ci
                    engine: MyISAM
```

I could finally convince the client to move to a good hoster and also do an update to Contao 4.9. First step was to move the existing installation to the new hoster, on Contao 4.6 and with this DBAL configuration. Then I updated to Contao 4.9 and did run `contao:migrate`, which fails at migrating the `tl_files` table:

>ALTER TABLE tl_files ENGINE = InnoDB ROW_FORMAT = DYNAMIC......FAILED
>[ERROR] An exception occurred while executing 'ALTER TABLE tl_files ENGINE = InnoDB ROW_FORMAT = DYNAMIC':                                                                                                                                  
        SQLSTATE[HY000]: General error: 1031 Table storage engine for '#sql-10b9_2e20d0a' doesn't have this option    

After a suggestion from @ausi I tried to set `ROW_FORMAT=DEFAULT` before changing the engine. Changing the row format worked, but I got a new error after that (this time in german due to phpMyAdmin).

> error 1478 - Speicher-Engine 'InnoDB' der Tabelle unterstützt die Option 'KEY_BLOCK_SIZE' nicht

I did some digging on the internet and found the following to MySQL issues:
 - https://bugs.mysql.com/bug.php?id=88843
 - https://bugs.mysql.com/bug.php?id=56628

Apparently, my database table has a `KEY_BLOCK_SIZE` defined, but that's not supported by an DYNAMIC (uncompressed) row format. According to the second bug report, setting `KEY_BLOCK_SIZE=0` is the correct way to remove that setting since MySQL 5.5.9

### DIFF
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -254,7 +254,11 @@ class Installer
 
             if (strtolower($tableOptions->Engine) !== strtolower($engine)) {
                 if ($innodb && $dynamic) {
-                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0';
+                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+
+                    if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
+                        $command .= ', KEY_BLOCK_SIZE=0';
+                    }
                 } else {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine;
                 }
@@ -263,7 +267,12 @@ class Installer
                 $alterTables[md5($command)] = $command;
             } elseif ($innodb && $dynamic) {
                 if (false === stripos($tableOptions->Create_options, 'row_format=dynamic')) {
-                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0';
+                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+
+                    if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
+                        $command .= ', KEY_BLOCK_SIZE=0';
+                    }
+
                     $alterTables[md5($command)] = $command;
                 }
             }
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -254,7 +254,7 @@ class Installer
 
             if (strtolower($tableOptions->Engine) !== strtolower($engine)) {
                 if ($innodb && $dynamic) {
-                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0';
                 } else {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine;
                 }
@@ -263,7 +263,7 @@ class Installer
                 $alterTables[md5($command)] = $command;
             } elseif ($innodb && $dynamic) {
                 if (false === stripos($tableOptions->Create_options, 'row_format=dynamic')) {
-                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+                    $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0';
                     $alterTables[md5($command)] = $command;
                 }
             }
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -257,7 +257,7 @@ class Installer
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
 
                     if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
-                        $command .= ', KEY_BLOCK_SIZE = 0';
+                        $command .= ' KEY_BLOCK_SIZE = 0';
                     }
                 } else {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine;
@@ -270,7 +270,7 @@ class Installer
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
 
                     if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
-                        $command .= ', KEY_BLOCK_SIZE = 0';
+                        $command .= ' KEY_BLOCK_SIZE = 0';
                     }
 
                     $alterTables[md5($command)] = $command;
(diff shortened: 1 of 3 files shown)


ausi:
Is setting `KEY_BLOCK_SIZE` supported in MySQL 5.1 too, or would this result in an error?

aschempp:
According to https://bugs.mysql.com/bug.php?id=56628 it has been fixed 5.1.55 and 5.5.9. Before that fix, it did trigger a warning for invalid value, but the keyword seems valid.

m-vo:
> Before that fix, it did trigger a warning for invalid value, but the keyword seems valid

Is this also true when running in `strict` mode? And with MariaDB? :see_no_evil:

aschempp:
🤷 we'll never know…

ausi:
I think we should check for `false !== stripos($tableOptions->Create_options, 'key_block_size=')` and only then add `KEY_BLOCK_SIZE=0` to the query. This should still fix the issue but not create any warnings or errors.

leofeyer:
>> Is this also true when running in `strict` mode? And with MariaDB? 🙈

> 🤷 we'll never know… 

Then we cannot merge this PR, can we? We would risk breaking installations.

ausi:
> Then we cannot merge this PR, can we? We would risk breaking installations.

I think we won’t break anything if we check the create options of the table as I suggested in https://github.com/contao/contao/pull/2363#issuecomment-699909628

If the `KEY_BLOCK_SIZE` is already present in the create options, it should be save to unset it (by setting it to `0`).

leofeyer (changes requested):
> I think we should check for `false !== stripos($tableOptions->Create_options, 'key_block_size=')` and only then add `KEY_BLOCK_SIZE=0` to the query. This should still fix the issue but not create any warnings or errors.

See https://github.com/contao/contao/pull/2363#issuecomment-699909628

aschempp:
PR updated accordingly. I have tested this on my customer's database, which worked correctly. To adjust the unit tests, I had to rewrite the `getInstaller` to use the `$fromSchema` as _source of truth_ instead of varying the configuration based on table name. Also noticed that the `installation-bundle` was missing in the `phpunit.xml.dist`, therefore coverage was not generated for it.

leofeyer:
> Also noticed that the `installation-bundle` was missing in the `phpunit.xml.dist`, therefore coverage was not generated for it.

That is intended.

aschempp:
Can you explain why that would be intended? Except for *we're too lazy to write tests* 😝

leofeyer:
The bundle is deprecated and will hopefully be replaced by the Contao Manager soon, therefore we had decided not to write tests and not to track coverage.

aschempp:
that sounds fair. But doesn't that mean the installer should be moved to the core-bundle then?

leofeyer:
Probably yes. I think there is a ticket for that already.

installation-bundle/src/Database/Installer.php
```diff
@@ -255,6 +255,10 @@ private function checkEngineAndCollation(array &$sql, array &$order, Schema $fro
             if (strtolower($tableOptions->Engine) !== strtolower($engine)) {
                 if ($innodb && $dynamic) {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+
+                    if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
+                        $command .= ', KEY_BLOCK_SIZE=0';
```
ausi:
```suggestion
                        $command .= ' KEY_BLOCK_SIZE = 0';
```
We should use consistend formatting of the query I guess.

installation-bundle/src/Database/Installer.php
```diff
@@ -264,6 +268,11 @@ private function checkEngineAndCollation(array &$sql, array &$order, Schema $fro
             } elseif ($innodb && $dynamic) {
                 if (false === stripos($tableOptions->Create_options, 'row_format=dynamic')) {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+
+                    if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
+                        $command .= ', KEY_BLOCK_SIZE=0';
```
ausi:
```suggestion
                        $command .= ' KEY_BLOCK_SIZE = 0';
```

installation-bundle/tests/Database/InstallerTest.php
```diff
@@ -46,26 +49,96 @@ public function testReturnsTheAlterTableCommands(): void
...
+
+        $toSchema
+            ->getTable('tl_foo')
+            ->addIndex(['foo'], 'foo_idx')
+        ;
+
+        $installer = $this->getInstaller($fromSchema, $toSchema, ['tl_foo']);
+        $commands = $installer->getCommands();
+
+        $this->assertHasStatement($commands['ALTER_TABLE'], 'ALTER TABLE tl_foo ENGINE = InnoDB ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0');
```
ausi:
```suggestion
        $this->assertHasStatement($commands['ALTER_TABLE'], 'ALTER TABLE tl_foo ENGINE = InnoDB ROW_FORMAT = DYNAMIC KEY_BLOCK_SIZE = 0');
```

installation-bundle/tests/Database/InstallerTest.php
```diff
@@ -168,11 +241,43 @@ public function testChangesTheRowFormatIfInnodbIsUsed(): void
...
+        ;
+
+        $installer = $this->getInstaller($fromSchema, $toSchema, ['tl_foo']);
+        $commands = $installer->getCommands();
+
+        $this->assertArrayHasKey('ALTER_TABLE', $commands);
+
+        $this->assertHasStatement(
+            $commands['ALTER_TABLE'],
+            'ALTER TABLE tl_bar ENGINE = InnoDB ROW_FORMAT = DYNAMIC, KEY_BLOCK_SIZE=0'
```
ausi:
```suggestion
            'ALTER TABLE tl_bar ENGINE = InnoDB ROW_FORMAT = DYNAMIC KEY_BLOCK_SIZE = 0'
```

ausi (changes requested):

installation-bundle/src/Database/Installer.php
```diff
@@ -257,7 +257,7 @@ private function checkEngineAndCollation(array &$sql, array &$order, Schema $fro
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
 
                     if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
-                        $command .= ', KEY_BLOCK_SIZE=0';
+                        $command .= ', KEY_BLOCK_SIZE = 0';
```
ausi: I’d remove the comma too 🤓
```suggestion
                        $command .= ' KEY_BLOCK_SIZE = 0';
```

ausi (dismissed):

installation-bundle/src/Database/Installer.php
```diff
@@ -255,6 +255,10 @@ private function checkEngineAndCollation(array &$sql, array &$order, Schema $fro
             if (strtolower($tableOptions->Engine) !== strtolower($engine)) {
                 if ($innodb && $dynamic) {
                     $command = 'ALTER TABLE '.$tableName.' ENGINE = '.$engine.' ROW_FORMAT = DYNAMIC';
+
+                    if (false !== stripos($tableOptions->Create_options, 'key_block_size=')) {
```
leofeyer, line 259: Can we be sure that there is never a space between `key_block_size` and `=`?

ausi: I think so, yes. In my tests there was never a space. But we could also check for `/key_block_size\s*=/i` instead to be sure.

ausi: But we would have to adjust the `row_format=dynamic` check too then.

aschempp: I think so too, otherwise `row_format=dynamic` would fail as well. As we never had issues with that, I suggest to keep it that way 🤷

PR #2365, opened 2020-09-25
Fix non-normalized file extension comparisons when building lightboxes

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2358
| Docs PR or issue | -

PR #2366, opened 2020-09-26
Harden non-normalized file extension comparisons in the LegacyResizer

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

This is another place (see #2365) where we're using `Path::getExtension()` without the second argument to normalize the case.


ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2368, opened 2020-09-27
Add a createIfDeferred() method to the ImageResult class

### LINKED ISSUE #2322: Consider adding createIfDeferred to `ImageResult`
We should imo consider adding a `createIfDeferred` method (working like the one in `File`) to the `ImageResult` class as well, so that you can easily use the `FigureBuilder` concept for documents, emails or other occations where the result shouldn't be fetched from the server.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Closes #2322
| Docs PR or issue | Todo

This adds the ability to synchronously trigger the creation of resized images from within an `ImageResult`. This is similar to `File#createIfDeferred()` but processes all resources (picture img + picture sources) instead of a single file.

Contrary to the method in `File` an exception will be thrown if the `contao.image.resizer` service does not support deferred resizing (`DeferredResizerInterface`) and exceptions won't be swallowed.

**Demo Usage:**
```php
$figure = $studio->createFigureBuilder()
    ->fromPath('files/public/foo.jpg')
    ->setSize('_my_size')
    ->build();

$figure->getImage()->createIfDeferred();

// It's now safe to access any of the generated image files directly
$pdf->addImage($figure);
```


core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -142,6 +145,38 @@ public function getFilePath($absolute = false): string
...
+     * being generated by the Contao\CoreBundle\Controller\ImagesController
+     * on first access.
+     */
+    public function createIfDeferred(): void
+    {
+        $resizer = $this->resizer();
+
+        if (!$resizer instanceof DeferredResizerInterface) {
+            throw new \RuntimeException('The "contao.image.resizer" service does not support deferred resizing.');
+        }
```
ausi: I think it should only throw if `$deferredImages` is not empty. If all images are already resized throwing an exception here doesn’t have any benefit, right?

core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -142,6 +145,38 @@ public function getFilePath($absolute = false): string
...
+
+        $deferredImages = array_filter(
+            array_column(
+                array_merge([$picture->getImg()], $picture->getSources()),
+                'src'
+            ),
+            static function ($image): bool {
+                return $image instanceof DeferredImageInterface;
+            }
+        );
```
ausi, line 181: I think the images from `srcset` are missing here?

core-bundle/src/Image/Studio/ImageResult.php
m-vo: My reasoning was: now you'll get noticed about a wrong usage even if the files are already processed. Not throwing would kind of hide this in some cases and you would probably only stumble across the error in production. :thinking:  

But I'm not 100% sure how a setup would look if you both have 'deferred' images + a resizer that does not support them.

ausi: Having deferred images but not a deferred resizer would be a very special edge case. But if I want to disable deferred images in my setup completley (wich is a valid use case IMO) I still want `->createIfDeferred()` to work and not throw exceptions.

core-bundle/src/Image/Studio/ImageResult.php
m-vo: :scream: Indeed. Fixed in 71ffe3fc3a9ae59369f3f7dc08e128d7c2ff1e59.

core-bundle/src/Image/Studio/ImageResult.php
m-vo: That's a valid point. I adjusted the behavior in b73ba710ac2965c1a0137b765101e7c9c797ad5d so that an exception will only get thrown if there *are* deferred images present.

core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -142,6 +145,67 @@ public function getFilePath($absolute = false): string
...
+                }
+            );
+        };
+
+        // Create an array of deferred images by looking through the 'src' and
+        // 'srcset' attributes of both the 'img' and all 'sources' tags.
+        $deferredImages = array_merge(
+            $findDeferredImages($picture->getImg()),
+            $array_flatten(array_map($findDeferredImages, $picture->getSources()))
+        );
```
ausi:
```suggestion
        $deferredImages = [];
        
        foreach ([$picture->getImg(), ...$picture->getSources()] as $source) {
            $deferredImages[] = $source['src'];
            foreach ($source['srcset'] ?? [] as $srcset) {
            	$deferredImages[] = $srcset[0];
            }
        }
```

Would be shorter and more easy to understand?

core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -154,25 +154,54 @@ public function getFilePath($absolute = false): string
...
-            array_column(
-                array_merge([$picture->getImg()], $picture->getSources()),
-                'src'
-            ),
-            static function ($image): bool {
-                return $image instanceof DeferredImageInterface;
+        /** @var DeferredImageInterface $deferredImage */
+        foreach ($deferredImages as $deferredImage) {
+            // Skip already processed images
+            if (isset($resizedPaths[$deferredImage->getPath()])) {
```
ausi: I think we can skip this check. The deferred resizer is able to handle this case.

core-bundle/src/Image/Studio/ImageResult.php
m-vo: :smile: Guess so. Besides that `...$picture->getSources()` only works in PHP7.4. But that's easily adjusted.

Is it safe to assume that a `src` key exists and it's always the first element in `$srcset` that contains the image?

ausi: I just saw that I removed the `instanceof DeferredImageInterface` check 🙈
With the instanceof check every incorrect value would be skipped automatically, we could use `$source['src'] ?? null` and `$srcset[0] ?? null` to allow malformed array structures.

m-vo: Ah yes, just read your comment. That's exactly what I have done in 4f30e2f9f8764de8f1ace572695d3c1e8cdb05a5. :smiley:

core-bundle/src/Image/Studio/ImageResult.php
m-vo: see 4f30e2f9f8764de8f1ace572695d3c1e8cdb05a5

ausi (dismissed):

core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -142,6 +145,49 @@ public function getFilePath($absolute = false): string
...
+            static function ($image): bool {
+                return $image instanceof DeferredImageInterface;
+            }
+        );
+
+        if (empty($deferredImages)) {
+            return;
+        }
+
+        $resizer = $this->resizer();
```
Toflar: Shouldn't we check this first? If the resizer isn't available, we don't need to collect any `$deferredImages`.

m-vo: see https://github.com/contao/contao/pull/2368#discussion_r495817375

leofeyer (dismissed):

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2072, opened 2020-08-04
Deprecate the Controller::addImageToTemplate() method

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

This deprecates `Controller::addImageToTemplate` and rewires all usages. What I really like is that you can now see what the element's/module's figure is built from. Like: Is there a resize happening? From what data? Lightbox? Overwritten mtadata? ....

**Example**

before:
```php
// Add an image
if ($this->addImage && $this->singleSRC != '')
{
    $objModel = FilesModel::findByUuid($this->singleSRC);

    if ($objModel !== null && is_file(System::getContainer()->getParameter('kernel.project_dir') . '/' . $objModel->path))
    {
        $this->singleSRC = $objModel->path;
        $this->addImageToTemplate($this->Template, $this->arrData, null, null, $objModel);
    }
}
```

after:
```php
 // Add an image
if ($this->addImage && null !== ($figureBuilder = $this->getFigureBuilderIfResourceExists($this->singleSRC)))
{
    $figureBuilder
        ->setSize($this->size)
        ->setMetaData($this->objModel->getOverwriteMetaData())
        ->enableLightbox($this->fullsize)
        ->build()
        ->applyLegacyTemplateData($this->Template, $this->imagemargin, $this->floating);
}
```

This PR needs to be rebased once #2054 + #2060 are merged upstream and depends on #2071 for BC compatibility.

PR #2373, opened 2020-09-29
Use $this->imageHref in the image.html5 template

### LINKED ISSUE #2127: Module Events: List: Teaser pic (full view = true) opens lightbox
**Affected version(s)**

Contao 4.9.4

**Description**
pre conditions in backend:
An event with a picture in the teaser.
Full view = checked

Event module: Event list. Template: event_list.html

Frontend:
If you click on the picture it does not open the event nor the picture itself in the lightbox.
It opens the event full view in the lightbox.

Template:
```php
<?php if ($this->addImage): ?>
    <?php $this->insert('image', $this->arrData); ?>
<?php endif; ?>
```

The same snippet works fine in news module.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2127
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/templates/image/image.html5
+++ b/core-bundle/src/Resources/contao/templates/image/image.html5
@@ -1,12 +1,12 @@
 
 <figure class="image_container<?= $this->floatClass ?>"<?php if ($this->margin): ?> style="<?= $this->margin ?>"<?php endif; ?>>
-  <?php if ($this->href): ?>
-    <a href="<?= $this->href ?>"<?php if ($this->linkTitle): ?> title="<?= $this->linkTitle ?>"<?php endif; ?><?= $this->attributes ?>>
+  <?php if ($this->imageHref || $this->href): ?>
+    <a href="<?= $this->imageHref ?: $this->href ?>"<?php if ($this->linkTitle): ?> title="<?= $this->linkTitle ?>"<?php endif; ?><?= $this->attributes ?>>
   <?php endif; ?>
 
   <?php $this->insert('picture_default', $this->picture); ?>
 
-  <?php if ($this->href): ?>
+  <?php if ($this->imageHref || $this->href): ?>
     </a>
   <?php endif; ?>
 


m-vo:
Imo this must be fixed in the `ModuleEventlist` class not in the template. There should already be a similar handling for this case in `ModuleNews`, I assume, as I cannot reproduce the behavior of the original ticket for news. 

Btw.: The `figure.html.twig` template expects `Figure#getLinkHref` to be the correct link as well. We should make sure this can still work as a drop-in replacement.

Memo to myself: #2072 probably needs to be adjusted to also represent the behavioral change.

----

Edit: Just realized that `imageHref` is the special key that gets used if `href` is already defined, so maybe it's still correct to apply this change. But we should still check why events and news behave differently here.

leofeyer:
> Imo this must be fixed in the `ModuleEventlist` class not in the template. There should already be a similar handling for this case in `ModuleNews`, I assume, as I cannot reproduce the behavior of the original ticket for news.

I initially though so, too. However, the news list only works by chance, because there is no `href` parameter by the time the module calls `Controller::addImageToTemplate()`. If there was, it would not work in the news list, either.

https://github.com/contao/contao/blob/9d448e5d0f4aba6c8224a06ffc5ad4fd5ae93944/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1723-L1724

So if there is a `href` parameter (the URL of the news item or event), Contao will add the image URL as `imageHref`. Therefore, in the `image.html5` template, if there is an `imageHref`, we have to use it.

m-vo:
Ok, the fix is correct and also works as intended with the newer additions in 4.10 and #2072.

Just for the record: the reason why news and events behave differently is because the news templates do *not* use the property `href` but the event templates do. And in our current way we're adding images to templates everything gets squished into one namespace and therefore we've got logic in place that then sets `imageHref` instead of `href` which wasn't reflected in the image template. 

4.9, `Controller::addImageToTemplate`

https://github.com/contao/contao/blob/9d448e5d0f4aba6c8224a06ffc5ad4fd5ae93944/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1723-L1724

4.10, `Figure#applyLegacyTemplateData`

https://github.com/contao/contao/blob/860fdcbb1ac5ff9e4c2cd46f69603047f27ba7d9/core-bundle/src/Image/Studio/Figure.php#L352-L356

This does not affect the `figure.html` Twig template because this template will never receive the full flat template data but only its specific `Figure`. 

---

Looking into the future (with #2072 done) we could additionally ship the figure to the templates (`$objTemplate->figure = $figure`) and enhance the `Template#insert` method to allow rendering Twig in place, so that you could do something like this:
```diff
-    $this->insert('image', $this->arrData);
+    $this->insert('@ContaoCore/Image/Studio/figure.html.twig', $this->arrData);

or

-    $this->insert('image', $this->arrData);
+    $this->insert('@ContaoCore/Image/Studio/figure.html.twig', ['figure' => $this->figure]);
```

This won't solve that we currently still need to support the old way for BC but would allow a transition to a more sane solution.

m-vo (approved):

PR #2372, opened 2020-09-29
Remove existing tags from annotations that could  be arrays

aschempp:
Fixes https://github.com/contao/contao/issues/2371

Let's see what https://github.com/symfony/symfony/issues/38339 brings up …

PR #2378, opened 2020-09-29
Decode entities for page types and increase page type length

fritzmg:
If you define a page controller like so:

```php
// src/Controller/Page/ExamplePageController.php
namespace App\Controller\Page;

use Contao\CoreBundle\ServiceAnnotation\Page;

/**
 * @Page(ExamplePageController::class)
 */
class ExamplePageController
{
    public function __invoke(Request $request): Response
    {
    }
}
```
then the selection in the back end is not accepted, because it compares the value with the special characters of the FQCN converted to html entities (`App\Controller\Page\ExamplePageController` vs. `App&#92;Controller&#92;Page&#92;ExamplePageController`).

Imho this should be supported though, wdyt?

(_Note:_ you need to apply https://github.com/contao/contao/pull/2372 in order to test this.)


aschempp:
I'm not sure this is/should be supported or is a valid use case. We don't do that for fragments either. Just don't add a type at all and it will be compiled from the class name.

fritzmg:
> I'm not sure this is/should be supported or is a valid use case. We don't do that for fragments either.

I do think it would be nice for fragments as well ;). I use it for route names as well, since it creates a much better DX.

> Just don't add a type at all and it will be compiled from the class name.

That does not work unfortunately:

```
LogicException:
@Page annotation requires a type property.

  at c410dev\vendor\contao\contao\core-bundle\src\ServiceAnnotation\Page.php:76
  at Contao\CoreBundle\ServiceAnnotation\Page->__construct()
     (c410dev\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\DocParser.php:819)
  at Doctrine\Common\Annotations\DocParser->Annotation()
     (c410dev\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\DocParser.php:653)
  at Doctrine\Common\Annotations\DocParser->Annotations()
     (c410dev\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\DocParser.php:359)
  at Doctrine\Common\Annotations\DocParser->parse()
     (c410dev\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\AnnotationReader.php:221)
  at Doctrine\Common\Annotations\AnnotationReader->getClassAnnotations()
     (c410dev\vendor\doctrine\annotations\lib\Doctrine\Common\Annotations\CachedReader.php:80)
  at Doctrine\Common\Annotations\CachedReader->getClassAnnotations()
     (c410dev\vendor\terminal42\service-annotation-bundle\src\DependencyInjection\Compiler\ServiceAnnotationPass.php:70)
  at Terminal42\ServiceAnnotationBundle\DependencyInjection\Compiler\ServiceAnnotationPass->parseClassAnnotations()
     (c410dev\vendor\terminal42\service-annotation-bundle\src\DependencyInjection\Compiler\ServiceAnnotationPass.php:62)
  at Terminal42\ServiceAnnotationBundle\DependencyInjection\Compiler\ServiceAnnotationPass->process()
     (c410dev\vendor\symfony\dependency-injection\Compiler\Compiler.php:94)
  at Symfony\Component\DependencyInjection\Compiler\Compiler->compile()
     (c410dev\vendor\symfony\dependency-injection\ContainerBuilder.php:762)
  at Symfony\Component\DependencyInjection\ContainerBuilder->compile()
     (c410dev\vendor\symfony\http-kernel\Kernel.php:596)
  at Symfony\Component\HttpKernel\Kernel->initializeContainer()
     (c410dev\vendor\contao\contao\manager-bundle\src\HttpKernel\ContaoKernel.php:315)
  at Contao\ManagerBundle\HttpKernel\ContaoKernel->initializeContainer()
     (c410dev\vendor\symfony\http-kernel\Kernel.php:136)
  at Symfony\Component\HttpKernel\Kernel->boot()
     (c410dev\vendor\symfony\http-kernel\Kernel.php:196)
     (c410dev\web\index.php:31)
  at require('c410dev\web\index.php')
     (c410dev\web\app.php:4)              
```

aschempp:
hah! That's actually wrong, because the page type is generated from the class name same as for fragments (see https://github.com/contao/contao/blob/master/core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php#L135)

I've included that fix change in https://github.com/contao/contao/pull/2372

aschempp:
Regarding using the class name: I think that's a good idea, it's just not 100% backwards compatible (= use the class name automatically). We also need to check for the field length in the database, I think it's currently too short for class names.

fritzmg:
True, I haven't thought of that.

What I do instead for front end modules and content elements currently is to simply use a public constant (e.g. `TYPE`) in the class, so I can reference it in the annotation as well es everywhere else where I need it (DCA palette & translations for example). I can do the same for page controllers of course, but since I am already so used to using the `::class` name for a controller's route (ADR pattern) for example, I wanted to do it for the page as well ;)

aschempp (dismissed):
Just approving here because I don't see a reason to encode the page type (which is the problem in this PR). We should still keep it for discussion regarding use of class names.

leofeyer:
> We also need to check for the field length in the database, I think it's currently too short for class names.

What about this?

fritzmg:
The current limit is `64`. I have increased it to `255`. Although I wonder if `128` would be sufficient for the majority of cases. With `128` you could have a FQCN like this:

```
SomeFairlyLongFoobarVendorName\SomeFairlyLongFoobarBundleName\Controller\Page\SomeFairlyLongFoobarPageController
```

(with 15 characters to spare).

And if we just look at the core and assume that we might need subnamespaces witin `Controller\Page\`, then `128` would also be enough:

```
Contao\CoreBundle\Controller\Page\SomeFairlyLongFoobarSubnamespace\SomeFairlyLongFoobarPageController
```

If an extension requires more, they can always easily adjust the DCA.

fritzmg:
On the other hand, may be it does not make sense to limit it to `128`, since `tl_page` is less likely to reach the row limit, at least compared to `tl_content` or `tl_module` (which is already often a problem).

aschempp:
Whatever the limit, can we validate it in the compiler pass to make sure noone runs into it accidentially? I'd be fine with 128 (which we can still increase in the unlikely event someone reports it).

fritzmg:
I have included a check in the compiler pass now.

core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php
```diff
@@ -62,6 +64,10 @@ protected function registerPages(ContainerBuilder $container): void
                 $class = $definition->getClass();
                 $type = $this->getPageType($class, $attributes);
 
+                if (\strlen($type) > self::MAX_PAGE_TYPE_LENGTH) {
```
m-vo, line 67: Shouldn't this consider the actual length of the field instead of a constant? Now, there is no way for me to increase the  length (SQL definition) if I run into the error.

fritzmg: Hm, I don't know, can we evaluate this during the compiler pass?

aschempp: not really, you can't initialize the framework in that state. As I said before, I don't think that's necessary either. If someone **really** runs into that issue we can still increase it for everyone in Contao.

aschempp (approved):
I was a bit confused at first, maybe you should rename the PR title 😆

leofeyer:
As discussed in Mumble on October 22rd, we do not want to support this for controllers, because it does not work for fragments and would thus be inconsistent. Also, it only takes two minor adjustments to the DCA to implement the feature in your local app.

PR #2379, opened 2020-09-30
Add the image_container class to the default figure template

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

I figured we should probably add the `image_container` class to the core's default `figure.html.twig` template as well so that people can use it as a drop-in-replacement for the `image.html5` template. 

If someone should not want this class to be present in their installation, only a minor template adjustment is needed.


ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2392, opened 2020-10-02
Iterate over all events to find the event boundaries

### LINKED ISSUE #1818: Calendar-Module ignores event coming from the get getAllEvents-Hook for boundaries calculation
In
https://github.com/contao/calendar-bundle/blob/f18b9130ba4d02ef0c189ed5fb16ef8d2922c8ec/src/Resources/contao/modules/ModuleCalendar.php#L116
the module tries to find its boundaries. This will be used to determine if there is a next and/or prev button.
But this calculation only takes events from the `tl_events` table into account.

In my case I wrote an extension that will load additional events from an given URL and adds them through the `getAllEvents`-Hook.
This all works fine. They will be displayed in the listing and in the calendar module. 
But the next and prev button in the calendar won't be generated as there are no current events in the database.

I think a good solution would be to iterate over all `$this->arrEvents` and find the boundaries.

Another case is if one used the `getAllEvents`-Hook to filter some events, e.g. all events. Then the next button will be generated, but there is no event to be displayed.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1818
| Docs PR or issue | -

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
@@ -43,6 +43,11 @@ class ModuleCalendar extends \Events
 	protected $strTemplate = 'mod_calendar';
 
 	/**
+	 * @var array
+	 */
+	protected $arrAllEvents;
+
+	/**
 	 * Do not show the module if no calendar has been selected
 	 *
 	 * @return string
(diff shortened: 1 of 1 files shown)


leofeyer:
@michb @bennyborn Can you please check if this PR fixes your issues?

bennyborn:
@leofeyer Unfortunately this makes it even worse since the query before

https://github.com/contao/contao/pull/2392/commits/c3bad5dda2927b4e134b7538fc4b70180cfecbf8#diff-9b19d869d387401fdb99f912875f1f27L115

tried to find the min and max date of **all events** where as now

https://github.com/contao/contao/pull/2392/commits/c3bad5dda2927b4e134b7538fc4b70180cfecbf8#diff-9b19d869d387401fdb99f912875f1f27R118

you're limiting it to events between the first and last day of the given month which results in not being able to see any events prior or after the current month.

leofeyer:
You are right. Then I guess we have to find another solution, because calling `getAllEvents()` for **all** events in the database is obviously not very performant and should be avoided.

bennyborn:
What about adding a new hook? After finding the initial boundaries

https://github.com/contao/contao/blob/3a29fe75aa8e2d39557e914603bf71be78ef44d1/calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php#L114-L115

a `getEventBoundaries` Hook could get the min- and max-dates found by the query and increase them if necessary.

Regarding performance: Setting up a `ModuleEventlist` which would show all events at once would also result in `getAllEvents` being triggered without a limit

https://github.com/contao/contao/blob/3a29fe75aa8e2d39557e914603bf71be78ef44d1/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php#L151-L154

https://github.com/contao/contao/blob/df942f20f591afd95ee3f8b8310f22ffc4edf60d/calendar-bundle/src/Resources/contao/classes/Events.php#L517-L518

Another idea: Why check if there are any events in the first place? We could spare all the hassle if we choose to let the user freely browse previous and upcoming months regardless if there are any events at all.

Issue #2254, opened 2020-09-08
Automatically add absolute flag for external redirects


qzminski:
When you create a new page of external redirect type, choosing a page in picker generates a `{{link_url::ID}}` insert tag. This is wrong as then the generated URL is relative and the redirect will point in fact to a 404 page (i.e. visitor is taken to `https://domain.tld/$redirectPageAlias/$redirectUrlTarget`). This is easily fixed by making the URL absolute `{{link_url::ID|absolute}}`, yet regular users tend to forget to add that. Could this flag be added automatically somehow?

Maybe it would be possible with some `onload_callback` or field's `load_callback` in combination with https://github.com/contao/contao/pull/450? One concern is that not all picker providers or insert tags support `|absolute` flag.

Tested on Contao 4.9.5.

/cc @Toflar

fritzmg:
> This is wrong as then the generated URL is relative and the redirect will point in fact to a 404 page

Why does it point to a 404 page? Any relative URL is prepended by the `<base href>` by the browser.

leofeyer:
Is this the same problem as in #2171?

qzminski:
@fritzmg I forgot to mention, it happens when you access the page URL directly. Then the redirect response being thrown/returned contains an invalid URL.

fritzmg:
I see, makes sense now :)

ausi:
I think instead of the `|absolute` flag we should resolve all relative URLs (after replacing the insert tag) so that manually entered relative URLs (and other/custom insert tags) also get absolutized.

bennyborn:
Haven't been able to reproduce this in the online demo. Using the Inser-Tag `{{link_url::xx}}` as well as opening the page directly leads to the desired url...

leofeyer:
The online demo is still at version 4.9.5. 🙈

bennyborn:
> The online demo is still at version 4.9.5. 🙈

So? @qzminski  said he tested it on 4.9.5 ;)

leofeyer:
You are right. My bad.

aschempp:
I generally wonder why we should not always use absolute URLs? I can't imagine it to ever harm and it would solve millions of issues (including https://github.com/contao/contao/pull/2262) 🤷

leofeyer:
> always use absolute URLs

I am for it. @contao/developers Any objections?

Toflar:
[...]

PR #2395, opened 2020-10-03
Only replace file insert tags for meta data

### LINKED ISSUE #2394: Image link to the start page with alias index does not work
**Affected version(s)**

Contao 4.10
Contao 4.9 ist nicht betroffen
Contao 4.4 habe ich nicht getestet

**Description**
Setzt man ein Bildlink zur Startseite mit dem Alias "index" in der Dateiverwaltung oder bei "Metadaten überschreiben" in einer einsprachigen Contao-Installation (ohne Sprachkürzel) dann wird kein Link auf die Startseite gesetzt. Das Bild erscheint im Großformat. Der Link wurde über den Seitenwähler gesetzt. Ein Link zu jeder beliebigen Seite mit anderem Alias als "index" funktioniert.
Das Problem ist im Forum aufgetreten https://community.contao.org/de/showthread.php?79031-Link-auf-Startseite-mit-Alias-quot-index-quot-wird-nicht-erstellt#post532698 und auch bei einem Kunden der Contao Academy und konnte auch von @fritzmg reproduziert werden.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2394
| Docs PR or issue | -

The reason for the bug is that `{{link_url::*}}` will return an empty string for the `index` page (if no prefix is defined in the website root) - and thus the whole meta data array is considered empty and no link will be generated. In this PR I limit replacing the Insert Tags to `{{file::*}}`, since the comment above says "Make sure we resolve insert tags pointing to files".

However: what is the reason for replacing the insert tags here early in the first place? I'd just not replace any insert tags at this point at all.

/cc @m-vo

### DIFF
--- a/core-bundle/src/Resources/contao/models/FilesModel.php
+++ b/core-bundle/src/Resources/contao/models/FilesModel.php
@@ -423,7 +423,7 @@ class FilesModel extends Model
 			}
 
 			// Make sure we resolve insert tags pointing to files
-			if (isset($data[Metadata::VALUE_URL]))
+			if (isset($data[Metadata::VALUE_URL]) && 0 === stripos($data[Metadata::VALUE_URL], '{{file'))
 			{
 				$data[Metadata::VALUE_URL] = Controller::replaceInsertTags($data[Metadata::VALUE_URL]);
 			}


m-vo:
> However: what is the reason for replacing the insert tags here early in the first place? I'd just not replace any insert tags at this point at all.

The link url can serve as the source for the lightbox image. We need to have a path to be able to do things like compiling resized images. And you should be able to use/create any insert tag that outputs a path (or opts to output nothing = empty string).

It's just unfortunate that the `{{link_url::*}}` insert tag resolves to an empty string for some valid cases. This makes it indistinguishable from an empty input and therefore we fall back to using the main image resource in `FigureBuilder#onDefineLightboxResult()`. This behavior is imo correct.

Couldn't the `{{link_url::*}}` insert tag output `/` instead?

m-vo:
The insert tag could also output an absolute url to begin with or the picker add the absolute flag `{{link_url::123|absolute}}`.

Reminds me of #2254 ...

fritzmg:
Hm, but it still works in Contao 4.9, which presumably also needs to know the image path for the lightbox image. How is it handled there?

m-vo:
It's a bit different in Contao 4.9: there the lightbox image wasn't resized if the resource was taken from metadata (which was kind of a bug/limitation).

aschempp:
I feel like this should be fixed in https://github.com/contao/contao/blob/master/core-bundle/src/Resources/contao/models/PageModel.php#L1137 so the page does not return an empty string for a root URL?

ausi:
How about using `NULL` as the default for the meta fields so that an empty string can be correctly interpreted as a relative URL?

fritzmg:
> How about using `NULL` as the default for the meta fields so that an empty string can be correctly interpreted as a relative URL?

I am not sure that will work. If you edit a file in the file manager and save it, without entering any meta data, meta data entries for each visible language in the meta data wizard will still be generated, albeit with an empty string (see also #2404).

leofeyer:
As discussed in Mumble on October 22nd, the page model should use `./` instead of an empty URL (see https://github.com/contao/core/commit/c9eab397c7c0859a6c4e0210f291b3bfdc628f01). Once we got rid of the base tag and use absolute URLs everywhere, we can revert this again.

leofeyer:
See #2555 for the follow-up PR.

PR #2397, opened 2020-10-05
Check allowed page types

### LINKED ISSUE #1521: Mehrere Fehlerseiten 404 unter einem Startpunkt
**Affected version(s)**

Festgestellt in Contao 4.9

**Description**

Unter einem Startpunkt können mehrere Fehlerseiten 404 (Seitentyp "404 Seite nicht gefunden") angelegt werden. Es wird dann offenbar stets die erste Seite angezeigt.

So wie unter einem Startpunkt nicht mehrere Seiten mit demselben Alias angelegt werden können, sollte es jedoch auch nicht möglich sein, mehrere Fehlerseiten 404 unter einem Startpunkt anzulegen.

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1521

With this PR Contao only allows acceptable page type options in the respective place:

1. The root node can only contain `root` (already present)
2. `root` is not allowed with a parent page
3. `error_401`, `error_403` and `error_404` are only allowed as direct descendents of type `root`
4. `error_401`, `error_403` and `error_404` can only exist once per root page

This PR is against Contao 4.10, because it added the feature to limit the page type options. I don't think we should invest into fixing this in Contao 4.9, also because it might introduce a slightly different behavior for people.

I'm not 100% sure if the error pages can only exist once per root page? What about permissions or unpublished pages? Also, this is not 100%, it does not prevent copying the page type etc., but I don't think it should.

### DIFF
--- a/core-bundle/src/EventListener/FilterPageTypeListener.php
+++ b/core-bundle/src/EventListener/FilterPageTypeListener.php
@@ -59,10 +59,10 @@ class FilterPageTypeListener
             return;
         }
 
-        $siblingTypes = $this->connection->executeQuery(
-            'SELECT DISTINCT(type) FROM tl_page WHERE pid=?',
-            [$dc->activeRecord->pid]
-        )->fetchAll(FetchMode::COLUMN);
+        $siblingTypes = $this->connection
+            ->executeQuery('SELECT DISTINCT(type) FROM tl_page WHERE pid=?', [$dc->activeRecord->pid])
+            ->fetchAll(FetchMode::COLUMN)
+        ;
 
         foreach (array_intersect(['error_401', 'error_403', 'error_404'], $siblingTypes) as $type) {
             $event->removeOption($type);
--- a/core-bundle/src/EventListener/FilterPageTypeListener.php
+++ b/core-bundle/src/EventListener/FilterPageTypeListener.php
@@ -13,12 +13,24 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\EventListener;
 
 use Contao\CoreBundle\Event\FilterPageTypeEvent;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\FetchMode;
 
 /**
  * @internal
  */
 class FilterPageTypeListener
 {
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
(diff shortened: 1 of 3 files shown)


core-bundle/src/EventListener/FilterPageTypeListener.php
```diff
@@ -27,11 +39,33 @@ public function __invoke(FilterPageTypeEvent $event): void
...
+        // The first level can only have root pages (see #6360)
+        if (!$dc->activeRecord->pid) {
             $event->setOptions(['root']);
+
+            return;
+        }
+
+        $event->removeOption('root');
+
+        $parentType = $this->connection->fetchColumn('SELECT type FROM tl_page WHERE id=?', [$dc->activeRecord->pid]);
```
leofeyer, line 51: The method is deprecated in favor of the `fetchOne()` method. Did you check if we can use the new method?

core-bundle/src/EventListener/FilterPageTypeListener.php
```diff
@@ -27,11 +39,33 @@ public function __invoke(FilterPageTypeEvent $event): void
...
+            $event->removeOption('error_401');
+            $event->removeOption('error_403');
+            $event->removeOption('error_404');
+
+            return;
+        }
+
+        $siblingTypes = $this->connection
+            ->executeQuery('SELECT DISTINCT(type) FROM tl_page WHERE pid=?', [$dc->activeRecord->pid])
+            ->fetchAll(FetchMode::COLUMN)
```
leofeyer, line 64: The method is deprecated in favor of the `fetchFirstColumn()` method. Did you check if we can use the new method?

leofeyer:
> I'm not 100% sure if the error pages can only exist once per root page? What about permissions or unpublished pages?

Contao does not check user permissions when looking for an error page, therefore one error page per root should be fine.

> Also, this is not 100%, it does not prevent copying the page type etc., but I don't think it should.

I think it should. 🙈 @contao/developers WDYT?

aschempp:
I don't think it should for multiple reasons. I might want to make a copy to prepare a new error page, and unpublish the existing one after that. Hiding the options is a _hint_ on what you _should_ do, but same as you can copy a root page into a root page, this should be allowed as well.

leofeyer:
Hm, I kind of agree. 😄

leofeyer (approved):

Issue #1669, opened 2020-04-20
Set helpful values to autocomplete attribute in user context


Mynyx:
Instead of ``autocomplete="off"`` the following values for the ``autocomplete`` attribute could be set in forms of the user context (e.g. registration, login).

Credential | Autocomplete Values
-- | --
User Name | username
Password | current-password
New Password | new-password
One-Time Code | one-time-code

https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_an_html_input_element

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

dmolineus:
Related: https://github.com/contao/contao/issues/598

PR #2399, opened 2020-10-05
Remove the last username from the session after use

### LINKED ISSUE #1627: Failed login attempt in front end module login form starts session
**Affected version(s)**

Contao 4.9

**Description**

Even if a login attempt in the front end module ``login form`` fails, a session is started and the ``PHPSESSID`` cookie is set.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1627

See https://github.com/contao/contao/issues/1627#issuecomment-669228176

### DIFF
--- a/core-bundle/src/EventListener/ClearSessionDataListener.php
+++ b/core-bundle/src/EventListener/ClearSessionDataListener.php
@@ -13,7 +13,9 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\EventListener;
 
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBag;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\Security\Core\Security;
 
 /**
  * @internal
@@ -37,6 +39,10 @@ class ClearSessionDataListener
 
         if (!$request->hasSession() || !$request->getSession()->isStarted()) {
             return;
+        }
+
+        if ($event->getResponse()->isSuccessful()) {
+            $this->clearLoginData($request->getSession());
         }
 
         $this->clearLegacyAttributeBags('FE_DATA');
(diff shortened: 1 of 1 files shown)


leofeyer:
I think the purpose of the "last username" feature is that when you log out and later go back to the login module, your username is already filled in. IIRC, this no longer works after merging this PR, does it?

asaage:
> I think the purpose of the "last username" feature is that when you log out and later go back to the login module, your username is already filled in. IIRC, this no longer works after merging this PR, does it?

to mitigate that i'd recommend autocomplete attributes on the inputs anyway:
```html
<input type="text" name="username" id="username" class="text" value="" required="" autocomplete="username">
```
```html
<input type="password" name="password" id="password" class="text password" value="" required="" autocomplete="current-password">
```
This is something chrome always complains about and i added that to the login template.
Similar autocomplete values can be added to the personal-data module formfields.

ausi:
> I think the purpose of the "last username" feature is that when you log out and later go back to the login module, your username is already filled in.

This does not work currently and I don’t think it is expected to work as Symfony removes the “last username” from the session as soon as your login was successful. See https://github.com/contao/contao/issues/1627#issuecomment-669155322

As far as I understand this feature, the “last username” is stored in the session so that if you mistype your password you don’t have to reenter your username for the second attempt.

fritzmg:
> As far as I understand this feature, the “last username” is stored in the session so that if you mistype your password you don’t have to reenter your username for the second attempt.

That's what I thought as well 👍 I've never seen the other use-case work, nor would I find it useful. In fact I would consider that a security issue. If I log out of a website on a public machine, I don't want my username to linger around.

> to mitigate that i'd recommend autocomplete attributes on the inputs anyway:

Oh yes, and please let's do it for the Install Tool as well. Browsers always confuse the input fields of the create admin user form 🙃

leofeyer:
Thank you @ausi.

leofeyer:
> to mitigate that i'd recommend autocomplete attributes on the inputs anyway:

@asaage Will you create a PR for that?

asaage:
> @asaage Will you create a PR for that?

didn't know that there was already a ticket for this (https://github.com/contao/contao/issues/1669) 🤷‍♂️

leofeyer:
Never mind, I did it myself: #2432

Issue #317, opened 2019-02-06
Handling missing messages in the contao translator


dmolineus:
**Description**  

I've struggled working with the custom translation message domain named `contao_example`. I just wondered why the message wasn't neighter translated nor any debug entry in the symfony console was created for a missing translation message.

@ausi revealed that all message domains starting with `contao_` are mapped to the legacy translator reading translations from `$GLOBALS['TL_LANG']`. If the message doesn't exist no debug notice is created nor the translation is delegated to the default translator.

**Suggested change**

I suggest to change following code so that the decorated translator is able to handle the message:

https://github.com/contao/contao/blob/e79f4cb153e09f34c678e500b9031990b8bd81c5/core-bundle/src/Translation/Translator.php#L57-L59

* There's no need to add any debug information in the Contao translator as the decorated translator would do it
* It's possible to use `contao_` as message prefix for custom domains again

```diff
-            return $id;
+            return $this->translator->trans($id, $parameters, 'contao_' . $domain, $locale);
```

ausi:
I’m not sure if this is the best way to improve this. It could lead to even more confusing behavior if some messages come from `$GLOBALS['TL_LANG']` and others from the Symfony translator for the same message domain. This would be way harder to debug IMO.

It would probably be better to add debug information similar to the Symfony translator which also explains that the message was tried to load from `$GLOBALS['TL_LANG']`.

leofeyer:
As discussed in Mumble on February 14th, we want to add a debug message in line 58 with the same text as used in the Symfony translator.

leofeyer:
It seems we have to adjust the [`translator.data_collector`](https://github.com/symfony/symfony/blob/4.2/src/Symfony/Bundle/FrameworkBundle/Resources/config/translation_debug.xml) service so it passes our translator object instead of the default Symfony one. Not 100% sure though.

aschempp:
can you explain?

leofeyer:
Symfony uses a special data collector, which adds missing translations to the profiler. It currently does not handle Contao translations, which I assume is because it gets the default translator object injected instead of our decorated one.

aschempp:
see https://github.com/contao/contao/pull/544

PR #2401, opened 2020-10-06
Check if catalogue has translation

fritzmg:
When creating a new extension for Contao 4.9 and up, I preferably want to use Symfony translations in YAML format, as they are easier to handle. For aesthetic  reasons I want the translation domain to consist of the `contao_` prefix plus the extension name. So for example if my extension's name is `Contao Example` and the package name is `inspiredminds/contao-example` and the extension's namespace is `InspiredMinds\ContaoExample\` I want the translation domain to be `contao_example`, with the translation files under `src/Resources/translations/contao_example.en.yaml` within the package, which you can then override in your application with `translations/contao_example.en.yaml` etc.

However - this does not work currently as any translation with a translation domain beginning with `contao_` is forced to be taken from `$GLOBALS['TL_LANG']`. Thus I would be forced to change the translation name to something else, i.e. `inspiredminds_contao_example`.

The question is: should any translation domain beginning with `contao_` really be only reserved to the `$GLOBALS['TL_LANG']` translations? It can be fixed by checking whether the catalogue has an actual entry for the requested translation id and translation domain combination, as I did in this PR. However I am not sure about performance penalties.

_Note:_ can someone explain what the `DataCollectorTranslator` is used for? I couldn't figure it out, but I did the same change there anyway.

### DIFF
--- a/core-bundle/src/Translation/DataCollectorTranslator.php
+++ b/core-bundle/src/Translation/DataCollectorTranslator.php
@@ -47,9 +47,10 @@ class DataCollectorTranslator extends SymfonyDataCollectorTranslator implements 
     public function trans($id, array $parameters = [], $domain = null, $locale = null): string
     {
         $translated = $this->translator->trans($id, $parameters, $domain, $locale);
+        $catalogue = $this->translator->getCatalogue();
 
         // Forward to the default translator
-        if (null === $domain || 0 !== strncmp($domain, 'contao_', 7) || $this->translator->getCatalogue()->has($id, $domain)) {
+        if (null === $domain || 0 !== strncmp($domain, 'contao_', 7) || null !== $catalogue && $catalogue->has($id, $domain)) {
             return $translated;
         }
 
--- a/core-bundle/src/Translation/DataCollectorTranslator.php
+++ b/core-bundle/src/Translation/DataCollectorTranslator.php
@@ -49,7 +49,7 @@ class DataCollectorTranslator extends SymfonyDataCollectorTranslator implements 
         $translated = $this->translator->trans($id, $parameters, $domain, $locale);
 
         // Forward to the default translator
-        if (null === $domain || 0 !== strncmp($domain, 'contao_', 7)) {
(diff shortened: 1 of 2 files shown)


ausi:
>  can someone explain what the `DataCollectorTranslator` is used for?

For debug information in the Symfony toolbar, see https://github.com/contao/contao/issues/317#issuecomment-506650346

ausi:
> should any translation domain beginning with `contao_` really be only reserved to the `$GLOBALS['TL_LANG']` translations?

I think so, yes. Mixing translations from `$GLOBALS['TL_LANG']` with the translations from Symfony could be very confusing I think. Which translation source should have precedence?

>  For aesthetic reasons I want the translation domain to consist of the `contao_` prefix plus the extension name.

For extensions you should probably use the bundle alias as the message domain to comply with the [Symfony best practices](https://symfony.com/doc/current/bundles/best_practices.html#translation-files). Following the [Symfony bundle naming rules](https://symfony.com/doc/current/bundles/best_practices.html#bundle-name) this should never start with `contao_`.

fritzmg:
> For debug information in the Symfony toolbar, see [#317 (comment)](https://github.com/contao/contao/issues/317#issuecomment-506650346)

Thank you :)

> For extensions you should probably use the bundle alias as the message domain to comply with the [Symfony best practices](https://symfony.com/doc/current/bundles/best_practices.html#translation-files). Following the [Symfony bundle naming rules](https://symfony.com/doc/current/bundles/best_practices.html#bundle-name) this should never start with `contao_`.

Well the bundle alias can be changed too ;). From a UX perspective - for a _Contao_ user, I think it would make sense if you configure everything related to Contao, _and its extensions_, within `contao` or `contao_*`.

leofeyer:
> For extensions you should probably use the bundle alias as the message domain to comply with the [Symfony best practices](https://symfony.com/doc/current/bundles/best_practices.html#translation-files). Following the [Symfony bundle naming rules](https://symfony.com/doc/current/bundles/best_practices.html#bundle-name) this should never start with `contao_`.

I strongly agree with this!

PR #2404, opened 2020-10-06
Remove the "add language" menu from the meta wizard

leofeyer:
Now that available languages are added to the meta wizard by default, we no longer need the "add language" menu (see #2329).


fritzmg:
Sorry for not following the previous discussions - but are the languages with this now always limited to the website roots? Just today I had a use-case where I needed to provide Meta Data for multiple languages, where most of these languages do not exist as website roots. It seems odd to me to also have to define an inaccessible website root just for that.

leofeyer:
> where I needed to provide Meta Data for multiple languages, where most of these languages do not exist as website roots

That does not make much sense, does it? These translations will never be used if there is no matching root page. 🤷🏻‍♂️ 

> It seems odd to me to also have to define an inaccessible website root just for that

It is still correct. And it has a big advantage: If there is a root page, the language will be added automatically. Otherwise you will have to use the dropdown menu to add the language to every file, which is cumbersome.

fritzmg:
In our use case, Contao's file manager is used to store meta data for multiple languages which are then used in other systems of the customer via a simple API.

ausi:
Do I get this right? 

- For the backend, the configurable `contao.locales` parameter holds the list of available languages for the backend. (Defaults to all languages where a translation file exists)
- For the frontend, the list of all languages comes from the settings of all root pages.

If so, we should probably improve the description of the `locales` configuration from `Allows to configure which languages can be used within Contao.` to something like `Allows to configure which languages can be used within the backend of Contao.`

leofeyer:
Yes, you got that right. And +1 for adjusting the help text as suggested.

ausi (dismissed):

fritzmg:
Any previously added languages, that are not present as a website root, will remain visible in the back end within the wizard, correct?

leofeyer:
I think so, but you should try it.

fritzmg:
Ok, I'll quickly test, before you merge :)

fritzmg:
👍  previously entered meta data is not lost. However, since the delete button is now also gone, it is now impossible to delete any meta data entries for not needed languages.

leofeyer:
What would be the point of deleting a language? It would automatically be re-added anyway.

fritzmg:
> What would be the point of deleting a language? It would automatically be re-added anyway.

No, I meant in the scenario I described.

* Consider a pre Contao 4.11 instance, with a single website root with the language `en`.
* Within the file manager, meta data in multiple languages are maintained, by manually adding them as needed for individual files.
* Then, after an update to Contao 4.11, the pre-existing language meta data will still be there, albeit no new languages can be added, without also adding a respective website root.
* However, since the delete button is gone, unneeded meta data of languages, that do not exist as a website root cannot be deleted.

fritzmg:
Or consider the following reproduction (with your PR):

1. Add a new website root, with the language `fr` for example.
2. Go into the file manager and edit the meta data of a file that does not have any pre-existing meta data yet.
3. Without filling out anything, click "Save and close".
4. Delete the `fr` website root again.
5. Go into the file manager and edit the same file from before.

Even though no meta data was entered before, the `fr` meta data input fields remain in the meta data wizard for this file - but not for the other files. It's nothing breaking, but still odd, as you now have no possibility of removing such empty or possibly superfluous entries.

leofeyer:
I have re-added the delete functionality in 1f43f131215f5024625662afe6ec763824e70e02.

ausi (approved):

PR #2416, opened 2020-10-08
Remove the Content-Length header when modifying the response

aschempp:
After searching for 2.5hrs, I tracked a network error in my browser down to this. The `CsrfTokenCookieSubscriber` modifies the response content. If the response already has a `Content-Length` header, the value will be incorrect and browser cannot correctly handle them (especially in HTTP2 fashion).

While debugging this with @leofeyer we also digged why Symfony does not set/fix that header automatically. Apparently there was some code in Symfony 2.0 (added in https://github.com/symfony/symfony/pull/1551) which caused some issues, and it was removed again in https://github.com/symfony/symfony/issues/1846

In my specific case (Isotope eCommerce), the issue was actually caused by https://github.com/codefog/contao-haste/blob/master/library/Haste/Http/Response/Response.php#L206

### DIFF
--- a/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php
+++ b/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php
@@ -139,13 +139,18 @@ class CsrfTokenCookieSubscriber implements EventSubscriberInterface
         }
 
         $content = $response->getContent();
+        $tokens = $this->tokenStorage->getUsedTokens();
 
-        foreach ($this->tokenStorage->getUsedTokens() as $value) {
-            $content = str_replace($value, '', $content);
+        if (!\is_string($content) || empty($tokens)) {
+            return;
         }
 
-        $response->setContent($content);
-        $response->headers->remove('Content-Length');
+        $content = str_replace($tokens, '', $content, $replacedCount);
+
+        if ($replacedCount > 0) {
+            $response->setContent($content);
+            $response->headers->remove('Content-Length');
+        }
     }
 
     private function removeCookies(Request $request, Response $response): void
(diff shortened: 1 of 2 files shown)


core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php
```diff
@@ -145,6 +145,7 @@ private function replaceTokenOccurrences(Response $response): void
         }
 
         $response->setContent($content);
+        $response->headers->remove('Content-Length');
```
m-vo: Might want to add a comment explaining why this is done?

ausi:
We might also want to watch out for other type of responses like streamed or binary file? They return `false` for `getContent()`.

```php
$content = $response->getContent();

if (!\is_string($content)) {
	return;
}
```

And only modify the headers if tokes actually got replaced?

```php
$content = str_replace($this->tokenStorage->getUsedTokens(), '', $content, $replacedCount);

if ($replacedCount === 0) {
	return;
}
```

aschempp:
PR updated 😊

ausi (dismissed):

Toflar (changes requested):
Why removing the header instead of setting it to the correct length?

aschempp:
Because we shouldn't - according to the Symfony issues linked.

leofeyer (dismissed):

core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php
leofeyer: Added in a8a07b6cc3c72549e353cc14e2958599bf6e6479.

leofeyer (dismissed):

leofeyer (approved):

PR #2417, opened 2020-10-08
Register globals in the fragments pass

### LINKED ISSUE #2415: Content element and front end module fragments not shown in select options
**Affected version(s)**

Contao 4.10.3

**Description**

I can confirm [this report](https://github.com/contao/contao/pull/2360#issuecomment-704910280). Content elements and front end modules implemented as fragments do not show up in the type selection anymore.

e.g.:

```php
// src/Controller/ContentElement/ExampleElementController.php

use Contao\ContentModel;
use Contao\Template;

/**
 * @ContentElement(category="texts")
 */
class ExampleElementController extends AbstractContentElementController
{
    {
    }
}
```

```php
// src/Controller/FrontendModule/ExampleModuleController.php
namespace App\Controller\FrontendModule;

use Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController;
use Contao\CoreBundle\ServiceAnnotation\FrontendModule;
use Contao\ModuleModel;
use Contao\Template;

/**
 * @FrontendModule(category="miscellaneous")
 */
class ExampleModuleController extends AbstractFrontendModuleController
{
    protected function getResponse(Template $template, ModuleModel $model, Request $request): ?Response
    {
    }
}
```

Switching back to Contao 4.10.2 shows `example_element`/`example_module` as an available option in the respective categories.

aschempp:
This fixes the issue in https://github.com/contao/contao/issues/2415
The changes are not _necessary_ for Contao 4.9, but I would merge it here for consistency. I don't know why we had two separate compiler passes for the same thing… 

additionally the `RegisterFragmentsPass` is _meant_ to be used by third-party developers to register additional fragment types. This now allows to register them in the globals as well.


core-bundle/tests/DependencyInjection/Compiler/RegisterFragmentsPassTest.php
```diff
@@ -177,6 +180,72 @@ public function testFailsIfThereIsNoPreHandlersDefinition(): void
...
+            ],
+            $definition->getArguments()[0]
+        );
+
+        $this->assertTrue($definition->isPublic());
+    }
+
+    public function testFailsToRegisterGlobalsMapListenerIfCategoryIsMissing(): void
+    {
+        $this->expectException(InvalidConfigurationException::class);
```
leofeyer: You should probably move this to line 246, so it becomes clear where exactly we expect the exception to be thrown. This will also ensure that we are not accidentally catching an earlier exception (although this is very unlikely in this specific case).

leofeyer (dismissed):

bytehead (dismissed):

aschempp:
@Toflar and I also noticed that child definitions do not inherit tags from parent. Because of this, tags added by e.g. autoconfiguration (`registerForAutoconfiguration`) for services – which includes the `ServiceSubscriberInterface` for our controllers by default – are not inherited to the child definition. I've fixed this in 0099dd2c2894a33772779ffd10121f2448a0e96a (@Toflar please confirm).

aschempp:
As noticed by @Toflar, be aware that https://github.com/contao/contao/commit/0099dd2c2894a33772779ffd10121f2448a0e96a fixed an actual bug in Contao 4.9, while the rest of the PR just adds "forward compatibility" and a fix for 4.10!

Toflar (changes requested):
I can confirm the changes fix the bug. But I think we should add a test to the `RegisterFragmentsPassTest` to make sure, tags are kept and prevent regression bugs.

Toflar (approved):

bytehead (approved):

PR #2419, opened 2020-10-10
Fix unknown option when creating or editing error pages

fritzmg:
This fixes a bug introduced in #2397. If you create a 401, 403 or 404 page, it will be shown as an "unknown option" in the type selection.

This is because the `FilterPageTypeListener` is supposed to remove any error pages that are already present within the same website root. However, it also removes the page type of the page being currently edited.

This PR adds an additional query condition to prevent that.


aschempp (approved):
This will not re-add e.g. the error404 type if the current level is not "one below root" but I think that's correct?

leofeyer:
What exactly do you mean?

leofeyer:
Thank you @fritzmg.

PR #2420, opened 2020-10-12
Fix issues found by the PhpStorm code inspector

leofeyer:


core-bundle/src/Util/SimpleTokenParser.php
```diff
@@ -254,7 +254,7 @@ private function getVariables(string $expression): array
 
         $variables = [];
 
-        for ($i = 0; $i < \count($tokens); ++$i) {
+        for ($i = 0, $c = \count($tokens); $i < $c; ++$i) {
```
m-vo, line 257: Personally I do not like these kind of optimizations. It's shaving off microseconds at best but makes the code less readable.

Imo these are optimizations that should be done at a (pre)compile step and in the future probably will be.

fritzmg: It also seems odd to have to do this. In other languages, stuff like that is optimized by the compiler. Why hasn't PHP optimised this already? 🤔

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -97,7 +97,7 @@ protected function compile()
 		$this->Template->results = '';
 
 		// Execute the search if there are keywords
-		if ((string) $strKeywords !== '' && $strKeywords != '*' && !$this->jumpTo)
+		if ($strKeywords !== '' && $strKeywords != '*' && !$this->jumpTo)
```
m-vo, line 100: If `$strKeywords` is always a string (and it should be because `trim` always outputs a string) - shouldn't the second comparison also be strict?

core-bundle/src/Util/SimpleTokenParser.php
leofeyer: > that should be done at a (pre)compile step and in the future probably will be

Yeah, but until then, it does not hurt to do it. Also, I find it more comprehensive this way. 🤷🏻‍♂️

m-vo: Well, besides your :eyes: when reading it. :smile: 

> Why hasn't PHP optimised this already? 

I assume the introduction of a JIT compiler with PHP8 will open up possibilities here...

ausi: The readability problem I see here is, that I have to look at the whole foreach now to see if `$c` gets modified somewhere.

ausi: And `$c` should probably be named `$tokenCount`?

leofeyer: I do not really understand the discussion. It seems to be quite a common pattern:

* [nikic/PHP-Parser: AttributeEmulator](https://github.com/nikic/PHP-Parser/blob/d520bc9e1d6203c35a1ba20675b79a051c821a9e/lib/PhpParser/Lexer/TokenEmulator/AttributeEmulator.php#L24)
* [FriendsOfPHP/PHP-CS-Fixer: NoTrailingCommaInSinglelineArrayFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3887f77656c953af697c2762e404f07e1875fe64/src/Fixer/ArrayNotation/NoTrailingCommaInSinglelineArrayFixer.php#L54)
* [vimeo/psalm: ScopeAnalyzer](https://github.com/vimeo/psalm/blob/bb7d7132f790a24d34aa8420620b4b53b0db423e/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php#L92)

And we already use it as well:

https://github.com/contao/contao/blob/9eb88f5de119c4533dadac6ac791cba9b900ca6f/core-bundle/src/Routing/Enhancer/InputEnhancer.php#L77

core-bundle/src/Resources/contao/modules/ModuleSearch.php
leofeyer: No. The first comparison is meant to be type safe, the second one is not. Of course, we know that `$strKeywords` is a string, so `!==` would work as well. But since it does not matter, there is no point in changing it. 🤷🏻‍♂️

core-bundle/src/Util/SimpleTokenParser.php
ausi: I don’t like single character variables. For me `$i` is one of the few exceptions (but I still try to not use it that often).

How about using a foreach instead?

```suggestion
        foreach ($tokens as $i => $token) {
```

But I’m also OK if this gets merged as it is now ☺️

m-vo: No, it must be a linear loop here, because you wouldn't be able to look ahead + skip (see L270).

PR #2425, opened 2020-10-13
Correctly generate the HTML module

### LINKED ISSUE #2424: ModuleHtml is not working since Contao 4.9.6
**Affected version(s)**

Contao 4.9.6 and higher

**Description**
Since the changes in 4.9.6 to use the ```contao.routing.scope_matcher``` the ModuleHtml is not working.
The ModuleHtml uses return statements inside the compile() function instead of assigning the html to the Template object:

https://github.com/contao/contao/commit/59d455b76fb8d91bcb5a0b4cbabcc24e9f8efe60#diff-48c3a8a044e2f16f1981b99bf39b1793e0ee48f047593ead1b4afbd198ce2a64

The ContentHtml uses the assigning to the Template object:

https://github.com/contao/contao/commit/59d455b76fb8d91bcb5a0b4cbabcc24e9f8efe60#diff-7d54af17e8850b869776f5769b59dc212126d99432e7994fc370d38303ee70ba

If I change the Module to use the Template object it works fine for me.

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/2424

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleHtml.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleHtml.php
@@ -24,18 +24,29 @@ class ModuleHtml extends Module
 	protected $strTemplate = 'mod_html';
 
 	/**
-	 * Generate the module
+	 * Parse the template
+	 *
+	 * @return string
 	 */
-	protected function compile()
+	public function generate()
 	{
+		$strBuffer = parent::generate();
 		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 
 		if ($request && System::getContainer()->get('contao.routing.scope_matcher')->isBackendRequest($request))
 		{
-			return htmlspecialchars($this->html);
+			return htmlspecialchars($strBuffer);
 		}
 
-		return $this->html;
+		return $strBuffer;
+	}
+
+	/**
+	 * Generate the module
+	 */
+	protected function compile()
+	{
+		$this->Template->html = $this->html;
 	}
 }
 


Toflar (approved):

bytehead (approved):

aschempp:
I wonder if we should change `ContentHtml` as well, because it will never show the correct template content.

leofeyer (changes requested):
I don't think that this is the correct fix. We do not need a `generate()` method; all we need to do is to assign the value instead of returning it. Basically the same code as in the `ContentHtml` class.

https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/elements/ContentHtml.php

leofeyer:
Also, according to @fritzmg, the original issue was not reproducible (see https://github.com/contao/contao/issues/2424#issuecomment-707764187). So which issue are you fixing here?

ausi:
> all we need to do is to assign the value instead of returning it.

I agree. This was changed recently in https://github.com/contao/contao/commit/59d455b76fb8d91bcb5a0b4cbabcc24e9f8efe60#diff-48c3a8a044e2f16f1981b99bf39b1793e0ee48f047593ead1b4afbd198ce2a64 probably by mistake?

dmolineus:
> Also, according to @fritzmg, the original issue was not reproducible (see #2424 (comment)). So which issue are you fixing here?

The issue only occured in the Contao backend where no escaping happened anymore (module included as content element) I guess @fritzmg only checked the frontend.

> We do not need a generate() method; all we need to do is to assign the value instead of returning it. 

The change of @aschempp would also include the generated html of the template. So I guess this would be an improvement of the current behaviour.

leofeyer:
What exactly do you mean?

aschempp:
The current behaviour is to escape the **content of the HTML field**. What we DO want to escape is the **output of the module**. That's not the same thing. And the `compile()` method can never return a value.

leofeyer:
I do not understand why we need to improve/change anything in a bugfix release. We simply want the module to behave like it did before, therefore we should merge #2430 instead of this PR IMHO.

PR #2427, opened 2020-10-14
Use binary comparison when looking for aliases in the database

### LINKED ISSUE #2319: Trailing spaces in news URL results in unexpected behavior
**Affected version(s)**

Contao 4.4 and Contao 4.9?

**Description**

When a trailing space or %20 is added to the end of an alias in the URL, the original news page is still loaded and shown instead of a 404 page.
This can result in duplicate content errors in e.g. Google search console as it is a different URL but with the exactly same content.

https://demo.contao.org/en/news-detail/new-contao-versions-in-short-intervalls.html
is the same as https://demo.contao.org/en/news-detail/new-contao-versions-in-short-intervalls%20.html
https://contao.org/de/news/rueckblick-auf-das-zweite-core-entwicklertreffen-2020.html
is the same as https://contao.org/de/news/rueckblick-auf-das-zweite-core-entwicklertreffen-2020%20.html
This only occurs on news pages, article pages are not affected by this.
https://demo.contao.org/en/content-elements.html works
https://demo.contao.org/en/content-elements%20.html does not work / 404 page not found

As far as I've found, it's because of how MySQL compares strings with the equal operator (https://blogs.agilefaqs.com/2011/06/22/trailing-white-spaces-are-ignored-by-mysql-for-comparison-in-a-select-statement/).
`Select 'test' = 'test ';` returns true
`Select 'test' like 'test ';`returns false

Changing from = to like in news-bundle/src/Resources/contao/models/NewsModel:190 solves the problem.
But maybe a validation check for the alias before `$objArticle = NewsModel::findPublishedByParentAndIdOrAlias(Input::get('items'), $this->news_archives);` (news-bundle/src/Resources/contao/modules/ModulesNewsReader:91) is more suitable?

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2319
| Docs PR or issue | -

### DIFF
--- a/calendar-bundle/src/Resources/contao/models/CalendarEventsModel.php
+++ b/calendar-bundle/src/Resources/contao/models/CalendarEventsModel.php
@@ -201,7 +201,7 @@ class CalendarEventsModel extends \Model
 		}
 
 		$t = static::$strTable;
-		$arrColumns = !is_numeric($varId) ? array("$t.alias=?") : array("$t.id=?");
+		$arrColumns = !is_numeric($varId) ? array("BINARY $t.alias=?") : array("$t.id=?");
 		$arrColumns[] = "$t.pid IN(" . implode(',', array_map('\intval', $arrPids)) . ")";
 
 		if (!static::isPreviewMode($arrOptions))
--- a/core-bundle/src/Resources/contao/library/Contao/Model.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Model.php
@@ -812,7 +812,7 @@ abstract class Model
 			array
 			(
 				'limit'  => 1,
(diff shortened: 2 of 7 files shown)


aschempp:
What's the test case to check if this solves the problem?

ausi:
Open a page and add `%20` before the `.html`, e.g. `example.com/foo%20.html`

ausi (approved):

aschempp:
> Open a page and add `%20` before the `.html`, e.g. `example.com/foo%20.html`

FYI that didn't cause any problem on demo.contao.org. But I could reproduce it on a news item.

ausi:
In Contao 4.4 the page alias was affected too, in Contao 4.9 it seems to only affected non-page aliases like news.

PR #2431, opened 2020-10-15
Remove the Google web fonts field from the page layout

### LINKED ISSUE #1791: Google-Fonts API Version 2 Handling
As mentioned in the [forum](https://community.contao.org/de/showthread.php?78134-Google-Fonts-API-Base-URL&p=526398&viewfull=1#post526398):

By using the suggested Google-Fonts embed parameters for example:
https://fonts.google.com/specimen/Oswald?sidebar.open&vfonly&query=Oswald&selection.family=Oswald:wght@200;380;400

within the Pagelayout/Webfonts settings we get unexpected results. This is because of the Google Fonts APIv2 - see: https://developers.google.com/fonts/docs/css2

So maybe this should be mentioned in the pagelayout help text. Difficult to describe ...
Or switch to the new API Version?
Or leaving a hint in the docs? 
...

leofeyer:
| Q                | A
| -----------------| ---
| Implemented feature     | #1791
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
+++ b/core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Migration\Version411;
+
+use Contao\CoreBundle\Migration\AbstractMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @internal
+ */
+class MergeWebfontsMigration extends AbstractMigration
+{
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+
+        if (!$schemaManager->tablesExist('tl_layout')) {
+            return false;
+        }
+
+        $columns = $schemaManager->listTableColumns('tl_layout');
+
+        if (!isset($columns['webfonts'])) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public function run(): MigrationResult
+    {
+        $tableQuoted = $this->connection->quoteIdentifier('tl_layout');
+        $webfontsFieldQuoted = $this->connection->quoteIdentifier('webfonts');
+        $headFieldQuoted = $this->connection->quoteIdentifier('head');
+
+        $rows = $this->connection->fetchAllAssociative("
+            SELECT
+                id, $webfontsFieldQuoted, $headFieldQuoted
+            FROM
+                $tableQuoted
+            WHERE
+                $webfontsFieldQuoted != ''
+        ");
+
+        foreach ($rows as $row) {
+            $this->connection
+                ->prepare("UPDATE $tableQuoted SET $headFieldQuoted = :head WHERE id = :id")
+                ->execute([
+                    ':id' => $row['id'],
+                    ':head' => $row['head']."\n".'<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=' . str_replace('|', '%7C', $row['webfonts']) . '">',
+                ])
+            ;
+        }
+
+        $this->connection->executeStatement("ALTER TABLE tl_layout DROP $webfontsFieldQuoted");
+
+        return $this->createResult(true);
+    }
+}
--- a/core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
+++ b/core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
@@ -50,22 +50,18 @@ class MergeWebfontsMigration extends AbstractMigration
 
     public function run(): MigrationResult
     {
-        $tableQuoted = $this->connection->quoteIdentifier('tl_layout');
-        $webfontsFieldQuoted = $this->connection->quoteIdentifier('webfonts');
-        $headFieldQuoted = $this->connection->quoteIdentifier('head');
-
         $rows = $this->connection->fetchAllAssociative("
             SELECT
-                id, $webfontsFieldQuoted, $headFieldQuoted
+                id, webfonts, head
             FROM
-                $tableQuoted
+                tl_layout
             WHERE
-                $webfontsFieldQuoted != ''
+                webfonts != ''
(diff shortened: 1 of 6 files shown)


fritzmg (dismissed):

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
```diff
@@ -0,0 +1,80 @@
...
+        $headFieldQuoted = $this->connection->quoteIdentifier('head');
+
+        $rows = $this->connection->fetchAllAssociative("
+            SELECT
+                id, $webfontsFieldQuoted, $headFieldQuoted
+            FROM
+                $tableQuoted
+            WHERE
+                $webfontsFieldQuoted != ''
+        ");
```
m-vo:
```suggestion
        $rows = $this->connection->fetchAllAssociative("SELECT id, webfonts, head FROM tl_layout WHERE webfonts != ''");
```

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
```diff
@@ -0,0 +1,80 @@
...
+            $this->connection
+                ->prepare("UPDATE $tableQuoted SET $headFieldQuoted = :head WHERE id = :id")
+                ->execute([
+                    ':id' => $row['id'],
+                    ':head' => $row['head']."\n".'<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=' . str_replace('|', '%7C', $row['webfonts']) . '">',
+                ])
+            ;
+        }
+
+        $this->connection->executeStatement("ALTER TABLE tl_layout DROP $webfontsFieldQuoted");
```
m-vo:
```suggestion
        $this->connection->executeStatement('ALTER TABLE tl_layout DROP webfonts');
```

m-vo (changes requested):

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
leofeyer: Not sure about this. I took the code from @ausi's migration:

https://github.com/contao/contao/blob/b0f4b53f4e869d63af16323aa24a4d1bd6adc421/core-bundle/src/Migration/Version410/OrderFieldMigration.php#L98-L112

m-vo: Mh, I don't like this added complexity - this also breaks SQL validation in PhpStorm.

If you still want to go down this route, there is a leftover `tl_layout` in the last query. :stuck_out_tongue:

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
```diff
@@ -0,0 +1,80 @@
...
+                id, $webfontsFieldQuoted, $headFieldQuoted
+            FROM
+                $tableQuoted
+            WHERE
+                $webfontsFieldQuoted != ''
+        ");
+
+        foreach ($rows as $row) {
+            $this->connection
+                ->prepare("UPDATE $tableQuoted SET $headFieldQuoted = :head WHERE id = :id")
```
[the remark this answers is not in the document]
m-vo: Might even want to change this to `$this->connection->update(...)`.

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
```diff
@@ -0,0 +1,80 @@
...
+                ->execute([
+                    ':id' => $row['id'],
+                    ':head' => $row['head']."\n".'<link rel="stylesheet" href="https://fonts.googleapis.com/css?family='.str_replace('|', '%7C', $row['webfonts']).'">',
+                ])
+            ;
```
m-vo:
```suggestion
            $head = $row['head']."\n".'<link rel="stylesheet" href="https://fonts.googleapis.com/css?family='.str_replace('|', '%7C', $row['webfonts']).'">';

            $this->connection->update('tl_layout', ['head' => $head], ['id' => $row['id']]);
```

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
ausi: `quoteIdentifier()` is only needed when the identifier is dynamic. For static identifiers like in this case we normaly write them directly into the query as suggested by @m-vo

leofeyer: Changed in 7e3c42d2bc2592001430c275161e5fe5e3b75bd6.

core-bundle/src/Migration/Version411/MergeWebfontsMigration.php
leofeyer: Changed in 7e3c42d2bc2592001430c275161e5fe5e3b75bd6.

aschempp (approved):
I would check if there's any layout with fonts in `shouldRun` but otherwise looks good

leofeyer:
The migration needs to run even if there are no layouts with web fonts, because it will remove the database column and thus prevent itself from running multiple times.

aschempp:
> The migration needs to run even if there are no layouts with web fonts, because it will remove the database column and thus prevent itself from running multiple times.

well if there's no field with data, it does not need to run, and it wouldn't run again because there's still no field with data 😉

frontendschlampe:
I think, to migrate the field isn't a good idea. I would give a hint to anybody, that the field will be deleted in Contao 4.11 and the webmaster should add the fonts again (or change it before update). If you update from 4.10 to 4.11, you have to do and check much more than just pressing the update button anyway. Maybe a hint on homescreen in the backend ...

aschempp:
@frontendschlampe not sure what you're referring to. Migrations are there to make sure people can upgrade without breaking their sites. You're suggesting we should break their site, which makes no sense except something like an "opinion" they _should_ switch to local fonts…

frontendschlampe:
First of all: we don't break any installation because of a webfont. Then it's an update which will surely require more adjustments than just pressing the update button in the Manager. In addition, third party extensions that might do something with the web fonts will get function problems because the field is no longer available or the desired function of the extension is no longer available - and this without any message. Not to mention the recommendation for local integration of web fonts.

But it's just my opinion. We have lot of other things to do, so you don't have to worry about copying web fonts. It's everytime the same discussion: should we migrate or not.

m-vo:
>  Then it's an update which will surely require more adjustments than just pressing the update button in the Manager

> We have lot of other things to do

I don't think so. After running the migrations you should be fine. Where are you seeing problems?

frontendschlampe:
> I don't think so. After running the migrations you should be fine. Where are you seeing problems?

We talk about 4.11 ... I don't know. We will see ...

> We have lot of other things to do

This sentence does not refer to an update, but to the load on the core team. ;-)

PR #2432, opened 2020-10-15
Use meaningful values for the "autocomplete" attributes

### LINKED ISSUE #1669: Set helpful values to autocomplete attribute in user context

leofeyer:
Implements #1669

### DIFF
--- a/core-bundle/src/Resources/contao/forms/FormPassword.php
+++ b/core-bundle/src/Resources/contao/forms/FormPassword.php
@@ -167,7 +167,7 @@ class FormPassword extends Widget
 	public function generate()
 	{
 		return sprintf(
-			'<input type="password" name="%s" id="ctrl_%s" class="text password%s" value=""%s%s',
+			'<input type="password" name="%s" id="ctrl_%s" class="text password%s" value="" autocomplete="new-password"%s%s',
 			$this->strName,
 			$this->strId,
 			($this->strClass ? ' ' . $this->strClass : ''),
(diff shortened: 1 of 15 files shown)


bytehead (dismissed):

bytehead (approved):

fritzmg (approved):

core-bundle/src/Resources/contao/forms/FormPassword.php
```diff
@@ -167,7 +167,7 @@ public function parse($arrAttributes=null)
 	public function generate()
 	{
 		return sprintf(
-			'<input type="password" name="%s" id="ctrl_%s" class="text password%s" value=""%s%s',
+			'<input type="password" name="%s" id="ctrl_%s" class="text password%s" value="" autocomplete="new-password"%s%s',
```
aschempp, line 170: are you sure this is always a new password? Like if I use the personal data module?

asaage: For the personal data module you would use this to change your pw - this seems correct.

But core-bundle/src/Resources/contao/forms/FormPassword.php would be used for every formgeneator password-field.
This comes always in pair with a pw-confirm field wich means it's aways a ```autocomplete="new-password"```.

The Question for me is: can i use the formgenerator to create a password-field without confirm-field (to use it in a processFormData hook for example)?

leofeyer: > are you sure this is always a new password?

Yes, it is. The widget always generates a password and a confirmation field.

PR #2438, opened 2020-10-19
Remove the CDN integration of MooTools and jQuery

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | 
| Docs PR or issue | not necessary

Google Chrome is now also forcing separation of local caches.
That means there's no point in loading assets from foreign CDN's anymore other than traffic reduction.

I think we should remove the option completely. It's also better in terms of GDPR compliance. If you want to reduce traffic, you can still put the whole server behind a CDN (which sounds better to me anyway because it will also cache regular pages if configured correctly).

Also see https://www.stefanjudis.com/notes/say-goodbye-to-resource-caching-across-sites-and-domains/


ausi (dismissed):

fritzmg (dismissed):

bytehead (dismissed):

aschempp (dismissed):
But can we also use the asset component to load the local files?

Toflar:
No idea, I just left things the way they are now.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #2453, opened 2020-10-22
Evaluate unknown simple tokens

### LINKED ISSUE #2402: Fix StringUtil::parseSimpleTokens evaluating unknown tokens as false
**Affected version(s)**

Contao 4.4

**Description**

from #2341:

> If we always return false if the condition contains unknown variables wouldn’t that flip the logic of inverted comparisons like {if foo!=""}?

We should define the values with `null` instead and continue with the evaluation. 

see https://github.com/contao/contao/pull/2341#discussion_r500259589

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2402
| Docs PR or issue | -

This is a followup to #2341 for Contao 4.4 - 4.10.


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #2459, opened 2020-10-26
Require pdo_mysql extension

aschempp:
see https://github.com/contao/contao-manager/issues/606

PR #2462, opened 2020-10-26
Remove hard dependency on PDO (4.4)

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2459
| Docs PR or issue | -

This would add the PDO options dynamically via the Manager Plugin of the Manager Bundle. It will not add anything if a driver other than `pdo_mysql` is used, or custom `options` have already been set.

### DIFF
--- a/manager-bundle/src/ContaoManager/Plugin.php
+++ b/manager-bundle/src/ContaoManager/Plugin.php
@@ -151,9 +151,9 @@ class Plugin implements BundlePluginInterface, ConfigPluginInterface, RoutingPlu
             return $extensionConfigs;
         }
 
-        $extensionConfigs = $this->setDatabaseServerVersion($extensionConfigs, $container);
+        $extensionConfigs = $this->addDefaultServerVersion($extensionConfigs, $container);
 
-        return $this->setPdoDriverOptions($extensionConfigs, $container);
+        return $this->addDefaultPdoDriverOptions($extensionConfigs, $container);
     }
 
     /**
@@ -171,7 +171,7 @@ class Plugin implements BundlePluginInterface, ConfigPluginInterface, RoutingPlu
      *
      * @return array
      */
-    private function setDatabaseServerVersion(array $extensionConfigs, PluginContainerBuilder $container)
+    private function addDefaultServerVersion(array $extensionConfigs, PluginContainerBuilder $container)
     {
         $params = [];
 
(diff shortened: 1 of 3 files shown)


manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -150,6 +151,28 @@ public function getExtensionConfig($extensionName, array $extensionConfigs, Plug
...
+
+        return $this->setPdoDriverOptions($extensionConfigs, $container);
+    }
+
+    /**
+     * Sets path to enable autoloading of legacy Contao modules.
+     *
+     * @param string $modulePath
+     */
+    public static function autoloadModules($modulePath)
```
leofeyer, line 163: Why did you add this method?

bytehead: It was already there, see lines 190-196 on the diff.

fritzmg: It's already there:

https://github.com/contao/contao/blob/8b07746dea9eecb9527a34fe05cb8ff824cd7aa2/manager-bundle/src/ContaoManager/Plugin.php#L189-L197

fritzmg: I originally inserted the new private methods before `public static function autoloadModules` - however, as per the code style rules, `public` methods come before `private` ones, therefore the CS fixer moved it around, resulting in this slightly more unreadable diff ;)

aschempp (dismissed):

fritzmg:
I'll provide a separate PR for Contao 4.9, as this cannot be merged upstream.

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -187,12 +210,41 @@ public function getExtensionConfig($extensionName, array $extensionConfigs, Plug
...
+        // Do not add PDO options, if constant does not exist
+        if (!\defined('PDO::MYSQL_ATTR_MULTI_STATEMENTS')) {
+            return $extensionConfigs;
+        }
+
+        foreach ($extensionConfigs as $extensionConfig) {
+            // Do not add PDO options, if selected driver is not pdo_mysql
+            if (isset($extensionConfig['dbal']['connections']['default']['driver']) && 'pdo_mysql' !== $extensionConfig['dbal']['connections']['default']['driver']) {
+                return $extensionConfigs;
+            }
```
fritzmg: A subsequent extension config could set the driver back to `pdo_mysql`. Should we consider this possibility? Seems unlikely though.

aschempp: Our only real problem is that the constants don't exist, but we check for that. So I'd be fine with that.

aschempp (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2473, opened 2020-10-30
Add a Cache-Control header to the back end response

### LINKED ISSUE #1928: No cache headers in back end responses
**Affected version(s)**

Contao 4.9

**Description**

In Contao 4.4, any back end request also returned the following headers:

```
Cache-Control: max-age=0, must-revalidate, private
Expires: Fri, 10 Jul 2020 12:56:44 GMT
```

(`Expires` will always be the current time). In Contao 4.9 however, this is not the case any more. This in itself is not problematic, but it breaks the back end, if someone happens to have

```
ExpiresByType text/html "access plus 6 hours"
```

in their `.htaccess`, since this causes the back end requests to be cached by the browser for the configured time.

Shouldn't we disallow caching of back end responses in the private cache with the appropriate response headers? Quite a few users in the Contao community already stumbled over this issue.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1928
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/classes/BackendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/BackendTemplate.php
@@ -49,8 +49,6 @@ class BackendTemplate extends Template
 	/**
 	 * Return a response object
 	 *
-	 * @param bool $blnForceCacheHeaders
-	 *
 	 * @return Response The response object
 	 */
 	public function getResponse()
--- a/core-bundle/src/Resources/contao/classes/BackendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/BackendTemplate.php
@@ -10,6 +10,8 @@
 
 namespace Contao;
 
+use Symfony\Component\HttpFoundation\Response;
+
 /**
  * Provide methods to handle back end templates.
  *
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

core-bundle/src/Resources/contao/classes/BackendTemplate.php
```diff
@@ -44,6 +46,21 @@ public function parse()
 		return $strBuffer;
 	}
 
+	/**
+	 * Return a response object
+	 *
+	 * @param bool $blnForceCacheHeaders
```
Toflar: ? :)

leofeyer: Removed in 2b1d383e9cbf0bddc29705bda455c93c58878ea5. 😊

Toflar (approved):

core-bundle/src/Resources/contao/classes/BackendTemplate.php
```diff
@@ -44,6 +46,19 @@ public function parse()
...
+	 * Return a response object
+	 *
+	 * @return Response The response object
+	 */
+	public function getResponse()
+	{
+		$response = parent::getResponse();
+		$response->headers->set('Cache-Control', 'no-cache, no-store');
+
+		return $response->setPrivate();
```
aschempp, line 59: why did you add `setPrivate`? The response should **not** be privately cacheable by the browser?

PR #2475, opened 2020-10-30
Adjust the PictureInterface for Contao 5

leofeyer:
This PR adds the missing argument of the `create()` method as comment, so we know how the interface will change in Contao 5.


core-bundle/src/Image/PictureFactoryInterface.php
```diff
@@ -35,5 +35,5 @@ public function setDefaultDensities($densities);
      *
      * @return PictureInterface
      */
-    public function create($path, $size = null);
+    public function create($path, $size = null/*, ResizeOptions $options = null*/);
```
ausi, line 40: Should we add `use Contao\Image\ResizeOptions;` to define which ResizeOptions are meant?
Or change it to this?
```suggestion
    public function create($path, $size = null/*, \Contao\Image\ResizeOptions $options = null*/);
```

leofeyer: I would just keep it as is, because that is how Symfony does it, too: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/DependencyInjection/Definition.php#L146

An unused use statement would be removed by the CS fixer anyway.

ausi: In the linked example the class is from the same namespace. I think we should add `@param ResizeOptions $options` too, this should stop the CS fixer from removing the use statement, the IDE can show hints for the third parameter and Symfony seems to be using `@param` this as well in this case.

leofeyer: But our rule is never to add `@param` if there is a type hint. 🤷🏻‍♂️

ausi: I thought we don’t add `@param` if it doesn’t provide additional information. We could add a short description that this param will be in the interface in Contao 5.0

m-vo: Yeah, we also do add `@param` annotations if there is a textual description or additional type information (like `array<string, Foo>`). Not having the namespace also doesn't allow you navigating to the file in your IDE.

leofeyer: Added in 2caac8f64c87718b283b6400ed9e7e4e77bc24c5.

leofeyer: 🤦🏻‍♂️

ausi (approved):

aschempp (approved):

PR #2477, opened 2020-10-31
Improve overwriting templates from within backend

### LINKED ISSUE #2452: Create Twig templates from within the backend
see https://github.com/contao/contao/issues/2428#issuecomment-712746509

> After #2051 what's still missing is creating Twig templates from the back end. May be we can provide such a functionality for some known templates? Like the error, maintenance and layout template for example.

&rarr; Maybe we can offer to create Twig templates in the backend. This would most likely target all bundles and they would need to get copied into `templates/bundles/<bundle>/<template>` on demand.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Closes #2452  
| Docs PR or issue | -

This PR is an attempt to allow both creating Contao templates (PHP) + bundle templates (Twig) from the backend. It also supports overwriting templates from symlinked bundles (currently breaking because of relative paths inside project root).

bundle_templates

This PR is using `symfony/form` and `symfony/validator` to create + handle the form which is something we did not do before. The idea is: You've got a `FormType` abstracting the form (that itself is a service and knows how to populate the choices). Bound to it is a DTO that has validation annotations on its properties. This makes the controller pretty simple: it only needs to show the form and copy a file + redirect when the the form is valid. The rendering of the form itself is done via a simple custom form theme that mimics the current backend markup. 

Let me know what you think. If we want this, I'll polish it + add tests.


leofeyer:
+1 for the concept. Not 100% sure about using Sf forms as you might have expected. 😄

fritzmg:
Since this is pretty independent from anything else, I wouldn't mind experimenting with Symfony Forms within Contao here ;)

m-vo:
Superseded by #7587

PR #2496, opened 2020-11-16
Re-use tl_member DCA password field in ModuleCloseAccount

### LINKED ISSUE #2492: ModuleCloseAccount does not use password field from tl_member
**Affected version(s)**

Contao >= 4.9

**Description**

`ModuleCloseAccount` should be consistent with all the other member related modules.

Each member related module (`ModulePersonalData`, `ModulePassword`, `ModuleChangePassword`,...) uses the fields located in the `tl_member` DCA - `ModuleCloseAccount` does not. There is currently no way to modify the password field in any way without simply copying the `ModuleCloseAccount` class.

In my case I needed to add a placeholder attribute to all fields from `tl_member` which are `feEditable`. This works in all modules other than `ModuleCloseAccount`.

I'd like to propose the following change:

```php
    /**
     * Generate the module
     */
    protected function compile()
    {
        $this->import(FrontendUser::class, 'User');

        System::loadLanguageFile('tl_member');
        $this->loadDataContainer('tl_member');

        // Initialize the password widget
        $arrField = $GLOBALS['TL_DCA']['tl_member']['fields']['password'];
```

bennyborn:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2492 

This PR re-uses the password field from tl_member DCA like instead of creating a new one. This way it's possible to modify the properties of the field like it's already the case with all other member related modules.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php
@@ -62,14 +62,11 @@ class ModuleCloseAccount extends Module
 	{
 		$this->import(FrontendUser::class, 'User');
 
+		System::loadLanguageFile('tl_member');
+		$this->loadDataContainer('tl_member');
+
 		// Initialize the password widget
-		$arrField = array
-		(
-			'name' => 'password',
-			'inputType' => 'text',
-			'label' => $GLOBALS['TL_LANG']['MSC']['password'][0],
-			'eval' => array('hideInput'=>true, 'preserveTags'=>true, 'mandatory'=>true, 'required'=>true)
-		);
+		$arrField = $GLOBALS['TL_DCA']['tl_member']['fields']['password'];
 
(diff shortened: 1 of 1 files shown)


bytehead:
You should rebase your base branch (`fix/membercloseaccount`) to the current upstream (`4.9`).

bennyborn:
> You should rebase your base branch (`fix/membercloseaccount`) to the current upstream (`4.9`).

Never really worked with rebasing. Any advice how to do this? Everything I figured out so far results in many conflicts :\

m-vo:
Have a look at the `--onto` option of `git rebase`. 

See https://git-scm.com/book/en/v2/Git-Branching-Rebasing for more details.

m-vo:
Btw.: Even without rebasing you could have reused this PR (instead of opening #2497) by force pushing:

`git push -f <remote> fix/membercloseaccount`

PR #2497, opened 2020-11-16
Re-use the tl_member.password field in ModuleCloseAccount

### LINKED ISSUE #2492: ModuleCloseAccount does not use password field from tl_member

bennyborn:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2492 

This PR re-uses the password field from tl_member DCA like instead of creating a new one. This way it's possible to modify the properties of the field like it's already the case with all other member related modules.


leofeyer:
The password field in the module uses different settings than the one in `tl_member`, therefore we are not re-using it currently. What exactly do you need to modify?

bennyborn:
> What exactly do you need to modify?

As I wrote in the original issue I needed to add a placeholder. 

But it doesn't matter what my exact use-case was rather than not being able to modify the field at all. Maybe someone wanted to give it a different label or overwrite the inputType to a custom widget. None of that is possible if the field is simply defined inline.

Furthermore I see no real reason not to re-use the field. 
The settings are almost the same except for some minor eval attributes (`minlength`, `feEditable`, `feGroup`, `tl_class`) which should cause no problems.

bennyborn:
@leofeyer I've just re-added the `hideInput` option in the eval and fixed the missing field name. This way it works like the original field with the added benefit to be able to modify it via the `tl_member` DCA.

aschempp (changes requested):
the field in `tl_member` also has a different input type, so this won't work either…

bennyborn:
> the field in `tl_member` also has a different input type, so this won't work either…

I noticed that too but actually there seems to be no need to change the type. It works absolutely fine.

fritzmg (approved):

aschempp:
I can hardly believe that, because the password field type generates two inputs (to confirm the password) 🤔

bennyborn:
Because @leofeyer just assigned me.... Where are we going from here? It seems to work from my side, @fritzmg also approved it. If @aschempp has any doubts be my guest testing the changes.

leofeyer:
Current configuration:

```php
		$arrField = array
 		(
 			'name' => 'password',
 			'inputType' => 'text',
 			'label' => $GLOBALS['TL_LANG']['MSC']['password'][0],
 			'eval' => array('hideInput'=>true, 'preserveTags'=>true, 'mandatory'=>true, 'required'=>true)
 		);
```

New configuration:

```php
		'password' => array
		(
			'label'                   => &$GLOBALS['TL_LANG']['MSC']['password'],
			'inputType'               => 'password',
			'eval'                    => array('mandatory'=>true, 'preserveTags'=>true, 'minlength'=>Contao\Config::get('minPasswordLength'), 'feEditable'=>true, 'feGroup'=>'login', 'tl_class'=>'clr', 'hideInput'=>true),
			'save_callback' => array
			(
				array('tl_member', 'setNewPassword')
			),
			'sql'                     => "varchar(255) NOT NULL default ''"
		),
```

This is quite different IMHO, starting with the field type currently being `text` (one input field) and afterwards being `password` (two input fields). I doubt that we can re-use the field.

aschempp:
I just tested this as suggested by @bennyborn and he's actually right, the input field "works". That's however because the field type is hardcoded in https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php#L75, so it would not be possible to override the widget (as mentioned in https://github.com/contao/contao/pull/2497#issuecomment-730320919).

bennyborn:
> [...] so it would not be possible to override the widget [...].

If we changed the type to `FormPassword` we'd have the problem of the unwanted confirmation field. Leaving it as `FormTextField` we IMHO have the best possibilities for customization.

leofeyer:
Thank you @bennyborn.

PR #2051, opened 2020-08-01
Allow editing Twig files in the template editor

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Closes #1399
| Docs PR or issue | -

This PR&hellip;
 - ~~adds `html5` and `twig` to the list of editable extensions~~
 - allows editing the source of ~~*any* file with an allowed extension~~ `html5` + `twig` files in `tl_templates`
 - purges the Twig cache after editing a file with a `.twig` extension

~~Is there a reason why we did not add `html5` to the editable files before but instead altered the config on the fly? The only thing I could think of is that people that adjusted their config would not have to do anything to be able to edit templates (they would now :grin:).~~

PR #2501, opened 2020-11-18
Handle altering Twig bundle paths at runtime

### LINKED ISSUE #2482: Remove or restrict editing templates in the backend
**Affected version(s)**

\>= Contao 4.9 (all versions with Symfony 4 / default template path `/templates`)

**Description**

We're currently allowing backend users to alter things inside the `templates` directory. This leads to a tricky problem because for some operations we would need to rebuild the container. 

If a user wants to overwrite a bundle template in Sf4 they put it under `templates/bundles/<Bundle>/<template>`. This folder is visible in the backend and you can rename or delete it. **But** this will break the application ([ref](https://github.com/contao/contao/pull/2051#issuecomment-719936643)).

**Reproducer**

 - add a  `service_unavailable.html.twig` to `templates/bundles/ContaoCoreBundle/Error`
 - rebuild the cache 
 - open the BE > Templates and rename or delete `ContaoCoreBundle`

**Background**

Here is why: The `TwigExtension` from the `symfony/twig-bundle` adds a method call `NativeFilesystemLoader#addPath` to the container for each found template path at compile time. This does include the ones in `templates/bundles`. `NativeFilesystemLoader#addPath` will throw an exception at runtime if it is given a non-existing path. So, for one: newly created folders won't be picked up as template paths. But worse: if we let the user remove/rename any folder inside `templates/bundles` in the backend the application will immediately crash.

In a dev setup you would just run `cache:clear` and done. In the application rebuilding the container seems impossible:

- It cannot be done at runtime (inside the app) because it'll change classes/paths and will break subsequent execution. It *can* be done to some extent in a very late `kernel.terminate` listener but the request in which this is happening would still need to be made separately (e.g. via ajax) and wouldn't have a way to signal when it has finished. Maybe it could be done with another entry point but that's imo a bit over the top.

- You can switch between `prod` and `dev` in the backend. So we might need to rebuild the cache for both environments, otherwise switching back and forth might be broken. There is imo no way to know the cache location of the 'other' environment, though. Running a command does not work because we do not know the console path. Creating a new Kernel in the respective env and asking it for the directory doesn't work either because we do not know how to instantiate the kernel class. We can only *assume* that it is in `<cacheOfEnvA>/../<cacheOfEnvB>`. 

**Possible solution**

Imo the cleanest solution would be to remove editing templates in the backend and migrate it to the manager. But I know there are lots of people using this functionality for all sorts of things. 

Other than that, I'd disallow editing/deleting any of the affected folders or skip the file operation and add a message that says something like "please remove the folder by hand + rebuild the cache". 

Wdyt? Or am I missing something completely and the problem is easily solveable? :crossed_fingers:  :smiley:

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2482
| Docs PR or issue | -

Please see #2482 + https://github.com/contao/contao/pull/2051#issuecomment-719936643 for reference.

#### What's the deal?

I came to the conclusion that rebuilding the app cache at runtime is a bad idea. Here is a workaround for the problem that prevents the hard fail and introduces a info message instead. Note that this logic will only kick in when renaming/deleting folders inside `templates/bundles`. Altering templates, even those inside this path, is unaffected.

Here are the changes:

1) This introduces a new fail tolerant twig filesystem loader that decorates `twig.loader.native_filesystem` and ignores missing paths in some cases. In order for this to work the `addPath` method calls made in compiler passes must also be rewired to the new loader (now done in a  `TwigPathsPass` compiler pass).

2) After altering bundle paths in the backend, a message is shown to the user that the application cache must be rebuilt. This is done by marking the `ApplicationCacheState` as dirty. A `BackendRebuildCacheMessageListener` then checks this state on each backend request and - if so - adds the respective flash message. The state is cached and thus will be reset when rebuilding the cache.


core-bundle/tests/Twig/FailTolerantFilesystemLoaderTest.php
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * todo: remove.
+     */
+    public function testCI(): void
+    {
+        // works
+        $this->assertDirectoryExists(Path::join($this->getTestRoot(), 'app'));
+
+        // fails
+        $this->assertDirectoryExists(Path::join($this->getTestRoot(), 'templates'));
```
m-vo: I have no clue why the tests are failing on CI? The template fixture folder I added seems missing. I created this temporary test to showcase the problem.

There are `.gitignore` files in them and they're clearly there if i clone the repo and checkout this branch....  What am I missing? :confused:

m-vo: The issue was indeed `core-bundle/tests/Contao/TemplateLoaderTest.php` removing the folder. :facepalm: 
Adjusted in 3b93f875f50bdfdacc51b1dd5e885e9c6e53c4da.

m-vo:
The reported CS issues seem unrelated to this PR.

This should be rebased onto 4.9 I guess? Or should we rather ignore the problem there by maybe hiding the `bundles` folder. (Then it would go into 4.10.)

leofeyer:
Based on the size of the PR, I lean towards merging it into Contao 4.11. @contao/developers WDYT?

leofeyer:
> The reported CS issues seem unrelated to this PR.

Check the .yml files: https://github.com/contao/contao/pull/2501/files#diff-2a0d76195d45f9f9b0486317e8d4a43845ef063101fa6ea0b3bbe3ef241e71a0

m-vo:
Ah you're right. I just had a look at the other two places, that are now resolved.

aschempp:
I don't understand why we need a service to manage a state, where a constant for the cache key would provide the same abstraction. Also, imho we should replace the class instead of decorate, then we don't need a compiler pass. Last thing, it now still hard fails if I edit templates on FTP, but I don't think it should?

m-vo:
@aschempp 

> Last thing, it now still hard fails if I edit templates on FTP, but I don't think it should?

You mean editing bundle folders? I think this is expected to fail. We're only entering 'fail safe mode' when a change is triggered in the backend (and the container state is marked dirty). Not doing this would imo degrade DX because exceptions will be swallowed.

aschempp:
What DX do you mean?

m-vo:
Expecting to see an exception if a bundle path does not exist anymore?

m-vo:
> Also, imho we should replace the class instead of decorate, then we don't need a compiler pass.

Imo this won't work (or at least I didn't get it working). The `TwigExtension` hardwires class names for instance, so the registered method calls do not reach the class when overwriting.

core-bundle/src/DependencyInjection/Compiler/TwigPathsPass.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Contao\CoreBundle\DependencyInjection\Compiler;
+
+use Contao\CoreBundle\Twig\FailTolerantFilesystemLoader;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @internal
+ */
+class TwigPathsPass implements CompilerPassInterface
```
leofeyer: I think we could improve the class name to say what the class does, e.g. `RewireTwigPathsPass` or `AddTwigPathsToFailTolerantLoaderPass` or something alike.

leofeyer (commented):
The concept makes sense to me. 👍🏻  @contao/developers Any objections?

core-bundle/src/DependencyInjection/Compiler/TwigPathsPass.php
m-vo: Agreed. I changed it to `RewireTwigPathsPass` for now.

see f9cccc33263f4231f9e0c1fb31cd5f8d5d261319

m-vo: Sorry for the force-push confusion - I accidentally had dropped the file when renaming :facepalm:.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #2502, opened 2020-11-19
Add font smoothing for Safari on retina displays

leofeyer:


m-vo:
Shouldn't this go into 4.4?

leofeyer:
No, because Contao 4.4 uses a web font instead of a native font stack.

bytehead (approved):

m-vo:
Ah, right. My bad.

ausi:
I think we should not add this. See https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

leofeyer:
I think so, too. It turned out to be a problem on my iMac. 🙈

PR #1935, opened 2020-07-15
Allow global page model access in fragments

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Currently fragment controllers for content elements and front end modules only receive the _ID_ of the page model of the current page in their `pageModel` request attribute, whereas everywhere else the `pageModel` request attribute is actually the global page model object itself.

This means that fragments are not able to access the global page model object at all, if they do not want to use `global $objPage` or `$GLOBALS['objPage']`, since the global page object is detached from the model registry. Thus you couldn't implement something like `ModuleNewsreader` as a fragment, without resorting to accessing the global variable, since the module needs to be able to modify the current global page object.

As far as I understand it, the `pageModel` only needs to be an ID for `esi` rendered fragments, but not really for `inline` or `forward` rendered fragments. Therefore this PR would change this behaviour.

However - why does this distinction need to be made at all? Isn't the global page object missing anyway (and thus also the `pageModel` request attribute), when rendering an `esi` fragment?

PR #2506, opened 2020-11-19
Handle the global page model in fragments

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1935

As discussed on Mumble on the 2020-11-19, this should make the (global) page model available in the fragment controllers.


core-bundle/src/HttpKernel/ModelArgumentResolver.php
```diff
@@ -79,6 +80,15 @@ private function fetchModel(Request $request, ArgumentMetadata $argument): ?Mode
             return $value;
         }
 
+        // Special handling for pageModel that could be globally registered
+        if (
+            isset($GLOBALS['objPage'])
+            && $GLOBALS['objPage'] instanceof PageModel
+            && $GLOBALS['objPage']->id === (int) $value
```
dmolineus: Afaik the id value is a string, so this comparision is very likely never true. I suggest to use `(int) $GLOBALS['objPage']->id === (int) $value`

aschempp: Fixed in a5c2326300d31a03b92cc81f12cab5b9fceaa9bc

leofeyer (approved):

PR #2508, opened 2020-11-19
hide manager link in back end

MDevster:
## feature

# Question
Is there a parameter for the config to hide the back end link for the manager?

# Solution
Now you can set manager_path to false in parameters.yml and the link will not be added.

```
parameters:
    contao_manager:
        manager_path: false

```

Note: The test should be viewed again by someone who is familiar with it ;)

### DIFF
--- a/manager-bundle/src/DependencyInjection/Compiler/ContaoManagerPass.php
+++ b/manager-bundle/src/DependencyInjection/Compiler/ContaoManagerPass.php
@@ -26,8 +26,7 @@ class ContaoManagerPass implements CompilerPassInterface
         $webDir = $container->getParameter('contao.web_dir');
         $managerPath = $container->getParameter('contao_manager.manager_path');
 
-        if('false' === $managerPath)
-        {
+        if('false' === $managerPath) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


baumannsven:
I think this must configure in the configuration for the contao manager bundle.

```yaml
contao_manager:
    manager_path: 'Enter a path where contao.manager.pahr.php does not exist.'
```

With this configuration is this pull request obsolete.

baumannsven:
https://github.com/contao/contao/blob/4cfc0ef8b3d69c28c14bfa50abc47d9780509190/manager-bundle/src/DependencyInjection/Compiler/ContaoManagerPass.php#L29

In the compiler pass the manager_path should be set to false, be allowed.

aschempp (changes requested):
I think @baumannsven is right, this would keep the default path and not remove the back end link.

leofeyer:
Does the following not work?

```yaml
parameters:
    contao_manager:
        manager_path: ~
```

aschempp:
`null` is the default value, so Contao would check if the file exists and display a link.

leofeyer:
@MDevster Do you know that you can rename the `contao-manager.phar.php` file (e.g. to `cm.phar.php`), which will make the link disappear?

We have discussed the issue in Mumble on April 8th. If we decide to merge the PR, we will have to use a real boolean `false` instead of `'false'`. Also, the validation of the path should be in the `Configuration.php` file instead of the pass.

Issue #2263, opened 2020-09-09
PHP 8.0 Roadmap


ausi:
The planned release date for PHP 8.0 is December 3rd.  
8.0.0-RC1 is planned for September 17th.

Contao 4.9, 4.10 and 4.11 will be compatible with PHP 8.0.  
Contao 4.4 will not be compatible.

### ToDos (to be extended):

- [x] Agree on `^7.2 || ^8.0` vs. `>=7.2` (see https://github.com/sebastianbergmann/phpunit/commit/fa50b8949b6f2f98a614fe66f54c5e6ea6a0f078 for example). Result: `^7.2 || ^8.0`
- [x] Check all dependencies and add incompatible ones to this list.

Our packages:
- [x] `ausi/slug-generator` add compatibility and release new version https://github.com/ausi/slug-generator/pull/23
- [x] `toflar/psr6-symfony-http-cache-store` add compatibility and release new version https://github.com/Toflar/psr6-symfony-http-cache-store/pull/36
- [x] `terminal42/escargot` add compatibility and release new version https://github.com/terminal42/escargot/pull/18
- [x] `terminal42/service-annotation-bundle` add compatibility and release new version https://github.com/terminal42/service-annotation-bundle/issues/5
- [x] `‌contao/image` add compatibility and release new version https://github.com/contao/image/pull/74
- [x] `contao/imagine-svg` (is already compatible)
- [x] `contao/manager-plugin` add compatibility and release new version https://github.com/contao/manager-plugin/pull/36
- [x] `contao/maintenance-bundle-deprecated` add compatibility and release new version https://github.com/contao/maintenance-bundle-deprecated/pull/1
- [x] `contao/mobile-page-layout-bundle` add compatibility and release new version https://github.com/contao/mobile-page-layout-bundle/pull/5
- [x] `contao/tcpdf-bundle` add compatibility and release new version https://github.com/contao/tcpdf-bundle/pull/1
- [x] `contao/easy-coding-standard` add compatibility and release new version https://github.com/contao/easy-coding-standard/issues/6
- [x] `contao/phpstan` add compatibility and release new version https://github.com/contao/phpstan/pull/20
- [x] `contao/test-case` add compatibility and release new version https://github.com/contao/test-case/issues/9
- [x] `contao/monorepo-tools` add compatibility (no release necessary) https://github.com/contao/monorepo-tools/pull/6
- [x] `contao/managed-edition` add compatibility and release new version https://github.com/contao/managed-edition/issues/56
[...]

PR #2509, opened 2020-11-20
Add compatibility with terminal42/escargot 1.0

ausi:
@Toflar anything more to do here or should version 1.0 be completely backwards compatible?

Related: #2263

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -110,7 +110,7 @@
         "symfony/var-dumper": "4.4.*",
         "symfony/web-profiler-bundle": "4.4.*",
         "symfony/yaml": "4.4.*",
-        "terminal42/escargot": "^0.6.0",
+        "terminal42/escargot": "^0.6.0 || ^1.0",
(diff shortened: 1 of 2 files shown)


Toflar:
We can require `^1.0` only imho. It's the same as the latest `0.6.x` other than php 8 compatibility.
Sorry, could've done the PR myself, thought you wanted to incorporate it into the other PR :)

ausi:
AFAIR we tried in the past to not increase dependency versions in bugfix versions unless necessary. This should keep updating more easy for people that use extensions that require escargot in version 0.6

ausi:
But we should create a PR for Contao 4.11 that removes `0.6` IMO. (Done: https://github.com/contao/contao/pull/2510 )

Toflar:
Fine for me either way

leofeyer (approved):

leofeyer:
Thank you @ausi.

Issue #2316, opened 2020-09-20
Tests should not rely on execution order


m-vo:
I sometimes get errors executing our tests - not as a suite but when running single classes/cases. There are some hidden dependencies between test cases or in other words: the execution order matters.

I think we should try to remove those dependencies: It can make writing new tests frustrating, because you now have to execute more then necessary to see if everything's green or not and you're possibly interfering with the original order creating new issues in seemingly unrelated tests. 

**How to reproduce:**

This yields no errors:
```bash
vendor/bin/phpunit --colors=always
```

But this will report over 200 errors. Most of them due to a missing `request_stack` service in the container. 
```bash
vendor/bin/phpunit --colors=always --order-by=random --random-order-seed=1 
```

Example of a single file causing errors when not tested in context:
```bash
vendor/bin/phpunit --colors=always core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php 
```

leofeyer:
Which Contao versions are affected?

m-vo:
I only tested it with the current master so far but I assume 4.4+ is affected.

ausi:
If it requires some amount of work, I don’t think we need to fix this in the 4.4 test suite.

leofeyer:
Contao 4.4 uses PHPUnit 5.7.27, which does not even support the `--order-by` flag, therefore I guess I will assign this issue to the `4.9` milestone.

m-vo:
I'm working on this. 

I will probably provide multiple PRs because of the various (rather independent) issues.

leofeyer:
Fixed in #4120.

PR #1918, opened 2020-07-07
Restore the old StringUtil logic (but compare on normalized strings)

m-vo:
see https://github.com/contao/contao/pull/1902#issuecomment-654896590

PR #1902, opened 2020-07-03
Use webmozart/path-util for all path operations

m-vo:
### Applied transformations

**Joins**
* `'$a/'.$b.'/c'` / `sprintf('%s/%s/c', $a, $b)` &rarr; `Path::join($a, $b, 'c')`

**Parent**
* `dirname($a)` &rarr; `Path::getDirectory($a)`
* `dirname($a).'/b'` &rarr; `Path::join($a, '../b')`

**Normalization**
* `strtr($a, '\\', '/')` &rarr; `Path::normalize($a)`
* simplified `Configuration#canonicalize()` by reusing `Path::canonicalize()`

**Absolute/relative paths**
* `$filesystem->makePathRelative($path, $root)` &rarr; `Path::makeRelative($path, $root)`
* `str_replace(strtr($root, '\\', '/').'/', '', strtr($path, '\\', '/'))` &rarr; `Path::makeRelative($path, $root)`
* `substr($path, \strlen($root) + 1)` &rarr; `Path::makeRelative($path, $root)`
* `0 === strncmp($p, 'system/modules/', 15)` &rarr; `Path::isBasePath('system/modules', $p)`
* `0 === strncmp($path, $dir.'/', \strlen($dir) + 1)` &rarr; `Path::isBasePath($dir, $path)`
* `0 === strpos($path, $dir.'/')` &rarr; `Path::isBasePath($dir, $path)`
* `$filesystem->isAbsolutePath($path)` &rarr; `Path::isAbsolute($path)` 

**File extension**
* `pathinfo($filename, PATHINFO_EXTENSION)` &rarr; `Path::getExtension($resource, true)` (also handles uppercase file ext. if not explicitly wrapping with `strtolower()`)
* `'.twig' === substr($path, -5)` &rarr; `'twig' === Path::getExtension($path, true)`

**General**
* no more `rtim($path, '/')` (joined paths are already canonicalized)
* I introduced variables wherever the same path was used more than once.
* Some bits are simplified (e.g. `SymlinksCommand#filterNestedPaths`) now that we can use the helpers.

## Scope

All non-legacy code excluding tests.

*Note:* Concatenated paths includung  `__DIR__ ` are left untouched (see https://github.com/contao/contao/pull/1902#discussion_r449650482).

PR #2516, opened 2020-11-21
Fix test order (1) - Never write into Fixtures dir

m-vo:
| Q                | A
| -----------------| ---
| Docs PR or issue | -

I'm working on test isolation. This part makes sure tests aren't reading temporary data from each other. This effectively means that a test should never write into the `Fixtures` dir (or anywhere else inside the source directories). If temporary files need to be created this now will happen inside the test's own `tmp` dir.

While working on these files, I'm also adjusting the following for consistency:

1) Refactoring path operations to use `Path`
1) Removing private property `$filesystem`  (if we want to share the `Filesystem` instance across tests, we should imo add it to `ContaoTestCase` as a static property)
1) Use `$filesystem` instead of `$fs` everywhere

Following changes needed for #2316 (but unrelated to writing files) are also part of this PR:

1) Reset template loader every time (was missing once, now in `tearDown`).
1) Do not share `ContaoCacheWarmer` instance.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
+++ b/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
@@ -12,6 +12,7 @@ namespace Contao;
 
 use Patchwork\Utf8;
 use Psr\Log\LogLevel;
+use Webmozart\PathUtil\Path;
 
 /**
  * Provides string manipulation methods
@@ -1112,10 +1113,12 @@ class StringUtil
 	 */
 	public static function stripRootDir($path)
 	{
-		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
+		// Compare normalized version of the paths
+		$projectDir = Path::normalize(System::getContainer()->getParameter('kernel.project_dir'));
+		$normalizedPath = Path::normalize($path);
 		$length = \strlen($projectDir);
 
-		if (strncmp($path, $projectDir, $length) !== 0 || \strlen($path) <= $length || ($path[$length] !== '/' && $path[$length] !== '\\'))
+		if (strncmp($normalizedPath, $projectDir, $length) !== 0 || \strlen($normalizedPath) <= $length || $normalizedPath[$length] !== '/')
 		{
 			throw new \InvalidArgumentException(sprintf('Path "%s" is not inside the Contao root dir "%s"', $path, $projectDir));
 		}
--- a/composer.json
+++ b/composer.json
@@ -118,6 +118,7 @@
         "twig/twig": "^2.7",
         "ua-parser/uap-php": "^3.9",
         "webignition/robots-txt-file": "^3.0",
+        "webmozart/path-util": "^2.0",
         "wikimedia/less.php": "^1.7"
     },
     "replace": {
--- a/composer.json
+++ b/composer.json
@@ -146,7 +146,7 @@
         "ext-fileinfo": "*",
         "bamarni/composer-bin-plugin": "^1.4",
         "composer/composer": "^1.0 || ^2.0",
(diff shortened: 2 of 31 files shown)


m-vo:
I had to backport the changes to `StringUtil#stripRootDir` from #1918 which fixed lots of tests on Windows. 

The ones that are still failing right now are a bit of a mystery for me, though. Maybe I'll have to check this PR out on windows and debug there. Or maybe someone else directly sees the problem? /cc @fritzmg :slightly_smiling_face:

leofeyer:
> 1. Refactoring path operations to use `Path`

Ok if the changes go into Contao 4.11 or 4.12 but not in Contao 4.9.

> 2. Removing private property `$filesystem`  (if we want to share the `Filesystem` instance across tests, we should imo add it to `ContaoTestCase` as a static property)

✅ 

> 3. Use `$filesystem` instead of `$fs` everywhere

I fail to see the improvement here…

> 1. Reset template loader every time (was missing once, now in `tearDown`).
> 2. Do not share `ContaoCacheWarmer` instance.

✅ 

> I had to backport the changes to StringUtil#stripRootDir from #1918

Another reason why we should probably only change this in Contao 4.11.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -1112,10 +1113,12 @@ public static function trimsplit($strPattern, $strString)
...
 	public static function stripRootDir($path)
 	{
-		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
+		// Compare normalized version of the paths
+		$projectDir = Path::normalize(System::getContainer()->getParameter('kernel.project_dir'));
+		$normalizedPath = Path::normalize($path);
 		$length = \strlen($projectDir);
 
-		if (strncmp($path, $projectDir, $length) !== 0 || \strlen($path) <= $length || ($path[$length] !== '/' && $path[$length] !== '\\'))
+		if (strncmp($normalizedPath, $projectDir, $length) !== 0 || \strlen($normalizedPath) <= $length || $normalizedPath[$length] !== '/')
```
leofeyer, line 1121: This change seems unrelated. Am I missing something?

Also – while we are at it – why are you not using `Path::isBasePath()`?

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
m-vo: It's needed because due to the normalization `\` &rarr; `/` together with legacy code paths using `\`, stripping the root dir could break on windows (analogous to #1902, https://github.com/contao/contao/pull/1902/commits/48b1f8998aca6a23c268927d6bf4666893421cb6).

 See this CI run: https://github.com/contao/contao/pull/2516/checks?check_run_id=2088691955

> Also – while we are at it – why are you not using Path::isBasePath()?

There are some edge cases to this. We fixed the same issue in https://github.com/contao/contao/pull/1918. Now the implementation should be identical to the upstream versions.

m-vo:
Thanks for fixing tests on windows @fritzmg! :heart:

fritzmg:
Never underestimate the value of Windows! There at least some errors don't fail silently as they do in *nix 😁 (regarding symlinks).

fritzmg (approved):
I am very much in favour of merging this PR into `4.9`, including all `Path` related refactors. It has been shown time and time again that _not_ using it leads to issues.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #2517, opened 2020-11-21
Fix test order (2) - Handle file scoped deprecations 

m-vo:
| Q                | A
| -----------------| ---
| Docs PR or issue | -

I'm working on test isolation. This part makes sure file scoped deprecations (= `trigger_error` outside the class definition) are only encountered in the static `setupBeforeClass` method while autoloading the file.

### DIFF
--- a/core-bundle/tests/Contao/GdImageTest.php
+++ b/core-bundle/tests/Contao/GdImageTest.php
@@ -19,6 +19,18 @@ use Contao\System;
 
 class GdImageTest extends TestCase
 {
+    /**
+     * @group legacy
+     *
+     * @expectedDeprecation Using the "Contao\GdImage" class has been deprecated %s.
+     */
+    public static function setUpBeforeClass(): void
+    {
+        parent::setUpBeforeClass();
+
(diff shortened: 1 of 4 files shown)


leofeyer:
Why do we need that? It feels quite unnatural to use the classes this way.

m-vo:
Because the deprecations aren't triggered within methods but when autoloading the classes. So they're only triggered once which either makes for unsilenced deprecations or failing tests if the order is turned around.

ausi (approved):

ausi:
Any reason why this wasn’t merged?

leofeyer:
I haven't had time to look into it yet. There are four PRs altogether and two of them are up for discussion.

ausi:
> There are four PRs altogether and two of them are up for discussion.

I think this part should be merged independently. ☺️

ausi:
> I think this part should be merged independently. ☺️

Changed my mind, we should merge #2863 instead 😅

m-vo:
Superseded by #2863.

PR #2518, opened 2020-11-21
Fix test order (3) - Reset singletons

m-vo:
| Q                | A
| -----------------| ---
| Docs PR or issue | -

I'm working on test isolation. This part adds the ability to reset the singleton instances. 

Without this the `Files` class would for instance store the `$projectDir` property and keep it forever although another test might have called `System::setContainer()` with another path in the meantime.

The following changes needed for #2316 are also part of this PR in order for the canonical order to pass:

1) Mocking models instead of real instances in the `Tests/Controller` namespace.
1) Mocking models instead of real instances in the `Tests/Fragment` namespace.

### DIFF
--- a/core-bundle/src/Framework/ContaoFramework.php
+++ b/core-bundle/src/Framework/ContaoFramework.php
@@ -19,10 +19,12 @@ use Contao\CoreBundle\Routing\ScopeMatcher;
 use Contao\CoreBundle\Security\Authentication\Token\TokenChecker;
 use Contao\CoreBundle\Session\LazySessionAccess;
 use Contao\Environment;
+use Contao\Files;
 use Contao\Input;
 use Contao\InsertTags;
 use Contao\Model\Registry;
 use Contao\RequestToken;
+use Contao\Session;
 use Contao\System;
 use Contao\TemplateLoader;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
@@ -119,6 +121,9 @@ class ContaoFramework implements ContaoFrameworkInterface, ContainerAwareInterfa
         Input::resetUnusedGet();
         InsertTags::reset();
         Registry::getInstance()->reset();
+        Files::reset();
+        Session::reset();
+        System::reset();
     }
 
     public function isInitialized(): bool
--- a/core-bundle/src/Framework/ContaoFramework.php
+++ b/core-bundle/src/Framework/ContaoFramework.php
@@ -117,11 +117,11 @@ class ContaoFramework implements ContaoFrameworkInterface, ContainerAwareInterfa
         }
 
         Environment::reset();
+        Files::reset();
         Input::resetCache();
         Input::resetUnusedGet();
         InsertTags::reset();
         Registry::getInstance()->reset();
-        Files::reset();
         Session::reset();
         System::reset();
     }
--- a/core-bundle/src/Framework/ContaoFramework.php
+++ b/core-bundle/src/Framework/ContaoFramework.php
@@ -18,6 +18,7 @@ use Contao\CoreBundle\Exception\RedirectResponseException;
 use Contao\CoreBundle\Routing\ScopeMatcher;
 use Contao\CoreBundle\Security\Authentication\Token\TokenChecker;
 use Contao\CoreBundle\Session\LazySessionAccess;
+use Contao\Database;
 use Contao\Environment;
 use Contao\Files;
 use Contao\Input;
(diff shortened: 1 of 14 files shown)


core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -329,6 +329,25 @@ public static function getContainer()
...
+					$singleton::reset();
+				}
+			}
+
+			self::$removedServiceIds = null;
+			self::$arrStaticObjects = array();
+			self::$arrSingletons = array();
+			self::$arrLanguages = array();
+			self::$arrLanguageFiles = array();
+			self::$arrImageSizes = array();
```
leofeyer: Can we be sure that this is not a BC break? `System::setContainer()` is used in the production code as well.

m-vo: It's a behavior change but also a bug to me. If you're re-setting the container to a new instance we should clear the related state. But maybe it would be better to test `$container !== static::$objContainer` instead, so we're gracefully ignoring if the same instance is set multiple times in a row.

Btw.: I think we're ever setting `System::setContainer` only once in production code anyway, do we?

core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -73,6 +73,14 @@ public static function getInstance()
 		return self::$objInstance;
 	}
 
+	/**
+	 * @internal
+	 */
+	public static function reset(): void
```
leofeyer, line 79: This method should also be called in `ContaoFramework::reset()`.

core-bundle/src/Resources/contao/library/Contao/Session.php
```diff
@@ -98,6 +98,14 @@ public static function getInstance()
 		return static::$objInstance;
 	}
 
+	/**
+	 * @internal
+	 */
+	public static function reset(): void
```
leofeyer, line 104: This method should also be called in `ContaoFramework::reset()`.

m-vo:
I'm now also unregistering and resetting the `ClassLoader`. :upside_down_face: 

Here is why: After correctly resetting `System` (and therefore unsetting the container), this caused lots of failures because the class loader was now called even with no framework booted. So this line tried to call something on `null`:

https://github.com/contao/contao/blob/69e7aa67b0402ab9abbb15a1436c36136344e0ba/core-bundle/src/Resources/contao/library/Contao/ClassLoader.php#L155

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -114,11 +116,15 @@ public function reset(): void
             return;
         }
 
+        ClassLoader::reset();
```
ausi, line 177: Did we have a misconception here? This reset doesn’t seem to uninitialize the framework which means after the first reset `ClassLoader::scanAndRegister()` will never be called again.

And `System::setContainer()` seems to have the same issue?

m-vo: @aschempp Should we 'uninitialize' the framework as well when calling `ContaoFramework::reset()`? Would sound sensible to me. :slightly_smiling_face:

aschempp: That's not fully possible, because initializing the framework will set some constants. What would you uninitialize?

m-vo: Sure, it won't work for constants and includes but should we still do the rest?

> What would you uninitialize?

All the class variables that are set in `initialize()`. We already do this for `$this->isFrontend` and `$this->adapterCache`, but not for `self::$initialized`, `$this->request`, `$this->hookListeners`.

Maybe we could even "un-iniSet" things but I'm not sure about that.

m-vo:
For the record: I'll never name a PR with a number in parenthesis at the end. Totally looks like 3 new messages all the time. :facepalm:  :smile:

core-bundle/src/Framework/ContaoFramework.php
ausi: The following are not considered “state” IMO (because set by the container or from `Config`) and should not be reset:

- `$this->requestStack`, `$this->scopeMatcher`, `$this->tokenChecker`, `$this->filesystem`, `$this->projectDir`, `$this->errorLevel`
- `$GLOBALS['TL_HOOKS']`,
- `error_reporting()`,
- `ini_set('date.timezone')`, `date_default_timezone_set()`

Needs to be reset (based on the request or method call):

- `$this->request`, `$this->isFrontend`, `$this->adapterCache`, 
- `self::$initialized`, 
- `$_SESSION` (if set to `LazySessionAccess` or `BE_DATA`, `FE_DATA` keys)
- `$GLOBALS['TL_LANGUAGE']`

> That's not fully possible, because initializing the framework will set some constants.

How about throwing an exception if the framework gets (re)initialized and the existing constants do not match the values that should get set (except `TL_START`)?

aschempp: sound like an idea. Should that also unset `$GLOBALS['TL_DCA']` and all the other known globals?

ausi: I don’t know, I think it might not make sense for all globals, but maybe for some.

SeverinGloeckle:
Lots of great stuff in this pull request! May I ask what the current status is here?

I would love to be able to cleanly reset `System::$objContainer` between unit tests, just like your `System::reset()` does.

If you need any assistance, I'd be happy to help.

m-vo:
> Lots of great stuff in this pull request! May I ask what the current status is here?

Missing time paired with a somewhat bottomless pit of legacy code… :see_no_evil: The old framework is only somewhat resettable, mostly due to singletons and constants. It could probably be improved a bit (thus this PR) but this will still be a lot of work.

> If you need any assistance, I'd be happy to help.

Feel absolutely free to base your work on this PR if you want to give it a try! :slightly_smiling_face:

SeverinGloeckle:
From my point of view, the current state looks good already. 

Since the adjustments primarily affect unit tests, I would propose a merge in 4.x in order to be able to use and test the resets in future unit tests. 

Or is there some context with open issues outside of the comments within this draft?

Note:
Before I came across this pull request, I was considering implementing the resets via utility methods in `ContaoTestCase`. 
However, it is much easier to implement a reset directly within the individual classes. Therefore, I like the approach in this draft better.



> Feel absolutely free to base your work on this PR if you want to give it a try! 🙂

I have added a pull request to merge the current 4.x branch into your draft to save you some conflict resolving.

leofeyer:
Fixed in #4120 now.

PR #2521, opened 2020-11-23
Do not try to generate fragments for generated fragments

aschempp:
If a fragment has multiple methods with different fragment type (e.g. a content element and a front end module), the first RegisterFragmentsPass creates a child definition that inherits all tags from the original. On the next run, the RegisterFragmentsPass would pick up the child definition and try to create duplicate fragments.

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php
@@ -83,10 +83,10 @@ class RegisterFragmentsPass implements CompilerPassInterface
         $command = $container->hasDefinition('contao.command.debug_fragments') ? $container->findDefinition('contao.command.debug_fragments') : null;
 
         foreach ($this->findAndSortTaggedServices($tag, $container) as $reference) {
-            // If a controller has multiple methods for different fragment type (e.g. a content element
-            // and a front end module), the first RegisterFragmentsPass creates a child definition that
-            // inherits all tags from the original. On the next run, the RegisterFragmentsPass would
-            // pick up the child definition and try to create duplicate fragments.
+            // If a controller has multiple methods for different fragment types (e.g. a content
+            // element and a front end module), the first pass creates a child definition that
+            // inherits all tags from the original. On the next run, the pass would pick up the
+            // child definition and try to create duplicate fragments.
             if (0 === strpos((string) $reference, 'contao.fragment._')) {
                 continue;
             }
--- a/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php
@@ -83,6 +83,14 @@ class RegisterFragmentsPass implements CompilerPassInterface
         $command = $container->hasDefinition('contao.command.debug_fragments') ? $container->findDefinition('contao.command.debug_fragments') : null;
 
         foreach ($this->findAndSortTaggedServices($tag, $container) as $reference) {
+            // If a controller has multiple methods for different fragment type (e.g. a content element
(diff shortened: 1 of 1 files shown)


core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php
```diff
@@ -83,6 +83,14 @@ protected function registerFragments(ContainerBuilder $container, string $tag):
         $command = $container->hasDefinition('contao.command.debug_fragments') ? $container->findDefinition('contao.command.debug_fragments') : null;
 
         foreach ($this->findAndSortTaggedServices($tag, $container) as $reference) {
+            // If a controller has multiple methods for different fragment type (e.g. a content element
+            // and a front end module), the first RegisterFragmentsPass creates a child definition that
+            // inherits all tags from the original. On the next run, the RegisterFragmentsPass would
+            // pick up the child definition and try to create duplicate fragments.
+            if (0 === strpos((string) $reference, 'contao.fragment._')) {
```
Toflar, line 90: Wouldn't it be safer to keep track of an array of fragments for which the child definition has already been created?

aschempp: that was actually my first draft, but it is much more complicated because
 1. that's not a service, we create multiple instances in the bundle
 2. we would need to _reset_ that, but again, not being a service makes that complicated.

Toflar: This prefix thing feels shaky to me. Why is there a second run anyway? Every pass is only passed once?

aschempp: A controller has two tags:
 - `contao.content_element`
 - `contao.frontend_module`

The bundle runs the pass for each fragment type
https://github.com/contao/contao/blob/master/core-bundle/src/ContaoCoreBundle.php#L89-L103

The original tags are duplicated to the child definitions because we need other tags (e.g. for monolog logger)
https://github.com/contao/contao/blob/master/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php#L117

Toflar: Ah, now I think I've understood the problem. Hmm, so the `RegisterFragmentsPass` itself is not a service which means we cannot store the fact that we've already created a child definition. But the container is. What do you think about adding another tag and checking for this then? E.g. when done creating the child class, calling `addTag(self::DUPLICATE_CHECK_TAG_NAME)` and check for this instead of a service id prefix?

aschempp: I don't see the point. The service ID prefix is unique to services generated by the pass, it would be a problem if its not. Adding a tag does not make it "more unique", anyone could abuse the tag the same way they can abuse the prefix …

Toflar: It's just nicer. Anyway, I'm fine with the prefix as well. However, we need it to be a class constant and tests.

leofeyer (changes requested):
Originally posted here: https://github.com/contao/contao/pull/2521/files#r529254297

aschempp:
We don't use class constants for anything atm. Using a private constant for something that is used once or twice does not seem useful to me (it would be different if it was public, but it's explicitly internal behaviour!).

I also tried to write a test for that, but I was not able to. The only valid test would need to mock the container to verify `setDefinition` is not called multiple times, but mocking the container breaks like everything in the run loop, or would require to reverse engineer the container and rebuild everything just for the test…

leofeyer (dismissed):

leofeyer (approved):

Issue #2197, opened 2020-08-26
HTML-tags in the field "Bildunterschrift" are not converted in HTML in Contao 4.10


jochi44:
**Affected version**

Contao 4.10

**Description**

In Contao 4.10 HTML-tags in the field "Bildunterschrift" (Inhaltselement "Text" -> "Ein Bild hinzufügen" -> "Metadateien überschreiben") are not converted in HTML anymore.

For example do I use something like this quite often for a "Bildunterschrift":
`Quelle: <a href="https://abc.to" target="_blank">abc.to</a>`

Unfortunately in Contao 4.10 the frontend-output shows
`Quelle: <a href="https://abc.to" target="_blank">abc.to</a>`
instead of the HTML
`Quelle: abc.to`

Im not shure but think the HTML-conversion worked until Version 4.9.

ausi:
Independent from the issue itself: You can use `{{link::https://abc.to}}` for this kind of use cases.

asaage:
I was not aware, this was even allowed by defaut - but it is...
https://github.com/contao/contao/blob/master/core-bundle/src/Resources/contao/dca/tl_content.php#L300

jochi44:
> Independent from the issue itself: You can use `{{link::https://abc.to}}` for this kind of use cases.

Thank you, @ausi , for this information.
Unfortunately I don't see a possibility to add the "target=_blank"-argument when using this insert tag.
The insert tag `{{link_target::*}}` does only work when linking a Contao intern site (if I read the Contao docs correctly).

PR #2527, opened 2020-11-27
Use a textarea for the image caption field

Toflar:
An image caption shouldn't be limited to 255 chars only, that makes no sense to me.

That's the result:

Bildschirmfoto 2020-11-27 um 13 12 19


fritzmg (dismissed):

m-vo:
What should happen to newlines in this field?

Toflar:
They’re kept? Why would I care?

asaage:
I wouldn't vote against it but that would be the only textarea in the BE without richtext-editor...
mabe revisit https://github.com/contao/contao/issues/2197
I personally would rather use a ce_text and add an image if there is so much text - just my opinion.

bytehead (dismissed):

asaage:
What about "overwrite metadata" in a ce_image?

leofeyer (approved):

bytehead (approved):

leofeyer:
Thank you @Toflar.

PR #2528, opened 2020-11-30
Increase the undo period

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

I really think the default undo period of 24h is not a good default (which is why I think it should be changed in 4.9 LTS too).

For me, there are two use cases our current undo system can cover:

1. The immediate "Oh, I deleted something which I shouldn't have" case: 24 hours are enough
2. The customer did something over the weekend and you get a report on Monday that something was deleted: 24 hours are likely not enough.

So 30 days by default seem totally okay to me. It makes no sense to store undo entries that date back a year or so.


bytehead (approved):

fritzmg (approved):

aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #2200, opened 2020-08-26
Load the default labels in the loadDcaFiles() method

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

In Contao **4.9** we do not have to specify

```php
'label' => &$GLOBALS['TL_LANG']['tl_module']['foo'],
```

any more in our DCAs, if the name of the translation is equal to the name of the field, because the default labels are automatically loaded from `$GLOBALS['TL_LANG']` within `\Contao\DcaLoader`.

However, this breaks under certain circumstances. For instance, if you install `codefog/contao-news_categories`, then any custom fields you add via `contao/dca/tl_content.php` or `contao/dca/tl_module.php` will not use the translation from `contao/languages/` automatically.

The reason is the following (example with a new field in `tl_module`):

1. The Contao Backend will first call `Controller::loadDataContainer('tl_module')`.
2. This loads all the `tl_module.php` files of all extensions in the defined order, with `contao/dca/tl_module.php` being last.
3. When `vendor/codefog/contao-news_categories/…/dca/tl_module.php` is called, it contains another call to `Controller::loadDataContainer('tl_content')` (because it depends on some data from that data container).
4. This loads all the `tl_content.php` files of all extensions in the defined order.
5. When `vendor/codefog/contao-news_categories/…/dca/tl_contao.php` is called, it contains another call to `Controller::loadDataContainer('tl_module')` (because it depends on some data from that data container).
6. In _that_ call, within `\Contao\DcaLoader::load`, `$this->loadDcaFiles();` is skipped, because it was already executed, however, `$this->addDefaultLabels();` is **not** skipped, because it was not yet executed at this point.
7. `\Contao\DcaLoader::addDefaultLabels` will then load the default labels for `tl_module`, but only for fields that are known at this point. i.e. any fields from DCA files that came _after_ `codefog/contao-news_categories` are not included.

You can also replicate the problem this way:

```php
// contao/dca/tl_module.php
use Contao\Controller;
use Contao\CoreBundle\DataContainer\PaletteManipulator;

Controller::loadDataContainer('tl_content');

$GLOBALS['TL_DCA']['tl_module']['fields']['foo'] = [
    'inputType' => 'text',
    'eval' => ['tl_class' => 'w50'],
];

PaletteManipulator::create()
    ->addField('foo', 'title_legend', PaletteManipulator::POSITION_APPEND)
    ->applyToPalette('navigation', 'tl_module')
;
```
```php
[...]

Issue #71, opened 2018-09-19
Custom navigation not possible for root pages


akroii:
**Affected version(s)**
4.4,4.6

**Description**  
If I use a custom navigation and select all my root-pages, the frontend navigation is empty.
I wonder why there is a separate [bundle](https://github.com/heimrichhannot/contao-rootnav) for this. Because I am able to check root pages in backend.
![bildschirmfoto 2018-09-19 um 16 52 44](https://user-images.githubusercontent.com/8830861/45761430-8cb42800-bc2c-11e8-983b-c45d0565491b.png)

**How to reproduce**  
1. Create new module -> custom navigation
2. select root-"pages" in navigation settings
3. embed your custom navigation module in frontend

xchs:
See https://github.com/contao/core-bundle/issues/1641

PR #2536, opened 2020-12-02
Ensure language files are loaded when adding default labels

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

In Contao 4.9 you do not have to provide a dedicated `label` key for each field of your data container in the data container array, as `DcaLoader::loadDcaFiles` will automatically add the labels from `$GLOBALS['TL_LANG']` when present.

However, this will fail if `Controller::loadDataContainer` or `DcaLoader::load` is executed for a DCA _before_ its language files have been loaded. Thus `DcaLoader::addDefaultLabels` also needs to ensure that the language files have already been loaded.

This might happen in unexpected circumstances. Suppose you have a `tl_foo` DCA with a `FooModel`. If you now execute the following somewhere:

```
$fooModel = new FooModel();

System::loadLanguageFile('tl_foo');
Controller::loadDataContainer('tl_foo');
```
then the labels within `$GLOBALS['TL_DCA']['tl_foo']['fields']` will be empty, because `\Contao\Model::__construct()` already loads the DCA of `tl_foo`, but before the language files of `tl_foo` have been loaded in this case.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/DcaLoader.php
+++ b/core-bundle/src/Resources/contao/library/Contao/DcaLoader.php
@@ -138,6 +138,8 @@ class DcaLoader extends Controller
 	 */
 	private function addDefaultLabels($blnNoCache)
 	{
+		$this->loadLanguageFile($this->strTable, null, $blnNoCache);
+
 		// Return if there are no labels
 		if (!isset(static::$arrLanguageFiles[$this->strTable]))
 		{


ausi (approved):

leofeyer (changes requested):
This is wrong, because it is not the responsibility of the DcaLoader to load language files. Also, we must not enforce loading the language files, because there are use cases in which we do not want to load them (e.g. in the DcaExtractor).

IMHO, the "correct" fix is to call `addDefaultLabels()` in line 71 so it is executed even if the DCA has been loaded before. The method will return early once the labels have been added.

aschempp (changes requested):
Imho the correct fix would be to not check if a label exists (https://github.com/contao/contao/pull/2536/files#diff-4600f89ff1d6f53ddc624cc0976a50e991196ebbad78e52aa3f6837af02f13aeR172). Thats the same a DCA does, it links to the language array by reference, so the labels are also assigned when loaded later.

fritzmg:
@leofeyer hm, yeah so this is a bug that was introduced with https://github.com/contao/contao/pull/2200, but moving it back to its old place would reintroduce that bug.

@aschempp you mean in addition to moving it back to line #71? That alone won't work (see #2200).

leofeyer:
Ok, so we have to figure out the correct solution. But this premise remains:

> it is not the responsibility of the DcaLoader to load language files. Also, we must not enforce loading the language files, because there are use cases in which we do not want to load them (e.g. in the DcaExtractor).

aschempp:
https://github.com/contao/contao/pull/2540 is my suggestion. Maybe this makes it clearer.

fritzmg:
> But this premise remains:
> 
> > it is not the responsibility of the DcaLoader to load language files. Also, we must not enforce loading the language files, because there are use cases in which we do not want to load them (e.g. in the DcaExtractor).

Thinking about it, I am not sure if we can do that actually? I mean the `addDefaultLabels` method adjusts the DCA according to data within `$GLOBALS['TL_LANG']`.

leofeyer:
See #2540.

ausi:
@fritzmg regarding your example from https://github.com/contao/contao/pull/2540#issuecomment-740709228 : you must not use references in your language file, then everything should work fine.

https://3v4l.org/QBXrr shows why this doesn’t work with references.

fritzmg:
@ausi but this has been working before and I have seen such references in other extensions as well. These might break then under certain circumstances as well, presumably.

ausi:
> but this has been working before

As far as I understand references in PHP, this could have never worked if the DCA was loaded before your language file.

I think we either have to define that using references in language files is wrong or otherwise we would have to make sure that the language files are always loaded before the DCA (but this would make using references in the DCA labels useless).

fritzmg:
> I think we either have to define that using references in language files is wrong or otherwise we would have to make sure that the language files are always loaded before the DCA.

Agreed. I am actually fine with saying that using `&$GLOBALS['TL_LANG']` within language files themselves is not really supported. After all, if you use XLF or YAML, you wouldn't be able to do that anyway. However, it is pretty convenient to not have to translate everything and rather rely on existing translations instead.

> (but this would make using references in the DCA labels useless)

Why would that be the case? If `$GLOBALS['TL_LANG']` is filled before the DCA, then refrences to TL_LANG or within TL_LANG should always work? After all TL_DCA references TL_LANG, but never the other way around.

ausi:
> However, it is pretty convenient to not have to translate everything and rather rely on existing translations instead.

This still works by assigning/copying the value without a reference from the other language file:

```php
\Contao\System::loadLanguageFile('tl_member');
$GLOBALS['TL_LANG']['tl_foo']['firstname'] = $GLOBALS['TL_LANG']['tl_member']['firstname'];
```

> > (but this would make using references in the DCA labels useless)
> 
> Why would that be the case?

If *language files are always loaded before the DCA* then *after loading the DCA $GLOBALS['TL_LANG'] will not change* and so there is no need in the DCA to use a reference.

> Agreed. I am actually fine with saying that using `&$GLOBALS['TL_LANG']` within language files themselves is not really supported.

👍 We can close this issue then I think. Or should we document this somewhere?

PR #2538, opened 2020-12-03
Fix the execution order of DCA callbacks registered via annotations or service tags

qzminski:
Up until now the priority of DCA callbacks registered via annotations or service tags was irrelevant:

```
tl_page => [
    list.operations.articles.button_callback => [
        0 => […],
        16 => […],
        -16 => […],
    ],    
],
```

This PR should fix it and put in correct order:

```
tl_page => [
    list.operations.articles.button_callback => [
        16 => […],
        0 => […],
        -16 => […],
    ],    
],
```

[According to the documentation](https://docs.contao.org/dev/framework/dca/#using-service-tagging) the callback with highest priority should be executed first.

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/DataContainerCallbackPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/DataContainerCallbackPass.php
@@ -86,6 +86,8 @@ class DataContainerCallbackPass implements CompilerPassInterface
             $serviceId,
             $this->getMethod($attributes, $class, $serviceId),
         ];
+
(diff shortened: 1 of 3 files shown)


bytehead (approved):

richardhj (approved):

aschempp (changes requested):
They are sorted in https://github.com/contao/contao/blob/master/core-bundle/src/EventListener/DataContainerCallbackListener.php#L90 because they are merged with existing items.

qzminski:
@aschempp you are actually right, and now I am no longer able to reproduce the problem with overriding a singleton button callback. Must have been some mistake on my side then.

Please consider this PR as irrelevant then.

PR #2543, opened 2020-12-04
Add a missing PHPStan dependency

leofeyer:


m-vo (approved):
Good catch. Just stumbled across this as well when trying to isolate the tooling.

PR #2547, opened 2020-12-07
Fix the compatibility with scssphp 1.4

### LINKED ISSUE #2495: New version of ScssPhp breaks scss import paths for Contao
In Contao 4.9 the import of relative paths over multiple bundles breaks now, because the path handling behaviour in https://github.com/scssphp/scssphp has changed in their newest version 1.4. If the current file path is not attached (additionally to `$objCompiler->setImportPaths`) it will always be hardcoded/relative to the webroot. 

If you for example include a variable from another bundle:
```SCSS
@import '../../../../../magmell-basic/src/Resources/public/css/variables';
```
the path gets resolved to `htdocs/web/../../../../../magmell-basic/src/Resources/public/css/_variables.scss` which results in an access violation for basedir restrictions (we would try to access 3 levels below basedir).

 The developer recognizes it as breaking change, but suggests a code fix, which will actually fix the problem to Combiner.php line 419: 
```PHP
return $this->fixPaths($objCompiler->compile($content, $this->strRootDir . '/' . $arrFile['name']), $arrFile);
```

With everything updated I can confirm that this fixes the problem - but don't know how to connected the necessary fix to only be active in ScssPhp versions of 1.4+

See also https://github.com/scssphp/scssphp/issues/271

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2495

### ToDos

- [x] Revert https://github.com/contao/conflicts/commit/820074e93850364cc3beb329d8ff0472a87b20c3 once this got merged and released (4.9 and upstream).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Combiner.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Combiner.php
@@ -416,7 +416,7 @@ class Combiner extends System
 				$objCompiler->setSourceMap(Compiler::SOURCE_MAP_INLINE);
 			}
 
-			return $this->fixPaths($objCompiler->compile($content), $arrFile);
+			return $this->fixPaths($objCompiler->compile($content, $this->strRootDir . '/' . $arrFile['name']), $arrFile);
 		}
 
 		$strPath = \dirname($arrFile['name']);


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2550, opened 2020-12-08
Do not index preview URLs for searching

### LINKED ISSUE #2507: Preview URLs in the search index
**Affected version(s)**

Contao 4.9+

**Description**

Preview URLs are currently indexed by our default indexer and shown in the search results. To fix this, we should set the `X-Robots-Tag` header as @Toflar suggested here:

Bildschirmfoto 2020-11-19 um 16 45 16

Also see: https://github.com/contao/contao/pull/1008#issuecomment-557467723

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2507
| Docs PR or issue | -

@Toflar: If the `X-Robots-Tag` header or the meta robots tag contains `noindex`, the listener will not pass the document to the indexer at all. This means that potentially existing entries will not be deleted. Is this OK?

### DIFF
--- a/core-bundle/src/EventListener/SearchIndexListener.php
+++ b/core-bundle/src/EventListener/SearchIndexListener.php
@@ -71,7 +71,26 @@ class SearchIndexListener
             return;
         }
 
-        $document = Document::createFromRequestResponse($request, $event->getResponse());
+        $response = $event->getResponse();
+
+        // Do not index if the X-Robots-Tag header contains "noindex"
+        if (false !== strpos($response->headers->get('X-Robots-Tag', ''), 'noindex')) {
+            return;
+        }
+
+        $document = Document::createFromRequestResponse($request, $response);
+
+        try {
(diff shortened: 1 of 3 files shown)


fritzmg:
> This means that potentially existing entries will not be deleted. Is this OK?

imho that's ok, as it was always like this. We could however implement a migration that simply deletes any URLs with `/preview.php` from the index?

Toflar:
That's okay to me too, just delete the search index manually. Should be done when updating or from time to time anyway. I don't think we need a migration for this. There's always something that could affect existing entries (e.g. if we change the output of a content element and add some `<!-- indexer:stop -->` comments etc.).
So either there is one that just always deletes the search index on every `contao:migrate` call or we just ignore it. If someone wants to do that on every deployment they can, as we have commands for that. 
So I'm fine with this PR, approving.

Toflar (approved):

PR #2551, opened 2020-12-08
Fix the cache tag invalidation

### LINKED ISSUE #2137: Cache tag invalidation
As discussed in https://github.com/contao/contao/pull/2103#issuecomment-671283944, we are currently always invalidating both the specific tag and the general "category" tag (e.g. `contao.db.tl_news.61` and `contao.db.tl_news`), which kind of renders the specific tag useless. Maybe we can improve this?

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2137
| Docs PR or issue | -

This is currently a proof of concept that should work for news and events. @Toflar @ausi Can you confirm that my implementation is what you had in mind?

### TODO

* [x] Add a general tag without ID for the top parent record
* [x] Implement cache tagging for the other elements and modules

### DIFF
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -10,8 +10,6 @@
 
 namespace Contao;
 
-use FOS\HttpCache\ResponseTagger;
-
 /**
  * Provide methods to get all events of a certain period from the database.
  *
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -284,7 +284,8 @@ abstract class Events extends Module
 		// Tag the event (see #2137)
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			System::getContainer()->get('fos_http_cache.http.symfony_response_tagger')->addTags(array('contao.db.tl_calendar_events.' . $objEvents->id));
+			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
+			$responseTagger->addTags(array('contao.db.tl_calendar_events.' . $objEvents->id));
 		}
 
 		// Store raw data
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -283,13 +283,21 @@ abstract class Events extends Module
 			}
 		}
 
-		// Tag the response
+		// Tag the event and its child content elements (see #2137). Calendars
+		// will be tagged separately in the related front end modules.
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			/** @var ResponseTagger $responseTagger */
-			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
-			$responseTagger->addTags(array('contao.db.tl_calendar_events.' . $objEvents->id));
-			$responseTagger->addTags(array('contao.db.tl_calendar.' . $objEvents->pid));
+			$arrTags = array('contao.db.tl_calendar_events.' . $objEvents->id);
+
+			if ($objElements = ContentModel::findPublishedByPidAndTable($objEvents->id, 'tl_calendar_events'))
+			{
+				foreach ($objElements as $objElement)
+				{
+					$arrTags[] = 'contao.db.tl_content.' . $objElement->id;
+				}
+			}
+
+			System::getContainer()->get('fos_http_cache.http.symfony_response_tagger')->addTags($arrTags);
 		}
 
 		// Store raw data
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -281,21 +281,10 @@ abstract class Events extends Module
 			}
 		}
 
-		// Tag the event and its child content elements (see #2137). Calendars
-		// will be tagged separately in the related front end modules.
+		// Tag the event (see #2137)
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			$arrTags = array('contao.db.tl_calendar_events.' . $objEvents->id);
-
-			if ($objElements = ContentModel::findPublishedByPidAndTable($objEvents->id, 'tl_calendar_events'))
-			{
-				foreach ($objElements as $objElement)
-				{
-					$arrTags[] = 'contao.db.tl_content.' . $objElement->id;
-				}
-			}
-
-			System::getContainer()->get('fos_http_cache.http.symfony_response_tagger')->addTags($arrTags);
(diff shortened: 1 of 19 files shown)


calendar-bundle/src/Resources/contao/classes/Events.php
```diff
@@ -283,13 +281,21 @@ protected function addEvent($objEvents, $intStart, $intEnd, $intBegin, $intLimit
...
+		// will be tagged separately in the related front end modules.
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			/** @var ResponseTagger $responseTagger */
-			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
-			$responseTagger->addTags(array('contao.db.tl_calendar_events.' . $objEvents->id));
-			$responseTagger->addTags(array('contao.db.tl_calendar.' . $objEvents->pid));
+			$arrTags = array('contao.db.tl_calendar_events.' . $objEvents->id);
+
+			if ($objElements = ContentModel::findPublishedByPidAndTable($objEvents->id, 'tl_calendar_events'))
```
Toflar: Why are you tagging the content elements here? They tag themselves in `Controller::getContentElement()`.

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -238,13 +237,20 @@ protected function compile()
...
+		// Tag the event and its child content elements (see #2137)
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			/** @var ResponseTagger $responseTagger */
-			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
-			$responseTagger->addTags(array('contao.db.tl_calendar_events.' . $objEvent->id));
-			$responseTagger->addTags(array('contao.db.tl_calendar.' . $objEvent->pid));
+			$arrTags = array('contao.db.tl_calendar_events.' . $objEvent->id);
+
+			if ($objElements = ContentModel::findPublishedByPidAndTable($objEvent->id, 'tl_calendar_events'))
```
Toflar: Same here, I don't think you need to tag the content elements.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1312,6 +1308,66 @@ public function invalidateCacheTags()
 		$cacheManager->invalidateTags($tags);
 	}
 
+	private function addPtableTags($strTable, $intId, &$tags)
+	{
+		if (empty($GLOBALS['TL_DCA'][$strTable]['config']['ptable']))
+		{
+			return;
```
Toflar, line 1317: Here would be the general tag, right?

Toflar (commented):
Looks good, except for the content element stuff. I've added my reviews.
Smartass move with the recursive calls there in DataContainer 👍 :)

calendar-bundle/src/Resources/contao/classes/Events.php
leofeyer: Ah, nice. 😎 Adjusted in 4b61ac2659d1dbb4d2809edd52acb6c9d468c4b0.

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
leofeyer: Adjusted in 4b61ac2659d1dbb4d2809edd52acb6c9d468c4b0.

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: Yes. Added in 5b4a03d8a5d5b6ea8b32c5375832499276e3eb1f.

Toflar:
Looks all correct to me now, ready to be applied to the rest. But let's maybe wait for @ausi's feedback.

comments-bundle/src/Resources/contao/classes/Comments.php
```diff
@@ -43,10 +42,9 @@ public function addCommentsToTemplate(FrontendTemplate $objTemplate, \stdClass $
 
 		$objTemplate->comments = array(); // see #4064
 
-		// Tag the response
+		// Tag the comment (see #2137)
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
-			/** @var ResponseTagger $responseTagger */
 			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
 			$responseTagger->addTags(array(sprintf('contao.comments.%s.%s', $strSource, $intParent)));
```
leofeyer, line 49: @Toflar Did you use a different namespace here on purpose? Or should this be `contao.db.`?

Toflar: This is different on purpose because of the source. Also see https://github.com/contao/contao/blob/master/comments-bundle/src/Resources/contao/dca/tl_comments.php#L788

leofeyer: I am not sure whether this is correct. If I delete a comment in the back end, Contao will automatically invalidate `['contao.db.tl_comments', 'contao.db.tl_comments.'.$id]`, so would it not make more sense to add a tag for each comment instead of `contao.comments.tl_news.12`?

Toflar: If you're the first one to comment on `tl_news.12`? How is that invalidated when the comment is published?

leofeyer: True. So we have to keep tagging the source record. 👍🏻 

Do you still want me to add tags for each individual comment (`contao.db.tl_comment.ID`) in case someone wants to build a "latest comments" module or something alike? Because there is no hook in the `addCommentsToTemplate()` method which could be used to add the functionality in a third-party bundle.

Toflar: I see, that makes sense I guess?

Toflar (approved):
LGTM. Would still love to have @ausi's feedback before merge, though.

comments-bundle/src/Resources/contao/classes/Comments.php
leofeyer: Added in 601247f2eedcfde9d2cc95bd1b764608b2fc6b26.

Last question: Since `tl_comments.source` always relates to a DB table, why do we not use the `contao.db.` prefix instead of the `contao.comments.` prefix? After all, it is a kind of "dynamic ptable" mechanism here as well.

Toflar: I don't know...I just went for this and nobody complained :) I wouldn't change it though as someone might already rely on it.

leofeyer: Valid point. I would still like to change it in one of the next feature releases. But I will create a separate ticket for that.

ausi (approved):
Nice Work! ❤

PR #2553, opened 2020-12-09
Strip port numbers from root page domains

### LINKED ISSUE #2214: Generated page urls contains host information twice
**Affected version(s)**

Contao 4.10.0

**Description**

I have a Contao setup where a custom port is used and as part of the root page dns (`PageModel::domain`). Contao isn't able to generate valid page urls anymore. A generated page url for the root page dns setting `localhost:8080` looks like `http://localhost:8080/localhost:8080:8080/example.html`

The issue seems to be be related to the host tokens of the compiled route of a page. Contao contains the port in the route host, where the `UrlGenerator` of symfony, which is used by the `PageUrlGenerator` of Contao, does not contain the port information and so a change of the reference type happens.

https://github.com/symfony/symfony/blob/69f8eaeec548566a25b5959b8f116e44cf2fe126/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L264-L269

It only happens for relative urls and not for urls linking to another root page.

I do not have a clue if this a Contao specific issue or a general Symfony routing issue.

/cc @frontendschlampe

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2214
| Docs PR or issue | -

Tested against https://regex101.com/r/jU0OrQ/2

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -1289,9 +1289,9 @@ class tl_page extends Backend
 
 		// The first part will match IPv6 addresses in square brackets. The
 		// second part will match domain names and IPv4 addresses.
-		preg_match('#(?:^(?:[a-z]+]://)?(\[[0-9a-f:]+]))|(?:^(?:[a-z]+://)?([\pN\pL._-]*))#ui', $varValue, $matches);
+		preg_match('#^(?:[a-z]+://)?(\[[0-9a-f:]+]|[\pN\pL._-]*)#ui', $varValue, $matches);
 
-		return Idna::encode($matches[2] ?? $matches[1]);
+		return Idna::encode($matches[1]);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Toflar:
Why not adding those tests to the already existing `ValidatorTest` :D

leofeyer:
Because we are extracting a substring here and not validating anything.

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1282,7 +1282,16 @@ public function loadDns($varValue)
...
+		if (!$varValue)
+		{
+			return '';
+		}
+
+		// The first part will match IPv6 addresses in square brackets. The
+		// second part will match domain names and IPv4 addresses.
+		preg_match('#(?:^(?:[a-z]+]://)?(\[[0-9a-f:]+]))|(?:^(?:[a-z]+://)?([\pN\pL._-]*))#ui', $varValue, $matches);
+
+		return Idna::encode($matches[2] ?? $matches[1]);
```
ausi:
```suggestion
		preg_match('#^(?:[a-z]+://)?(\[[0-9a-f:]+]|[\pN\pL._-]*)#ui', $varValue, $matches);

		return Idna::encode($matches[1]);
```

leofeyer: 🦊

Toflar (approved):

ausi (approved):

fritzmg (approved):

PR #2554, opened 2020-12-10
Support native transport and other mailer transport options

### LINKED ISSUE #2164: Set MAILER_DSN default to native://default
This is a reminder issue for myself in order to set the `MAILER_DSN` env var by default to `native://default`, once Symfony 5.2 has been released: https://github.com/symfony/symfony/pull/36131

This will configure the Mailer to use the "native configuration" from the `php.ini` - which in turn will reduce the likeliness of having to define an SMTP server yourself on some hosting environments. The reason being that apparently on some hosting environments using `sendmail` would still work - but not with the default sendmail command parameters that the Swiftmailer Bundle uses for instance. So it would work with `/usr/sbin/sendmail -t` instead of `/usr/sbin/sendmail -bs` for example.

Under the assumption this is correctly configured by the hoster in the `php.ini`, sending mails should then work out of the box with Contao 4.11 in such cases, without the need to either set the sendmail command parameters or define an SMTP server yourself.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2164
| Docs PR or issue | -

This PR adds support for the `native://default` mailer transport, which uses the PHP `sendmail_path` configuration. Some notes:

* This feature is available in Symfony `5.2.0` and up. Since the `symfony/mailer` dependency is independent from other Symfony packages, it can be allowed to be installed, while everything else stays at Symfony `4.4.*`.
* Since the actual transport factory for the native transport is however defined by the `symfony/framework-bundle`, an additional compiler pass is necessary, to add said transport factory to the container, if supported and not already present. This can be removed once the core requires at least Symfony `5.2.*`.
* Changing the `symfony/mailer` dependency to include `5.2.*` also allows you to use [additional options](https://github.com/symfony/symfony/pull/37432) for your transport (like a local domain for an SMTP relay or a custom sendmail command).

It would be important to include this in the upcoming Contao 4.11 version, as otherwise there exists no possibility of using a custom sendmail command and thus you are stuck with the hardcoded one.


core-bundle/src/DependencyInjection/Compiler/AddNativeTransportFactoryPass.php
```diff
@@ -0,0 +1,45 @@
...
+            ->setClass(NativeTransportFactory::class)
+            ->addTag('mailer.transport_factory')
+        ;
+
+        $container->setDefinition('mailer.transport_factory.native', $definition);
+    }
+
+    private function supportsNativeMailer(): bool
+    {
+        return version_compare(PackageUtil::getNormalizedVersion('symfony/mailer'), '5.2.0', '>=');
```
ausi: I think it would be better to use `class_exists(NativeTransportFactory::class)` here.

ausi (dismissed):

aschempp (dismissed):

Toflar (dismissed):

composer.json
```diff
@@ -179,7 +179,7 @@
             "contao/newsletter-bundle": "Contao\\NewsletterBundle\\ContaoManager\\Plugin"
         },
         "symfony": {
-            "require": "4.4.*"
+            "require": "4.4.* || 5.2.*"
```
leofeyer:
```suggestion
            "require": "4.4.* || ^5.2"
```

This will also allow Symfony packages that are no direct requirements in version 5.3.

core-bundle/composer.json
```diff
@@ -139,7 +139,7 @@
         },
         "contao-manager-plugin": "Contao\\CoreBundle\\ContaoManager\\Plugin",
         "symfony": {
-            "require": "4.4.*"
+            "require": "4.4.* || 5.2.*"
```
leofeyer:
```suggestion
            "require": "4.4.* || ^5.2"
```

core-bundle/src/DependencyInjection/Compiler/AddNativeTransportFactoryPass.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Component\Mailer\Transport\NativeTransportFactory;
+
+/**
+ * @internal
+ */
+class AddNativeTransportFactoryPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('mailer.transport_factory.native') || !$this->supportsNativeMailer()) {
```
leofeyer: What is the advantage of encapsulating a single `class_exists()` call in a separate method? It is never re-used IIRC.

manager-bundle/composer.json
```diff
@@ -77,7 +77,7 @@
         },
         "contao-manager-plugin": "Contao\\ManagerBundle\\ContaoManager\\Plugin",
         "symfony": {
-            "require": "4.4.*"
+            "require": "4.4.* || 5.2.*"
```
leofeyer:
```suggestion
            "require": "4.4.* || ^5.2"
```

leofeyer (changes requested):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2555, opened 2020-12-10
Do not return empty strings in the link_url insert tag

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2395
| Docs PR or issue | -


fritzmg:
Hm, what about other insert tags, that might generate a link to the start page? e.g. `{{news_url::*}}` and `{{event_url::*}}`, if the news or event points to the start page (for whatever reason). Shouldn't `Page::generateFrontendUrl` generally return `./` instead, to cover all possibilities?

leofeyer:
News and events will always have an auto_item and can therefore never generate an empty URL.

fritzmg:
> News and events will always have an auto_item and can therefore never generate an empty URL.

If the `source` is `internal` for example, the `{{news_url::*}}` might be an empty string, if the selected page is the `index` page. Though that's of course an unlikely edge case.

ausi:
A news can have a jumpTo page, if this page is the home page it would be the empty url.

ausi:
> Though that's of course an unlikely edge case.

Yeah, but IMO it would make sense to fix all cases by adjusting the source of the URL `Page::generateFrontendUrl`.

leofeyer:
> If the `source` is `internal` for example, the `{{news_url::*}}` might be an empty string, if the selected page is the `index` page. Though that's of course an unlikely edge case.

No, that is not correct. The path would be `index/news-alias.html` in this case.

fritzmg:
> No, that is not correct. The path would be `index/news-alias.html` in this case.

No it would not. You are confusing this with `source = default` where the redirect page of the news archive is pointing to `index`. I am talking about `source = internal` - i.e. selecting a target page within the news article itself.

leofeyer:
I see. My bad. 🙈

aschempp (dismissed):
I think this can be merged regardless of any other changes, it's a safety guard against an invalid URL. But we sure should fix the other places and `PageModel::getFrontendUrl`.

leofeyer:
Support for news, events and FAQs has been added in 9639096c3ebc8b678824b0329d0ad8982a996e62.

fritzmg (approved):

PR #2558, opened 2020-12-10
Fix phpstan (ThrowMustBundlePreviousExceptionRule not found)

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

The `thecodingmachine/phpstan-strict-rules` dev dependency is missing in `4.10` and `master`, which is required for

https://github.com/contao/contao/blob/90caa2a8e8cd3f7e066628f7a84f5c12c32e15f3/phpstan.neon.dist#L10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -159,6 +159,7 @@
         "slam/phpstan-extensions": "^5.0",
         "symfony/browser-kit": "4.4.*",
         "symfony/phpunit-bridge": "4.4.*",
+        "thecodingmachine/phpstan-strict-rules": "^0.12",
         "weirdan/doctrine-psalm-plugin": "^0.11"
     },
     "extra": {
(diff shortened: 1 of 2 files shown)


fritzmg:
Oh I see, `monorepo-tools composer-json` is removing this dev dependency automatically. Though I do not really get why?

ausi:
You have to add it here https://github.com/contao/contao/blob/master/monorepo.yml#L11 too so that the monorepo tools know that this dependency is only for the mono repository and is not needed in the bundles.

fritzmg:
Ah ok, thank you, didn't know :)

leofeyer:
This has already been fixed in #2543 and will be merged upstream automatically.

fritzmg:
Ah ok, missed that 👍

PR #2564, opened 2020-12-11
Clean up after the upstream merge

leofeyer:


core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -349,7 +349,7 @@ public static function indexPage($arrData)
 					tl_search_index.pid,
 					SQRT(SUM(POW(
 						(1 + LOG(relevance)) * LOG((
-							" . (int) ($objDatabase->query("SELECT COUNT(*) as count FROM tl_search")->count + 1) . "
+							" . ($objDatabase->query("SELECT COUNT(*) as count FROM tl_search")->count + 1) . "
```
ausi, line 352: Why got this removed? I think we should cast the `->count` then if we don’t cast the whole expression.

leofeyer: The `+ 1` automatically casts the previous expression, so `(int)` is not necessary.

ausi: `+ 1` could also result in a `float` or `NAN` or `INF`. But I agree, in this case probably very unlikely.

PR #2572, opened 2020-12-14
Revert #2476 + work around for PHPStan

### LINKED ISSUE #2570: Registering callbacks/hooks/picker providers with a priority via annotations is broken
**Affected version(s)**

Contao 4.10

**Description**
This [commit](https://github.com/contao/contao/commit/e68ed7fbf66d2c6d1cfa3a880f96fc3501592b2b) from #2476 breaks defining callbacks, hooks and picker providers with a priority. This will fail silently and the elements won't be registered at all. :scream: 

Here is an example:
```
[Type Error] Attribute "priority" of @Callback declared on method Contao\CoreBundle\EventListener\DataContainer\ContentCompositionListener::generateArticleForPage() expects a(n) int|null, but got integer.
```

This is because of this check in the doc parser that now fails because `'int' !== 'int|null'`: https://github.com/doctrine/annotations/blob/984d7dc8d61c3566768379fe25aa531b54b8130d/lib/Doctrine/Common/Annotations/DocParser.php#L902

Doctrine Annotations should probably handle this but we should imo fix this asap as this renders lots of things unusable in the current 4.10.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2570
| Docs PR or issue | -

This works around the issue in #2570. Besides that we should probably investigate if `doctrine/annotations` has a bug and/or check  our default values.

### DIFF
--- a/core-bundle/tests/EventListener/DataContainer/DisableAppConfiguredSettingsListenerTest.php
+++ b/core-bundle/tests/EventListener/DataContainer/DisableAppConfiguredSettingsListenerTest.php
@@ -24,10 +24,9 @@ class DisableAppConfiguredSettingsListenerTest extends TestCase
     public function testAnnotatedCallbacks(): void
     {
         $listener = $this->createListener();
-
         $annotationReader = new AnnotationReader();
 
-        /** @var callback $annotation */
+        /** @var Callback $annotation */
         $annotation = $annotationReader->getClassAnnotation(new \ReflectionClass($listener), Callback::class);
 
         $this->assertSame('tl_settings', $annotation->table);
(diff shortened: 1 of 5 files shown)


aschempp (dismissed):

core-bundle/tests/EventListener/DataContainer/DisableAppConfiguredSettingsListenerTest.php
```diff
@@ -26,16 +26,12 @@ public function testAnnotatedCallbacks(): void
         $listener = $this->createListener();
 
         $annotationReader = new AnnotationReader();
+
+        /** @var callback $annotation */
```
m-vo: Just in case you're wondering as well: I just noticed that the CS fixer's `PhpCsFixer\Fixer\Phpdoc\PhpdocTypesFixer` transforms `Callback` to `callback` :man_facepalming:. It doesn't really matter though, because the annotation still works.

leofeyer: I have adjusted this in contao/easy-coding-standard@251e23cf9435e7e9dfd867f5606024df5424ca32.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #2582, opened 2020-12-18
Use composer-bin-plugin to isolate the dev dependencies

leofeyer:
This PR implements dev dependency isolation as discussed in yesterday's Mumble call.

### TODO

- [x] Make `contao/test-case` compatible with `vendor-bin` so it can be removed as dev dependency in the bundles.


m-vo:
Psalm schema location should probably be adjusted as well:

```diff
- xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+ xsi:schemaLocation="https://getpsalm.org/schema/config vendor-bin/code-quality/vendor/vimeo/psalm/config.xsd"
```

leofeyer:
Fixed in c31e44b0bdde3f28b476a3670d3ffed606cfe3e5.

m-vo:
What about adding a sentence to the ['Running scripts'](https://github.com/contao/contao#running-scripts) section of the README as well?

Something like this:
```diff
- You can use Composer to run the code quality scripts:
+ Run `composer bin all update` to install the code quality tools. You can then use the following shortcuts to run them:
```

leofeyer:
Done.

m-vo (dismissed):

m-vo:
I really like this iteration. :+1: 

This is now calling the tools from `tools/*/vendor/bin` instead of (symlinked) from within `vendor/bin`. We might want to double check this still works for every tool + platform (Windows?). Probably everyone's using `getcwd()` internally and it's just fine.

leofeyer:
At least the CI chain does not break on Windows, so this should not be an issue.

PR #2586, opened 2020-12-21
Fix require-dev and re-enable require-dev merging in monorepo tools

ausi:
Follow up to #2582

I think we should revert https://github.com/contao/monorepo-tools/commit/a4ff52a5b125c50e22d43b22ccdcb7a5a9801095 as this helps us spot issues like the one (`"composer/composer": "^1.0 || ^2.0"`) from this PR.

### DIFF
--- a/tools/monorepo/composer.json
+++ b/tools/monorepo/composer.json
@@ -1,5 +1,5 @@
 {
     "require": {
-        "contao/monorepo-tools": "dev-master#608d7dee82a1dba95be989b0b843634e9a2da085"
+        "contao/monorepo-tools": "dev-master"
     }
 }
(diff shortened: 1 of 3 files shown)


tools/monorepo/composer.json
```diff
@@ -1,5 +1,5 @@
 {
     "require": {
-        "contao/monorepo-tools": "dev-master"
+        "contao/monorepo-tools": "dev-master#608d7dee82a1dba95be989b0b843634e9a2da085"
```
ausi: Should get reverted after we reverted https://github.com/contao/monorepo-tools/commit/a4ff52a5b125c50e22d43b22ccdcb7a5a9801095
```suggestion
        "contao/monorepo-tools": "dev-master"
```

leofeyer:
Thank you @ausi.

PR #2591, opened 2020-12-23
Retrieve the PageModel from the current request

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2506
| Docs PR or issue | https://github.com/contao/docs/pull/629

The changes introduced in #2506 cause the following warning if the `__invoke` method is extended.

`
Warning: Declaration of App\Controller\FrontendModule\MyController::__invoke(Symfony\Component\HttpFoundation\Request $request, Contao\ModuleModel $model, string $section, ?array $classes = NULL): Symfony\Component\HttpFoundation\Response should be compatible with Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController::__invoke(Symfony\Component\HttpFoundation\Request $request, Contao\ModuleModel $model, string $section, ?array $classes = NULL, ?Contao\PageModel $pageModel = NULL): Symfony\Component\HttpFoundation\Response  
`

This PR fixes that by not relying on the argument resolver but duplicate the functionality inside the abstract controller. While implementing it, I notices the asset context would not behave correctly in a fragment case either.

I think we should introduce a service in Contao 4.1x to get the page model, and possibly also set the page title etc. But this would certainly be a feature we can't have in 4.9 just yet.

### DIFF
--- a/core-bundle/src/Asset/ContaoContext.php
+++ b/core-bundle/src/Asset/ContaoContext.php
@@ -12,10 +12,10 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Asset;
 
+use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\PageModel;
 use Symfony\Component\Asset\Context\ContextInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Contao\CoreBundle\Framework\ContaoFramework;
 
 /**
  * @internal Do not use this class in your code; use the "contao.assets.assets_context" or "contao.assets.files_context" service instead
@@ -102,6 +102,10 @@ class ContaoContext implements ContextInterface
         $request = $this->requestStack->getCurrentRequest();
 
         if (null === $request || !$request->attributes->has('pageModel')) {
+            if (isset($GLOBALS['objPage']) && $GLOBALS['objPage'] instanceof PageModel) {
+                return $GLOBALS['objPage'];
+            }
+
             return null;
         }
 
--- a/core-bundle/src/Asset/ContaoContext.php
+++ b/core-bundle/src/Asset/ContaoContext.php
@@ -15,13 +15,22 @@ namespace Contao\CoreBundle\Asset;
 use Contao\PageModel;
 use Symfony\Component\Asset\Context\ContextInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Contao\CoreBundle\Framework\ContaoFramework;
 
+/**
+ * @internal Do not use this class in your code; use the "contao.assets.assets_context" or "contao.assets.files_context" service instead
+ */
 class ContaoContext implements ContextInterface
 {
     /**
      * @var RequestStack
      */
     private $requestStack;
+
+    /**
+     * @var ContaoFramework
+     */
+    private $framework;
 
     /**
      * @var string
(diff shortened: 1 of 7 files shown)


fritzmg:
This always returns the `PageModel` instance from the `pageModel` attribute - which in turn is not the global `PageModel` instance. Shouldn't it always return the global `PageModel` instance instead, if present?

aschempp:
yes, but isn't that the same as we currently have (through the controller resolver)?

leofeyer:
The PR is marked as draft and almost all CI jobs fail, so I do not understand why you have requested a review? Do you want me to provide feedback on the concept?

aschempp:
Yes. The code fix can be reviewed. The PR is draft because tests are not update and it can't/shouldn't be merged as-is.

core-bundle/src/Asset/ContaoContext.php
```diff
@@ -89,11 +99,32 @@ public function getStaticUrl(): string
...
+            return null;
+        }
+
+        $pageModel = $request->attributes->get('pageModel');
+
+        if ($pageModel instanceof PageModel) {
+            return $pageModel;
+        }
+
+        if (
```
leofeyer, line 118: This if condition will never be evaluated if `$request` is `null` or does not have an attribute named `pageModel`. Intended?

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -27,24 +28,48 @@ abstract class AbstractFragmentController extends AbstractController implements
...
+            return null;
+        }
+
+        $pageModel = $request->attributes->get('pageModel');
+
+        if ($pageModel instanceof PageModel) {
+            return $pageModel;
+        }
+
+        if (
```
leofeyer, line 59: Same here.

leofeyer:
Besides my two comments, the concept looks correct.

core-bundle/src/Asset/ContaoContext.php
aschempp: yes, the if-condition refers to `(int) $pageModel` (from the request), we can't operate without that. The method is not supposed to _always_ get the (global) page model, but to return the one from the current request.

fritzmg (dismissed):
> This always returns the `PageModel` instance from the `pageModel` attribute - which in turn is not the global `PageModel` instance. Shouldn't it always return the global `PageModel` instance instead, if present?

As far as I tested, the page model of the request attribute is the same as in `global $objPage` 👍

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -27,24 +28,48 @@ abstract class AbstractFragmentController extends AbstractController implements
...
     }
 
     protected function getPageModel(): ?PageModel
     {
-        return $this->pageModel;
+        $request = $this->get('request_stack')->getCurrentRequest();
+
+        if (null === $request || !$request->attributes->has('pageModel')) {
+            return null;
+        }
```
leofeyer, line 51: The equivalent in the `ContaoContext` class looks like this:

```php
        if (null === $request || !$request->attributes->has('pageModel')) {
            if (isset($GLOBALS['objPage']) && $GLOBALS['objPage'] instanceof PageModel) {
                return $GLOBALS['objPage'];
            }

            return null;
        }
```

Do we not need the `$GLOBALS['objPage']` check here?

aschempp: No I don't think we want that in a controller. In a controller, the page model should always relate to the current request. If I call `$this->getPageModel()` in my controller method, but my request/route does not contain that information, I would expect to get a `null` value returned.

leofeyer (approved):

PR #2592, opened 2020-12-23
Update scheb/2fa-bundle and lcobucci/jwt for PHP 8

ausi:
It seems that we need to update to scheb/2fa-bundle 5.0 in order to be compatible with PHP 8. Because the two-factor-bundle 4 requires `lcobucci/jwt` in version 3 which is not (and will not be) PHP 8 ready.

Related #2263


bytehead:
@ausi should I take care of this?

ausi:
@bytehead that would be super awesome 😍

leofeyer:
@bytehead I had already started working on this, so you might want to use https://github.com/leofeyer/contao/commit/ecaa4735cbd3c604c488dc9d32d735c7800674c3 as starting point.

bytehead:
@leofeyer thanks, I'll check!

bytehead:
@leofeyer did you intentionally branch from master? Or should this be a bugfix for Contao 4.9?

ausi:
Needs to be a bugfix for 4.9 IMO because we want 4.9 to be PHP 8 compatible.

bytehead:
Closing this in favor of #2617.

PR #2593, opened 2020-12-23
Replace highlight.js with highlight.php

leofeyer:
This PR replaces `contao-components/highlight.js` with its PHP clone `scrivo/highlight.php`.


aschempp:
does this mean highlighting now works on the server instead of client side? How can I disable code highlighting in the element now (or use something else than highlight.js)?

leofeyer:
You have to create your own content element, just like before. The integration in Contao is tailored to highlight.js (CSS classes, file names, style sheets etc.) and we can only switch because highlight.php aims to be an exact clone.

core-bundle/src/Migration/Version411/RemoveJsHighlightMigration.php
```diff
@@ -0,0 +1,97 @@
...
+        $layout = $this->connection->fetchAssociative("
+            SELECT
+                COUNT(*) AS cnt
+            FROM
+                tl_layout
+            WHERE
+                scripts LIKE '%js_highlight%'
+        ");
+
+        return $layout['cnt'] > 0;
```
ausi: I’d rename `cnt` to `count` or use this code instead 🤓:

```suggestion
        $count = (int) $this->connection->fetchOne("
            SELECT
                COUNT(*)
            FROM
                tl_layout
            WHERE
                scripts LIKE '%js_highlight%'
        ");

        return $count > 0;
```

leofeyer: Changed in de3e3c299969f5d8bcf9d2b180612e0e9b4b7914.

ausi (approved):

PR #2568, opened 2020-12-11
Restore compatibility with E_WARNING

leofeyer:
This PR aims to revert the PHP error_level changes we had to make in #2535 to be compatible with PHP 8.

@ausi and I have also discussed making the code compatible with `E_ALL`, however, that is a lot of work and I am not sure if the Contao ecosystem will ever be compatible given that we have been ignoring `E_NOTICE` ever since. But let us see how this PR develops and maybe we can create another PR for `E_ALL`.

PR #2594, opened 2020-12-23
Fix two more E_WARNING issues

leofeyer:
First of presumably several follow-up issues on #2568. 🙈

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_layout.php
+++ b/core-bundle/src/Resources/contao/dca/tl_layout.php
@@ -507,7 +507,7 @@ class tl_layout extends Backend
 	 */
 	public function getStyleSheets(DataContainer $dc)
 	{
-		$intPid = $dc->activeRecord->pid;
+		$intPid = $dc->activeRecord->pid ?? null;
 
 		if (Input::get('act') == 'overrideAll')
 		{
--- a/core-bundle/src/Resources/contao/widgets/ModuleWizard.php
+++ b/core-bundle/src/Resources/contao/widgets/ModuleWizard.php
@@ -220,7 +220,7 @@ class ModuleWizard extends Widget
 				}
 				elseif ($button == 'enable')
 				{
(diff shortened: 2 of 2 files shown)


core-bundle/src/Resources/contao/dca/tl_layout.php
```diff
@@ -507,7 +507,7 @@ public function checkPermission()
 	 */
 	public function getStyleSheets(DataContainer $dc)
 	{
-		$intPid = $dc->activeRecord->pid;
+		$intPid = $dc->activeRecord->pid ?? null;
```
aschempp, line 510: I would use `0` instead of `null` to generate a valid SQL query

core-bundle/src/Resources/contao/widgets/ModuleWizard.php
```diff
@@ -220,7 +220,7 @@ public function generate()
 				}
 				elseif ($button == 'enable')
 				{
-					$return .= ' <button type="button" data-command="enable" class="mw_enable" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['mw_enable']) . '">' . Image::getHtml((($this->varValue[$i]['enable']) ? 'visible.svg' : 'invisible.svg')) . '</button><input name="' . $this->strId . '[' . $i . '][enable]" type="checkbox" class="tl_checkbox mw_enable" value="1" onfocus="Backend.getScrollOffset()"' . (($this->varValue[$i]['enable']) ? ' checked' : '') . '>';
+					$return .= ' <button type="button" data-command="enable" class="mw_enable" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['mw_enable']) . '">' . Image::getHtml((($this->varValue[$i]['enable'] ?? null) ? 'visible.svg' : 'invisible.svg')) . '</button><input name="' . $this->strId . '[' . $i . '][enable]" type="checkbox" class="tl_checkbox mw_enable" value="1" onfocus="Backend.getScrollOffset()"' . (($this->varValue[$i]['enable'] ?? null) ? ' checked' : '') . '>';
```
aschempp, line 223: I would use `false` instead of `null` for a boolean operation

core-bundle/src/Resources/contao/dca/tl_layout.php
leofeyer: We have agreed on always using `?? null`, because anything else like `?? ''` or `?? false` is a behavior change. The discussion is here: https://github.com/contao/contao/pull/2568#issuecomment-747391516

ausi (approved):

PR #2599, opened 2020-12-28
Fix more E_WARNING issues

leofeyer:
Another follow-up issue on #2568.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -1641,7 +1641,7 @@ class tl_content extends Backend
 		$arrPids = array();
 		$arrArticle = array();
 		$arrRoot = array();
-		$intPid = $dc->activeRecord->pid;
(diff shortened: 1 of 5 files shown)


core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -1641,7 +1641,7 @@ public function getArticles(DataContainer $dc)
 		$arrPids = array();
 		$arrArticle = array();
 		$arrRoot = array();
-		$intPid = $dc->activeRecord->pid;
+		$intPid = $dc->activeRecord->pid ?? null;
```
aschempp, line 1644: I still think the "fallback value" of a field named `intPid` should be `0` and not `null`

ausi: But this would be a (potential) behavior change and we intentionally wanted to keep the code working the same as before for the PHP 8 warning patches.

PR #2600, opened 2020-12-28
Remove the Composer script handler

leofeyer:
I think it is time to remove the old Composer script handler in the core bundle. We had added it in case someone wanted to add the Contao core bundle to their Symfony 3 installation, but all this has long since been replaced by our [Symfony flex recipe](https://github.com/symfony/recipes-contrib/tree/master/contao/core-bundle/4.6) for Symfony 4 and 5.

Theoretically, we could remove the script handler in Contao 4.9 as well, because Contao 4.9 is no longer compatible with Symfony 3. However, just to be sure, I have assigned the PR to the `4.11` milestone.


m-vo (dismissed):

aschempp (approved):

composer.json
```diff
@@ -147,7 +147,6 @@
     "require-dev": {
         "ext-fileinfo": "*",
         "bamarni/composer-bin-plugin": "^1.4",
-        "composer/composer": "^1.0 || ^2.0",
```
aschempp: I think we still need this for the manager bundle?

aschempp (changes requested):

composer.json
leofeyer: Seems the dependency was missing there all the time. 😄

leofeyer: Re-added in edcc0a5b4f8684e86995324e5a5ab44bf0f90043.

aschempp (approved):

m-vo (approved):

PR #2601, opened 2020-12-28
Cast numeric identifiers to int in the FigureBuilder

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

When auto detecting the resource type in the `FigureBuilder` a numeric identifier should always be cast to an integer. Otherwise `fromId()` will fail with a `TypeError` when called with a numeric string.


ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2607, opened 2020-12-28
Add a "figure" insert tag

### LINKED ISSUE #2466: Refactor image/picture insert tags
We're talking about this:

https://github.com/contao/contao/blob/a9791371c698b060dd97a1839022d8da67a1dee7/core-bundle/src/Resources/contao/library/Contao/InsertTags.php#L767-L931

Looks like there is quite some redundant code that could be removed by using the image studio classes. This way we would also normalize the behavior for edge cases.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2466 
| Docs PR or issue | todo

This deprecates the current `image` and `picture` insert tags in favor of a new one called `figure`. Its usage is analogous to the `contao_figure` Twig function. 

#### Syntax
```
{{figure::<from>[?<key>=<value>,[&<key>=<value>]*]}}
```

#### Arrays
The `<size>` and `<configuration>` can accept (nested) arrays that can be defined like so:
```
Format:
  - Arrays must be surrounded with "[" and "]"
  - Elements need to be separated with a ","
  - Keys are optional; if used, keys and values must be separated with a ":"
  - Keys are always treated as strings
  - Strings can be delimited with single or double quotes
  - Non delimited values are treated as int/float/bool and fallback to string
  - Non delimited strings can only contain "\w" characters
  - Floats must use the "." as decimal separator

Examples:
  '[a, b]'              --> [0 => 'a', 1 => 'b']
  '[a: "foo", b]'       --> ['a' => 'foo', 1 => 'b']
  '[a: 2, b: true]'     --> ['a' => 2, 'b' => true]
  '[[a, b], foo: [c]]'  --> [0 => [0 => 'a', 1 => 'b' ], 'foo' => [0 => 'c']]
```
For the usage inside the insert tag, the outermost 'array brackets' can be omitted.

#### Examples
Altogether this allows using the full `FigureBuilder` feature set:

```
{{figure::123?size=[100,200,crop]&enableLightbox=true}}
{{figure::files/cat.jpg?size=_my_size&metadata[title]=foo%20bar&options[attr][class]=bar}}
```

**Edit**: Syntax adjusted to the final version on February 15th.


m-vo:
We could also enforce using quotes around strings so that `int`/`float`/`string`/`bool` becomes distinguishable ('castable'). 

```
{{figure::1::[200, 200, 'crop']::[enableLightbox:true, metadata:[title: 'foo'], options: [attr: [class: 'true']]]}}
```
&rarr;
```
$size = [200, 200, 'crop'];

$configuration = [
  'enableLightbox' => true,
  'metadata' => ['title' => 'foo'],
  'options' => ['attr' => ['class' => 'true']]
];
```

m-vo:
The array output is now typed. I updated the initial description.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -764,8 +767,58 @@ protected function doReplace($strBuffer, $blnCache)
...
+					catch (\Throwable $e)
+					{
+						$arrCache[$strTag] = '';
+					}
+
+					break;
+
 				case 'image':
 				case 'picture':
+					trigger_deprecation('contao/core-bundle', '4.11', 'Using the "image" and "picture" insert tags has been deprecated and will no longer work in Contao 5.0. Use the "figure" insert tag instead.');
```
ausi: Are we sure we want to deprecate image and picture insert tags? What if I want to include an image without the `<figure>` image container?

ausi (changes requested):
Do we really need another format for data to parse with `ArrayString`? I would prefer to use something existing as parsing data structures is quite complex and we would need to make sure that the syntax is unambiguous and allows escaping for special characters and so on...

I guess JSON and YAML cannot be used because they require `{` and `}` which is not allowed in insert tags?

The current image/picture insert tags use URL parameters. Can’t we reuse this for `figure` too?

```
{{figure::files/cat.jpg?size=_my_size&metadata[title]=true%20foo&options[attr][class]=bar}}
```

m-vo:
> Do we really need another format for data to parse with ArrayString? 

Reusing something else was really was first attempt. But I didn't find anything that both supported nesting and doesn't include curly braces or indentation. JSON would have been nice, but is not compatible with the insert tag curly braces. :cry:.

I agree that URL parameters would work in general, but this…

> `{{figure::files/cat.jpg?size=_my_size&metadata[title]=true%20foo&options[attr][class]=bar}}`

… is IMHO really hard to read and even harder to write. Especially with the need for urlencoding stuff.

We could probably also go full PHP for the syntax, nikic/token-parse the result for sanitizing, and then reemit and evaluate it.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
m-vo: That's a valid point for the `image` insert tag, I guess. Although you could as well use an alternative template that simply outputs the image instead.

ausi:
> > `{{figure::files/cat.jpg?size=_my_size&metadata[title]=true%20foo&options[attr][class]=bar}}`
> 
> … is IMHO really hard to read and even harder to write.

I don’t think it’s that much harder to read than `{{figure::"files/cat.jpg"::"_my_size"::metadata:[title:'true foo'],options:[attr:[class:bar]]}}` but the URL parameter syntax is standardized and most users understand it.

> Especially with the need for urlencoding stuff.

Most characters don’t need to be URLencoded, the `%20` from my example can be kept as a plain space character. And the `ArrayString` would also need some sort of encoding for `[`, `]`, `"` and `::`, right?

> We could probably also go full PHP for the syntax, nikic/token-parse the result for sanitizing, and then reemit and evaluate it.

I’m not really a fan of that either 😕

I think we need to agree on how we want to handle komplex parameter structures for insert tags in general @contao/developers

*(I should’ve reviewed this before yesterdays call 🙈, sorry for that...)*

Toflar:
> I think we need to agree on how we want to handle komplex parameter structures for insert tags in general @contao/developers

We already use the url stuff. As you stated yourself in 

> The current image/picture insert tags use URL parameters. Can’t we reuse this for figure too?

The same also applies for `{{file}}`.
Both formats are cumbersome to write and it's a matter of documenting it properly with multiple examples.
I don't think we should switch to yet another format.

leofeyer:
> > > `{{figure::files/cat.jpg?size=_my_size&metadata[title]=true%20foo&options[attr][class]=bar}}`
> > 
> > … is IMHO really hard to read and even harder to write.
> 
> I don’t think it’s that much harder to read than `{{figure::"files/cat.jpg"::"_my_size"::metadata:[title:'true foo'],options:[attr:[class:bar]]}}` but the URL parameter syntax is standardized and most users understand it.

I would be fine with the URL syntax, too.

> > We could probably also go full PHP for the syntax, nikic/token-parse the result for sanitizing, and then reemit and evaluate it.
> 
> I’m not really a fan of that either 😕

ausi:
Then I think we should go with the URL-parameters.

It probably would make sense to keep overlapping parameters with `{{image}}` and `{{picture}}` consistent.

And I would not deprecate `{{image}}` and `{{picture}}`, but maybe instead refactor them to use the same basis as `{{figure}}` does. These insert tags should play well together I think:

1. Use `{{image}}` if you only want a single non-responsive ``
2. Use `{{picture}}` to get a responsive image with all its features
3. Use `{{figure}}` to get a `<figure class="image_container">` with all its meta data.

m-vo:
Ok, Ill refactor the PR to use URL parameters. :upside_down_face:  

>but maybe instead refactor them to use the same basis as {{figure}}

That was my initial attempt. But the edge cases were exploding there. I'll give it another shot, though.

m-vo:
The PR is rebased + updated to use URL parameters now. I did not yet try refactoring the other bits as this could IMHO be done in an independent second step.

> It probably would make sense to keep overlapping parameters with {{image}} and {{picture}} consistent.

I'd rather keep this consistent with the other places for `FigureBuilder` usage like in the PHP and Twig template functions (and the `FigureBuilder` itself of course). The other insert tags for instance allow setting things like `width` and `height` individually which doesn't play well with predefined settings from `config.yaml`.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -765,6 +766,43 @@ protected function doReplace($strBuffer, $blnCache)
 					break;
 
 				// Images
+				case 'figure':
+					// Expected format: figure::<from>[?<key>=<value>,[&<key>=<value>]*][::<template>]
```
leofeyer: Why is `template` not one of the key/value options?

core-bundle/src/Twig/Runtime/FigureRendererRuntime.php
```diff
@@ -23,7 +23,10 @@
 use Twig\Environment;
 use Twig\Extension\RuntimeExtensionInterface;
 
-final class FigureRendererRuntime implements RuntimeExtensionInterface
+/**
+ * @final
```
leofeyer: If we cannot declare the class as final, we should declare it as `@internal`.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
m-vo: The idea was to keep it in line with the definitions of the `Template#figure()` and `contao_figure` Twig functions which accept the template as last parameter.

As we're not respecting this for the `size` as individual argument either, we could as well allow `template=<template>' in the query string. :thinking:

leofeyer: Yes please. 😊

core-bundle/src/Twig/Runtime/FigureRendererRuntime.php
m-vo: This would IMHO be a BC break. Final classes aren't meant to be extended (= no BC for internal fields) while internal ones aren't meant to be used (= no BC for the API).

leofeyer: Why not keep the `final class` then?

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
m-vo: Changed in 77119d19edb103f1e2c141a3a1d93f9a66b07806.

core-bundle/src/Twig/Runtime/FigureRendererRuntime.php
m-vo: :bird: &larr; mocky, the mockingbird

m-vo: Alternatively, we could duplicate the code or wrap it in a service (maybe shared across all three usages). For this use case the implementation of the `FigureRendererRuntime` already does the right thing, including the automatic wrapping of metadata into an object, so that's why I went for it.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -765,6 +766,43 @@ protected function doReplace($strBuffer, $blnCache)
...
+					if (null === $from || 2 !== \count($elements))
+					{
+						$arrCache[$strTag] = '';
+
+						break;
+					}
+
+					// Build arguments for FigureRendererRuntime#render
+					$size = $configuration['size'] ?? null;
+					$template = $configuration['template'] ?? null;
```
ausi: This only supports twig templates not Contao html5 templates, right?

m-vo: It's now refactored and all three* usages accept both.

*) the `contao_figure` Twig function, the `Template#figure` function and the `figure` insert tag

see a07d8b43fdbda28530683d0212b684029713bb2f

core-bundle/src/Twig/Runtime/FigureRendererRuntime.php
m-vo: changes are reverted, see a07d8b43fdbda28530683d0212b684029713bb2f for refactored solution

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -1123,6 +1152,19 @@ protected function doReplace($strBuffer, $blnCache)
...
+	 * @return array<string|null, array>
+	 */
+	private function parseUrlWithQueryString(string $url): array
+	{
+		$base = parse_url($url, PHP_URL_PATH) ?: null;
+		$query = parse_url($url, PHP_URL_QUERY) ?: '';
+
+		parse_str($query, $attributes);
+
+		return array($base, $attributes);
```
leofeyer, line 1166: If I compare this with the code of the `image` tag, I notice that there are some differences.

https://github.com/contao/contao/blob/6d3ccc6a115719c0f1a5a4e0be90ba7928750a56/core-bundle/src/Resources/contao/library/Contao/InsertTags.php#L783-L786

What about decoding and replacing `[&]` with `&`? Is this not necessary here?

leofeyer:
Thank you @m-vo.

PR #2609, opened 2020-12-29
Correctly validate min and max values in text fields

aschempp:
Without the config value, the input is not validated on the server (e.g. if browser side validation is disabled): https://github.com/contao/contao/blob/master/core-bundle/src/Resources/contao/library/Contao/Widget.php#L840

### DIFF
--- a/core-bundle/src/Resources/contao/forms/FormTextField.php
+++ b/core-bundle/src/Resources/contao/forms/FormTextField.php
@@ -202,8 +202,8 @@ class FormTextField extends Widget
 				{
 					return $this->arrAttributes['min'];
 				}
-				return parent::__get($strKey);
 
+				return parent::__get($strKey);
 
 			case 'max':
 			case 'maxval':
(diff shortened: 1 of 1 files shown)


fritzmg (changes requested):
That should not be necessary. Accessing `$this->minval`/`$this->maxval` will load the value either from `$this->arrAttributes` or `$this->arrConfiguration`:

https://github.com/contao/contao/blob/0e7bb7ab60634f1266c92bbccb643e098e98acf4/core-bundle/src/Resources/contao/library/Contao/Widget.php#L411-L421

aschempp:
you're right, but it refers to `minval` but we're only setting `min` (and max accordingly)

fritzmg:
Oh yeah, you are right. But may be we should instead normalise this in `__get`?

leofeyer (changes requested):

fritzmg (dismissed):

leofeyer (approved):

PR #2612, opened 2020-12-30
Allow using Twig templates

### LINKED ISSUE #2303: Allow using Twig templates to implement/replace Contao templates
Some time ago I created a [bundle](https://github.com/m-vo/contao-twig) that allows using Twig templates instead of or to overwrite Contao PHP templates. This concept really works out nicely if you want a 'twig-only' approach without having to change every single template. 

One thing that bothers me, though, is that a [proxy PHP template](https://github.com/m-vo/contao-twig/blob/master/src/Resources/contao/templates/twig_template_proxy.html5) needs to be rendered along each Twig template. If we would implement this in the core instead, the correct renderer could be chosen from the beginning.

This would basically mean that templates inside the `/templates` folder can either have a `.html5` or `.twig` extension (same logic). If both exist I'd give the Twig version precedence.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Closes #2303 
| Docs PR or issue | todo

This PR adds the ability to replace PHP templates with Twig templates in your application. If a template `<template>.html.twig` is registered in the main Twig template path, rendering is forwarded to Twig instead of rendering natively.

There are a few tricks involved to make the full context available **without the need of the raw filter**:

* All data stored in `$arrData` as well as some context sensitive functions from the `FrontendTemplate` class will end up in the context.
* However, before passing them to Twig, proxy classes wrapping the values are created (where needed). This allows accessing callables like in Contao templates as well as \*drumroll\* - bypassing Twigs escaping mechanism. 

#### Callables
If a callable is just a lazy replacement for a string (i.e no arguments are needed), they can just be accessed as if they were strings:
```php
$template->my_callable = function(): string { return 'foo'; };
```
```twig
Will output "foo":
{{ my_callable }}
```

If arguments are needed, the function can be called via the proxy's `invoke` method:
```php
$template->my_callable = function($a): string { return $a . 'bar'; };
```
```twig
Will output "foobar":
{{ my_callable.invoke('foo') }}
```

#### Input encoding magic
The proxy objects are tagged with a special `SafeHTMLValueHolderInterface` which gets registered as a `safe class` for the `html` strategy in Twigs `EscaperExtension`.

In order to support arrays or objects that can generate values themselves the proxy will intercept calls and wrap the results. This way the effect is transitive on child elements. That's the reason why there are different value holders implemented. The `ObjectValueHolder` is the most sophisticated one because it needs to reimplement parts of Twigs access strategy, the other ones are quite straight-forward. 

It's possible to tag arbitrary classes with the `SafeHTMLValueHolderInterface` (e.g. for performance reasons) - they will get skipped when creating value holders.


core-bundle/src/Twig/Extension/ContaoTemplateExtension.php
```diff
@@ -76,4 +91,18 @@ public function renderContaoBackendTemplate(array $blocks = []): string
...
+     * with a harmless identity function (i.e. removing 'is_safe') once we're
+     * switching to output encoding.
+     */
+    public function handleInputEncoding(array $context, string $input): string
+    {
+        if (($context['_contao_encoding'] ?? null) !== 'input') {
+            throw new \LogicException("The filter 'c' must only be applied to Contao template context. Did you mean 'raw'?");
+        }
+
+        return $input;
```
ausi: How about using

```suggestion
        return htmlspecialchars($input, ENT_QUOTES, null, false);
```

With `double_encode` set to `false`. This should keep all “input encoded” strings the same and would force you to use `|raw` if the actual raw value needs to be written to the output.

This would reduce the risk of template developers learing the wrong way to work with twig templates. The only difference to “regular” twig templates would be to use `|c` everywhere. Maybe we could make the `c` filter the default ouput filter for Contao twig templates?

m-vo: Imho this is the wrong way round - we need a filter that allows passing "unsafe" HTML because it's already pre-rendered and sanitized. With `htmlspecialchars` your would get the same as defining the filter without `'is_safe' => 'html`. 

See https://twig.symfony.com/doc/3.x/advanced.html#automatic-escaping

> Maybe we could make the c filter the default ouput filter for Contao twig templates?

This should be possible on a template level with something like the following if we'd register an `Escaper`:

```twig
{% autoescape 'contao' %}
<h2>My Template</h2>
Some {{ content }} that is treated like being passed through the raw filter.
{% endautoescape %}
```

I think `is_safe` can also be evaluated by a callback (context sensitive), so we could maybe check the context before blindly allowing anything.

m-vo: Ah! I just found out that you can mark classes as safe for the escaper via `EscaperExtension::addSafeClass():` - I'll investigate if this could be the solution. Maybe we can simply proxy everything through this class when building the context from within a Contao template. :smile:

m-vo: Yes, this actually works!! :tada: :tada:
I'll finish this tomorrow.

m-vo: Done :star:

m-vo:
Twig for Contao templates:  no `|raw` needed anymore. :partying_face: :tada:

I updated the description accordingly.

m-vo:
Just letting this here for further discussions: There is also the option to register a `NodeVisitor` that's active in the compilation phase. This feature isn't documented but gets mentioned in the section about [extensions](https://twig.symfony.com/doc/3.x/advanced.html#creating-an-extension). Probably allows altering nodes or the node tree itself.

core-bundle/src/Twig/Compat/ArrayValueHolder.php
```diff
@@ -0,0 +1,58 @@
...
+
+final class ArrayValueHolder extends \ArrayIterator implements SafeHTMLValueHolderInterface
+{
+    /**
+     * @var string
+     */
+    private $valueName;
+
+    /**
+     * @internal
```
leofeyer, line 23: Shouldn't this be at class level?

core-bundle/src/Twig/Compat/ProxyFactory.php
```diff
@@ -0,0 +1,66 @@
...
+        // Objects that are explicitly marked
+        if ($value instanceof SafeHTMLValueHolderInterface) {
+            return false;
+        }
+
+        // Empty strings/arrays
+        return !empty($value);
+
+        // A more sophisticated check could go here in the future that further
+        // limits the amount of proxy objects that need to be created.
```
leofeyer, line 64: This should be a `@todo` phpDoc comment.

core-bundle/src/Twig/Extension/ContaoCompatExtension.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Contao\CoreBundle\Twig\Compat\SafeHTMLValueHolderInterface;
+use Twig\Environment;
+use Twig\Extension\AbstractExtension;
+use Twig\Extension\EscaperExtension;
+
+class ContaoCompatExtension extends AbstractExtension
+{
+    /**
+     * @internal
```
leofeyer, line 23: Why not at class level?

core-bundle/src/Twig/Compat/ArrayValueHolder.php
m-vo: Imho it can't be class level because you need to consume this class (but not instantiate it).

ausi:
> Twig for Contao templates: no `|raw` needed anymore. 🥳 🎉

I’m not sure if that’s what we want. IMO we should integrate twig templates in a way that they would not have to be changed if we switch from input to output encoding.

How about adding an escaper strategy like `html_contao` that we enable by default for twig templates that are used in place of a Contao template? In this escaper we would use `htmlspecialchars()` with `double_encode` set to `false`. We could then write the templates the same way as we would with output encoding:

```twig
<h1>{{ headline }}</h1> 
<div>{{ content|raw }}</div>
```

In this example the `headline` would already be (input)-encoded by Contao but because of `double_encode: false` escaping doesn’t actually change something but still guarantees the safety of the output.

The `content` would e.g. be from a tinyMCE field which requires the template author to use `|raw`, the same way as it would be with regular twig templates using output encoding.

core-bundle/src/Twig/Compat/ArrayValueHolder.php
leofeyer: But how is someone supposed to use the class if the constructor is internal?

core-bundle/src/Twig/Extension/ContaoCompatExtension.php
bytehead: Your IDE at least marks it when you create an object 🙃

core-bundle/src/Twig/Compat/ArrayValueHolder.php
bytehead: Your IDE at least marks it when you create an object 🙃

m-vo: I think it's the same with every class that has a factory. The constructor isn't part of the BC promise, the factory method constructing it, however, is. In this case you would use `ProxyFactory::createValueHolder(…)` instead of `new ArrayValueHolder(…)`.

We've got the same in place for the studio classes for instance. :slightly_smiling_face:

leofeyer: I see. Maybe we should add a comment: `@internal Do not instantiate this class in your code; use ProxyFactory::createValueHolder() instead`

m-vo:
> I’m not sure if that’s what we want. IMO we should integrate twig templates in a way that they would not have to be changed if we switch from input to output encoding.

I generally agree with your point (and I already started implementing some thing like this but didn't get too far, yet). :smiley_cat: 

IMO, one of our current problems is having prerendered content at all. The output of the RTE should probably be the only exception. And *maybe* it would be valid solution to have a simple proxy object for this case later on as well. In Symfony land I'll rarely ever have to use `|raw` as a consumer. The building blocks (like forms for instance) are really built inside Twig with the help of filters and functions. This is a good paradigm, because it shifts the responsibility of building things to the template: HTML should almost exclusively originate from templates.

So, ideally, we would not only drop input encoding in the future but also migrate away from prerendered content. Together with a special handling for the RTE output a regular user wouldn't have to use encoding filters then. I think it all boils down to two things:
* What are the things we want users to learn about Twig templates and encoding? This is especially important if they start using their own Twig templates (e.g. within a fragment) that then behave differently.
* Would it be helpful in terms of 'user awareness' about encoding if a change *is* needed at certain places when we're switching. Note, that it makes a huge difference to add `|raw` to sanitized tinyMCE output versus to any HTML that was sanitized now but not anymore in the future.

aschempp:
I absolutely agree with @ausi that we should try to be forward-compatible with this Twig integration. I think we had several proposals on that already, but none seems to have worked was well as this (thanks @m-vo!).

I missing here how this is supposed to be the default in the future. If I understand correctly, this would only allow one to override a PHP template with Twig. But would it also allow extensions to use Twig (while fixing the same input/output encoding issues)? And on that spot, Twig does not allow to override a template from another extension, right?

m-vo:
> And on that spot, Twig does not allow to override a template from another extension, right?

It's possible. You need to preprend configuration that includes a new namespace path in your extension, though.

Edit: here is an example: https://stackoverflow.com/a/52693472/3146819

m-vo:
> But would it also allow extensions to use Twig (while fixing the same input/output encoding issues)?

You mean an extension overwriting a core template? Imo this shouldn't make a difference.

The question to me is: Should there also be any support for things like `tinyMCE` output, if you're using Twig directly (i.e. from your fragment), not via `\Contao\Template`. :thinking: And "support" would probably mean "default in the future".

leofeyer:
As discussed in Mumble on January 14th:

- HTML is a separate data type (like text or integer) and should be sanitized before it is stored in the DB.
- If we want to output a HTML string in Twig, we have to add `|raw`.
- A custom Twig encoder that does not double encode could be used to handle strings that are already encoded in the DB (see https://github.com/contao/contao/pull/2612#discussion_r550298990).

leofeyer:
As discussed in Mumble on January 14th, @Toflar is going to create a ticket in which we define the Twig implementation requirements. Then we can discuss again and eventually implement Twig according to these requirements.

m-vo:
superseded by #2988

PR #2614, opened 2020-12-30
Fix the file manager performance with large non-image files

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

If you have one or more large non-image files within `files/`, the TTFB in the file manager of the back end will be significantly impacted with each large file since `DC_Folder::generateTree` always accesses `$objFile->width` and `$objFile->height` for each and every file. This in turn will execute `getimagesize`, even if the file is not actually an image, but for example a large MP4 file instead.

» [blackfire.io profiling](https://blackfire.io/profiles/f454b839-686b-41d2-b3cd-d9b93f53b0b7/graph)

This PR reduces the TTFB significantly by adding a `$this->isImage` check within `__get('imageSize')`.

For example, in my test setup, I had 6 MP4 files in a test folder within `files/`, each ~2.04 GiB in size. With this folder open in the file manager, the TTFB was between _50 to 60 **seconds**_.

With this change, the TTFB is cut down to only only _400 to 800 **ms**_ (note: my setup is fairly slow for some reason, but that's besides the point).

Also this makes me wonder if we should store the image size in the database as well, in order to improve the performance of the file manager even more?


Toflar:
Is it due to open_basedir? ;)

fritzmg:
> Is it due to open_basedir? ;)

No, no `open_basedir` setting active.

ausi:
> This PR reduces the TTFB significantly by adding a `$this->isImage` check within `__get('imageSize')`.

I think we should instead remove the line https://github.com/contao/contao/blob/0e7bb7ab60634f1266c92bbccb643e098e98acf4/core-bundle/src/Resources/contao/library/Contao/File.php#L291 as this only acts as a fallback for images that the image factory cannot handle.

fritzmg:
@ausi but wouldn't it be better in any case to have an early out, if the file extension does not match a supported image type?

ausi:
I’d like to give the image factory the responsibility for that check. The image factory has already an early out for that case in https://github.com/contao/contao/blob/0e7bb7ab60634f1266c92bbccb643e098e98acf4/core-bundle/src/Image/ImageFactory.php#L130

fritzmg:
Alright, changed in c234541900860d67a6120c8f1687fae1fd9e045c

leofeyer:
@fritzmg Can you please fix the CI chain?

fritzmg:
In order to fix the problem reported by the tests I have implemented another bug fix: before `$file->imageViewSize` could be an `array` or `false` - even though we promise it to always be an array. I have changed the responsible code so that instead of setting `$this->arrImageViewSize` to `false` (which is confusing anyway) it is instead again reset to an empty array.

leofeyer (changes requested):
I'm afraid this is a BC break that we must not merge. People might rely on `$file->imageViewSize` to return `false`, so if at all, we should fix the `@property` annotation instead of adjusting the code behavior.

fritzmg:
> Also this makes me wonder if we should store the image size in the database as well, in order to improve the performance of the file manager even more?

I made some tests. Even with a big number of large image files, or a huge number of small image files, there is no discernible performance improvement.

core-bundle/src/Resources/contao/library/Contao/File.php
```diff
@@ -349,10 +344,10 @@ public function __get($strKey)
 				return $this->arrImageViewSize;
 
 			case 'viewWidth':
-				return $this->imageViewSize !== false ? $this->imageViewSize[0] : null;
+				return !empty($this->imageViewSize) ? $this->imageViewSize[0] : null;
```
leofeyer: Why?

fritzmg:
> I'm afraid this is a BC break that we must not merge. People might rely on `$file->imageViewSize` to return `false`, so if at all, we should fix the `@property` annotation instead of adjusting the code behavior.

Done. I haven't touched the `@property` annotations though, as there are more candidates that would need to be fixed.

core-bundle/src/Resources/contao/library/Contao/File.php
fritzmg: `$this->imageViewSize` can be one of the following:

* false
* an empty array
* an array with two elements (width and height)

ausi: I don’t think this is true. `$this->imageViewSize` can never be an empty array. This is only true for `$this->arrImageViewSize` I think.

fritzmg: `$this->imageViewSize` will return an empty array under the following conditions in this PR:

* `$this->imageSize` is falsy 
* `$this->isSvgImage` is falsy

Both of these will be true, if the file is not an image.

fritzmg: Without this change the `ImageFactoryTest::testExecutesTheExecuteResizeHook` test fails for example. But this is actually another problem: within the `ImageFactoryTest::testExecutesTheExecuteResizeHook` test, `$this->imageSize` will return an empty array, even though it is an image - because the service container is missing the `@filesystem` service, which is required by the  `@contao.image.image_factory` service, which is used by `\Contao\File` (or rather the container is missing `@contao.image.image_factory` as a whole). This throws an exception - however, that exception is simply suppressed within `__get('imageSize')` and thus the test does not actually fail because of that. Only later on when `$this->imageViewSize` is accessed.

The test did not fail before this PR, because the `getimagesize` fallback would then be in use, which this PR removes.

fritzmg: The question is now: since we removed the `getimagesize` fallback, should we also remove the try-catch block? We do effectively rely on the image factory to provide the correct image size. Any errors occurring there should not be hidden, I think?

ausi: > `$this->imageViewSize` will return an empty array under the following conditions

You’re right, I missed that 🙈

> since we removed the `getimagesize` fallback, should we also remove the try-catch block?

I don’t think we can. For BC reasons accessing `->imageSize` should not throw an exception I think.

ausi (dismissed):

fritzmg:
I have also included a fix in the `ImageFactoryTest`, which fixes a problem that is revealed due to the changes in this PR + when the try catch block in `__get('imageSize')` would be removed (see https://github.com/contao/contao/pull/2614#discussion_r554393161).

core-bundle/src/Resources/contao/library/Contao/File.php
fritzmg: What if we at least moved `System::getContainer()->get('contao.image.image_factory')` out of the try catch block? That at least reveals the issue in the tests which is otherwise just hidden.

```php
$imageFactory = System::getContainer()->get('contao.image.image_factory');

try
{
	$dimensions = $imageFactory->create($this->strRootDir . '/' . $this->strFile)->getDimensions();

	// …
}
catch (\Exception $e)
{
	// ignore
}
```

core-bundle/src/Resources/contao/library/Contao/File.php
```diff
@@ -349,10 +346,14 @@ public function __get($strKey)
 				return $this->arrImageViewSize;
 
 			case 'viewWidth':
-				return $this->imageViewSize !== false ? $this->imageViewSize[0] : null;
+				$imageViewSize = $this->imageViewSize;
```
leofeyer, line 350: Why did you add a separate variable here?

fritzmg: `empty($this->imageViewSize)` will always be true, because it accesses `__isset` first and not `__get` and `__isset` is not implemented in `Contao\File`.

leofeyer: I see. We should definitely add a comment here. 😄

ausi (dismissed):

ausi (approved):

bytehead (approved):

leofeyer:
Thank you @fritzmg.

PR #2617, opened 2020-12-31
Upgrade scheb/2fa to version 5 (PHP 8 compatibility)

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | n/a
| Docs PR or issue | n/a

This is a follow up from #2592 and is related to #2263.


manager-bundle/src/HttpKernel/JwtManager.php
```diff
@@ -26,30 +26,13 @@ class JwtManager
...
-     */
-    private $parser;
-
-    /**
-     * @var string
-     */
-    private $secret;
-
-    public function __construct(string $projectDir, Signer $signer = null, Builder $builder = null, Parser $parser = null, Filesystem $filesystem = null)
+    public function __construct(string $projectDir, Configuration $config = null, Filesystem $filesystem = null)
```
bytehead: @leofeyer this is a BC break, right?

bytehead: Nvm, got it in 25a0b71.

leofeyer: It is a BC break, however, adding another constructor argument is also a BC break IMHO, because it will result in a "declaration should be compatible with parent" error. We should probably mark the `JwtManager` class as `@internal`, because people are supposed to use the service and not the class anyway.

@aschempp @ausi @Toflar WDYT?

aschempp: `JwtManager` is not a service, it is injected by the kernel. But I would mark the class as `@internal` so we're free to modify that. The class itself certainly isn't API to me.

bytehead: The same question applies to the [`ContaoLoginAuthenticationListener`](https://github.com/bytehead/contao/blob/fix/2fa-php8/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php).

bytehead: @aschempp The `JwtManager` is registered as a service (and public): https://github.com/contao/contao/blob/4.9/manager-bundle/src/Resources/config/services.yml#L25, but directly injected by the kernel: https://github.com/contao/contao/blob/4.9/manager-bundle/src/HttpKernel/ContaoKernel.php#L263

ausi: > because it will result in a "declaration should be compatible with parent" error

This is not true for `__construct()` AFAIK https://3v4l.org/kFcii

leofeyer: Except if the constructor is part of the interface (see https://3v4l.org/mQRa8), which is not the case here I think, is it?

ausi: > which is not the case here I think, is it?

Correct.

ausi (dismissed):
Looks great! 😎

With this PR merged, Contao 4.9 is finally PHP 8.0 compatible 🎉

ausi (dismissed):

core-bundle/src/Resources/config/services.yml
```diff
@@ -628,7 +628,7 @@ services:
             - '@contao.security.authentication_success_handler'
             - '@contao.security.authentication_failure_handler'
             - []
-            - '@scheb_two_factor.token_factory'
+            - ~ # Backwards compatibility
```
leofeyer: There is no need to keep the argument, because event listeners are explicitly excluded from out BC promise: https://docs.contao.org/dev/internals/bc-promise/

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
```diff
@@ -31,24 +31,25 @@
 use Symfony\Component\Security\Http\Session\SessionAuthenticationStrategyInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
+/**
+ * @internal
+ */
```
leofeyer: This is not needed, because event listeners are explicitly excluded from our BC promise: https://docs.contao.org/dev/internals/bc-promise/

core-bundle/tests/Security/Authentication/ContaoLoginAuthenticationListenerTest.php
```diff
@@ -126,19 +125,14 @@ public function testStoresLastUsernameInSession(): void
...
     {
-        $authenticatedToken = $this->createMock(UsernamePasswordToken::class);
-
         $currentToken = $this->createMock(TwoFactorTokenInterface::class);
-        $currentToken
-            ->expects($this->once())
-            ->method('getAuthenticatedToken')
-            ->willReturn($authenticatedToken)
-        ;
+        $currentTokenWithCredentials = $currentToken->createWithCredentials('123456');
```
leofeyer: If you want to use a method of the original class, you have to use `createPartialMock()` IMHO.

manager-bundle/src/HttpKernel/JwtManager.php
```diff
@@ -58,13 +47,40 @@ public function __construct(string $projectDir, Signer $signer = null, Builder $
...
+            $secret = file_get_contents($secretFile);
+        }
+
+        if (!\is_string($secret) || 64 !== \strlen($secret)) {
+            $secret = bin2hex(random_bytes(32));
+            $filesystem->dumpFile($secretFile, $secret);
+        }
+
+        if (null !== $signer) {
+            @trigger_error('Second argument ($signer) of JwtManager::__construct(...) is deprecated since Contao 4.9, to be removed in Contao 5.0. Use the Configuration object instead.', E_USER_DEPRECATED);
```
leofeyer: Since the class is internal, we do not have to keep the old constructor arguments.

leofeyer (changes requested):

core-bundle/src/Resources/config/services.yml
bytehead: I'll remove it then 👍

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
bytehead: 👍 I'll remove it.

manager-bundle/src/HttpKernel/JwtManager.php
bytehead: 👍 I'll remove it.

core-bundle/src/Resources/config/services.yml
bytehead: Fixed in 9ae780b55.

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
bytehead: Fixed in 9ae780b55.

manager-bundle/src/HttpKernel/JwtManager.php
bytehead: Fixed in 2f4a70de1.

core-bundle/tests/Security/Authentication/ContaoLoginAuthenticationListenerTest.php
bytehead: I've mocked the second token as well in 9d7eb5f66.

leofeyer (approved):

leofeyer:
Thanks a lot @bytehead. 🚀

bytehead:
Nice 💪 🎉

m-vo:
Yeah, PHP 8.0 is coming! Loving it.

bytehead:
It's around the corner in #2263 😬

scheb:
Author of `scheb/2fa-bundle` here. Are you sure you want to reference `scheb/2fa` as a dependency? Because that's the all-in-one package, which includes all sub-packages with all their dependencies and all sub-packages that might be added in the future - even when you're not using it. It's the easy & quick solution, but you're likely pulling in a certain overhead of dependencies.

Instead, I'd recommend you use `scheb/2fa-bundle` and the sub-packages that you specifically need for Contao:

```
scheb/2fa-backup-code            # Add backup code feature
scheb/2fa-trusted-device         # Add trusted devices feature
scheb/2fa-totp                   # Add two-factor authentication using TOTP
scheb/2fa-google-authenticator   # Add two-factor authentication with Google Authenticator
scheb/2fa-email                  # Add two-factor authentication using email
scheb/2fa-qr-code                # Add to render QR-codes for Google Authenticator / TOTP
```

bytehead:
Hey @scheb

Thanks for the hint. I remember thinking about the topic as well while I was implementing the changes - don't know why I've missed it tough.

As far as I can see this, we should use

- `scheb/2fa-bundle`
- `scheb/2fa-backup-code`
- `scheb/2fa-trusted-device`

instead of `scheb/2fa`.

WDYT @contao/developers?

Toflar:
Hey @scheb, thanks for the feedback, it's highly appreciated! Sure, we should remove the dependency on the meta package and only require what we really need! I didn't notice that yet :)

bytehead:
@Toflar I'll create a PR for that 👍

Toflar:
Cool, thanks! I'll try to remember we owe Christian a few beers on the next occasion ;)

bytehead:
>  I'll try to remember we owe Christian a few beers on the next occasion ;)

Definitely!

m-vo:
Btw.: is there a reason we're using `bacon/qr-code` directly not via `scheb/2fa-qr-code`?

bytehead:
@m-vo This is new in version 5.x, I could also adapt this. I'll check it in the PR.

scheb:
> Cool, thanks! I'll try to remember we owe Christian a few beers on the next occasion ;)

Beer is always appreciated ;)

> Btw.: is there a reason we're using `bacon/qr-code` directly not via `scheb/2fa-qr-code`?

The package was added with 5.x as a ready-to-go solution for QR codes. In 4.x it was all up to you to implement the QR code.

bytehead:
@scheb thanks, now I remember 😄

PR #2632, opened 2021-01-12
Deserialize the image size when using maxWidth

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

We must deserialize the image size when someone is using the legacy `maxWidth`/`maxImageWidth` parameter/config option in `Controller::addImageToTemplate`.

Proof of failing test is in 92ad5f338ae79130fb7a6d3806c299564c15acc1.


fritzmg (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1589,6 +1589,8 @@ public static function addImageToTemplate($template, array $rowData, $maxWidth =
 			}
 
 			// Normalize size
+			$size = StringUtil::deserialize($size);
```
leofeyer: This cannot be the correct place to deserialize the string. It should be in line 1605 inside the `else` condition IMHO, don't you think?

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/Controller.php
m-vo: We're also deserializing any input in the `ImageFactory` and that's what was in place [before](https://github.com/contao/contao/blob/5feb9da96b80ffad5791d8c1b14f363ca2f22d79/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1536) as well. 

So - in theory - someone could rely on it working with a single serialized integer for example.

leofeyer: BTW, did you mean to pass `$rowData['size'] ?? ''`? It seems that `$size` is not declared at this point.

leofeyer: That is weird as well then. 😄 If we know that `$size` can be numeric or an object, why would we try to deserialize it without checking if deserializing is necessary first?

m-vo: It could for instance be a numeric value (`"123"`) or a serialized numeric value (`s:3:"123";`). This is at least what the old implementation allowed and the current `PictureFactory` is allowing.

In fact `$size` can imo be one of `int|string|array|PictureConfiguration|null` with `string` including all possible serialized versions of`int|string|array|null`.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: You are missing the point.

```php
$size = StringUtil::deserialize($size);

if ($size instanceof PictureConfiguration)
{
    return array($size, $margin);
}
```

`StringUtil::deserialize()` does not allow serialized objects, so this condition can only be true if `$size` is already an object **before** it is passed to `StringUtil::deserialize()`. It therefore does not make sense to have the `instanceof` check **after** `StringUtil::deserialize()`. It should be:

```php
if ($size instanceof PictureConfiguration)
{
    return array($size, $margin);
}

$size = StringUtil::deserialize($size);
```

m-vo: Yes, I know and `StringUtil::deserialize()` is the identity function for objects :slightly_smiling_face:. I was referring to your initial comment suggesting to put the line in the `else` clause where it would miss the serialized numeric case.

Nevermind, I moved it after the `instanceof` check in 1b0bc6b6694237d5cd4cebe007d58ac7863cc670.

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #2635, opened 2021-01-14
Loosen type constraint for `$rowData['fullsize']`

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

We should not force `$rowData['fullsize']` to be exactly `'1'` but also allow other trueish values.


ausi (approved):

bytehead (approved):

leofeyer:
Thank you @m-vo.

PR #2636, opened 2021-01-14
Use the correct User-Agent request header in Escargot

qzminski:
Up until now the Escargot crawler used `Symfony HttpClient/Curl` as a User-Agent header. This PR fixes it to use the correct `contao/crawler` value.

### DIFF
--- a/core-bundle/src/Crawl/Escargot/Factory.php
+++ b/core-bundle/src/Crawl/Escargot/Factory.php
@@ -165,10 +165,7 @@ class Factory
      */
     public function create(BaseUriCollection $baseUris, QueueInterface $queue, array $selectedSubscribers, array $clientOptions = []): Escargot
     {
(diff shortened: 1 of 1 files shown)


Toflar (approved):
That change is correct.
I've also added a check in Escargot itself to prevent this from happening in the future: https://github.com/terminal42/escargot/commit/0090c57f5879350f759696283c5be18e2448dd7c

Note: I did not release this yet as it would break all 4.9 setups. It's not a BC break on Escargot because the wrong usage has been documented all along. However, I think it's worth waiting for the next 4.9 release :)

Toflar:
BTW, this PR also fixes this check from not being applied: https://github.com/contao/contao/blob/master/core-bundle/src/EventListener/SearchIndexListener.php#L65 :)

leofeyer:
Thank you @qzminski.

PR #2639, opened 2021-01-14
Correctly match the language URL if there is no trailing slash

aschempp:
If legacy mode is disabled, Contao 4.10 does currently not correctly redirect `https://example.org/de` to `https://example.org/de/`. Symfony actually performs a redirect automatically if the slash is missing, but we need to make sure to find the respective root/index page through our route candidates.


ausi (approved):

aschempp:
Failing tests are not related to this PR

bytehead (approved):

PR #2660, opened 2021-01-20
Remove the separate CI run for PHP 8

leofeyer:


ausi:
Should we add a run for `8.1` to test the nightly builds (with allowing failure)?

leofeyer:
See #3036

PR #2661, opened 2021-01-20
Use type="url" for httpurl text fields

fritzmg:
When using `rgxp => httpurl` the form field type in the front end will be `type="text"` instead of `type="url"` currently. This PR fixes that.


leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #1909, opened 2020-07-05
Correctly handle custom default templates of fragments

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1819 #1905
| Docs PR or issue | -

This PR replaces all `customTpl.options_callback`s with a service. It will use the defined custom default template of a fragment for `Controller::getTemplateGroup`.

Since the custom template that is registered via the service tag isn't accessible anywhere I decided to simply register those templates directly in the callback service during the compiler pass.

This includes a fix for #1905, which will override the template group to `form_textfield`, if the form field type is `text`.

This also includes a fix for the `overrideAll` mode of `tl_content` and `tl_module` (same as the one already used for `tl_form_field`).

### Content Element

```php
// src/Controller/ContentElement/ExampleElementController.php
namespace App\Controller\ContentElement;

use Contao\ContentModel;
use Contao\CoreBundle\Controller\ContentElement\AbstractContentElementController;
use Contao\CoreBundle\ServiceAnnotation\ContentElement;
use Contao\Template;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;

/**
 * @ContentElement(category="texts", template="ce_foo")
 */
class ExampleElementController extends AbstractContentElementController
{
    protected function getResponse(Template $template, ContentModel $model, Request $request): ?Response
    {
        return new Response('<p>Hello World!</p>');
    }
}
```

### Front end module

```php
// src/Controller/FrontendModule/ExampleModuleController .php
namespace App\Controller\FrontendModule;

use Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController;
use Contao\CoreBundle\ServiceAnnotation\FrontendModule;
use Contao\ModuleModel;
use Contao\Template;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;

/**
 * @FrontendModule(category="miscellaneous", template="mod_foo")
 */
class ExampleModuleController extends AbstractFrontendModuleController
{
    protected function getResponse(Template $template, ModuleModel $model, Request $request): ?Response
    {
        return new Response('<p>Hello World!</p>');
    }
}
```

PR #2666, opened 2021-01-20
Reset the custom template if the element type changes

### LINKED ISSUE #2567: Unbekannte Option bei Template Auswahl
**Affected version(s)**

Contao 4.9.9

**Description**

Beispielszenario:

1. Es gibt für ce_headline ein eigenes Template "ce_headline_test.html" (nur anderer Name, Inhalt gleich)
2. Ich lege ein Element Überschrift an, weise ihm das Template zu und speichere
3. Ich ändere den Elementtyp auf Text
4. Im Dropdown "Inhaltselemente-Template" steht nun "Unbekannte Option"
5. Ich speichere
6.  Im Frontend wird nun folgendes Markup erzeugt:

`<h2 class="ce_text first last"></h2>`

Das führt natürlich zu Fehldarstellungen auf der Website.

step1
step2

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2567
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -25,6 +25,10 @@ $GLOBALS['TL_DCA']['tl_content'] = array
 			array('tl_content', 'filterContentElements'),
 			array('tl_content', 'preserveReferenced')
 		),
+		'onsubmit_callback'             => array
+		(
+			array('tl_content', 'resetTemplate')
+		),
 		'sql' => array
 		(
 			'keys' => array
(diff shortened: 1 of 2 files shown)


fritzmg (changes requested):
That won't work for elements that use a custom template. If we want to do this, then I think we need to create an independent "default template finder" service which will then used by this callback and the callback from https://github.com/contao/contao/pull/1909

leofeyer:
You are right. I have reworked the implementation in a76a3fead81903ea2460a98b74399624929dc722 so it no longer requires to know the default template name.

fritzmg:
New implementation looks good to me. However should we really introduce new functionality in the legacy `tl_` classes? I think any new callbacks should always be implemented as a service with appropriate service tags.

leofeyer:
If that is our claim, such small issues will never be fixed anymore. The relationship between effort and benefit should be reasonable. 🤷🏻‍♂️

fritzmg (approved):
The effort would be low if we enabled auto-wiring for event listeners 🙈 🐱‍👤

leofeyer:
Maybe, if it were not for the effort of discussing this and convincing everyone. 😉

PR #2667, opened 2021-01-20
Ignore Monolog log files when rotating log files

### LINKED ISSUE #2579: Contao\Automator::logRotate prevents monolog log rotate
**Affected version(s)**

Contao 4.x

**Description**

The `Contao\Automator::logRotate` rotates all log files in `system/logs`. In fact `system/logs` is a symlink to `var/logs` so it affects all log files generated by monolog. The issue occurs because of the different log file formats. Contao expects a fixed log file name, like `error.log`. The monolog configuration creates a date based error log file `prod-YYYY-MM-DD.log`. The Contao log rotation adds a numeric suffix to the files, which doesn't make any sense for date based files. As the result the the log directory is spamed with date based log files having the suffix `.1`.

This issues does not appear in every Contao project but it depends on the order of log entries and the time when the Contao cron job is called. In our system, it's triggered with a cronjob at `00:00:00` (last run finished at `2020-12-17 00:00:17`).

**Proposed sulition**

The `Contao\Automator::logRotate` has to ignore all files having a date based format.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2579
| Docs PR or issue | -

@dmolineus Does this fix your issue?

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -472,7 +472,7 @@ class Automator extends System
 		foreach ($arrFiles as $strFile)
 		{
 			// Ignore Monolog log files (see #2579)
-			if (preg_match('/^[a-z]+-\d{4}-\d{2}-\d{2}\.log$/', $strFile))
+			if (preg_match('/-\d{4}-\d{2}-\d{2}\.log$/', $strFile))
 			{
 				continue;
 			}
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -471,6 +471,12 @@ class Automator extends System
 
 		foreach ($arrFiles as $strFile)
 		{
+			// Ignore Monolog log files (see #2579)
+			if (preg_match('/^[a-z]+-\d{4}-\d{2}-\d{2}\.log$/', $strFile))
+			{
+				continue;
+			}
+
 			$objFile = new File('system/logs/' . $strFile . '.9');
 
 			// Delete the oldest file


core-bundle/src/Resources/contao/library/Contao/Automator.php
```diff
@@ -471,6 +471,12 @@ public function rotateLogs()
 
 		foreach ($arrFiles as $strFile)
 		{
+			// Ignore Monolog log files (see #2579)
+			if (preg_match('/^[a-z]+-\d{4}-\d{2}-\d{2}\.log$/', $strFile))
```
aschempp: this wouldn't match `isotope_postsale-2020-01-01.log`. I think we should only match on the date at the end of the file name, wdyt?

leofeyer: Changed in babdc77f550dd602153eb3e26ab7dcd03b2ee360.

core-bundle/src/Resources/contao/library/Contao/Automator.php
```diff
@@ -471,6 +471,12 @@ public function rotateLogs()
 
 		foreach ($arrFiles as $strFile)
 		{
+			// Ignore Monolog log files (see #2579)
+			if (preg_match('/-\d{4}-\d{2}-\d{2}\.log$/', $strFile))
```
aschempp, line 475:
```suggestion
			if (preg_match('/(-|_)\d{1,4}(-|_|\.)\d{1,2}(-|_|\.)\d{1,4}\.log$/', $strFile))
```

This would also match
 - `foo_2020-01-01.log`
 - `foo-2020.01.01.log`
 - `foo-01.01.2020.log`
 - `foo__2020-1-1.log`
 - `foo__1_1_2020.log`

leofeyer: Why? We never use these formats, do we?

aschempp: This isn't about "we" the Contao core, because we don't use the old log handler anymore? `log_message` does not create a file with a date, so such files would "manually" be created by developers use the old method, and they're free to choose the date format?

leofeyer: I think you are missing the point of the if clause. It does not define which files to **handle** but which files to **ignore**. And in my opinion, we must only ignore the Monolog date format, which is:

* RotatingFileHandler::FILE_PER_DAY ("Y-m-d")
* RotatingFileHandler::FILE_PER_MONTH ("Y-m")
* RotatingFileHandler::FILE_PER_YEAR ("Y")

https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/RotatingFileHandler.php#L87-L90

So if a log file is named `foo-01.01.2020.log`, we can assume that it is **not** handled by Monolog and that we have to rotate it ourselves.

aschempp: My initial thought was to not rotate files that already include a date, because they are "recreated" each day anyway. However, that would not **delete** those files after 10 days, which is what the Automator is supposed to do. Therefore I agree we should only match on the date format of Monolog.

aschempp (approved):

dmolineus:
I don't like the solution because it assumes monolog uses the default date format. If anyone decides to [customize it](https://github.com/symfony/monolog-bundle/blob/b9b322d5aab3aff15d3055b990b8f93276cb6079/DependencyInjection/Configuration.php#L67) and the issue is back again.

aschempp:
> I don't like the solution because it assumes monolog uses the default date format. If anyone decides to [customize it](https://github.com/symfony/monolog-bundle/blob/b9b322d5aab3aff15d3055b990b8f93276cb6079/DependencyInjection/Configuration.php#L67) and the issue is back again.

Yes, this handles the default Managed Edition. Any idea for improvements? Also, did you ever override the default date format?

Issue #2637, opened 2021-01-14
Improve handling of boolean entity fields (DCA)


m-vo:
**Affected version(s)**

Contao 4.9

**Description**

This bit in `DC_Table` sets a default value if the current one is considered *empty*: 

https://github.com/contao/contao/blob/5fd76d55ab9266027585eea6d7ace08730af49a7/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3172-L3176

That's a bit of a bummer when using DCA with entities as their data definition. With a checkbox + boolean field type, I already need to keep adding a `save_callback` like this to get the right type:
```php
'save_callback' => [static fn ($v) => '1' === $v]`
```

But due to the string cast, yet another `sql` definition in the DCA is needed just to prevent `DC_Table` from saving an empty string again:
```php
'sql' => ['type' => 'boolean', 'default' => false]`
```

Maybe there's a clever way to improve this situation a but without breaking anything else. :smiley:

fritzmg:
I suppose we have to query the database directly for the field's type, instead of relying on `sql` within the DCA?

aschempp:
Contao uses "" and "1" for Boolean, maybe we could fix that by providing a `boolean_string` datat type (similar as our `binary_string`)?

fritzmg:
> Contao uses "" and "1" for Boolean

It works with `'sql' => ['type' => 'boolean', 'default' => false]` too though.

m-vo:
But not without the sql definition which is the default when using a DCA on entities.

aschempp:
Ah yes, our default value parser does not use the metadata. I've run into this as well with null-defaults for integer fields for example.

leofeyer:
@aschempp Why did you remove the "up for discussion" label? Is there a solution already?

fritzmg:
Related issue: even if you use a DCA, but with `'sql' => ['type' => 'boolean', 'default' => false]` (instead of Contao's usual `char(1)`), the following error will occur in the back end when trying to save a record:

```
An exception occurred while executing 'UPDATE tl_calendar_events SET `foobar`='' WHERE id='15'':

SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect integer value: '' for column `foobar`.`tl_calendar_events`.`foobar` at row 1
```

This was actually introduced by https://github.com/contao/contao/pull/2588
Before that fix, the empty value would have been set to `false`, rather than an empty string.

[...]

PR #2668, opened 2021-01-20
Fix empty value for boolean type

fritzmg:
See https://github.com/contao/contao/issues/2637#issuecomment-763886537

If you use `'sql' => ['type' => 'boolean', 'default' => false]` for your checkbox widget in `4.9.x-dev` or `4.11.0-RC1`, the unchecked value cannot be saved, because `Widget::getEmptyValueByFieldType()` would default to an empty string now.

This PR adds support for boolean field types.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -1454,6 +1454,11 @@ abstract class Widget extends Controller
 					return 0;
 				}
 
+				if ($sql['type'] === Type::BOOLEAN)
+				{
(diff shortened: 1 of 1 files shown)


leofeyer (approved):

aschempp:
Can't this be solved by using doctrine metadata to get the actual SQL default?

leofeyer:
Thank you @fritzmg.

leofeyer:
@aschempp This is only a follow-up fix on #2588 (see https://github.com/contao/contao/issues/2637#issuecomment-763886537). We should discuss your question in #2637.

PR #2669, opened 2021-01-21
Use a listener to reset custom templates

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2666

I'm sorry for the late response about #2666, @fritzmg is right there. The behaviour of adding new methods to `tl_` classes is discouraged and something we agreed not to do. Doing the same in a listener is much easier and more universal, see this PR changes. It also means we don't need to duplicate the code in multiple places.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/ResetCustomTemplateListener.php
+++ b/core-bundle/src/EventListener/DataContainer/ResetCustomTemplateListener.php
@@ -12,15 +12,15 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\EventListener\DataContainer;
 
+use Contao\CoreBundle\ServiceAnnotation\Callback;
 use Contao\DataContainer;
 use Doctrine\DBAL\Connection;
-use Contao\CoreBundle\ServiceAnnotation\Callback;
 
 /**
  * @internal
  *
- * @Callback(table="tl_content", target="fields.customTpl.save")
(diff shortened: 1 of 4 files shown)


core-bundle/src/EventListener/DataContainer/ResetCustomTemplateListener.php
```diff
@@ -0,0 +1,70 @@
...
+
+use Contao\DataContainer;
+use Doctrine\DBAL\Connection;
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+
+/**
+ * @internal
+ *
+ * @Callback(table="tl_content", target="fields.customTpl.save")
+ * @Callback(table="tl_module", target="fields.customTpl.save"
```
leofeyer: The callback is registered on the `type` field, not on the `customTpl` field.

leofeyer (changes requested):

leofeyer:
I have fixed the implementation, however, I actually cannot merge it, because the unit tests are missing. And here at least it becomes clear what I meant by "the relationship between effort and benefit should be reasonable". 🤷🏻‍♂️

aschempp:
Maybe we should rethink the test policy if that's the reason to write code in the legacy code space 🤷

I fully agree on _the relationship between effort and benefit should be reasonable_, that's why I don't think this listener deserves a test. But the good thing is, we can always still add one, **because it is a listener and not legacy code** 😉

PR #2674, opened 2021-01-21
Handle image URLs with a {{file::*}} insert tag in the lightbox

fritzmg:
By default `Controller::addImageToTemplate` acts in such a way that if you provided an `imageUrl` with a local image and enabled `fullsize` then the `lightbox` attribute is automatically added, so that the target image is displayed in the lightbox and not in a new browser window/tab.

For example:

1. Create a new image content element.
2. Enable _Overwrite meta_.
3. Set the image URL to `files/foobar/some-image.jpg`.
4. Enable _Full size/new window_.

This will open the image under the provided image URL in the lightbox.

**However**, if you instead do the following:

1. Create a new image content element.
2. Enable _Overwrite meta_.
3. Select the same file from the previous example, but this time via the universal picker in the file manager.
4. Enable _Full size/new window_.

then the same image from before will only open in a new tab/window, instead of in a lightbox.

The reason is that the logic in `Controller::addImageToTemplate` checks whether the provided image URL is a local image - but in the latter example the provided image URL is a `{{file::*}}` insert tag and thus it is not detected a a local image.

This PR fixes that by replacing the insert tag beforehand.


fritzmg:
Contao 4.10+ is also affected by this. @m-vo can you work on that (if this gets merged)? 🙃

m-vo:
Shouldn't the insert tag be replaced beforehand? We're currently only replacing an insert tag automatically when taking the resource from file metadata.

fritzmg:
Beforehand where? It needs to be done somewhere here: 

https://github.com/contao/contao/blob/7bc1599f4ea3f26c5402efc2ef6c8e1a3b222d16/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1695-L1740

So after the meta data (and overwriteMeta) has been loaded/applied and before `imageUrl` is analysed.

m-vo:
> Beforehand where?  

Before passing it to `Controller::addImageToTemplate` ideally, I think.

fritzmg:
> Before passing it to `Controller::addImageToTemplate` ideally, I think.

That won't suffice. The same issue arises if you define an `imageUrl` in the file manager. `Controller::addImageToTemplate` fetches the meta data from the file model where the URL will still be an insert tag.

m-vo:
What I'm thinking is (>= 4.10): We could easily handle insert tags in the `FigureBuilder#from` method that auto-detects the resource type. But  I'm not sure if this should be the responsibility of the `FigureBuilder` or rather the consumer calling it. It could well be that implementing the 'consumer side' solution is way more cumbersome because it's used all over the place, because we're not using `Controller::addImageToTemplate` as a 'legacy bridge' anymore.

ausi:
Shouldn’t this already be fixed in Contao 4.10 in the following line? https://github.com/contao/contao/blob/8f91578a8ff1025227ea05e35c2b42584dcbbc0b/core-bundle/src/File/ModelMetadataTrait.php#L50

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1733,15 +1733,17 @@ public static function addImageToTemplate($objTemplate, $arrItem, $intMaxWidth=n
 		// Image link
 		if (TL_MODE == 'FE' && $arrItem['imageUrl'])
 		{
-			$objTemplate->$strHrefKey = $arrItem['imageUrl'];
+			$imageUrl = urldecode(self::replaceInsertTags($arrItem['imageUrl']));
```
ausi: I don’t think we can use `urldecode` here because it would potentially break external URLs.

fritzmg: Hm, yeah, the problem is that we generally cannot know, whether the content in `$arrItem` is url encoded or not, ~~no matter if it comes from an Insert Tag or if the URL is put in manually~~. Actually it only affects Insert Tag, since the back end field requires you to have the URL already encoded.

fritzmg:
> Shouldn’t this already be fixed in Contao 4.10 in the following line?

Actually in Contao `4.10.x-dev` neither version works, as far as I have tested. Contao never generates a Lightbox Link.

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1762,7 +1764,7 @@ public static function addImageToTemplate($objTemplate, $arrItem, $intMaxWidth=n
 						}
 						catch (\Exception $e)
 						{
-							$objTemplate->$strHrefKey = static::addFilesUrlTo(System::urlEncode($arrItem['imageUrl']));
+							$objTemplate->$strHrefKey = static::addFilesUrlTo($arrItem['imageUrl']);
```
fritzmg: I have removed the call to `System::urlEncode` here as it seems wrong to me. `$arrItem['imageUrl']` _must_ already be URL encoded, if a user enters a relative or absolute URL in the back end. If the content is an insert tag, then it _must_ be assumed that the insert tag outputs url encoded content (like the `{{file::*}}` insert tag does).

core-bundle/src/Resources/contao/library/Contao/Controller.php
fritzmg: I have updated the PR to only specially handle `{{file::*}}` insert tags instead, where we know the output is url encoded and can be safely decoded.

m-vo:
I tried to reproduce this in the Contao demo. But no matter if I put `files/contaodemo/media/content_images/DSC_5316.jpg` in the `Image link target` field or use the file picker (which results in `{{file::c67514df-d9c8-11e3-8217-000c29e44aea}}`) I'll get a working lightbox. Both in 4.10 as well as 4.11.

fritzmg:
@m-vo hm, not sure what would be different in my case. My local `4.10.x-dev` installation has no significant adjustments and currently no extensions installed. When I use an image content element with the following settings:

Screenshot_2021-01-25 Articles › Home › Content elements › Edit content element ID 26 c410 local

Then the resulting HTML will be

```html
<div class="ce_image block">
  <figure class="image_container">
    <a href="files/dts/Death_to_stock_photography_weekend_work%20%284%20of%2010%29.jpg" target="_blank">
      <img src="assets/images/0/Death_to_stock_photography_weekend_work%20%283%20of%2010%29-a7fa5d04.jpg" alt="" itemprop="image" width="125" height="83">
    </a>
  </figure>
</div>
```

fritzmg:
@m-vo you need to test with a file path that contains special characters like spaces. Then this function

https://github.com/contao/contao/blob/563b6c1bf2949cdb48a57266ff4a319ed89f2f32/core-bundle/src/Image/Studio/FigureBuilder.php#L586-L611

will always return `[null, null]`, because the path is url encoded, which in turn will cause `FigureBuilder::onDefineLightboxResult` to return null.

m-vo:
Followup PR for 4.10 and later: https://github.com/contao/contao/pull/2716

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: Not sure about that. We generally should not remove code just because "it seems wrong". At least search the history and check why the code has been added.

fritzmg: Reproduction on the `4.9` branch or `4.9.11` tag:

* Put a file called `foo bar.jpg` into the `web/` folder.
* Create a new image content element.
* Enable **Overwrite meta data**.
* Set the **Image link target** to the publicly available file under `web/`, i.e. `foo%20bar.jpg`.
* Open the front end.

Generated output:

```html
<figure class="image_container">
  <a href="foo%2520bar.jpg" data-lightbox="eafe6b">
    <img src="…" alt="" itemprop="image">
  </a>
</figure>
```

The **Image link target** is erroneously double encoded due to

https://github.com/contao/contao/blob/b8f28f3fd33c52e6f0095deb7bebf431bd2a0bb0/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1765

However, this section of code is only ever reached if an error occurs when processing the linked image - e.g. if the **Image link target** is url encoded (which it actually always is).

core-bundle/src/Resources/contao/library/Contao/Controller.php
fritzmg: I don't like this solution either though, because it will only handle `{{file::*}}` properly. But if you set the **Image link target** to `files/foo%20bar/foo%20bar.jpg` for example then this file is again not processed by Contao, even though it could.

Should we instead simply url decode and encode all **Image link target**s that are _not_ external?

leofeyer (changes requested):
The fix seems wrong to me. 🙈

An image that is added via `{{file::` can never be an external image, therefore, instead of replacing the insert tag, we can simply assume that `$blnIsExternal` is `false`.

```php
$blnIsExternal = strncmp($arrItem['imageUrl'], '{{file::', 8) !== 0 && (strncmp($arrItem['imageUrl'], 'http://', 7) === 0 || strncmp($arrItem['imageUrl'], 'https://', 8) === 0);
```

(Not sure if three consecutive `strncmp()` are the most effective way to check the preconditions here.)

fritzmg:
> An image that is added via `{{file::` can never be an external image, therefore, instead of replacing the insert tag, we can simply assume that `$blnIsExternal` is `false`.

Yes, see also https://github.com/contao/contao/pull/2674#discussion_r576212518

core-bundle/src/Resources/contao/library/Contao/Controller.php
ausi: I think we should use the same logic as in #2716. First check if the URL is external, if not, urldecode and check if the path exists on disk.

fritzmg: Should we always replace all insert tags though? I don't know how this is handled in the Image Studio.

leofeyer: Replacing insert tags is expensive and should only be done if really necessary.

fritzmg: > Replacing insert tags is expensive and should only be done if really necessary.

But we already do it in Contao 4.10 and up: https://github.com/contao/contao/pull/2674#issuecomment-766356141

leofeyer: I see. We should not do it there, either (if possible).

fritzmg: We could limit it to `{{file::*}}`. But then any custom insert tag that would return a valid, url encoded, local path to an image will not get processed.

ausi: IMO replacing all insert tags is required because of the possibility of custom insert tags (e.g. created via a picker).

> Replacing insert tags is expensive

Why is that? If `preg_split` doesn’t match anything is sould return immediately and should not cost much performance IMO.

m-vo: > I don't know how this is handled in the Image Studio.

There is no replacing of insert tags in the Image Studio code but in the `ModelMetadataTrait`. There we *do* call `replaceInsertTags()` on each `Metadata::VALUE_URL` though in order to guarantee that we're dealing with valid paths.

leofeyer: Ok, in this case I guess it is fine to replace all insert tags.

fritzmg:
I have incorporated the changes following the discussion above:

* Insert Tags for the `imageUrl` are now always replaced when processing the `fullsize` part.
* The content of the `imageUrl` is only decoded before being passed to the `PictureFactory`.

aschempp:
Can someone explain how **not** overriding metadata could ever work with a `{{file::` insert tag? Also, I'm not sure if the Controller method should be responsible for all possible use cases, or (as said before) the method user should make sure the data is useful (e.g. replace insert tags).

ausi (dismissed):

fritzmg:
> Can someone explain how not overriding metadata could ever work with a {{file:: insert tag?

What do you mean? It doesn't matter if you use `overwriteMeta` or not. The result is the same.

> Also, I'm not sure if the Controller method should be responsible for all possible use cases, or (as said before) the method user should make sure the data is useful (e.g. replace insert tags).

You mean changing it in all content elements and all modules? I am not sure that's worth it.

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1762,7 +1767,7 @@ public static function addImageToTemplate($objTemplate, $arrItem, $intMaxWidth=n
...
```
leofeyer: Shouldn't this be `$imageUrl` instead of `$arrItem['imageUrl']`?

fritzmg: Changed in fe3df76

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2685, opened 2021-01-25
Use the request attribute to determine preview mode

### LINKED ISSUE #2460: No preview bar on 404 pages
**Affected version(s)**

Contao 4.9+

**Description**

Since the preview bar is rendered using javascript, it is no longer possible to preview an unpublished page. This is especially true if the root page is not published, which means the whole domain cannot be previewed.

In the iframe context, previewing a non-published root page meant the preview iframe would load, but the page content would be empty. With the new setup, the preview bar is not added to the page at all.

No idea how/if we can fix this @richardhj

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2460

This fixes #2460 because of https://github.com/contao/contao/pull/2685/files#diff-d8b64ca1e6f2e35530ef0c68a0f1dd2c5978bad268025a592d6c194883026b44R60
It adds the preview toolbar on every html response that is not a back end request. I'm not sure that's what we really want, but why would anyone use the `preview.php` to get a non-Contao page?

### DIFF
--- a/core-bundle/src/EventListener/PreviewAuthenticationListener.php
+++ b/core-bundle/src/EventListener/PreviewAuthenticationListener.php
@@ -58,6 +58,7 @@ class PreviewAuthenticationListener
 
         if (
             !$request->attributes->get('_preview', false)
+            || $this->scopeMatcher->isBackendRequest($request)
             || $this->tokenChecker->hasBackendUser()
         ) {
             return;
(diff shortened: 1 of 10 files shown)


Toflar:
I think that makes sense. The `preview.php` boots the `ContaoKernel` anyway so it's already Contao specific.

Toflar (approved):

PR #2694, opened 2021-01-26
Remove the legacy encryption logic

Toflar:
We've deprecated it a while ago but that doesn't really help us much because it will break as of PHP 7.2 and we're now at a minimum required version of 7.3 for 4.11.

The deprecation message and all the code that still resides in the class somewhat still implies that it works. Which is does absolutely not. So I prefer to remove it completely from the core.

### DIFF
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -23,11 +23,3 @@ types, e.g. from an article to a news item or from a news item to an event.
 They can only copy or move elements from e.g. one article to another article.
 
 More information: https://github.com/contao/core/issues/5234
-
-## Unique checks on encrypted data
-
-The DCA option `'encrypt'=>true` cannot be used together with `'unique'=>true`,
-because there is no effective way to check the unencrypted values for
-duplicate entries.
-
-More information: https://github.com/contao/core/issues/8144
--- a/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
+++ b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
@@ -66,7 +66,6 @@ class ContaoCoreExtension extends Extension
         $loader->load('migrations.yml');
 
         $container->setParameter('contao.web_dir', $config['web_dir']);
-        $container->setParameter('contao.encryption_key', $config['encryption_key']);
         $container->setParameter('contao.upload_path', $config['upload_path']);
         $container->setParameter('contao.editable_files', $config['editable_files']);
         $container->setParameter('contao.preview_script', $config['preview_script']);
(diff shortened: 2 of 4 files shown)


core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -1400,12 +1388,6 @@ public static function getAttributesFromDca($arrData, $strName, $varValue=null,
 			}
 		}
 
-		// Warn if someone uses the "encrypt" flag (see #8589)
```
aschempp: shouldn't we keep this and warn that encryption is not supported anymore?

aschempp (dismissed):

core-bundle/src/Resources/contao/library/Contao/Widget.php
Toflar: As I stated: It doesn't help anything at all. It does not work. Why would I have a warning for anything that does not work?

aschempp: because people might expect it to work? That's why we use deprecations, to tell people about changes in our code base …

Toflar: You use deprecation messages to inform people about something they shouldn't use anymore but it still works. Which is what we did for the last 3 years. But now it's not deprecated anymore. It doesn't work.

fritzmg: May be it should be converted to an exception now instead, not just a warning.

Toflar: No. That's what you guys don't get. The exception was thrown ever since people used PHP 7.2. The `if (!\in_array('mcrypt', get_loaded_extensions()))` was always `true` so there is no way there's any installation out there still using `encrypt -> true` in their DCA.

fritzmg: > so there is no way there's any installation out there still using encrypt -> true in their DCA.

Why would you assume that? If the extension or application only allowed or used older PHP versions so far and then updated to PHP >=7.3 and Contao >=4.11, there would be no indication that using `encrypt => true` does nothing.

ausi: I’m with @fritzmg here. If someone uses `'encrypt' => true` (because of an update or an outdated tutorial/article) we should throw an exception.

Toflar: I've updated the PR to keep all the `encrpyt -> true` references which will now result in an exception right away.

ausi (dismissed):

bytehead (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #2698, opened 2021-01-27
Handle root pages without hostname in the SitemapController

### LINKED ISSUE #2692: SitemapController returns 404 for root pages with empty hostname
**Affected version(s)**

Contao 4.11-RC1

**Description**

In https://github.com/contao/contao/blob/ac8a518c3cdfddd1c67c12143a9684c2dab4449e/core-bundle/src/Controller/SitemapController.php#L41 we only search for root pages with matching host name. This fails for setups with an empty hostname.

If we find no matching root page, we should probably fallback to the root page with `dns = ''`.

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/2692

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -41,7 +41,12 @@ class SitemapController extends AbstractController
         $rootPages = $pageModel->findPublishedRootPages(['dns' => $request->server->get('HTTP_HOST')]);
 
         if (null === $rootPages) {
-            return new Response('', Response::HTTP_NOT_FOUND);
+            // We did not find root pages by matching host name, let's fetch those that do not have any domain configured
+            $rootPages = $pageModel->findPublishedRootPages(['dns' => '']);
+
+            if (null === $rootPages) {
+                return new Response('', Response::HTTP_NOT_FOUND);
+            }
         }
 
         $pages = [];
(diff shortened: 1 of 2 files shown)


core-bundle/src/Controller/SitemapController.php
```diff
@@ -41,7 +41,12 @@ public function __invoke(Request $request): Response
         $rootPages = $pageModel->findPublishedRootPages(['dns' => $request->server->get('HTTP_HOST')]);
 
         if (null === $rootPages) {
-            return new Response('', Response::HTTP_NOT_FOUND);
+            // We did not find root pages by matching host name, let's fetch those that do not have any domain configured
+            $rootPages = $pageModel->findPublishedRootPages(['dns' => '']);
```
Toflar, line 45: FYI: At first, I wanted to combine it in one query but that wouldn't work correctly. If you had one root page with a domain configured and one not, the one with the domain configured would then also include the ones from the one without domain configured which is not what we want.

fritzmg (dismissed):

aschempp (dismissed):
Not sure if filtering the domain later would be more efficient than two database calls, but approve because this certainly works.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #2704, opened 2021-01-29
Fix the rgxp=>httpurl implementation

### LINKED ISSUE #2629: validation of comment form website-field
Just stumbled about this:
The "website"-field in the comment forms accept **mailto:** urls.
I don't think that's desired but if so - the email-address should at least be encoded in the output.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2629
| Docs PR or issue | -

This PR fixes the rgxp=>httpurl implementation, which was missing a couple of checks. It also uses the new rgxp for all fields that expect an absolute URL (see #2629).

I am not a big fan of having class constants for 2 of 10 rgxp options, though. Either all or none. @contao/developers WDYT?


fritzmg:
I am in favour of constants for everything. Not sure where to put them though.

fritzmg (approved):

PR #2706, opened 2021-01-29
Add support for namespaced DC drivers

Toflar:
Requiring custom DC's to reside in the global namespace is still kind of PITA as you'll have to register that class differently from everything else in your bundle in your `composer.json` autoload section.

Of course, we'll have to come up with an alternative via DI container eventually. But let's at least support placing it in a proper namespace like everything else until then.

Usage:

```php
use Vendor\Driver\FoobarDriver;

$GLOBALS['TL_DCA']['tl_foobar']['config']['dataContainer'] = FoobarDriver::class;
```


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1388,6 +1388,24 @@ abstract public function getPalette();
 	 * @throws \Exception
 	 */
 	abstract protected function save($varValue);
+
+	/**
+	 * Return the class name for .
```
fritzmg:
```suggestion
	 * Return the class name of the DataContainer driver for the given table.
```

aschempp:
How about using a factory (`DataContainer::create`) instead?

ausi (dismissed):

Toflar:
Can't use a factory. The constructor params are not standardized across the data containers.

leofeyer:
This is very nice. If you fix the CI chain, I'll happily merge the PR. 👍🏻

bytehead (dismissed):

Toflar:
Done.

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @Toflar.

ausi:
Did we forget to backport this for Contao 4.9?

fritzmg:
I don't think it was "forgotten" but merely regarded as a new feature. But @ausi and me were discussing whether we should backport this simple change to the current LTS, since it truly is a PITA to actually get a driver to be registered otherwise, if you want to follow modern programming principles. @contao/developers wdyt?

leofeyer:
Ok with me.

PR #2716, opened 2021-02-08
Urldecode file paths from metadata URLs in the FigureBuilder

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Followup to #2674 for 4.10 and later

~~Imo it's safe to always urldecode the output of `Metadata#getUrl()`. This way the `is_file()` check in `FigureBuilder#onDefineLightboxResult()` will succeed and we do not need to handle the case separately.~~

This PR urldecodes file paths taken from `Metadata#getUrl()` inside the `FigureBuilder`.


ausi:
> Imo it's safe to always urldecode the output of `Metadata#getUrl()`.

Why? What if someone links to external urls like `https://www.google.com/search?q=foo%26bar`?

m-vo:
> Why? What if someone links to external urls like https://www.google.com/search?q=foo%26bar?

Yeah, silly me. Force pushed into oblivion. It's changed now.

fritzmg (approved):

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2717, opened 2021-02-09
Use the chained router to find root pages

### LINKED ISSUE #2499: Show error page (404) of fallback language, for not defined language.
When your installation is a multilingual and uses a `prepend_locale: true`, then langauge is added to the url. And also say you have defined a page not found for all these roots (let's say, we have roots for `en` and `de`).

Now, if someone access any non available resource on these roots, for example  `mydomain.com/de/junklink` or `mydomain.com/en/junklink`, where `junklink`is non-existing and fictive. Contao shows correct 404 page from respective roots.

But if someone access a page with not defined root for a language, for example `mydomain.com/jp/junklink`. Then Contao shows standard error message `Oops! An Error Occurred`. Here it would be nice to show the error page from the fallback root.

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2499

I had this issue in a client application, debugging it showed that the fallback 404 was actually found correctly, but `Frontend::getRootPageFromUrl` used by  `PageError404` did not resolve correctly. Most likely because the error404 pages are not present in the regular matcher but only in the chained router/matcher.


Toflar (dismissed):

aschempp:
The failing tests look like an issue in the GitHub Actions?

m-vo:
> The failing tests look like an issue in the GitHub Actions?

Yes, looks similar in #2716. Except for the 7.4 + 8.0 test.

leofeyer:
I have fixed the unrelated failures. The remaining failures are related to the PR, I'm afraid. 🙈

aschempp:
I have updated the implementation. The failing test showed why using the router is not ideal:
 1. it will use _all_ routers, e.g. also local and custom ones. That means it _might_ match a root page that does not come from Contao. While that is correct, it is not what the method is looking for, so using the matcher is better.
 2. **because** it uses the router, it also runs the route enhancers, which modify the `$_GET` parameters based on matched results. We certainly don't want that here though.

I also added a test for the current issue. This lead me to notice we only check a 404 error is thrown (the Symfony one) and not actually rendering the Contao 404 case. This is now all resolved.

core-bundle/src/Resources/contao/classes/Frontend.php
```diff
@@ -332,24 +332,32 @@ public static function getRootPageFromUrl()
...
+				$logger->log(LogLevel::ERROR, $strError, array('contao' => new ContaoContext(__METHOD__, 'ERROR')));
+
+				throw new NoRootPageFoundException('No root page found', 0, $exception);
+			}
+		}
+
+		$objRootPage = $arrParameters['pageModel'] ?? null;
 
-			throw new NoRootPageFoundException('No root page found', 0, $exception);
+		if (!$objRootPage instanceof PageModel)
```
leofeyer, line 358: Why not `if (!isset($arrParameters['pageModel'])) {`? Then we would not need the `?? null` part. And the error message would be more accurate, because now `$objRootPage` could be a non-PageModel object and the error would still be "must have a pageModel parameter".

aschempp: I think mostly because that tells my IDE that `$objRootPage` is a `PageModel` later in the code 🤔

aschempp: also, this has been the code before, that didn't change (only rearranged) 😉

PR #2724, opened 2021-02-11
Support extracting the canonical URI from a document

Toflar:
I think once we start working on e.g. a new search implementation, this might come in handy.

/cc @ausi


core-bundle/src/Search/Document.php
```diff
@@ -90,6 +90,30 @@ public function getContentCrawler(): Crawler
         return $this->crawler;
     }
 
+    public function extractCanonicalUri(): ?UriInterface
+    {
+        foreach ($this->getHeaders() as $key => $values) {
+            if ('link' === $key) {
+                foreach ($values as $value) {
+                    if (preg_match('@<(https?://(.+))>; rel="canonical"@', $value, $matches)) {
```
ausi:
```suggestion
                    if (preg_match('@<(https?://(.+))>;\s*rel="canonical"@', $value, $matches)) {
```

as far as I can see from the spec, the space before `rel=` seems to be optional? Not sure if spaces between around `=` are allowed too? `rel = "canonical"`

ausi (commented):
`rel="canonical"` links don’t have to be absolute, so if we want a robust implementation we should probably resolve relative URLs to absolute ones?

Toflar:
> `rel="canonical"` links don’t have to be absolute, so if we want a robust implementation we should probably resolve relative URLs to absolute ones?

According to Google, they have to be: https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-link-method

I think that's good practice. And if you don't, then it's just not gonna work - no big deal.

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #2731, opened 2021-02-11
Do not replace template data recursively when applying legacy template data

### LINKED ISSUE #2730: array_replace_recursive in applyLegacyTemplateData
**Affected version(s)**

Contao 4.10

**Description**

I think we should check if the recursive part is really needed here https://github.com/contao/contao/blob/d381a110efbc704e9793e229e98a9c0dffd16717/core-bundle/src/Image/Studio/Figure.php#L360

/cc @m-vo

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2730
| Docs PR or issue | -


ausi (approved):

leofeyer:
Thank you @m-vo.

Issue #2709, opened 2021-02-01
Image studio improvements


m-vo:
**Affected version(s)**

Contao >=4.10

**Description**

I've been intensively using the image studio stuff in some projects now and while I'm happy with the overall solutions there are a few things I'd like to adjust.

### Dealing with invalid resources

When outputting images that were selected in the BE it quite often happens, that a resource is or becomes invalid. E.g. because  incomplete/invalid datasets were stored or if a file was deleted later on. We're throwing a custom `InvalidResourceException` in the `FigureBuilder` in these cases (which is a nice thing in general).

1. IMHO we should adjust the `FigureRenderer` to tolerate these exceptions when running in `prod`: an empty string would be returned in this case (and maybe a log entry made). As this feature is only used in template/insert tag context, I'd consider this a bugfix/no BC break because you would (in general) not have a way to handle an exception there.

2. Dealing with potentially invalid resources in the `FigureBuilder` leads to a rather ugly pattern that breaks the otherwise nice fluent interface. This happens if you're fine with assigning a `Figure` or `null` to the template's context and letting the template handle the `null` case.  Here is an example from a typical scenario in a content element:
    ```php
    function getFigure($resource): ?Figure
    {
        $figureBuilder = $this->studio
            ->createFigureBuilder()
            ->setSize('_my_size')
        ;
    
        try {
            $figureBuilder->from($resource);
        } catch (InvalidResourceException $e) {
            return null;
        }
    
        return $figureBuilder->build;
    }
    ```

    IMHO we should *optionally* introduce the possibility to build figures from invalid resources. In this case calling `build()` would return a `Figure` or `null` and no `InvalidResourceException` exception would be thrown when defining the resource. We could either add an argument to the factory method…

    ```php
    function getFigure($resource): ?Figure
    {
        return $this->studio
            ->createFigureBuilder(allowInvalidResources: true)
            ->setSize('_my_size')
            ->from($resource)
            ->build()
        ;
    }
    ```
    … or add a `FailTolerantFigureBuilder`. 
    ```php
    function getFigure($resource): ?Figure
    {
        return $this->studio
            ->createFailTolerantFigureBuilder()
            ->setSize('_my_size')
            ->from($resource)
            ->build()
        ;
    }
    ```
    (The latter would be more effort to implement but would allow the original one to keep its strong type return type hint for the `build` function.)

### Figure in legacy template data
[...]

PR #2734, opened 2021-02-11
Enable accessing Figure in all templates

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Partly solves #2709
| Docs PR or issue | todo

Addresses https://github.com/contao/contao/issues/2709#issuecomment-777626885 (last bit).

With this PR the `Figure` can be accessed in all custom templates. I added a reference to `applyLegacyTemplateData` and also manually added references where `getLegacyTemplateData` was used directly in our content elements/modules.

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentGallery.php
+++ b/core-bundle/src/Resources/contao/elements/ContentGallery.php
@@ -263,7 +263,7 @@ class ContentGallery extends ContentElement
 						->build();
 
 					$cellData = $figure->getLegacyTemplateData($this->imagemargin);
-					$cellData['figure'] = $this;
+					$cellData['figure'] = $figure;
 				}
 				else
 				{
--- a/core-bundle/src/Resources/contao/elements/ContentGallery.php
+++ b/core-bundle/src/Resources/contao/elements/ContentGallery.php
@@ -258,10 +258,12 @@ class ContentGallery extends ContentElement
 				// Image / empty cell
 				if (($j + $i) < $limit && null !== ($image = $images[$i + $j] ?? null))
 				{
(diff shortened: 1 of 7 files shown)


ausi (dismissed):

m-vo:
This one is rebased + imo good to go as well.

ausi (dismissed):

core-bundle/src/Resources/contao/elements/ContentGallery.php
```diff
@@ -258,10 +258,12 @@ protected function compile()
...
 				{
-					$cellData = $figureBuilder
+					$figure = $figureBuilder
 						->fromFilesModel($image)
-						->build()
-						->getLegacyTemplateData($this->imagemargin);
+						->build();
+
+					$cellData = $figure->getLegacyTemplateData($this->imagemargin);
+					$cellData['figure'] = $this;
```
leofeyer, line 266:
```suggestion
					$cellData['figure'] = $figure;
```

Shouldn't this be `$figure` instead of `$this`?

leofeyer:
```suggestion
					$cellData['figure'] = $figure;
```

Shouldn't this be `$figure` instead of `$this`?

leofeyer (changes requested):

core-bundle/src/Resources/contao/elements/ContentGallery.php
m-vo: Good catch!

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #2735, opened 2021-02-11
Handle non-regex values in the search filter

### LINKED ISSUE #2498: Backend filter does not work when using special chars - brackets
**Affected version(s)**

Contao 4.10.4

**Description**
Filter result is empty when using special chars (brackets) in backendfilter "title"

My event titles are named as follows: **MyDate (teacher's abbreviation/exam quarter)** so **terminxyz (MM/Q4)**

If I only search for "**terminxyz**" the filter is correct. As soon as I take the bracket "**terminxyz (MM**" (abbreviation for Max Muster) the result is zero.

Searching "**MM/Q4**" its working. Brackets seem to be the problem here.

greetings
Frank

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/2498#issuecomment-777595404


fritzmg (dismissed):
🦊

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -5200,11 +5186,50 @@ protected function searchMenu()
...
+			$searchValue = $session['search'][$this->strTable]['value'];
+
+			try
+			{
+				$this->Database->prepare("SELECT '' REGEXP ?")->execute($searchValue);
+			}
+			catch (DriverException $exception)
+			{
+				// Quote search string if it is not a valid regular expression
+				$searchValue = preg_quote($session['search'][$this->strTable]['value']);
```
```suggestion
				$searchValue = preg_quote($searchValue);
```

Toflar (dismissed):
🦊

fritzmg (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -5200,11 +5186,50 @@ protected function searchMenu()
...
+
+			$strPattern = "$strReplacePrefix CAST(%s AS CHAR) $strReplaceSuffix REGEXP ?";
 
 			if (substr(Config::get('dbCollation'), -3) == '_ci')
 			{
-				$strPattern = "LOWER(CAST(%s AS CHAR)) REGEXP LOWER(?)";
+				$strPattern = "$strReplacePrefix LOWER(CAST(%s AS CHAR)) $strReplaceSuffix REGEXP LOWER(?)";
 			}
 
 			$fld = $session['search'][$this->strTable]['field'];
```
leofeyer: Can we move this up to line 5190 and replace `$GLOBALS['TL_DCA'][$this->strTable]['fields'][$session['search'][$this->strTable]['field']]['eval']['decodeEntities']` with `$GLOBALS['TL_DCA'][$this->strTable]['fields'][$fld]['eval']['decodeEntities']` in line 5205?

ausi: Done in 692f47f8bd04318446bfd1526ec676748ba28663

Toflar:
Ah, don't we have the same issue in `DC_Folder`.

leofeyer:
Oh yes, we do. And in `PageSelector` and `FileSelector`.

aschempp:
The `REPLACE` is what I suggested, but didn't you argue this would break any database indexes?

ausi:
> The `REPLACE` is what I suggested, but didn't you argue this would break any database indexes?

No, I answered that `REGEXP` doesn’t use the index anyway (unless you are matching from the start of the string, which we do not in this case).

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -5200,27 +5186,65 @@ protected function searchMenu()
...
+				{
+					$strReplaceSuffix .= ", '$strSource', '$strTarget')";
+				}
+			}
+
+			$strPattern = "$strReplacePrefix CAST(%s AS CHAR) $strReplaceSuffix REGEXP ?";
+
+			if (substr(Config::get('dbCollation'), -3) == '_ci')
+			{
+				$strPattern = "$strReplacePrefix LOWER(CAST(%s AS CHAR)) $strReplaceSuffix REGEXP LOWER(?)";
```
aschempp, line 5233: I understand that `REPLACE` is case-sensitive according to the docs, but the replaces characters are never upper/lowercase, so what's the `LOWER` for?

ausi: The `LOWER()` is for the rest of the string in order to compare it case insensitive. This is exactly as before, I didn’t change the `LOWER(CAST(%s AS CHAR))` part.

aschempp: But only if the collation matched? https://github.com/contao/contao/pull/2735/files#diff-e4bf1535e70e3fbc2b9692ba8add4c45237305ff815df14c7ba9c9ce53dc9204L5205

aschempp: Only if the collation was `case insensitive`. Not sure why that would need `LOWER()` at all 😅

ausi: I would guess that older MySQL versions ignored the collation for `REGEXP` comparisons. But that is not part of this PR IMO.

aschempp: Agreed, but you're now changing the behavior?

ausi: I don’t understand what you mean, this code is exactly the same as it was before (except the replace prefix and suffix): https://github.com/contao/contao/blob/ffebcd7fa47ce6576d311cf73c2a9eda49ae7113/core-bundle/src/Resources/contao/drivers/DC_Table.php#L5203-L5208

aschempp: Me neither, looking at it now 😂 Sorry for some reason I was blind to see line 5205

aschempp (dismissed):

ausi:
> Ah, don't we have the same issue in `DC_Folder`.

> Oh yes, we do. And in `PageSelector` and `FileSelector`.

Done. This should be RTM now I think.

Toflar (approved):

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #2737, opened 2021-02-12
Use the internal page title for the search index

### LINKED ISSUE #2487: Search results show page title + root page title
**Affected version(s)**

Contao 4.9

**Description**

Since Contao 4.9, the search results include the website root title in the title of the result (if configured like that in the page layout, which is the default). After upgrading to Contao 4.9 our customers are complaining that this leads to a lot of superfluous information in the search results.

Before Contao 4.9 the search results might have looked like this:

* Lorem ipsum dolor
* Consectetuer adipiscing elitr
* Nullam quis
* Maecenas tempus
* …

i.e. just the pages' titles, which is expected. Now they look like this:

* Lorem ipsum dolor - Vendor Website Title
* Consectetuer adipiscing elitr - Vendor Website Title
* Nullam quis - Vendor Website Title
* Maecenas tempus - Vendor Website Title
* …

The " - Vendor Website Title" will always be the same for each page, unless you configured the search module to include more than one website root of course (which is not the default).

This happens due to the switch to the crawler, which simply takes an URL's title from the meta data - which in turn is generated (by default) through the insert tag `{{page::pageTitle}} - {{page::rootPageTitle}}`.

I would propose to additionally add `contao:pageTitle` to the JSON-LD, so that this can be stored in the `tl_search` table as before.

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2487
| Docs PR or issue | -

### DIFF
--- a/core-bundle/tests/Search/Indexer/DefaultIndexerTest.php
+++ b/core-bundle/tests/Search/Indexer/DefaultIndexerTest.php
@@ -125,6 +125,21 @@ class DefaultIndexerTest extends ContaoTestCase
             null,
             true,
         ];
+
+        yield 'Test valid index with page title' => [
+            new Document(new Uri('https://example.com/valid'), 200, [], '<html lang="de"><head><title>HTML page title</title></head><body><script type="application/ld+json">{"@context":{"contao":"https:\/\/schema.contao.org\/"},"@type":"contao:Page","contao:title":"JSON-LD page title","contao:pageId":2,"contao:noSearch":false,"contao:protected":true,"contao:groups":[42],"contao:fePreview":false}</script></body></html>'),
+            [
+                'url' => 'https://example.com/valid',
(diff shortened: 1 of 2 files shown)


Toflar:
We must not forget to update the schema, once this is released :)

core-bundle/tests/Search/Indexer/DefaultIndexerTest.php
```diff
@@ -133,7 +148,7 @@ public function indexProvider(): \Generator
      *
      * @expectedDeprecation Using the JSON-LD type "RegularPage" has been deprecated and will no longer work in Contao 5.0. Use "Page" instead.
      */
-    public function testIndexesADocumentWithDeprecatedJsonLd(Document $document, ?array $expectedIndexParams, string $expectedMessage = null, bool $indexProtected = false): void
+    public function IndexesADocumentWithDeprecatedJsonLd(Document $document, ?array $expectedIndexParams, string $expectedMessage = null, bool $indexProtected = false): void
```
ausi: Why did you remove the `test` prefix?

ausi (approved):

leofeyer:
> We must not forget to update the schema, once this is released :)

See https://github.com/contao/schema.contao.org/pull/3. Can we merge this in advance?

ausi:
> Can we merge this in advance?

IMO yes.

leofeyer:
Thank you @Toflar.

PR #2743, opened 2021-02-12
Change the default URL suffix

leofeyer:
As discussed in yesterday's Mumble call, this PR changes the default URL suffix from `'.html'` to `''`.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -471,7 +471,7 @@ $GLOBALS['TL_DCA']['tl_page'] = array
 			'label'                   => &$GLOBALS['TL_LANG']['tl_page']['urlSuffix'],
 			'inputType'               => 'text',
 			'eval'                    => array('nospace'=>'true', 'maxlength'=>16, 'tl_class'=>'w50'),
-			'sql'                     => "varchar(16) NOT NULL default '.html'"
+			'sql'                     => "varchar(16) NOT NULL default ''"
 		),
 		'useSSL' => array
 		(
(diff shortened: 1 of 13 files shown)


fritzmg (approved):

asaage:
🤨 realy ?
Well - i still could not convince plesk+nginxonly v-host setups to work with empty URL-suffixes.
I mean i could - but it requires server-admin-status and cannot be done by a subscription-owner.
Any quick install on such systems will then require some sort of troubleshooting.

ausi:
@asaage How do you login to the backend on such systems?

asaage:
>  How do you login to the backend on such systems?

I have to reevaluate that - unfortunately that will take 2-3weeks.
The only thing i remember is that i had all sorts of problems with empty suffix - sometimes the 404PageType didn't work correctly sometimes folderUrls and/or the deferred ImageResize did't work - i think i switched from nginx to apache in some cases.

leofeyer:
No one forces you to use an empty URL suffix. Just enter `.html` into the field like you do now and you will be fine.

asaage:
> Just enter ```.html``` into the field

Im fine with that - it was just pretty hard to figure out why things did not work as expected in this case - especially if you don't know that the suffix was changed.

ausi:
> especially if you don't know that the suffix was changed.

This pull request only changes the default, existing installations will stay the same.

> The only thing i remember is that i had all sorts of problems with empty suffix

If there are problems with an empty URL suffix, we should check if we could fix them. But as `/contao/login` has no URL suffix for a very long time now, I guess if these problems would occur frequently we would already have got more feedback on that.

> and/or the deferred ImageResize did't work

That’s most likely due to image suffixes like `.jpg` are handled differently, should not be related to empty suffix.

PR #2748, opened 2021-02-14
Refactored backend controllers and backend module fragments

richardhj:
| Q                | A
| -----------------| ---
| Fixed issues     | N/A
| Docs PR or issue | tba.

This PR reworks the backend controller and fragment handling.

More precisely,
- The `BackendController` was moved into its own namespace. This allows making use of the route `prefix` option. (1c41e34)
- The old `BackendIndex` and `BackendMain` controllers have been refactored (b40abd2)
- Developers can add backend modules by tagging a fragment controller with the `contao.backend_module` tag (09f4810)
- `Backend::getBackendModule()` is deprecated and has been moved to its own controller (bc18aa7)
- The Welcome Screen was refactored and becomes capabilities to be extended by additional fragments, so-called dashboard widgets (e4a6d08).

New features
-------------

### Register backend module fragment

Since contao/core-bundle#512 it is possible, to use custom backend routes. However, this approach has some downsides in practice:
1. Your controller is responsible for the entire backend page, even if you want to add just the backend module section.
2. You need to add the route to the navigation manually.
3. Custom routes do not support defining permissions.

The new approach is to just provide the backend module as a fragment, as we already know for frontend modules.

```php
/**
 * @\Contao\CoreBundle\ServiceAnnotation\BackendModule("custom", category="content", hideInNavigation=false, disablePermissionChecks=false)
 */
class CustomBackendModuleController extends \Contao\CoreBundle\Controller\BackendModule\AbstractBackendModuleController
{
    {
        return new Response('This section will be available under /contao/?do=custom and is automatically displayed in the navigation');
    }
}
```

The new backend module will be automatically added to `BE_MOD`, so it is automatically added to the backend navigation and the user permissions. You can define the category (as in `$GLOBALS['BE_MOD']['category']`) as a service tag argument. Further, you can define some options, i.e. `disablePermissionChecks` and `hideInNavigation`, as service tag attributes that will be passed to the `$GLOBALS['BE_MOD']` array.

### Add widgets to the dashboard

With the same approach as above, it is possible to add new widgets/fragments to the dashboard. Often, extension developers used to use the `getSystemMessages` hook to show information to the user on the home screen. This always has been a hack.

```php
/**
 * @DashboardWidget()
 */
class CustomWidgetController extends AbstractDashboardWidgetController
{
    {
        return new Response('This section will be added on the backend dashboard.');
    }
}
```

This is a screenshot of the new feature, with two added fragments under the welcome screen. It's just a quick mockup, using the non-default backend theme. I hope, you get the idea. 

Screenshot 2021-02-14 at 01 28 18

Each of those widgets can be a turbo frame, meaning that the widget can be fully interactive, include input masks and many more: https://github.com/symfony/ux/blob/f7c6e637d36ae94d4f9f735a0d20009af5bd67bd/src/Turbo/README.md#2-create-interactive-components-with-turbo-frames 🚀 

### Change route prefix

You can change the prefix of the backend routes (f5daea4).

```yml
contao:
  backend:
    route_prefix: '/gondao'
```

The backend will then be available under `/gondao/login` instead of `/contao/login`.

-------

This PR is a POC and RFC.
This PR only includes the bare minimum technical changes. The potential of this PR will evolve with a new Backend theme. We must remove the white background of the container, so that custom fragments can add their own stylings on the page. This will be addressed in #2697.


aschempp:
If I'm understanding correctly, is there a reason this PR contains three (totally) separate things (renaming backend route, improve welcome screen, support backend module fragments)?

richardhj:
I can split this PR into two separate PRs. However, some parts are depending on each other (and GitHub only supports depending PRs for local branches, not forks).
Will you comment on this PR anyway? This is a POC (surely some things missing).

Issue #2686, opened 2021-01-25
Main heading in the backend is cut off at the bottom


rabauss:
**Affected version(s)**

Contao 4.9+

**Description**

Unfortunately, since https://github.com/contao/contao/pull/2655 the line height is to low with overflow and the main heading in the backend is cut off at the bottom - see:
![image](https://user-images.githubusercontent.com/14016098/105688676-2d503c00-5efa-11eb-9f9a-8d376623bbaa.png)

leofeyer:
Can you please check if the following CSS fixes the issue for you:

```css
#main_headline {
    margin: 18px 0;
}
#main_headline span {
    line-height: 20px;
}
```

rabauss:
Not completely in the demo.contao.org it looks like this:

![image](https://user-images.githubusercontent.com/14016098/105696534-c2a3fe00-5f03-11eb-935c-bcef8bca00a7.png)

To reach nearly the state before 4.9.11 I think it should be:
```css
#main_headline {
    margin: 16px 0;
}
#main_headline span {
    line-height: 24px;
}
```

bohnmedia:
The seperator is too low now.

![image](https://user-images.githubusercontent.com/9419148/108196425-648cc400-7119-11eb-87ad-492bdffe76e3.png)

You should remove line-height from "#main_headline span", give it "vertical-align:top" and assign the line height to "#main_headline".

```stylesheet
#main_headline {
    margin: 18px 0;
    padding: 0 16px;
    font-size: 1.1rem;
    line-height: 22px;
}

#main_headline span {
    display: inline-block;
    max-width: 30%;
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
    vertical-align: top;
}
```

![image](https://user-images.githubusercontent.com/9419148/108196862-e67ced00-7119-11eb-8678-c04d6c6186de.png)

bohnmedia:
The misplaced seperator is in the last commit. https://github.com/contao/contao/commit/2d3f99ebae4755ce743cf799643498ea24bfbf66

asaage:
https://github.com/contao/contao/pull/2751#issuecomment-779337265

leofeyer:
@bohnmedia Can you please open a new issue regarding this?

Also, which OS and browser are you using? Because I cannot reproduce the issue on my system (Safari 14.0.3 on macOS 11.2.1).

[...]

PR #2655, opened 2021-01-20
Improve rendering long titles and file names

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2575
| Docs PR or issue | -

PR #2751, opened 2021-02-15
Fixed cutting off the descenders in backend breadcrumb

ghost:
I changed the main.css of the flexible theme to avoid cutting off the descenders in the backend breadcrumb.

Before
image

After
image


bohnmedia:
Issue: #2686

asaage:
That's not just headings and breadcrumbs.
It's an annoyance throughout the whole BE like here: Sitestructure
image
I guess that was introduced here: https://github.com/contao/contao/pull/2655 and i don't like it at all.

leofeyer:
See #2875.

PR #2752, opened 2021-02-15
Symlink highlight.php as highlight_php

leofeyer:
This will save us a lot of headaches with servers interpreting `https://domain.tld/vendor/scrivo/highlight.php/styles/foundation.css` as request for `vendor/scrivo/highlight.php`. 🙈


ausi (approved):

fritzmg (approved):

richardhj:
Don't we need to symlink the old path as well? (I know, there is no BC promise on templates…)
Nvm, seems to be a new feature :D

PR #2755, opened 2021-02-16
Remove an unnecessary loadLanguageFile() call

leofeyer:
Follow-up on #2497

The label ist taken from `$GLOBALS['TL_LANG']['MSC']`, therefore we do not need to load the `tl_member` language files.


fritzmg (approved):

ausi:
> The label ist taken from `$GLOBALS['TL_LANG']['MSC']`

But this could be overwritten via a custom DCA file to something from `$GLOBALS['TL_LANG']['tl_member']`

leofeyer:
That did not work before, either:

https://github.com/contao/contao/blob/2ab244e7362b463b8b2caa15953934a4a67d8475/core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php#L70

PR #2762, opened 2021-02-16
Fix loading database.sql files in the DcaExtractor

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

The `DcaExtractor` uses this check (that is run on each field of a DCA) to check if a `database.sql` file should be parsed:

https://github.com/contao/contao/blob/712de2476dc6100cad1876da1f22b809f7558d5c/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L400-L404

However, this check is always true in case of an entity backed DCA or if there are any fields without a SQL definition like explanations or fields that are populated/stored by a `load/save_callback`.  We're then falsely triggering a deprecation notice (L439) and an invalid array access occurs (L462).

This PR always tries to locate the `database.sql` files and only continues processing if there are actually any.


fritzmg (dismissed):

aschempp (commented):
This means an (expensive) File IO search is done for every DCA, even though that might not be necessary. Shouldn't we rather adjust the error message to only appear if there actually are any files (probably around (old) line 453?

m-vo:
> This means an (expensive) File IO search is done for every DCA, even though that might not be necessary.

This should only happen during cache warmup, should it?

We could cache the output of `System::getContainer()->get('contao.resource_locator')->locate('config/database.sql', null, false);` in a static variable as it should not change across calls. This is probably OK because there is already static state in the object.

But IMHO we need to look at least once. Or else what would be the criteria to do so?

m-vo:
> We could cache the output of System::getContainer()->get('contao.resource_locator')->locate('config/database.sql', null, false); in a static variable as it should not change across calls.

Done so in 0651990b36b5351b99c75160c1ee959c72f08c8e

fritzmg (dismissed):

m-vo:
The failing test seems unrelated.

aschempp:
I still don't see why we need to change the behaviour? If the DCA has all the SQL needed, we don't need to look for SQL files. Your case of using Doctrine entities means we _should_ look for SQL files if the DCA does not have SQL definitions, but **only show a warning** when an SQL file is actually there?

m-vo:
The logic currently in place (without this PR) is that `database.sql` files are expected to exist if *at least one* field does not have a SQL definition. IMHO this means for BC: `database.sql` files were always allowed as a complimentary source.

If we, however, say that you were never intended to mix these resources, we could alter the logic to: If at least one SQL definition is there - do not expect any `.database.sql` files. If we would want to include entity backed DCAs we would also need to check for class metadata. 

So what do we want?

m-vo:
> If the DCA has all the SQL needed, we don't need to look for SQL files. 

btw.: To me, this is only one possible case. Even the core works around it in the serp field by setting `sql` to `null`.

aschempp:
> If we, however, say that you were never intended to mix these resources, we could alter the logic to: If at least one SQL definition is there - do not expect any .database.sql files.

That's a wrong assumption, because DCAs can be extended by SQL files, or they can be extended with something that includes SQL (but the origin DCA of the table uses SQL files).

m-vo:
> That's a wrong assumption, because DCAs can be extended by SQL files, or they can be extended with something that includes SQL (but the origin DCA of the table uses SQL files).

The PR isn't following this logic. I thought that was what you were suggesting in order to reduce the needed file access. Sorry, if I got you wrong.

We could of course swallow the exception/readd the condition if you prefer. To me this logic is flawed and only holds true for some cases or allows weird behavior (i.e. why should you not be able to use `database.sql` files as soon as you extend your DCA with an explanation field).

leofeyer:
@contao/developers /cc

fritzmg:
This is also a problem in Contao 4.11 under PHP 8. If you have a DCA field without an `sql` definition, it will lead to `Warning: Undefined array key` [here](https://github.com/contao/contao/blob/1873bcc9ea728a08ca8755f6e7251d999d7b316b/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L462).

m-vo:
Indeed, that's the very reason I opened up this PR. :upside_down_face:

bytehead:
> This is also a problem in Contao 4.11 under PHP 8. If you have a DCA field without an `sql` definition, it will lead to `Warning: Undefined array key` [here](https://github.com/contao/contao/blob/1873bcc9ea728a08ca8755f6e7251d999d7b316b/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L462).

Good catch! I do have fields without an `sql` definition when using entities 😎

aschempp:
> @contao/developers /cc

I guess we need to decide whether to change the behaviour/logic or just fix the bug/invalid deprecation (in 4.9!)?
👍 for behaviour change
👎 to fix deprecation only

leofeyer:
As discussed in Mumble on March 11, we want to keep the current behavior as much as possible but prevent the deprecation warning if there are no database.sql files.

core-bundle/tests/Cache/ContaoCacheWarmerTest.php
```diff
@@ -46,8 +47,16 @@ protected function tearDown(): void
 
     public function testCreatesTheCacheFolder(): void
     {
+        $resourceLocator = $this->createMock(FileLocator::class);
+        $resourceLocator
+            ->method('locate')
+            ->with('config/database.sql', null, false)
+            ->willThrowException(new FileLocatorFileNotFoundException())
```
leofeyer, line 54: Why not `->willReturn([])`?

m-vo: Because that's also what Symfony's file locator does if the file is missing. See https://github.com/symfony/symfony/blob/950e1444cb53780543beb4968e83f13679bd5983/src/Symfony/Component/Config/FileLocator.php#L43-L45

leofeyer (approved):

leofeyer:
Thank you @m-vo.

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -592,6 +591,25 @@ protected function createExtract()
...
+		if (null !== self::$arrDatabaseSqlFiles)
+		{
+			return self::$arrDatabaseSqlFiles;
+		}
+
+		try
+		{
+			$files = System::getContainer()->get('contao.resource_locator')->locate('config/database.sql', null, false);
+		}
+		catch (FileLocatorFileNotFoundException $e)
```
aschempp, line 606: this now catches a specific exception instead of `\InvalidArgumentException`, is that intentional?

m-vo: Yes, that's the one that Symfony file locator is throwing if a file does not exist.

PR #2763, opened 2021-02-16
Fix an illegal array access in BackendUser::navigation()

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     |  -
| Docs PR or issue | -

This PR hardens `BackendUser#navigation()` against  invalid array access. Normally I wouldn't care too much about those things in the legacy code base but this can break the back end under PHP8 when the `knp_menu_get()` Twig function gets executed.

To trigger the problem it's enough to have a back end module with a language definition like this:
```php
$GLOBALS['TL_LANG']['MOD']['my_group'] = ['Group']`; // array notation without description
```

### DIFF
--- a/core-bundle/src/Resources/contao/classes/BackendUser.php
+++ b/core-bundle/src/Resources/contao/classes/BackendUser.php
@@ -519,7 +519,7 @@ class BackendUser extends User
 			{
 				$arrModules[$strGroupName]['class'] = 'group-' . $strGroupName . ' node-expanded';
 				$arrModules[$strGroupName]['title'] = StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['collapseNode']);
-				$arrModules[$strGroupName]['label'] = ($label = \is_array($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null) ? $GLOBALS['TL_LANG']['MOD'][$strGroupName][0] ?? '' : ($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null)) ? $label : $strGroupName;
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/BackendUser.php
```diff
@@ -519,7 +519,7 @@ public function navigation($blnShowAll=false)
 			{
 				$arrModules[$strGroupName]['class'] = 'group-' . $strGroupName . ' node-expanded';
 				$arrModules[$strGroupName]['title'] = StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['collapseNode']);
-				$arrModules[$strGroupName]['label'] = ($label = \is_array($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null) ? $GLOBALS['TL_LANG']['MOD'][$strGroupName][0] : ($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null)) ? $label : $strGroupName;
+				$arrModules[$strGroupName]['label'] = ($label = \is_array($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null) ? $GLOBALS['TL_LANG']['MOD'][$strGroupName][0] ?? '' : ($GLOBALS['TL_LANG']['MOD'][$strGroupName] ?? null)) ? $label : $strGroupName;
```
leofeyer: Again, using `?? ''` instead of `?? null` is a behavior change. It is ok to use it if the variable is passed to a function that expects a string (e.g. StringUtil::specialchars()) but rather not in assignments like this.

leofeyer (dismissed):

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #1999, opened 2020-07-26
Add configuration options for the back end theme

rabauss:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #403 

This PR is an improved solution of https://github.com/contao/contao/pull/1604!

Some config params for the back end theme were added:

```yaml
#config/config.yml or config/parameters.yml
contao:
    theme:
        brand:
            color: '#f47c00' # brand color for header etc.
            whitelabel: false # hide contao logo in header + login form
        header:
            title: 'Demo' # title for the header
            invert: false # invert the text + icon color from white to black
        environment:
            title: 'production' # environment title for the badge - is also set as standardized css class to the body
            background_color: '#0F1C26' # background color for the environment badge
            text_color: '#fff' # text color for the environment badge
        main:
            background_color: '#eaeaec' # background color of the main container
```

image
image

With the following example config you can "fully" cutomize your backend as shown in the next screenshots:

```yaml
contao:
    theme:
        brand:
            color: '#FFED00'
            whitelabel: true
        header:
            title: 'Tüdelü'
            invert: true
        environment:
            title: 'develop'
            background_color: '#fff'
            text_color: '#cc0000'
        main:
            background_color: '#F4CCCC'
```

image
image

I kept the previous color settings without css variables as fallback for IE11 ;-)

PR #2766, opened 2021-02-17
Automatically prefix the back end attributes

leofeyer:
Follow-up on #1999

### Currently

```yaml
contao:
    backend:
        attributes:
            data-app-name: Foobar
            data-app-version: 1.2.3
            style: 'background: #f00'
```

Will be converted to `data-app-name="Foobar" data-app-version="1.2.3" style="background: #f00"`.

### New

```yaml
contao:
    backend:
        attributes:
            app-name: Foobar
            app-version: 1.2.3
            style: 'background: #f00'
```

Will be converted to `data-app-name="Foobar" data-app-version="1.2.3" data-style="background: #f00"`.

So by automatically prefixing the attributes with `data-`, we can not only simplify the configuration but also prevent users from accidentally overwriting attributes such as `style`, `class` or `id`.

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -456,7 +456,7 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->arrayNode('attributes')
                     ->info('Adds HTML attributes to the <body> tag in the back end.')
-                    ->example(['data-app-name' => 'My App', 'data-app-version' => '1.2.3'])
+                    ->example(['app-name' => 'My App', 'app-version' => '1.2.3'])
                     ->validate()
                     ->always(
                         static function (array $attributes): array {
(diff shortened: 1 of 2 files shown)


Toflar (approved):

m-vo:
Imo that should have gone into the template so that you can overwrite this logic if you use the attributes at other places than HTML attributes. 🤔

 I guess we can still implement this later should we move to an array object for the attributes but maybe the API would become more complicated.

m-vo:
Thought about this again: This will make the attributes used in the back end main template incompatible with the other usages like in `ce_player` or `mod_randomImage` where there is a `<?= $this->attributes ?>` in the template.

We either won't be able to use the same logic everywhere or there is no way to use the attributes without prefix. Also: why shouldn't it be possible to set a class?

leofeyer:
Because that is what we have agreed to in the Mumble call (see https://github.com/contao/contao/pull/1999#issuecomment-730438526).

m-vo:
Got you, sorry. I really thought the conclusion of the call was something else. :see_no_evil:

PR #2777, opened 2021-02-19
Ignore empty authorization headers in the MakeResponsePrivateListener

ausi:
After debugging a non-working cache on a Contao website I noticed that I always got the following headers in the response:

```
contao-cache: miss
contao-private-response-reason: authorization
```

After debugging I found out that `$_SERVER['HTTP_AUTHORIZATION']` is set to an empty string. So I guess we should ignore empty authorization headers in the `MakeResponsePrivateListener`.

### DIFF
--- a/core-bundle/src/EventListener/MakeResponsePrivateListener.php
+++ b/core-bundle/src/EventListener/MakeResponsePrivateListener.php
@@ -39,7 +39,7 @@ class MakeResponsePrivateListener
      * Make sure that the current response becomes a private response if any
      * of the following conditions are true.
      *
-     *   1. An Authorization header is present
+     *   1. An Authorization header is present and not empty
      *   2. The session was started
      *   3. The response sets a cookie (same reason as 2 but for other cookies than the session cookie)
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

fritzmg (dismissed):

aschempp (dismissed):

core-bundle/src/EventListener/MakeResponsePrivateListener.php
```diff
@@ -64,8 +64,8 @@ public function __invoke(ResponseEvent $event): void
             return;
         }
 
-        // 1) An Authorization header is present
-        if ($request->headers->has('Authorization')) {
+        // 1) An Authorization header is present and not empty
+        if ($request->headers->has('Authorization') && '' !== $request->headers->get('Authorization')) {
```
ausi:
```suggestion
        if ('' !== (string) $request->headers->get('Authorization')) {
```

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #2307, opened 2020-09-17
Change condition from empty to "empty" string

zonky2:
Change condition from empty to "empty" string - now it´s possible to set 0 as default value

C 4.9

PR #2779, opened 2021-02-19
Allow 0 as default value for range sliders

### LINKED ISSUE #2778: Range slider cannot bet set to a value of zero
We just noticed that it is not possible to give an range slider (`FormRange`) a value of zero.
The cause of this is a simple mistake in the `form_range.html5` where the according if-statement is a bit to lax.

```php
<?php if ($this->value): ?> value="<?= (float) $this->value ?>"<?php endif; ?>
```

The better way in this case would be to use `isset` instead:

```php
<?php if (isset($this->value)): ?> value="<?= (float) $this->value ?>" <?php endif; ?>
```

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2778
| Docs PR or issue | -

In order to be able to set a **Default value** of `0` for a _Range slider_ form field in the form generator, we need to check for an empty string.

### DIFF
--- a/core-bundle/src/Resources/contao/templates/forms/form_range.html5
+++ b/core-bundle/src/Resources/contao/templates/forms/form_range.html5
@@ -1,4 +1,8 @@
-<?php $this->extend('form_row'); ?>
+<?php
+
+use Contao\FormRange;
+
+$this->extend('form_row'); ?>
 
 <?php $this->block('label'); ?>
   <?php if ($this->label): ?>
@@ -17,5 +21,5 @@
     <p class="error"><?= $this->getErrorAsString() ?></p>
   <?php endif; ?>
 
-  <input type="<?= $this->type ?>" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="range<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>"<?php if ($this->value != ''): ?> value="<?= (float) $this->value ?>"<?php endif; ?><?= $this->getAttributes() ?>>
+  <input type="<?= $this->type ?>" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="range<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>"<?php if ('' != $this->value): ?> value="<?= (float) $this->value ?>"<?php endif; ?><?= $this->getAttributes() ?>>
 <?php $this->endblock(); ?>
--- a/core-bundle/src/Resources/contao/templates/forms/form_range.html5
+++ b/core-bundle/src/Resources/contao/templates/forms/form_range.html5
@@ -1,8 +1,4 @@
-<?php
-
-use Contao\FormRange;
-
-$this->extend('form_row'); ?>
+<?php $this->extend('form_row'); ?>
 
 <?php $this->block('label'); ?>
   <?php if ($this->label): ?>
--- a/core-bundle/src/Resources/contao/templates/forms/form_range.html5
+++ b/core-bundle/src/Resources/contao/templates/forms/form_range.html5
@@ -17,5 +17,5 @@
     <p class="error"><?= $this->getErrorAsString() ?></p>
   <?php endif; ?>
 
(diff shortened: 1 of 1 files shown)


ausi:
Duplicate of #2307 ?

fritzmg:
> Duplicate of #2307 ?

True, but that PR has not seen any action since November last year. I can update that PR instead. I am using `'' != $this->value`, which covers `$this->value` being `''` or `null`.

core-bundle/src/Resources/contao/templates/forms/form_range.html5
```diff
@@ -1,4 +1,8 @@
-<?php $this->extend('form_row'); ?>
+<?php
+
+use Contao\FormRange;
```
ausi: Why do you need this `use` here?

fritzmg: That was an accident 🙃

ausi (dismissed):

bennyborn:
> I am using `'' != $this->value`, which covers `$this->value` being `''` or `null`.

Are you sure this works? It seems `''!=0` returns `false` on PHP 7.x and `true` on PHP 8.x
The proposed solution using `isset` works in both versions.

fritzmg:
> Are you sure this works?

Yes: https://3v4l.org/JCUJ4

bennyborn:
> > Are you sure this works?
> 
> Yes: https://3v4l.org/JCUJ4

Maybe I'm missing something but the original problem wasn't about an empty string but a numeric zero as a value https://3v4l.org/AreNO

fritzmg:
> The proposed solution using `isset` works in both versions.

`isset` would not work. If you do not set a default value in the back end (which is equivalent to setting it to an empty string), then `isset($this->value)` will be `true`.

bennyborn:
> `isset` would not work. If you do not set a default value in the back end (which is equivalent to setting it to an empty string), then `isset($this->value)` will be `true`.

True. But how to handle that? What about smth like `isset($this->value) && '' !== $this->value`?

fritzmg:
The current implementation in this PR does have one problem though: it does not work (in PHP <8) if you set `$widget->value = 0;`: https://3v4l.org/LtaYF

fritzmg:
I guess using `'' !== $this->value && null !== $this->value` is the only way to go: https://3v4l.org/f51b7

bennyborn:
> The current implementation in this PR does have one problem though: it does not work (in PHP <8) if you set `$widget->value = 0;`: https://3v4l.org/LtaYF

Thats what I was trying to tell you the whole time :D

fritzmg:
> Thats what I was trying to tell you the whole time :D

I only considered `'0'` since that's what would happen if you set it to `0` in the back end in the form generator. But not necessarily, when you instantiate the widget via PHP, yes.

m-vo:
`is_numeric($this->value)` would probably work as well…

ausi (dismissed):

core-bundle/src/Resources/contao/templates/forms/form_range.html5
```diff
@@ -17,5 +17,5 @@
     <p class="error"><?= $this->getErrorAsString() ?></p>
   <?php endif; ?>
 
-  <input type="<?= $this->type ?>" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="range<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>"<?php if ($this->value): ?> value="<?= (float) $this->value ?>"<?php endif; ?><?= $this->getAttributes() ?>>
+  <input type="<?= $this->type ?>" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="range<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>"<?php if ('' !== $this->value && null !== $this->value): ?> value="<?= (float) $this->value ?>"<?php endif; ?><?= $this->getAttributes() ?>>
```
leofeyer:
```suggestion
  <input type="<?= $this->type ?>" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="range<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>"<?php if ('' !== (string) $this->value): ?> value="<?= (float) $this->value ?>"<?php endif; ?><?= $this->getAttributes() ?>>
```

This is how we have solved it everywhere else. It will also handle `false`, which can never be a valid option.

leofeyer (changes requested):

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #2780, opened 2021-02-20
Handle session being null in System::getReferer

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Yet another array access problem notable with PHP8: in case `$session` is `null`, we should not try to access `$session['current']` / `$session['last']`…


m-vo:
Failing test seems unrelated.

ausi (dismissed):

leofeyer:
How do I reproduce this? Because `$session` should never be `null` at this point.

m-vo:
This happened when I had a BE page open in the browser and - after the session had expired for a while - I started working at it again. IIRC this error happened after the redirect from the login page.

leofeyer:
Does the issue occur in Contao 4.9 as well?

m-vo:
Likely, but it will only be noticed with PHP8 + default error reporting level.

leofeyer:
Then we should fix it in Contao 4.9.

m-vo:
It's rebased onto 4.9, now. 

I'm not sure what the failing tests are trying to tell me or if they are all unrelated.

ausi (dismissed):

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2783, opened 2021-02-21
Correctly merge image size _defaults

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Currently defining `formats` under the image size `_defaults` is broken. This is because the processed config values of the individual size definitions already contain an empty `formats` array which then wins in  `array_merge`. This PR fixes this problem by giving the values under `_defaults` precedence over empty arrays.

Unfortunately this case was missing from the tests and went unnoticed (added now)…


fritzmg (dismissed):

richardhj:
Why is this bugfix targeted 4.12? 😃

m-vo:
> Why is this bugfix targeted 4.12? smiley

It shouldn't. :smile:

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #2784, opened 2021-02-22
Correctly sort the pages if the URL suffix is empty

### LINKED ISSUE #2775: 404 error for Folder URLs without 'URL suffix'
**Affected version(s)**

Contao 4.11

**Description**

After updating to Contao 4.11, Folder URLs return 404 error [when URL's suffix is empty]. 

https://domain.com                      => OK
https://domain.com/about            => OK
https://domian.com/about/ceo     => error 404

When URLs suffix is changed to '/' or '.html' or something else, everything works properly.

https://domain.com/about.html            => OK
https://domian.com/about/ceo.html     => OK
or
https://domain.com/about/            => OK
https://domian.com/about/ceo/     => OK

Don't matter legacy_routing is false or true.

This website is hosted in Cpanel Server (Couldlinux with Litespeed [PHP info](https://mahanpt.com/info.php)). This problem don't occur in my local Contao installtion (php 7.4 and apache v2.4).

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2775

Apparently, `substr('foo', 0, -0)` is empty 🙈


aschempp:
Failling tests are not related

core-bundle/src/Routing/AbstractPageRouteProvider.php
```diff
@@ -167,8 +167,8 @@ protected function compareRoutes(Route $a, Route $b, array $languages = null): i
             return 1;
         }
 
-        $pathA = $a instanceof PageRoute ? substr($a->getPath(), 0, -\strlen($a->getUrlSuffix())) : $a->getPath();
-        $pathB = $b instanceof PageRoute ? substr($b->getPath(), 0, -\strlen($b->getUrlSuffix())) : $b->getPath();
+        $pathA = $a instanceof PageRoute && $a->getUrlSuffix() ? substr($a->getPath(), 0, -\strlen($a->getUrlSuffix())) : $a->getPath();
+        $pathB = $b instanceof PageRoute && $b->getUrlSuffix() ? substr($b->getPath(), 0, -\strlen($b->getUrlSuffix())) : $b->getPath();
```
m-vo, line 174: This could also work without calling `getUrlSuffix()` twice:

```suggestion
        $pathA = $a instanceof PageRoute ? substr($a->getPath(), 0, -\strlen($a->getUrlSuffix()) ?: null) : $a->getPath();
        $pathB = $b instanceof PageRoute ? substr($b->getPath(), 0, -\strlen($b->getUrlSuffix()) ?: null) : $b->getPath();
```

ausi (approved):

core-bundle/src/Routing/AbstractPageRouteProvider.php
aschempp: technically true, but we never want two inline comparison on the same line

leofeyer: @aschempp What exactly do you mean?

aschempp: There would be two `… ? … : …` on the same line, which is confusing and reported by EA-inspections (and probably other tools)

m-vo: Why would it be fine to use multiple null coalescing operators but not elvis? I think the idea of these rules is that you should avoid nesting ternary operators where it would be hard to spot their associativity.

PR #2789, opened 2021-02-22
Fix a PHPStan false-positive with changing values in $GLOBALS

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Fixes the issues reported in the CI.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -29,17 +29,17 @@ parameters:
             message: '#PHPDoc tag @param references unknown parameter: \$options#'
             path: %currentWorkingDirectory%/core-bundle/src/Image/PictureFactoryInterface.php
 
+        # Ignore PHPStan not being able to trace changes in $GLOBALS
+        -
+            message: '#Call to method PHPUnit\\Framework\\Assert::assertSame\(\) with .* will always evaluate to false\.#'
+            path: %currentWorkingDirectory%/core-bundle/tests/Framework/ContaoFrameworkTest.php
+
         # Ignore the wrong return type hint of the UrlGeneratorInterface::generate() method
         - '#Method Contao\\CoreBundle\\Picker\\AbstractPickerProvider::generateUrl\(\) never returns null so it can be removed from the return typehint\.#'
 
         # Ignore the missing transChoice() method of the Translator class
         - '#Call to an undefined method .*::transChoice\(\)\.#'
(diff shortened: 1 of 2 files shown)


m-vo:
Not sure if this is also a problem in 4.9.

phpstan.neon.dist
```diff
@@ -36,6 +36,11 @@ parameters:
         - '#Call to an undefined method .*::transChoice\(\)\.#'
         - '#Trying to mock an undefined method transChoice\(\) on class Symfony\\Component\\Translation\\Translator\.#'
 
+        # Ignore PHPStan not being able to trace changes in $GLOBALS
+        -
+            message: '#Call to method PHPUnit\\Framework\\Assert::assertSame\(\) with .* will always evaluate to false\.#'
```
leofeyer:
```suggestion
            message: '#Call to method PHPUnit\\Framework\\Assert::assertSame\(\) with \$GLOBALS.* will always evaluate to false\.#'
```

Limiting the regex will prevent false positives that are not related to `$GLOBALS`.

m-vo: Should we also drop the `path` then?

m-vo: This unfortunately won't work as the message looks like this:
```
Call to method PHPUnit\Framework\Assert::assertSame() with array(array('test.listener.a', 'onGetPageLayout'), array('test.listener.c', 'onGetPageLayout'), array('test.listener.b',            
         'onGetPageLayout')) and array(array('test.listener.c', 'onGetPageLayout')) will always evaluate to false.   
```

leofeyer: I see. 🤷🏻‍♂️

leofeyer (changes requested):
If this really only occurs in one single place, we should use `@phpstan-ignore-next-line` instead of a global exception rule.

bytehead (approved):

m-vo:
> If this really only occurs in one single place, we should use `@phpstan-ignore-next-line` instead of a global exception rule.

It's more than one place. :grimacing:

leofeyer:
Thank you @m-vo.

m-vo:
Just noticed that the issue from the `XliffFileLoaderTest` also happens in 4.9 :disappointed: 

@leofeyer Shall I make a second PR?

leofeyer:
Yes please.

m-vo:
see https://github.com/contao/contao/pull/2793

PR #2792, opened 2021-02-22
Fix order of ALTER statements being reversed

### LINKED ISSUE #2790: SQL alter table command split may cause an error
**Affected version(s)**

- Contao `^4.9`

**Related issues**

 - #169
 - #1651

**Description**

The SQL alter command is split into sub statements which leads to invalid sql statements. Adding a primary key does not work without a field, also creating an `AUTO_INCREMENT` field does not work without a primary key (mysql strict mode, not sure if it works without the strict mode):

For example a primary should be added before the field is created:

```sql
  ALTER TABLE tl_xyz_subscription ADD PRIMARY KEY (id)
  ALTER TABLE tl_xyz_subscription ADD state VARCHAR(255) DEFAULT '' NOT NULL
  ALTER TABLE tl_xyz_subscription ADD tstamp INT UNSIGNED DEFAULT 0 NOT NULL
  ALTER TABLE tl_xyz_subscription ADD id INT UNSIGNED AUTO_INCREMENT NOT NULL
```

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2790 
| Docs PR or issue | -

@dmolineus Could you check if this solves your problem?

### DIFF
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -169,6 +169,7 @@ class Installer
                     $prefix = $matches[1];
                     $sql = substr($sql, \strlen($prefix));
                     $parts = array_reverse(array_map('trim', explode(',', $sql)));
+                    $partsOrder = [];
 
                     for ($i = 0, $count = \count($parts); $i < $count; ++$i) {
                         $part = $parts[$i];
@@ -177,18 +178,18 @@ class Installer
                         switch (true) {
                             case 0 === strncmp($part, 'DROP ', 5):
                                 $return['ALTER_DROP'][md5($command)] = $command;
-                                $order[] = md5($command);
+                                $partsOrder[] = md5($command);
                                 break;
 
                             case 0 === strncmp($part, 'ADD ', 4):
                                 $return['ALTER_ADD'][md5($command)] = $command;
-                                $order[] = md5($command);
+                                $partsOrder[] = md5($command);
                                 break;
 
                             case 0 === strncmp($part, 'CHANGE ', 7):
                             case 0 === strncmp($part, 'RENAME ', 7):
                                 $return['ALTER_CHANGE'][md5($command)] = $command;
-                                $order[] = md5($command);
+                                $partsOrder[] = md5($command);
                                 break;
 
                             default:
@@ -196,6 +197,9 @@ class Installer
                                 break;
                         }
                     }
+
+                    $order += array_reverse($partsOrder);
+
                     break;
 
                 default:
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -198,7 +198,7 @@ class Installer
                         }
                     }
 
-                    $order += array_reverse($partsOrder);
+                    $order = array_merge($order, array_reverse($partsOrder));
 
                     break;
 
--- a/installation-bundle/src/Database/Installer.php
+++ b/installation-bundle/src/Database/Installer.php
@@ -213,6 +213,8 @@ class Installer
 
         // HOOK: allow third-party developers to modify the array (see #3281)
         if (isset($GLOBALS['TL_HOOKS']['sqlCompileCommands']) && \is_array($GLOBALS['TL_HOOKS']['sqlCompileCommands'])) {
+            @trigger_error('Using the "sqlCompileCommands" hook has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+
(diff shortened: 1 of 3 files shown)


m-vo:
As briefly discussed in slack, this unfortunately does not fully solve the issue. 

Doctrine actually behaves correctly by aggregating multiple `ALTER statements. Our strategy of splitting them for a nicer presentation is actually the bug. So, we should IMHO drop this 'feature' and keep the statements as they are.

If you agree, I'll change the behavior in this PR.
/cc @ausi

installation-bundle/src/Database/Installer.php
```diff
@@ -177,25 +178,28 @@ public function compileCommands(): void
...
+                                $partsOrder[] = md5($command);
                                 break;
 
                             default:
                                 $parts[$i + 1] .= ','.$part;
                                 break;
                         }
                     }
+
+                    $order += array_reverse($partsOrder);
```
ausi: I don’t think this works as expected, you need array_merge here IMO

ausi:
AFAIK the split feature is needed so skip single columns from being added or deleted. I’m not sure if we can drop this.

installation-bundle/src/Database/Installer.php
m-vo: You are right. I had the associative array in mind, but this is numeric. :facepalm: 

Adjusted in add21b0067de89ded475117bb2407267bf8ca461.

m-vo:
> AFAIK the split feature is needed so skip single columns from being added or deleted. I’m not sure if we can drop this.

Just tried + can confirm this, unfortunately. This wouldn't be possible anymore.

The logic is baked into `\Doctrine\DBAL\Platforms\MySqlPlatform#getAlterTableSQL()`. We would need to create our own platform and overwrite this logic to be able to get another change set. IMHO that's over the top.

The only other option I see would be to somehow temporarily disable certain checks (like you possibly would with FKs when importing large datasets) before executing the migrations. But that has downsides as well.

m-vo:
Do we plan to support selecting individual statements from the command as well? If not we could at least try to fix it there (or some hacky thing where we execute them at once if not selected individually…).

ausi:
> Do we plan to support selecting individual statements from the command as well?

I don’t think so. The command line has currently only the option to execute all or to skip `DELETE` queries. If we add the migrate command to the manager, keeping these two options should be enough I think.

ausi:
We could check for `empty($GLOBALS['TL_HOOKS']['sqlCompileCommands'])` in https://github.com/contao/contao/blob/6cdb7fce22c393ba19e0ca3b9237398b9bbba050/core-bundle/src/Command/MigrateCommand.php#L203 and skip the installer entirely if no hook is used.  
And deprecate the `sqlCompileCommands` hook.

m-vo:
> and skip the installer entirely if no hook is used.

Unfortunately there is also some magic in `checkEngineAndCollation()`. So I went with just skipping the splitting of `ALTER TABLE` statements depending on a new flag in `compileCommands()`. 

The `contao:migrate` command now favors combined statements if no `sqlCompileCommands` hooks are registered.

ausi:
> Unfortunately there is also some magic in `checkEngineAndCollation()`.

Is there a possibility to register a doctrine listener for that?

ausi:
There is another problem with we don’t split, `DROP COLUMN` would now get executed even if the migration runs “without deletes” 😕

m-vo:
Turns out this is kind of a bummer and unfortunately not that easy to fix. Schema updates are a complex thing...

Here is how we could approach it still:

- reimplement `Installer#checkEngineAndCollation()` as some suitable doctrine DBAL listener/extension/service
- stop using the `Installer` completely in the `contao:migrate` command *if* the (then deprecated) `sqlCompileCommands` hook is not registered
- never split/recombine any of the SQL that's output from the schema diff but alter the schemas instead &mdash; this means: instead of removing the drop statements when running `--schema-only` we would instead remove the columns from the 'from schema' (or add them to the 'to schema') and let doctrine handle order and grouping to make it work.

m-vo:
@dmolineus Are you still encountering the problems of #2790? And if so - did you find a suitable workaround?

This problem is a bit cumbersome to solve - I'd rather look at it again in Contao 5 when the `sqlCompileCommands` was  removed (and the install tool might be finally gone) if it doesn't do to much harm.

m-vo:
I'm closing this as the issue seems to have disappeared. 
At least for me and - judging by the "stale" state - likely everyone else.

PR #2800, opened 2021-02-24
Fix the json+ld schema extraction

Toflar:
It currently does not work when not providing a trailing slash which is totally valid though.

### DIFF
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -167,7 +167,7 @@ class Document
         }
 
         if (\is_string($data['@context'])) {
-            $data['@context'] = rtrim($data['@context'], '/') .'/';
+            $data['@context'] = rtrim($data['@context'], '/').'/';
 
             foreach ($data as $key => $value) {
                 if ('@type' === $key) {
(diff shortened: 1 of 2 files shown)


ausi:
For the prefix case (`contao:…`) the slash seems to be required.

Bildschirmfoto 2021-02-24 um 21 27 34
Bildschirmfoto 2021-02-24 um 21 27 25

But using it in `@context` directly seems to work with or without a trailing slash

Bildschirmfoto 2021-02-24 um 21 30 50

Toflar:
That might be an issue in the tool though. I didn't see anything requiring the trailing slash in the spec (https://www.w3.org/2018/jsonld-cg-reports/json-ld-api/#context-processing-algorithm). I think it's okay if we support both, isn't it?

ausi:
> I think it's okay if we support both, isn't it?

I agree. But we should use the trailing slash always ourselfs IMO to make sure other parsers understand our data too.

ausi (dismissed):

leofeyer:
Can you please rebase this onto the `4.9` branch?

Toflar:
I've actually based it on 4.9, only the target branch was wrong.

leofeyer:
The CS check is still failing. 🙈

Toflar:
Fixed :)

aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #2733, opened 2021-02-11
Make the FigureBuilder more fail tolerant

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Partly solves #2709 
| Docs PR or issue | todo

This PR addresses https://github.com/contao/contao/issues/2709#issuecomment-777626885 (1.). There now is an alternative `buildIfResourceExists()` and throwing of `InvalidResourceExceptions` is deferred to the `build()` method.

**Todo:**
 - [x] Adjust usage after #2732 is merged upstream

PR #2801, opened 2021-02-24
Add JSON+LD metadata to the search index

### LINKED ISSUE #2738: Allow custom metadata in search index
The goal is to index all JSON LD data found on a page alongside the search index entry.
That data must be available in the template so people can add custom data such as images etc. to a search result.

Toflar:
Closes https://github.com/contao/contao/issues/2738.
The actual part of just extracting the meta data and storing it is super easy but I thought we also need a use case for it.

One thing I keep adding for our customers is displaying an image to the search results. I've checked schema.org if there's a standardized way as to how you can declare a primary image for a webpage and turns out there is: https://schema.org/primaryImageOfPage

So I've implemented that by default.
How that information is added to the page is a different story and not part of this PR.


core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -321,6 +325,48 @@ protected function compile()
...
+							if (isset($v['https://schema.org/primaryImageOfPage']['alternateName']))
+							{
+								$figureMeta[Metadata::VALUE_ALT]  = $v['https://schema.org/primaryImageOfPage']['alternateName'];
+							}
+
+							$objTemplate->image = (object) $figureBuilder
+								->setSize($this->imgSize)
+								->setMetadata(new Metadata($figureMeta))
+								->build()
+								->getLegacyTemplateData();
```
ausi: Why not use `->applyLegacyTemplateData($objTemplate)` here?

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -321,6 +325,48 @@ protected function compile()
 					$objTemplate->hasContext = true;
 				}
 
+				// Add image to search results
+				$objTemplate->hasImage = false;
+				if (isset($arrResult[$i]['meta']) && null !== $arrResult[$i]['meta'])
+				{
+					$meta = json_decode($arrResult[$i]['meta'], true);
+					// Reverse so the last element wins
+					$meta = array_reverse($meta);
```
Toflar: Not sure if this is correct? One has to win for this case but is it the last or the first appearance on the page? /cc @ausi

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: I wanted to make sure it's not top level but resides inside `->image` instead so that if others want to add their own images, it feels more consistent.

core-bundle/src/Resources/contao/templates/search/search_default.html5
```diff
@@ -1,8 +1,13 @@
 
 <div class="<?= $this->class ?>">
 
-  <h3><a href="<?= $this->href ?>" title="<?= $this->title ?>"><?= $this->link ?></a> <span class="relevance">[<?= $this->relevance ?>]</span></h3>
-
+  <a href="<?= $this->href ?>" title="<?= $this->title ?>">
```
Toflar: Is this a good default? Guess most people will likely want to position the image next to the search results but I don't know which variant we should choose to provide the best styling flexibility?

core-bundle/src/Resources/contao/modules/ModuleSearch.php
ausi: How about an images array?

Toflar: I don't think there's any value in that. It's named `primaryImageOfPage`. There can only ever be one primary image. And moving it to the template would only delegate the problem to the user. They would pick the first array element anyway.

ausi: I think we should choose the first one then.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -20,6 +22,8 @@
  */
 class ModuleSearch extends Module
 {
+	use LegacyFigureBuilderTrait;
+
```
m-vo: We should IMHO not add new code for 4.12 using this trait but rather use [FigureBuilder#buildIfResourceExists()](https://github.com/contao/contao/pull/2733).

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -321,13 +325,69 @@ protected function compile()
...
+			}
+
+			$template->hasImage = true;
+			$template->image = (object) $figureBuilder
+				->setSize($this->imgSize)
+				->setMetadata(new Metadata($figureMeta))
+				->build()
+				->getLegacyTemplateData();
+
+			return;
```
m-vo, line 370: Let's add the built `Figure` to the template as well (like in https://github.com/contao/contao/pull/2734).

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: It's not merged yet so I couldn't use it :) We can remove that trait in all of the core once the other PR is merged.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -321,6 +325,48 @@ protected function compile()
...
+							}
+							if (isset($v['https://schema.org/primaryImageOfPage']['name']))
+							{
+								$figureMeta[Metadata::VALUE_TITLE]  = $v['https://schema.org/primaryImageOfPage']['name'];
+							}
+
+							if (isset($v['https://schema.org/primaryImageOfPage']['alternateName']))
+							{
+								$figureMeta[Metadata::VALUE_ALT]  = $v['https://schema.org/primaryImageOfPage']['alternateName'];
+							}
```
m-vo: How about this? Might be a bit easier to read:

```suggestion
							$figureMeta = new Metadata(array_filter([
								Metadata::VALUE_CAPTION => $v['https://schema.org/primaryImageOfPage']['caption'] ?? null,
								Metadata::VALUE_TITLE => $v['https://schema.org/primaryImageOfPage']['name'] ?? null,
								Metadata::VALUE_ALT => $v['https://schema.org/primaryImageOfPage']['alternateName'] ?? null,
							]));
```

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: Done.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: Done

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: Done.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -313,13 +315,73 @@ protected function compile()
...
+		$meta = array_reverse($meta);
+
+		foreach ($meta as $v)
+		{
+			if (!isset($v['https://schema.org/primaryImageOfPage']))
+			{
+				continue;
+			}
+
+			$figureBuilder = $this->getFigureBuilderIfResourceExists($v['https://schema.org/primaryImageOfPage']['contentUrl']);
```
m-vo: Leftover :slightly_smiling_face: 
```suggestion
```

Toflar: Yeah I've already fixed that locally but the CS fixer takes forever to run :D

core-bundle/src/Resources/contao/modules/ModuleSearch.php
Toflar: Nice, that's done too.

Toflar:
This PR is ready. The only thing imho left to worry about would be the default markup of the template.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -313,13 +315,60 @@ protected function compile()
...
+		if (!isset($result['meta']) || null == $result['meta'])
+		{
+			return;
+		}
+
+		$meta = json_decode($result['meta'], true);
+
+		foreach ($meta as $v)
+		{
+			if (!isset($v['https://schema.org/primaryImageOfPage']))
```
ausi:
```suggestion
			if (!isset($v['https://schema.org/primaryImageOfPage']['contentUrl']))
```
So that we cannot get an undefined array key error in line 347.

core-bundle/src/Resources/contao/templates/search/search_default.html5
ausi: I think I would keep the `<h3>` as it was before and add the image with its own `<a>` before the headline. 

We could even set the URL in the figure builder with `$figureBuilder->setLinkHref($result['url'])` in `ModuleSearch:355`.

Toflar: Updated in 49ef4f677061c7d114c9ef6a591fcf248d8e6827.

core-bundle/src/Resources/contao/templates/search/search_default.html5
```diff
@@ -1,6 +1,10 @@
 
 <div class="<?= $this->class ?>">
 
+  <?php if ($this->hasImage): ?>
+    <?php $this->insert('picture_default', $this->image->picture) ?>
```
ausi:
```suggestion
    <?php $this->insert('image', (array) $this->image) ?>
```

Toflar: Ah yeah right :D Done

ausi (dismissed):

m-vo (dismissed):

ausi:
CI error is unrelated I think, see #2879

Toflar:
I re-ran the jobs, it's green now :)

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -313,13 +315,61 @@ protected function compile()
...
 			$this->Template->header = vsprintf($GLOBALS['TL_LANG']['MSC']['sResults'], array($from, $to, $count, $strKeywords));
 			$this->Template->duration = System::getFormattedNumber($query_endtime - $query_starttime, 3) . ' ' . $GLOBALS['TL_LANG']['MSC']['seconds'];
 		}
 	}
+
+	protected function addImageToTemplateFromSearchResult(array $result, Template $template): void
+	{
+		$template->hasImage = false;
+
+		if (!isset($result['meta']) || null == $result['meta'])
```
leofeyer: This does not make sense. If `$result['meta']` is `null`, the `!isset()` check would already match. Did you mean `!$result['meta']` instead? In case it is an empty string or false?

ausi: But it’s only a `==` not `===` so it would also match `false`, `0` and more...

Toflar: Improved in 9d8bd2be28ec1c46475512b2ba94e73c0f09803c.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -313,13 +315,61 @@ protected function compile()
...
+
+			$figureMeta = new Metadata(array_filter(array(
+				Metadata::VALUE_CAPTION => $v['https://schema.org/primaryImageOfPage']['caption'] ?? null,
+				Metadata::VALUE_TITLE => $v['https://schema.org/primaryImageOfPage']['name'] ?? null,
+				Metadata::VALUE_ALT => $v['https://schema.org/primaryImageOfPage']['alternateName'] ?? null,
+			)));
+
+			$figure = $figureBuilder
+				->setSize($this->imgSize)
+				->setMetadata(new Metadata($figureMeta))
```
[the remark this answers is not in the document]
Toflar: Fixed in 9d8bd2be28ec1c46475512b2ba94e73c0f09803c.

ausi (dismissed):

aschempp (dismissed):
Great idea! We might add a page image (including `og:image`) setting at some point 😊

ausi (dismissed):

core-bundle/src/Resources/contao/modules/ModuleSearch.php
leofeyer: I still do not get it. Isn't `if (!array_key_exists('meta', $result) || null === $result['meta'])` the same as `if (!isset($result['meta']))`?

leofeyer: Yes, it is: https://3v4l.org/GDsMU

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -313,13 +315,61 @@ protected function compile()
...
 			$this->Template->header = vsprintf($GLOBALS['TL_LANG']['MSC']['sResults'], array($from, $to, $count, $strKeywords));
 			$this->Template->duration = System::getFormattedNumber($query_endtime - $query_starttime, 3) . ' ' . $GLOBALS['TL_LANG']['MSC']['seconds'];
 		}
 	}
+
+	protected function addImageToTemplateFromSearchResult(array $result, Template $template): void
+	{
+		$template->hasImage = false;
+
+		if (!\array_key_exists('meta', $result) || null === $result['meta'])
```
leofeyer:
```suggestion
		if (!isset($result['meta']))
```

See https://3v4l.org/GDsMU

leofeyer (changes requested):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #2807, opened 2021-02-25
Fix the legacy entry points not working in Contao 4.11

qzminski:
The first error I have encountered when trying to open a script that uses a legacy entry point was:

```
"Contao\CoreBundle\Controller\InitializeController" has no container set, did you forget to define it as a service subscriber?
```

Apparently in Contao 4.11 that controller is no longer considered by Symfony as a service and thus the container is not injected. Thus my adjustments to the `services.yml` file.

Then another error popped up:

```
Service "contao.framework" not found: even though it exists in the app's container, the container inside "Contao\CoreBundle\Controller\InitializeController" is a smaller service locator that only knows about the "doctrine", "http_kernel", "parameter_bag", "request_stack", "router", "security.authorization_checker", "security.csrf.token_manager", "security.token_storage", "session" and "twig" services. Try using dependency injection instead.
```

It seems that other legacy Contao controllers (such as `BackendController`) do extend the Contao's `AbstractController`, whereas the `InitializeController` extends the Symfony's one. I am not sure if it's intended or not, but having the class extend the Contao's controller does the job.

Please double check this PR as I am not 100% sure if it has any negative impacts on other stuff.

/cc @aschempp

### DIFF
--- a/core-bundle/src/Controller/InitializeController.php
+++ b/core-bundle/src/Controller/InitializeController.php
@@ -12,7 +12,9 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Controller;
 
+use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\CoreBundle\Response\InitializeControllerResponse;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
@@ -121,6 +123,16 @@ class InitializeController extends AbstractController
         );
 
         return new InitializeControllerResponse('', 204);
+    }
+
+    public static function getSubscribedServices()
+    {
+        $services = parent::getSubscribedServices();
+
+        $services['contao.framework'] = ContaoFramework::class;
+        $services['kernel'] = KernelInterface::class;
+
+        return $services;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


qzminski:
Also, the `kernel` service does not seem to longer exist (at least in the injected container). Shouldn't it be `http_kernel`?

https://github.com/contao/contao/blob/b4e1c78754608e37459d5022a6ca6618c04b78a7/core-bundle/src/Controller/InitializeController.php#L179

https://github.com/contao/contao/blob/b4e1c78754608e37459d5022a6ca6618c04b78a7/core-bundle/src/Controller/InitializeController.php#L211

aschempp (changes requested):
Veery interesting why this ever worked. Looks like Symfony 5 changed this behaviour and previously injected the full container for any controller.

Anyway, extending the `BackendController` certainly isn't correct, because the `initialize.php` can be both front or back end (and most likely is front end). To fix the issues, we should extend `public static function getSubscribedServices()` and request the necessary services (`kernel` and `contao.framework` I think).

qzminski:
I don't want to extend `BackendController` but Contao's `AbstractController` …

aschempp (changes requested):
This PR now changes the parent class from Symfony's `AbstractController` to Contao's `AbstractController` (I didn't notice that from seeing just the `use` statement being removed).

This will fix the `contao.framework` not being available "by accident". It will not provide `kernel` though, as that is still not in the subscribed services. We should correctly implement the `getSubscribedServices` (similar to https://github.com/contao/contao/blob/4.x/core-bundle/src/Controller/AbstractController.php#L22-L31) and not change the parent class.

aschempp (approved):
This implementation is correct imho. https://github.com/contao/contao/pull/2833 would be an alternative, having the advantage that our dependencies are "hardcoded" and we don't rely on any parent class behaviour.

bytehead (approved):

leofeyer:
I have merged #2833 now.

Issue #937, opened 2019-11-06
Set width and height attributes as an aspect ratio for images


ausi:
Browsers now seem to change the behavior of how `width` and `height` attributes are handled in combination with responsive CSS, see https://github.com/whatwg/html/pull/4952

The new behavior seems to be the way that @fritzmg already suggested five years ago in https://github.com/contao/core/issues/7500#issuecomment-65933582 which means you can use `width` and `height` now to define an aspect ratio for the image.

To improve our usage of this feature in Contao we should now always use `width` and `height` attributes as long as all `<source>`s have the same aspect ratio.

This is a bugfix IMO because the behavior of browsers has changed.

For more details about the changes see <https://bugzilla.mozilla.org/show_bug.cgi?id=1547231>, <https://bugs.chromium.org/p/chromium/issues/detail?id=979891>, <https://bugs.webkit.org/show_bug.cgi?id=201641>, <https://twitter.com/tomayac/status/1165902659083280384>, <https://www.youtube.com/watch?v=4-d_SoCHeWE>.

There are also plans to add `width` and `height` support for `<source>` tags in the future, see https://github.com/whatwg/html/issues/4968

leofeyer:
What exactly needs to be changed on our end?

ausi:
> To improve our usage of this feature in Contao we should now always use `width` and `height` attributes as long as all `<source>`s have the same aspect ratio.

I will provide a pull request once we agreed on bugfix vs. feature 🙃

leofeyer:
I guess this is a bugfix.

Toflar:
Bugfix

ausi:
Pull requests: #940 (4.8) #939 (4.8)

We should probably wait until one of the browsers released a stable version with this feature activated. Firefox 71 will be released in december AFAIK.

leofeyer:
Would it break something if we released it now?

ausi:
It shouldn’t break something in general IMO. There might be some edge cases with using the `sizes` attribute, but they should be easy to fix with CSS I think.

I’d still wait for the first released browser version to be more certain that this new behavior stays, and so that we can test our implementation.

fritzmg:
I did find a problem. Consider the following setup:

![Screenshot_2019-12-18 Themes › Contao Official Demo › Image sizes › Edit image size ID 1 demo contao org(2)](https://user-images.githubusercontent.com/4970961/71093994-82c00280-21aa-11ea-835a-2f5425a5dea5.png)

[...]

PR #2811, opened 2021-02-26
Add width/height attributes to the picture source

### LINKED ISSUE #2746: Set picture source width/height attributes once supported by browers
We added `width` and `height` attributes to the `` element in #940

It looks like browsers will support these attributes on `<source>` elements too in the future. See https://github.com/whatwg/html/pull/5894 and https://groups.google.com/a/chromium.org/g/blink-dev/c/GEjjJE0LCnM/m/0p8CJ48hAgAJ

As soon as the first browser ships this in a stable version we should add these attributes I think.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2746

We should probably wait for Chrome 90 to be released on April 13th before merging this. https://chromestatus.com/feature/5737185317748736

UPDATE: Implementation tickets for Safari and Firefox:

- <https://bugzilla.mozilla.org/show_bug.cgi?id=1694741>
- <https://bugs.webkit.org/show_bug.cgi?id=222368>


m-vo:
This will also need to be adjusted in `Image/Studio/_macros.html.twig` for Contao 4.11.

core-bundle/src/Resources/contao/templates/picture/picture_default.html5
```diff
@@ -2,11 +2,11 @@
...
+<img src="<?= $this->img['src'] ?>"<?php if ($this->img['srcset'] !== $this->img['src']): ?> srcset="<?= $this->img['srcset'] ?>"<?php endif; ?><?php if (!empty($this->img['sizes'])): ?> sizes="<?= $this->img['sizes'] ?>"<?php endif; ?><?php if (!empty($this->img['width']) && !empty($this->img['height'])): ?> width="<?= $this->img['width'] ?>" height="<?= $this->img['height'] ?>"<?php endif; ?> alt="<?= $this->alt ?>"<?php if (!empty($this->img['loading'])): ?> loading="<?= $this->img['loading'] ?>"<?php endif; ?><?php if ($this->title): ?> title="<?= $this->title ?>"<?php endif; ?><?php if ($this->class || !empty($this->img['class'])): ?> class="<?= trim($this->class.' '.$this->img['class']) ?>"<?php endif; ?><?= $this->attributes ?> itemprop="image">
```
m-vo, line 9: So images where `$this->img['hasSingleAspectRatio'])` is not set will now receive a value here, right? 

Doesn't this affect rendering for browsers that do not understand aspect ratios from `source` entries? :thinking:

ausi: > So images where `$this->img['hasSingleAspectRatio'])` is not set will now receive a value here, right?

Correct.

> Doesn't this affect rendering for browsers that do not understand aspect ratios from `source` entries? 🤔

It shouldn’t matter as you size your images via CSS. We had a similar discussion in https://github.com/contao/contao/issues/937#issuecomment-568181620

ausi:
Chrome 90 is released, so I think we can merge this. 🎉

fritzmg (approved):

Toflar (approved):

m-vo (changes requested):
> This will also need to be adjusted in Image/Studio/_macros.html.twig for Contao 4.11.

Do we want to do this in this PR or in another one?

ausi:
> Do we want to do this in this PR or in another one?

Another one I think because this one is for version 4.9.

ausi:
PR for Contao 4.11: #2982

leofeyer:
Thank you @ausi.

PR #2815, opened 2021-02-28
Do not tag contao.db.tl_module.<id> in articles

### LINKED ISSUE #2814: Do not tag contao.db.tl_module.<id> for articles
**Affected version(s)**

Contao >=4.9 

**Description**

Every generated `Module` tags the response with `contao.db.tl_module.<id>`, even when rendering articles:

https://github.com/contao/contao/blob/712de2476dc6100cad1876da1f22b809f7558d5c/core-bundle/src/Resources/contao/modules/Module.php#L242

This way the `x-cache-tags` of a page displaying article `ID 5` will look like this even though module `ID 5` is not present:
```
contao.db.tl_article.5,[…],contao.db.tl_module.5
```

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2814 
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -236,7 +236,7 @@ abstract class Module extends Frontend
 		}
 
 		// Tag the module (see #2137)
-		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
+		if (!$this instanceof ModuleArticle && System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
 			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
 			$responseTagger->addTags(array('contao.db.tl_module.' . $this->id));
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -236,10 +236,10 @@ abstract class Module extends Frontend
 		}
 
 		// Tag the module (see #2137)
-		if (!$this instanceof ModuleArticle && System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
+		if (!empty($tags = $this->getResponseCacheTags()) && System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 		{
 			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
-			$responseTagger->addTags(array('contao.db.tl_module.' . $this->id));
+			$responseTagger->addTags($tags);
 		}
 
 		return $this->Template->parse();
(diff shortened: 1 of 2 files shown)


aschempp (dismissed):

fritzmg (dismissed):

Toflar:
This looks incorrect to me. Why would the `Module` know about `ModuleArticle`? And what if there are other (legacy) modules, that have the same issue?
The only way to fix this properly would be to extract the logic to a `protected function getResponseCacheTags()` and override this in `ModuleArticle` to return nothing or so.
I know we're talking legacy modules here but still, it's worth fixing correctly imho.

Toflar (changes requested):
This looks incorrect to me. Why would the `Module` know about `ModuleArticle`? And what if there are other (legacy) modules, that have the same issue?
The only way to fix this properly would be to extract the logic to a `protected function getResponseCacheTags()` and override this in `ModuleArticle` to return nothing or so.
I know we're talking legacy modules here but still, it's worth fixing correctly imho.

m-vo:
> The only way to fix this properly would be to extract the logic to a protected function getResponseCacheTags() and override this in ModuleArticle to return nothing or so.

Thought about this as well and and I agree in general… Although I think articles are the only "virtual" modules that are ever rendered without a `tl_module` context, so this might be overkill. :thinking:

fritzmg:
`Module` is a legacy parent class that already provides functionality for specific classes that are extending it. See for example `Module::renderNavigation` which is not used by `Module` itself, but by `ModuleNavigation` and `ModuleSitemap`. So imho it would not be far fetched to implement special handling for the special `ModuleArticle`.

But otherwise creating a `protected function getResponseCacheTags()` (and then overriding it in `Contao\ModuleArticle`) is easy to do 👍

leofeyer:
I am fine either way.

aschempp:
> But otherwise creating a protected function getResponseCacheTags() (and then overriding it in Contao\ModuleArticle) is easy to do

It is. It just won't work with fragments. I'd go with the "simple" version, even though we know it's not "ideal". `ModuleArticle` isn't really a module anyway …

fritzmg:
> It is. It just won't work with fragments.

Why? This does not affect fragments. Fragments tag their response in `AbstractFrontendModuleController::__invoke`.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -236,10 +236,10 @@ public function generate()
 		}
 
 		// Tag the module (see #2137)
-		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
+		if (!empty($tags = $this->getResponseCacheTags()) && System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
```
Toflar: Should be the other way around :) We don't need to collect the tags if there's no response tagger.

Toflar (commented):
Just a minor nitpick. Then we can get this merged :)

core-bundle/src/Resources/contao/modules/Module.php
m-vo: see 708a32aed64d7eaa2834c037ba88adc9c4579138

Toflar (approved):
Thank you @m-vo <3

leofeyer:
Thank you @m-vo.

PR #2818, opened 2021-03-01
Fix the logout handler in Symfony 5

fritzmg:
This PR fixes the following error in Contao **4.11.0**:

```
TypeError:
Argument 2 passed to Contao\CoreBundle\Security\Logout\LogoutHandler::logout() must be an instance of Symfony\Component\HttpFoundation\Response, null given, called in vendor\symfony\security-bundle\Security\LegacyLogoutHandlerListener.php on line 42

  at vendor\contao\contao\core-bundle\src\Security\Logout\LogoutHandler.php:50
  at Contao\CoreBundle\Security\Logout\LogoutHandler->logout()
     (vendor\symfony\security-bundle\Security\LegacyLogoutHandlerListener.php:42)
  at Symfony\Bundle\SecurityBundle\Security\LegacyLogoutHandlerListener->onLogout()
     (vendor\symfony\event-dispatcher\EventDispatcher.php:270)
  at Symfony\Component\EventDispatcher\EventDispatcher::Symfony\Component\EventDispatcher\{closure}()
     (vendor\symfony\event-dispatcher\EventDispatcher.php:230)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
     (vendor\symfony\event-dispatcher\EventDispatcher.php:59)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
     (vendor\symfony\security-http\Firewall\LogoutListener.php:122)
  at Symfony\Component\Security\Http\Firewall\LogoutListener->authenticate()
     (vendor\symfony\security-bundle\Debug\WrappedLazyListener.php:49)
  at Symfony\Bundle\SecurityBundle\Debug\WrappedLazyListener->authenticate()
     (vendor\symfony\security-http\Firewall\AbstractListener.php:26)
  at Symfony\Component\Security\Http\Firewall\AbstractListener->__invoke()
     (vendor\symfony\security-bundle\Debug\TraceableFirewallListener.php:62)
  at Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener->callListeners()
     (vendor\symfony\security-http\Firewall.php:86)
  at Symfony\Component\Security\Http\Firewall->onKernelRequest()
     (vendor\symfony\event-dispatcher\Debug\WrappedListener.php:117)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke()
     (vendor\symfony\event-dispatcher\EventDispatcher.php:230)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
     (vendor\symfony\event-dispatcher\EventDispatcher.php:59)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
     (vendor\symfony\event-dispatcher\Debug\TraceableEventDispatcher.php:151)
  at Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch()
     (vendor\symfony\http-kernel\HttpKernel.php:133)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
     (vendor\symfony\http-kernel\HttpKernel.php:79)
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (vendor\symfony\http-kernel\Kernel.php:195)
     (web\index.php:31)
  at require('web\\index.php')
     (web\app.php:4)           
```

**Reproduction**:

1. Create a login form, login page and member.
2. Create a logout page.
3. Log into the front end.
4. Use the link in the menu for the logout.

Apparently the `LogoutEvent` might not contain a `Response` object any more.


bytehead:
Does this not have to follow the interface? 
https://github.com/symfony/security-http/blob/be7540a95c445bd82ad088d7e236d0463b526b26/Logout/LogoutHandlerInterface.php#L32

core-bundle/src/Security/Logout/LogoutHandler.php
```diff
@@ -47,7 +47,7 @@ public function __construct(ContaoFramework $framework, LoggerInterface $logger
         $this->logger = $logger;
     }
 
-    public function logout(Request $request, Response $response, TokenInterface $token): void
+    public function logout(Request $request, ?Response $response, TokenInterface $token): void
```
bytehead, line 50: Does this method not have to follow the interface: https://github.com/symfony/security-http/blob/5.x/Logout/LogoutHandlerInterface.php#L32 ?

bytehead (changes requested):

core-bundle/src/Security/Logout/LogoutHandler.php
fritzmg: Hm, interesting. PHP did not complain during testing 🤔 . Then I suppose the real error is that the `LogoutEvent` is missing a response.

fritzmg: To me this looks like an error in Symfony. `LogoutEvent::getResponse()` returns either a `Response` object or `null`. However, [`LegacyLogoutHandlerListener::onLogout`](https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Bundle/SecurityBundle/Security/LegacyLogoutHandlerListener.php#L41-L43) relies on this method always returning a `Response` object, since the registered `logoutHandler` is supposed to implement `LogoutHandlerInterface` where the second parameter cannot be null (although apparently it can 🤷).

fritzmg:
I think this is definitely a bug in Symfony. For backwards compatibility with the old security components Symfony implements a single [`LegacyLogoutHandlerListener`](https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Bundle/SecurityBundle/Security/LegacyLogoutHandlerListener.php) as an event _subscriber_. The [`SecurityExtension`](https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php) then dynamically creates multiple instances of this service for each service that is a [logout handler](https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php#L422-L429) _or_ a [success handler](https://github.com/symfony/symfony/blob/252f85c2c249da7d8f2a490b57cc71199d51fcc9/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php#L454-L459).

But the logout handler depends on the success handler. See [here in Symfony 4](https://github.com/symfony/symfony/blob/07d7f6bad7d4adce5004eaf7dd539b4f0769c6a9/src/Symfony/Component/Security/Http/Firewall/LogoutListener.php#L95-L105) - the success handler is always executed _before_ the logout handler, because the success handler returns the `Response` object for the registered logout handlers. 

Since this is now handled within one single service subscriber, no priority can be set for either instance. Thus there is no guarantee that the success handler is executed before the logout handler - and thus the `Response` object can be missing.

To fix this within Symfony I have altered the `SecurityExtension` and `LegacyLogoutHandlerListener` so that it is implemented as a `kernel.event_listener` - where the success handler services will be given a higher priority than the logout handler services.

core-bundle/src/Security/Logout/LogoutHandler.php
bytehead: could `public function logout(Request $request, Response $response = null, TokenInterface $token): void` be a solution?

fritzmg: Both `?Response $resopnse` and `Response $response = null` are actually valid: https://3v4l.org/R45Tn

bytehead: Looks good to me then! 🎉

bytehead (approved):

leofeyer (changes requested):
Overwriting the method signature of the interface cannot be the correct fix. @aschempp /cc

aschempp (approved):
I agree with @leofeyer that this is a Symfony bug. However, the proposed changes would (temporary) fix our issue, and we do not need the response anyway, I would be fine to fix it this way. We should however also open an issue or PR with Symfony to get this fixed.

leofeyer:
Agreed. Interesting that PHP allows to nullify the parameter in the concrete class but not the other way round:

```php
interface Foo {
    public function bar(string $str, Response $response);
}

class FooClass implements Foo {
    // This works
    public function bar(string $str, ?Response $response) {}
}
```

```php
interface Foo {
    public function bar(string $str, ?Response $response);
}

class FooClass implements Foo {
    // This does not
    public function bar(string $str, Response $response) {}
}
```

See https://3v4l.org/XD3Mp

leofeyer (approved):

bytehead:
Yep, the behaviour is quite strange.

leofeyer:
Thank you @fritzmg.

PR #2827, opened 2021-03-03
Fix generating error pages by name

### LINKED ISSUE #2809: Routing debug issues
**Affected version(s)**

Contao 4.9

**Description**

I tried to debug some routes with `vendor/bin/contao-console route:match` but get following errors:

`vendor/bin/contao-console route:match /preview.php/_wdt/205c80`
```
In RouteProvider.php line 115:
                                                                                                                                                   
  Return value of Contao\CoreBundle\Routing\RouteProvider::getRouteByName() must be an instance of Symfony\Component\Routing\Route, null returned                                        
```

I locally fixed it with following condition, I'm able to select a route:
```php
if (!isset($routes[$name])) {
    throw new RouteNotFoundException('Route name does not match a page ID');
}
```

```
 [OK] Route "tl_page.893.error_404" matches                                                                             
                                                                                                                        
 Select one of the matching routes:
  [0] tl_page.893.error_404.locale
  [1] tl_page.893.error_404
```

If I select any I get another error now:

```
  The route "tl_page.893.error_404" does not exist.  
```

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2809

### DIFF
--- a/core-bundle/src/Routing/Route404Provider.php
+++ b/core-bundle/src/Routing/Route404Provider.php
@@ -56,17 +56,24 @@ class Route404Provider implements RouteProviderInterface
 
     public function getRouteByName($name): Route
     {
-        throw new RouteNotFoundException('This router cannot load routes by name');
+        $routes = $this->getRoutes();
+
+        if (!isset($routes[$name])) {
+            throw new RouteNotFoundException('Route name "'.$name.'" is not supported by '.__METHOD__);
+        }
+
+        return $routes[$name];
     }
 
     public function getRoutesByNames($names): array
     {
-        // Support console and web inspector profiling
+        $routes = $this->getRoutes();
+
         if (null === $names) {
-            return $this->getRoutes();
+            return $routes;
         }
 
-        return [];
+        return array_intersect_key($routes, array_flip($names));
     }
 
     private function getRoutes(array $languages = null): array
--- a/core-bundle/src/Routing/RouteProvider.php
+++ b/core-bundle/src/Routing/RouteProvider.php
@@ -339,6 +339,7 @@ class RouteProvider implements RouteProviderInterface
 
         foreach ($names as $name) {
             $parts = explode('.', $name);
+
             if ('tl_page' !== ($parts[0] ?? '') || 'error_404' === ($parts[2] ?? '')) {
                 continue;
             }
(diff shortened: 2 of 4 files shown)


core-bundle/src/Routing/RouteProvider.php
```diff
@@ -338,7 +338,8 @@ private function getPageIdsFromNames(array $names): array
         $ids = [];
 
         foreach ($names as $name) {
-            if (0 !== strncmp($name, 'tl_page.', 8)) {
+            $parts = explode('.', $name);
```
bytehead, line 341: CS (newline)?

bytehead (changes requested):

core-bundle/src/Routing/RouteProvider.php
aschempp: already fixed 😊

core-bundle/src/Routing/Route404Provider.php
```diff
@@ -56,17 +56,24 @@ public function getRouteCollectionForRequest(Request $request): RouteCollection
...
-        // Support console and web inspector profiling
+        $routes = $this->getRoutes();
+
         if (null === $names) {
-            return $this->getRoutes();
+            return $routes;
         }
 
-        return [];
+        return array_intersect_key($routes, array_flip($names));
```
leofeyer, line 76: Do we need to cast `(array) $names` here? Because even though the phpDoc says `array|null`, the type is not enforced and `$names` could be anything (except `null` of course).

core-bundle/src/Routing/RouteProvider.php
```diff
@@ -338,7 +338,9 @@ private function getPageIdsFromNames(array $names): array
...
+
+            if ('tl_page' !== ($parts[0] ?? '') || 'error_404' === ($parts[2] ?? '')) {
```
leofeyer:
```suggestion
            if ('tl_page' !== $parts[0] || 'error_404' === ($parts[2] ?? '')) {
```

`$parts[0]` will always be set, even if the string is empty or does not contain the split character (see https://3v4l.org/O2Qvv).

core-bundle/src/Routing/Route404Provider.php
aschempp: if the interface would enforce it, you would get an error. Now you get an error as well if you don't follow the docs…

leofeyer (approved):

PR #2829, opened 2021-03-03
Start the session to check for a user token

### LINKED ISSUE #2810: Preview listener does not show the debug bar
**Affected version(s)**

Contao 4.9 and later

**Description**

1. I have the debug mode enabled by switching the app mode in the `.env` file
2. I use the preview link from the backend
3. I expect to see the debug toolbar which does not appear

After debugging it a bit I can provide following additional information:

1. The markup for the toolbar is injected
2. The toolbar tries to call `/preview.php/_wdt/xyz` which ends in a endless redirect
3. The redirect is caused by the `Contao\CoreBundle\EventListener\PreviewAuthenticationListener` which try to login the user as the route is not defined as a backend route

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2810

Explicitly bypassing the `_wdt` route as suggested in https://github.com/contao/contao/issues/2810#issuecomment-786569685 is not the correct solution. The back end user is always available from the session, but we need to start the session if no other service has done that before the `TokenChecker`.

### DIFF
--- a/core-bundle/src/Security/Authentication/Token/TokenChecker.php
+++ b/core-bundle/src/Security/Authentication/Token/TokenChecker.php
@@ -174,7 +174,16 @@ class TokenChecker
 
     private function getTokenFromSession(string $sessionKey): ?TokenInterface
     {
-        if (!$this->session->isStarted() || !$this->session->has($sessionKey)) {
+        if (!$this->session->isStarted()) {
+            $request = $this->requestStack->getMasterRequest();
+
+            if (!$request || !$request->hasPreviousSession()) {
+                return null;
+            }
+        }
+
+        // This will start the session if Request::hasPreviousSession() was true
+        if (!$this->session->has($sessionKey)) {
             return null;
         }
 


leofeyer:
This looks correct. @dmolineus Can you confirm that the changes fix your issue?

core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -174,7 +174,16 @@ private function getTokenFromStorage(string $context): ?TokenInterface
...
+        if (!$this->session->isStarted()) {
+            $request = $this->requestStack->getMasterRequest();
+
+            if (!$request || !$request->hasPreviousSession()) {
+                return null;
+            }
+        }
+
+        // This will start the session if Request::hasPreviousSession() was true
+        if (!$this->session->has($sessionKey)) {
```
dmolineus, line 186: I don't like trusting on the internal be behaviour of the session component. I prefer starting the session explicitly. The behavior is not documented in the `SessionInterface`.

dmolineus (commented):
> This looks correct. @dmolineus Can you confirm that the changes fix your issue?

Yes, it does. I'd love to see it unit tested however. :-)

leofeyer (approved):

core-bundle/src/Security/Authentication/Token/TokenChecker.php
aschempp: Have you ever started the session manually before accessing it? I don't think anyone does, the only check I ever see is **not** accessing the session if it hasn't been started (because accessing it would start it).

dmolineus: Yes, I have seen it before (e.g. https://github.com/m-vo/contao-survey/pull/31/files#diff-51927511979faf55d65c36a68217dc98328ead09b2fdca0d71b57bf8ef3b4d72R248) and usually do it myself. I started doing such stuff which looks not necessary at the first time because the relying on internal implementations causes bugs when things change in the future. Maybe they will, but if they do, I don't have to care. But okay, it's not my code base, just my opinion. :-)

fritzmg: > The behavior is not documented in the `SessionInterface`.

It is documented [here](https://symfony.com/doc/current/session.html#avoid-starting-sessions-for-anonymous-users):

> Sessions are automatically started whenever you read, write or even check for the existence of data in the session.

m-vo: And still there are edge cases: In the referenced example where we're reading/writing into a custom attribute bag the session is not started for instance.

fritzmg: Then Symfony either needs to fix that or update its documentation :). Can you provide a PR? 😁

m-vo: Maybe, yes. :smile: But on the other hand you're technically not dealing with the `SessionInterface` directly in this case.

PR #2833, opened 2021-03-04
Use dependency injection for the InitializeController

aschempp:
As discussed with @qzminski, this is an alternative to https://github.com/contao/contao/pull/2807. As we were not sure what's better, we decided to present both implementations to decide on 😊

### DIFF
--- a/core-bundle/src/Controller/InitializeController.php
+++ b/core-bundle/src/Controller/InitializeController.php
@@ -12,9 +12,11 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Controller;
 
+use Contao\CoreBundle\Framework\ContaoFramework;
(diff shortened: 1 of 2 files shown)


fritzmg:
Personally I always use DI in my controllers as it's more convenient and also easier to read and the (my) IDE will know which service is which. However, there is probably some advantage for using `getSubscribedServices` instead?

qzminski (approved):
Works just fine.

aschempp:
> However, there is probably some advantage for using `getSubscribedServices` instead?

The only advantage is the services are not instantiated if you don't use them. So if your controller has multiple actions, or an otherwise complex setup, or parent methods like the `AbstractController` that are mostly unused, then `getSubscribedServices` can help. But otherwise I agree DI is better/more explicit.

bytehead (approved):

PR #2834, opened 2021-03-04
Allow passing an array of IDs to User::isMemberOf()

### LINKED ISSUE #2753: Show protected pages only in sitemap for contao searchcrawler
**Affected version(s)**
Contao 4.11

**Description**
With the new dynamic sitemap-generator it should be possible to fix this old ticket (https://github.com/contao/contao/issues/1407)

**How to reproduce**

1. Set `index_protected:true` in `config.yml`
2. Set robots-tag to `noindex,follow`
3. set `Protect page`

The page would appear in public sitemap.xml, but should only appear for the contaocrawler.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2753
| Docs PR or issue | -

To fix #2753 (do not show protected pages in the sitemap if there is no front end user or the user groups do not match), I have adjusted the `User::isMemberOf()` method to also accept an array of group IDs. This allows us to greatly simplify other checks as well:

### Current

```php
$blnFeUserLoggedIn = System::getContainer()->get('contao.security.token_checker')->hasFrontendUser();

while ($objCalendar->next())
{
	if ($objCalendar->protected)
	{
		if (!$blnFeUserLoggedIn || !\is_array($this->User->groups))
		{
			continue;
		}

		$groups = StringUtil::deserialize($objCalendar->groups);

		if (empty($groups) || !\is_array($groups) || \count(array_intersect($groups, $this->User->groups)) < 1)
		{
			continue;
		}
	}

	// …
}
```

### New

```php
$user = null;

if (System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
{
	$user = FrontendUser::getInstance();
}

while ($objCalendar->next())
{
	if ($objCalendar->protected && (!$user || !$user->isMemberOf(StringUtil::deserialize($objCalendar->groups))))
	{
		continue;
	}

	// …
}
```

### DIFF
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -63,29 +63,23 @@ abstract class Events extends Module
 			return $arrCalendars;
 		}
 
-		$this->import(FrontendUser::class, 'User');
 		$objCalendar = CalendarModel::findMultipleByIds($arrCalendars);
 		$arrCalendars = array();
 
 		if ($objCalendar !== null)
 		{
-			$blnFeUserLoggedIn = System::getContainer()->get('contao.security.token_checker')->hasFrontendUser();
+			$user = null;
+
+			if (System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
+			{
+				$user = FrontendUser::getInstance();
+			}
 
 			while ($objCalendar->next())
 			{
-				if ($objCalendar->protected)
+				if ($objCalendar->protected && (!$user || !$user->isMemberOf(StringUtil::deserialize($objCalendar->groups))))
 				{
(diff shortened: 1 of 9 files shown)


Toflar (approved):

aschempp (changes requested):
Shouldn't we have a security voter instead of a new ~static~ method?

leofeyer:
Where do you see a new static method?

aschempp (approved):
Sorry, I didn't realize the `isMemberOf` method already exists 😆

ausi (approved):

fritzmg (approved):

fritzmg:
Does this actually solve #2753? I don't see any changes regarding that, but I suppose I am missing something?

leofeyer:
Yes, it does. The changes are in the `Backend` class.

PR #2835, opened 2021-03-04
Fix an illegal object access in the Versions class

### LINKED ISSUE #2832: Illegal object access in Versions#getEditUrl()
**Affected version(s)**

Contao 4.11 + PHP8

**Description**

`Warning: Attempt to read property "id" on null` appears here when accessing the user profile in the backend:
https://github.com/contao/contao/blob/fb3c730abf2ae8a1be948e2f87be7c7ddcbb285b/core-bundle/src/Resources/contao/classes/Versions.php#L772-L776

Apparently `$this->User` can be `null`.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2832
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -771,6 +771,8 @@ class Versions extends Controller
 		// Adjust the URL of the "personal data" module (see #7987)
 		if (preg_match('/do=login(&|$)/', $strUrl))
 		{
+			$this->import(BackendUser::class, 'User');
+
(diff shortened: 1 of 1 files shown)


bytehead (approved):

aschempp:
The code example in the issue is not the same as this PR? I don't think that problem still exists in the current code, so the import is useless?

fritzmg (approved):

bytehead:
@aschempp it covers exactly [these lines](https://github.com/contao/contao/blob/fb3c730abf2ae8a1be948e2f87be7c7ddcbb285b/core-bundle/src/Resources/contao/classes/Versions.php#L772-L776) ... 🤔

PR #2837, opened 2021-03-04
Use eval.context in the picker if set

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/madeyourday/contao-rocksolid-custom-elements/issues/129

For dynamic DCA fields (like in multicolumn wizard, custom elements and so on) there is no relation information stored in the DCA extractor. The picker widget supports setting `eval.context` to something like `dc.tl_content`, but when rendering the values this `context` is ignored. This PR changes that so that `context` is always used and the relation data as fallback.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -174,11 +174,18 @@ class Picker extends Widget
 
 	protected function generateValues($blnHasOrder): array
 	{
-		$strRelatedTable = $this->getRelatedTable();
+		if (substr($this->context ?? '', 0, 3) === 'dc.')
+		{
+			$strRelatedTable = substr($this->context, 3);
+		}
+		else
+		{
+			$strRelatedTable = $this->getRelatedTable();
+		}
 
 		if (!$strRelatedTable)
 		{
-			return (array) $this->varValue;
+			return array_combine((array) $this->varValue, (array) $this->varValue);
 		}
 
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/widgets/Picker.php
```diff
@@ -174,11 +174,18 @@ public function generate()
...
+		}
+		else
+		{
+			$strRelatedTable = $this->getRelatedTable();
+		}
 
 		if (!$strRelatedTable)
 		{
-			return (array) $this->varValue;
+			return array_combine((array) $this->varValue, (array) $this->varValue);
```
ausi, line 188: I’m not sure if this change is correct. But the widget seems to be expecting the values as keys. Without this change a selected value does not get stored because the key is `0` instead of the actual selected ID value.

leofeyer, line 188: Is this correct?

leofeyer: If `$this->varValue` is an array of IDs, you are right. Not sure for UUIDs though.

ausi: Looking at like 121, I’m pretty sure the array keys are meant to be the “raw values” and the array values are the labels shown to the user.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2845, opened 2021-03-04
Always show all errors in the contao-setup binary

### LINKED ISSUE #2844: post-update-cmd fails with error code 255 (4.11)
When upgrading managed edition 4.10 to 4.11, `composer update` fails with `Script @php vendor/bin/contao-setup handling the post-update-cmd event returned with error code 255`:

```console
...
Generating autoload files
> post-autoload-dump: PackageVersions\Installer->dumpVersionsClass
composer/package-versions-deprecated: Generating version class...
composer/package-versions-deprecated: ...done generating version class
85 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
> post-update-cmd: ContaoCommunityAlliance\Composer\Plugin\Plugin->dumpRunonce
> post-update-cmd: Contao\ManagerPlugin\Composer\ManagerPluginInstaller->dumpPlugins
contao/manager-plugin: Generating plugin class...
 - Added plugin for contao/calendar-bundle
 - Added plugin for contao/comments-bundle
 - Added plugin for contao/core-bundle
 - Added plugin for contao/faq-bundle
 - Added plugin for contao/installation-bundle
 - Added plugin for contao/listing-bundle
 - Added plugin for contao/manager-bundle
 - Added plugin for contao/news-bundle
 - Added plugin for terminal42/contao-fineuploader
 - Added plugin for terminal42/contao-folderpage
 - Added plugin for terminal42/contao-url-rewrite
 - Added plugin for terminal42/dc_multilingual
contao/manager-plugin: ...done generating plugin class
> post-update-cmd: Google\Task\Composer::cleanup
> post-update-cmd: @php vendor/bin/contao-setup
Executing command (CWD): '/Applications/MAMP/bin/php/php7.3.9/bin/php' -d allow_url_fopen='1' -d disable_functions='' -d memory_limit='-1' vendor/bin/contao-setup
Script @php vendor/bin/contao-setup handling the post-update-cmd event returned with error code 255
```

This does not happen on a fresh install (i.e. `composer create-project contao/managed-edition folder 4.11`)

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2844
| Docs PR or issue | -

We should IMHO be verbose with errors happening in the `contao-setup` binary. By default. As #2844 nicely demonstrates, this could otherwise quickly lead to a bad debugging experience. 

btw.:This is something else than the errors from the sub processes (from within the command) that we're already delegating.

### DIFF
--- a/manager-bundle/bin/contao-setup
+++ b/manager-bundle/bin/contao-setup
@@ -16,7 +16,10 @@ use Contao\ManagerBundle\HttpKernel\ContaoKernel;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Filesystem\Filesystem;
 
+error_reporting(-1);
 set_time_limit(0);
+@ini_set('display_startup_errors', '1');
+@ini_set('display_errors', '1');
 @ini_set('zlib.output_compression', '0');
 
 if (file_exists(__DIR__.'/../autoload.php')) {


fritzmg (approved):

bytehead (approved):

Toflar (approved):

aschempp (commented):
I'm not sure I like this change...

First of all, this does not "fix" the mentioned issue, it merely allows the issue in #2844 to be seen. Also, this only helps "by accident", if you use the mentioned binary. If one uses the `contao-console` to rebuild the cache, exactly the same issue would happen but it won't be visible by this change.

Also, it reminds me that we discussed the `contao-setup` binary shouldn't use the kernel, but I don't remember what the agreement was on that thinkering 🙃 if that was the case, the `contao-setup` binary would not have failed – but the error would just happen in the delegated commands that still won't show it …

m-vo:
> First of all, this does not "fix" the mentioned issue

[It](https://github.com/contao/contao/issues/2844#issuecomment-790995975) [does](https://github.com/contao/contao/issues/2844#issuecomment-791001138)? :smiley_cat: 

> If one uses the contao-console to rebuild the cache, exactly the same issue would happen but it won't be visible by this change.

That's right. IMHO we should add these lines to to our `contao-console` binary as well. Wdyt?

> Also, it reminds me that we discussed the contao-setup binary shouldn't use the kernel, but I don't remember what the agreement was on that thinkering 

Even if we would not use the framework/container in the binary nor the thing that executes the commands, running the first command would build the container with our kernel anyway.

leofeyer:
As discussed in Mumble on March 11th, we want to merge the PR as is and see if and which problems occur.

leofeyer:
Thank you @m-vo.

aschempp:
Here's what I meant by _this does not really fix the issue_ 😉 

https://community.contao.org/de/showthread.php?80487-Contao-Manager-1-4-2-–-Abgesicherter-Modus-lässt-sich-nicht-abschalten&p=541628&viewfull=1#post541628

fritzmg:
But that just proves we definitely should show the errors, even for `contao-console`?

PR #2846, opened 2021-03-05
Upgrade simplify/easy-coding-standard to version 9

leofeyer:
See contao/easy-coding-standard#7


bytehead (approved):

leofeyer:
Actually, we do not even need this PR. Composer will use version 1.3 automatically. 😄

PR #2628, opened 2021-01-08
Use the token checker instead of FE_USER_LOGGED_IN constant

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2576
| Docs PR or issue | -

This replaces the usage of the `FE_USER_LOGGED_IN` constant with the `contao.security.token_checker` service as discussed.

PR #2848, opened 2021-03-07
Fix PHP 8 warnings

mroi:
I have upgraded my Contao installation to PHP 8 and got warnings during regular operations. I have fixed all that I have seen so far. For some of my fixes, more seasoned Contao developers should probably check:
* whether this is the right way of fixing the issue
* if this warning condition should actually occur or if this points to some problem with my installation.

I have targeted this PR to the 4.x branch, since this is not strictly fixing a bug. Please tell me, if I should target 4.11 instead and I will gladly adjust.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -71,7 +71,7 @@ abstract class Controller extends System
 			/** @var PageModel $objPage */
 			global $objPage;
 
-			if ($objPage->templateGroup)
+			if ($objPage && $objPage->templateGroup)
 			{
 				if (Validator::isInsecurePath($objPage->templateGroup))
 				{
--- a/core-bundle/src/Resources/contao/library/Contao/Date.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Date.php
@@ -638,7 +638,7 @@ class Date
 
 		foreach ($chunks as $chunk)
 		{
-			list($index, $flag) = explode('::', $chunk);
+			list($index, $flag) = explode('::', $chunk) + array(null, null);
 
 			switch ($flag)
 			{
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -170,7 +170,7 @@ class InsertTags extends Controller
 
 				// Accessibility tags
 				case 'lang':
-					if (!$elements[1])
+					if (empty($elements[1]))
 					{
 						$arrCache[$strTag] = '</span>';
 					}
@@ -189,7 +189,7 @@ class InsertTags extends Controller
 				case 'email':
 				case 'email_open':
 				case 'email_url':
-					if (!$elements[1])
+					if (empty($elements[1]))
 					{
 						$arrCache[$strTag] = '';
 						break;
@@ -742,7 +742,7 @@ class InsertTags extends Controller
 				case 'ua':
 					$ua = Environment::get('agent');
 
-					if ($elements[1])
+					if (!empty($elements[1]))
 					{
 						$arrCache[$strTag] = $ua->{$elements[1]};
 					}
@@ -755,7 +755,7 @@ class InsertTags extends Controller
 				// Abbreviations
 				case 'abbr':
 				case 'acronym':
-					if ($elements[1])
+					if (!empty($elements[1]))
 					{
 						$arrCache[$strTag] = '<abbr title="' . StringUtil::specialchars($elements[1]) . '">';
 					}
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -55,7 +55,7 @@ class Search
 		$arrSet['url'] = $arrData['url'];
 		$arrSet['title'] = $arrData['title'];
 		$arrSet['protected'] = $arrData['protected'];
-		$arrSet['filesize'] = $arrData['filesize'];
+		$arrSet['filesize'] = $arrData['filesize'] ?? null;
 		$arrSet['groups'] = $arrData['groups'];
 		$arrSet['pid'] = $arrData['pid'];
 		$arrSet['language'] = $arrData['language'];
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -154,6 +154,8 @@ class Search
 			$arrData['description'] = trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($tags[1])));
 		}
 
+		$arrData['keywords'] = '';
+
 		// Get the keywords
 		if (preg_match('/<meta[^>]+name="keywords"[^>]+content="([^"]*)"[^>]*>/i', $strHead, $tags))
 		{
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -159,7 +159,7 @@ class Search
 		// Get the keywords
 		if (preg_match('/<meta[^>]+name="keywords"[^>]+content="([^"]*)"[^>]*>/i', $strHead, $tags))
 		{
-			$arrData['keywords'] = trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($tags[1])));
+			$arrData['keywords'] .= trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($tags[1])));
 		}
 
 		// Read the title and alt attributes
--- a/core-bundle/src/Resources/contao/modules/ModuleArticle.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleArticle.php
@@ -121,7 +121,7 @@ class ModuleArticle extends Module
 
 		// Add the modification date
 		$this->Template->timestamp = $this->tstamp;
-		$this->Template->date = Date::parse($objPage->datimFormat, $this->tstamp);
+		$this->Template->date = Date::parse($objPage ? $objPage->datimFormat : null, $this->tstamp);
 
 		// Clean the RTE output
 		$this->teaser = StringUtil::toHtml5($this->teaser);
--- a/news-bundle/src/Resources/contao/modules/ModuleNews.php
+++ b/news-bundle/src/Resources/contao/modules/ModuleNews.php
@@ -148,12 +148,12 @@ abstract class ModuleNews extends Module
 		$arrMeta = $this->getMetaFields($objArticle);
 
 		// Add the meta information
-		$objTemplate->date = $arrMeta['date'];
+		$objTemplate->date = $arrMeta['date'] ?? null;
 		$objTemplate->hasMetaFields = !empty($arrMeta);
 		$objTemplate->numberOfComments = $arrMeta['ccount'] ?? null;
 		$objTemplate->commentCount = $arrMeta['comments'] ?? null;
 		$objTemplate->timestamp = $objArticle->date;
-		$objTemplate->author = $arrMeta['author'];
+		$objTemplate->author = $arrMeta['author'] ?? null;
 		$objTemplate->datetime = date('Y-m-d\TH:i:sP', $objArticle->date);
 		$objTemplate->addImage = false;
 		$objTemplate->addBefore = false;
--- a/core-bundle/src/Command/DebugPagesCommand.php
+++ b/core-bundle/src/Command/DebugPagesCommand.php
@@ -1,0 +1,129 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Command;
+
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\Routing\Page\DynamicRouteInterface;
+use Contao\CoreBundle\Routing\Page\PageRegistry;
+use Contao\CoreBundle\Routing\Page\RouteConfig;
+use Contao\PageModel;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class DebugPagesCommand extends Command
+{
+    protected static $defaultName = 'debug:pages';
+
+    /**
+     * @var ContaoFramework
+     */
+    private $framework;
+
+    /**
+     * @var PageRegistry
+     */
+    private $pageRegistry;
+
+    /**
+     * @var array<RouteConfig>
+     */
+    private $routeConfigs = [];
+
+    /**
+     * @var array<DynamicRouteInterface>
+     */
+    private $routeEnhancers = [];
+
+    public function __construct(ContaoFramework $framework, PageRegistry $pageRegistry)
+    {
+        parent::__construct();
+
+        $this->framework = $framework;
+        $this->pageRegistry = $pageRegistry;
+    }
+
+    public function add(string $type, RouteConfig $config, DynamicRouteInterface $routeEnhancer = null): void
+    {
+        $this->routeConfigs[$type] = $config;
+
+        if (null !== $routeEnhancer) {
+            $this->routeEnhancers[$type] = $routeEnhancer;
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this->setDescription('Displays the page controller configuration.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->framework->initialize();
+
+        $io = new SymfonyStyle($input, $output);
+
+        $rows = [];
+        $types = array_keys($GLOBALS['TL_PTY']);
+        natsort($types);
+
+        foreach ($types as $type) {
+            $config = $this->routeConfigs[$type] ?? null;
+
+            $page = new PageModel();
+            $page->type = $type;
+
+            $rows[] = [
+                $type,
+                $config ? $config->getPath() : '*',
+                $config ? $config->getUrlSuffix() : '*',
+                $this->pageRegistry->supportsContentComposition($page) ? 'yes' : 'no',
+                isset($this->routeEnhancers[$type]) ? \get_class($this->routeEnhancers[$type]) : '-',
+                $config ? $this->generateArray($config->getRequirements()) : '-',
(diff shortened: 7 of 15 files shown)


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -71,7 +71,7 @@ public static function getTemplate($strTemplate)
 			/** @var PageModel $objPage */
 			global $objPage;
 
-			if ($objPage->templateGroup)
+			if ($objPage && $objPage->templateGroup)
```
mroi, line 74: This change is where I am most uncertain. Can `$objPage` legally be `null` here or is this a problem with my setup?

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -55,7 +55,7 @@ public static function indexPage($arrData)
 		$arrSet['url'] = $arrData['url'];
 		$arrSet['title'] = $arrData['title'];
 		$arrSet['protected'] = $arrData['protected'];
-		$arrSet['filesize'] = $arrData['filesize'];
+		$arrSet['filesize'] = $arrData['filesize'] ?? 0;
```
m-vo: Should probably default to `null` for BC:
```suggestion
		$arrSet['filesize'] = $arrData['filesize'] ?? null;
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -170,7 +170,7 @@ protected function doReplace($strBuffer, $blnCache)
 
 				// Accessibility tags
 				case 'lang':
-					if (!$elements[1])
+					if (empty($elements[1]))
```
m-vo: IMHO this should be `$elements[1] ?? null` everywhere, so that the illegal array access is prevented but the logic stays the same. (And because using `empty` ist kind of an antipattern for nearly everything. :slightly_smiling_face:.)

core-bundle/src/Resources/contao/library/Contao/Controller.php
m-vo: When does this occur in your setup?

m-vo (changes requested):
LGTM in general. This should be targeted at 4.11 as it *is* a bugfix. :upside_down_face:

mroi:
Updated to target the `4.11` branch.

mroi:
Updated to include suggested changes.

mroi:
It looks like I made a mess with my branches. Sorry for the noise.

core-bundle/src/Resources/contao/library/Contao/Controller.php
mroi: I saw it in my log, but did not see an obvious context. I’ll try to reproduce this, please stand by.

core-bundle/src/Resources/contao/modules/ModuleArticle.php
```diff
@@ -121,7 +121,7 @@ protected function compile()
 
 		// Add the modification date
 		$this->Template->timestamp = $this->tstamp;
-		$this->Template->date = Date::parse($objPage->datimFormat, $this->tstamp);
+		$this->Template->date = Date::parse($objPage ? $objPage->datimFormat : null, $this->tstamp);
```
aschempp, line 124: I don't think it's valid to not have a an `$objPage` in this case. Dunno why it was missing though.

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -148,12 +148,12 @@ protected function parseArticle($objArticle, $blnAddArchive=false, $strClass='',
 		$arrMeta = $this->getMetaFields($objArticle);
```
aschempp: Looks like `getMetaFields` should always return all properties (`date` and `author` in your case) as empty strings, instead of checking for null-values.

core-bundle/src/Resources/contao/library/Contao/Date.php
```diff
@@ -638,7 +638,7 @@ protected static function resolveCustomModifiers($strDate)
 
 		foreach ($chunks as $chunk)
 		{
-			list($index, $flag) = explode('::', $chunk);
+			list($index, $flag) = array_pad(explode('::', $chunk), 2, null);
```
aschempp: If I understand correctly, we should probably check for `::` in the `$chunk` and if none is present just add an early-return (add `$chunk` to `$strReturn` and `continue`)?

aschempp (changes requested):

core-bundle/src/Resources/contao/library/Contao/Controller.php
mroi: From studying the logs, it happens while the daily cron jobs are executed. `$strTemplate` is set to `mod_article`. Maybe that rings a bell for someone? Otherwise, I’ll add more instrumentation.

fritzmg: Do you know which cronjob causes that? You can also look at `tl_cron_job` to see which cronjobs are getting executed in your system and when.

mroi: I looked in that table and when the log messages appeared, the minutely, hourly and daily jobs executed (the weekly or monthly ones did *not* run). I then re-ran the minutely and hourly ones and nothing happened in the log. I therefore suspect the daily jobs.

news-bundle/src/Resources/contao/modules/ModuleNews.php
mroi: I pushed a new commit. Do you mean like this? Wouldn’t this be a behaviour change, since templates might treat empty string authors and `null` authors differently?

core-bundle/src/Resources/contao/modules/ModuleArticle.php
mroi: I suspect this has the same root cause as what’s discussed above for `Controller.php`. The respective log messages appear together.

core-bundle/src/Resources/contao/library/Contao/Controller.php
fritzmg: In theory this should never happen though. Even if a cron job renders some templates (for instance `News::generateFeeds` for the content elements of a news entry), `TL_MODE` shouldn't be defined there. You would need to debug which cron job causes this and  figure out why `TL_MODE` is set to `FE`.

mroi: So here’s what happens:
* your hunch was right, `generateFeeds` is called
* I have configured the feed description using an insert tag: `{{insert_article::…}}` (Aha!)
* this invokes `Controller::getArticle` and through various layers of template engine …
* … ends up here with `$objPage` being `null`

Unfortunately, I don’t know the Contao architecture well enough to suggest a course of action. 😞

core-bundle/src/Resources/contao/library/Contao/Date.php
mroi: I pushed a new commit. But is this really better? There is code duplication with the default case of the switch now. I guess it’s a matter of taste.

core-bundle/src/Resources/contao/library/Contao/Controller.php
fritzmg: I see, thank you for debugging :).

@contao/developers I am also not sure what the best course of action is here. On the one hand I think we should replace these instances of

```php
if (\defined('TL_MODE') && TL_MODE == 'FE')
```

with

```php

if ($request && System::getContainer()->get('contao.routing.scope_matcher')->isFrontendRequest($request))
```

similar to what we already did in https://github.com/contao/contao/pull/2628. However, an additional check for `null !== $objPage` would certainly not hurt too - but I am not sure this would really be the best way.

core-bundle/src/Resources/contao/library/Contao/Date.php
```diff
@@ -638,6 +638,12 @@ protected static function resolveCustomModifiers($strDate)
 
 		foreach ($chunks as $chunk)
 		{
+			if (strpos($chunk, '::') === false)
+			{
+				$strReturn .= $chunk;
+				continue;
+			}
+
 			list($index, $flag) = explode('::', $chunk);
```
leofeyer:
```suggestion
			list($index, $flag) = explode('::', $chunk) + array(null, null);
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -170,7 +170,7 @@ protected function doReplace($strBuffer, $blnCache)
 
 				// Accessibility tags
 				case 'lang':
-					if (!$elements[1])
+					if (!($elements[1] ?? null))
```
leofeyer:
```suggestion
					if (empty($elements[1]))
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -189,7 +189,7 @@ protected function doReplace($strBuffer, $blnCache)
 				case 'email':
 				case 'email_open':
 				case 'email_url':
-					if (!$elements[1])
+					if (!($elements[1] ?? null))
```
leofeyer:
```suggestion
					if (empty($elements[1]))
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -742,7 +742,7 @@ protected function doReplace($strBuffer, $blnCache)
 				case 'ua':
 					$ua = Environment::get('agent');
 
-					if ($elements[1])
+					if ($elements[1] ?? null)
```
leofeyer:
```suggestion
					if (!empty($elements[1]))
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -755,7 +755,7 @@ protected function doReplace($strBuffer, $blnCache)
 				// Abbreviations
 				case 'abbr':
 				case 'acronym':
-					if ($elements[1])
+					if ($elements[1] ?? null)
```
leofeyer:
```suggestion
					if (!empty($elements[1]))
```

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -159,6 +159,10 @@ public static function indexPage($arrData)
 		{
 			$arrData['keywords'] = trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($tags[1])));
 		}
+		else
+		{
+			$arrData['keywords'] = '';
+		}
```
leofeyer: We should probably initialize this around line 61, shouldn't we?

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -280,16 +280,16 @@ protected function getMetaFields($objArticle)
 	{
 		$meta = StringUtil::deserialize($this->news_metaFields);
 
+		$return = array('date' => '', 'author' => '');
```
leofeyer: I don't think that this is correct. What problem does it fix?

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: How would replacing the instances fix the issue?

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/Date.php
mroi: I fixed this earlier using `array_pad`, which makes explicit what you did manually. But I guess both are fine.

core-bundle/src/Resources/contao/library/Contao/Search.php
mroi: But lines 54 to 61 initialize `arrSet`. Setting `$arrData['keywords']` there would feel out-of-place to me.

I would alternatively suggest fixing it by `?? null` in line 174. Right now, this is the only place, where `$arrData['keywords']` is read.

news-bundle/src/Resources/contao/modules/ModuleNews.php
mroi: This was suggested in an earlier review comment and I [pushed an additional commit](https://github.com/contao/contao/pull/2848/commits/bf5a1da42c610a9085fb9348759e8adfea9d073b) as a result. Previously, I was fixing the problem with `?? null` in lines 151 and 156. Do you prefer that?

core-bundle/src/Resources/contao/library/Contao/Search.php
leofeyer: Good point. How about this:

```php
$arrData['keywords'] = '';

// Get the keywords
if (preg_match('/<meta[^>]+name="keywords"[^>]+content="([^"]*)"[^>]*>/i', $strHead, $tags))
{
	$arrData['keywords'] .= trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($tags[1])));
}

// Read the title and alt attributes
if (preg_match_all('/<* (title|alt)="([^"]*)"[^>]*>/i', $strBody, $tags))
{
	$arrData['keywords'] .= ' ' . implode(', ', array_unique($tags[2]));
}
```

news-bundle/src/Resources/contao/modules/ModuleNews.php
leofeyer: Yes.

core-bundle/src/Resources/contao/library/Contao/Search.php
mroi: Done as suggested.

news-bundle/src/Resources/contao/modules/ModuleNews.php
mroi: Done. I removed the review commit and cleaned up my commits a bit.

core-bundle/src/Resources/contao/library/Contao/Date.php
mroi: Solved differently based on review by @leofeyer.

news-bundle/src/Resources/contao/modules/ModuleNews.php
mroi: Solved differently based on review by @leofeyer.

core-bundle/src/Resources/contao/library/Contao/Controller.php
mroi: I also don’t see the connection. From my very cursory understanding of Contao’s internals, if `getArticle` is invoked because on an insert tag, it is somewhat correct that `$objPage` is `null`.

leofeyer:
@mroi I have incorporated the uncontroversial changes of this PR in #2911 so we can merge them already. The `$objPage` related changes need to be discussed.

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer, line 74:
```suggestion
			if ($objPage->templateGroup ?? null)
```

leofeyer (changes requested):

core-bundle/src/Resources/contao/modules/ModuleArticle.php
leofeyer: As discussed in Mumble on April 8th, we have to revert this change and make sure to set `$objPage` here: 

https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/news-bundle/src/Resources/contao/classes/News.php#L157

@fritzmg will create a PR to fix this.

fritzmg: Done: https://github.com/contao/contao/pull/2946

mroi:
I have looked at #2946. This merged PR fixes the warning we are discussing whenever an insert tag is used in teaser texts of articles. When those articles are used in a calendar or news feed, an `$objPage` is now present to avoid the warning.

However, in my case I am using an insert tag in the feed description. In this setup, I still see the warning:
```
Attempt to read property "datimFormat" on null
```

Since this is now much more specific than the original topic of this PR, should I close this and open a new PR?

fritzmg:
@mroi this issue should not occur anymore. Which Contao version are you using?

mroi:
I’m running 4.11.5. I have checked that #2946 is in the files I have on the web server. I understood this PR such that it sets up a page context for each feed entry based on the `jumpTo` for that entry? But I use insert tags in the feed description. I guess there  is still no `$objPage` there?

fritzmg:
Oh, in the _feed's_ description. Which insert tag are you using there?

mroi:
> Oh, in the feed's description.

Yes, precisely. I’m using an `{{insert_article::*}}` insert tag there. Functionally it is working fine, I just get the known warning.

fritzmg:
If you use `{{insert_article::*}}` in the feed's description, then this means your feed's description will have HTML tags. I am not sure this is supported by the RSS standard?

mroi:
You are probably right there and I should be using `{{insert_article::*|specialchars}}` instead. I just checked and this causes the same warning.

fritzmg:
> You are probably right there and I should be using `{{insert_article::*|specialchars}}` instead.

Well no, that would just convert `<` etc. to their HTML entities which is certainly not what you want ;)

mroi:
I know, I know. My point was just that it’s possible to use an insert tag inside the feed description and generate a valid RSS feed. But if the verdict is that it’s not worth constructing a gnarly fix for a warning in such a fringe use case, I would be fine with that decision.

m-vo:
> My point was just that it’s possible to use an insert tag inside the feed description

IMHO it's very unfortunate that you can enter _any_ insert tag in any field. Don't expect it to work though.

But even if the content does not make sense it should not produce a PHP warning. :+1:

fritzmg:
Insert tags like `{{insert_article::*}}`, which enable you to insert a _page_ article anywhere, are imho only supposed to work in a _page_ context and not necessarily outside of it. If you want to store text resources to be used in different contexts, do not use _page_ articles for that. Build your own solution (your own DCA) or use extensions likes `terminal42/contao-node`.

leofeyer:
The PR is now targeted against an EOL version of Contao. It seems that most of the issues have been fixed in #2911 and #2946, though.

Does the problem with `$objPage` being `null` still exist in Contao 4.12?

fritzmg:
> Does the problem with `$objPage` being `null` still exist in Contao 4.12?

Yes, but I would consider the instance described above as known limitation.

leofeyer:
You are referring to this, right?

> Insert tags like `{{insert_article::*}}`, which enable you to insert a _page_ article anywhere, are imho only supposed to work in a _page_ context and not necessarily outside of it. If you want to store text resources to be used in different contexts, do not use _page_ articles for that. Build your own solution (your own DCA) or use extensions likes `terminal42/contao-node`.

If so, then I agree.

mroi:
The remaining issue has a different root cause from what the PR originally was about. As the original requestor, I would be fine with closing this.

PR #2851, opened 2021-03-08
Pass cssID from module and alias elements to fragments

### LINKED ISSUE #2823: Content include element broken with fragments
**Affected version(s)**

Contao 4.9 or Contao 4.10?

**Description**

If I include a content element which is a fragment controller with the content alias element Contao selects the wrong element class for the rendering. The issue appears because the element id is overridden here: https://github.com/contao/contao/blob/991282912ef0c590b92d25b67785463516a38890/core-bundle/src/Resources/contao/elements/ContentAlias.php#L42

### LINKED ISSUE #2840: cssID not available in Module
**Affected version(s)**

Contao 4.9

**Description**

Die eingegebenen Werte bei CSS-id bzw. Css-Klasse für ein FE-Modul werden nicht im template ausgespielt.

Siehe: https://community.contao.org/de/showthread.php?80427-CSS-des-Inhaltselement-mit-eingeb-Modul-sind-im-Template-nicht-verf%C3%BCgbar

Referenz auf dieses Problem wohl auch hier  https://github.com/contao/contao/issues/2533#issuecomment-782172069

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2840


core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -114,11 +114,16 @@ protected function addHeadlineToTemplate(Template $template, $headline): void
     /**
      * @param string|array $cssID
      */
-    protected function addCssAttributesToTemplate(Template $template, string $templateName, $cssID, array $classes = null): void
+    protected function addCssAttributesToTemplate(Template $template, string $templateName, $cssID, array $classes = null, string $overrideCssID = null): void
```
aschempp, line 117: Are we allowed to add additional (nullable) properties to this method, or would that be a BC break if someone overrides it? Then we might need to go with Symfony's `..., /** string $overrideCssID */` and `func_get_arg()`?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -26,7 +26,7 @@ public function __invoke(Request $request, ContentModel $model, string $section,
         $template = $this->createTemplate($model, 'ce_'.$type);
 
         $this->addHeadlineToTemplate($template, $model->headline);
-        $this->addCssAttributesToTemplate($template, 'ce_'.$type, $model->cssID, $classes);
+        $this->addCssAttributesToTemplate($template, 'ce_'.$type, $model->cssID, $classes, $request->attributes->get('cssID'));
```
dmolineus, line 29: We could replace the third parameter with `$request->attributes->get('cssID', $model->cssID)` so that no extra flag is required. The attribute has to contain the css id and classes of course.

core-bundle/src/Controller/AbstractFragmentController.php
richardhj: A BC break only in PHP 8: https://3v4l.org/jAen0

So, a breaking behavior in the "real world" is unlikely IMHO.

Anyway, this method could have been `private`, no?

fritzmg: It needs to be a `protected` method because you are supposed to be able to use it yourself if you implement your own `__invoke` function for example.

dmolineus:
This pull request does not fix #2823 but #2840.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
richardhj: I like this approach more

aschempp: I intentionally didn't write the serialized/array data into the field. We could still use the fallback value, but I think it would not handle it the same if `cssID` attribute was an empty string …

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -114,11 +114,16 @@ protected function addHeadlineToTemplate(Template $template, $headline): void
...
     {
         $data = StringUtil::deserialize($cssID, true);
         $template->class = trim($templateName.' '.($data[1] ?? ''));
-        $template->cssID = !empty($data[0]) ? ' id="'.$data[0].'"' : '';
+
+        if (!empty($overrideCssID)) {
+            $template->cssID = $overrideCssID;
```
leofeyer: No surrounding `id=""` markup here?

leofeyer:
Are you sure that this PR fixes the issue #2823? Shouldn't we not overwrite the element ID or use `origId` instead?

dmolineus:
> Are you sure that this PR fixes the issue #2823? Shouldn't we not overwrite the element ID or use `origId` instead?

No, it does not fix the issue. You're right, we need to change the handling of the `origId`.

core-bundle/src/Resources/contao/elements/ContentAlias.php
```diff
@@ -38,12 +38,30 @@ public function generate()
...
+		if (is_a($strClass, ContentProxy::class, true))
+		{
+			if (!empty($this->cssID[1]))
+			{
+				$objElement->classes = array_merge((array) $objElement->classes, [$this->cssID[1]]);
+			}
+
+			/** @var ContentProxy $proxy */
+			$proxy = new $strClass($objElement, $this->strColumn);
+
```
dmolineus, line 50:
```suggestion
$proxy->addFragmentAttributes(['typePrefix' => 'ce_']);

```

core-bundle/src/Resources/contao/elements/ContentAlias.php
```diff
@@ -38,12 +38,30 @@ public function generate()
...
+			{
+				$proxy->addFragmentAttributes(['cssID' => $this->cssID[0]]);
+			}
+
+			return $proxy->generate();
+		}
+
 		$objElement->origId = $objElement->origId ?: $objElement->id;
 		$objElement->id = $this->id;
 		$objElement->typePrefix = 'ce_';
```
dmolineus, line 61:
```suggestion
        else
        {
		$objElement->origId = $objElement->origId ?: $objElement->id;
     		$objElement->id = $this->id;
	    	$objElement->typePrefix = 'ce_';
	}
```

core-bundle/src/Resources/contao/elements/ContentAlias.php
dmolineus: Didn't find any places where the type prefix is actually used. Maybe we can drop it completely for fragments.

core-bundle/src/Resources/contao/elements/ContentAlias.php
dmolineus: We may not override the element id for content proxies. Furthermore setting `origId` and `typePrefix` doesn't have any effect for content proxies. Maybe we might set the id of the `ContentAlias` element as fragment attribute as well.

dmolineus (changes requested):
I added some comments which needs to be changed to fix #2823 as well. If I should create a new pull request, let me know.

aschempp:
How does it not fix #2823 if the code that overrides the ID is not executed for fragments at all?

core-bundle/src/Controller/AbstractFragmentController.php
aschempp: good catch, that certainly is a bug!

core-bundle/src/Controller/AbstractFragmentController.php
leofeyer: Since this is an abstract class, which is intended to be extended, we must not change the method signature in a minor or patch version. We have to go with Symfony's `..., /** string $overrideCssID */` and `func_get_arg()`.

core-bundle/src/Resources/contao/elements/ContentAlias.php
aschempp: I think so too, that's only used for Hybrids

core-bundle/src/Resources/contao/elements/ContentAlias.php
aschempp: not sure what you're referring to, there's an early exit in the if-condition above?

core-bundle/src/Controller/AbstractFragmentController.php
aschempp: fixed in 7507e0d42c939eb3dcfe043825bdb4500dd16420

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -114,11 +114,16 @@ protected function addHeadlineToTemplate(Template $template, $headline): void
...
      * @param string|array $cssID
      */
-    protected function addCssAttributesToTemplate(Template $template, string $templateName, $cssID, array $classes = null): void
+    protected function addCssAttributesToTemplate(Template $template, string $templateName, $cssID, array $classes = null, string $overrideCssID = null): void
     {
         $data = StringUtil::deserialize($cssID, true);
         $template->class = trim($templateName.' '.($data[1] ?? ''));
-        $template->cssID = !empty($data[0]) ? ' id="'.$data[0].'"' : '';
+
+        if (!empty($overrideCssID)) {
```
ausi, line 122:
```suggestion
        if (null !== $overrideCssID) {
            $template->cssID = !empty($overrideCssID) ? ' id="'.$overrideCssID.'"' : '';
```
shouldn’t this be a null check so that I can overwrite the CssID with an empty string?

aschempp: I don't think that's intended. We're passing the CSS ID from an alias element to the included element, only if the value is not empty there.

aschempp:
see https://github.com/contao/contao/pull/2930 for an updated PR

Toflar:
I think we should close this, the other PR is way cleaner :)

aschempp:
Closed in favour of https://github.com/contao/contao/pull/2930

PR #2854, opened 2021-03-08
Respect the label field order in the picker

### LINKED ISSUE #2853: Universal picker renders label fields in wrong order
**Affected version(s)**

Contao 4.9

**Description**

Hello, I use the Universal Picker in a custom data container array. I use it with two label fields and a custom format:

```php
// tl_my_dca.php
// ...
'fields' => ['title', 'type'],
'format' => '%s <span class="tl_gray">(%s)</span>'
// ...
```

The universal picker however applies the fields to the placeholders in reverse order. 

The formatting should be the other way around: `local` is the `type` and should be in light gray.

This happens due to the [flipping and intersecting in the widget](https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/widgets/Picker.php#L265): After `array_intersect_key` the resulting array contains the label field keys in the order of appearance in `$arrRow`, rather than the order provided by my dca config. 

I'll look into this and provide a fix 👍

bezin:
| Q                | A
| -----------------| ---
| Fixed issues     | #2853
| Docs PR or issue | -

Maintains the order of the keys in the picker label generator function.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -265,7 +265,7 @@ class Picker extends Widget
 		$labelValues = [];
 
 		foreach ($labelConfig['fields'] as $field) {
-			$labelValues[] = $arrRow[$field];
+			$labelValues[] = $arrRow[$field] ?? '';
 		}
 
 		$label = vsprintf($labelConfig['format'], $labelValues);
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -262,9 +262,10 @@ class Picker extends Widget
 		}
 
 		$labelConfig = &$GLOBALS['TL_DCA'][$dc->table]['list']['label'];
-		$labelValues = [];
+		$labelValues = array();
 
-		foreach ($labelConfig['fields'] as $field) {
+		foreach ($labelConfig['fields'] as $field)
+		{
 			$labelValues[] = $arrRow[$field] ?? '';
 		}
 
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -262,7 +262,13 @@ class Picker extends Widget
 		}
 
 		$labelConfig = &$GLOBALS['TL_DCA'][$dc->table]['list']['label'];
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/widgets/Picker.php
```diff
@@ -262,7 +262,13 @@ protected function renderLabel(array $arrRow, DataContainer $dc)
 		}
 
 		$labelConfig = &$GLOBALS['TL_DCA'][$dc->table]['list']['label'];
-		$label = vsprintf($labelConfig['format'], array_intersect_key($arrRow, array_flip($labelConfig['fields'])));
+		$labelValues = [];
+
+		foreach ($labelConfig['fields'] as $field) {
+			$labelValues[] = $arrRow[$field];
```
ausi: Could the field be missing?

```suggestion
			$labelValues[] = $arrRow[$field] ?? '';
```

core-bundle/src/Resources/contao/widgets/Picker.php
```diff
@@ -262,7 +262,13 @@ protected function renderLabel(array $arrRow, DataContainer $dc)
 		}
 
 		$labelConfig = &$GLOBALS['TL_DCA'][$dc->table]['list']['label'];
```
ausi, line 271:
```suggestion
		$labelConfig = $GLOBALS['TL_DCA'][$dc->table]['list']['label'];
```

I think this reference needs to be removed, but this might be out of scope for this PR :)

core-bundle/src/Resources/contao/widgets/Picker.php
bezin: Yes, it could. However, the previous implementation did not care for this either so I kinda went along with it 😂 
But I will improve this!

bezin:
What's with the code style errors? Is this because the widget classes are legacy and that is why I should use `array()` rather than `[]` and have the curly brace on a new line for the `foreach` statement?

core-bundle/src/Resources/contao/widgets/Picker.php
bezin: Yeah, I was wondering the same but yet I felt I should not touch this ;-)

ausi:
Yes, you can run `composer run cs-fixer` locally to fix these issues automatically.

ausi (approved):

leofeyer:
Thank you @bezin.

Issue #2608, opened 2020-12-28
Search engine: to less characters => Error


math-GH:
**Affected version(s)**

Contao 4.9.10

**Description**
When I search for a "too short" string, than this error appears:
![grafik](https://user-images.githubusercontent.com/1645099/103244710-2b787480-495e-11eb-8835-bbe831278fba.png)

"too short" examples: "a" or "ab" or "abc" (shorter then minimum length)
![grafik](https://user-images.githubusercontent.com/1645099/103244816-7a260e80-495e-11eb-9ef2-907546369a69.png)

"abcd" (minimum length): it works fine

ausi:
To fix this syntax error we should change https://github.com/contao/contao/blob/918884ce919f753e8e7fd68c76396bf92cb5703d/core-bundle/src/Resources/contao/library/Contao/Search.php#L461 to something like this I think: 

```php
$strQuery .= " FROM (SELECT word FROM tl_search_index WHERE (" . (implode(' OR ', $arrAllKeywords) ?: "0") . ") GROUP BY word) words JOIN tl_search_index ON tl_search_index.word = words.word WHERE 1";
```

However the result would still be the same for the end user because no result would be found.

PR #2856, opened 2021-03-08
Fix the search query if there are no keywords

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2657 #2608

Fixes the same issue as #2657 but for Contao 4.11


bytehead (approved):
LGTM, works in my case.

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #2857, opened 2021-03-08
Fix deprecations in unit tests

ausi:
This should fix the CI chain. Not sure if we should add `TODO` comments so that we don’t forget to actually change the code according to the deprecation messages?

Related: https://github.com/contao/contao/pull/2517


ausi:
The CI is green now again 🎉

bytehead (approved):

ausi:
> Not sure if we should add `TODO` comments so that we don’t forget to actually change the code according to the deprecation messages?

Created a seperate issue for the affected classes: #2862

ausi:
Closed in favor of #2870

PR #2863, opened 2021-03-10
Add a PHPUnit extension to test for deprecated classes

ausi:
This is an alternative to #2517

It adds a before-all-tests and an after-all-tests hook to phpunit and tests the deprecated classes before and throws an error after the tests. This has several benefits:

- Test execution order (or using `--filter`) is no longer a problem as all deprecated classes get loaded (and verified) before the first test is executed
- Fixes deprecations for functional tests too, where other methods (like #2517) have problems.
- Single point in the codebase where we can check if we need to actually react to a deprecation when migrating to a new Symfony version for example.

For clarification: This is for file scoped deprecations (= `trigger_error` outside the class definition) that get triggerd only once when the class file gets autoloaded.


m-vo (dismissed):

ausi:
@leofeyer should I create this PR for Contao 4.11 too before we merge this one to verify that the concept works there too?

leofeyer:
I don't think that is necessary.

ausi:
Depends on contao/test-case#11 being merged and released as 4.3.0

m-vo (approved):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2864, opened 2021-03-10
Use the image template everywhere

fritzmg:
The `gallery_default` and `mod_randomImage` templates still do not use the `image` template and instead output

```
<figure><a>…</a><figcaption></figcaption></figure>
```

themselves. There does not seem to be any reason for it as the HTML structure is exactly the same as with the `image` template.

In order to reduce the complexity of these templates and make the HTML output of images more consistent (in cases where you may want to alter the `image` template for example) this PR also uses the `image` template for these templates instead.

The only exception left is `ce_hyperlink` - which needs `embed_pre` and `embed_post` before/after the image itself. Honestly though I have never used the image feature in the hyperlink element and I consider it somewhat of a frankenstein thing anyway 😁 

I have based this against the `4.9` branch as I consider it a bug that if you create a custom `image` template these changes are not automatically reflected in the `gallery` content element or the `randomImage` module. But that's up for discussion of course.

### DIFF
--- a/core-bundle/src/Resources/contao/templates/gallery/gallery_default.html5
+++ b/core-bundle/src/Resources/contao/templates/gallery/gallery_default.html5
@@ -4,16 +4,7 @@
     <?php foreach ($row as $col): ?>
       <?php if ($col->addImage): ?>
         <li class="<?= $class ?> <?= $col->class ?>">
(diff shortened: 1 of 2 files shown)


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #164, opened 2018-11-03
Set shared max age for content element controllers with start or stop date.

richardhj:
See #13.

PR #2867, opened 2021-03-11
Recognize visibility settings in abstract content element controller

### LINKED ISSUE #2533: AbstractContentElementController does not recognize visibility settings
**Affected versions**

Contao `^4.6`

**Description**

Legacy content elements are in order to apply the visibility settings (`invisible`, `start`, `stop`). 

https://github.com/contao/contao/blob/7e6bacd4e334bdd1a32996dab12d3b05e1e161cc/core-bundle/src/Resources/contao/elements/ContentElement.php#L245-L248

The new [abstract controller](https://github.com/contao/contao/blob/4.9/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php) doesn't contain the logic nor the [`ContentProxy`](https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/elements/ContentProxy.php) applies this setting. As a result content elements defined as controllers are always visible.

While discussing the issue on the slack channel the question was raised if the content element should be responsible for it or better be applied from the outside like handling the permission check and the `isElementVisible` hook?

Is the concrete content element (controller) the right place to apply the visibility check or should it be handled by the `Controller::getContentElement()`? If we agree that a content element is responsible for it, a content element might not implement the visibility handling. Otherwise every content element should define the fields in the backend palette.

Depending on the question above this issue could be fixed in the `ContentProxy`, the `AbstractContentElementController` or the `Controller::getContentElement` method.

dmolineus:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2533

This pull requests implements the visibility check as discussed here https://github.com/contao/contao/issues/2533#issuecomment-747503667. It does not include the permission check as proposed in the mentioned comment. It's currently done by the `isVisibleElement` check outside of the controller.

This change is based against the `4.x` branch on purpose. Adding new methods to an abstract controller have to be considered as BC break. So adding it in a bugfix release is unfortunately no option even though it's a bugfix.


fritzmg:
> This change is based against the `4.x` branch on purpose. Adding new methods to an abstract controller have to be considered as BC break.

We already added new methods to the `AbstractFragmentController` in a bugfix release, so I think that would be fine. It doesn't actually break backwards compatibility, does it?

dmolineus:
> It doesn't actually break backwards compatibility, does it?

It will break projects where I implemented the missing visibility check slightely different where the method signature is different.

leofeyer:
> Adding new methods to an abstract controller have to be considered as BC break.

I don't think so. Adding new **abstract** methods to an abstract controller would be a BC break, but new regular methods should be fine.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -36,6 +43,8 @@ public function __invoke(Request $request, ContentModel $model, string $section,
             $response = $template->getResponse();
         }
 
+        $this->addSharedMaxAgeToResponse($response, $model);
```
aschempp, line 46: I don't think this is related to the PR?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -59,5 +68,34 @@ protected function addSharedMaxAgeToResponse(Response $response, ContentModel $m
...
+
+        // The element is visible, so show it
+        if (!$isInvisible) {
+            return true;
+        }
+
+        return $this->isFrontendPreview($request);
+    }
+
+    protected function isFrontendPreview(Request $request): bool
```
aschempp: did you intentionally add this as a separate method because it could be used separately?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -59,5 +68,34 @@ protected function addSharedMaxAgeToResponse(Response $response, ContentModel $m
...
+    {
+        $tokenChecker = $this->get('contao.security.token_checker');
+
+        // Preview mode is enabled, so show the element
+        if ($tokenChecker->hasBackendUser() && $tokenChecker->isPreviewMode()) {
+            return true;
+        }
+
+        // We are in the back end, so show the element
+        if ($this->get('contao.routing.scope_matcher')->isBackendRequest($request)) {
```
aschempp: I would check for this before anything else ("first line" in `isVisible`)

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -59,5 +68,34 @@ protected function addSharedMaxAgeToResponse(Response $response, ContentModel $m
...
+
+        return $this->isFrontendPreview($request);
+    }
+
+    protected function isFrontendPreview(Request $request): bool
+    {
+        $tokenChecker = $this->get('contao.security.token_checker');
+
+        // Preview mode is enabled, so show the element
+        if ($tokenChecker->hasBackendUser() && $tokenChecker->isPreviewMode()) {
```
aschempp, line 88: Does `isPreviewMode` not work without `hasBackendUser()` check first?

aschempp (changes requested):
Generally looks good to me. I noticed the hook does not work anymore, should we add a new event?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
dmolineus: Imo it is. If a fragment controller now decides not to be shown for a time frame it needs to sets the shared max age as well. Otherwise the element might be visible even though the stop time is reached.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
dmolineus: I just followed the proposed solution mentioned here https://github.com/contao/contao/issues/2533#issuecomment-747503667

dmolineus: Removed it in 0d18831.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
dmolineus: I don't know. It used the same way in other places, for instance here: https://github.com/contao/contao/blob/2bbc4fd07bf29f8ad2c3f92853070f1b4d836a9b/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L87

dmolineus:
> > Adding new methods to an abstract controller have to be considered as BC break.
> 
> I don't think so. Adding new **abstract** methods to an abstract controller would be a BC break, but new regular methods should be fine.

The consequence of this approach is that I'm never safe to run a bugfix update if I ever add a custom method in my implementations based on abstract classes Contao provides. Could we at least formulate what backward compatibility means for Contao so that upgrade policies could rely on it? This should be discussed in another issue I guess.

@leofeyer So I need to rebase the PR against 4.9?

leofeyer:
> @leofeyer So I need to rebase the PR against 4.9?

I guess so. @contao/developers Any objections?

Toflar:
I'm fine with considering this a bugfix.

dmolineus:
I rebased the pull request onto 4.9 now.

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
```diff
@@ -25,11 +26,31 @@
...
     {
         parent::setUp();
 
-        System::setContainer($this->getContainerWithContaoConfiguration());
+        $container = $this->getContainerWithContaoConfiguration();
+        $container->set('request_stack', new RequestStack());
+
+        // Prevent that the dca extractor tries to load language file and dca files
+        // See Contao\DcaExtractor::createExtract()
+        $GLOBALS['TL_LANG']['MSC']['foo'] = 'bar';
```
Toflar: This should be reset in `tearDown()` otherwise @m-vo will have more work with cleaning up the test suite again :D

dmolineus: Done with eec3137.

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
```diff
@@ -14,22 +14,56 @@
...
+        $GLOBALS['TL_LANG']['MSC'] = ['foo' => 'bar'];
+
+        System::setContainer($container);
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        unset($GLOBALS['TL_LANG']['MSC']);
```
ausi:
```suggestion
        unset($GLOBALS['TL_LANG']);
```
Otherwise `$GLOBALS['TL_LANG']` would stay as an empty array.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
aschempp: Here's what was discussed when we added the method: https://github.com/contao/contao/pull/164#issuecomment-438633583

ausi:
Do we need to add the `tstamp=0` logic from https://github.com/contao/contao/pull/2994#issuecomment-839763083 here too?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -22,6 +22,13 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
 {
     public function __invoke(Request $request, ContentModel $model, string $section, array $classes = null): Response
     {
+        if (!$this->isVisible($model, $request)) {
```
rabauss, line 25: Good plan for member groups!
But unfortunately you will not get there with set `invisible`, `start`, `stop` because the element will be never loaded previously because of `ContentModel::findPublishedByPidAndTable` - see for example in the article (there are some other places!): https://github.com/contao/contao/blob/37e1a09d99f6bbde827d1e4e7f47ed41341d6bcb/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L189

Especially for a start date in the future we need to find another solution! Because otherwise the shared max age will be never added to the response, so the page cache won't be invalid on the content element start date!
Only for the member groups this will work but not for `invisible`, `start`, `stop`

Toflar: I guess for BC reasons, we need to make a difference between old style elements/modules and new fragments.
New fragments should always be loaded and the response should be generated. In case of an invisible one, it should return an empty response. This would allow to also modify the caching headers for `start` and `stop`.
I'll mark this issue as up for discussion.

aschempp:
As discussed in Mumble today, I see 3 points that should be addressed:
 1. it should be possible for an element to bypass the check, e.g. through a new `shouldCheckVisibility` method. Maybe it can be achieved by overriding the `isVisible` check, e.g. by returning NULL.
 2. The same should be applied for frontend module, so some stuff might need to go in the `AbstractFragmentController`
 3. We must not check `Controller::isVisibleElement` for elements of type proxy (but probably not inside that method but before calling it)

aschempp:
~I also noticed that a protected element should not be cacheable at all, I guess that needs to be checked before tagging the response.~

However, we have a `MakeResponsePrivateListener` 🤦

rabauss:
It's not solved, is it? 
Or did @Toflar implement something in this direction?

dmolineus:
No, afaik it's not solved. It's still on my todo list to complete this PR.

dmolineus:
This pull request will very likely be supoerfluous with #4407.

leofeyer:

PR #2876, opened 2021-03-15
Adjust the Google Analytics strip cookie regex

### LINKED ISSUE #2820: Cookies set via JavaScript cause InvalidRequestTokenException
**Affected version(s)**

Contao 4.9.12+

**Description**

If you visit a page with a form where some JavaScript generates a cookie within the site's domain, the form cannot be sent as the newly set cookie triggers the `InvalidRequestTokenException` - if the previous request had no CSRF token cookie in its response.

**Reproduction**

1. Create a simple test form (`POST`) and integrate in into a page.
2. Integrate a JavaScript like `document.cookie = "foobar=foobar";` into the same page.
3. Clear the cookies for the site's domain.
4. Visit the page containing the form in the front end.
5. Submit the form.

If the initial request did not return `true` for `CsrfTokenCookieSubscriber::requiresCsrf` then the subsequent `POST` request will trigger the `InvalidRequestTokenException`, since the newly, client side set cookie now requires a CSRF token to be present, but none was set before of course.

This is a case reported on the [community forum](https://community.contao.org/de/showthread.php?79555-Ung%C3%BCltiges-Anfrage-Token-mal-wieder&p=540676&viewfull=1#post540676) where the cookie is actually set by the Google Tag Manager for example. So when visiting the page for the first time (or after clearing the cookies), form submission will not work for anyone.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes [#2820](https://github.com/contao/contao/issues/2820#issuecomment-796813028)
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/EventListener/HttpCache/StripCookiesSubscriber.php
+++ b/core-bundle/src/EventListener/HttpCache/StripCookiesSubscriber.php
@@ -30,9 +30,8 @@ class StripCookiesSubscriber implements EventSubscriberInterface
         '(.*)?modal(.*)?',
 
         // Google Analytics (https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage)
-        '_ga',
+        '_ga.*',
         '_gid',
-        '_gat',
         '_dc_gtm_.+',
         'AMP_TOKEN',
         '_gac_.+',


m-vo (approved):

fritzmg (approved):

Toflar (approved):

core-bundle/src/EventListener/HttpCache/StripCookiesSubscriber.php
```diff
@@ -30,9 +30,8 @@ class StripCookiesSubscriber implements EventSubscriberInterface
         '(.*)?modal(.*)?',
 
         // Google Analytics (https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage)
-        '_ga',
+        '_ga.*',
         '_gid',
-        '_gat',
         '_dc_gtm_.+',
         'AMP_TOKEN',
         '_gac_.+',
```
ausi, line 37: This can be removed too I think.

PR #2879, opened 2021-03-16
Make sure test runs are deterministic

ausi:
To prevent random test failures like https://github.com/contao/contao/pull/2801/checks?check_run_id=2119404288

### DIFF
--- a/core-bundle/tests/Routing/RouteProviderTest.php
+++ b/core-bundle/tests/Routing/RouteProviderTest.php
@@ -27,6 +27,15 @@ use Symfony\Component\Routing\Route;
 
 class RouteProviderTest extends TestCase
 {
+    private $pageModelAutoIncrement = 0;
+
+    protected function tearDown(): void
+    {
+        $this->pageModelAutoIncrement = 0;
+
+        parent::setUp();
+    }
+
     public function testGetsARouteByName(): void
     {
         /** @var PageModel&MockObject $page */
@@ -664,7 +673,7 @@ class RouteProviderTest extends TestCase
     {
         /** @var PageModel&MockObject $page */
         $page = $this->mockClassWithProperties(PageModel::class);
-        $page->id = random_int(1, 10000);
+        $page->id = ++$this->pageModelAutoIncrement;
         $page->rootId = 1;
         $page->type = 'regular';
         $page->alias = $alias;
@@ -672,6 +681,7 @@ class RouteProviderTest extends TestCase
         $page->rootLanguage = $language;
         $page->rootIsFallback = $fallback;
         $page->rootUseSSL = 'https' === $scheme;
+        mt_srand($this->pageModelAutoIncrement);
         $page->rootSorting = mt_rand();
 
         return $page;
--- a/core-bundle/tests/Routing/RouteProviderTest.php
+++ b/core-bundle/tests/Routing/RouteProviderTest.php
@@ -31,9 +31,9 @@ class RouteProviderTest extends TestCase
 
     protected function tearDown(): void
     {
-        $this->pageModelAutoIncrement = 0;
+        parent::tearDown();
 
-        parent::setUp();
+        $this->pageModelAutoIncrement = 0;
     }
 
     public function testGetsARouteByName(): void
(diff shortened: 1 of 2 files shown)


core-bundle/tests/Routing/RouteProviderTest.php
```diff
@@ -664,14 +673,15 @@ private function createPage(string $language, string $alias, bool $fallback = tr
     {
         /** @var PageModel&MockObject $page */
         $page = $this->mockClassWithProperties(PageModel::class);
-        $page->id = random_int(1, 10000);
+        $page->id = ++$this->pageModelAutoIncrement;
```
ausi: @aschempp Is it OK to just count upwards here? I’m not sure how this can or should affect the test results.

core-bundle/tests/Routing/RouteProviderTest.php
```diff
@@ -664,14 +673,15 @@ private function createPage(string $language, string $alias, bool $fallback = tr
...
         $page->rootId = 1;
         $page->type = 'regular';
         $page->alias = $alias;
         $page->domain = $domain;
         $page->rootLanguage = $language;
         $page->rootIsFallback = $fallback;
         $page->rootUseSSL = 'https' === $scheme;
+        mt_srand($this->pageModelAutoIncrement);
```
m-vo: Why do we need to seed here?

ausi: So that the `mt_rand()` in the next line is deterministic.

m-vo: Sorry, missed that line completely. :see_no_evil:

leofeyer (dismissed):
Looks good to me. Waiting for @aschempp's answer before merging this.

core-bundle/tests/Routing/RouteProviderTest.php
```diff
@@ -662,9 +671,11 @@ private function mockFramework(Adapter $pageAdapter = null, Adapter $configAdapt
      */
     private function createPage(string $language, string $alias, bool $fallback = true, string $domain = '', string $scheme = null): PageModel
     {
+        mt_srand(++$this->pageModelAutoIncrement);
```
aschempp, line 674: why do we seed the random number generator here? We seed it again in randomBytes?

core-bundle/tests/Routing/RouteProviderTest.php
```diff
@@ -662,9 +671,11 @@ private function mockFramework(Adapter $pageAdapter = null, Adapter $configAdapt
...
+
         /** @var PageModel&MockObject $page */
         $page = $this->mockClassWithProperties(PageModel::class);
-        $page->id = random_int(1, 10000);
+        $page->id = $this->pageModelAutoIncrement;
```
aschempp, line 678: This should be fine, I guess we just need different page IDs on avery `createPage` call. However, this does not really make the tests deterministic, because the page ID will be different depending on the tests run? Maybe we just need to pass an ID to the method or generate it from the arguments?

ausi: > However, this does not really make the tests deterministic, because the page ID will be different depending on the tests run?

Why? The `pageModelAutoIncrement` member gets reset for every single test.

core-bundle/tests/Routing/RouteProviderTest.php
ausi: For the line `$page->rootSorting = mt_rand();` further down. I think @leofeyer moved the line up here in [`a54cea5` (#2879)](https://github.com/contao/contao/pull/2879/commits/a54cea5e9396faaf85fa615c7f7dc23bfcae8bab)

core-bundle/tests/Security/TwoFactor/AuthenticatorTest.php
```diff
@@ -126,4 +126,17 @@ public function testGeneratesTheQrCode(): void
         $this->assertSame(5897, \strlen($svg));
         $this->assertSame(0, strpos($svg, $beginSvg));
     }
+
+    private function randomBytes(int $bytesCount, int $seed): string
```
Toflar: Can we rename this? I know I'm being picky but it's not "random" anymore now. And it doesn't have to be. Let's just name it `generateSecret()` and add a comment about it having to be deterministic for tests or so. Otherwise someone will read this code and go like "wtf?".

ausi: Done in c10dc43c55f7e49aeaf7a8c13deb10bd5d308482

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #2883, opened 2021-03-16
Return a 404 status code if an image file does not exist

### LINKED ISSUE #1995: Missing images do not generate 404 page
**Affected version(s)**

Contao 4.9

**Description**

If I request an image file that does not exist on the server, I get the following in my log: `InvalidArgumentException: "files/example.jpg does not exist" at vendor/contao/image/src/Image.php line 61`

Shouldn't we throw a 404 error instead or convert this somewhere in the deferred image controller?

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1995 
| Docs PR or issue | contao/docs#688

~~Depends on contao/image#79~~

The image controller now returns a 404 status code if the deferred image json exists but the source image file is missing.

The `resize-images` command now has an option `--preserve-missing` as discussed in https://github.com/contao/contao/issues/1995#issuecomment-697168755


core-bundle/src/Command/ResizeImagesCommand.php
```diff
@@ -107,6 +108,7 @@ protected function configure(): void
             ->addOption('throttle', 't', InputOption::VALUE_OPTIONAL, '(Deprecated) Use the concurrent option instead', false)
             ->addOption('image', null, InputArgument::OPTIONAL, 'Image name to resize a single image')
             ->addOption('no-sub-process', null, InputOption::VALUE_NONE, 'Do not start a sub process per resize')
+            ->addOption('delete-missing', null, InputOption::VALUE_NONE, 'Delete deferred image references to images that no longer exist')
```
Toflar: I still think it should be the opposite and missing should delete by default.

ausi: @contao/developers WDYT?

leofeyer: What is the motivation to keep references to images that no longer exist?

ausi: If the image is only missing temporarily.

leofeyer: I think I am with @Toflar on this one. Delete by default but provide a `--preserve-missing` or `--no-deletes` flag for these special situations.

ausi: Done in b947f1158095e585cdd1404b499a2dbe1db2c4e9

aschempp (dismissed):
Looks good to me. I would prever `--no-deletes` as I feel this exists somewhere else, but obviously both work.

> now returns a 404 

that was misleading, I was expecting a response and wanted to complain you should throw an exception but you already did 😂

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2884, opened 2021-03-16
Fix the version 4.8.0 migration

### LINKED ISSUE #2881: Provide migration for failed Version480Update migration
Due to https://github.com/contao/contao/issues/2822 there are some Contao 4 installations out there now which were not migrated correctly. There are two problems to solve:

* If the `Version480Update` fails, it will not be offered to be run again, since `ALTER TABLE tl_layout DROP picturefill` is executed right at the beginning of the migration. This should be moved to the end.
* Since the issue in #2822 happened in [this line](https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Migration/Version408/Version480Update.php#L176) everything after it was not executed. We should implement a migration that detects a failed migration state and runs the necessary updates again (e.g. important part transformations).

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2881

This is currently untested, but it should be much more robust this way I think.


ausi:
An alternative would be to split the migration up into 7 different migrations. Should be quite easy to do now.

fritzmg:
Ah yes, splitting the migration up into different components is much better 👍 

> An alternative would be to split the migration up into 7 different migrations. Should be quite easy to do now.

That would be at least the preferred way for future migrations, since we already do it this way (see the `Version409` subnamespace for example). It would also provide better feedback to the user, what kind of tasks are actually run within a migration.

core-bundle/src/Migration/Version408/Version480Update.php
```diff
@@ -144,7 +209,45 @@ public function run(): MigrationResult
...
+            SELECT EXISTS(
+                SELECT id
+                FROM tl_files
+                WHERE
+                    importantPartX > 1
+                    OR importantPartY > 1
+                    OR importantPartWidth > 1
+                    OR importantPartHeight > 1
+            )
+        ')->fetchColumn();
```
fritzmg: In theory there could be edge cases where the important part is still absolute instead of relative - yet has a coordinate of `0,0` and a width of `0,0` or `1,0` or `1,1` 😬

leofeyer:
It seems that "Add a PHPUnit extension to test for deprecated classes (see #2863)" is not the correct issue title, is it?

core-bundle/src/Migration/Version408/Version480Update.php
ausi: I know :(
Do you think we should instead check if the column type is an integer (instead of `DOUBLE`). But this would not work then for everyone that has already converted the data type but still has the old “absolute” values.

ausi: Or maybe better: If the column type is integer we migrate, otherwise we only migrate if there are actual values that are integer 2 or higher...

fritzmg: I think that would be the way to go yes. However there might still be edge cases where the migration failed (for whatever reason) _after_ the database field type was already altered. At this point you still have the old absolute integer values in the new double database field.

But I think that is the best we can do for automatic detection.

ausi: This should be better now with 325525c5144939ec568f85a58be50bea878b2399

ausi: I think/hope I got all edge cases now in d74fcec5bc5fb430d05a1bfbd96a08c404cba912

ausi:
I think this should be RTM now. It would be great if someone could test these changes in an upadate of a real installation from 4.4 to 4.9.

fritzmg:
I have one customer that is still on 4.4 and has lots and lots of images with an important part for example. I'll try the new migration implementation with that database later today.

fritzmg (dismissed):
lgtm

I tested with the database of the customer and everything looks fine. The only issue I had is with files that have special characters in them, as these are not cross-platform compatible (between Windows' NTFS and a Linux FS) and thus cannot be found in my local environment and therefore get their important part removed, which is expected.

leofeyer (dismissed):

leofeyer:
Thank you @ausi.

PR #2892, opened 2021-03-19
Add constants for the DCA sorting modes and flags

bezin:
This PR adds constants for sorting mode and flags, that can be used in the data container arrays. This gives us a little bit more autocomplete in IDEs and does not need us to remember every single mode and flag or look it up in the docs every time ;-)

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -36,6 +36,101 @@ use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 abstract class DataContainer extends Backend
 {
 	/**
+	 * Records are not sorted
+	 */
+	public const MODE_UNSORTED = 0;
+
+	/**
+	 * Records are sorted by a fixed field
+	 */
+	public const MODE_SORTED = 1;
+
+	/**
+	 * Records are sorted by a switchable field
+	 */
+	public const MODE_SORTABLE = 2;
+
+	/**
+	 * Records are sorted by the parent table
+	 */
+	public const MODE_SORTED_PARENT = 3;
+
+	/**
+	 * Displays the child records of a parent record (see content elements)
+	 */
+	public const MODE_PARENT = 4;
+
+	/**
+	 * Records are displayed as tree (see site structure)
+	 */
+	public const MODE_TREE = 5;
+
+	/**
+	 * Displays the child records within a tree structure (see articles module)
+	 */
+	public const MODE_TREE_EXTENDED = 6;
+
+	/**
+	 * Sort by initial letter ascending
+	 */
+	public const SORT_INITIAL_LETTER_ASC = 1;
+
+	/**
+	 * Sort by initial letter descending
+	 */
+	public const SORT_INITIAL_LETTER_DESC = 2;
+
+	/**
+	 * Sort by initial two letters ascending
+	 */
+	public const SORT_INITIAL_LETTERS_ASC = 3;
+
+	/**
+	 * Sort by initial two letters descending
+	 */
+	public const SORT_INITIAL_LETTERS_DESC = 4;
+
+	/**
+	 * Sort by day ascending
+	 */
+	public const SORT_DAY_ASC = 5;
+
+	/**
+	 * Sort by day descending
+	 */
+	public const SORT_DAY_DESC = 6;
+
+	/**
+	 * Sort by month ascending
+	 */
+	public const SORT_MONTH_ASC = 7;
+
+	/**
+	 * Sort by month descending
+	 */
+	public const SORT_MONTH_DESC = 8;
+
+	/**
+	 * Sort by year ascending
+	 */
+	public const SORT_YEAR_ASC = 9;
+
+	/**
+	 * Sort by year descending
+	 */
+	public const SORT_YEAR_DESC = 10;
+
+	/**
+	 * Sort ascending
+	 */
+	public const SORT_ASC = 11;
+
+	/**
+	 * Sort descending
+	 */
+	public const SORT_DESC = 12;
+
+	/**
 	 * Current ID
 	 * @var integer|string
 	 */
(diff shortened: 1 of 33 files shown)


m-vo:
That's a really helpful addition!

I'd put it into `DC_Table` though, as it's only relevant there.  E.g. with a prefix similar to the config key name: `SORTING_MODE__TREE`, `SORTING_FLAG__INITIAL_LETTER_ASC` and so on.

fritzmg:
I prefer having separate classes, rather than packing everything into `DC_Table`.

m-vo:
> I prefer having separate classes, rather than packing everything into DC_Table.

But then you need to know the classes…

bezin:
For me typing `SortingMode` or `SortingFlag` came more naturally when I am in the  DCA `sorting` section configuring `mode` or `sorting`. Also, I did not have a Kneifzange at hand to work on `DC_Table` 😛 

I would prefer this: 


over this:

fritzmg:
> But then you need to know the classes…

But that's also the case with `DC_Table`. You don't write DCAs within the `DC_Table` or `DataContainer` class, so you wouldn't be able to just type `self::` anyway. You would need to have knowledge about where the flags are stored in any case.

aschempp:
This way, I have to remember 1,2,3 different class names! I would prefer it in `DataContainer` as well, e.g. `DataContainer::SORT_ INITIAL_LETTER_ASC`.

rabauss:
I'm torn! On the one hand, you have to remember the classes, but on the other hand, what are the other large lists of constants for DCAs? Actually only the input types, right? If we add it to the `DataContainer` it should be added via traits!

bezin:
Yes, good point. I would like to add these as well. Wouldn't it make sense, to add the input type constants to every widget class? Which, I must admit, would make the case for adding the sorting mode and flag constants to the DataContainer/DC_Table 🤔

leofeyer:
> This way, I have to remember 1,2,3 different class names! I would prefer it in `DataContainer`

Me too.

aschempp:
> Wouldn't it make sense, to add the input type constants to every widget class? 

It would actually make sense to use the class name instead of an arbitrary string for the widget types… not sure if we can add that in a BC-way 🤔

fritzmg:
> It would actually make sense to use the class name instead of an arbitrary string for the widget types…

I thought we did not want to support this: https://github.com/contao/contao/pull/2378#issuecomment-714511735

bezin:
I deleted the new classes and moved everything to `DC_Table` – as @m-vo already mentioned, the constants are only used there but not in `DC_Folder` or `DC_File`.

leofeyer:
> I'd put it into `DC_Table` though, as it's only relevant there. E.g. with a prefix similar to the config key name: `SORTING_MODE__TREE`, `SORTING_FLAG__INITIAL_LETTER_ASC` and so on.

I don't like this. Yes, the constants might not be used in DC_Folder, but they might be used in a custom child class of DC_Folder. In the end, it is more about in the scope in which the constants make sense – and that is definitely in any data container. Thus `DataContainer::SORT_ASC` instead of `DC_Table::SORT_ASC`. (Files could be sorted descending, too.)

leofeyer:
As discussed in Mumble on April 8th, two things should be changed:

1. The constants should be in the `DataContainer` class.
2. The constants should be used in the code. 😄

bezin:
Hello there, thanks for your comments (unfortunately I could not make it to the Mumble call as I was on my holidays). 
I updated as you discussed. I hope I got them all – please, somebody double check :-)

One minor thing came up: The sorting flags `3` and `4` differ a little between the list sorting flag and the per-field sorting flag: For the list it is always `Sort by initial first two letters` (hence the name), but for a single field you may configure how many letters should be taken into account with the `length` option.

I still used the sorting flag there, as I thought everything else would be more confusing. If somebody has an idea for a name that reflects both possibilities, we could update the naming.

bezin:
Hi there, I actually thought that this was finished already. 

Anyways, I renamed `SORT_INITIAL_TWO_LETTERS_ASC` to `SORT_INITIAL_LETTERS_ASC` so we now have the following options:

* **SORT_INITIAL_LETTER_ASC (or_DESC)**
  Sort by the initial letter
* **SORT_INITIAL_LETTERS_ASC (or _DESC)**
   Sort by initial letters, which is either fixed to the first two characters in list sorting or can be configured with `length` when used in a field configuration.

It's final from my side now, let me know if there is anything I can do.

Cheers 👋

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -143,6 +143,101 @@ abstract class DataContainer extends Backend
...
 
+	/**
+	 * Records are not sorted
+	 */
+	public const MODE_UNSORTED = 0;
+
+	/**
+	 * Records are sorted by a fixed field
+	 */
+	public const MODE_FIXED_FIELD = 1;
```
```suggestion
	public const MODE_SORTED = 1;
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -143,6 +143,101 @@ abstract class DataContainer extends Backend
...
+
+	/**
+	 * Records are sorted by a switchable field
+	 */
+	public const MODE_SWITCHABLE_FIELD = 2;
```
```suggestion
	public const MODE_SORTABLE = 2;
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -143,6 +143,101 @@ abstract class DataContainer extends Backend
...
+
+	/**
+	 * Records are sorted by the parent table
+	 */
+	public const MODE_PARENT_TABLE = 3;
```
```suggestion
	public const MODE_SORT_PARENT = 3;
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -143,6 +143,101 @@ abstract class DataContainer extends Backend
...
+
+	/**
+	 * Displays the child records of a parent record (see style sheets module)
+	 */
+	public const MODE_CHILD_RECORD = 4;
```
```suggestion
	public const MODE_CHILDREN = 4;
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -143,6 +143,101 @@ abstract class DataContainer extends Backend
...
+	 */
+	public const MODE_SWITCHABLE_FIELD = 2;
+
+	/**
+	 * Records are sorted by the parent table
+	 */
+	public const MODE_PARENT_TABLE = 3;
+
+	/**
+	 * Displays the child records of a parent record (see style sheets module)
```
```suggestion
	 * Displays the child records of a parent record (see content elements)
```

bezin:
I do like all of your suggestions 👍 Will update them later.

bezin:
Everything is updated.

aschempp (dismissed):

leofeyer (changes requested):
I am not happy with `MODE_CHILDREN` and `MODE_CHILD_TREE`. Can we come up with something else? @contao/developers /cc

leofeyer:
@bezin Can you please rebase the PR on the current `4.x` branch and resolve the conflicts?

bezin:
What about `MODE_CHILD_TREE` and `MODE_CHILD_LIST` instead of `MODE_CHILDREN`?

aschempp (changes requested):
```php
public const MODE_UNSORTED = 0;
public const MODE_SORTED = 1;
public const MODE_SORTABLE = 2;
public const MODE_SORTED_PARENT = 3;
public const MODE_PARENT = 4;
public const MODE_TREE = 5;
public const MODE_TREE_EXTENDED = 6;
```

bezin:
All done here from my point of view 👍

Toflar (dismissed):
Nice work @bezin <3

Toflar:
Need to fix the CI tests though ;)

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4800,7 +4800,7 @@ protected function listView()
...
-<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < 4) ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1] ?? '') . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
+<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < self::MODE_PARENT ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1] ?? '') . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
```
bytehead: There's a `)` missing after `self::MODE_PARENT`

bytehead (changes requested):

core-bundle/src/Resources/contao/drivers/DC_Table.php
bytehead:
```suggestion
<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < self::MODE_PARENT) ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1] ?? '') . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
```

bezin: Thanks, just spotted this myself 👍

bezin:
Done 👍

bytehead (dismissed):

leofeyer:
I would also change this: https://github.com/bezin/contao/pull/1

leofeyer (approved):

leofeyer:
Thank you @bezin.

PR #2894, opened 2021-03-19
Fix a normalize whitespace error in the DomCrawler

fritzmg:
Currently the following error will occur when running the Crawler:

```
"exception" => Exception {
    #message: ""Symfony\Component\DomCrawler\Crawler::text()" will normalize whitespaces by default in Symfony 5.0, set the second "$normalizeWhitespace" argument se to retrieve the non-normalized version of the text."
    #code: 16384
    #file: "vendor\terminal42\escargot\src\Subscriber\RobotsSubscriber.php"
    #line: 244
    trace: {
        vendor\terminal42\escargot\src\Subscriber\RobotsSubscriber.php:244 { …}
        Terminal42\Escargot\Subscriber\RobotsSubscriber->Terminal42\Escargot\Subscriber\{closure}() {}
        vendor\symfony\dom-crawler\Crawler.php:622 { …}
        vendor\contao\contao\core-bundle\src\Search\Document.php:113 { …}
        vendor\symfony\dom-crawler\Crawler.php:355 { …}
        vendor\contao\contao\core-bundle\src\Search\Document.php:120 { …}
        vendor\contao\contao\core-bundle\src\Search\Indexer\DefaultIndexer.php:146 { …}
        vendor\contao\contao\core-bundle\src\Search\Indexer\DefaultIndexer.php:76 { …}
        vendor\contao\contao\core-bundle\src\Search\Indexer\DelegatingIndexer.php:34 { …}
        vendor\contao\contao\core-bundle\src\Crawl\Escargot\Subscriber\SearchIndexSubscriber.php:165 { …}
        vendor\terminal42\escargot\src\Escargot.php:515 { …}
        vendor\terminal42\escargot\src\Escargot.php:457 { …}
        vendor\terminal42\escargot\src\Escargot.php:330 { …}
        vendor\contao\contao\core-bundle\src\Command\CrawlCommand.php:135 { …}
        vendor\symfony\console\Command\Command.php:255 { …}
        vendor\symfony\console\Application.php:1027 { …}
        vendor\symfony\framework-bundle\Console\Application.php:97 { …}
        vendor\symfony\console\Application.php:273 { …}
        vendor\symfony\framework-bundle\Console\Application.php:83 { …}
        vendor\symfony\console\Application.php:149 { …}
        vendor\bin\contao-console:21 { …}
        vendor\bin\contao-console:21 { …}
    }
}
```

This PR fixes this by specifically setting the second parameter of `Symfony\Component\DomCrawler\Crawler::text` to `true` (which was previously the default).

### DIFF
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -110,7 +110,7 @@ class Document
             ->filterXPath('descendant-or-self::script[@type = "application/ld+json"]')
             ->each(
                 static function (Crawler $node) {
-                    $data = json_decode($node->text(), true);
+                    $data = json_decode($node->text(null, true), true);
 
                     if (JSON_ERROR_NONE !== json_last_error()) {
                         return null;
--- a/core-bundle/src/Search/Indexer/DefaultIndexer.php
+++ b/core-bundle/src/Search/Indexer/DefaultIndexer.php
@@ -55,7 +55,7 @@ class DefaultIndexer implements IndexerInterface
         }
 
         try {
-            $title = $document->getContentCrawler()->filterXPath('//head/title')->first()->text();
+            $title = $document->getContentCrawler()->filterXPath('//head/title')->first()->text(null, true);
         } catch (\Exception $e) {
             $title = 'undefined';
         }


Toflar:
We should have noticed this with tests. Apparently that was not covered, can you add a test case for that?

fritzmg:
Hm, the tests would already cover this - but no error occurs there. Presumably because `trigger_error` is already silenced in Symfony. However, I suppose that may be the [custom error handler in Escargot](https://github.com/terminal42/escargot/blob/faad8af1d62835cbe7f33530eb42b7609849f143/src/Subscriber/RobotsSubscriber.php#L244-L246) makes the error then actually visible. May be Escargot should restore the error handler after [this block](https://github.com/terminal42/escargot/blob/faad8af1d62835cbe7f33530eb42b7609849f143/src/Subscriber/RobotsSubscriber.php#L244-L253)?

fritzmg:
That would explain why the error only now surfaces. The error handler in Escargot was only added 10 days ago (in https://github.com/terminal42/escargot/commit/ef7d131f91224d4a74578b34aadefcf7496c3053#diff-e66b75aabb721cde7baf72cca5b86fcc33eb03e24ed5b5d550d0a0b2c476972c).

@Toflar do you want me to create an issue there?

core-bundle/src/Search/Document.php
```diff
@@ -110,7 +110,7 @@ public function extractJsonLdScripts(string $context = '', string $type = ''): a
             ->filterXPath('descendant-or-self::script[@type = "application/ld+json"]')
             ->each(
                 static function (Crawler $node) {
-                    $data = json_decode($node->text(), true);
+                    $data = json_decode($node->text(null, true), true);
```
leofeyer, line 113: I don't understand. If `true` was the default and will still be the default in the future, why do we have to set it explicitly?

fritzmg: It is not technically necessary with the release of `terminal42/escargot` `1.0.5`. However, not setting the parameters will still result in 

```php
@trigger_error(sprintf('"%s()" will normalize whitespaces by default in Symfony 5.0, set the second "$normalizeWhitespace" argument to false to retrieve the non-normalized version of the text.', __METHOD__), \E_USER_DEPRECATED);
```

being triggered under certain circumstances.

leofeyer: Then we should ignore the deprecation in these places as we do not want the non-normalized version of the text.

ausi: If we do not want the non-normalized version we have to use `->text(null, true)` in Symfony 4.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #2899, opened 2021-03-21
Fix error when using return => Array for models

### LINKED ISSUE #2897: Model::findMultipleByIds() fails when asked to return an array
**Affected version(s)**

Contao 4.9.12

**Description**

When called with the option `'return' => 'Array'`, findMultipleByIds throws an error:

> "Call to a member function next() on array" at contao/core-bundle/src/Resources/contao/library/Contao/Model.php line 872.

There's simply no logic in place to handle an array returned from the call to `find()` in line 868.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2897
| Docs PR or issue | -

When using

```php
Model::findMultipleByIds([…], ['return' => 'Array']);
```

the following error occurs:

```
Error:
Call to a member function next() on array

  at vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Model.php:872
  at Contao\Model::findMultipleByIds()
```

This PR changes the affected code to always use `foreach` instead of `while` so that it does not matter whether `Array` or `Collection` is used for the `return` option.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Model.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Model.php
@@ -869,10 +869,10 @@ abstract class Model
 
 			if ($objMissing !== null)
 			{
-				foreach ($objMissing as $objMissingRecord)
+				foreach ($objMissing as $objCurrent)
 				{
-					$intId = $objMissingRecord->{static::$strPk};
-					$arrRegistered[$intId] = $objMissingRecord;
+					$intId = $objCurrent->{static::$strPk};
+					$arrRegistered[$intId] = $objCurrent;
 				}
 			}
 		}
--- a/core-bundle/src/Resources/contao/library/Contao/Model.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Model.php
@@ -869,10 +869,10 @@ abstract class Model
 
 			if ($objMissing !== null)
 			{
-				while ($objMissing->next())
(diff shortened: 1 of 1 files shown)


aschempp (dismissed):

bytehead (dismissed):

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -869,10 +869,10 @@ public static function findMultipleByIds($arrIds, array $arrOptions=array())
 
 			if ($objMissing !== null)
 			{
-				while ($objMissing->next())
+				foreach ($objMissing as $objMissingRecord)
```
leofeyer: Why did you choose `$objMissingRecord` instead of something short like `$model` or `$current` (or `$objModel` if you want to stick to the old naming schema)?

fritzmg: I wanted to stick to the old naming schema and use a descriptive variable name. Since `$objMissing` is the collection then iterating over the collection gives you a single record - hence the name `$objMissingRecord`. Otherwise I would go for `$objCurrent`.

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #2901, opened 2021-03-22
Respect _target_path on logout if set

### LINKED ISSUE #2855: Logout does not redirect to home page on protected pages
**Affected version(s)**

Contao 4.9.12 (Symfony 4.4.19)
Contao 4.11.1 (Symfony 5.2.3)

**Description**

If you click the logout button in the login module on a protected page you get redirected to the protected page which results in a 401 response.

The `_target_path` hidden field gets set to the home page correctly (if you are on a protected page), but it seems like symfony ignores it.

`_always_use_target_path` is set to the empty string, but setting it to `1` manually doesn’t seem to have an effect either.

bytehead:
See #2855

- [x] Add a test if a decision in the solution is made

### DIFF
--- a/core-bundle/src/Security/Logout/LogoutSuccessHandler.php
+++ b/core-bundle/src/Security/Logout/LogoutSuccessHandler.php
@@ -45,6 +45,12 @@ class LogoutSuccessHandler extends DefaultLogoutSuccessHandler
             return $this->httpUtils->createRedirectResponse($request, $targetUrl);
         }
 
+        $targetUrl = $request->request->get('_target_path');
+
+        if (null !== $targetUrl && '' !== $targetUrl) {
+            return $this->httpUtils->createRedirectResponse($request, $targetUrl);
+        }
+
         if ($targetUrl = $request->headers->get('Referer')) {
             return $this->httpUtils->createRedirectResponse($request, $targetUrl);
         }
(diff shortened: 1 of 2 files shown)


ausi:
Should we maybe use the `redirect` query parameter instead? https://github.com/contao/contao/issues/2855#issuecomment-793113648

I’m a little bit confused by `_target_url` and `redirect`, are they both the same? Is one from us and the other from Symfony? Or should they do different things?

core-bundle/src/Security/Logout/LogoutSuccessHandler.php
```diff
@@ -45,6 +45,12 @@ public function onLogoutSuccess(Request $request): Response
             return $this->httpUtils->createRedirectResponse($request, $targetUrl);
         }
 
+        $targetUrl = $request->request->get('_target_path');
+
+        if (null !== $targetUrl && '' !== $targetUrl) {
```
aschempp: Looks like this could be simplified the same code as on line 44? Also, I would respect the POST parameter before a query, wdyt?

bytehead: was not sure about the precedence here. 🤷‍♂️

bytehead: and the use of `_target_path` only makes sense when it's not empty string, doesn't it?

ausi: Using `if ($targetUrl = $request->request->get('_target_path')) {` would also skip the empty string so I think we can use the same coding style here.

bytehead: Adjusted in f88532321.

ausi (approved):

aschempp (approved):

leofeyer:
Thank you @bytehead.

PR #2903, opened 2021-03-23
Fixed some error notices

aschempp:


core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -1072,7 +1072,7 @@ protected static function find(array $arrOptions)
 		}
 
 		$objStatement = static::preFind($objStatement);
-		$objResult = $objStatement->execute($arrOptions['value']);
+		$objResult = $objStatement->execute($arrOptions['value'] ?? null);
 
 		if ($objResult->numRows < 1)
```
bytehead, line 1077: if `$objResult` is null, access to an undefined property will raise a warning in PHP 8.

```suggestion
		if (null !== $objResult && $objResult->numRows < 1)
```

bytehead (changes requested):

core-bundle/src/Resources/contao/library/Contao/Model.php
aschempp: why would `$objStatement->execute()` return `null`? It's a database query, not a model collection 😉

bytehead: Ops, I've missed the parenthesis 🙈 😂

bytehead (dismissed):

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -421,7 +423,7 @@ protected function createExtract()
 						$table = substr($config['foreignKey'], 0, strrpos($config['foreignKey'], '.'));
 					}
 
-					$arrRelations[$field] = array_merge(array('table'=>$table, 'field'=>'id'), $config['relation']);
+					$arrRelations[$field] = array_merge(array('table'=>$table, 'field'=>'id', 'load'=>'lazy', 'type'=>null), $config['relation']);
```
leofeyer: This should be fixed where the values are read and not where they are written.

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
aschempp: Shouldn't we make sure the relation data is always consistent?

leofeyer: No, because that is a behavior change and thus a possible BC break.

aschempp: Changed in acc9110172ba8f3aa53526b159221c94013b0bb7

core-bundle/src/Resources/contao/library/Contao/Model/QueryBuilder.php
```diff
@@ -45,9 +45,9 @@ public static function find(array $arrOptions)
 			foreach ($objBase->getRelations() as $strKey=>$arrConfig)
 			{
 				// Automatically join the single-relation records
-				if ($arrConfig['load'] == 'eager' || $arrOptions['eager'])
+				if (($arrConfig['load'] ?? null) == 'eager' || ($arrOptions['eager'] ?? false))
```
leofeyer:
```suggestion
				if (($arrConfig['load'] ?? null) == 'eager' || ($arrOptions['eager'] ?? null))
```

Using `?? false` is a behavior change again, even though I don't think it makes a difference. @ausi How would you handle this?

ausi: I would use `?? null` for consistency but it is exactly the same in this case I think.

m-vo: I remember this discussion :upside_down_face:

aschempp: Changed in 9cd2ac02efc6ba2e8b36a22be0659c01e32c870d

leofeyer:
Please rebase the PR onto the `4.11` branch.

aschempp:
All changes have already been applied (probably through other PRs) in 4.11 😆

PR #2904, opened 2021-03-23
Add type constant to each widget class

bezin:
For the same reasons as in #2892 I would like to add a public constant `type` to every widget class to ease the pain of writing a DCA a little :-)

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -197,6 +197,62 @@ abstract class Widget extends Controller
 	 */
 	protected $objDca;
 
+	public const TYPE_CHECKBOX = 'checkbox';
+
+	public const TYPE_CHECKBOX_WIZARD = 'checkboxWizard';
+
+	public const TYPE_CHMOD_TABLE = 'chmod';
+
+	public const TYPE_FILE_SELECTOR = 'fileSelector';
+
+	public const FILE_TREE_TYPE = 'fileTree';
+
+	public const TYPE_IMAGE_SIZE = 'imageSize';
+
+	public const TYPE_INPUT_UNIT = 'inputUnit';
+
+	public const TYPE_KEY_VALUE_WIZARD = 'keyValueWizard';
+
+	public const TYPE_LIST_WIZARD = 'listWizard';
+
+	public const TYPE_META_WIZARD = 'metaWizard';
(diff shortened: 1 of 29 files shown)


bytehead:
Shouldn't the available constants be in the class `Widget` to use them like `Widget::TYPE_CHECKBOX`? Otherwise you always need to know/guess the correct widget class :)

bezin:
You are obviously right. moved them to the Widget class. I was wondering if there is a better name for the TRBLWidget, but I figured `TYPE_TOP_RIGHT_BOTTOM_LEFT` is a little to long? It would be more self-explanatory though. 
But who is using this widget anyways ;-)

bytehead:
I didn't even know `trbl` did exist 😅

bytehead (approved):
LGTM apart from the empty newlines between the constants. But CS in legacy code is part of @leofeyer 😬

m-vo:
`trbl` - ah the dreaded *trouble* field. Maybe we could write this as `TYPE_TOP_RIGHT_BOTTOM_LEFT` or `TYPE_MARGIN` so that anyone could have a realistic chance of understanding what it does.

fritzmg:
Unless we switch to class names for widget types, I don't think this can be solved in a unified way. Because with this solution you would have to use `Widget::TYPE_TEXT` for a Contao Core widget, but `CustomWidget::TYPE` (or whatever the custom widget provides, if anything) for any other widget.

bytehead:
> Unless we switch to class names for widget types, I don't think this can be solved in a unified way. 

True 😞

m-vo:
> Because with this solution you would have to use Widget::TYPE_TEXT for a Contao Core widget, but CustomWidget::TYPE (or whatever the custom widget provides, if anything) for any other widget.

But tbh that's a pattern that is very common… We even have similar things in the core: take the `@property` annotations in models for instance that do not include custom columns. 

You could in theory create a `class \App\Widget extends \Contao\Widget { /* more constants go here */ }` if you want auto-completion in your app. :upside_down_face:  Not working without code generation for extensions, though.

leofeyer:
As discussed in Mumble on April 8th, we do not want to merge this, because it does not work for third-party widgets. We have also discussed adding a constant to every widget class or using the FQCN, but these alternatives have their own disadvantages.

bezin:
All right, I will close this then. Thanks for the discussion 👍

PR #2905, opened 2021-03-23
Fix service tagging for url_callback + title_tag_callback

rabauss:
| Q                | A
| -----------------| ---
| Docs PR or issue | contao/docs#694

Fix the service tagging for `url_callback` + `title_tag_callback` of the serp preview widget.

### DIFF
--- a/core-bundle/src/EventListener/DataContainerCallbackListener.php
+++ b/core-bundle/src/EventListener/DataContainerCallbackListener.php
@@ -27,6 +27,8 @@ class DataContainerCallbackListener
         'input_field_callback',
         'options_callback',
         'group_callback',
+        'url_callback',
+        'title_tag_callback',
     ];
 
     /**


aschempp (approved):

fritzmg:
~~Is that enough to fix that? What exactly do you have to define for `target` in this case?~~ nvm, this works automatically through

```php
$keys = explode('.', $target);
$dcaRef = &$this->getDcaReference($table, $keys);
```

with `target=fields.<FIELD>.eval.url` (as you already mentioned in https://github.com/contao/docs/pull/694 👍 )

aschempp:
The behaviour change is that these callbacks are not array of callbacks, but just one. Everything else should just work as it is.

fritzmg (approved):

leofeyer:
Thank you @rabauss.

PR #2906, opened 2021-03-23
Fix label callback for tree view in picker widget

### LINKED ISSUE #2882: Inconsistent parameters of label callback for tree view in picker widget
**Affected version(s)**

Contao 4.9+

**Description**

Unfortunately, the parameters for the call of `label_callback` are inconsistent for the tree view in the picker widget - see:
https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Resources/contao/widgets/Picker.php#L281-L291

versus:
https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4004-L4014

How should we solve this? I guess we need to add a further switch for mode 5 + 6 in the picker widget!

rabauss:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2882 

Fix the call of the label callbacks of tree views in the picker widget.
With the `isProtected` boolean I was not sure, if we need to solve that recursive.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -284,9 +284,7 @@ class Picker extends Widget
 
 			if (\in_array($mode, array(5, 6)))
 			{
-				$isProtected = 'tl_page' === $dc->table && $arrRow['protected'];
-
-				return $this->{$labelConfig['label_callback'][0]}->{$labelConfig['label_callback'][1]}($arrRow, $label, $dc, '', false, $isProtected);
+				return $this->{$labelConfig['label_callback'][0]}->{$labelConfig['label_callback'][1]}($arrRow, $label, $dc, '', false, null);
 			}
 
 			return $this->{$labelConfig['label_callback'][0]}->{$labelConfig['label_callback'][1]}($arrRow, $label, $dc, $arrRow);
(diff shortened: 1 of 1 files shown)


aschempp:
> With the isProtected boolean I was not sure, if we need to solve that recursive.

Can you point to the code how this is done in DC_Table?

rabauss:
In the `DC_Table` the method `generateTree` is called recursive and then the `$blnProtected` is "merged" - see:
https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3917-L3923

and:
https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4105
https://github.com/contao/contao/blob/cd9121439b195c2588fb741ba704a2e42e7bf729/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4125

EDIT:

In the picker widget we have no recursive call, because there we only have the currently selected value(s).
And in the `DC_Table` this is only a edge case for the site structure (`tl_page`) to use the correct protected icon for child pages.
I'm really not sure how to solve that!
Is the current solution sufficient or should we say this variable is always false in picker context?

aschempp:
I would be fine with always being false or even `null` because it seems to be an edge-case that shouldn't have been there in the first place …

rabauss:
@aschempp thanks for your feedback! Setting the flag to `null` seemed to be the best solution for me ;-)

aschempp (approved):

leofeyer:
Thank you @rabauss.

PR #2917, opened 2021-03-25
Remove the search results cache

### LINKED ISSUE #2869: Remove the search cache
Now that the search is fast enough, we no longer need the search cache.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2869
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -46,12 +46,6 @@ class Automator extends System
 		// Clear the index
 		$searchIndexer->clear();
 
-		$strCachePath = StringUtil::stripRootDir(System::getContainer()->getParameter('kernel.cache_dir'));
-
-		// Purge the cache folder
-		$objFolder = new Folder($strCachePath . '/contao/search');
-		$objFolder->purge();
-
 		// Add a log entry
 		$this->log('Purged the search tables', __METHOD__, TL_CRON);
 	}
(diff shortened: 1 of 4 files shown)


Toflar (dismissed):

fritzmg (dismissed):

bytehead (dismissed):

core-bundle/src/Resources/contao/library/Contao/Automator.php
```diff
@@ -194,9 +188,13 @@ public function purgePageCache()
 
 	/**
 	 * Purge the search cache
+	 *
+	 * @deprecated Deprecated since Contao 4.0, to be removed in Contao 5.0.
```
ausi:
```suggestion
	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
```

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -147,41 +147,17 @@ protected function compile()
 			}
 
 			$arrResult = null;
```
ausi: This line can be removed I think because it gets set both in the `try` and in the `catch` block.

core-bundle/src/Resources/contao/library/Contao/Automator.php
leofeyer: Fixed in 4116d124af8d2e63077846cdad61a147e3c87045.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
leofeyer: Changed in 4116d124af8d2e63077846cdad61a147e3c87045.

ausi (approved):

PR #2918, opened 2021-03-26
Use `origId` if defined by alias element

### LINKED ISSUE #2823: Content include element broken with fragments

discordier:
ContentAlias set's `origId` property for the id of the aliased element.
Therefore we need to use that one in the proxy, as otherwise the alias will get rendered by the proxy instead of the "aliased" element.

| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2823


aschempp:
Looks like a duplicate of https://github.com/contao/contao/pull/2851 to me? 🙃

aschempp:
Sorry thats not #2851. I can't find my other one (yet) 😂

aschempp:
Well 🙄 it is actually solved by #2851 …

discordier:
> Well  it is actually solved by #2851 …

Feel free to close then.

leofeyer (commented):
Besides the failing CS job, the fix looks correct to me. We should merge it independently from #2851 IMHO.

leofeyer (approved):

aschempp (changes requested):
I disagree. `origId` is not a valid model property, it is "randomly" set by the alias element. We shouldn't introduce a "feature" to use another ID than the model ID, because someone might have modified the model ID.

In the context of fragments, setting properties on the source model is pointless, because only the ID is taken and forwarded to the fragment handler (e.g. for ESI cases). There's a concept in place to forward other properties, that's what #2930 (successor of #2851)  implements.

leofeyer:
Closing this in favor of #2930 then.

PR #2919, opened 2021-03-26
Prevent an 'Undefined array key "id"' warning in the clipboard

leofeyer:
It is so nice that undefined array keys are now warnings in PHP 8. 😉

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3544,6 +3544,10 @@ class DC_Table extends DataContainer implements \listable, \editable
 			$blnClipboard = true;
 			$arrClipboard = $arrClipboard[$this->strTable];
 		}
+		else
+		{
+			$arrClipboard = null;
+		}
 
 		if (isset($GLOBALS['TL_DCA'][$table]['config']['label']))
 		{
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4163,11 +4171,15 @@ protected function parentView()
 			$blnClipboard = true;
 			$arrClipboard = $arrClipboard[$table];
 
-			if (\is_array($arrClipboard['id']))
+			if (!empty($arrClipboard['id']) && \is_array($arrClipboard['id']))
```
m-vo: Could be shortened to this:
```suggestion
			if (\is_array($arrClipboard['id'] ?? null))
```
(`??` works on arrays as well as on `null`, https://3v4l.org/HEvNi)

leofeyer: Yeah, but there is a subtle difference:

```php
$arrClipboard['id'] = [];

if (\is_array($arrClipboard['id'] ?? null)) {
    // will be executed
}

if (!empty($arrClipboard['id']) && \is_array($arrClipboard['id'])) {
    // will not be executed
}
```

And since we do not want `$blnMultiboard` to be `true` if the array is empty, we need both checks here.

leofeyer: On the other hand, we have always set `$blnMultiboard` to `true` even if the array was empty, so my change actually implies a behavior change. We should merge your suggestion instead. 👍🏻

bytehead (approved):

ausi (approved):

PR #2923, opened 2021-03-29
Fix more PHP 8 undefined array index errors

ausi:
Some more ☺️

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentAccordion.php
+++ b/core-bundle/src/Resources/contao/elements/ContentAccordion.php
@@ -50,7 +50,7 @@ class ContentAccordion extends ContentElement
 				->applyLegacyTemplateData($this->Template, $this->imagemargin, $this->floating);
 		}
 
-		$classes = StringUtil::deserialize($this->mooClasses, true) + array('', '');
+		$classes = StringUtil::deserialize($this->mooClasses, true) + array(null, null);
 
 		$this->Template->toggler = $classes[0] ?: 'toggler';
 		$this->Template->accordion = $classes[1] ?: 'accordion';
--- a/core-bundle/src/Resources/contao/elements/ContentAccordion.php
+++ b/core-bundle/src/Resources/contao/elements/ContentAccordion.php
@@ -50,7 +50,7 @@ class ContentAccordion extends ContentElement
 				->applyLegacyTemplateData($this->Template, $this->imagemargin, $this->floating);
 		}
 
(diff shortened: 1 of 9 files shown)


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1506,7 +1506,7 @@ public static function addImageToTemplate($template, array $rowData, $maxWidth =
 				Metadata::VALUE_ALT => $rowData['alt'] ?? '',
 				Metadata::VALUE_TITLE => $rowData['imageTitle'] ?? '',
 				Metadata::VALUE_URL => self::replaceInsertTags($rowData['imageUrl'] ?? ''),
-				'linkTitle' => $rowData['linkTitle'] ?: '',
+				'linkTitle' => $rowData['linkTitle'] ?? '',
```
ausi: @m-vo this correct?

core-bundle/src/Resources/contao/elements/ContentAccordion.php
```diff
@@ -50,7 +50,7 @@ protected function compile()
 				->applyLegacyTemplateData($this->Template, $this->imagemargin, $this->floating);
 		}
 
-		$classes = StringUtil::deserialize($this->mooClasses);
+		$classes = StringUtil::deserialize($this->mooClasses, true) + array('', '');
```
leofeyer: Isn't `array('', '')` a behavior change compared to `array(null, null)`?

core-bundle/src/Resources/contao/library/Contao/Controller.php
m-vo: Probably. In theory someone could have set `linkTitle` to `false` or `0` which would have worked in the old code but would now result in an error.

 If we wanted to be extra safe, we could for instance do one of these:

```suggestion
				'linkTitle' => ($rowData['linkTitle'] ?? null) ?: '',
```

```suggestion
				'linkTitle' => (string)($rowData['linkTitle'] ?? ''),
```

core-bundle/src/Resources/contao/elements/ContentAccordion.php
ausi: Fixed in 93139b85368a3aa735f527f91f9dba3be7353084

m-vo (approved):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2930, opened 2021-03-31
Add template properties to fragment proxy

aschempp:
This is an alternative implementation to https://github.com/contao/contao/pull/2851 I came up with while fixing issues in https://github.com/contao/contao/pull/2851.

Basically, instead of adding more _arbritrary_ known properties from the proxy to the fragment, it implements a way of passing any property to the fragment template, even across ESI requests.

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentProxy.php
+++ b/core-bundle/src/Resources/contao/elements/ContentProxy.php
@@ -63,17 +63,17 @@ class ContentProxy extends ContentElement
 
 	public function __set($strKey, $varValue)
 	{
-		$this->reference->attributes['templateProps'][$strKey] = $varValue;
+		$this->reference->attributes['templateProperties'][$strKey] = $varValue;
 	}
 
 	public function __get($strKey)
 	{
-		return $this->reference->attributes[$strKey]['templateProps'];
+		return $this->reference->attributes['templateProperties'][$strKey];
 	}
 
 	public function __isset($strKey)
 	{
-		return isset($this->reference->attributes['templateProps'][$strKey]);
+		return isset($this->reference->attributes['templateProperties'][$strKey]);
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/modules/ModuleProxy.php
```diff
@@ -19,20 +20,60 @@
...
+	}
+
+	public function __set($strKey, $varValue)
+	{
+		$this->reference->attributes['templateProps'][$strKey] = $varValue;
+	}
+
+	public function __get($strKey)
+	{
+		return $this->reference->attributes[$strKey]['templateProps'];
```
leofeyer: Should this not be `->attributes['templateProps'][$strKey]` like above and below?

core-bundle/src/Resources/contao/elements/ContentProxy.php
```diff
@@ -19,20 +20,60 @@
...
+	}
+
+	public function __set($strKey, $varValue)
+	{
+		$this->reference->attributes['templateProps'][$strKey] = $varValue;
+	}
+
+	public function __get($strKey)
+	{
+		return $this->reference->attributes[$strKey]['templateProps'];
```
leofeyer: Should this not be `->attributes['templateProps'][$strKey]` as above?

leofeyer (changes requested):
I think I would prefer `templateProperties` instead of `templateProps` but I understand that `templateProps` is shorter, so whatever.

core-bundle/src/Resources/contao/modules/ModuleProxy.php
aschempp: Good catch! Fixed in f4209135eeb5b2d2b742d0b11f508201ea4af652

core-bundle/src/Resources/contao/elements/ContentProxy.php
aschempp: Good catch! Fixed in f4209135eeb5b2d2b742d0b11f508201ea4af652

leofeyer (approved):

zonky2:
it is not fixed if this is a CE-MM list/filter or FE module MM list/filter

PR #2931, opened 2021-03-31
Fix filtering for recurring events

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

In 7670ee1e9fc5adf63a73ce662e57d73ab63138f9 `filter => true` was replaced with `search => true` for `tl_calendar_events.recurring`. However this field is a checkbox and you cannot search through such fields. I think this has been changed on accident - @bytehead or was there a specific reason? Currently you can neither search nor filter for events that have been set to _recurring_.

Keep in mind that re-enabling the filtering for _recurring_ will increase the amount of filtering options for Contao 4.11, making the filters take up two rows instead of one. But that's just how it is if we want to be able to keep filtering for the appropriate fields.

### DIFF
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -520,7 +520,6 @@ $GLOBALS['TL_DCA']['tl_calendar_events'] = array
 		'noComments' => array
 		(
 			'exclude'                 => true,
-			'filter'                  => true,
 			'inputType'               => 'checkbox',
 			'eval'                    => array('tl_class'=>'w50 clr'),
 			'sql'                     => "char(1) NOT NULL default ''"
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -426,7 +426,7 @@ $GLOBALS['TL_DCA']['tl_calendar_events'] = array
 		'recurring' => array
 		(
 			'exclude'                 => true,
-			'search'                  => true,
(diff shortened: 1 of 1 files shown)


bytehead:
I can't remember but I would say this is a bug and it was changed by accident.

bytehead:
May @leofeyer can shed some light on this?

fritzmg:
Failing tests are unrelated.

bytehead (dismissed):

leofeyer:
I think I have removed the filter because there are six filters already and I did not want a line break. Maybe we can remove the filter on the `author` field instead? The authors are already searchable and we probably do not need both options.

fritzmg:
> Maybe we can remove the filter on the `author` field instead? The authors are already searchable and we probably do not need both options.

Personally I prefer having the author as a filter as it provides you with a list of authors to filter by - and thus also shows you which authors (back end users) actually have created events (or news). Using the search requires you to do trial & error on that and you would need to look up which authors are available.

aschempp (dismissed):

leofeyer:
> However this field is a checkbox and you cannot search through such fields.

This is not correct BTW. You can search through these fields by using `1` (is recurring) or `^$` (is not recurring). But that is not very intuitive of course.

leofeyer:
Maybe we can omit the "Disable comments" filter instead?

fritzmg:
Fine with me at least. I never ever had to use comments for events for a customer's site so far.

And if someone still needs it, it's pretty easy to customize.

fritzmg:
PR updated accordingly. Failing test is unrelated.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #2933, opened 2021-03-31
Compile the limit menu after all other filter panels

### LINKED ISSUE #2932: DCA panels depend on specific order
**Affected Versions**
Contao 4.9

Hello there,

in a recent project we had the issue that `DC_Table` ignored our `filter`and `search` panels for the `limit` panel and back-end pagination. The search by e-mail-address results in three entries, which are correctly listed (not visible in the screenshot). The pagination and the `limit` panel however still take all some hundred entries into account.

The panel configuration in this case was `limit;search;sort`. Because the search panel sets the actual search conditions and values, the limit menu rendered as no search took place. The limit menu then sets the limit for the pagination, hence the pagination is off, too. Defining the panels with `search;sort;limit` fixed this issue.

We should compile the panels in the order necessary to properly work and than reorder them to what was defined by the developer. I'll provide a PR for this as soon as I find the time :-)

Cheers

bezin:
Fixes #2932. Basically if there is a limit menu, I insert a placeholder and skip compiling the menu at first. After all other panels were compiled, especially search and filter, I compile the limit menu and replace the placeholder.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -1181,7 +1181,7 @@ abstract class DataContainer extends Backend
 
 				// The limit menu depends on other panels that may set a filter query, e.g. search and filter.
 				// In order to correctly calculate the total row count, the limit menu must be compiled last.
-				// We isnert a placeholder here and compile the limit menu after all other panels.
+				// We insert a placeholder here and compile the limit menu after all other panels.
 				if ($strSubPanel == 'limit')
 				{
 					// Set placeholder to inject compiled limit menu later
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -1179,40 +1179,41 @@ abstract class DataContainer extends Backend
 			{
 				$panel = '';
 
-				// The limit menu depends on other panels that may set a filter query, e.g. search and filter.
-				// In order to correctly calculate the total row count, the limit menu must be compiled last.
-				// We insert a placeholder here and compile the limit menu after all other panels.
-				if ($strSubPanel == 'limit')
+				switch ($strSubPanel)
 				{
-					// Set placeholder to inject compiled limit menu later
-					$panel = '###limit_menu###';
-				}
-				// Regular panels
-				elseif ($strSubPanel == 'search' || $strSubPanel == 'sort')
-				{
-					$panel = $this->{$strSubPanel . 'Menu'}();
-				}
+					case 'limit':
+						// The limit menu depends on other panels that may set a filter query, e.g. search and filter.
+						// In order to correctly calculate the total row count, the limit menu must be compiled last.
+						// We insert a placeholder here and compile the limit menu after all other panels.
+						$panel = '###limit_menu###';
+						break;
 
-				// Multiple filter subpanels can be defined to split the fields across panels
-				elseif ($strSubPanel == 'filter')
-				{
-					$panel = $this->{$strSubPanel . 'Menu'}(++$intFilterPanel);
-				}
+					case 'search':
+						$panel = $this->searchMenu();
+						break;
 
-				// Call the panel_callback
-				else
-				{
-					$arrCallback = $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['panel_callback'][$strSubPanel];
+					case 'sort':
+						$panel = $this->sortMenu();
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1179,8 +1179,16 @@ protected function panel()
 			{
 				$panel = '';
 
+				// The limit menu depends on other panels that may set a filter query, e.g. search and filter.
+				// In order to correctly calculate the total row count, the limit menu must be compiled last.
+				// We isnert a placeholder here and compile the limit menu after all other panels.
```
Toflar:
```suggestion
				// We insert a placeholder here and compile the limit menu after all other panels.
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1179,8 +1179,16 @@ protected function panel()
...
+				// In order to correctly calculate the total row count, the limit menu must be compiled last.
+				// We insert a placeholder here and compile the limit menu after all other panels.
+				if ($strSubPanel == 'limit')
+				{
+					// Set placeholder to inject compiled limit menu later
+					$panel = '###limit_menu###';
+				}
 				// Regular panels
-				if ($strSubPanel == 'search' || $strSubPanel == 'limit' || $strSubPanel == 'sort')
+				elseif ($strSubPanel == 'search' || $strSubPanel == 'sort')
```
m-vo: Now that nearly all `$strSubPanel` variants are handled individually, we could maybe replace the last two `$panel = $this->{$strSubPanel . 'Menu'}(…);` expressions with real function calls? Took me a while to get what this was doing.

bezin: Yeah, I agree this takes a moment to figure that out. Would you continue with if {} elseif {} in this case?

m-vo: Yes, or a `switch` statement. But let's hear what others think about this first.

aschempp: Doesn't sound like something related to this PR (fixing the limit menu) 😇

leofeyer: What's wrong with `$this->{$strSubPanel . 'Menu'}()`? It is a real function call, isn't it? 😄 

I'm ok with a switch statement as well, though.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1226,6 +1234,18 @@ protected function panel()
...
+		// Compile limit menu if placeholder is present
+		foreach ($arrPanels as $key => $strPanel)
+		{
+			if (strpos('###limit_menu###', $strPanel) === false)
+			{
+				continue;
+			}
+
+			$limitPanel = $this->limitMenu();
+			$arrPanels[$key] = str_replace('###limit_menu###', $limitPanel, $strPanel);
```
```suggestion
			$arrPanels[$key] = str_replace('###limit_menu###', $this->limitMenu(), $strPanel);
```

No need for a variable here, is there?

core-bundle/src/Resources/contao/classes/DataContainer.php
bezin: I refactored this part to a switch statement. Definitely helps with readability.

m-vo: > What's wrong with $this->{$strSubPanel . 'Menu'}()? It is a real function call, isn't it? smile

:smile: I was more referring to things like this where the dynamic part is unnecessary:

```php
elseif ($strSubPanel == 'filter')
{
    $panel = $this->{$strSubPanel . 'Menu'}(++$intFilterPanel);
}
```
&rarr;
```php
elseif ($strSubPanel == 'filter')
{
    $panel = $this->filterMenu(++$intFilterPanel);
}
```

aschempp (approved):

Toflar (approved):

leofeyer:
Thank you @bezin.

fritzmg:
It looks like this in the current `4.9` branch for me:



Can anyone confirm?

bezin:
~I cannot. Which view is this? Could you share your DCA config?~

Yes, I can. The parameters passed to `strpos()` are in the wrong order 🙈 Hence, the code does not work if the limit menu is in the same row with another menu. I'll fix that.

fritzmg:
That's the view of a news archive. No customisation.

edit: ok :)

PR #2935, opened 2021-04-01
Fix the PageModel registry state

### LINKED ISSUE #2934: Models in Registry include additional data for non-existent columns
```php
PageModel::findByPk(2)->subpages; // null
PageModel::findPublishedSubpagesWithoutGuestsByPid(1);
PageModel::findByPk(2)->subpages; // "5"
```

This can lead to very weird bugs.
_(I had an issue where a navigation module displayed subpages only on a page where a sitemap module was present)_

To fix this, I think we should remove the subpages column before creating the collection in https://github.com/contao/contao/blob/9864858d36a8319b68b21d3c0f7b912a9d205fc7/core-bundle/src/Resources/contao/models/PageModel.php#L612 and add them back to copies of the models before returning.

This way we can keep the performance benefit of the registry.

Also I’m a bit confused by https://github.com/contao/contao/blob/9864858d36a8319b68b21d3c0f7b912a9d205fc7/core-bundle/src/Resources/contao/models/PageModel.php#L95 
Does someone know a case where `$pageModel->subpages` is actually an array instead of a string-int?

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2934

~~Work in progress. Adds a tests that shows the issue from #2934~~

With these changes, the models get stored in the registry without the `subpages` column.

### DIFF
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -11,7 +11,6 @@
 namespace Contao;
 
 use Contao\CoreBundle\Exception\NoRootPageFoundException;
-use Contao\Database\Result;
 use Contao\Model\Collection;
 use Contao\Model\Registry;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
@@ -93,7 +92,6 @@ use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
  * @property boolean $rootIsFallback
  * @property boolean $rootUseSSL
  * @property string  $rootFallbackLanguage
- * @property integer $subpages
  * @property boolean $minifyMarkup
  * @property integer $layoutId
  * @property boolean $hasJQuery
@@ -596,9 +594,13 @@ class PageModel extends Model
 	 * @param boolean $blnIsSitemap  If true, the sitemap settings apply
 	 *
 	 * @return Collection|PageModel[]|PageModel|null A collection of models or null if there are no pages
+	 *
+	 * @deprecated Deprecated since Contao 4.9, to be removed in Contao 5.0. Use PageModel::findPublishedWithoutGuestsByPid() instead.
 	 */
 	public static function findPublishedSubpagesWithoutGuestsByPid($intPid, $blnShowHidden=false, $blnIsSitemap=false)
 	{
+		@trigger_error('Using PageModel::findPublishedSubpagesWithoutGuestsByPid() has been deprecated and will no longer work Contao 5.0. Use PageModel::findPublishedWithoutGuestsByPid() instead.', E_USER_DEPRECATED);
+
 		$time = Date::floorToMinute();
 		$tokenChecker = System::getContainer()->get('contao.security.token_checker');
 		$blnFeUserLoggedIn = $tokenChecker->hasFrontendUser();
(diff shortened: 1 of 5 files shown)


core-bundle/tests/Contao/PageModelTest.php
```diff
@@ -0,0 +1,174 @@
...
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        Registry::getInstance()->reset();
+
+        // Reset database instance
+        $property = (new \ReflectionClass(Database::class))->getProperty('arrInstances');
+        $property->setAccessible(true);
+        $property->setValue([]);
```
m-vo, line 70: We should probably add a `reset` method to the `Database` class as well. Wdyt?
(Or I'll do it in #2518 - that one's still in progress because things turned out way more complicated than I thought)

ausi: Yes, adding it to #2518 would make sense I think.

aschempp (commented):
I don't really like this. Models can always have additional properties in Contao, think of `loadDetails` etc. The relevant issue is that `subpages` _might_ not be set if the model is loaded from registry. So we should only make sure to add the property from the query data, but not clone the original models.

ausi:
I think `loadDetails` is different, because it always loads the same data. But `subpages` is a computed column that depends on the parameters given to `findPublishedSubpagesWithoutGuestsByPid()`.

If I would write the method from scratch today I would not return `subpages` as part of the entities but as a separate array.

> So we should only make sure to add the property from the query data, but not clone the original models.

This is exactly what the old code does. But don’t you think the example code from https://github.com/contao/contao/issues/2934#issue-846757522 is problematic?

aschempp:
> This is exactly what the old code does. But don’t you think the example code from #2934 (comment) is problematic?

No I don't think that's problematic, simply because that never was a problem and works everywhere the same in all Contao models, e.g. in https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/elements/ContentAlias.php#L41-L43

However, I doubt the current code would merge `subpages` from the database result into the model, if the model already exists in the database?

ausi:
> No I don't think that's problematic, simply because that never was a problem

Why did we fix this in other places then? https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Resources/contao/elements/ContentModule.php#L60-L63

Also, it caused a very weird bug in one of my extensions, that’s how I noticed this inconsistency ( https://github.com/madeyourday/contao-rocksolid-mega-menu/commit/f5ac4e0b087e067a129e015a89f9148565e45552 )

> However, I doubt the current code would merge `subpages` from the database result into the model, if the model already exists in the database?

https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Resources/contao/library/Contao/Model/Collection.php#L136

ausi:
As discussed in Mumble on April 8th, we should check if it is possible to replace `findPublishedSubpagesWithoutGuestsByPid()` with a better alternative and deprecate the old method.

aschempp (commented):
hmm, is that really what we discussed? I thought we'd write the DB query manually in the navigation module where we need subpages so we don't have a repository method that returns "fake" data.

ausi:
> I thought we'd write the DB query manually in the navigation module

A protected method in `Module` so that `ModuleBooknav` can reuse it?

aschempp:
well shoot, I didn't notice it's used in multiple places. But a model/registry method that does not return an array of models but an array of props seems more confusing than a new method for that case. It could also be a static method in `ModuleNavigation` imho 🤷

ausi:
> It could also be a static method in `ModuleNavigation` imho 🤷

Well it is used in the base `Module` class, so adding it to `Module` would be a better fit I think.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -595,21 +594,67 @@ public static function findPublishedByIdOrAlias($varId, array $arrOptions=array(
...
 											  ->execute($intPid);
 
 		if ($objSubpages->numRows < 1)
 		{
 			return null;
 		}
 
-		return static::createCollectionFromDbResult($objSubpages, 'tl_page');
+		return static::createCollectionFromDbResult($objSubpages, static::$strTable);
```
```suggestion
		return static::createCollectionFromDbResult($objSubpages, 'tl_page');
```

I would revert this, because we are already using `tl_page` in the DB queries and thus it is not possible to use another table anyway.

leofeyer (dismissed):
LGTM. Not sure about @aschempp's objections though.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2937, opened 2021-04-04
Fix PHP 8 compat of DC_Table/Environment

### LINKED ISSUE #2936: [C4.11 + PHP8] Warnings for undefined array keys of server
**Affected version(s)**

Contao 4.11  with PHP 8 in Docker environment

**Description**

I'm getting the following warnings in my docker dev environment:
```php
PHP Warning:  Undefined array key "SERVER_NAME" in /app/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Environment.php on line 299

PHP Warning:  Undefined array key "SERVER_PORT" in /app/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Environment.php on line 301

PHP Warning:  Undefined array key "SERVER_PORT" in /app/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Environment.php on line 303
```

rabauss:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2936

Fix array key checks for php 8 compat of DC_Table and Environment


core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -290,15 +290,15 @@ protected static function httpUserAgent()
 	 */
 	protected static function httpHost()
 	{
-		if (!empty($_SERVER['HTTP_HOST']))
+		if (isset($_SERVER['HTTP_HOST']) && !empty($_SERVER['HTTP_HOST']))
```
fritzmg: `empty()` already incoroporates `isset`, so this is not necessary.

> No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to `!isset($var) || $var == false`. 

https://www.php.net/manual/en/function.empty.php

fritzmg (changes requested):

core-bundle/src/Resources/contao/library/Contao/Environment.php
rabauss: Thanks! I should not do that in the middle of the night :-D

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -296,9 +296,9 @@ protected static function httpHost()
 		}
 		else
 		{
-			$host = $_SERVER['SERVER_NAME'];
+			$host = $_SERVER['SERVER_NAME'] ?? '';
 
-			if ($_SERVER['SERVER_PORT'] != 80)
+			if (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] != 80)
```
aschempp: Our PHP8 adjustments always try to mimic the previous behaviour, always cast to `null`

```suggestion
			if (($_SERVER['SERVER_PORT'] ?? null) != 80)
```

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -296,9 +296,9 @@ protected static function httpHost()
 		}
 		else
 		{
-			$host = $_SERVER['SERVER_NAME'];
+			$host = $_SERVER['SERVER_NAME'] ?? '';
```
aschempp:
```suggestion
			$host = $_SERVER['SERVER_NAME'] ?? null;
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -5000,7 +5000,7 @@ protected function listView()
 				{
 					$current = $row[$firstOrderBy];
 					$orderBy = $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['fields'] ?? array();
-					$sortingMode = (\count($orderBy) == 1 && $firstOrderBy == $orderBy[0] && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'] ?? null)) ? $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] : $GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'];
+					$sortingMode = (\count($orderBy) == 1 && $firstOrderBy == $orderBy[0] && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'] ?? null)) ? $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] : $GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'] ?? null;
```
aschempp:
```suggestion
					$sortingMode = (\count($orderBy) == 1 && $firstOrderBy == $orderBy[0] && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'] ?? null)) ? $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] : ($GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['flag'] ?? null);
```

aschempp (changes requested):

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -296,9 +296,9 @@ protected static function httpHost()
 		}
 		else
 		{
-			$host = $_SERVER['SERVER_NAME'];
+			$host = $_SERVER['SERVER_NAME'] ?? null;
 
-			if ($_SERVER['SERVER_PORT'] != 80)
+			if (($_SERVER['SERVER_PORT'] ?? null) != 80)
```
leofeyer: These env variables should always exist. Why did you have to change this?

rabauss: Sorry, I wasn't sure if it does - and it was close to the affected code `$_SERVER['SERVER_NAME']`. Shall I remove the check?

fritzmg: Since other methods already use `System::getContainer()->get('request_stack')->getCurrentRequest()` may be the remaining methods should do the same? Then this issue would not occur.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4825,7 +4825,7 @@ protected function listView()
...
-<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < 4) ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1]) . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
+<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < 4) ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1] ?? $labelNew[0]) . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
```
rabauss: One extension had no translation key `tl_table.new.1` so the Fallback should be `$labelNew[0]`, should it?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -892,7 +892,7 @@ protected function generateGlobalButtons()
 
 			$v = \is_array($v) ? $v : array($v);
 			$label = \is_array($v['label']) ? $v['label'][0] : $v['label'];
-			$title = \is_array($v['label']) ? $v['label'][1] : $v['label'];
+			$title = \is_array($v['label']) ? ($v['label'][1] ?? '') : $v['label'];
```
rabauss: fix for extension see: https://github.com/oveleon/contao-component-style-manager/pull/46/files#diff-bc2c37cb346d11ba0a90db5779f412b8d9f79de1ed6a2a58451fe22a6efaf989R25

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -776,8 +776,8 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 			{
 				if (\is_array($v['label']))
 				{
-					$label = $v['label'][0];
-					$title = sprintf($v['label'][1], $id);
+					$label = $v['label'][0] ?? $k;
+					$title = sprintf($v['label'][1] ?? $k, $id);
```
rabauss: fix for extension see: https://github.com/oveleon/contao-component-style-manager/pull/46/files#diff-bc2c37cb346d11ba0a90db5779f412b8d9f79de1ed6a2a58451fe22a6efaf989L33

doishub: I think this could cause problems, because `$k` ~~does not necessarily~~ never contain a specifier (sprintf).

core-bundle/src/Resources/contao/classes/DataContainer.php
doishub: Fixed in the extension see: https://github.com/oveleon/contao-component-style-manager/pull/46/commits/7f9110518fd05bd391c6b9829d99fda51df8f035#diff-5c38568b17b8589bfb3893db5658f251b4cf63594660bc5c0238d9d9d9d24271

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: I don't think it can cause problems, but it is at the least a behaviour change. Should be changed to:

```suggestion
					$title = sprintf($v['label'][1] ?? '', $id);
```

doishub: That's right, I remembered that `sprintf` can't handle strings that don't contain a specifier and return `false`. However, empty values can be handled here, so I prefer the variant to return an empty string to avoid errors. 

[With PHP 8.0 `false` is no longer returned on errors.](https://www.php.net/manual/en/function.sprintf.php#refsect1-function.sprintf-changelog)

fritzmg: Hm, I cannot confirm that. It shows no error and does not return `false` in any version in such a case: https://3v4l.org/1DRKh

doishub: Yes that's right, that was only in my mind 🙃

That's why I prefer your suggestion to return an empty string, since the key (`$k`) may contain values we can't consider. But yes, no one will create keys with a specifier.

rabauss: I just used `$k` because it is the initial fallback: https://github.com/contao/contao/blob/d56bc8067cf5b2c3255fd1eed2bdad3766bcf05c/core-bundle/src/Resources/contao/classes/DataContainer.php#L773

It could be solved outside `sprintf` but as @fritzmg confirmed it doesn't matter if there is no specifier included in `$k`

But of course this is a behavior change so the `''` would be the right solution!

core-bundle/src/Resources/contao/drivers/DC_Table.php
rabauss: or empty string as fallback?
```suggestion
<a href="' . ($this->ptable ? $this->addToUrl('act=create' . ((($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) < 4) ? '&amp;mode=2' : '') . '&amp;pid=' . $this->intId) : $this->addToUrl('act=create')) . '" class="header_new" title="' . StringUtil::specialchars($labelNew[1] ?? '') . '" accesskey="n" onclick="Backend.getScrollOffset()">' . $labelNew[0] . '</a> ' : '') . $this->generateGlobalButtons() . '
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -776,8 +776,8 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 			{
 				if (\is_array($v['label']))
 				{
-					$label = $v['label'][0];
-					$title = sprintf($v['label'][1], $id);
+					$label = $v['label'][0] ?? $k;
```
rabauss: For no behavior change it should be maybe:
```suggestion
					$label = $v['label'][0] ?? null;
```
But an operation without label makes no sense!

ausi:
Could you please also change https://github.com/contao/contao/blob/a00c8d7ed3fae6cc24f33e56e9217ea79e591f2f/core-bundle/src/Resources/contao/drivers/DC_Table.php#L6007 to:

```php
list($offset, $limit) = explode(',', $this->limit) + array(null, null);
```

rabauss:
Done! `array_pad` would be also a possible solution, but as there was allready ` + array(null, null)`  in the `DC_TABLE` I used the suggested solution from @ausi

core-bundle/src/Resources/contao/library/Contao/Environment.php
rabauss: so the solution would be this? 

```php
	protected static function httpHost()
	{

		if ($request === null)
		{
			return null;
		}

		return $request->getHost();
	}
```

fritzmg: You might still need to append the port manually though. Though I don't quite understand the old code. It appends the port if `HTTP_HOST` is empty and the `SERVER_PORT` is not `80`? That means it could return the port with an SSL connection?

rabauss: You're absolutely right I used the wrong method - it has to be `getHttpHost`, which I commited! Successfully tested with docker environment with ports 😉

core-bundle/src/Resources/contao/classes/DataContainer.php
```suggestion
			$title = \is_array($v['label']) ? ($v['label'][1] ?? null) : $v['label'];
```

Otherwise it is a behavior change.

leofeyer (changes requested):
The changes to the `Environment::httpHost()` method need to be reverted, because the method has to work without a Symfony request, too.

core-bundle/src/Resources/contao/library/Contao/Environment.php
rabauss: I reverted the changes with Symfony request!
It is really also neccessary to check `SERVER_PORT` - because on CLI while cache clear it is not defined in my docker environment!

ausi (approved):

leofeyer:
Thank you @rabauss.

PR #2941, opened 2021-04-07
Remove the system log entry for 'no root page found'

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

In Contao `4.2.1` (1811b8b632cfa59a2e8f9f8f488249fe444f430a) all system log entries related to 404 responses were removed in order to not spam the system log with such entries anymore.

However, one such log entry remains. Whenever someone randomly accesses a language that does not exist in a multi-language Contao setup (with the language parameter enabled), a 404 response is generated - but still also the following log entry is generated:

```
No root page found (host "example.com", language "xx")
```

While this can be prevented by creating a dedicated 404 page within your site structure for each website root, I think this system log entry should also be removed, as bots can again spam such errors on sites where there happens to be no dedicated 404 page configured (which is not required anyway).

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Frontend.php
+++ b/core-bundle/src/Resources/contao/classes/Frontend.php
@@ -347,8 +347,6 @@ abstract class Frontend extends Controller
 			}
 			catch (RoutingExceptionInterface $exception)
 			{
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

bytehead (approved):

aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #2945, opened 2021-04-08
Add the `--migrations-only` option to the migrate command

### LINKED ISSUE #2760: Keep migration_versions when migrating
**Affected version(s)**

Contao 4.11

**Description**

Now that we dropped the schema filter, the `contao:migrate` command will offer to drop the `migration_versions` table that keeps track of those migrations executed via `doctrine:migration:migrate`. This is because this table doesn't have any associated metadata. IMHO we should have a look how doctrine handles this case and exclude it as well.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2760
| Docs PR or issue | contao/docs#705

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -19,6 +19,7 @@ use Contao\InstallationBundle\Database\Installer;
 use Symfony\Component\Config\Exception\FileLocatorFileNotFoundException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Exception\InvalidOptionException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


core-bundle/src/Command/MigrateCommand.php
```diff
@@ -75,6 +76,7 @@ protected function configure(): void
         $this
             ->addOption('with-deletes', null, InputOption::VALUE_NONE, 'Execute all database migrations including DROP queries. Can be used together with --no-interaction.')
             ->addOption('schema-only', null, InputOption::VALUE_NONE, 'Execute database schema migration only.')
+            ->addOption('migrations-only', null, InputOption::VALUE_NONE, 'Execute migrations only and skip schema diff.')
             ->setDescription('Executes migrations and the database schema diff.')
```
leofeyer:
```suggestion
            ->addOption('schema-only', null, InputOption::VALUE_NONE, 'Only update the database schema.')
            ->addOption('migrations-only', null, InputOption::VALUE_NONE, 'Only execute the migrations.')
            ->setDescription('Executes migrations and updates the database schema.')
```

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #2948, opened 2021-04-13
Fix array to string conversion in the picker

ausi:
The source code for the order fields currently looks like:

```html
<input type="hidden" name="orderPages" id="ctrl_orderPages" value="Array">
```

This was probably not noticed so far because the `Backend.makeMultiSrcSortable()` directly fixes the value on load.


fritzmg (approved):

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #2952, opened 2021-04-13
Set a label for the Contao Manager back end menu entry

cliffparnitzky:
Avoids missing message error in debug mode.

### DIFF
--- a/manager-bundle/src/EventListener/BackendMenuListener.php
+++ b/manager-bundle/src/EventListener/BackendMenuListener.php
@@ -157,11 +157,11 @@ class BackendMenuListener
 
         $item = $event->getFactory()
             ->createItem('contao_manager')
-            ->setLabel('contao_manager')
+            ->setLabel('Contao Manager')
             ->setUri('/'.$this->managerPath)
             ->setLinkAttribute('class', 'navigation contao_manager')
             ->setLinkAttribute('title', $this->translator->trans('contao_manager_title', [], 'ContaoManagerBundle'))
-            ->setExtra('translation_domain', 'ContaoManagerBundle')
+            ->setExtra('translation_domain', false)
         ;
 
         $categoryNode->addChild($item);
--- a/manager-bundle/src/EventListener/BackendMenuListener.php
+++ b/manager-bundle/src/EventListener/BackendMenuListener.php
@@ -157,9 +157,11 @@ class BackendMenuListener
 
         $item = $event->getFactory()
             ->createItem('contao_manager')
-            ->setLabel('Contao Manager')
+            ->setLabel('contao_manager')
             ->setUri('/'.$this->managerPath)
             ->setLinkAttribute('class', 'navigation contao_manager')
+            ->setLinkAttribute('title', $this->translator->trans('contao_manager', [], 'ContaoManagerBundle'))
+            ->setExtra('translation_domain', 'ContaoManagerBundle')
         ;
 
         $categoryNode->addChild($item);
--- a/manager-bundle/src/EventListener/BackendMenuListener.php
+++ b/manager-bundle/src/EventListener/BackendMenuListener.php
@@ -160,7 +160,7 @@ class BackendMenuListener
             ->setLabel('contao_manager')
             ->setUri('/'.$this->managerPath)
             ->setLinkAttribute('class', 'navigation contao_manager')
-            ->setLinkAttribute('title', $this->translator->trans('contao_manager', [], 'ContaoManagerBundle'))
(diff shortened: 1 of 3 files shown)


aschempp:
hmm, thats kinda stupid. I don't think `Contao Manager` should ever be translated, as it is a product name. The missing translation message isn't better though. Isn't there a way to _not_ translate the label in KnpMenu?

cliffparnitzky:
On the one hand I'm with you. If it could be solved with a configuration to avoid the missing message error, this would also be fine. On the other hand this offers the possibility to define a Tooltip (which I currently forgot to add).

aschempp:
by tooltip you mean the title attribute? A title attribute that is the same as the link content does not make much sense to me?

cliffparnitzky:
That's absolutly right. So, it would be the best to have a title like `Open the Contao Manager to administrate the Contao system`.

leofeyer:
@cliffparnitzky Isn't this a bugfix?

cliffparnitzky:
I'm not sure. I'm totally fine with your decision. Shall I change the base to `4.9`? I can also add a tooltip, as discussed above.

leofeyer:
@contao/developers WDYT?

ausi:
👍 for bugfix and 👍 for the tooltip

Toflar:
I've updated the base branch to point to 4.9 then. @cliffparnitzky do you mind bringing the PR up to date then? :)

cliffparnitzky:
I've added the tooltip.

Toflar:
By "bringing up to date" I meant rebase/merge so it's ready to be sqashed into 4.9 :)

cliffparnitzky:
Hopefully cherry picking was successful ;-)

Toflar (dismissed):
Nice work 👍

ausi (dismissed):

manager-bundle/src/Resources/translations/ContaoManagerBundle.en.xlf
```diff
@@ -5,6 +5,12 @@
       <trans-unit id="1" resname="debug_mode">
         <source>Debug mode</source>
       </trans-unit>
+      <trans-unit id="2" resname="contao_manager">
+        <source>Contao Manager</source>
+      </trans-unit>
+      <trans-unit id="3" resname="contao_manager_title">
+        <source>Open the Contao Manager to administrate the Contao system.</source>
```
leofeyer: What about just "Open the Contao Manager" or "Manage the Contao installation" or "Manage updates and extensions and maintain the application"?

cliffparnitzky: I'm fine with all suggestions, especially the last one. But we need a good text to avoid @aschempp questioning the sense ;-).

manager-bundle/src/EventListener/BackendMenuListener.php
```diff
@@ -157,9 +157,11 @@ private function addManagerLink(MenuEvent $event): void
 
         $item = $event->getFactory()
             ->createItem('contao_manager')
-            ->setLabel('Contao Manager')
+            ->setLabel('contao_manager')
             ->setUri('/'.$this->managerPath)
             ->setLinkAttribute('class', 'navigation contao_manager')
+            ->setLinkAttribute('title', $this->translator->trans('contao_manager_title', [], 'ContaoManagerBundle'))
+            ->setExtra('translation_domain', 'ContaoManagerBundle')
```
aschempp: Setting `translation_domain` to `false` will disable the label translation

aschempp (changes requested):
I think adding the title is a very good idea, and I'm fine with all versions. But `Contao Manager` should not be translatable because it's a product name. We should keep the original but fix the translator error by disabling it (see code hint)

manager-bundle/src/EventListener/BackendMenuListener.php
cliffparnitzky: Jep, this works. Great. My suggestion:
- [x] I change the translation domain proposed by @aschempp 
- [x] I fix set set Contao Manager label in the code
- [x] I remove the translation for the label
- [x] I change the title to "Manage updates and extensions and maintain the application"
Any hints, remarks, doubts ?

leofeyer: Sounds good. 👍🏻

cliffparnitzky: Everything checked in.

aschempp (approved):
Great work! Thank you @cliffparnitzky

leofeyer:
Thank you @cliffparnitzky.

PR #2954, opened 2021-04-15
Do not start the session to check if a property exists

aschempp:
If legacy code calls `exists($_SESSION['foo'])`, we're starting the session in our `LegacySessionAccess`. However, this is totally unnecessary if there is no previous session, because the value can obviously never be there.


ausi:
Did you check if code like `isset($_SESSION['FORM_DATA'][$strKey])` does not start the session now?

Toflar:
That looks correct to me. Just wondering if we should inject the Request (nullable) as we're injecting the whole session too.

aschempp:
> That looks correct to me. Just wondering if we should inject the Request (nullable) as we're injecting the whole session too.

Then we wouldn't need to inject the session. But that would be unnecessary coupling to me and make tests harder. I don't think we'll ever need the full request?

Toflar (dismissed):
LGTM but @ausi's question should be double checked.

leofeyer (approved):

PR #2957, opened 2021-04-17
Show zero values in the list view and DCA filters

fritzmg:
If you add something like this to your DCA:

```php
// contao/dca/tl_member.php

$GLOBALS['TL_DCA']['tl_member']['fields']['some_number'] = [
    'label' => ['Some number', 'Some foobar number.'],
    'filter' => true,
    'inputType' => 'text',
    'eval' => ['rgxp' => 'digit', 'tl_class' => 'w50', 'maxlength' => 32],
    'sql' => ['type' => 'string', 'length' => 32, 'default' => ''],
];

$GLOBALS['TL_DCA']['tl_member']['list']['label']['fields'][5] = 'some_number';

PaletteManipulator::create()
    ->addField('some_number', 'personal_legend', PaletteManipulator::POSITION_APPEND)
    ->applyToPalette('default', 'tl_member')
;
```` 

and then enter a zero (`0`) into the new field, the back end will show `-` instead of `0` both in the list view when `showColumns` is enabled and in the filter menu of that field.

In my opinion it only makes sense to display `-` instead of `0` if the field acts as a foreign key to another table. Otherwise no presumptions can be made about the semantics of the field.

This PR would show `-` whenever `foreignKey` is set for the field and its content is falsy and otherwise it will only show `-` whenever the field is an empty string.

**Before:**

Screenshot_2021-04-17 Mitglieder c49 local(1) 

**After:**

Screenshot_2021-04-17 Mitglieder c49 local

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -5038,7 +5038,18 @@ class DC_Table extends DataContainer implements \listable, \editable
 				{
 					foreach ($args as $j=>$arg)
 					{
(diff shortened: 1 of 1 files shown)


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #2958, opened 2021-04-17
Fix the tooltips in the JavaScript wizards

### LINKED ISSUE #732: deleting select options causes weird symbols to appear
__Reproduction:__

1. Log into the Contao Demo.
2. Go to _Form generator_ » _Contact_ and edit the `salutation` field.
3. Delete the _second_ option (`mr`).

Weird symbols will appear in the upper left corner of the browser window:

screenshot_2018-08-10 form generator contact fields edit field id 15 - contao open source cms

However, I cannot reproduce it all of the time.

rabauss:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #732

- [x] Do not show tooltip with empty title
- [x] Hide interactive help on delete + init tooltip on actions of new row of optionWizard
- [x] Hide interactive help on delete + init tooltip on actions of new row of moduleWizard
- [x] Hide interactive help on delete + init tooltip on actions of new row of tableWizard
- [x] Hide interactive help on delete + init tooltip on actions of new row of keyValueWizard
- [x] Hide interactive help on delete + init tooltip on actions of new row of sectionWizard


core-bundle/src/Resources/public/mootao.js
```diff
@@ -118,6 +118,10 @@ Tips.Contao = new Class(
 	options: {
 		id: 'tip',
 		onShow: function() {
+			var title = this.tip.getElement('div.tip-title');
+			if (!title || !title.innerHTML) {
+				return;
+			}
```
rabauss, line 124: Maybe we could even hide the tooltip if `event.page[y] == 1` in the `position` method, because it seems to be an error if a tooltip should be show on position 0

core-bundle/src/Resources/public/core.js
```diff
@@ -1329,6 +1329,34 @@ var Backend =
...
+		buttons && buttons.each(function(button) {
+			var img = button.getElement('img');
+			img && button.set('title', img.get('alt'));
+		});
+	},
+
+	/**
+	 * Hide the interactive help
+	 *
+	 * @author Ralf Baussenwein
```
aschempp: please don't add @author, we have a Git commit log for that 😊

rabauss: okay okay, I thought this is the code style in the js file, because there are lots of copyrights 😁

core-bundle/src/Resources/contao/widgets/ModuleWizard.php
```diff
@@ -212,19 +212,19 @@ public function generate()
...
+					$return .= ' <a href="contao/main.php?do=themes&amp;table=tl_module&amp;act=edit&amp;id=' . $this->varValue[$i]['mod'] . '&amp;popup=1&amp;nb=1&amp;rt=' . REQUEST_TOKEN . '" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['tl_layout']['edit_module']) . '" class="module_link" ' . (($this->varValue[$i]['mod'] > 0) ? '' : ' style="display:none"') . ' onclick="Backend.openModalIframe({\'title\':\'' . StringUtil::specialchars(str_replace("'", "\\'", $GLOBALS['TL_LANG']['tl_layout']['edit_module'])) . '\',\'url\':this.href});return false">' . Image::getHtml('edit.svg', str_replace("'", "\\'", $GLOBALS['TL_LANG']['tl_layout']['edit_module'])) . '</a>' . Image::getHtml('edit_.svg', '', 'class="module_image"' . (($this->varValue[$i]['mod'] > 0) ? ' style="display:none"' : ''));
```
leofeyer:
```suggestion
					$return .= ' <a href="contao/main.php?do=themes&amp;table=tl_module&amp;act=edit&amp;id=' . $this->varValue[$i]['mod'] . '&amp;popup=1&amp;nb=1&amp;rt=' . REQUEST_TOKEN . '" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['tl_layout']['edit_module']) . '" class="module_link" ' . (($this->varValue[$i]['mod'] > 0) ? '' : ' style="display:none"') . ' onclick="Backend.openModalIframe({\'title\':\'' . StringUtil::specialchars(str_replace("'", "\\'", $GLOBALS['TL_LANG']['tl_layout']['edit_module'])) . '\',\'url\':this.href});return false">' . Image::getHtml('edit.svg', $GLOBALS['TL_LANG']['tl_layout']['edit_module']) . '</a>' . Image::getHtml('edit_.svg', '', 'class="module_image"' . (($this->varValue[$i]['mod'] > 0) ? ' style="display:none"' : ''));
```

core-bundle/src/Resources/public/core.js
```diff
@@ -1329,6 +1329,30 @@ var Backend =
 		});
 	},
 
+	/**
+	 * Copy the interactive help from the button image
+	 */
+	copyInteractiveHelpFromButtonsImage: function (buttons) {
```
leofeyer: I would prefer `copyInteractiveHelpFromButtonImage` like you wrote in the doc comment.

leofeyer (changes requested):
Thank you so much for debugging and fixing this! 💪🏻 It has been an issue for much too long now.

core-bundle/src/Resources/public/core.js
rabauss: Well, now we can discuss if it's wrong in the docs or if we should rename the method. It really copies the `alt` text from one or multiple buttons - what's convention?

leofeyer: I guess both "the button image" and "the button's image" are valid, however, with 250.000 Google results to 10.000 Google results, "the button image" seems to be more common. Also, it is an elliptical sentence and "copy interactive help from button's image" does not sound correct to me. But maybe @blairwinans can help us out here?

blairwinans: Oh wow, this is a tough one. My preference is `copyInteractiveHelpFromButtonImage` because there is no way to add an apostrophe to show possession. Since it takes an array you could argue that it should just be `copyInteractiveHelpFromButtons` since that is what is being passed, but if you feel the need to include the Image, I would make it singular with `copyInteractiveHelpFromButtonImage`. That my two cents!

rabauss: Okay, I'm fine with `copyInteractiveHelpFromButtonImage` - it was even my first idea before the previous committed one 🙃

core-bundle/src/Resources/contao/widgets/KeyValueWizard.php
```diff
@@ -139,11 +139,11 @@ public function generate()
 			{
 				if ($button == 'drag')
 				{
-					$return .= ' <button type="button" class="drag-handle" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['move']) . '" aria-hidden="true">' . Image::getHtml('drag.svg') . '</button>';
+					$return .= ' <button type="button" class="drag-handle" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['move']) . '" aria-hidden="true">' . Image::getHtml('drag.svg', $GLOBALS['TL_LANG']['MSC']['move']) . '</button>';
```
leofeyer: I don't like these changes. We have deliberately not added the button title to the button image as well, because screen readers will read it twice otherwise. Why is this necessary?

leofeyer (changes requested):

core-bundle/src/Resources/contao/widgets/KeyValueWizard.php
rabauss: Unfortunately, the current tooltip implementation removes the title attribute on init, and so we cannot simply copy the previous line and call `addInteractiveHelp` again. I thought it would be usefull to have the text also as `alt` and copy it before tooltip init from the button to the link (`copyInteractiveHelpFromButtonImage`). If this is a bad solution for screen readers, we should choose a different way!

Of course we could also add another `data` attribute (to be defined!) with the text instead of using the `alt` of the image and then copy the help from there or we need to change the tooltip init!

leofeyer: According to the source code, the attribute is stored as `tip:native`: https://github.com/mootools/mootools-more/blob/687363b141c7d6abb89e6716462bbd99545f81e5/Source/Interface/Tips.js#L134 Maybe we can retrieve the value this way?

rabauss: Thanks for the hint! I could really retrieve the help before the clones and refactored the PR!

core-bundle/src/Resources/public/core.js
```diff
@@ -1329,6 +1329,31 @@ var Backend =
 		});
 	},
 
+	/**
+	 * Retrieve the interactive help
+	 */
+	retrieveInteractiveHelp: function (element) {
+		var buttons = element.getElements('button,a');
```
rabauss: Should we maybe use the `buttons` as input variable for better reusability?

core-bundle/src/Resources/public/core.js
```diff
@@ -1724,6 +1748,7 @@ var Backend =
 								ntr = new Element('tr');
 								childs = tr.getChildren();
 								for (i=0; i<childs.length; i++) {
+									Backend.retrieveInteractiveHelp(childs[i].getElements('button,a'));
```
leofeyer, line 1750: Shouldn't this be `'a[title]', 'input[title]', 'button[title]', 'time[title]'` like in line 1323?

rabauss: if then `'a', 'input', 'button', 'time'`, because we want to retrieve the missing title! but in the widget context there are only `button,a` as actions in the markup! see e.g.: https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Resources/contao/widgets/TableWizard.php

core-bundle/src/Resources/public/core.js
```diff
@@ -1329,6 +1329,30 @@ var Backend =
...
+	 */
+	hideInteractiveHelp: function () {
+		var hideTips = function () {
+			document.querySelectorAll('.tip-wrap').forEach(function (tip) {
+				tip.setStyle('display', 'none');
+			});
+		};
+		hideTips();
+		// hide delayed tips - timeout: Tips.Contao.options.showDelay
+		setTimeout(hideTips, 1000);
```
leofeyer: What if someone overwrites `Tips.Contao.options.showDelay` in their installation? Can we not use the variable instead of a fixed value here?

rabauss: I'm no javascript freak 😇 - maybe someone from @contao/developers can support.
I did not get the value of the comment! To be honest, I would prefer another solution without delay!

leofeyer: @ausi /cc

ausi: `(new Tips.Contao()).options.showDelay`

The correct way would be to delete the actual timers of the tooltips with `clearTimeout()`, but I didn’t find a way to get access to them.

rabauss: Yes, the only way would be if we save the initiated tooltips while `addInteractiveHelp`. Currently we don't have access to them because of the closures e.g. https://github.com/contao/contao/blob/a70561370fcf46a6148c1ca899d1d48c5ec2d023/core-bundle/src/Resources/public/core.js#L1324

rabauss: By the way the delay variable is now set with the recommended code and I don't want to touch the initiation of tooltips to get a better solution for the delayed tooltips!

rabauss:
Any suggestions to merge this PR?

leofeyer (approved):

leofeyer:
Thank you @rabauss.

PR #2959, opened 2021-04-17
Add the back end attributes and badge title to the preview toolbar

rabauss:
This is a reverted commit from contao/pull/1999

I added the badge + attributes also to the preview toolbar - but I'm not sure about the position in the toolbar.
The attributes are set on container `cto-toolbar__inside`

At the moment it would look like this:
image

image

see discussion: https://github.com/contao/contao/pull/1999#issuecomment-756710937

/cc @bezin @denniserdmann


rabauss:
Any suggestions to merge this PR?

aschempp (dismissed):

leofeyer:
As discussed in Mumble on July 29th, we want to add the badges to the preview toolbar on the left side. @rabauss Can you adjust the CSS accordingly and also resolve the existing conflict?

rabauss:
image
image

Now the badge will be added to the top left corner of the toolbar in the frontend.
The attributes are added raw to the toolbar, too!

aschempp:
Nice! I would prefer it to be centered in the desktop bar (or better said male it have the same line height as the button)

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -177,6 +177,24 @@
     .cto-toolbar .tl_submit:active {
         color: #aaa;
     }
+
+    .cto-toolbar .badge-title {
+        align-self: flex-start;
```
rabauss:
```suggestion
        align-self: center;
```
@aschempp  would you like to have the badge centered instead of top? in mobile the badge would be floated to the top left though!

aschempp: Top left on mobile seems good to me, yes I did mean on desktop centered in the bar. Not sure if that CSS does that 😂  if you maybe have another screenshot at hand?

rabauss: With the proposed change it would look like this:
image

aschempp: looks good to me

aschempp:
@rabauss I hope we can get this into 4.13! Could you rebase again and/or fix the tests? 😊

aschempp:
The `contao.backend` parameter has been removed recently. Looks like a BC break to me @contao/developers?

rabauss:
Obviously, the removal of `contao.backend` was intentional because of #3653 - so I updated my PR 
For similar code as in `BackendTemplate` I used the `ContainerInterface` - although I would prefer `ParameterBagInterface`

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -94,6 +98,20 @@ private function renderToolbar(): string
...
         $showUnpublished = $this->tokenChecker->isPreviewMode();
 
+        $attributes = '';
+
+        if ($this->container->hasParameter('contao.backend.attributes')) {
+            $attributes = $this->container->getParameter('contao.backend.attributes');
+
+            if (!empty($attributes) && \is_array($attributes)) {
+                $attributes = ' '.implode(' ', array_map(
+                    static fn ($v, $k) => sprintf('data-%s="%s"', $k, $v),
```
aschempp: we should probably encode `StringUtil::specialchars` on the value to make sure quotes etc. are correctly encoded.

aschempp (changes requested):
shouldn't the badge title and attributes be injected into the constructor (dependency injection)? Alternatively we might use method injection to make it simpler (https://symfony.com/doc/current/service_container/calls.html)

core-bundle/src/Controller/BackendPreviewSwitchController.php
rabauss: Well, it's just copied from the implementation in the backend template: https://github.com/contao/contao/blob/4fa153799e2a3443c55a1f13d961df79b838fdde/core-bundle/src/Resources/contao/classes/BackendTemplate.php#L182

And there is a validation in the configuration: https://github.com/contao/contao/blob/4fa153799e2a3443c55a1f13d961df79b838fdde/core-bundle/src/DependencyInjection/Configuration.php#L538

rabauss:
Of course we could inject the `backend.attributes` and `backend.badge_title` but I thought we keep the code similar to the `BackendTemplate` class with the `container`, don't we?

aschempp:
> Of course we could inject the `backend.attributes` and `backend.badge_title` but I thought we keep the code similar to the `BackendTemplate` class with the `container`, don't we?

I don't think so. `BackendTemplate` only does that because its a legacy class that does not support dependency injection. In anything that does we should use it.

rabauss:
Okay once again some refactoring with DI 😮‍💨

aschempp (dismissed):
thank you @rabauss !

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -88,12 +90,28 @@ public function __invoke(Request $request): Response
...
+        $attributes = '';
+
+        if (!empty($this->backendAttributes)) {
+            $attributes = ' '.implode(' ', array_map(
+                static fn ($v, $k) => sprintf('data-%s="%s"', $k, $v),
+                $this->backendAttributes,
+                array_keys($this->backendAttributes)
+            ));
+        }
+
```
m-vo: HTML composition should be done in the template, not in the controller. IMHO this part can be dropped and the `backendAttributes` directly passed to the template. 

Once we introduce a `html_attr` function or something alike (compare to [here](http://craftsnippets.com/articles/using-attr-function-to-render-html-attributes-in-craft-cms) or [here](https://www.drupal.org/docs/8/theming-drupal-8/using-attributes-in-templates)) this allows a simple use inside templates. Plus you can semantically add/remove things.

core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
```diff
@@ -2,9 +2,12 @@
 <div class="cto-toolbar__open">
     <a href="#" title="{{ 'MSC.openToolbar'|trans }}">{{ include('@ContaoCore/Collector/contao.svg') }}</a>
 </div>
-<div class="cto-toolbar__inside">
+<div class="cto-toolbar__inside" {{ attributes|raw }}>
```
m-vo: Without a helper function and passing the associative array, this would become something like that:
```suggestion
{% set attributes = attributes|merge({class: 'cto-toolbar__open' }) %}
<div {{ attributes|map((v, k) => '%s="%s"'|format(k,v))|join(' ')|raw }}>
```

If we do not trust the input are safe attributes we could even do this, i.e. escaping the attribute parts (btw., this is exactly why the template should compose, not the code. :slightly_smiling_face:):
```twig
attributes|map((v, k) => '%s="%s"'|format(k|e('html_attr'), v|e('html')))|join(' ')|raw
```
/cc @ausi

m-vo (changes requested):
Nice addition, thanks! I only have a remark about the template context.

core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
ausi: I think I would prefer something like this:

```suggestion
<div
    class="cto-toolbar__inside"
    {% for key, value in attributes %}
        data-{{ key|e('html_attr') }}="{{ value }}"
    {% endfor %}
>
```

rabauss:
Sorry guys, I don't have the time for another rework! Please suggest what we need to commit, to get this PR merged!

fritzmg (dismissed):

core-bundle/src/Controller/BackendPreviewSwitchController.php
rabauss: I refactored the template variable for the attributes to an array. Once the new twig function is available it can be refactored in the template!

m-vo (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @rabauss.

PR #2960, opened 2021-04-18
Protect users against Google FLoC

aschempp:
Google’s Federated Learning of Cohorts is a new initiative to track internet users without their consent. This feature is already enabled by default for a small percentage of Google Chrome users, and is intended to be enabled for all users at the end of the experiment.

see https://github.com/api-platform/api-platform/pull/1879 and https://github.com/symfony/symfony/issues/40835

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -320,6 +320,7 @@ abstract class Template extends Controller
 
 		$response = new Response($this->strBuffer);
 		$response->headers->set('Content-Type', $this->strContentType . '; charset=' . Config::get('characterSet'));
+		$response->headers->set('Permissions-Policy', 'interest-cohort=()');
 
 		// Mark this response to affect the caching of the current page but remove any default cache headers
 		$response->headers->set(SubrequestCacheSubscriber::MERGE_CACHE_HEADER, true);


Toflar (approved):

leofeyer:
There are some very active discussions in other communities (e. g. [Drupal](https://www.drupal.org/project/drupal/issues/3209628) and [Wordpress](https://make.wordpress.org/core/2021/04/18/proposal-treat-floc-as-a-security-concern/)) and there is an article by Rohan Kumar with more details that everyone participating in this discussion should read:

https://seirdy.one/2021/04/16/permissions-policy-floc-misinfo.html

I suggest that we wait a little until there is a general consensus. Also, let us wait and see if [Symfony](https://github.com/symfony/symfony/issues/40835) implements it.

bytehead:
For the case we'll add the header I would prefer to set it in the `.htaccess`. It would apply for all requests, not just the PHP ones.

fritzmg:
> I suggest that we wait a little until there is a general consensus. Also, let us wait and see if Symfony implements it.

PR to follow: https://github.com/symfony/symfony/pull/40989

math-GH:
> 
> 
> For the case we'll add the header I would prefer to set it in the `.htaccess`. It would apply for all requests, not just the PHP ones.

.htaccess does not work with every hoster :(
see my forum post: https://community.contao.org/de/showthread.php?80897-FLoC-blockieren-(Hoster-Alfahosting)

Toflar:
I guess "Configuring in the .htaccess" has become a synonym for "webserver configuration". It just means to leave it to the users to configure the server accordingly.

aschempp:
Just an update about other CMSes 😊 
https://wptavern.com/joomla-blocks-floc-by-default-drupal-moves-to-block-floc-in-upcoming-9-2-release

ausi:
> … let us wait and see if [Symfony](https://github.com/symfony/symfony/issues/40835) implements it.

👍

Toflar:
Symfony is irrelevant for us. It's a framework. We're not.

ausi:
I’m against this change, but no hard feelings if it gets merged ☺️

Total-Reality:
We get errors (Console in Chrome) in Contao 4.9.15 on every page because of this change here: https://github.com/contao/contao/commit/f30a5287a307673af92c3782fab5f9816374fba9

"_Error with Permissions-Policy header: Unrecognized feature: 'interest-cohort'._"

bytehead:
If the message is displayed the browser does not know about the new header. Already tried to update Chrome?

fritzmg:
This is expected behaviour. Google Chrome will display this warning in the browser console, even in the newest version.

ausi:
> We get errors (Console in Chrome)

I cannot reproduce this in Firefox or Safari so I think this issue needs to be reported in the Google Chrome issue tracker.

aschempp:
https://digiday.com/marketing/google-switch-floc-cookie-replacement-fingerprinting-potential/

PR #2015, opened 2020-07-28
Replace legacy page types

aschempp:
This is an early draft extracted from the routing PR (https://github.com/contao/contao/pull/1516) to replace the existing page types with real page controllers. We might need a layout framework before we can actually implement this so we don't need to use `PageRegular` to render the page anymore.

PR #2962, opened 2021-04-20
Centralize the JSON-LD metadata management

Toflar:
I would like to enhance Contao's core mechanisms and add a convenient way to manage JSON-LD data for developers so we can provide better machine readability for meta data and provide better out-of-the-box search engine relevant data.
This PR builds on top of the new `ResponseContext` concept introduced in https://github.com/contao/contao/pull/2975.

General remarks and concept background:

- I wanted to have an abstraction layer for everything schema.org related so that we can give extension developers ways to easily add new schema but also extend and or modify schemas that have already been added by the core and/or other developers/extensions. For that I found `spatie/schema-org` which is a vendor people should be familiar with (they're very active in the Laravel community and build excellent libraries). It's been installed about 1.5 mio. times so that feels like a safe bet to me.

  - I've adjusted news templates and `ModuleNews` to showcase how adding JSON-LD data now works.

Decisions to make:

Future:

I would like to extend JSON-LD data for images so we can provide licensing information etc. based on this PR.


ausi:
I’m not sure if this is the right approach for the schema.org metadata. I see benefits in the metadata being directly connected to the markup:

1. If an element gets generated but the output does not end up in the final HTML, the metadata in the markup would be always in sync automatically.
2. Same if I remove some blocks from a template, the connected metadata would be removed too.
3. It is easy to customize and extend by editing the templates.
4. Less complex and better to handle for non-developers (I think?)

This POC doesn’t seem to connect the image data with the associated parent entity. Our current markup-based approach does that, so that the image of a news article is connected to the article it appears in.

For other meta data that is related to the whole page, like the schema.contao.org data, this approach makes perfect sense IMO.

Toflar:
I basically disagree with most of what you wrote because if I'd agree, I wouldn't have taken the time to work on this POC in the first place.
I absolutely **need** a central place where I can access the metadata and change **all of it** before it's output. Having everything in a central place allows me to extend existing data, remove data and even add completely new data.
If it's scattered all over the place in templates, I have to adjust way too many things. I cannot even provide an extension without asking people to adjust `n` templates.

However, I have now updated the PR to also allow to manage the JSON-LD data from within the templates. This allows everybody to customize the output and add additional meta data. Whether or not you put that in the template or hardcode it in your modules/elements etc. is up to the developers of the extensions.
Look at the news templates now. They can be easily extended and customized but we still have the advantage of it all being collected at one central point -> dispatch events -> and only then output the data.

> This POC doesn’t seem to connect the image data with the associated parent entity. Our current markup-based approach does that, so that the image of a news article is connected to the article it appears in.

That's fixed now.

ausi:
@Toflar enlightened me on slack ☺️ summary:

Points 3 (easy to customize in templates) and 4 (Less complex for non-developers) and fixed now.

Points 1 and 2 are basically features of this PR because it is hard to add metadata in the microdata format that is not visible on the page. It might need a little more attention from template editors, but it should also be way easier to understand because the HTML for the user and the additional metadata for search engines are separated now.

---

> Do we even adjust the current templates and remove the schema.org HTML markup or do we duplicate the information

We should remove the microdata then I think.

> Should we dispatch another event in the `AddJsonLdListener` so everybody can still add JSON-LD data one final time before it's added to the HTML.

IMO yes.

Is injecting the json-ld via response event the best fit here? How about adding it to `fe_page` instead?

m-vo:
I did not think everything from this PR though yet but :+1: in general.

I'm not 100% sure about the "do it in the template" approach, especially because it would need to be duplicated for Twig. :upside_down_face:  But I get @ausi's point and tend to agree. Maybe we could have a factory for JSON LD data and use it in both a thin Contao and Twig template function or use shortcuts like  `$figure->getJsonLd()` to ease things.

Regarding the event (although factored out for now): I like this idea :+1: . Funny enough I had a similar/related need where I wanted to adjust the image metadata landing in the `Figure`s at a central place. We would probably need to pass the `FigureBuilder` instance instead of the already built `Figure` for this, though (because it's virtually immutable).

leofeyer:
> > Do we even adjust the current templates and remove the schema.org HTML markup or do we duplicate the information
> 
> We should remove the microdata then I think.

Can we be sure that this does not affect screen readers or other assistive tools?

Toflar:
A little update on this one:

* @m-vo will you create a separate PR then? I agree it would be a useful event although not required for this PR anymore.
* @leofeyer JSON-LD is [the preferred way to add meta data to a page](https://www.google.com/search?q=google+json+ld+preferred) so I expect them to be able to handle that, yes. It's also way easier to parse and understand.
* To everyone: I've talked to @aschempp about this today for 2 hours and he's in favour of that JSON-LD solution as well (some remarks which I will work on next week). The major pain point we were discussing is what I currently named `PageMetadataContainer` and we've discussed how to get rid of our `$GLOBALS` for JS, CSS etc. and page title again. We think we have a pretty neat concept which will also play nicely with [his work on the new page controllers](https://github.com/contao/contao/pull/2015) :-) I'll create a separate PR for this first so we can - hopefully - get this in first and then build on top of that.

leofeyer:
> @leofeyer JSON-LD is the preferred way to add meta data to a page so I expect them to be able to handle that, yes. It's also way easier to parse and understand.

I think we've been doing the job long enough to know that our expectations are often different from the reality; especially when it comes to adopting web standards. 🤷🏻‍♂️ 

@ausi @MacKP Do you know anything about this?

MacKP:
I must say: I am a fan of schema.org direct in the HTML. But on the other Hand it can be a little bit ugly ;-) I see no Problem with changing it to JSON-LD (the only thing could be, that People can forget to change these things parallel to the Template stuff...).
And for Screenreaders this should be no Problem as far as i know. I've tested it with NVDA. The key is (as always) to have propper HTML for the screenreader and that's all. The extra stuff for machines isn't important here.
Kind regards

Toflar:
This PR now builds on top of https://github.com/contao/contao/pull/2975 which means that it includes all the code of it as well.
So if you want to compare them, here's the link: https://github.com/Toflar/contao/compare/feature/response-context...Toflar:feature/proper-page-metadata

As you can see, this PR now extends the `WebpageContext` and adds the `JsonLdManager` on top of it (the reason why we need the providers/factories). And it's super easy to access the current response context to add JSON-LD (or not) :-)

@m-vo the template functions are now super easy and can be easily replicated in Twig. But I agree, we should also build internal reusable stuff - I'll have a look at it again :)

fritzmg:
I also much prefer JSON-LD metadata over markup. It's a much cleaner way of providing as much information as possible to whichever service wants to process that data. So 👍 for me here, especially with the new ResponseContext approach :)

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -124,6 +125,19 @@ public function getMetadata(): Metadata
...
+    {
+        $schemaBase = $request ? $request->getSchemeAndHttpHost() : '';
+
+        $jsonLd = [
+            '@type' => 'ImageObject',
+            'identifier' => $schemaBase.'#/schema/image/'.$this->getImage()->getImageSrc(), // TODO: replace with UUID once this is available
+            'contentUrl' => $this->getImage()->getImageSrc(),
+        ];
+
+        return array_merge($jsonLd, $this->getMetadata()->getJsonLd('ImageObject'));
```
m-vo: We should check if metadata exists, otherwise there would be an exception:

```suggestion
        if(!$this->hasMetadata()) {
            return $jsonLd;
        }

        return array_merge($jsonLd, $this->getMetadata()->getJsonLd('ImageObject'));
```

core-bundle/src/File/Metadata.php
```diff
@@ -36,12 +36,27 @@ class Metadata
...
+     * @var array<string, array>
+     */
+    private $jsonLd;
+
     /**
      * @param array<string, mixed> $values
+     * @param array<string, array> $jsonLd
      */
-    public function __construct(array $values)
+    public function __construct(array $values, array $jsonLd = null)
```
Toflar: `Metadata` is immutable and its internal representation must not necessarily represent its JSON-LD representation. It is file meta data so the structure depends on whether you want it to be output as an `ImageObject` or e.g. a `MediaObject`. This logic will allow developers full control over the meta data if they want to, thanks to https://github.com/contao/contao/pull/3019.

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -124,6 +132,29 @@ public function getMetadata(): Metadata
         return $this->metadata;
     }
 
+    public function getJsonLd(): array
+    {
+        $imageIdentifier = $this->getImage()->getImageSrc();
+
+        if ($this->hasMetadata() && $this->getMetadata()->has(Metadata::VALUE_UUID)) {
+            $imageIdentifier = '#/schema/image/'.$this->getMetadata()->getUuid();
```
Toflar, line 132: I've searched the web on what's the best way to give unique identifiers to things like images. By definition, it should ideally be an URI because that "namespaces" it automatically. E.g. if we used something else as an identifier (https://schema.org/identifier) like e.g. `PropertyValue` we would end up having potential duplicates all over the web. So it's best to use an URI which points to the current website.
That's one thing. The second is that we can use our own UUID as a proper identifier if available. This is better than `getImage()->getImageSrc()` because that contains the `assets/images` URL and might change over time. But our UUID only exists if the file was synchronized, hence the fallback to the URL of the image.
We can use the `#/schema/image` approach for the UUID. The format doesn't really matter - the string just has to be unique.
This approach is recommended by the SEO experts at Yoast: https://developer.yoast.com/features/schema/pieces/image/#extended-criteria

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -124,6 +132,29 @@ public function getMetadata(): Metadata
...
+            '@type' => 'ImageObject',
+            'identifier' => $imageIdentifier,
+            'contentUrl' => $this->getImage()->getImageSrc(),
+        ];
+
+        if (!$this->hasMetadata()) {
+            return $jsonLd;
+        }
+
+        return array_merge($jsonLd, $this->getMetadata()->getJsonLd('ImageObject'));
```
Toflar: With this approach, you don't even need the JSON-LD event if you want to add data to the `ImageObject`. You can use the event introduced in https://github.com/contao/contao/pull/3019 and it will all work out of the box :)

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -392,6 +396,58 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
 		return System::getContainer()->get('translator')->trans($strId, $arrParams, $strDomain);
 	}
 
+	/**
+	 * Decodes a value for raw usage (DO NOT USE THIS IN HTML OUTPUT IF THE SOURCE IS UNKNOWN)
+	 */
+	public function raw(string $value, bool $stripTags = true): string
+	{
+		// TODO: Use StringUtil::getRawDecodedValue() once https://github.com/contao/contao/pull/3014 is merged.
```
Toflar: This remains a TODO but shows the importance of https://github.com/contao/contao/pull/3014! @ausi did a wonderful job there! <3

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -44,9 +62,34 @@ public function createResponseContext(): ResponseContext
 
     public function createContaoWebpageResponseContext(PageModel $pageModel): ContaoWebpageResponseContext
     {
-        $context = new ContaoWebpageResponseContext($pageModel);
+        $context = new ContaoWebpageResponseContext(new JsonLdManager($this->eventDispatcher), $pageModel);
```
Toflar: This shows a design issue in the `ResponseContext`. It's not possible to extend an existing context without knowing what constructor arguments it needs. We might still require composition here in the future but that's not part of this PR - it just shows we should think about it (I will, in a follow up PR).

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -44,9 +62,34 @@ public function createResponseContext(): ResponseContext
...
         $this->responseContextAccessor->setResponseContext($context);
 
         return $context;
     }
+
+    public function onJsonLd(JsonLdEvent $event): void
```
Toflar: I figured this is actually a beautiful design desicion. The factory itself knows that if someone created a `ContaoWebpageResponseContext`, it should add its own JSON-LD schema directly. So no page controller ever needs to worry about the JSON-LD schema for `schema.contao.org`. As soon as it creates a `ContaoWebpageResponseContext`, the JSON-LD schema for it will be ensured - all it needs to do is to output it.

core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return 'https://schema.contao.org/';
+    }
+
+    public function getType(): string
+    {
+        return 'contao:Page';
+    }
+
+    public function toArray(): array
```
Toflar, line 75: I had to override the core functionality here because `spatie/schema-org` does not support the namespaced format. But it's no problem at all here because that does not need to be extensible - it's our own schema and only needed for our search engine. What needs to be exensible is the schema.org stuff and that perfectly works :)

core-bundle/src/Routing/ResponseContext/JsonLdProvidingResponseContextInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Routing\ResponseContext;
+
+use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
+
+interface JsonLdProvidingResponseContextInterface extends ResponseContextInterface
```
Toflar: Alternative naming proposals are welcome :)

news-bundle/src/Resources/contao/templates/news/news_latest.html5
```diff
@@ -1,22 +1,22 @@
 
-<div class="layout_latest arc_<?= $this->archive->id ?> block<?= $this->class ?>" itemscope itemtype="http://schema.org/Article">
+<div class="layout_latest arc_<?= $this->archive->id ?> block<?= $this->class ?>">
```
Toflar: schema.org data only makes sense in detail templates. There is no such thing as describing lists. This data has been irrelevant ever since. What matters is that the search engines know if a page is describing a product or a news article etc. - the list views themselves do not matter.
Also see https://developers.google.com/search/docs/data-types/product
> We recommend focusing on adding markup to product pages instead of pages that list products or a category of products.

Toflar (commented):
This PR is ready for a first review to see if you like the approach. I'll then adjust and finish the stuff with tests. I've reviewed the PR myself and added comments to explain the thoughts behind some design decisions - I hope that makes things clearer :)

Toflar:
Initial PR description updated because lots of it has become irrelevant - also most comments in this PR do not apply anymore. Just keep that in mind while reading through the stuff :)

core-bundle/src/File/Metadata.php
```diff
@@ -118,4 +133,28 @@ public function empty(): bool
...
+    public function getJsonLd(string $type = null): array
+    {
+        if (null === $type) {
+            return $this->jsonLd;
+        }
+
+        return $this->jsonLd[$type] ?? [];
+    }
+
+    public static function extractBasicJsonLd(self $metadata): array
```
Toflar: I made this `public static` on purpose so that if you want to access this somewhere else, you can. We can also make this `private` though.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
aschempp: Shouldn't that be its own event listener? `Contao\CoreBundle\EventListner\JsonLdSchema\ContaoPageSchemaListener.php`

Toflar: I don't know why you ask me this at the exact spot where I wrote that I find this "a beautiful design decision"?
So again, no I don't think it should go to its separate listener because again, I find this a beautiful design decision to have the things together here.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: https://github.com/contao/contao/pull/3032 will solve this.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
aschempp: so I guess the beauty is that the same class handles creating and updating the request context? Because the fact that a `ContaoWebpageResponseContext` would be handled automatically would be the same with a listener. The factory (in your implementation) does not _know_ about that `ContaoWebpageResponseContext`, it just handles all of them. If your intention is for the factory to actually only handle what's been created _through_ it, we would need to store some object hash map to only handle its own. Not sure if that's worth anything though …

comments-bundle/src/Resources/contao/classes/Comments.php
```diff
@@ -293,12 +293,6 @@ protected function renderCommentForm(FrontendTemplate $objTemplate, \stdClass $o
 
 			if ($flashBag->has('comment_added'))
 			{
-				/** @var PageModel $objPage */
-				global $objPage;
-
-				$objPage->noSearch = 1;
```
Toflar, line 299: No need to set `noSearch` here. The template is exempt from the indexer anyway and disabling the cache is not needed as the session is started.

Toflar: All the other appearances will be removed with https://github.com/contao/contao/pull/3085.

core-bundle/src/EventListener/ContaoJsonLdSchemaListener.php
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        /** @var HtmlHeadBag $htmlHeadBag */
+        $htmlHeadBag = $responseContext->get(HtmlHeadBag::class);
+
+        /** @var ContaoPageSchema $schema */
+        $schema = $event->getJsonLdManager()->getGraphForSchema(JsonLdManager::SCHEMA_CONTAO)->get(ContaoPageSchema::class);
+
+        $schema->updateFromHtmlHeadBag($htmlHeadBag);
```
Toflar, line 46: Right now, this is only doing `setTitle()` but who knows. Feels better like that.

core-bundle/src/Image/Studio/Figure.php
Toflar: @contao/developers What's your take on this. According to schema.org, it has to be `URL`. But it seems to be also valid when it's relative so we could maybe omit the request being injected?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -327,6 +327,10 @@ public function __set($strKey, $varValue)
 					case 'robots':
 						$htmlHeadBag->setMetaRobots($varValue);
 						break;
+
+					case 'noSearch':
+						$htmlHeadBag->setMetaRobots($varValue ? 'index,nofollow' : 'noindex,nofollow');
```
Toflar: Makes no sense to differ between `follow` and `nofollow` here. Imho if something is dynamically set to `noindex` (which I consider to be not a use case anymore - the core doesn't need it itself), it's irrelevant whether it was set to be followed or not.

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -215,6 +215,12 @@ protected function prepare($objPage)
 		// Meta robots tag
 		$this->Template->robots = $responseContext->get(HtmlHeadBag::class)->getMetaRobots();
 
+		// Do not search the page if the query has a key that is in TL_NOINDEX_KEYS
+		if (preg_grep('/^(' . implode('|', $GLOBALS['TL_NOINDEX_KEYS']) . ')$/', array_keys($_GET)))
```
Toflar, line 227: I would like to get rid of this too. Makes no sense. It seems we're using this to somehow prevent indexing for filtering (pagination, calendar range filters etc.) rather than correctly setting the right canonical link which would have been the right solution ever since. But that's for another PR thus that remains here for now.

Toflar:
@contao/developers this is ready for review. I've added loads of comments to explain my ideas and completely got rid of the news-bundle adjustments. For now, only our own schema.contao.org plus image schema meta data is added. It should make the review process easier.
Tests will only follow once concept is approved :)

core-bundle/src/Image/Studio/Figure.php
aschempp: Since the image is always available and the same on all domains, using a relative path shouldn't be a problem?

core-bundle/src/Resources/contao/models/PageModel.php
aschempp: If `noSearch === true` it sets `index/nofollow`? And `noindex...` if `noSearch === false`?

Toflar: Yeah that's wrong of course.

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: I also don't understand why indexing a pagination (e.g. product list) wouldn't make sense... 🤷‍♂️

core-bundle/src/Resources/contao/templates/image/image.html5
```diff
@@ -14,3 +14,7 @@
     <figcaption class="caption"><?= $this->caption ?></figcaption>
   <?php endif; ?>
 </figure>
+
+<?php
+
+$this->jsonLd($this->figure->getJsonLd());
```
aschempp: This would probably fail if the image template is used outside the figure builder?

core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
```diff
@@ -0,0 +1,99 @@
...
+     * @var bool
+     */
+    private $fePreview;
+
+    /**
+     * ContaoPageSchema constructor.
+     *
+     * @param array<int> $groups
+     */
+    public function __construct(string $title, int $pageId, bool $noSearch, bool $protected, array $groups, bool $fePreview)
```
aschempp, line 55: Can't we pass the PageModel here?

core-bundle/src/Resources/contao/templates/image/image.html5
Toflar: `$this->figure` always exists and yes, if you do not have it, you cannot use it.

core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
Toflar: No, because of the whole encoding/decoding stuff.

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: Fixed.

core-bundle/src/Image/Studio/Figure.php
Toflar: It's relative now. https://validator.schema.org validates perfectly fine.

core-bundle/src/Resources/contao/templates/image/image.html5
Toflar: You mean you want to wrap it into an `if ($this->figure)`?

m-vo: > $this->figure always exists

This is a rather recent addition, though. It might fail for people that use `$this->insert('image')` from legacy contexts, so I'd rather check for it as well.

Toflar: Added in d273e2a48412e77de9de705db7b1b0f5d6f8f26c.

core-bundle/src/File/Metadata.php
```diff
@@ -124,4 +134,39 @@ public function empty(): bool
...
+        if ($this->has(self::VALUE_CAPTION)) {
+            $this->jsonLd['AudioObject']['caption'] = $this->getCaption();
+            $this->jsonLd['ImageObject']['caption'] = $this->getCaption();
+            $this->jsonLd['MediaObject']['caption'] = $this->getCaption();
+        }
+
+        if ($this->has(self::VALUE_LICENSE)) {
+            $this->jsonLd['AudioObject']['license'] = $this->getCaption();
+            $this->jsonLd['ImageObject']['license'] = $this->getCaption();
+            $this->jsonLd['MediaObject']['license'] = $this->getCaption();
```
ausi:
```suggestion
            $this->jsonLd['AudioObject']['license'] = $this->getLicense();
            $this->jsonLd['ImageObject']['license'] = $this->getLicense();
            $this->jsonLd['MediaObject']['license'] = $this->getLicense();
```

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -327,6 +327,10 @@ public function __set($strKey, $varValue)
 					case 'robots':
 						$htmlHeadBag->setMetaRobots($varValue);
 						break;
+
+					case 'noSearch':
+						$htmlHeadBag->setMetaRobots($varValue ? 'noindex,nofollow' : 'index,nofollow');
```
ausi: Shouldn’t this modify the JSON-LD `noSearch` for schema.contao.org/Page instead of setting the meta robots?

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -215,6 +215,12 @@ protected function prepare($objPage)
...
+		{
+			$this->Template->robots = 'noindex,nofollow';
```
ausi: Same here, I think this should set the JSON-LD 'noSearch' attribute.

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -395,6 +398,50 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
...
+		$jsonLdManager
+			->getGraphForSchema($schema)
+			->set($type, $id);
+	}
+
+	/**
+	 * Finalizes the JSON-LD data collected during the current response context
+	 * and returns the scripts ready to be output within HTML context.
+	 */
+	public function finalizeJsonLd(): string
```
ausi: I would prefer removing the `Template::finalizeJsonLd()` method and set the script string explicitly in `PageRegular` to something like `$this->Template->jsonLd` (we do the same/similar with the `HtmlHeadBag` IMO).

core-bundle/src/File/Metadata.php
```diff
@@ -38,11 +38,21 @@ class Metadata
     private $values;
 
     /**
-     * @param array<string, mixed> $values
+     * JSON-LD data where the key matches the schema.org
+     * type.
+     *
+     * @var array<string, array>
+     */
+    private $jsonLd;
```
ausi: The `Metadata::$jsonLd` property is now hardwired with schema.org. Should we change this to something like the following to be open to support other schemas in the future?

```
[
	{
		"@type": "https://schema.org/MediaObject",
		"name": …,
		"caption": …,
		/…
	},
	…
]
```

core-bundle/src/Event/JsonLdEvent.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Event;
+
+use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
+use Symfony\Contracts\EventDispatcher\Event;
+
+class JsonLdEvent extends Event
```
ausi: Should we name the `JsonLdEvent` more specific like `JsonLdFinalizeEvent` or `CollectFinalJsonLdEvent`?

core-bundle/src/Resources/contao/library/Contao/Template.php
aschempp: the idea was that this could be relevant for other controllers as well, e.g. an ESI fragment should render its own JSON-LD

core-bundle/src/File/Metadata.php
Toflar: That would make things quite a bit more complicated for something we don't know if it'll ever happen. schema.org has been around since 2011 and there's no sign whatsoever for an alternative to pop up. We can still implement that when the day comes (by adding the requested schema to as an argument to the `getJsonLd()` method or so.

core-bundle/src/Event/JsonLdEvent.php
Toflar: I named it `JsonLdEvent` because I cannot think of any use case for another JSON-LD related event at all. I don't really like the proposals tbh.

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: It would be way more complicated so I decided against it. Especially because the use case to set `noSearch = true` has never been "disable the internal search only". It has always been "disable search indexing". All the cases I've encountered where we used `->noSearch` (which are all gone in core now so the use case is questionable anyway), we wanted to exempt the response from being indexed by **any** search engine. 
Hence, my decision to not go down the complex road. It's not worth it.

fritzmg: I have to disagree here, we specifically distinct in the page settings between robots and Contao's internal search engine, i.e. you can set a page to `noindex`, but still allow it to be indexed by Contao's search index and you can set it to `index` and disallow the page to be indexed by Contao's search index.

core-bundle/src/Resources/contao/library/Contao/Template.php
ausi: And shouldn’t the fragment controller then itself just call `$responseContext->get(JsonLdManager::class)->collectFinalScriptFromGraphs()` and give it to its template?

core-bundle/src/Resources/contao/models/PageModel.php
ausi: I’m pretty sure that setting all pages that match one of the `TL_NOINDEX_KEYS` to `noindex,nofollow` would have very bad effects for SEO.

Toflar: It's changed now.

core-bundle/src/Resources/contao/library/Contao/Template.php
Toflar: No the template defines where it wants to output the JSON-LD.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -52,6 +73,11 @@ public function has(string $serviceId): bool
         return isset($this->current[$serviceId]);
     }
 
+    public function isInitialized(string $serviceId): bool
```
Toflar, line 76: I've introduced that to improve performance. When setting data, this is not relevant (`has()` is enough). But if you want to call methods on a service like we do when we finally render the JSON-LD, we need this. This is to cover the use case where the controller has provided the `JsonLdManager` with `addLazy()` but no module/content element or whatever ever accessed it (= it was never initialized).
We would then intantiate the service only to see that it was empty. With the additional `isInitialized()` we can improve this :)

core-bundle/src/Resources/contao/library/Contao/Template.php
ausi: Agree, it should define *where* to output the script, but it schould not generate it itself IMO.

Toflar (commented):
New event logic. The event dispatcher is now part of the response context as well so any other service can easily dispatch response context events :) Developed that concept with @aschempp this morning - this is a lot cleaner now :)

core-bundle/src/Resources/contao/library/Contao/Template.php
Toflar: Well it's just a function to make things lazy. The same as we use in the `news` for `text` (https://github.com/contao/contao/blob/4.x/news-bundle/src/Resources/contao/modules/ModuleNews.php#L124) for example. Otherwise it would always be generated even if not used. Maybe it's just the name or the fact that it's a function call that puzzles you? What about `<?= $this->jsonLdScripts ?>`?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -327,10 +336,19 @@ public function __set($strKey, $varValue)
...
+			if ('noSearch' === $strKey && $responseContext->has(JsonLdManager::class))
+			{
+				/** @var JsonLdManager $jsonLdManager */
+				$jsonLdManager = $responseContext->get(JsonLdManager::class);
+
+				if ($jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_CONTAO)->has(ContaoPageSchema::class))
+				{
+					/** @var ContaoPageSchema $schema */
+					$schema = $jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_CONTAO)->get(ContaoPageSchema::class);
+					$schema->setNoSearch($varValue);
```
```suggestion
					$schema->setNoSearch((bool) $varValue);
```
should we cast to boolean here?

core-bundle/src/Resources/contao/library/Contao/Template.php
ausi: What I don’t like, is that it is a method on the Template class that accesses the “global” container to get and generate data.

> The same as we use in the `news` for `text`

There the news module sets (or injects) the `text` property to the template. If we do this in `PageRegular` with something like `$this->template->jsonLdScripts = function () {…}` I would be totally happy ☺️

Toflar: But then everybody needs to copy that code?

ausi: They have to create the response context themselves, and therefore should know if it has a JsonLdManager and it would only be a single line of code then:

```
$responseContext->get(JsonLdManager::class)->collectFinalScriptFromGraphs()
```

Toflar: That would look like c40216ffa28991da2087fc8cbcaca45c8fd6d335 then. I don't mind, is that okay with you too @aschempp ?

ausi: We could pass `$responseContext` as a parameter to `createFooterScripts()` so that we don’t have to ask the container. But I’m fine with how it is now too. 🎉

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Template.php
Toflar: We can't because they were extensible. But I've fixed that with 6664322671c4c34587b015118413b2a18b2a653c now.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Template.php
ausi: approved ✅

ausi (dismissed):

core-bundle/src/Resources/contao/templates/image/image.html5
```diff
@@ -14,3 +14,9 @@
     <figcaption class="caption"><?= $this->caption ?></figcaption>
   <?php endif; ?>
 </figure>
+
+<?php
+
+if ($this->figure) {
+    $this->jsonLd($this->figure->getJsonLd());
+}
```
fritzmg, line 22: Personally I would prefer having

```php
<?php

if ($this->figure) {
    $this->jsonLd($this->figure->getJsonLd());
}

?>
```

at the beginning of the template, but since we do not really have a convention for that (yet) I am fine with this. Only question remains: since templates use two spaces for indentation, should PHP code within `<?php ?>` in templates also use two spaces or four spaces like here?

@m-vo how would the implementation look in an `image.html.twig`?

Toflar: I prefer it to be at the end because it's meta data. and the indendation is set to 4 by the cs fixer :)

core-bundle/src/EventListener/ContaoJsonLdSchemaListener.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Event\JsonLdEvent;
+use Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag\HtmlHeadBag;
+use Contao\CoreBundle\Routing\ResponseContext\JsonLd\ContaoPageSchema;
+use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
+
+/**
+ * Updates the schema.contao.org schema before rendering it
```
aschempp, line 21: sure this is `schema.contao.org`?

core-bundle/src/Resources/contao/templates/image/image.html5
m-vo: There is the `do` tag in Twig to execute things. We would need a `jsonLd` template function and then write this:

```twig
{% do jsonLd(figure.jsonLd) %}
```

@toflar We should IMO add the Twig function to this PR as well. I can contribute this if you want to.

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -395,6 +399,32 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
 		return System::getContainer()->get('translator')->trans($strId, $arrParams, $strDomain);
 	}
 
+	/**
+	 * Adds JSON-LD data to the current page.
+	 */
+	public function jsonLd(array $jsonLd, string $schema = JsonLdManager::SCHEMA_ORG): void
```
m-vo: What about naming this `addJsonLd()` instead? To me that would sound less as it would return something.

core-bundle/src/Resources/contao/templates/image/image.html5
Toflar: No, I don't want any Twig here please. There's a gazillion other places where we will be able to add JSON-LD and make use of the new functionality. Those are future PR's. Feel free to create one as soon as this is merged :)

fritzmg: > and the indendation is set to 4 by the cs fixer :)

Ah ok, didn't realise the CS fixer already covers this :)

m-vo: > There's a gazillion other places where we will be able to add JSON-LD.

Then this should imo become a service and not live inside `Template`.

Toflar: Yes, which it already is. It's a one-liner.

fritzmg (dismissed):

core-bundle/src/EventListener/ContaoJsonLdSchemaListener.php
```diff
@@ -0,0 +1,48 @@
...
+ * with the current data from the HtmlHeadBag.
+ */
+class ContaoJsonLdSchemaListener
```
aschempp, line 24: This is only about the `ContaoPageSchema` or is the listener supposed to handle any other Contao-related JSON schema as well?

core-bundle/src/Resources/contao/library/Contao/Template.php
aschempp: I feel like `addSchemaData` would be a better name. Since Contao does not support any other Schema type the whole JsonLD stuff feels a bit superfluous.

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -692,39 +714,26 @@ protected function createFooterScripts($objLayout, $objPage = null)
...
-
 		// Add the custom JavaScript
 		if ($objLayout->script)
 		{
 			$strScripts .= "\n" . trim($objLayout->script) . "\n";
 		}
 
 		$this->Template->mootools = $strScripts;
+
+		$this->Template->jsonLdScripts = function ()
```
aschempp, line 725: this will not work if someone adds JSONLD data in the fe_page (or any included) templates. It will also be a problem when insert tags are being replaced on the response… maybe we need to lazy-generate the JSONLD?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -52,6 +73,11 @@ public function has(string $serviceId): bool
...
+    {
+        return !$this->services[$serviceId] instanceof \Closure;
```
aschempp, line 82: This should also check if the service exists at all (`isset()`)

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -70,7 +96,7 @@ public function get(string $serviceId)
         $serviceId = $this->current[$serviceId];
 
         // Lazy load the ones with factories
-        if ($this->services[$serviceId] instanceof \Closure) {
+        if (!$this->isInitialized($serviceId)) {
```
aschempp, line 103: this won't work if the service does not exist at all. I would keep the old line

aschempp (changes requested):

core-bundle/src/EventListener/ContaoJsonLdSchemaListener.php
Toflar: Hm? Yes?

core-bundle/src/EventListener/ContaoJsonLdSchemaListener.php
Toflar: Only the `ContaoPageSchema`.

core-bundle/src/Resources/contao/library/Contao/Template.php
Toflar: Disagree but well

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: /cc @ausi?

ausi: > this will not work if someone adds JSONLD data in the fe_page (or any included) templates.

Why? The callback function is only executed at the point where it gets inserted into the template and as this is just before `</body>` all other included template code would already have been executed.

> It will also be a problem when insert tags are being replaced on the response…

Same with page title or everything else so IMO a known limitation.

Toflar:
I've renamed the methods so they do more accurately represent that they only work for schema.org. The JsonLdManager can manage other schema too but the methods in `Template`, `Figure` and `Metadata` don't. 

Also added unit tests for full test coverage. 😅 
This can be reviewed now!

fritzmg (dismissed):

ausi (dismissed):

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -23,6 +26,11 @@
  */
 class PageRegular extends Frontend
 {
+	/**
+	 * @var ResponseContext
+	 */
+	protected $responseContext;
```
aschempp, line 32: could someone possible change the response context and this would unintentionally keep state in the object?

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: I suggest to create a separate issue to discuss the `TL_NOINDEX_KEYS`?

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: we use `[[TL_BODY]]` and `[[TL_MOOTOOLS]]` etc. to support this, because otherwise e.g. a `{{insert_content::X}}` could not add global scripts. I suggest we use a `[[TL_JSONLD]]` to achieve the same? Maybe keep state in the class whether the closure was called at all and JSONLD should be replaced?

aschempp (commented):
Looks very good! Only a few minor comments 😊

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: I have to allow child classes access to it and there's no way for a readonly property in PHP. So yes, that's possible. But they can also override all sorts of other methods and cause their version of PageRegular to not work anymore so I don't see any problem here.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: In which case we're back to `str_replace` and could also have a general `FinalizeResponseContextEvent` instead of `JsonLdEvent` which we explicitly did not want.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Done https://github.com/contao/contao/issues/3097

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: no I didn't mean that, I mean if someone switched the request context in the request. Maybe it would be better to access a "fresh copy" of the request context in every method that uses it instead of keeping state.

Toflar: No. It's the controller. It sets and ends the state. If a child class wants to tamper with it, then that's the fault of the child class. Simple as that.

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: I don't think thats the same. We would stick to existing implementation details, but the template decides where and if the JSON is added.

aschempp: But thats ultimately up 2 u. I can guarantee this will come up pretty quickly and will require a fix, but the fix would still be easy and we can merge as is 😊

aschempp (dismissed):

core-bundle/src/Resources/contao/templates/image/image.html5
leofeyer: I would also prefer to have this at the beginning of the template, because that is where we usually add metadata such as `$GLOBALS['TL_CSS']` etc. @Toflar Mind if I change this?

Toflar: The reason I put it at the bottom is because it might get a lot longer than just one line in other templates. With various `if` statments. Dunno if you remember the adjusted `news_full` template but it looked (and will look as I will provide PRs that actually use the new feature in other places soon) something like this:

```php
<?php

$jsonLd = [
    '@type' => 'NewsArticle',
    'identifier' => $this->absoluteLink,
    'url' => $this->absoluteLink,
    'headline' => $this->raw($this->headline),
    'datePublished' => $this->datetime,
];

if ($this->hasTeaser) {
    $jsonLd['description'] = $this->raw($this->teaser);
}

if ($this->hasText()) {
    $jsonLd['text'] = $this->raw($this->text);
}

if ($this->addImage) {
    $jsonLd['image'] = $this->figure->getJsonLd();
}

if ($this->authorModel) {
    $jsonLd['author'] = [
        '@type' => 'Person',
        'name' => $this->authorModel->name,
    ];
}

$this->jsonLd($jsonLd);
```

So you'd want to have this whole block bevore any HTML? Because for consistency reasons, we should handle all templates the same.

fritzmg: It could be shortened to

```php
<?php

$this->jsonLd(array_filter([
    '@type' => 'NewsArticle',
    'identifier' => $this->absoluteLink,
    'url' => $this->absoluteLink,
    'headline' => $this->raw($this->headline),
    'datePublished' => $this->datetime,
    'description' => $this->hasTeaser ? $this->raw($this->teaser) : null,
    'text' => $this->hasText() ? $this->raw($this->text) : null,
    'image' => $this->addImage ? $this->figure->getJsonLd() : null,
    'author' => $this->authorModel ? ['@type' => 'Person', 'name' => $this->authorModel->name] : null,
]));
```

but I see your point.

Toflar: I know but remember that regular users are editing templates. That's why I would not use `array_filter()` stuff here. It sould be as simple as it can get so everybody understands what it does.

leofeyer: Is it unavoidable to have this code in a template? Or could we encapsulate it in a helper class or something alike?

fritzmg: It's there so that you can easily change or extend it by altering the template - just as you would be able to before the JSON-LD management by editing the HTML tags and attributes.

Toflar: We may sure think about also providing a similar helper like we did on `Figure` in this PR on `NewsModel` so the basic data can be reused in a Twig template as well. However, the whole purpose of having the `addSchemaOrg()` method in the template is for it to be fully customizable. So the idea really is that you can modify the array to your personal needs and this might easily exceed a few lines of code. That's why I placed it at the end and I still prefer this. The primary use case of the template is to manage the HTML. That's gonna be the 90% use case. So HTML should be first in the template.

m-vo: Maybe it would be easier to understand in templates if the method would (additionally) allow to define the entries one by one?

```php
$this->jsonLd('@type', 'NewsArticle');
$this->jsonLd('url', $this->absoluteLink);
```

```twig
{% do addJsonLd('@type', 'NewsArticle') %}
{% do addJsonLd('url', absoluteLink) %}
```

m-vo: Another thing I was wondering: How is inheritance affected by this? In a child template everything outside blocks is ignored, so you could only alter JSON-LD data inside any block. It could neither be at the top or bottom of the document then.

Workaround would be to always put the call(s) into blocks in the first place.

fritzmg: > In a child template everything outside blocks is ignored

Only after `$this->extend()` - so any PHP code before that is still evaluated.

m-vo: That's not the case in Twig, though, and trying to support it would be unnecessarily complex.

m-vo: Just checked: `do` works outside blocks as it does not emit any output.

core-bundle/src/Resources/contao/library/Contao/Template.php
m-vo: Was this closed intentionally? :slightly_smiling_face:

Toflar: Yes because I renamed it.

m-vo: Sorry, missed that!

leofeyer:
Two questions remain on my side:

> I've adjusted news templates and ModuleNews to showcase how adding JSON-LD data now works.

1. Where do I find these changes? They are not in the PR, are they?

2. The HTML markup still includes a lot of `schema.org` tags (see below). Will they be centralized as well?

* `itemtype="http://schema.org/SiteNavigationElement"`
* `itemtype="http://schema.org/Article"`
* `itemtype="http://schema.org/Event"`

Toflar:
>     1. Where do I find these changes? They are not in the PR, are they?

No, I've removed them so this PR can be better reviewed. I wrote that in some comment as well.



>     2\. The HTML markup still includes a lot of `schema.org` tags (see below). Will they be centralized as well?

We'll see. This PR allows for it :)

leofeyer (approved):

leofeyer:
Thank you @Toflar.

Toflar:
YESS!!! For everybody that did not follow this PR: This work (and all the related PRs which were worked on before) will unlock a whole new world for us developers! We'll be able to get rid of a lot more `$GLOBALS` stuff in the future and we'll be able to ship Contao setups with absolutely excellent schema.org data! 😊 🕺 🥳

PR #2967, opened 2021-04-22
Use the date formats from the page context for insert tags

fritzmg:
Currently the following insert tags will only fall back to the back end date and time formats if no specific date format is given in the insert tag:

* `{{date}}`
* `{{last_update}}`
* `{{user::*}}` if `*` is a `tl_member` DCA field with a `rgxp` of either `date`, `time` or `datim`.

In the front end context the date formats of the page should always be used when displaying dates.

_Note:_ this automatically falls back to the date formats of the back end (the local config), see [here](https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Resources/contao/models/PageModel.php#L1057-L1071).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -170,7 +170,7 @@ class InsertTags extends Controller
 			{
 				// Date
 				case 'date':
-					$arrCache[$strTag] = Date::parse($elements[1] ?: Config::get('dateFormat'));
+					$arrCache[$strTag] = Date::parse($elements[1] ?: $objPage->dateFormat ?: Config::get('dateFormat'));
 					break;
 
 				// Accessibility tags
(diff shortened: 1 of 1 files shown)


Toflar (approved):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -329,15 +329,15 @@ protected function doReplace($strBuffer, $blnCache)
 
 						if ($rgxp == 'date')
 						{
-							$arrCache[$strTag] = Date::parse(Config::get('dateFormat'), $value);
+							$arrCache[$strTag] = Date::parse($objPage->dateFormat, $value);
```
ausi, line 332: Should we use the config as a fall back if `$objPage` does not exist?
```suggestion
							$arrCache[$strTag] = Date::parse($objPage->dateFormat ?? Config::get('dateFormat'), $value);
```

ausi: ~~Or maybe fallback to `null` because the date class handles the fallback itself?~~ no it doesn’t

fritzmg: I was thinking about the same, but the rest of the code currently already relies on `$objPage` not being null. Though generally I am in favour of doing it this way.

ausi: I see, the `date` tag already seems to rely on the `$objPage` in the cache handling section. So I guess it’s OK then.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #2970, opened 2021-04-23
Load protected status from parent pages in customnav

patrickjDE:
The custom navigation module does render protected pages if not logged in, even with the "Show protected items" option disabled, if they do not have the protected status set on the page itself, but rather inherited the status from one of its ancestor pages.

This PR adds a call to `PageModel::loadDetails` to actually inherit the protected status and associated groups setting from parent pages before checking for them, to correctly hide protected pages.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleCustomnav.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleCustomnav.php
@@ -116,10 +116,10 @@ class ModuleCustomnav extends Module
 		/** @var PageModel[] $arrPages */
 		foreach ($arrPages as $objModel)
 		{
-			$_groups = StringUtil::deserialize($objModel->groups);
-
-			// Inherit protected status from the parent pages
+			// Inherit settings from the parent pages
 			$objModel->loadDetails();
+
+			$_groups = StringUtil::deserialize($objModel->groups);
 
 			// Do not show protected pages unless a front end user is logged in
 			if (!$objModel->protected || $this->showProtected || (\is_array($_groups) && \is_array($groups) && \count(array_intersect($_groups, $groups))))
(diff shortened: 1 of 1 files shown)


aschempp:
This is likely a big performance penalty, but I can't think of any other solution 😢

leofeyer:
I am not sure if this is correct. In the custom navigation, there is no page hierarchy and thus no inheritance – at least this is how the module was designed. Of course, we can discuss this and agree to change it in the future.

patrickjDE:
> In the custom navigation, there is no page hierarchy and thus no inheritance – at least this is how the module was designed.

That might have been the original intention, but it just cannot work. Which navigation module was used has no say in the protection status of a page.

When accessing the page in question, protected status from the page's hierachy is taken into account anyway, as you cannot tell from which navigation (if any) that request originated.

aschempp:
@leofeyer is right here. A page that is published is publicly available on the website, regardless of a published parent page. Therefore it should also appear in the custom navigation. The regular navigation module does not _show_ them by accident, because it does not render the parent node and therefore it's children.

patrickjDE:
> A page that is published is publicly available on the website, regardless of a published parent page.

~~Am I missing something here? 🤔~~

~~A page is not publicly accessible, if one of its parent pages are protected. It's in the docs, it works like this in the official demo, and did so since what feels like forever.~~

EDIT: Ah, yes, I misread. Didn't notice you were talking about a *published* parent page - which, in turn, is not what this PR is about, as @fritzmg already pointed out.

fritzmg:
> @leofeyer is right here. A page that is published is publicly available on the website, regardless of a published parent page.

@aschempp this is about the **protected** status, not the **published** status. I agree with @patrickjDE that pages that the front end user does not have access to should _not_ show up.

leofeyer:
As discussed in Mumble on May 12th, we want to merge this as is.

leofeyer (changes requested):
As discussed in Mumble on May 12th, we want to merge this PR. However, the `->loadDetails()` call should be above the `$_groups = StringUtil::deserialize($objModel->groups);` statement.

patrickjDE:
Updated the PR as requested

aschempp (approved):
Sorry about my confusion as well 😄 looks good now!

ausi (approved):

leofeyer:
Thank you @patrickjDE.

PR #2975, opened 2021-04-26
Add the response context

Toflar:
This is a POC (though thoroughly discussed with @aschempp so it's probably a bit more than just a POC :)) to have a new way to **add or change page meta data dynamically**.
This includes all the things like

* Changing the page tilte, description and robots meta content (included in this PR for demonstration purposes)
* Adding `<script>` tags on the fly (replacing `$GLOBALS['TL_JAVASCRIPT']`)
* Adding `<link>` tags on the fly (replacing `$GLOBALS['TL_CSS']` or `$GLOBALS['TL_HEAD']`)
* etc.

While discussing with @aschempp what all of this represents, we noticed that all of it is basically

> adding information to the future response which is not generated yet.

Examples:

* When you set `$objPage->pageTitle` you do so, because the `Response` has not been generated yet and you cannot set `<title>` right away.
* When you use `$GLOBALS['TL_JAVASCRIPT']` you do so, because there's no way for you to access the current `Response` and add this to the future HTML somehow because it's not even generated yet. 
* etc.

Moreover:

> You don't even know if a script, page title, page description etc. is even going to be used or if you potentially waste resources generating something that's then never required.

Hence, here's what we noticed: What we need is a **ResponseContext**.
A page controller cannot generate the `Response` for you beforehand but it can tell, what context is going to be applied and thus, what options you have:

* Is it a regular `WebpageContext` (terminology taken from schema.org) you're going to generate? So a typical HTML page? If so there sure is a `setTitle()` and a `setDescription()`.
* Is it an ESI fragment? It would probably use an `EsiContext` because `setTitle()` or `setDescription()` typically make no sense.

Proof (or rather, why we think) that this concept is going to work:

* It will play nicely with the future real page controllers, @aschempp is working on in https://github.com/contao/contao/pull/2015. The process for a page controller would be like so:
    1. Set the `ResponseContext` according to what the controller does. Will it output `XML` only? Then don't set it to `WebpageContext` because offering `setTitle()` and `setDescription()` makes no sense. But maybe something else makes sense? 
    2. Generate the modules, content elements (if you want to do that in your page type). Anything collected here will write to the `ResponseContext` if provided and implementing the correct interfaces.
    3. The resulting `ResponseContext` has to be applied to the final response by the page controller. It's entirely the developer's responsibility to apply this. In other words, as a developer you specify the `ResponseContext` you **are going to** to return. Then you generate code you **do not control** like third party modules which **have access to the current context and can thus decide what to do**. Finally, you apply the changes **you said you were going to pass on** to the final response.
* It also plays very nicely with the centralized JSON-LD PR proposed in https://github.com/contao/contao/pull/2962. I will update the PR shortly to show how this now works together beautifully. The context specifies that it understands JSON-LD which enables modules to access the `JsonLdManager` and add more information. And here's the beauty in this concept. Both, the `WebpageContext` as well as a possible `EsiContext` can implement the correct interface and thus **indicate to modules, elements, templates** that **they are going to** use this data. For the `WebpageContext` this is rather obvious because the data will be displayed before the closing `</body>` but JSON-LD works perfectly fine for an ESI fragment as well. You just have to make sure to output it inline. :-)

## Design decisions:

* The `ResponseContext` is deliberately optional. A page controller is not required to provide a `ResponseContext` at all. If there is none, then no module, content element etc. can add anything to the context which is the intended behaviour.
* There are multiple providers for different `ResponseContextInterface` implementations. These serve as factories. This enables container access, you'll see a use case in the JSON-LD PR where we need access to the EventDispatcher.
* The `WebpageContext` does not have any interface because I don't feel like it's worth it. I don't want to have 30 interfaces for every single feature a `ResponseContext` might support. We can still extract them in the future if we see that something can be re-used a lot across different contexts (such as JSON-LD in the other PR).

## Additional notes

* I've only replaced the deprecations in the `news-bundle` so far to show how easy it is to get rid of those and switch to the new context. We should ideally get rid of all deprecation notices but it's not a requirement to finish this PR.
* No unit tests so far as I would like to know if the Core team likes the concept first. I'll clean up afterwards.


leofeyer:
I like the concept. Way to go! 👍🏻

m-vo:
Nice concept! :heart: 

I've got a few questions/remarks, but also - very likely - I'm misunderstanding something. :relaxed: 

What about extending a context? Is this sth that should be supported? Take open graph tags for instance that I may want to set from within some content element fragments and or event listeners and that should be output in every regular page template. These should also live in the `WebpageContext`, right? If I get this correctly, I could in theory register a provider with a higher priority that also supports this context and returns an instance of something that extends the class. But that's not something multiple extensions could do. Also how would I render them to the response? Would the page controller e.g. issue a render event or would the context be available in the template?

> The WebpageContext does not have any interface

But this also means any implementer adding things to a context must know the actual type in advance, while an interface would allow composition, wouldn't it?

I wonder if the asynchronous nature of the problem could benefit from lazy evaluation (closures?) or events. Like: only evaluate a certain value as soon as it is requested. Then you would maybe set a property with `$context->setExpensiveValue(fn() => calculateValue())` or by implementing a `provideContextValue` listener.

Toflar:
> What about extending a context?

I think what you're misunderstanding here is that this use case does not exist :-D You just cannot extend a context.
The process is split up into three steps:

1. The controller announces its response context by setting it.
2. Anybody can access and use it accordingly.
3. The controller maps the context to the response (whatever that means).

So if you could dynamically extend the context, you would adjust step 1 but the controller wouldn't know what to do in step 3. 
I know we could provide ways to work around that and have something like initializers for step 1 and finalizers for step 3 but this would make the whole concept **massively** more complex and I also don't think that it will happen that often in the wild. Your open graph example should 100% go to `Webpage` and be contributed to the core. If you want to provide a custom way around it you can still set your own request attribute, like e.g. Symfony does for `Link` headers :-)

> But this also means any implementer adding things to a context must know the actual type in advance, while an interface would allow composition, wouldn't it?

Yes. As I wrote, this is a concious design desicion and goes with the answer to the previous question. There is only one `Webpage` context. HTML is a standard, it can only ever do what it can. In 99.999% of all use cases you will use the `Webpage` class. If there's a new feature we need to include, we can extend it and have to make sure, our controllers use it.
I just don't think the additional complexity (which will be huge) is worth it. We can still extract parts of it, if we see use cases popping up but we should not provide interfaces for everything just for the sake of it. I already see the documentation nightmare.

> I wonder if the asynchronous nature of the problem could benefit from lazy evaluation (closures?) or events.

Thanks to the ResponseContext you already know if a value is going to be used. No need for lazy evaluations.
Whether or not such a lazy evaluation is supported is up to the developer. See my JSON-LD PR. We could totally turn

```php
$this->jsonLd($array);
```

into 

```php
if ($this->supportsJsonLd()) {
    $this->jsonLd($array);
}
```

or

```php
$this->jsonLd(function() {
   return $array;
});
```

Whether or not this is worth the DX pain is up to the developer. It's not related to the overall concept here :)

ausi:
The concept LGTM 🎉

m-vo:
Ah, thanks for the explanation! :heart: 
Makes sense now! 

>> I wonder if the asynchronous nature of the problem could benefit from lazy evaluation (closures?) or events.
>
> Thanks to the ResponseContext you already know if a value is going to be used. No need for lazy evaluations.

I was thinking of cases where the *actual* usage in the template makes the difference. E.g. if you do not request an og:image in the template you would not even create one, like the image studio does. But I agree that might be too niche or rather unnecessarily add complexity. :slightly_smiling_face: :

fritzmg:
I also really like the concept 👍 🚀

Toflar:
We have 5/7 core devs that are happy with the concept. I'll thus work on polishing this PR and I am assigning it to 4.12.
Thanks for the quick feedback ❤️

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -721,11 +723,7 @@ protected function doReplace($strBuffer, $blnCache)
 
 				// Page
 				case 'page':
-					if (!$objPage->pageTitle && $elements[1] == 'pageTitle')
```
aschempp, line 724: shouldn't we continue to access the page model if there is no context? Also, how does this work BC if someone modifies the page model?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -299,6 +302,34 @@ class PageModel extends Model
...
+				switch ($strKey)
+				{
+					case 'pageTitle':
+						$responseContext->setTitle($varValue);
+						break;
+					case 'description':
+						$responseContext->setDescription($varValue);
+						break;
+					case 'robots':
+						$responseContext->setRobotsMetaTagContent($varValue);
```
aschempp: I feel like a `WebpageContext` could have a generic _Meta tag bag_, similar to a header bag where I can define any meta headers? Could be used to set `og:image` and everything else.

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -71,6 +82,14 @@ protected function prepare($objPage)
 		// Static URLs
 		$this->setStaticUrls();
 
+		// Set response context
+		/** @var WebpageContext $responseContext */
+		$responseContext = $container->get(ResponseContextFactory::class)->createAndSetCurrent(WebpageContext::class);
```
aschempp: isn't that a bit late and should be done in the controller?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * @var ResponseHeaderBag|null
+     */
+    private $headerBag;
+
+    public function getHeaderBag(): ResponseHeaderBag
+    {
+        if (null === $this->headerBag) {
+            $this->headerBag = new ResponseHeaderBag();
```
aschempp: we must make sure this `ResponseHeaderBag` is empty so it only contains what is really set. The default class is not empty!

news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
```diff
@@ -134,29 +136,33 @@ protected function compile()
 		$arrArticle = $this->parseArticle($objArticle);
 		$this->Template->articles = $arrArticle;
 
-		// Overwrite the page title (see #2853, #4955 and #87)
-		if ($objArticle->pageTitle)
-		{
-			$objPage->pageTitle = $objArticle->pageTitle;
```
aschempp, line 140: for BC reasons, I think we should do this in addition to the response context

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
Toflar: It does work because the `PageModel` sets it on the context. And there cannot be no context because all legacy page types get one.

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: Yes, not part of this PR though.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: Indeed but that's a problem. That's impossible. There's always at least `Date`. It's forced on the `ResponseHeaderBag`.

Toflar: If we take the default `HeaderBag` though, we don't have the cache control shortcuts.

news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
Toflar: I don't think that makes sense, no. It would keep generating deprecation notices.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: It's exactly where it's supposed to be. In the respective page type at the very beginning?

aschempp: The actual controller is here: https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/controllers/FrontendIndex.php#L88
this one is used by Symfony. However I agree `PageRegular` is _kind of_ a sub-controller for each page type. But I'm pretty sure there are use-cases to create your own instance of `PageRegular`, and that should never override the response context imho?

core-bundle/src/Routing/ResponseContext/ResponseContextAccessor.php
```diff
@@ -0,0 +1,53 @@
...
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Routing\ResponseContext;
+
+use Symfony\Component\HttpFoundation\RequestStack;
+
+class ResponseContextAccessor
```
aschempp, line 18: I find this class name rather strange. Also, shouldn't these methods just live the factory? So you only have one class to get/set/create a response context?

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return $this->robotsMetaTagContent;
+    }
+
+    public function setRobotsMetaTagContent(string $robotsMetaTagContent): self
+    {
+        $this->robotsMetaTagContent = $robotsMetaTagContent;
+
+        return $this;
+    }
```
aschempp: Not sure if that got lost: I think we should have a generic meta attribute parameter bag which would allow us to add arbitrary meta tags. This would also support @m-vo's use case.

aschempp (commented):
I would like to see a `ContaoWebpageContext` that extends `WebpageContext`. It would accept the current page model (which is already present when routing matched a page), initialize the known properties from it (pageTitle) and keep them in sync. This could be a massive BC benefit. But not sure how we can prevent endless loop if context => pagemodel => context 😂

core-bundle/src/Routing/ResponseContext/ResponseContextAccessor.php
Toflar: I thought about that too but I decided against it because as a user you are normally only accessing the current context. It would be super strange to inject a factory to access the context. That's why they are split and I like that better.

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
Toflar: It did not. I just said this was not part of this PR and resolved the conversation.
There's a lot more stuff that needs to go in the context like scripts etc. but again, not part of this PR.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Yeah where we set the context in this PR already, I just did it in PageRegular again. But I don't really care how we do that here TBH because it's going to be different with your PR anyway.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -299,6 +301,34 @@ class PageModel extends Model
 	 */
 	protected $blnDetailsLoaded = false;
 
+	public function __set($strKey, $varValue)
+	{
+		if (\in_array($strKey, array('pageTitle', 'description', 'robots'), true))
+		{
+			trigger_deprecation('contao/core-bundle', '4.12', sprintf('Overriding "%s" is deprecated and will not work in Contao 5.0 anymore. Use the ResponseContext instead.', $strKey));
```
fritzmg, line 308: I am not sure about this deprecation. What's deprecated is overriding these values for the pre-existing _global_ page object. But something like

```php
$newPage = new PageModel();
$newPage->pageTitle = 'foo';
```
or
```php
$somePage = PageModel::findById(…);
$somePage->description = 'New description';
$somePage->save();
```

should not trigger a deprecation, should it?

Toflar: True, but there's no way for me to know what you're going to do :D

fritzmg: May be add `$GLOBALS['objPage'] ?? null === $this` to the condition (not sure if that works, but you get the idea).

Toflar:
The failing tests are unrelated, this PR is ready to go to the final review round :)

Note regarding future plans on extending the `WebpageRequestContext`(or any subclass) for other meta tags (other than robots and description included in this PR), scripts, links etc.:
I think this does not belong into this PR for the sake of the size of it. I certainly want to work on it though.
However, I don't think that the current state of the PR prevents from doing that. E.g. I want to have some kind of head tag bag for `<meta>` and `<scripts>` etc. with priorities to order them and maybe searching CSS query like. But having the current `setDescription()` or `setRobotsMetaTagContent()` doesn't prevent us from doing that. They can just become shortcuts later on.
So

```php
    public function setDescription(string $description): self
    {
        $this->description = $description;

        return $this;
    }
```

would become something like this:

```php
    public function setDescription(string $description): self
    {
        $descriptionTag = $this->getHtmlHeadTagsBag()->find('meta[name="description"]');

        if (null === $descriptionTag) {
            $descriptionTag = new Meta('description'........);
            $this->getHtmlHeadTagsBag()->add($descriptionTag);
        }

        $descriptionTag->setContent($description); 

        return $this;
    }
```

fritzmg:
👍  I agree that introducing setting/getting a HTML head tag bag should be done in a follow-up PR.

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
aschempp: then we shouldn't add the meta robots tag neither, should we?

Toflar: Why not? It works the same as description.

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: This works indeed. Fixed in 83bbaccf31b47001288e80cfa1072f9e5cfe12d7. Will mark the discussion as resolved.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: I have fixed that by having our own `PartialResponseHeaderBag` in 5e69b26d2dc6c46aee4d2eb0800a84cb0c53f15e. I'll mark this discussion as resolved then, feel free to create a new one.

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
Toflar: See my latest comment in the PR about this. There's quite a few places in our code where we want to override the robots meta as well as the description so I think they deserve the shortcut methods we currently have. We can map them nicely onto a future whatever-head-tag-bag. Marking this conversation as resolved as well.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -299,6 +301,35 @@ class PageModel extends Model
 	 */
 	protected $blnDetailsLoaded = false;
 
+	public function __set($strKey, $varValue)
+	{
+		// Deprecate setting dynamic page attributes if they are set on the global $objPage
+		if (\in_array($strKey, array('pageTitle', 'description', 'robots'), true) && ($GLOBALS['objPage'] ?? null === $this))
```
m-vo:
```suggestion
		if (\in_array($strKey, array('pageTitle', 'description', 'robots'), true) && (($GLOBALS['objPage'] ?? null) === $this))
```
(because `===` binds tighter than `??`)

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -196,15 +215,15 @@ protected function prepare($objPage)
 
 		// Set the page title and description AFTER the modules have been generated
 		$this->Template->mainTitle = $objPage->rootPageTitle;
-		$this->Template->pageTitle = $objPage->pageTitle ?: $objPage->title;
-
-		// Meta robots tag
-		$this->Template->robots = $objPage->robots ?: 'index,follow';
+		$this->Template->pageTitle = $responseContext->getTitle(); // $objPage->pageTitle ?: $objPage->title;
```
m-vo: IMHO we can omit this comment :smiley:

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+namespace Contao\CoreBundle\Routing\ResponseContext\Factory\Provider;
+
+use Contao\CoreBundle\Routing\ResponseContext\ResponseContextInterface;
+
+interface ResponseContextProviderInterface
+{
+    public function supports(string $responseContextClassName): bool;
+
```
m-vo: Not sure what we decided about psalm template annotations, but this would be the perfect example for a `class-string` annotation. :smile_cat: 

```suggestion
    /**
     * @template T of ResponseContextInterface
     * @psalm-param class-string<T> $responseContextClassName
     * @return T
     */
```

m-vo (changes requested):

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: Hmm, we can omit the outer brackets then though?

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
Toflar: As we're using it in our tools, we might as well use them in docblocks if it makes things clearer. Added in 58f3728aa3e6ff1603dc7bfd0e80ec96df4993db

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Lol, leftover. Fixed in 58f3728aa3e6ff1603dc7bfd0e80ec96df4993db

core-bundle/src/Resources/contao/models/PageModel.php
Toflar: Fixed in 58f3728aa3e6ff1603dc7bfd0e80ec96df4993db.

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
```diff
@@ -0,0 +1,32 @@
...
+use Contao\CoreBundle\Routing\ResponseContext\ResponseContextInterface;
+
+interface ResponseContextProviderInterface
+{
+    /**
+     * @template T of ResponseContextInterface
+     * @psalm-param class-string<T> $responseContextClassName
+     * @return T
+     */
```
m-vo: This was probably added by accident for the `supports` function as well. But that one is returning a `bool` :slightly_smiling_face:

m-vo (changes requested):

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
Toflar: Ah, right. I'm new to Psalm :D Is https://github.com/contao/contao/pull/2975/commits/44e2445cf658a638a8184109aa92a94cdcb55668 correct?

m-vo: LGTM!

m-vo: Ah, hold on. Psalm actually can't infer things on an interface like this. :unamused: 

see https://psalm.dev/r/020d52cea5 vs https://psalm.dev/r/af4b7450d7

IMO we would need to implement the 2nd version in this case (using `@template T` on a method level):
```php
interface ResponseContextProviderInterface
{
    /**
     * @template T of ResponseContextInterface
     * @psalm-param class-string<T> $responseContextClassName
     */
    public function supports(string $responseContextClassName): bool;

    /**
     * @template T of ResponseContextInterface
     * @psalm-param class-string<T> $responseContextClassName
     * @return T
     */
    public function create(string $responseContextClassName): ResponseContextInterface;
}
```

m-vo (dismissed):

fritzmg (dismissed):

Toflar:
I've had a (imho great) idea overnight: The response context can now be terminated in which case modifications should not be possible anymore (immutability is hard though, especially when re-using other code like the `ResponseHeaderBag` - we'd need to override all setters etc. so I made this a `SHOULD` requirement until PHP supports something like this natively  - see `ResponseContextInterface`).

So basically the controller that sets the `ResponseContext` is expected to `->terminate($response)` it.
I've implemented dispatching a `TerminateResponseContext` event there but that's totally optional for any context to use (but it will allow me to use that general event to access the JSON-LD data instead of having a lot of specific events).
After a response context is terminated. Any listener/service coming after the controller (mainly `kernel.response` listeners) can still access the context but `$context->isTerminated()` should be `true` then. They can use the context to apply the logic to the output, do some logging etc. But they cannot modify it anymore.

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -71,6 +82,14 @@ protected function prepare($objPage)
 		// Static URLs
 		$this->setStaticUrls();
 
+		// Set response context
+		/** @var WebpageResponseContext $responseContext */
+		$responseContext = $container->get(ResponseContextFactory::class)->createAndSetCurrent(WebpageResponseContext::class);
```
aschempp: We cannot create a request context here. It would override anything already created by the controller. If my controller creates a context and then uses `PageRegular` to generate a response, it would be overwritten.

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
```diff
@@ -0,0 +1,32 @@
...
+     * @template T of ResponseContextInterface
+     * @psalm-param class-string<T> $responseContextClassName
+     */
+    public function supports(string $responseContextClassName): bool;
+
+    /**
+     * @template T of ResponseContextInterface
+     * @psalm-param class-string<T> $responseContextClassName
+     *
+     * @return T
```
aschempp: is that `@return` necessary?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+            $this->headerBag = new PartialResponseHeaderBag();
+        }
+
+        return $this->headerBag;
+    }
+
+    public function terminate(Response $response): void
+    {
+        if (null !== $this->serviceLocator && $this->serviceLocator->has('event_dispatcher')) {
+            $this->serviceLocator->get('event_dispatcher')->dispatch(new TerminateResponseContextEvent($this));
```
aschempp: The event should also contain the response, shouldn't it? And probably be applied after the default stuff (header bag).

core-bundle/src/Routing/ResponseContext/ResponseContextInterface.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Contao\CoreBundle\Routing\ResponseContext;
+
+use Symfony\Component\HttpFoundation\Response;
+
+interface ResponseContextInterface
+{
+    public const REQUEST_ATTRIBUTE_NAME = '_contao_response_context';
+
+    public function getHeaderBag(): PartialResponseHeaderBag;
```
aschempp: Do we need to be specific to `PartialResponseHeaderBag`?

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function setTitle(string $title): self
+    {
+        $this->title = $title;
+
+        return $this;
+    }
+
+    public function getDescription(): string
```
aschempp:
```suggestion
    public function getMetaDescription(): string
```

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return $this->description;
+    }
+
+    public function setDescription(string $description): self
```
aschempp:
```suggestion
    public function setMetaDescription(string $description): self
```

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->description = $description;
+
+        return $this;
+    }
+
+    public function getRobotsMetaTagContent(): string
```
aschempp:
```suggestion
    public function getMetaRobots(): string
```

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return $this->robotsMetaTagContent;
+    }
+
+    public function setRobotsMetaTagContent(string $robotsMetaTagContent): self
```
aschempp:
```suggestion
    public function setMetaRobots(string $robotsMetaTagContent): self
```

aschempp (commented):
I **really** like the idea of terminating the response context. I **really dislike** the term _terminate_ for it 😂 can we `apply` a response context or something?

core-bundle/src/Routing/ResponseContext/Factory/Provider/ResponseContextProviderInterface.php
Toflar: Yes :) It lets Psalm know that what's returned is an instance of `class-string` you passed in as an argument :)

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Where then?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: I want to have the event before so I can add headers. If you do not want to add any, you can remove them all. So this is definitely the better place. I guess passing on the response makes sense though, yes. Implemented in 8a1691325b40008ae37334b44b5bf13f691e6e2c.

core-bundle/src/Routing/ResponseContext/ResponseContextInterface.php
Toflar: I thought we should. That's why I did :D

Toflar:
`apply`ing can happen multiple times. `terminate` happens once. I thought about `finalize()` but in the HttpKernel it's called `terminate()` too so I thought I just go with what people might already know.
I'm open for alternative suggestions but "apply" is really not correct here. I want to close it. But `close()` does not tell that there's still something happening. So I like `terminate` or `finalize`.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Ah, `FrontendIndex::renderPage()` is the controller for legacy page types. I understand now. So I guess 8901784467137ed6c6fad1a1a067b3535f08afdb should be it then?

fritzmg:
(I think I am also in favor of `finalize` 😄)

m-vo:
According to [this site](https://wikidiff.com/terminated/finalized) terminate means "To end, especially in an incomplete state" while finalize means "To make final or firm; to finish or complete". So, although I agree that reusing the term of the HttpKernel is a good idea, maybe finalize is the better fitting word. 

What about freeze? :grin:

Toflar:
"freeze" has the same issue as "close". It doesn't tell that it's doing something and only then is finishing it.

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -45,7 +47,15 @@ public function getResponse($objPage, $blnCheckRequest=false)
 	{
 		$this->prepare($objPage);
 
-		return $this->Template->getResponse($blnCheckRequest);
+		$response = $this->Template->getResponse($blnCheckRequest);
+		$responseContext = System::getContainer()->get(ResponseContextAccessor::class)->getResponseContext();
+
+		if ($responseContext instanceof WebpageResponseContext)
```
aschempp: why are we checking the instanceof here?

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -213,7 +230,7 @@ protected function prepare($objPage)
 
 		// Assign the title and description
 		$this->Template->title = strip_tags($this->replaceInsertTags($objLayout->titleTag));
-		$this->Template->description = str_replace(array("\n", "\r", '"'), array(' ', '', ''), $objPage->description);
+		$this->Template->description = str_replace(array("\n", "\r", '"'), array(' ', '', ''), $responseContext->getMetaDescription());
```
aschempp, line 224: didn't we already sanitize the description when it was added to the response context?

aschempp (commented):
How about `finishResponse` instead of `terminate`?

I also wonder whether we actually need a generic factory service with providers, since every controller **knows** what context instance it wants. Maybe we can discuss this tomorrow at the public call and finalise the PR 😊

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Because we are accessing methods of it afterwards. Any service that gets the current response context must test if it's the correct one and either do nothing (in case of our future JSON-LD implementation) or throw an exception if it was unexpected. Nobody should be doing `new PageRegular()` without setting the context previously. If I did not check here, they would get a PHP error saying that you cannot call `getMetaDescription()` on `null` ;)

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: I think we should keep it as is because it's the way it was before. I guess if e.g. the news sets the description with its default process now it might contain newlines. I don't want to change that here now for BC reasons. But I agree that we should expect the context to be clean in future proper page controllers.

Toflar:
> How about `finishResponse` instead of `terminate`?

Again, `finishResponse` does not describe what it does correctly. The response is not finished. The context is finalized (not finished). So either `finalize()` or `terminate()`. Those are the only two viable options to me.

> I also wonder whether we actually need a generic factory service with providers, since every controller **knows** what context instance it wants. Maybe we can discuss this tomorrow at the public call and finalise the PR 😊

Yes, we definitely do need this because it's the correct architectural decision. Every controller knows what context it wants. However, it does **NOT** know, what the context itself requires. Hence, the factory.
`PageRegular` does not need to know what `WebpageResponseContext` needs. It only needs to know that it wants a `WebpageResponseContext`. Maybe we want to add new arguments and dependencies to `WebpageResponseContext` in the future because of a new HTML standard - the page controller should not need to care.

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: ah so this is not to check the interface, but to check against NULL values (not having a request context)?

aschempp:
> Yes, we definitely do need this because it's the correct architectural decision. Every controller knows what context it wants. However, it does NOT know, what the context itself requires. Hence, the factory.
PageRegular does not need to know what WebpageResponseContext needs. It only needs to know that it wants a WebpageResponseContext. Maybe we want to add new arguments and dependencies to WebpageResponseContext in the future because of a new HTML standard - the page controller should not need to care.

I absolutely understand and agree to that. But `PageRegular` **knowns** what factory provides this request context, so it could directly use that factory and not a generic "provider" implementation. The major advantage would be that the `WebpageContextFactory` could have a method that accepts the `PageModel` and sets the known properties. Otherwise your code to set the title needs to be duplicated on every front end controller.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: It makes sure that we have a `WebpageResponseContext`, yes. In a new page controller, creating the context would be something of the first things you would do in your invoication method and then you know it must be there. Here, however, we have a "legacy" page type. Its controller resides in `FrontendIndex::renderPage()` which means that in theory someone could've called `new PageRegular()` instead of properly using the Symfony controller. I want to make sure that people notice this error. Calling `new PageRegular()` manually has always been wrong, imho. You should've called the controller for page `xy` :)

Toflar:
Reworked the concept again. We don't need any finalizing stuff anymore and the event was nonsense because

- the event will be triggered AFTER the response is generated. So it is essentially the same as `kernel.response`. I will need my own event anyway when generating the scritps for JSON-LD.
- The finalizing/terminating concept is gone because instead of closing the context, it should actually be removed. So the controller sets it when it starts working and applies and unsets it, when generating the response.
- Using concrete factories now which simplifies things a lot.


Now that should be it :D

core-bundle/src/Routing/ResponseContext/ResponseContextAccessor.php
```diff
@@ -0,0 +1,71 @@
...
+        $request = $this->requestStack->getCurrentRequest();
+
+        if (null !== $request) {
+            $request->attributes->set(ResponseContextInterface::REQUEST_ATTRIBUTE_NAME, $responseContext);
+        }
+
+        return $this;
+    }
+
+    public function finalize(Response $response): self
```
aschempp: Shouldn't this be part of the response context itself? Only the context knows what can/should be done on the response, maybe my custom context needs to do something else than (just) applying headers?

core-bundle/src/Routing/ResponseContext/ContaoWebpageResponseContext.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Contao\PageModel;
+
+class ContaoWebpageResponseContext extends WebpageResponseContext
+{
+    public function __construct(PageModel $pageModel)
+    {
+        $this
+            ->setTitle($pageModel->pageTitle ?: $pageModel->title ?: '')
+            ->setMetaDescription(str_replace(["\n", "\r", '"'], [' ', '', ''], $pageModel->description ?: ''))
```
aschempp, line 23: what about meta robots here?

Toflar: Fixed in d829f75ae177c68c310bf3f52f495a853b39e26b.

core-bundle/src/Routing/ResponseContext/ResponseContextAccessor.php
Toflar: Adjusted in eae2ddc09c8ed126150ed37ca19cf8f96931aacb.

Toflar:
All done here. Remaining failing CS issue is not related to this PR.

aschempp (dismissed):

Toflar:
All green now - I'm done here :D

fritzmg (dismissed):
lgtm 👍 Let's not forget to also make the same changes in the calendar- and faq-bundle, as well as [`ModuleArticle`](https://github.com/contao/contao/blob/204fb5a978e1cd49bc2819bdc34df421a1cd1296/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L164-L173) in a follow up PR though.

aschempp (dismissed):

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * @param string $key
+     */
+    public function remove($key): void
+    {
+        parent::remove($key);
+
+        $uniqueKey = strtr($key, self::UPPER, self::LOWER);
```
leofeyer, line 39: This will trigger a warning if `$key` is not a string in PHP 8.

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
```diff
@@ -0,0 +1,38 @@
...
+        $responseAccessor = $this->createMock(ResponseContextAccessor::class);
+        $responseAccessor
+            ->expects($this->exactly(3))
+            ->method('setResponseContext')
+        ;
+
+        $factory = new CoreResponseContextFactory($responseAccessor);
+
+        $this->assertInstanceOf(ResponseContext::class, $factory->createResponseContext());
+        $this->assertInstanceOf(WebpageResponseContext::class, $factory->createWebpageResponseContext());
```
leofeyer: These tests are meaningless, because the return type is already guaranteed by the method signature.

leofeyer (changes requested):

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
Toflar: It's copied from Symfony Core. `$key` must be a string according to the docs ;)

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
Toflar: Only if someone calls them. If we don't, there's no coverage. And there's nothing else to test against here - I don't want to test the specific implementations. I want to test whether the factories do what they are supposed to do.

leofeyer:
> lgtm 👍 Let's not forget to also make the same changes in the calendar- and faq-bundle, as well as [`ModuleArticle`](https://github.com/contao/contao/blob/204fb5a978e1cd49bc2819bdc34df421a1cd1296/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L164-L173) in a follow up PR though.

Why don't we make the changes here so we can get this issue done?

Toflar:
I only did it in news-bundle to illustrate the new way. The old way has to work as well (and does) which is why we needed both cases. Getting rid of the deprecation messages would pollute this PR (apart from the fact that there's **a lot** more than just page title, description and meta tags to be added on top of that concept here in the future).

leofeyer:
The changes in the news module look pretty straight forward to me. If we do not change the other classes now, I am afraid we never will. Therefore I would prefer to have this done here. 🙈

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
leofeyer: It is still not enforced, therefore we either have to handle the `!is_string($key)` case or cast the variable explicitly.

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
m-vo: IMHO the main problem isn't that this is testing against a known return type but that PHPUnit does not allow testing valid behavior of things that it cannot observe.

(see https://github.com/sebastianbergmann/phpunit-documentation/issues/171).

leofeyer:
I have added the changes in 753a048320b9ca887441d9520b7f0bd0cb7e50f3. Please review the commit.

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
leofeyer: Isn't that what the `@covers` annotation is for?

Toflar:
They look good to me :)

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
Toflar: It's not enforced in Symfony either. The phpdoc clearly state `@param string`. If you pass something else then it doesn't work - yes :D In newer versions Symfony has a type hint anway but I couldn't use that because of being compatible with lower versions. But feel free to add a `(string)` type cast if you prefer that.

m-vo (dismissed):

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
leofeyer: @ausi WDYT? Should we be defensive here or is it OK to handle it like Symfony does?

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
ausi: You could call the factory methods without the instanceof check in the test method and add the test with `@doesNotPerformAssertions` I think.

core-bundle/src/Routing/ResponseContext/PartialResponseHeaderBag.php
ausi: I would keep it as it is because with Symfony 5 the `parent::remove($key)` would already trigger the type error and we would be best compatible with Symfony this way IMO.

core-bundle/tests/Routing/ResponseContext/CoreResponseFactoryTest.php
leofeyer: Changed in bd9a4a30e6e616ed086f59a501735d6499b4902a.

leofeyer (approved):

fritzmg (approved):

aschempp (approved):

leofeyer:
Thank you @Toflar.

ausi (approved):

PR #2615, opened 2020-12-31
Fix the news link markup

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/2396#issuecomment-703212985
| Docs PR or issue | -

This fixes the issue mentioned [here](https://github.com/contao/contao/issues/2396#issuecomment-703212985).

##### Before

```html
<div class="layout_latest arc_1 block odd" itemscope="" itemtype="http://schema.org/Article">
  <h2 itemprop="name">
    <a href="en/news/lorem-ipsum-foobar" title="Read the article: Lorem ipsum foobar" itemprop="url">
      <span itemprop="headline">Lorem ipsum foobar</span>
    </a>
  </h2>
  <p class="more">
    <a href="en/news/lorem-ipsum-foobar" title="Read the article: Lorem ipsum foobar" itemprop="url">
      <span itemprop="headline">Read more …</span>
      <span class="invisible"> Lorem ipsum foobar</span>
    </a>
  </p>
</div>
```

##### After

```html
<div class="layout_latest arc_1 block odd" itemscope="" itemtype="http://schema.org/Article">
  <h2 itemprop="name">
    <a href="en/news/lorem-ipsum-foobar" title="Read the article: Lorem ipsum foobar" itemprop="url">
      <span itemprop="headline">Lorem ipsum foobar</span>
    </a>
  </h2>
  <p class="more">
    <a href="en/news/lorem-ipsum-foobar" title="Read the article: Lorem ipsum foobar" itemprop="url">
      Read more …
      <span class="invisible"> Lorem ipsum foobar</span>
    </a>
  </p>
</div>
```

##### Before

```html
<div class="layout_latest arc_1 block first even" itemscope="" itemtype="http://schema.org/Article">
  <h2 itemprop="name">
    <a href="https://twitter.com/" title="Read more on https://twitter.com/" itemprop="url">
      <span itemprop="headline">External</span>
    </a>
  </h2>
  <p class="more">
    <a href="https://twitter.com/" title="Read more on https://twitter.com/" itemprop="url">
      <span itemprop="headline">Read more …</span>
    </a>
  </p>
</div>
```

##### After

```html
<div class="layout_latest arc_1 block first even" itemscope="" itemtype="http://schema.org/Article">
  <h2 itemprop="name">
    <a href="https://twitter.com/" title="Read more on https://twitter.com/" itemprop="url">
      <span itemprop="headline">External</span>
    </a>
  </h2>
[...]

PR #2977, opened 2021-04-27
Fix external news targets not opening in a new window

### LINKED ISSUE #2947: respect news redirect target
**Affected version(s)**

Tested in demo (4.9.12)

**Description**

News-Item:
Redirect target->Custom URL->Open in new window
The link does **not** open in new window (neither layout_simple nor layout_short).
It works for events though.

image

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2947
| Docs PR or issue | -

The implementation for opening external target URLs for news articles in a new window was erroneously removed in https://github.com/contao/contao/pull/2615. This PR fixes that.

### DIFF
--- a/news-bundle/src/Resources/contao/modules/ModuleNews.php
+++ b/news-bundle/src/Resources/contao/modules/ModuleNews.php
@@ -369,9 +369,10 @@ abstract class ModuleNews extends Module
 		$strArticleUrl = News::generateNewsUrl($objArticle, $blnAddArchive);
 
 		return sprintf(
-			'<a href="%s" title="%s" itemprop="url">%s%s</a>',
+			'<a href="%s" title="%s"%s itemprop="url">%s%s</a>',
 			$strArticleUrl,
 			StringUtil::specialchars(sprintf($strReadMore, $blnIsInternal ? $objArticle->headline : $strArticleUrl), true),
+			($objArticle->target && !$blnIsInternal ? ' target="_blank" rel="noreferrer noopener"' : ''),
 			($blnIsReadMore ? $strLink : '<span itemprop="headline">' . $strLink . '</span>'),
 			($blnIsReadMore && $blnIsInternal ? '<span class="invisible"> ' . $objArticle->headline . '</span>' : '')
 		);


news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -369,9 +369,10 @@ protected function generateLink($strLink, $objArticle, $blnAddArchive=false, $bl
 		$strArticleUrl = News::generateNewsUrl($objArticle, $blnAddArchive);
 
 		return sprintf(
-			'<a href="%s" title="%s" itemprop="url">%s%s</a>',
+			'<a href="%s" title="%s"%s itemprop="url">%s%s</a>',
 			$strArticleUrl,
 			StringUtil::specialchars(sprintf($strReadMore, $blnIsInternal ? $objArticle->headline : $strArticleUrl), true),
+			($objArticle->target && !$blnIsInternal ? ' target="_blank" rel="noreferrer noopener"' : ''),
```
m-vo, line 375: Shouldn't `rel="noreferrer noopener"` be added to each external link (independent of the target)?

fritzmg: That's how it was previously. I assume the thought behind it that there is no real reason to add `noreferrer noopener` for links that point to your own site, as it is not an untrusted link.

m-vo: You could also add an external link and not set it to open in a new window, could you? This is probably what you want for UX anyway.

I know it was like this before but IMHO we should change it. :slightly_smiling_face:

leofeyer (approved):

news-bundle/src/Resources/contao/modules/ModuleNews.php
leofeyer: @contao/developers WDYT?

ausi: AFAIK `noopener` only makes sense for `target=_blank` links as links in the same page don’t get an opener reference anyway. I would keep it as it is now.

leofeyer:
Thank you @fritzmg.

PR #2980, opened 2021-04-27
[POC] Extendable image studio

m-vo:
This is a rough outline how the image studio concept could be extended to allow arbitrary media instead of "just" images that are processed by our image factories as well as an event based system (inspired by @Toflar's #2962) to alter the output. 

I've got a use case in my mind where I wanted to display [lottie files](https://lottiefiles.com/web-player) and want them to act like regular images.  These are `json` files that must not be put through the image processing pipelines. I try to find out if things like this could be made possible. If so, you would be able to reuse the infrastructure, meta data and so on. :slightly_smiling_face: 

That said, at this point it's a rather quick draft to get the discussion going.

### DIFF
--- a/core-bundle/src/Image/Studio/Event/AbstractFigureBuilderEvent.php
+++ b/core-bundle/src/Image/Studio/Event/AbstractFigureBuilderEvent.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Image\Studio\Event;
+
+use Contao\CoreBundle\Image\Studio\Figure;
+use Contao\CoreBundle\Image\Studio\FigureBuilder;
+
+abstract class AbstractFigureBuilderEvent
+{
+    /**
+     * @var FigureBuilder
+     */
+    private $figureBuilder;
+
+    /**
+     * @var Figure
+     */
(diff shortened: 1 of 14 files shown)


Toflar:
I like the idea of generally abstracting media (just like abstracting the filesystem would be a great idea). However, I don't think extending the Studio like that makes a lot of sense. It's in the `Image` namespace and really works for images only at the moment which is also okay to me. It's an image studio :-)
Maybe your general concept could sit on top of the image studio rather than making the studio the top level?

m-vo:
My thinking was: Why should it matter if an image is an animated `svg` or a `json` file (that in effect looks the same)? If you'd have existing code like this (e.g. from a core content element)…

```php
// existing code (unchanged)
$figure = $figureBuilder->from($resource)->setSize($size)->build();
$this->render('template.html.twig', ['figure' => $figure]);
```
… you would be able to introduce new capabilities in an extension or your app without the need to change it. For the lottie file example you could implement a `DefineMediaResultEvent` that creates an instance of `LottieResult implement MediaResultInterface` in case there's a matching source file set. And then you'd get everything you need inside the `Figure` container in the templates. 

I kind of think the same about the namespace. :-/ Maybe the figure concept should have been media agnostic from the beginning (images, videos, maps?, whatnot…). Not sure if it's worth abstracting, though.

m-vo:
I'm closing this in favor of #3019. This followup PR only implements an event for metadata which is probably the most important part still.

Extending the concept isn't forgotten :smile: but may need another approach then just a pure event based solution (think of multiple `FigureBuilders` or the like).

PR #2983, opened 2021-04-29
Unset noComments field if comments bundle is not installed

### LINKED ISSUE #1472: Events Feld "Kommentare deaktivieren" ohne comments-bundle
In den Einstellungen von Events wird das Feld ``Kommentare deaktivieren`` auch angezeigt, wenn das ``comments-bundle`` nicht installiert ist.

gmpf:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1472

Also add a `clr` class to the field that follows the removed `noComments` field (currently: `featured`) so that we don't end up with a checkbox field floating awkwardly next to the previous field (currently: `cssClass`).


leofeyer:
This is the wrong approach IMHO. Instead of defining and removing the field in the news/calendar/FAQ bundle, it should be defined in the comments bundle. Then, if the comments bundle is not installed, the field will not be present at all.

gmpf:
Ok, I was just going by what was already being done for the `allowComments` field:
https://github.com/contao/contao/blob/814d0a07e5e31c5e49f5685b2eb92ed5f9555f60/news-bundle/src/Resources/contao/dca/tl_news_archive.php#L260-L266

However, I didn't notice that the actual removal of the field was implemented such a long time ago; I can see why you'd want to change the whole concept.

> Instead of defining and removing the field in the news/calendar/FAQ bundle, it should be defined in the comments bundle. Then, if the comments bundle is not installed, the field will not be present at all.

So the task is to provide those fields generically so that the comments bundle doesn't have to know which other bundles use it, right?

leofeyer:
Exactly.

aschempp:
I'm not sure @leofeyer is right here. News bundle integrates with comments, not the other way around. `ModuleNewsReader` handles whether the comments bundle is installed, so it should handle this too.

leofeyer:
> `ModuleNewsReader` handles whether the comments bundle is installed, so it should handle this too.

I figured we had agreed that this is wrong as well?

aschempp:
Well it depends if 
 1. comments integrates with news
 2. or news integrates with comments

It would probably be much simpler to do the second than the first

gmpf:
I agree that the bundles that use comments have a soft requirement for the comments bundle, not the other way around.

The difference between the checks in the modules vs the DCAs is that in the modules, there is a check whether comments is available, then something is done in that case; whereas in the DCAs, the field is added first, then removed in the callback if comments is _not_ available.

I thought @leofeyer's initial reply meant that the comments bundle should have something that provides the field configuration for noComments and allowComments, as well as a means of using that provision without much code in the bundles that use it, so that we don't have as much repetition of field definition and removal in the DCAs. Of course, that provision would still have to be used in the other bundles, it would just reduce the amount of code in events/faq/news DCAs.

leofeyer:
As discussed in Mumble on May 12th, the current integration is correct and the PR should be merged.

Toflar:
Can we change the implementation in such a way that the fields are not removed from the DCA if the comments bundle does not exist but instead only added if the bundle exists? :) That would be a bit cleaner I guess and you can use the `PaletteManipulator` for that :)

leofeyer:
BTW, a much simpler fix would be to exchange the "Disable comments" and the "Feature event" fields and have the `clr` class on the latter.

leofeyer:
Closed in favor of #3508

PR #2985, opened 2021-04-30
Commit the composer.lock files of the tools

leofeyer:
As discussed with @Toflar, this should prevent the CI chain from suddenly failing if one of the tools releases a new version.


aschempp (dismissed):
do we need a way to make sure these are kept up to date? Like a weekly/montly CI ?

Toflar:
I agree with @aschempp. It's great to lock the tools to a certain version so we don't have failing PRs all the time. However, we should ensure to run updates on them from time to time because the updates also come with better code quality for us :)

m-vo:
> Like a weekly/montly CI ?

This would basically need a periodic job that mirrors the current CI but just runs `update` instead of `install`, right?

```diff
-   composer bin ecs install --no-interaction --no-progress
+   composer bin ecs update --no-interaction --no-progress

[…]
```

IMHO this could replace the current schedule to save resources.

leofeyer:
I have added a cronjob that runs the CS job with updated dependencies. I have also removed the composer.lock file of the monorepo-tools, because we control the package ourselves.

bytehead (approved):

Toflar (approved):

m-vo (approved):

PR #2988, opened 2021-05-04
Add Twig support

### LINKED ISSUE #2303: Allow using Twig templates to implement/replace Contao templates

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2303 and potentially #2638
| Docs PR or issue | see https://github.com/contao/docs/pull/799

supersedes #2612 

This PR aims at implementing first steps of native Twig support for Contao.

# What's inside
### Encoding
You now can write Twig templates like so…
```twig
<h1>{{ headline }}</h1>
<p class="rte">{{ content|raw }}</p>
```

… and provide input encoded data for the variables. They won't be encoded twice thanks to setting `double_encode: false` (idea &copy; @ausi). This means you do not have to use `|raw` for normal values or, put differently, only where you indeed intend to output the raw data - for example when dealing with rte data. This way templates can be written like they would with output encoding and we get a nice upgrade path.

This logic is implemented in a `NodeVisitor` that rewrites the escaping strategy to our own one called `contao_html`. It is only applied to template names that follow a certain rule set (`'%^@Contao(_[a-zA-Z0-9_-]*)?/%'`) but you can add your own rules as well.

### Loader
Twig is configured with a `ChainLoader` that supports multiple loaders by default. We're using this fact and are now providing two additional loaders with a higher priority (default has priority 0):

1. `FailTolerantFilesystemLoader` (priority: 1) - this is just a simple wrapper around the original implementation that doesn't care about adding missing paths. Paths that Symfony adds at compile (i.e. all the Twig bundle paths) do get rewired to this loader in a compiler pass. We kind of used this mechanism already (only with decoration instead of a new loader).

2. `ContaoFilesystemLoader` (priority: 2) - this is where all the magic happens. This loader will provide all Contao templates from bundles, the app, global templates, theme folders… It is also designed to be altered at runtime and uses a cache pool to store its state. Here are some features that differ from Twig's default filesystem loader:
    - If you request a template but the current page context is 'a theme' and there is a theme variant of the template, you'll get this instead. In fact that's the case for every operation (`getCacheKey`, `getSourceContext`, `exists`, `isFresh`).
    - When adding template paths (namespace + directory) there is a switch to enable 'tracking templates'. If enabled, the content of the path will be scanned and found templates will be put in a 'hierarchy' that represents a possible inheritance chain (see below). Tracking templates also allows invalidation (see `isFresh`) for associated variants. 

Our filesystem loader gets primed by a mandatory cache warmer (`ContaoFilesystemLoaderWarmer`) that'll add all the paths and namespaces. Speaking of namespaces, these are the ones getting registered:

1. `Contao_Theme_<themeSlug>` for each theme folder
2. `Contao_Global` for the global templates folder
3. `Contao_<bundleName>` for each bundle path (could be multiple per bundle); this includes `Contao_App` for the app resources (i.e. in a `AppBundle` or `/contao/templates`).
4. additionally `Contao` for all resources

Also see the `TemplateLocator` helper service for reference. Btw.: The existing `AddResourcesPass` also searches these paths but behaves differently. Maybe we could merge those two things in the future. But I'm not planning to do this in this PR.

### Inheritance
Twig allows inheriting multiple times (`A extends B extends C extends …`) which is a very helpful concept when building complex components. However this doesn't play nicely with an extension system like ours where the code does not know how it will be orchestrated.  Imagine you've got an original resource and you are adjusting it in the app:

```twig
{# the original thing.html.twig #}
{% block content %}{{ foo }}{% endblock %}
<footer>{% block footer %}Footer{% endblock %}</footer>
```

```twig
{# App #}
{% extends "@Contao/thing.html.twig" %}
{% block content %}<section>{{ parent() }}</section>{% endblock %}
{% block footer %}My own footer{% footer %}
```

Now you install an extension (think of an extension to a bigger extension) that brings its own modifications:

```twig
{# BarExtension #}
{% extends "@Contao/thing.html.twig" %}
{% block content %}{{ parent() }}? Bar!{% endblock %}
```

You would probably expect this output:
```twig
<section>{{ foo }}? Bar!</section>
<footer>My own footer</footer>
```

And in fact, this is what you would get with this PR. :grin:

You may have noticed the namespace every template is extending from is `@Contao`. In regular Twig land there can only ever win a single template (with the same name) in each namespace, in fact this is why Symfony registers each bundle namespace twice (`@<bundle>` and `@!<bundle>`), so that you can overwrite a template and still extend from it (using the `!` one). 

We're working around this limitation using two things: we know about templates not only template paths (see tracking in previous section) and we're providing our own version of the `ExtendsTokenParser` that - whenever a `@Contao` template is targeted - will rewrite it to a specific namespace according to the inheritance hierarchy. 

So `@Contao/foo.html.twig` would for instance become `@Contao_App/foo.html.twig` in the theme template, `@Contao_FooBundle/foo.html.twig` in the app template and `@Contao_CoreBundle/foo.html.twig` in the FooBundle template. 

### PHP Template Interop

Twig and our current templates are similar in a way that at the end PHP code is evaluated and both support template inheritance. But that's kind of it. :smile: To be able to make them extend from each other seamlessly we IMHO have to transpile PHP to Twig and process them like regular Twig templates. But that's something for the future. 

What **is** possible with this PR is extending a `.html5` template in Twig (like `{% extends "@Contao\nav_default.html5" %}`) but bear in mind that this won't allow certain dynamic things like introducing new blocks and extending again.

This part is probably the thing that involves most "magic". Here is the gist how it works: We're also registering `.html5` templates as if they were Twig templates in our loader. The loader, however, strips the PHP code when returning a source context to not confuse Twig's lexer/parser chain but simply keeps the identified block names in there. Later on - in a node visitor - we're then dressing the `ModuleNode` of this template to include a proxy call and inject `BlockNode` s that return `[[TL_PARENT]]` and that are rendered when running the proxy code. If a block was overwritten by a child template it will win, otherwise the placeholder will be output. Contao's template engine then gets run with the blocks prepopulated and will output the result.


core-bundle/src/Twig/Interop/ContaoEscaperNodeVisitor.php
```diff
@@ -0,0 +1,112 @@
...
+            foreach ($this->escaperFilterNodes as $escaperFilterNode) {
+                $this->setContaoEscaperArguments($escaperFilterNode);
+            }
+
+            $this->escaperFilterNodes = null;
+        }
+
+        return $node;
+    }
+
```
m-vo, line 92: The current implementation opens up a collection when an affected module is encountered, collects references to filter nodes and modifies them at once when leaving the module.

We could also toggle a flag when entering an affected module and directly modify each filter node when leaving it as long as the flag is set. Maybe this is cleaner because no references are needed. But I'm not sure.

ausi:
I like this! 😍

The escaping looks great and I think this can be the perfect approach to slowly move to Twig and to output encoding. 💪

Not sure how good I can review the Twig integration part. Should we set this to “up for discussion”?

m-vo:
> Not sure how good I can review the Twig integration part. Should we set this to “up for discussion”?

Sounds good to me. I'm currently working on the inheritance chain that most likely will need some explanation + decision making anyway.

m-vo:
> Currently only Twig templates are supported, but I'm working on arbitrary extends so that you can mix .html5 and .html.twig templates in the chain. Not sure this can be fully supported - it's the bit with most magic involved.

Small update: I built a PHP2Twig transpiler. :see_no_evil:  Not yet ready, but this might actually work.

m-vo:
I reworked the concept as discussed in our Mumble call on 12th of May. 

I'm also really happy how much cleaner this iteration is than the previous one. :tada: The `ContaoFilesystemLoader` now encapsulates most of the logic and is completely uncoupled from the container.

Here's the gist:

1) I am now storing the loader's state (paths + template hierarchy) in a cache pool and use a cache warmer to prime it. This allows altering and refreshing the state at any time. We probably want to implement a 'always warm' or 'watch files' strategy when in dev mode and also invalidate the cache when editing things in the backend. Not there yet, but now we can. :slightly_smiling_face: 
2) This also means the compiler pass isn't needed anymore. I'm still migrating the "addPath" calls from Symfony's loader, because ours is fail tolerant against missing paths. We could IMHO discuss if we want to drop the support to alter arbitrary bundle templates in the backend once we've got the `@Contao` namespace solution. Then we could leave Symfony's bundle logic untouched as well.
3) Speaking of namespaces: I'm now searching in all of Contao's resource paths and resolve sub directories like discussed. (Unfortunately I had to replicate the logic inside `AddResourcesPass` because `%contao.resources_paths%` does not provide enough information about bundles/namespaces… but it's only a few lines.) Theme templates are now be resolved without the need to specify them: the loader checks the page context and automatically provides the right cache keys/source contexts. This means you render `@Contao/ce_text.html.twig` but you may get `@Contao_Theme_FooTheme/ce_text.html.twig` like you would with Contao templates.
4) I merged the template hierarchy logic into the loader as well. You can now simply enable "tracking templates" when adding paths which will put templates of this path into the hierarchy. We're evaluating the inheritance chain at template compile time. Instead of extending from the bundle template, every `text.html5.twig` template extends from `@Contao/text.html5.twig` - the individual extends will then be resolved to: `@Contao_Theme_Baz/text.html5.twig` extends `@Contao_Global/text` (/template/) extends `@Contao_App/text.html5.twig` (/contao/templates/) extends `@Contao_LateBundle/text.html5.twig`  (vendor/…) extends `@Contao_OriginalBundle/text.html5.twig`. You get the point. :smile:  This drastically improves the possibilities to adjust small things and promotes the use of blocks.
5) I replaced the registry of templates that the `contao_html` escaper mechanism will act on with a rule based solution. Now, by default, every template in the `@Contao` and `@Contao_*` namespaces will be processed. Own filters (regular expressions on the filename) can be added at runtime.

A bit more to come, mostly adding/fixing tests and first steps of integration with the `TemplateLoader` / backend. Feedback is welcome.

m-vo:
6. We're now cache busting all templates in the Contao namespace by appending a hash of the template hierarchy. So, if the hierarchy is updated/refreshed the environment will compile new templates as well.

7. I added a simple `debug:contao-twig` command to visualize the template hierarchy. Adding `--refresh` or `-r` rebuilds it.

m-vo:
8. Now, everything should be automatically refreshed in dev mode.

9. Contao Templates are now also located in the template hierarchy. When loading, their (PHP) content will be stripped and later on (when parsing) be replaced with a proxy to the old template system. This has its limitations (like: you cannot introduce new blocks when inheriting) but at least allows extending them.

10. I improved the cache busting multiple times until I ultimately dropped it and implemented our own version of `Loader#isFresh` that simply checks all templates in a hierarchy…

Allright, I pretty much went through all the rabbit holes there are. Kind of exhausted but I hope it's pretty close. I'll update the initial description to reflect the state. Ping me, if you're interested in looking through this PR and have questions (I'm sure there are).

m-vo:
>  I'll update the initial description to reflect the state.

Done.

composer.json
```diff
@@ -87,7 +87,7 @@
         "symfony/console": "4.4.* || 5.2.*",
         "symfony/debug": "4.4.* || 5.2.*",
         "symfony/debug-bundle": "4.4.* || 5.2.*",
-        "symfony/dependency-injection": "4.4.* || 5.2.*",
+        "symfony/dependency-injection": "^4.4.20 || ^5.2.4",
```
```suggestion
        "symfony/dependency-injection": "~4.4.20 || ~5.2.4",
```

As we explicitly want to disallow `5.3.X`

core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
```diff
@@ -0,0 +1,59 @@
...
+            ->write('echo $this->extensions[')
+            ->repr($this->getAttribute('extension_name'))
+            ->raw(']->renderLegacyTemplate('."\n")
+            ->indent()
+            ->write('$this->getTemplateName(),'."\n")
+            ->write('array_map('."\n")
+            ->indent()
+            ->write('static function(callable $block) use ($context): string {'."\n")
+            ->indent()
+            ->write('ob_start(); $block($context); return ob_get_clean();'."\n")
```
ausi: Can/should we wrap the closing of the output buffer in a `finally {}`?

```suggestion
            ->write('ob_start(); try { $block($context); return ob_get_contents(); } finally { ob_end_clean(); }'."\n")
```

composer.json
m-vo: Thanks, yes! Adjusted in 832b16e205dab5fa1155fa0b1cb3ef35450170d3.

core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
m-vo: Twig does it like this for its own "renderBlock" (see \Twig\Template):

```php
if ($this->env->isDebug()) {
    ob_start();
} else {
    ob_start(function () { return ''; });
}

// render the thing…

return ob_get_clean();
```

Maybe we should go this route as well? I'm not sure I fully understand the consequences of each version. :slightly_smiling_face:

ausi: I think we should do both:

```php
if ($this->env->isDebug()) {
    ob_start();
} else {
    ob_start(function () { return ''; });
}

try { 
    $block($context);
    return ob_get_contents(); 
} finally {
    ob_end_clean();
}
```

As far as I understand the code: The empty handler function ensures that the output never gets accidentally output to the browser, even in cases like memory limit, timeout, segfault. The `finally` ensures that the opened output buffer gets closed even if the render code throws an unhandled exception.

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -0,0 +1,426 @@
...
+        // framework via a PhpTemplateProxyNode. We're removing the source to
+        // not confuse Twig's lexer and parser and just keep the block names.
+        // At some point we may transpile the source to valid Twig instead and
+        // drop the proxy.
+        if ('html5' !== Path::getExtension($source->getPath(), true)) {
+            return $source;
+        }
+
+        preg_match_all(
+            '/\$this->block\s*\(\s*[\'"]([a-zA-Z0-9_-]+)[\'"]\s*\)/i',
```
ausi:
```suggestion
            '/\$this\s*->\s*block\s*\(\s*[\'"]([a-z0-9_-]+)[\'"]\s*\)/i',
```

core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
m-vo: implemented in  e63889a63c48c0fac6253ed0d6bcedf81f704e84

core-bundle/src/DependencyInjection/Compiler/RewireTwigPathsPass.php
```diff
@@ -22,7 +22,10 @@
 class RewireTwigPathsPass implements CompilerPassInterface
 {
     /**
-     * Rewires the registered "addPath" method calls to our decorated service.
+     * Rewires Symfony's "addPath" method calls to our fail tolerant version
+     * of a filesystem loader which tolerates missing paths. Those will occur
+     * as soon as a user removes registered directories (e.g. from within the
+     * backend) and that would otherwise require the container to be rebuilt.
```
m-vo, line 28: No functional changes here, we're now simply replacing the default filesystem loader instead of decorating it.

core-bundle/src/DependencyInjection/Compiler/RewireTwigPathsPass.php
```diff
@@ -39,14 +42,12 @@ static function (array $call): bool {
             return;
         }
 
-        do {
-            $original->removeMethodCall('addPath');
-        } while ($original->hasMethodCall('addPath'));
```
m-vo, line 44: This was fixed in the di component (>= 4.4.20), so no need for the hack anymore.

core-bundle/src/EventListener/BackendRebuildCacheMessageListener.php
```diff
@@ -24,6 +23,8 @@
  */
 class BackendRebuildCacheMessageListener
 {
+    public const CACHE_DIRTY_FLAG = 'contao.template_path_cache_dirty';
+
```
m-vo, line 27: I decided that the default filesystem loader should always ignore missing paths and thus moved the cache flag into here. We might want to revert this, but in practice this made no difference for me but reduced some hard to find errors.

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -0,0 +1,126 @@
...
+            new DynamicExtendsTokenParser($this->hierarchy),
+        ];
+    }
+
+    /**
+     * @see \Contao\CoreBundle\Twig\Interop\PhpTemplateProxyNode
+     * @see \Contao\CoreBundle\Twig\Interop\PhpTemplateProxyNodeVisitor
+     *
+     * @internal
+     */
```
m-vo, line 168: This function is called from within Twig templates if a Contao template should get rendered (there are similar constructs in the Symfony's Twig profiler extension). This basically calls the old template engine with prepopulated blocks and block names and returns the rendered ("parsed") result.

core-bundle/src/Twig/Inheritance/DynamicExtendsTokenParser.php
```diff
@@ -0,0 +1,104 @@
...
+
+        if (1 !== preg_match('%^@Contao/(.*)%', $parent->getAttribute('value'), $matches)) {
+            return;
+        }
+
+        $shortName = $matches[1];
+        $sourcePath = $stream->getSourceContext()->getPath();
+        $parentName = $this->hierarchy->getDynamicParent($matches[1], $sourcePath);
+
+        // Detect loops
```
m-vo: This is just a safety net that prevents hard to find errors (endless loops, memory exhausted) in case something goes wrong.

core-bundle/src/Twig/Inheritance/TemplateHierarchyInterface.php
```diff
@@ -0,0 +1,39 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Inheritance;
+
+interface TemplateHierarchyInterface
```
m-vo, line 18: We technically do not need an interface for this. But it felt wrong to reference the `ContaoFilesystemLoader` (which currently implements the functionality), thus decoupling it.

core-bundle/src/Twig/Loader/TemplateLocator.php
```diff
@@ -0,0 +1,157 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Loader;
+
+use Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle;
+use Symfony\Component\Finder\Finder;
+use Webmozart\PathUtil\Path;
+
+class TemplateLocator
```
m-vo, line 22: This is functionally related to the `AddResourcesPass` but currently cannot reuse due to different requirements. We might want to revisit this (in the future) and try to homogenize it.

m-vo (commented):
I added some remarks/explanations in form of a review. Hopefully makes it easier to go through it.

core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $compiler
+            ->write('echo $this->extensions[')
+            ->repr($this->getAttribute('extension_name'))
+            ->raw(']->renderLegacyTemplate('."\n")
+            ->indent()
+            ->write('$this->getTemplateName(),'."\n")
+            ->write('array_map('."\n")
+            ->indent()
+            ->write('static function(callable $block) use ($context): string {'."\n")
```
ausi:
```suggestion
            ->write('function(callable $block) use ($context): string {'."\n")
```
this must not be static IMO, otherwise the `$this->env->isDebug()` triggers an error.

ausi (commented):
Two things/bugs I noticed:

`{% extends '@Contao/picture_default' %}` works, but `{% include '@Contao/picture_default.html5' %}` requires the `.html5` suffix to find the template. And if there is an overwritten `picture_default` in twig this will not be used then. Can we fix this to work the same? I guess this could otherwise cause confusion.

Extending `fe_page.html5` didn’t work because `An exception has been thrown during the rendering of a template ("sections is not set or not a callable") in "@Contao_ContaoCoreBundle/fe_page.html5".` I guess this is because https://github.com/contao/contao/blob/f277abad6b6a66ace61804b4017e7e2bf8f6c99d/core-bundle/src/Resources/contao/templates/frontend/fe_page.html5#L27 calls the method https://github.com/contao/contao/blob/f277abad6b6a66ace61804b4017e7e2bf8f6c99d/core-bundle/src/Resources/contao/classes/FrontendTemplate.php#L202 on `FrontendTemplate` and you are using the generate `Template` class?

ausi:
I checked this branch out now and played around with it a little bit.

I really like the encoding stuff. This just feels like real data in real twig templates. Trying things out I forgot that we allow HTML in the image caption, and the twig template fixed it for me by still encoding it properly. When I then allowed HTML by adding the the `|raw` filter even the Contao code editor marked it in red for me. This feels so much more sane and secure. 😍

Template folder hierarchy with themes works like a charm! Extending the template from another theme or from the global *templates* folder just works like one would expect. This is a great selling point for getting users switching to twig IMO.

I think this is the way to go, WDYT @contao/developers ?

leofeyer:
> I think this is the way to go, WDYT @contao/developers ?

I think so, too. Should we assign this to the `4.12` milestone?

fritzmg:
> I think so, too. Should we assign this to the `4.12` milestone?

As opposed to `4.13` you mean? Yes, definitely 🙂 . This gives us time to play around with it and work out further kinks while `4.12` is active and provide more additions that will then make it into the next LTS.

core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
m-vo: Good catch, fixed in 816a7c247e13d767bb758f633591f15f8ddcb9a0

leofeyer:
> Yes, definitely 🙂 

I think that is @m-vo's decision to make, because we have only two more weeks before the release of Contao 4.12.0-RC1 and only he can say whether he has the time and energy to finalize the PR until then.

bytehead:
I would like to see it as an experimental feature in 4.12 - but it's up to @m-vo 😎

m-vo:
> I think that is @m-vo's decision to make, because we have only two more weeks before the release of Contao 4.12.0-RC1 and only he can say whether he has the time and energy to finalize the PR until then.

Let's do it! :rocket:  :slightly_smiling_face: 

 The most fundamental things should already be there but tests + things from feedback loops. What's missing but should IMO go into followup PR are a better integration in the BE + Contao specific functions/filters (e.g. for the JSON LD stuff). But I think it would be ok these things won't make it into the 4.12 release.

Toflar:
I'm +1 for experimental support in 4.12. I'm trusting on @ausi and @m-vo here and I have no doubt it'll make a great addition! This will give us time to polish things for the next LTS.

m-vo:
This is basically ready for a first round of reviews, now. I'm still going to…

* add more tests
* try to implement a replacement function for the `section` template method (see todo)
* improve comments/descriptions
 
… in the next days, but this should be rather independent.

m-vo:
image

:muscle:

m-vo:
https://github.com/contao/contao/pull/3170#issue-686209471

> I registered the addSchemaOrg Twig function in the ContaoTemplateExtension extension for now. Note, that this is only temporary. Once #2988 is merged, I'd like to move everything in a single ContaoExtension. Having multiple runtimes is a good idea for SOC, but it turned out that multiple extensions doesn't make any sense for our use case (IMHO BC is not relevant in this case).

This is now already refactored in this PR.

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -68,13 +70,24 @@ trait TemplateInheritance
 	 */
 	protected $intBufferLevel = 0;
 
+	/**
+	 * Allow to delegate rendering to Twig
+	 * @var bool
+	 */
+	protected $blnEnableTwigSurrogateRendering = true;
+
```
ausi:
```suggestion
```

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -68,13 +70,24 @@ trait TemplateInheritance
...
 	/**
 	 * Parse the template file and return it as string
 	 *
 	 * @return string The template markup
 	 */
 	public function inherit()
 	{
+		if ($this->blnEnableTwigSurrogateRendering && null !== ($result = $this->renderTwigSurrogateIfExists()))
```
ausi:
```suggestion
		if (null !== ($result = $this->renderTwigSurrogateIfExists()))
```

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -323,6 +336,23 @@ protected function getTemplatePath($strTemplate, $strFormat='html5', $blnDefault
 
 		return Controller::getTemplate($strTemplate);
 	}
+
+	private function renderTwigSurrogateIfExists(): ?string
```
ausi:
```suggestion
	protected function renderTwigSurrogateIfExists(): ?string
```

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -0,0 +1,209 @@
...
+    public function renderLegacyTemplate(string $name, array $blocks, array $context): string
+    {
+        $template = Path::getFilenameWithoutExtension($name);
+
+        $partialTemplate = new class($template) extends FrontendTemplate {
+            protected $blnEnableTwigSurrogateRendering = false;
```
ausi:
```suggestion
            protected function renderTwigSurrogateIfExists(): ?string
            {
                return null;
            }
```

Would this be a more clean approach? (Not sure though)

core-bundle/src/Twig/Interop/ContextHelper.php
```diff
@@ -0,0 +1,103 @@
...
+     * Adjust Contao template data to seamlessly work as a Twig context.
+     */
+    public static function fromContaoTemplate(Template $template): array
+    {
+        $context = $template->getData();
+
+        array_walk_recursive(
+            $context,
+            static function (&$value, $key): void {
+                if (\is_callable($value)) {
```
ausi: This needs to check for closures I think. Otherwise string values like `'strtolower'` produce unexpected results 😳

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -0,0 +1,438 @@
...
+        $chain = $this->getInheritanceChains()[$this->getIdentifier($name)] ?? [];
+
+        foreach (array_keys($chain) as $path) {
+            try {
+                if (filemtime($path) > $time) {
+                    return false;
+                }
+            } catch (\Exception $e) {
+                return false;
+            }
```
ausi: Does `filemtime()` throw exceptions?

core-bundle/src/Twig/Loader/FailTolerantFilesystemLoader.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param string $namespace A path namespace
+     */
+    public function addPath($path, $namespace = self::MAIN_NAMESPACE): void
+    {
+        try {
+            parent::addPath($path, $namespace);
+        } catch (LoaderError $error) {
+            // Ignore
+
+            return;
```
ausi:
```suggestion
```

ausi:
After https://github.com/contao/contao/pull/2988#discussion_r669456479 is solved this is RTM! 🎉

Awesome work @m-vo !!!!!!!11elf 💪

core-bundle/src/Twig/Interop/ContextHelper.php
m-vo: Good catch! 

Changed in 1c9f019978e0eb34f7faabf42965259d7abe4eeb.

core-bundle/src/Twig/Extension/ContaoExtension.php
m-vo: I like this! :+1: 

Changed in 9edace8392c6dfd05c2a38b1b39361e9b74f7434.

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
m-vo: Handled in 80dcb4fbd13c020ea498146743ef0c29336c2aee.

ausi (dismissed):

core-bundle/src/Twig/Interop/PhpTemplateProxyNodeVisitor.php
```diff
@@ -0,0 +1,76 @@
...
+        return $node;
+    }
+
+    /**
+     * We are replacing the module body with a PhpTemplateProxyNode that will
+     * delegate rendering to the Contao framework on the fly. To support blocks
+     * we're also injecting a BlockNode for each block in the original source
+     * that will return the default Contao block placeholder when called.
+     */
+    private function configurePhpTemplateProxy(ModuleNode $node): ModuleNode
```
leofeyer: The return value of this method is never used (see line 51). Is this correct?

leofeyer (changes requested):
Found one unused return value in an otherwise flawless PR. Great work! 💪🏻

core-bundle/src/Twig/Interop/PhpTemplateProxyNodeVisitor.php
m-vo: Ah good catch! The method is operating on the reference.

Fixed in f65506b0a9a7186af604ba0051db1bffcd02fe77

leofeyer (approved):

leofeyer:
Thank you @m-vo.

m-vo:
Wohoo! :partying_face: :tada:

PR #2991, opened 2021-05-07
Enable "useSSL" by default unless the backend request is insecure

### LINKED ISSUE #2968: Make HTTPS the default protocol for root pages
**Affected version(s)**

Contao 4.11

**Description**

Since #2323 you have to choose between using HTTP and HTTPS for every root page. I think it would make sense to change the default to HTTPS, because thats the recommeded protocol and nowadays most webspaces already use HTTPS by default AFAIK.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2968

This defaults the `useSSL` to HTTPS and only switches to HTTP as default if the backend request itself uses HTTP.

This should lower the chance to run into endless redirects between HTTP and HTTPS.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -1095,7 +1095,8 @@ class tl_page extends Backend
 	{
 		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 
-		if ($request && !$request->isSecure()) {
+		if ($request && !$request->isSecure())
+		{
 			$GLOBALS['TL_DCA']['tl_page']['fields']['useSSL']['default'] = '';
 		}
 	}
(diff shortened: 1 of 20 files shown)


bytehead (dismissed):

Toflar (dismissed):

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1087,6 +1088,19 @@ public function makeRedirectPageMandatory(DataContainer $dc)
 		}
 	}
 
+	/**
+	 * Set useSSL to HTTP if the current request does not use HTTPS
+	 */
+	public function setUseSslDefault(DataContainer $dc)
```
aschempp: this should be a listener and not a DCA method

aschempp (changes requested):

core-bundle/src/Resources/contao/dca/tl_page.php
ausi: Done in 30036212ba38b55b4d3a30d1d669b59e462287b6

aschempp (dismissed):

leofeyer:
Would it make more sense to replace all occurrences of `http://` with `https://` in the functional tests? Then we would not have to add `useSSL: ''` in all the .yml files.

aschempp:
We could also revert the database to not enable SSL by default but use the listener to do so in the back end?

ausi:
I would prefer to use HTTPS as the default in the database because it is the better default IMO. 
But I can change the functional tests to HTTPS if you want?

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2994, opened 2021-05-11
Hide unsaved content elements in the front end

### LINKED ISSUE #2708: Hide unsaved content elements in the front end
**Affected version(s)**

Contao 4.9 (maybe later)

**Description**

**Folgender Sachverhalt:**
Kopiert man ein Inhaltselement im Backend und fügt es an gewünschter Stelle im Backend wieder ein, dann öffnet sich ja wie gewünscht die Bearbeitungsmaske. In diesem Moment wird das kopierte Element auch schon in der Datenbank angelegt, allerdings mit tstamp=0.
Speichert man das Element aber **NICHT** und klickt beispielsweise dann, weil mach sich vielleicht umentschieden hat :-), auf die Dateiveraltung oder ähnliches, dann ist das Element **ONLINE**! Das Element ist dann solange ONLINE bis man wieder auf irgendeinen Artikel/Modul im Backend in die Detailansicht (childTable tl_content) klickt. Hier wird dann wohl eine Query gestartet und alle Elemente mit tstamp=0 gelöscht!

Ich weiss nicht ob es wirklich ein Bug ist aber wenn der Kunde das Frontend anschaut ist das kopierte Element da und will er es dann im Backend löschen ist es weg! Ist irgendwie inkonsistent.

Meine Meinung:
Wenn das Element im Frontend erscheint sollte es dann auch im Backend da sein ohne das man es gespeichert hat **oder** es sollte im Frontend gar nicht erscheinen...!

Danke und Grüße

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2708
| Docs PR or issue | -

This only affects the article module and does not change the behavior of `Controller::getContentElement()`, which must remain backwards compatible.

### DIFF
--- a/core-bundle/src/Resources/contao/models/ContentModel.php
+++ b/core-bundle/src/Resources/contao/models/ContentModel.php
@@ -425,6 +425,9 @@ class ContentModel extends Model
 			$arrColumns[] = "$t.invisible='' AND ($t.start='' OR $t.start<='$time') AND ($t.stop='' OR $t.stop>'$time')";
 		}
 
+		// Skip unsaved elements (see #2708)
+		$arrColumns[] = "NOT ($t.tstamp=0 AND $t.invisible='')";
+
 		if (!isset($arrOptions['order']))
 		{
 			$arrOptions['order'] = "$t.sorting";
@@ -461,6 +464,9 @@ class ContentModel extends Model
 			$time = Date::floorToMinute();
 			$arrColumns[] = "$t.invisible='' AND ($t.start='' OR $t.start<='$time') AND ($t.stop='' OR $t.stop>'$time')";
 		}
+
+		// Skip unsaved elements (see #2708)
+		$arrColumns[] = "NOT ($t.tstamp=0 AND $t.invisible='')";
 
 		if (!isset($arrOptions['order']))
 		{
--- a/core-bundle/src/Resources/contao/models/ContentModel.php
+++ b/core-bundle/src/Resources/contao/models/ContentModel.php
@@ -426,7 +426,7 @@ class ContentModel extends Model
 		}
 
 		// Skip unsaved elements (see #2708)
-		$arrColumns[] = "NOT ($t.tstamp=0 AND $t.invisible='')";
+		$arrColumns[] = "$t.tstamp!=0";
 
 		if (!isset($arrOptions['order']))
 		{
@@ -466,7 +466,7 @@ class ContentModel extends Model
 		}
 
 		// Skip unsaved elements (see #2708)
-		$arrColumns[] = "NOT ($t.tstamp=0 AND $t.invisible='')";
+		$arrColumns[] = "$t.tstamp!=0";
 
 		if (!isset($arrOptions['order']))
 		{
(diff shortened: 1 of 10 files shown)


bytehead (dismissed):

fritzmg:
What about news and events though? I think this should rather be fixed in the `ContentModel::findPublishedByPidAndTable` method.

fritzmg (changes requested):
What about news and events though? I think this should rather be fixed in the `ContentModel::findPublishedByPidAndTable` and `ContentModel::countPublishedByPidAndTable` method.

leofeyer:
No, we explicitly agreed on not adding this logic to anything that has a `published` field (see the original ticket). Also, we must not change the behavior of the existing methods in a bugfix release.

fritzmg:
> No, we explicitly agreed on not adding this logic to anything that has a `published` field (see the original ticket).

Yes, content elements do not have a `published` field, what's your point there? My suggestion would not change that.

> Also, we must not change the behavior of the existing methods in a bugfix release.

Unless the current behavior is bugged, no? If you do not want to change it in the model method, then any bundle, including `contao/core-bundle`, `contao/news-bundle`, `contao/calendar-bundle` and any other third party bundle using content elements are themselves responsible for checking for an empty `tstamp`, which I find quite inconvenient and only leads to inconsistent behaviour.

leofeyer:
I agree that this only solves the problem for the core distribution and that this might be inconsistent. Let's therefore discuss again it in the next Mumble call.

aschempp:
As @leofeyer already said, we have already discussed this. All other entities have a `published` field if relevant. They are unpublished by default, so it can/should/will never happen to be "publicly available" with a `tstamp=0` (except e.g. for database manipulations, which are not our scope). Only content elements are `NOT invisible` by default, therefore they. must not show up until someone saved them (where `invisible` is possibly activated).

fritzmg:
@aschempp but that's not what we are discussing here. We are discussing _where_ the `tstamp` check for content elements should be implemented.

leofeyer:
As discussed in Mumble on May 12th, we want to exclude elements with `tstamp=0` and `invisible=''` in the ContentModel and FormFieldModel `findPublishedBy*()` methods and in the `ContentElement::isHidden()` method.

leofeyer:
I have adjusted the implementation in daa5b606c841e3bc3cbaf9a4e084668ed223d16a.

core-bundle/src/Resources/contao/models/ContentModel.php
```diff
@@ -425,6 +425,9 @@ public static function findPublishedByPidAndTable($intPid, $strParentTable, arra
 			$arrColumns[] = "$t.invisible='' AND ($t.start='' OR $t.start<='$time') AND ($t.stop='' OR $t.stop>'$time')";
 		}
 
+		// Skip unsaved elements (see #2708)
+		$arrColumns[] = "NOT ($t.tstamp=0 AND $t.invisible='')";
```
leofeyer: @contao/developers Why exactly have we concluded that we need to check both the timestamp and the invisible field? Shouldn't we ignore all content elements with `tstamp=0`?

fritzmg: Yeah I think the additional `invisible` check is superfluous there 🤔

ausi: Yes, the `invisible` check is already done 4 lines above so we only have to check for `tstamp!=0` IMO.

fritzmg (approved):

aschempp (approved):

PR #2996, opened 2021-05-11
Correctly strip the root path in the FilesModel and Dbafs classes

### LINKED ISSUE #2824: FilesModel::findByPath() does not normalize windows paths
**Affected version(s)**

Contao 4.11 (most probably 4.9 too)

**Description**

In the `ImageFactory` the `uploadPath` can contain backslashes like `C:\foo\www/files` but the image path is mostly in the normalized form like `C:/foo/www/files/bar.jpg`. This creates problems in https://github.com/contao/contao/blob/1229843fcd1ec18522b85287ee0ebd915a950c18/core-bundle/src/Resources/contao/models/FilesModel.php#L254 because the root path is not stripped correctly. This can result in missing important parts for images like it was the case in the following discussion on slack: https://contao.slack.com/archives/CK4J0KNDB/p1614694011186200

I think we should use the pathutils library to strip the directories there to fix the issue.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2824
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
@@ -12,6 +12,7 @@ namespace Contao;
 
 use Contao\Filter\SyncExclude;
 use Contao\Model\Collection;
+use Webmozart\PathUtil\Path;
 
 /**
  * Handles the database assisted file system (DBAFS)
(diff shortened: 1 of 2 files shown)


fritzmg (approved):

ausi (approved):

leofeyer:
The change seems to cause new issues: https://github.com/contao/contao/issues/3001 😢

PR #3004, opened 2021-05-13
Fix an inconsistency in the routing functional tests

Toflar:
Currently, there's no distinction between our own 404 page type and the 404 exception. 
Our tests are green only because we use `assertStringContainsString` and our functional tests use `(404 Not Found)` in the Contao 404 page type which happens to be the same as what the general 404 Symfony response creates.

I've noticed this with https://github.com/contao/contao/pull/2975 where only the `prefer-lowest` tests were failing instead of all the tests which would have been correct. Seems like Symfony removed the default `Accept` header in some version so we should have had failing tests all along as our `PrettyErrorScreenListener` was not triggered.
However, nobody ever noticed because `(404 Not Found)` occurs in both, our own page type test and the exception.
This is now distinguished and fixed properly.

### DIFF
--- a/core-bundle/tests/Fixtures/Functional/Routing/root-with-home.yml
+++ b/core-bundle/tests/Fixtures/Functional/Routing/root-with-home.yml
@@ -28,7 +28,7 @@ tl_page:
       pid: 1
       sorting: 256
       tstamp: 1539698035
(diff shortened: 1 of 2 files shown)


aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #3006, opened 2021-05-14
Allows to save the value 0 in the keyValueWizard

doishub:
This PR fixes the bug that 0 values cannot be entered in the KeyValueWizard. 

It seems that this possibility was [removed in the preparation for PHP8](https://github.com/contao/contao/blame/e8491d709c5055fbdfe54161553b3888210044b7/core-bundle/src/Resources/contao/widgets/KeyValueWizard.php#L68).
In further versions of contao 4.9 it was possible to set 0 as value.

(https://github.com/contao/contao/issues/2347, https://github.com/contao/contao/pull/2342)

PR #3007, opened 2021-05-17
Allow saving the value `0` in the KeyValueWizard

doishub:
Replaces the pull request #3006.

---

This PR fixes the bug that 0 values cannot be entered in the KeyValueWizard.

It seems that this possibility was removed in the preparation for PHP8.
In further versions of contao 4.9 it was possible to set 0 as value.

(#2347, #2342)

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/KeyValueWizard.php
+++ b/core-bundle/src/Resources/contao/widgets/KeyValueWizard.php
@@ -66,7 +66,7 @@ class KeyValueWizard extends Widget
 			foreach ($options as $key=>$option)
 			{
 				// Unset empty rows
-				if (!$option['key'])
+				if (trim($option['key']) === '')
 				{
 					unset($options[$key]);
 					continue;


core-bundle/src/Resources/contao/widgets/KeyValueWizard.php
```diff
@@ -66,7 +66,7 @@ public function validate()
 			foreach ($options as $key=>$option)
 			{
 				// Unset empty rows
-				if (!$option['key'])
+				if (trim($option['key']) === '')
```
leofeyer, line 69: Is `$option['key']` always set? Otherwise we should probably add `!isset($option['key']) || `.

doishub: I think it's fine this way. If no values are passed, an array with empty values is returned:
image

ausi:
```suggestion
				if (trim($option['key'] ?? '') === '')
```

doishub: From my point of view, `$option['key']` can never be `null` in the `validate` method, because the `keyValueWizard` widget always passes an empty array. Should I take this anyway?

ausi: Agree, `trim($option['key']) === ''` should be fine then.

leofeyer:
Thank you @doishub.

PR #3014, opened 2021-05-18
Ensure clean response context and unify meta handling

Toflar:
This PR

* ensures the new `ResponseContext` does contain clean and unencoded values in preparation for the future (one thing less to migrate)
* unifies the handling of insert tags for meta description. Sometimes it was stripped, sometimes replaced


fritzmg:
`Frontend::prepareMetaDescription` also limits the meta description to 320 characters. With this PR no such limit would be in place at all anymore. If we still do want to limit the meta description to an arbitrary character count, then it should probably be implemented in `PageRegular.php#224`.

Generally, I am in favour of **not** arbitrarily limiting the meta description character count. However, for things like news and events where the meta description is automatically derived from the teaser (if no dedicated meta description is given) that should probably still be limited?

Toflar:
The limit is still applied in the context itself. If there's an agreement on what makes sense, I think it's fine to keep that as is.

core-bundle/src/Routing/ResponseContext/WebpageResponseContext.php
```diff
@@ -47,6 +51,9 @@ public function getMetaDescription(): string
 
     public function setMetaDescription(string $metaDescription): self
     {
+        $metaDescription = self::cleanString($metaDescription);
+        $metaDescription = StringUtil::substr($metaDescription, 320);
```
Toflar: Here @fritzmg

fritzmg: Ah yes, sorry, I missed that.

fritzmg (dismissed):

m-vo:
IMHO truncating strings should be done in the template. Then it's easily adjustable and you could even reuse the original data.

ausi:
Stripping the tags should not be done in the `ResponseContext` I think. It receives raw (=decoded) content that can contain `<` or `>` that has a different meaning from HTML.

For example if I enter `I <3 Contao` in the Backend, Contao stores this as `I &lt;3 Contao` in the Database, with `StringUtil::decodeEntities($pageModel->title)` it gets converted back to `"I <3 Contao"`. But now with the `cleanString()` method it tries to remove tags and ends up with `"I "`.

Stripping tags should be done before decoding entities and it should be the responsibility of the Modules and the code that reads from `$pageModel` IMO.

> IMHO truncating strings should be done in the template.

Agree, I would put the stripping of newlines and truncating the description in the template (or to `PageRegular` when Setting `$this->Template->…`).

Toflar:
So first, this PR does not introduce nor change the current behaviour. The length was hardcoded so far and we have been using `strip_tags()` before. All it does is ensuring that it works the same everywhere and `ResponseContext` contains unencoded data.

While I agree that we might want to make things dynamically and optimize, it's not really related to this PR.
What actually happened was that because things are stored encoded, `strip_tags()` did not strip the HTML so it ended up landing in the description - is that the desired outcome? Then we can just remove the "cleaning" and we're done.

ausi:
> So first, this PR does not introduce nor change the current behaviour.

I may be mistaken here, but entering `I <3 Contao` in a News title works currently in Contao 4.11 and as far as I understand the code of this PR it does not work anymore after these changes.

> What actually happened was that because things are stored encoded, `strip_tags()` did not strip the HTML so it ended up landing in the description - is that the desired outcome?

IMO yes. Currently a news with `Multiline{{br}} News-Title` as title gets converted to `Multiline News-Title` correctly. The same happens if I enable `allowHtml` for the title and enter `Multiline<br> News-Title`.

Toflar:
I have moved the substr to `fe_page` after a quick discussion with @ausi.
Really looking forward to Twig because `{{ title[:320] }}` would have been a lot nicer :)

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
```diff
@@ -9,7 +9,7 @@
 
     <?php $this->block('meta'); ?>
       <meta name="robots" content="<?= $this->robots ?>">
-      <meta name="description" content="<?= $this->description ?>">
+      <meta name="description" content="<?= mb_substr($this->description, 0, 320) ?>">
```
fritzmg: That would be a behaviour change. You need to use `StringUtil::substr($this->description, 320)`, otherwise the ellipsis won't get added and it would otherwise cut off words.

fritzmg (changes requested):

fritzmg:
> Really looking forward to Twig because `{{ title[:320] }}` would have been a lot nicer :)

But would not have the same result 🙃 (see comment)

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
Toflar: Yes, which we don't want. It makes no sense at all and `StringUtil::substr()` does way too much bad stuff. E.g. it's also calling `strip_tags()` for who knows what reason.

fritzmg: > It makes no sense at all

Imho it does make sense to create a description that reads

> Auch gibt es niemanden, der den Schmerz an sich liebt, sucht oder wünscht, nur, weil er Schmerz ist, es sei denn, es kommt zu zufälligen Umständen, in denen Mühen und Schmerz ihm große Freude bereiten können. Um ein triviales Beispiel zu nehmen, wer von uns unterzieht sich je anstrengender körperlicher Betätigung, außer …

instead of

> Auch gibt es niemanden, der den Schmerz an sich liebt, sucht oder wünscht, nur, weil er Schmerz ist, es sei denn, es kommt zu zufälligen Umständen, in denen Mühen und Schmerz ihm große Freude bereiten können. Um ein triviales Beispiel zu nehmen, wer von uns unterzieht sich je anstrengender körperlicher Betätigung, auße

fritzmg: > E.g. it's also calling strip_tags() for who knows what reason.

It strips tags to avoid creating invalid HTML. If you want to clip text with HTML, you need to use `StringUtil::substrHtml`.

m-vo:
For the record, you can easily do all of this (and more) with Twig as well using the unicode filter `|u`.
https://twig.symfony.com/doc/3.x/filters/u.html - see truncate

Toflar:
@leofeyer please make sure to merge this as authored by @ausi when squashing because he did all the work :)

ausi (dismissed):

ausi (dismissed):

fritzmg:
I am a bit confused by the new `StringUtil` methods. Their naming and description are not hinting at the fact, that any HTML tags would also be stripped from the string.

Could you may be give some hints why this elaborate solution is now necessary anyway? Regarding the `I <3 Contao` use-case I would have thought that simply stripping HTML tags first and then decoding the entities would suffice.

ausi:
> I am a bit confused by the new `StringUtil` methods. Their naming and description are not hinting at the fact, that any HTML tags would also be stripped from the string.

The method `getRawDecodedValue` *Converts an input-encoded string back to its raw UTF-8 value it originated from.* The tags it strips are from the insert tags because an input-encoded string itself cannot have HTML tags.

With `getRawDecodedValueFromHtml()` we should probably improve the description of the method. How about: *Gets the raw text value of an HTML string with tags removed entities decoded and multiple white space characters combined to single spaces or newlines.*

> Could you may be give some hints why this elaborate solution is now necessary anyway?

Did you take a look at the `StringUtilTest`? It should show why all these steps are necessary.

> Regarding the `I <3 Contao` use-case I would have thought that simply stripping HTML tags first and then decoding the entities would suffice.

That’s basically all we do now. The `Ensure valid UTF-8` and `[{]` parts are for security reasons.

ausi (dismissed):

fritzmg (dismissed):

m-vo (dismissed):
Special love for all the encoding voodoo. ❤️

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -971,6 +971,77 @@ public static function ampersand($strString, $blnEncode=true): string
...
+		$strValue = static::restoreBasicEntities($strValue);
+		$strValue = static::decodeEntities($strValue);
+
+		// Ensure valid UTF-8
+		if (preg_match('//u', $strValue) !== 1)
+		{
+			$substituteCharacter = mb_substitute_character();
+			mb_substitute_character(0xFFFD);
+
+			$strValue = mb_convert_encoding($strValue, 'UTF-8', 'UTF-8');
```
aschempp, line 991: this means we no longer support anything else than UTF-8 right? I'm totally fine with that, but isn't there a localconfig for the character encoding?

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
```diff
@@ -94,4 +94,4 @@
 
   <?= $this->mootools ?>
 </body>
-</html>
```
aschempp: this change should be reverted. You might want to add an `.editorconfig` plugin to your IDE 😉

aschempp (commented):
The decoding looks good in general, but I don't like 2 things
 1. The StringUtil methods do not make it clear that this is about cleaning up a string, e.g. `getRawDecodedValueFromHtml` strips als HTML tags and returns a text with line breaks instead of `<br>`.
 2. We're making hardcoded assumptions about the DCA configuration of the `tl_page` fields. DCA fields can be reconfigured to e.g. allow HTML and then the method calls would be wrong.

I also wonder if there is a reason to distinguish between with or without HTML text, can't the same method just "clean up" both cases? What if I want to replace insert tags and then strip HTML to have line breaks?

ausi:
> The StringUtil methods do not make it clear that this is about cleaning up a string

How would you change the method description *“Gets the raw text value of an HTML string with normalized white space.”* to make that more clear?

> DCA fields can be reconfigured to e.g. allow HTML and then the method calls would be wrong.

This should work fine, as both methods strip HTML tags.

> I also wonder if there is a reason to distinguish between with or without HTML text, can't the same method just "clean up" both cases? What if I want to replace insert tags and then strip HTML to have line breaks?

The difference is line breaks and white space normalization. `\n` in HTML is not a line break, but it is in a non-HTML textarea field.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: I think we should deprecate the character encoding in the config and force `UTF-8`. See also #3015

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
Toflar: That's bundled with phpstorm, so 🤷

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
aschempp: deprecate means we still need to support it though? Or we agree that this is an acceptable break in the next minor 🙃

Toflar: Honestly, nothing ever worked in anything other than UTF-8 because why would it? Imho this config can be removed anywhere. I don't even understand why anyone would have ever used that.

leofeyer: Back in 2006, not everyone had UTF-8 or was using UTF-8, therefore the charset was configurable (and still is). While I agree that we could enforce UTF-8, it would still be an unnecessary BC break.

ausi: I can add a check if the config charset is `utf-8` and only then execute this part of the code.

aschempp:
> > DCA fields can be reconfigured to e.g. allow HTML and then the method calls would be wrong.

> This should work fine, as both methods strip HTML tags.

Yes, but it would not add line breaks for HTML tags?

ausi:
> Yes, but it would not add line breaks for HTML tags?

Correct. But allowing HTML in the alias field for example also would not result in HTML suddenly working in URLs.
IMO if someone reconfigures the DCA, they are responsible for possible side effects.

Toflar:
I would like to remind you that this was also the case ever since. The existing code did not consider any DCA settings either.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: So, what do we do here? I would leave it as it is in this PR and add the deprecation for the config charset in #3015

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
aschempp: Two options for me:
 1. can we just use the `Config::get('characterSet')` as argument for `mb_convert_encoding`?
 2. Alternatively, let's create a new issue assigned to 4.12 to remove encoding (or check on this again)

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -221,7 +221,7 @@ protected function prepare($objPage)
 
 		// Assign the title and description
 		$this->Template->title = strip_tags($this->replaceInsertTags($objLayout->titleTag));
```
aschempp, line 223: Looks like you need to use the new method here as well?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
aschempp: Also, did you see there's a `StringUtil::convertEncoding()` method that does pretty much the same?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -971,6 +971,77 @@ public static function ampersand($strString, $blnEncode=true): string
...
 	}
+
+	/**
+	 * Converts an input-encoded string back to its raw UTF-8 value it originated from.
+	 *
+	 * It handles all Contao input encoding specifics like insert tags, basic entities and encoded entities.
+	 *
+	 * @param bool $blnRemoveInsertTags True to remove insert tags instead of replacing them
+	 */
+	public static function getRawDecodedValue(string $strValue, bool $blnRemoveInsertTags = false): string
```
aschempp: This method name needs to be changed. First of all, the `StringUtil` class does not have **any** `get…` method, because you're not _getting_ anything from it. Secondly, its unclear what _raw decoded_ means, and since this is the `StringUtil` class, what's _value_ but a string?.

Here some suggestions
 1. `revertInputEncoding`
 2. `decodeInput` / `restoreInput`
 3. `convertToPlainText` (which mean the other method would be `convertToPlainHtml`)

I like the third option the most, at least if I understand the method correctly. _plain text_ would make me understand this might strip html and replace insert tags for example.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: > Alternatively, let's create a new issue assigned to 4.12 to remove encoding (or check on this again)

Done https://github.com/contao/contao/issues/3050

core-bundle/src/Resources/contao/pages/PageRegular.php
ausi: I don’t think so.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: If we go with `convertToPlainText` the second method should be named `convertToPlainTextFromHtml` IMO. WDYT?

Toflar: It's not "plain text" though. It's raw. Not plain... :)

ausi: `convertToRawText()` and `convertToRawTextFromHtml()` or 
`decodeToRawText()` and `decodeToRawTextFromHtml()` ?

Toflar: I'm fine with either of them, yes :) @aschempp ?

aschempp: @Toflar Do you refer to `raw` from Twig? Then that would make sense. Otherwise to me, plain text means no encoding, no HTML etc., just what you would write in a (plain text) email.

In almost any case, running either of the method results in the same output, right? Expect the newline stuff for HTML, which is not "obvious" by "...FromHtml`. Maybe we can just have one method with an argument to `$blnConvertHtmlToNewline` ?

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: We almost exclusively use the new methods to convert something from user input for one of the meta attributes (title, description). In this case, that logic wouldn't be applied if I use anything but the standard `{{page::pageTitle}}` insert tag?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
Toflar: No, I refer to English - the language. "Plain" means not encrypted to me.

Toflar: 👎 for more arguments. Methods are better for that.

ausi: `$blnConvertHtmlToNewline` is very confusing IMO, `decodeToRawTextFromHtml()` with the current method description should be way better understandable I think.

core-bundle/src/Resources/contao/pages/PageRegular.php
ausi: `$this->Template->title` needs to be encoded, the new methods are used to get the “decoded” raw value.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
aschempp: https://en.wikipedia.org/wiki/Plain_text 🤷 
Also, searching for "raw text" on [DDG](https://duckduckgo.com) returns that wikipedia article for me 😂 

a better name for the argument would be `$preserveHtmlNewlines`. If looking at both method names, I would assume I must use the `FromHtml` version if my text contains HTML, but that wouldn't be correct.

Toflar: Yeah can we FINALLY just do something?

core-bundle/src/Resources/contao/pages/PageRegular.php
aschempp: Then maybe I don't understand the purpose of the new methods at all. Isn't that exactly what https://github.com/contao/contao/pull/3014/files#diff-0d3a685427658e56073dd7043ebcaa18132e0c77bc6cd9bb56409274f2503f6dR207 does a few lines above? The user input is first converted to plain text through the new methods, and then encoded with `htmlspecialchars`?

So the new methods are not to be used in general whenever we need to "convert" a user input to a plain text, but only in the specific case of the response context?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: > I would assume I must use the `FromHtml` version if my text contains HTML, but that wouldn't be correct.

That would be correct.

> Yeah can we FINALLY just do something?

I’m happy with any name for the two methods.

aschempp: > > I would assume I must use the FromHtml version if my text contains HTML, but that wouldn't be correct.

> That would be correct.

No it would not? It would only be correct if I want to keep newlines. Which for a page title I do not want. Both methods strip the HTML?

core-bundle/src/Resources/contao/pages/PageRegular.php
ausi: The new methods are meant to be used everywhere where we want “raw” non-encoded data. In new data objects like the response context, we want to already use raw data instead of input-encoded data. In the far future we want “raw” non-encoded data everywhere and only encode as a final step in the template rendering code.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: We want to name the methods `inputEncodedToPlainText()` and `htmlToPlainText()`.
And to have a new method `revertInputEncoding()` that only reverts the encoding

ausi (dismissed):

fritzmg (dismissed):

aschempp (dismissed):

leofeyer:
Thank you @Toflar.

Toflar:
Just to make sure I don't take any credit for other people's achievements: This here is all @ausi's wonderful work. :heart:
It would've taken me ages to wrap my head around all the encoding/decoding stuff.
Glad to have you on our team, Martin!

PR #3017, opened 2021-05-20
Handle UUIDs in the file metadata class

m-vo:
| Q                | A
| -----------------| ---

This PR adds a new `uuid` key to our file metadata, that gets set inside the `FigureBuilder` when it is defining meta data. 

As `Metadata` is immutable we now also have a `with()` method that allows merging by returning a new instance with the merged values.

### DIFF
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -45,8 +45,8 @@ class Metadata
     }
 
     /**
-     * Get a new metadata representation that also contains the given values.
-     * Existing keys will get overwritten.
+     * Returns a new metadata representation that also contains the given
+     * values. Existing keys will be overwritten.
      *
      * @param array<string, mixed> $values
      */
@@ -88,7 +88,7 @@ class Metadata
     }
 
     /**
-     * Get a UUID reference in ASCII format or null if not set.
+     * Returns a UUID reference in ASCII format or null if not set.
      */
     public function getUuid(): ?string
     {
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -27,6 +27,7 @@ class Metadata
     public const VALUE_CAPTION = 'caption';
     public const VALUE_TITLE = 'title';
     public const VALUE_URL = 'link';
+    public const VALUE_UUID = 'uuid';
 
     /**
      * Key-value pairs of metadata.
@@ -84,6 +85,11 @@ class Metadata
     public function getUrl(): string
     {
         return $this->values[self::VALUE_URL] ?? '';
+    }
+
+    public function getUuid(): ?string
+    {
+        return $this->values[self::VALUE_UUID] ?? null;
     }
 
     /**
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -87,6 +87,9 @@ class Metadata
         return $this->values[self::VALUE_URL] ?? '';
     }
 
+    /**
+     * Get a UUID reference in ASCII format or null if not set.
+     */
     public function getUuid(): ?string
     {
         return $this->values[self::VALUE_UUID] ?? null;
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -35,9 +35,27 @@ class Metadata
      */
     private $values;
 
+    /**
+     * @param array<string, mixed> $values
+     */
     public function __construct(array $values)
     {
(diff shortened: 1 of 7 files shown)


Toflar (dismissed):

core-bundle/src/File/Metadata.php
```diff
@@ -68,6 +87,11 @@ public function getUrl(): string
         return $this->values[self::VALUE_URL] ?? '';
     }
 
+    public function getUuid(): ?string
```
ausi, line 93: Should we add a doccomment that tells us if it is a binary 16-byte UUID or in the ASCII format?

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -597,14 +603,19 @@ private function onDefineMetadata(): ?Metadata
...
         // container from the default meta fields with empty values instead
         $metaFields = $this->filesModelAdapter()->getMetaFields();
 
-        return new Metadata(array_combine($metaFields, array_fill(0, \count($metaFields), '')));
+        $data = array_merge(
+            array_combine($metaFields, array_fill(0, \count($metaFields), '')),
+            $fileReferenceData
+        );
+
+        return new Metadata($data);
```
fritzmg, line 626:
```suggestion
        $metadata = new Metadata(array_combine($metaFields, array_fill(0, \count($metaFields), '')));

        return $metadata->with($fileReferenceData);
```

May be? Then the merging is consistently handled via `Metadata::with`.

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -584,8 +584,14 @@ private function onDefineMetadata(): ?Metadata
...
 
+        $fileReferenceData = array_filter([
+            Metadata::VALUE_UUID => null !== $this->filesModel ? $this->filesModel->uuid : null,
+        ]);
+
         if (null !== $this->metadata) {
-            return $this->metadata;
+            return $this->metadata
+                ->with($fileReferenceData)
+            ;
```
fritzmg:
```suggestion
            return $this->metadata->with($fileReferenceData);
```

It's a single line in line 606 as well, so may be a single line here as well?

core-bundle/src/Image/Studio/FigureBuilder.php
m-vo: Then you'd potentially create two instances vs one though. :fox_face:

core-bundle/src/Image/Studio/FigureBuilder.php
m-vo: The CS fixer did this. :cry:

core-bundle/src/File/Metadata.php
m-vo: Is the one inside `FilesModel->uuid` always in ASCII format or do we need to enforce this when setting it?

ausi: I think `FilesModel->uuid` is always in binary format.

Toflar: Let's enforce ASCII

core-bundle/src/Image/Studio/FigureBuilder.php
fritzmg: True, but that's the case with the previous returns anyway, isn't it?

core-bundle/src/File/Metadata.php
ausi: If we plan to use objects for UUIDs in the future we could return an uuid object here?
`\Symfony\Component\Uid\Uuid::fromString(...)`

core-bundle/src/Image/Studio/FigureBuilder.php
m-vo: The other two cases aren't build from scratch but from an explicitly set metadata object, though. There we need to go this route whereas in this case we're constructing the set on the fly.

I mean, at the end it's what you prefer, the memory + performance difference is negligible.

core-bundle/src/File/Metadata.php
m-vo: Done in 121f44875077ab1dce100ca2b1ed9245f5e9ff74 + e4a7bfee65eb89354e5614bc4807b2a6dc2c894b.

ausi (dismissed):

fritzmg (dismissed):

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3019, opened 2021-05-20
Dispatch an event to define metadata inside the FigureBuilder

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | todo

This introduces a `DefineMetadataEvent` that gets dispatched inside the `FigureBuilder` as soon as metadata is requested. It is run *after* the `FigureBuilder` logic and allows to remove/replace/add `Metadata`.

### DIFF
--- a/core-bundle/src/Image/Studio/FigureBuilder.php
+++ b/core-bundle/src/Image/Studio/FigureBuilder.php
@@ -23,6 +23,7 @@ use Contao\PageModel;
 use Contao\Validator;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Webmozart\PathUtil\Path;
 
 /**
@@ -710,7 +711,7 @@ class FigureBuilder
         return $framework->getAdapter(Validator::class);
     }
 
-    private function eventDispatcher()
+    private function eventDispatcher(): EventDispatcherInterface
     {
         return $this->locator->get('event_dispatcher');
     }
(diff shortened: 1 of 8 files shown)


Toflar (dismissed):

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -705,6 +710,11 @@ private function validatorAdapter()
         return $framework->getAdapter(Validator::class);
     }
 
+    private function eventDispatcher()
```
ausi:
```suggestion
    private function eventDispatcher(): EventDispatcherInterface
```

Should we use the `Symfony\Contracts\EventDispatcher\EventDispatcherInterface` here?

m-vo: sounds good

ausi (dismissed):

Toflar (dismissed):

leofeyer:
I would prefer `FigureMetadataEvent` or `FigureBuilderMetadataEvent` instead of `DefineMetadataEvent`, because the latter does not really say what kind of metadata it handles (there could be page metadata, news metadata, file metadata etc.). Or is this a general event that we can use for any metadata? Then I would just call it `MetadataEvent`.

Toflar:
It is `FileMetadataEvent`, I guess. The class name is `Contao\CoreBundle\File\Metadata`  :) (it represents meta data of `tl_files`).

m-vo:
We could also put the event in the File namespace. But that's not consistent with the other ones.

I have no strong preferences but would try to avoid having the FigureBuilder in the name.

leofeyer:
Then let's go with `FileMetadataEvent` but keep the class in the `Event` namespace.

m-vo:
Now renamed to `FileMetadataEvent` in 4babc1e59f1936c24f80c9496271e842af908220. I also added a small test for the event class in 1ea7904bd20701226529d385028863dcec2f9b7f.

ausi (dismissed):

Toflar (dismissed):

leofeyer:
Thank you @m-vo.

PR #3028, opened 2021-05-26
Do not require ocramius/proxy-manager

leofeyer:
As suggested here: https://github.com/contao/conflicts/pull/26#issuecomment-848089167


aschempp (dismissed):
Idk what we use the proxy-manager for at all, but I'm fine with the change 😊

ausi:
> Idk what we use the proxy-manager for at all

We only use it here I think: https://github.com/contao/contao/blob/ad86686181b7bd85ad1fa00d37a497a9e4e84e70/manager-bundle/src/HttpKernel/ContaoKernel.php#L312-L314

aschempp:
so that should actually be a dev requirement?

Toflar (changes requested):
We use it because we want to support lazy services in the container.
However, it seems like we could drop this dependency and the method call altogether in favor of the `symfony/proxy-manager-bridge`. The container will then automatically support lazy services - a `setProxyInstantiator()` call is not necessary anymore.


See https://symfony.com/doc/current/service_container/lazy_services.html.

Toflar:
Ah, that's not true. We still need to set it as we're overriding the whole method: https://github.com/symfony/symfony/blob/e34cd7dd2c6d0b30d24cad443b8f964daa841d71/src/Symfony/Component/HttpKernel/Kernel.php#L701

But we might still require the bridge instead (that already requires the correct lts dependency).

Toflar:
@aschempp why do we override the whole method there and not call parent? Is that desired?

leofeyer:
I have removed the requirement in 7a9deebd07e1c0a70cb08d8a505d1711ca45190a, because we already require `symfony/proxy-manager-bridge`.

Toflar (approved):

PR #3029, opened 2021-05-26
Do not modify the select name at runtime

### LINKED ISSUE #2622: Side-effects in FormSelectMenu::generate()
Hello,

while working with Haste Forms and Contao Widgets (in this case `FormSelectMenu` with `multiple: true`) I experienced the following:

~Firstly:~ When you call the `generate` method of `FormSelectMenu` more than once, the array brackets will be added multiple times to the widget instance name (`field_name[]`, `field_name[][]`, `field_name[][][]`).

I expect this method to always return the same result. Same applies for the `parse()` method: It also modifies the widget instance name. I would expect this to only parse the template with no side-effects. 

~Secondly, the `generate` method is only called in `form_explanation.html5`, but nowhere else (in the Core Bundle at least). Shouldn't this method be deprecated as apparently the prefered way is calling `parse()`?~

Edit: I forgot about the back-end 🙈 In the `be_widget*.html5` templates the `generateWithErrors()` method is called (which calls `generate`).

Cheerio

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2622

@bezin can you please test this?

### DIFF
--- a/core-bundle/src/Resources/contao/forms/FormSelectMenu.php
+++ b/core-bundle/src/Resources/contao/forms/FormSelectMenu.php
@@ -162,9 +162,13 @@ class FormSelectMenu extends Widget
 	 */
 	public function __get($strKey)
 	{
-		if ($strKey == 'options')
+		switch ($strKey)
 		{
-			return $this->arrOptions;
+			case 'options':
+				return $this->arrOptions;
+
+			case 'name':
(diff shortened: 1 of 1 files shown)


ausi (approved):

bezin:
LGTM, thank you!

PR #3032, opened 2021-05-26
ResponseContext: Use service composition instead of concrete implementations

Toflar:
https://github.com/contao/contao/pull/2962 made it evident that we need to adjust the current implementation. 
The problem we're facing is that by extending another response context, you have to know what this response context itself requires as constructor arguments.
For example, the `WebpageResponseContext` in this PR will get a new `JsonLdMangager`. The `ContaoWebpageResponseContext` then has to be extended too which is no problem for the core but it shows that we would introduce a BC break for anybody else that extended that context.

Composition is the solution here and it's introduced in this PR.
The `HtmlHeadManager` is the equivalent of the future `JsonLdManager` and it will be responsible for all the other stuff like `<script>`, `<link>` etc.
I have already prepared this in yet another feature branch but I'll only PR this once this one here is done :-)


core-bundle/src/Routing/ResponseContext/HtmlHeadManager/HtmlHeadManager.php
```diff
@@ -0,0 +1,66 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Routing\ResponseContext\HtmlHeadManager;
```
aschempp, line 13: why is this in a subnamespace (but the interface is not)?

core-bundle/src/Routing/ResponseContext/ContaoWebpageResponseContext.php
```diff
@@ -12,19 +12,58 @@
...
+            ;
         }
     }
+
+    public function getHtmlHeadManager(): HtmlHeadManager
+    {
+        return $this->inner->getHtmlHeadManager();
+    }
+
+    public function getHeaderBag(): ResponseHeaderBag
```
aschempp, line 58: Looks like this should be `getResponseHeaderBag` to be clearer

aschempp (commented):
I really **really** REALLY don't like all that `Manager` and `Providing` nonsense... can't we come up with something better? If we don't, I would prefer `HtmlHeadInterface` in the `ResponseContext` namespace …

Also, be aware that I don't think adding scripts and stylesheets to that head manager would be correct, at least not necessarily. They both have a lot more functionality, like being combined and being appended to the body.

I'm also not sure what the `HtmlHeadManager` is really for? I mean it's not like you could reuse the `HtmlHeadManager` in any other response context than an HTML response. And any HTML response should be based on the same (base) context, shouldn't it?

m-vo:
Not a big fan of the naming either tbh. But nothing that can't be solved. :relaxed: 

What I'm wondering is: There is a `ResponseContext(Interface)`, a `WebpageResponseContent` and a `ContaoResponseContent` - which of these things should embrace composition? I feel like implementing a certain context (e.g. the webpage one) is already composing bits and pieces and the extra level of `$inner` nesting could maybe be omitted. I understand that it's there for typesafety against interfaces...

Just brainstorming, what if there is a generic `ResponseContext` (no interface) that allows registering arbitrary 'attribute' classes (ignore the naming for now) and you could interact with it like this?
```php
if($responseContext->supports(HtmlHeadAttributes::class) {
  // get registered instance that implements the class/interface, @returns class-string($type)
  $responseContext->get(HtmlHeadAttributes::class)->setTitle('foo');
}
```

Toflar:
Closing in favor of https://github.com/contao/contao/pull/3066

PR #3042, opened 2021-05-31
Handle non-string arguments in the `FilesModel::findByPath()` method

### LINKED ISSUE #3001: Latest change to FilesModel may lead to exception
**Affected version(s)**

Contao 4.9.14

**Description**

The changes to FilesModel in 4.9.14 lead to different outcome on method calls (only tested for findByPath). Until 4.9.13 passing an non-string value lead to method returning null, now an exception (InvalidArgumentException of webmozart Assert) is thrown. This breaks code which trust in this behaivior (which is same for most (all?) findBy-Model-Methods.

This is the change: https://github.com/contao/contao/commit/c9b9ba7a8b83885683476a961968dd935252874b#

This is the line leading to an error now: https://github.com/contao/contao/blob/7ee9472cb592cc0de678d3b16af53fb763c64afc/core-bundle/src/Resources/contao/models/FilesModel.php#L260

A simple try-catch should fix the problem.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3001
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/models/FilesModel.php
+++ b/core-bundle/src/Resources/contao/models/FilesModel.php
@@ -254,6 +254,13 @@ class FilesModel extends Model
 	 */
 	public static function findByPath($path, array $arrOptions=array())
 	{
+		if (!\is_string($path))
+		{
+			trigger_deprecation('contao/core-bundle', '4.9', 'Using Contao\FilesModel::findByPath() with a non-string argument has been deprecated and will no longer work in Contao 5.0.');
+
+			return null;
+		}
+
 		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
 		$uploadPath = System::getContainer()->getParameter('contao.upload_path');
 
--- a/core-bundle/src/Resources/contao/models/FilesModel.php
+++ b/core-bundle/src/Resources/contao/models/FilesModel.php
@@ -256,8 +256,6 @@ class FilesModel extends Model
 	{
 		if (!\is_string($path))
 		{
-			trigger_deprecation('contao/core-bundle', '4.9', 'Using Contao\FilesModel::findByPath() with a non-string argument has been deprecated and will no longer work in Contao 5.0.');
-
 			return null;
 		}
 


core-bundle/src/Resources/contao/models/FilesModel.php
```diff
@@ -254,6 +254,13 @@ public static function findMultipleByUuids($arrUuids, array $arrOptions=array())
 	 */
 	public static function findByPath($path, array $arrOptions=array())
 	{
+		if (!\is_string($path))
+		{
+			trigger_deprecation('contao/core-bundle', '4.9', 'Using Contao\FilesModel::findByPath() with a non-string argument has been deprecated and will no longer work in Contao 5.0.');
```
fritzmg: Do we really need to trigger a deprecation, even though the docblock already says that it should be a string?

Toflar: Personally, I would have closed the other issue without doing anything as well. The phpdoc has always been correct. You were never supposed to pass anything else - that has always been a bug and even if it worked fine previously, that was never the idea. I don't like this deprecation, tbh.

ausi: I think we allow `null` to be passed in more places even if the docblock says `string`. So calling `findByPath($path)` without doing a null check on `$path` first could be quite common.

Toflar: The method says `findByPath()`. It never makes sense to call `findByPath(null)`. It's absolute nonsense. And the docblock states `@param string`. I don't care if it worked with `null` by chance before. It's something we could've simply ignored.

m-vo: Are we going to replace our docbloc typehints with real typehints at any point in time? Or do we rather deprecate the whole structures?

> The phpdoc has always been correct.

I think I tried to raise Psalm/PHPStan levels at one time to the point where docbloc typehints were considered "the truth" which produced lots of problems. We're not that consistent in the old codebase and I fear the same is true for the usage of it.

That said, I still think we should either "fix" the type hint (`string|null` in this case) and not trigger a deprecation or do nothing at all.

m-vo: > It never makes sense to call findByPath(null)

Actually, I agree. :smile: But I wouldn't bet the core does not do this as well.

leofeyer: As discussed in Mumble on June 2nd, we want to remove the deprecation.

leofeyer: Removed in 320efada356aa86898c3a4984f2e85cdbec3394d.

Toflar (approved):

PR #3043, opened 2021-05-31
Rename the close button at the bottom of the modal window

### LINKED ISSUE #2646: modalIFrame: OK button not displayed
**Affected version(s)**

Contao 4.9

**Description**
1) File manager
2) open info box
3) modalIFrame does not display the OK-Button (see 4)
grafik
(screen shot from demo.contao.org)

Does it have any reason? Or is it a bug? Is it possible to display this button via parameter like title, width and height?

```
onclick="Backend.openModalIframe({'title':'content-images','url':this.href});return false"
```

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2646
| Docs PR or issue | -

@contao/developers Are we sure that the double close button provides a better UX?


aschempp:
To me, only modals with an action should have buttons. The info modal does not have an action, it should be closed by clicking the cross or the background overlay (like anywhere else on _the internet®_ ).

leofeyer:
As discussed in Mumble on June 2nd, we do not want to add a close button and instead rename the existing "Close" button to "Cancel".

leofeyer:
Changed in b6904ab69b404c9de0e2ca9c920ea715dab8060c.

core-bundle/src/Resources/contao/classes/BackendTemplate.php
```diff
@@ -180,7 +180,7 @@ protected function getLocaleString()
 			'var Contao={'
 				. 'theme:"' . Backend::getTheme() . '",'
 				. 'lang:{'
-					. 'close:"' . $GLOBALS['TL_LANG']['MSC']['close'] . '",'
```
Toflar, line 183: I would keep both here. I'm sure someone else is using the close label.

Toflar (changes requested):

core-bundle/src/Resources/contao/classes/BackendTemplate.php
leofeyer: I am not a big fan of sending data that is nowhere used, but for the sake of BC, I have re-added the label in 9296b6640ade7e23768e6f9008ff67123dfeb42d.

Toflar (approved):

Issue #2490, opened 2020-11-10
Hook similar to isVisibleElement but for pages


bennyborn:
From time to time we have projects where it's not sufficient to create a set of member groups which have access to specific pages. For example: Some pages might only be accessed if the member activated a specific option in his personal data or bought a specific product in an online shop.

Setting up something like this currently requires a modification to the core-bundle because while we're already able to prevent access to a specific page via `generatePage` hook we have no method to prevent the page from showing up in any navigation module.

Because of this we'd like to propose a new hook `isVisiblePage` which we currently add manually into the into `renderNavigation` method of `modules/Module.php` like this:

```php
    // Browse subpages
    foreach ($objSubpages as $objSubpage)
    {

        $blnIsVisible = true;

        // HOOK: add custom logic
        if (isset($GLOBALS['TL_HOOKS']['isVisiblePage']) && is_array($GLOBALS['TL_HOOKS']['isVisiblePage']))
        {
            foreach ($GLOBALS['TL_HOOKS']['isVisiblePage'] as $callback)
            {
                $blnIsVisible = static::importStatic($callback[0])->{$callback[1]}($objSubpage,$blnIsVisible);
            }
        }

        // Skip hidden sitemap pages
        if (!$blnIsVisible || ($this instanceof \ModuleSitemap && $objSubpage->sitemap == 'map_never'))
        {
            continue;
        }
```

Toflar:
If you're going to work on this, I think the better option would be to convert it to a knp menu where we have all the events and abstraction at hand to add and/or remove items dynamically. See https://github.com/contao/contao/pull/860 and https://github.com/contao/contao/pull/1101 for reference how that happened for the back end navigation. It's e.g. used to dynamically add the `Contao Manager` link if it's present etc.

bennyborn:
@Toflar Judging by the amount of files / code in Leo's commits it seems like a task that is better suited for someone being more familiar with KNP and Symfony itself. Nevertheless I'd be happy to provide a simple PR that implements the proposed hook.

aschempp:
It should be fairly simple to hide the page in a navigation template, have you tried that?

bennyborn:
> It should be fairly simple to hide the page in a navigation template, have you tried that?

Sure but that's not a very pleasant solution and would require a bit more customization on the template side to create a "clean" markup (regarding CSS classes and such). One could also do the same for content elements instead of using the `isVisibleElement` hook.

[...]

PR #3046, opened 2021-06-01
Add the "findCalendarBoundaries" hook

### LINKED ISSUE #1818: Calendar-Module ignores event coming from the get getAllEvents-Hook for boundaries calculation

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1818
| Docs PR or issue | contao/docs#734

@michb Does this solve your issue?


bennyborn:
I guess that would be a nice solution but I thought there will be no new hooks anymore? (https://github.com/contao/contao/issues/2490#issuecomment-727641705)

michb:
Haven't tested it, but it's looking fine and would solve my issue.

Toflar (approved):

ausi (approved):

fritzmg (approved):

PR #3047, opened 2021-06-01
Update the YouTube options

### LINKED ISSUE #3012: Adjust YouTube-Player parameters
**Affected version(s)**
Contao 4.9+

**Description**
I have checked all the options of the YouTube-player and found out the following:

- **Autoplay**: only works under certain conditions (browser autoplay policy) maybe there should be a hint (https://developers.google.com/web/updates/2017/09/autoplay-policy-changes)
- **Hide the YouTube logo**: Only prevents the YouTube logo from displaying in the control bar, a small YouTube text label will still display in the corner of a paused video
- **Hide the top bar in the preview**: not working anymore, parameter _showinfo_ is deprecated since 2018 and was removed

Source: https://developers.google.com/youtube/player_parameters

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3012
| Docs PR or issue | -

@ausi Will autoplay work now with the added `allow` attribute or do we still have to add a hint? And if yes, how would the hint look like?


ausi:
We should still add a hint.

Something like this? 
*Autoplay of non-muted videos only works under certain conditions in most browsers. Sometimes a user interaction might be required in order to play the video.*

fritzmg:
If we add such a hint, then it should also be shown for Vimeo and regular videos.

It also affects muted videos (e.g. in battery saver mode on iOS).

core-bundle/src/Resources/contao/templates/elements/ce_youtube.html5
```diff
@@ -17,12 +17,13 @@
...
             iframe.width = '<?= $this->width ?>';
             iframe.height = '<?= $this->height ?>';
+            iframe.setAttribute('allow', 'autoplay; encrypted-media; picture-in-picture; fullscreen');
             iframe.setAttribute('allowfullscreen', '');
             this.parentNode.replaceChild(iframe, this);
           });
         </script>
       <?php else: ?>
-        <iframe<?= $this->size ?> src="<?= $this->src ?>" allowfullscreen></iframe>
+        <iframe<?= $this->size ?> src="<?= $this->src ?>" allow="autoplay; encrypted-media; picture-in-picture; fullscreen" allowfullscreen></iframe>
```
leofeyer, line 26: Is this change even right? Because according to https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#new-behaviors, _autoplay with sound will only work if the user has interacted with the domain (click, tap, etc.)_. This is not the case in the `else` condition where there is no splash image.

ausi: IMO yes. Other browsers might act differently and AFAIK you can allow specific domains to always autoplay videos for certain domains.

leofeyer:
I will merge this PR now so we have the fixes in the next release. Let's discuss and add the hint(s) in a separate PR.

ausi (approved):

ausi:
Thank you @leofeyer.

PR #3049, opened 2021-06-01
Get rid of more unit test deprecations

leofeyer:


core-bundle/src/EventListener/TwoFactorFrontendListener.php
```diff
@@ -121,7 +121,12 @@ public function __invoke(RequestEvent $event): void
             return;
         }
 
-        $targetPath = $this->getTargetPath($request->getSession(), $token->getProviderKey());
+        // Backwards compatibility with scheb/2fa-bundle <5.8
```
Toflar: What's the reason not to simply increase the dependency? Why do we want to remain compatible with a lower version? I understand the check for Symfony <5.2 but for this bundle?

leofeyer: Changed in 765ed583c2fdfcd8f0d6a28b4c362032bf937cd4.

Toflar (approved):

PR #3052, opened 2021-06-02
Remove the password confirmation field

### LINKED ISSUE #3040: Remove the password confirmation field
See #2972

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Implements #3040
| Docs PR or issue | -

This PR adds a show/hide password toggle to every password field that previously had a confirmation field. We still have to discuss whether we want to add a toggle to the login screen, however, we will discuss this in a separate issue.


core-bundle/src/Resources/contao/templates/forms/form_password.html5
```diff
@@ -19,19 +19,3 @@
 
   <input type="password" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="text password<?php if ($this->class): ?> <?= $this->class ?><?php endif; ?>" value="" autocomplete="new-password"<?= $this->getAttributes() ?>>
 <?php $this->endblock(); ?>
-
-<?php $this->block('label2'); ?>
```
aschempp, line 23: I think this template should not extend `form_row_double` anymore (topmost line, but I can't comment there)

leofeyer: Removed in a95d1df23874b52fdc95fa9703e43dac010ea5a8

leofeyer:
So far, this is working nice and neat and I was able to add the new show/hide toggle to every field in the back end where you set a new password:







However, there are some issues now:

1. If we toggle the password in the back end, it will show the placeholder `*****`. I assume that this is OK?
2. What about the login screen and other places where you have to enter your password? Do we want a toggle there as well?

aschempp:
I would place the toggle field inside the input element, not beside it. It's related to the content, not a separate wizard.
Bildschirmfoto 2021-06-10 um 17 45 14

leofeyer:
The picker wizard is no longer shown inline in Contao 4.12, either:



I think it is more consistent this way. Otherwise we would have to discuss which wizards are "related to the content" and which are not. 😉

ausi (approved):

bytehead (approved):

PR #3053, opened 2021-06-04
Add a page controller for root pages

aschempp:
This is the first PR to replace one of our page types with a page controller. It superseeds https://github.com/contao/contao/pull/2015 and should be much easier to review and get started 😊


core-bundle/src/Resources/contao/pages/PageRoot.php
```diff
@@ -51,7 +54,7 @@ public function generate($rootPageId, $blnReturn=false, $blnPreferAlias=false)
 	 */
 	public function getResponse($rootPageId)
 	{
-		return new RedirectResponse($this->getRedirectUrl($rootPageId));
+		return new RedirectResponse($this->getRedirectUrl($rootPageId), Response::HTTP_SEE_OTHER);
```
Toflar: Are you sure this is the correct code? I remember we had a long discussion about this and changed it at some point. Cannot find the relevant issue anymore though.

aschempp: That's a very good question, and I think it depends 😂 I've actually changed this in 352b10b because the functional tests expect a 303 response. However, some places (e.g. `PageRoot::getResponse()` would have returned a 302, but I think that method was never used …

leofeyer: We have come to the conclusion that it does not matter whether it is 302 or 303, therefore the legacy code uses the one status code and the Symfony based code uses the other.

aschempp: Then I would suggest to stick to the default Symfony code (302) instead of explicitly defining it?

leofeyer: Yes. 👍🏻

aschempp: Changed in 38324ccc195a25a0d4fe3654512d66cf2d37b2c6

core-bundle/src/Resources/config/controller.yml
```diff
@@ -0,0 +1,12 @@
...
+        autoconfigure: true
+
+    _instanceof:
+        Symfony\Bundle\FrameworkBundle\Controller\AbstractController:
+            calls:
+                - [setContainer, ['@Psr\Container\ContainerInterface']]
+
+    Contao\CoreBundle\Controller\Page\RootPageController:
+        tags:
+            - { name: contao.page, contentComposition: false }
```
fritzmg: We are using Annotations everywhere else now. Why explicitly tag it here? For consistency I would also use annotations for the page controllers and change this to `Contao\CoreBundle\Controller\Page\RootPageController: ~`

core-bundle/src/Resources/contao/pages/PageRoot.php
fritzmg: But this line still explicitly defines it. Shouldn't it be left as it was, as it is also unrelated to the PR?

core-bundle/src/Resources/config/controller.yml
aschempp: good question. I'm not sure we have a consistency here, we just merged https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/config/services.yml#L111-L113 for example. But we seem to use it for event listeners (hooks & callbacks) a lot …

fritzmg: Ah yeah, that went under my radar too.

aschempp: Changed in 5dd58a27a0e60ea8a1c26b5811bb9a3672390fbf
I'll do a follow-up PR once this is merged to clean it up everywhere

core-bundle/src/Resources/contao/pages/PageRoot.php
leofeyer: It sure should.

fritzmg (dismissed):

core-bundle/src/Resources/contao/pages/PageRoot.php
leofeyer: Reverted in 4963145ed257d7f9b588096b9868cf7cf672091e.

leofeyer:
The tests are still failing, because the functional tests still use the old `PageRoot` class instead of the new page controller. @aschempp Can you please fix this?

aschempp:
Tests were failing because service annotations did not work in our test kernel. @leofeyer I've reverted your changes on the status code.

aschempp:
Failing tests seem unrelated (or I don't understand them 🤷 )

PR #3055, opened 2021-06-04
Add an enhanced Markdown content element

### LINKED ISSUE #3037: Feature Request: Change Markdown Library
Change Markdown Library to use some either features like `target:_blank` or `rel="noopener noreferrer"`.

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/3037
| Docs PR or issue | TODO

This will greatly enhance our current Markdown content element by supporting additional features such as

- Full GitHub flavoured Markdown support like tables or task lists
- Supports including file content now too
- Autodetects external links and assigns external CSS classes, `rel="noopener noreferrer"` and `target="_blank"`.

This:

```md
# Headline

| foo | bar |
| --- | --- |
| baz | bim |


[I'm an external link](https://www.contao.org/foobar.html)

[I'm an internal link](https://contao4.wip/foobar.html)
```

Now turns into this:

```html
<h1>Headline</h1>
<table>
   <thead>
      <tr>
         <th>foo</th>
         <th>bar</th>
      </tr>
   </thead>
   <tbody>
      <tr>
         <td>baz</td>
         <td>bim</td>
      </tr>
   </tbody>
</table>
<p><a rel="noopener noreferrer" target="_blank" class="external-link" href="https://www.contao.org/foobar.html">I'm an external link</a></p>
<p><a href="https://contao4.wip/foobar.html">I'm an internal link</a></p>
```

Tbd:

Should we keep the `strip_tags($this->code, Config::get('allowedTags'));`? I understand the idea but I'm not sure if it makes sense. We do not strip any tags in the HTML content element either. If I give access to this element to my users I should know they can use any HTML. That's how both, the HTML element as well as Markdown work 🤷‍♂️


asaage:
.md Files should be added to the defaultlist of uploadable and editable filetypes.

asaage:
For convenience: - maybe add .txt files to the eval extensions? (changing the fileextension in BE is not possible)

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function __construct(MarkdownConverterInterface $converter)
+    {
+        $this->converter = $converter;
+    }
+
+    public static function createController(RequestStack $requestStack): self
+    {
+        $environment = Environment::createCommonMarkEnvironment();
+        $environment->addExtension(new GithubFlavoredMarkdownExtension()); // Support GitHub flavoured Markdown
```
m-vo: There's a shortcut for this :smiley: 

```suggestion
        $environment = Environment::createGFMEnvironment(); // Support GitHub flavoured Markdown
```

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,80 @@
...
+
+        return new self($converter);
+    }
+
+    protected function getResponse(Template $template, ContentModel $model, Request $request): ?Response
+    {
+        $markdown = '';
+
+        if ('sourceFile' === $model->markdownSource) {
+            /** @var FilesModel $filesModel */
```
m-vo:
```suggestion
            /** @var FilesModel|null $filesModel */
```

m-vo:
> If I give access to this element to my users I should know they can use any HTML.

The GFM extension already brings the [Disallowed Raw HTML Extension](https://commonmark.thephpleague.com/1.6/extensions/disallowed-raw-html/) which filters certain HTML tags. 

IIRC there are also `html_input` options that can be configured on an environment level like `HTML_INPUT_ESCAPE` and `HTML_INPUT_STRIP` (default: `HTML_INPUT_ALLOW`) that decide what should happen to HTML in the input.

ausi:
> We do not strip any tags in the HTML content element either.

We encode all not-allowed tags in the `Input` class, the content `<div><script>` gets converted to `<div>&lt;script>` for example.

> That's how both, the HTML element as well as Markdown work 🤷‍♂️

Currently the HTML element encodes and the markdown element strips all not-allowed tags (after conversion). I think we should keep it that way.

Toflar:
The `html_input` filter does not work for us. It's disabled (= allows everything) by default to support the full spec which is okay for us. I've implemented stripping according to the allowed tags now.
`md` files are allowed to be uploaded and downloaded by default now too.
I don't think allowing to select `txt` files makes sense, so I did not add this.

m-vo:
The following tags are stripped nevertheless by The `DisallowedRawHtml*Renderer`:
```
title|textarea|style|xmp|iframe|noembed|noframes|script|plaintext
```

Not sure if this matters that much but this would for instance disallow using iframes or js.

Toflar:
Yeah, we could easily allow those, if we want that. We probably should for BC reasons, wdyt?

leofeyer:
Is it safe to allow them (especially `iframe` and `script`)?

Toflar:
They are allowed now with our existing markdown parser unless disallowed in our configuration (which is activated by default).

asaage:
Ok then no ```txt```-files...
I just wonder: where did ```$GLOBALS['TL_CONFIG']['editableFiles']  ``` go?

Toflar:
Bundle configuration - where all the settings are slowly migrated to until there's no more back end system configuration.

Toflar:
Updated the extension config. This is ready imho.

core-bundle/src/Resources/config/services.yml
```diff
@@ -108,6 +108,14 @@ services:
         #    message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "Contao\CoreBundle\Controller\BackendCsvImportController" instead.
         public: true
 
+    Contao\CoreBundle\Controller\ContentElement\MarkdownController:
+        factory: ['Contao\CoreBundle\Controller\ContentElement\MarkdownController', 'createController']
+        arguments:
+            - '@request_stack'
+            - '%kernel.project'
```
ausi: This parameter is not needed I think?

Toflar: Yes, just removed that in my latest CS commit :)

ausi (dismissed):

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,93 @@
...
+            if (null !== $filesModel) {
+                $markdown = (string) $filesModel->getContent();
+            }
+        } else {
+            $markdown = $model->code;
+        }
+
+        $html = $this->converter->convertToHtml($markdown);
+
+        return new Response(strip_tags($html, $this->get('contao.framework')->getAdapter(Config::class)->get('allowedTags')));
```
m-vo: Just thinking: If we wanted to open up the environment in the future (e.g. by using a factory or dispatching an event that allows adding your own extensions) it would probably be better to use a renderer to check for allowed tags. Then the output of other inline elements/blocks could very well safely use any tag but a user still could not inject them.

m-vo (dismissed):

core-bundle/src/Resources/contao/models/FilesModel.php
```diff
@@ -96,6 +96,31 @@ class FilesModel extends Model
...
+	public function getContent(): ?string
+	{
+		$path = $this->getAbsolutePath();
+
+		if (!file_exists($path))
+		{
+			return null;
+		}
+
+		return (string) file_get_contents($path);
```
leofeyer: I do not like this. The model should not be responsible to read the file. I would prefer to have the controller do this. WDYT?

leofeyer (changes requested):

core-bundle/src/Controller/ContentElement/MarkdownController.php
Toflar: Yeah, that's a new feature though. We haven't had that up until now and I don't need that so feel free, if you need it :)

leofeyer:
Also, there is this issue in my installation:

core-bundle/src/Resources/contao/models/FilesModel.php
Toflar: I put it here because I was fed up by writing the same code over and over again. It deserves its own shortcut.

Toflar:
How can I reproduce the error? It runs just fine on my system. Maybe that's an unrelated PHP 8 issue?

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,93 @@
...
+        $markdown = '';
+
+        if ('sourceFile' === $model->markdownSource) {
+            /** @var FilesModel|null $filesModel */
+            $filesModel = $this->get('contao.framework')->getAdapter(FilesModel::class)->findByPk($model->singleSRC);
+
+            if (null !== $filesModel) {
+                $markdown = (string) $filesModel->getContent();
+            }
+        } else {
```
leofeyer:
```suggestion
        } elseif ('sourceText' === $model->markdownSource && $model->code !== '') {
```

Otherwise we will end up with an exception if someone creates a new markdown element and does not add any code.

core-bundle/src/Resources/contao/languages/en/tl_content.xlf
```diff
@@ -218,6 +218,18 @@
       <trans-unit id="tl_content.code.1">
         <source>Note that the code will not be executed. Use F11 to toggle the fullscreen mode if you are using the code editor.</source>
       </trans-unit>
+      <trans-unit id="tl_content.markdownSource.0">
+        <source>Markdown source</source>
+      </trans-unit>
+      <trans-unit id="tl_content.markdownSource.1">
+        <source>Please chose the source of the Markdown content.</source>
```
leofeyer:
```suggestion
        <source>Please choose the Markdown source.</source>
```

leofeyer (changes requested):

core-bundle/src/Resources/contao/models/FilesModel.php
leofeyer: I disagree. Not the model's responsibility. @contao/developers WDYT?

leofeyer:
> How can I reproduce the error? It runs just fine on my system. Maybe that's an unrelated PHP 8 issue?

It occurs here:

https://github.com/contao/contao/blob/5fa134166a7b52f2b94bf7d5e384e233bc4a4e2d/core-bundle/src/Resources/contao/library/Contao/Controller.php#L568

The `ContentElementReference` object does not have a `protected` (magic) property.

core-bundle/src/Resources/contao/models/FilesModel.php
m-vo: :thinking:  IMO one could argue that the model is wrapper around the resource. So this way it wouldn't matter if the content comes from the database, is a resource handle that will be read or whatever (implementation detail). I had the same discussion over entities and it boiled down to whether you wanted rich entities (models) or POPOs.

composer.json
```diff
@@ -59,10 +59,10 @@
         "knplabs/knp-menu-bundle": "^2.1 || ^3.0",
         "knplabs/knp-time-bundle": "^1.10",
         "lcobucci/jwt": "^3.4 || ^4.0",
-        "league/commonmark": "^1.6",
         "league/uri-components": "^1.8",
         "league/uri-schemes": "^1.2",
         "matthiasmullie/minify": "^1.3",
+        "michelf/php-markdown": "^1.4",
```
leofeyer: Why did you revert this change? Now the old lib is installed again.

Toflar:
> The `ContentElementReference` object does not have a `protected` (magic) property.

I did not change anything there so I guess that's an unrelated bug.

leofeyer:
@aschempp /cc

core-bundle/src/Resources/contao/models/FilesModel.php
aschempp: I don't really think this should be part of the model either, also because it's basically a one-line?

Toflar: But I can leave the absolute path there? That's the one I care about more because I don't want to inject the project_dir everywhere

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,93 @@
...
+     * @var MarkdownConverterInterface
+     */
+    private $converter;
+
+    public function __construct(MarkdownConverterInterface $converter)
+    {
+        $this->converter = $converter;
+    }
+
+    public static function createController(RequestStack $requestStack): self
```
aschempp: This factory method looks wrong to me. I can imagine a lot of these options being runtime configuration for the content element (e.g. whether to generate `target=_blank`), I feel like this should go into the response method (or be a private/protected method that's being used at runtime).

core-bundle/src/Controller/ContentElement/MarkdownController.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->initializeContaoFramework();
+        $markdown = '';
+
+        if ('sourceFile' === $model->markdownSource) {
+            /** @var FilesModel|null $filesModel */
+            $filesModel = $this->get('contao.framework')->getAdapter(FilesModel::class)->findByPk($model->singleSRC);
+
+            if (null !== $filesModel) {
+                $markdown = (string) $filesModel->getContent();
+            }
```
aschempp: We can safely return an empty response here without calling the converter.

```suggestion
            if (null === $filesModel) {
                return new Response('');
            }
            
            $markdown = (string) $filesModel->getContent();
```

core-bundle/src/Controller/ContentElement/MarkdownController.php
aschempp: I think you should place the markdown content in the template and return the template response, otherwise it wouldn't have a `ce_markdown` div etc.

core-bundle/src/Resources/config/services.yml
```diff
@@ -108,6 +108,13 @@ services:
         #    message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "Contao\CoreBundle\Controller\BackendCsvImportController" instead.
         public: true
 
+    Contao\CoreBundle\Controller\ContentElement\MarkdownController:
+        factory: ['Contao\CoreBundle\Controller\ContentElement\MarkdownController', 'createController']
+        arguments:
+            - '@request_stack'
```
aschempp: This will make the controller/service be factored according to the first request that needs it, I don't think that's correct (also see my comment about the factory method).

core-bundle/src/Resources/contao/elements/ContentMarkdown.php
```diff
@@ -19,6 +19,13 @@
  */
 class ContentMarkdown extends ContentElement
 {
+	public function __construct($objElement, $strColumn = 'main')
+	{
+		trigger_deprecation('contao/core-bundle', '4.12', 'ContentMarkdown has been deprecated and will be removed in Contao 5.0.');
```
aschempp: We usually just place this inside the class file, as soon as the class is loaded the deprecation is triggered. No need for the constructor method.

aschempp (changes requested):
Where do we require the commonmark library?

core-bundle/src/Controller/ContentElement/MarkdownController.php
Toflar: That would mean we would create the instance of the MarkdownConverter on the fly rather than injecting it as a dependency. So in the `getResponse()` we'd have `new MarkdownConverter()` which felt a bit wrong to me.

Toflar:
Completely reworked the implementation, I guess I addressed all comments :)

fritzmg (dismissed):

leofeyer (changes requested):
The implementation looks good, however, the content element does not work neither in the back end nor in the front end.

### Back end error



### Front end error

Toflar:
Then please debug because I can still not reproduce this issue. It works perfetly fine here.

leofeyer:
It seems that my demo DB was causing the issue. The element works now, however, the unit tests are missing and therefore the `codecov/project` job fails.

Toflar:
I don't see how they would add value here.

m-vo:
What about an end to end test to ensure the config is correct? (markdown in, check against expected html)

Toflar:
Added tests in https://github.com/contao/contao/pull/3055/commits/77409ae699c01e2cb2f5af4a05d58f6f929a4f3f. You can call me king of mocks now.

leofeyer:
Thank you @Toflar.

PR #3056, opened 2021-06-05
Deprecate already moved config options

### LINKED ISSUE #3050: Deprecate/remove characterSet config option
As discussed in https://github.com/contao/contao/pull/3014#discussion_r643119966 the characterEncoding should always be `utf-8`.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3050
| Docs PR or issue | contao/docs#769

Working on #3050 I noticed that we don’t yet trigger deprecations for `Contao::get()` and `set()` calls to configs we moved to the container.

I added these deprecations in this PR and mapped `characterSet` to `kernel.charset`.

I also added a warning if someone uses a custom kernel with the `Kernel::getCharset()` method overwritten and returning something different than `"UTF-8"`.

And moved the configuration of the DC_Folder uploadPath and editableFiles to the DCA config, best seen in the [*tl_templates.php* diff](https://github.com/contao/contao/pull/3056/files#diff-d38dc1262eca6e265e775fc25926e0d111687e593eb0b03d844f809884ffee53)

### DIFF
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -344,7 +344,7 @@ $GLOBALS['TL_DCA']['tl_calendar_events'] = array
 			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['singleSRC'],
 			'exclude'                 => true,
 			'inputType'               => 'fileTree',
-			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>implode(',', System::getContainer()->getParameter('contao.image.valid_extensions')), 'mandatory'=>true),
+			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>'%contao.image.valid_extensions%', 'mandatory'=>true),
 			'sql'                     => "binary(16) NULL"
 		),
 		'alt' => array
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -344,7 +344,7 @@ $GLOBALS['TL_DCA']['tl_calendar_events'] = array
 			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['singleSRC'],
 			'exclude'                 => true,
 			'inputType'               => 'fileTree',
-			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>Config::get('validImageTypes'), 'mandatory'=>true),
+			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>implode(',', System::getContainer()->getParameter('contao.image.valid_extensions')), 'mandatory'=>true),
 			'sql'                     => "binary(16) NULL"
 		),
 		'alt' => array
--- a/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
+++ b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
@@ -48,6 +48,10 @@ class ContaoCoreExtension extends Extension
 
     public function load(array $configs, ContainerBuilder $container): void
     {
+        if ('UTF-8' !== $container->getParameter('kernel.charset')) {
+            trigger_error(sprintf('Using the charset "%s" is not supported by Contao, use "UTF-8" instead.', $container->getParameter('kernel.charset')), E_USER_WARNING);
+        }
+
         $configuration = new Configuration(
             $container->getParameter('kernel.project_dir'),
             $container->getParameter('kernel.default_locale')
--- a/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
+++ b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
@@ -49,7 +49,7 @@ class ContaoCoreExtension extends Extension
     public function load(array $configs, ContainerBuilder $container): void
     {
         if ('UTF-8' !== $container->getParameter('kernel.charset')) {
-            trigger_error(sprintf('Using the charset "%s" is not supported by Contao, use "UTF-8" instead.', $container->getParameter('kernel.charset')), E_USER_WARNING);
+            trigger_deprecation('contao/core-bundle', '4.12', 'Using the charset "%s" is not supported, use "UTF-8" instead. In Contao 5.0 an exception will be thrown for unsupported charsets.', $container->getParameter('kernel.charset'));
         }
 
         $configuration = new Configuration(
--- a/core-bundle/src/Resources/contao/classes/Frontend.php
+++ b/core-bundle/src/Resources/contao/classes/Frontend.php
@@ -321,9 +321,10 @@ abstract class Frontend extends Controller
 		$host = Environment::get('host');
 		$logger = System::getContainer()->get('monolog.logger.contao');
 		$accept_language = Environment::get('httpAcceptLanguage');
+		$blnAddLanguageToUrl = System::getContainer()->getParameter('contao.prepend_locale');
 
 		// Get the language from the URL if it is not set (see #456)
-		if (!isset($_GET['language']) && $GLOBALS['TL_CONFIG']['addLanguageToUrl'])
+		if (!isset($_GET['language']) && $blnAddLanguageToUrl)
 		{
 			$arrMatches = array();
 
@@ -337,7 +338,7 @@ abstract class Frontend extends Controller
 		}
 
 		// The language is set in the URL
-		if (!empty($_GET['language']) && $GLOBALS['TL_CONFIG']['addLanguageToUrl'])
+		if (!empty($_GET['language']) && $blnAddLanguageToUrl)
 		{
 			$strUri = Environment::get('url') . '/' . Input::get('language') . '/';
 		}
@@ -377,7 +378,7 @@ abstract class Frontend extends Controller
 		// Redirect to the website root or language root (e.g. en/)
 		if (!Environment::get('relativeRequest'))
 		{
-			if ($GLOBALS['TL_CONFIG']['addLanguageToUrl'])
+			if ($blnAddLanguageToUrl)
 			{
 				$arrParams = array('_locale' => $objRootPage->language);
 
--- a/core-bundle/src/Resources/contao/classes/Frontend.php
+++ b/core-bundle/src/Resources/contao/classes/Frontend.php
@@ -323,7 +323,7 @@ abstract class Frontend extends Controller
 		$accept_language = Environment::get('httpAcceptLanguage');
 
 		// Get the language from the URL if it is not set (see #456)
-		if (!isset($_GET['language']) && Config::get('addLanguageToUrl'))
+		if (!isset($_GET['language']) && $GLOBALS['TL_CONFIG']['addLanguageToUrl'])
 		{
 			$arrMatches = array();
 
@@ -337,7 +337,7 @@ abstract class Frontend extends Controller
 		}
 
 		// The language is set in the URL
-		if (!empty($_GET['language']) && Config::get('addLanguageToUrl'))
+		if (!empty($_GET['language']) && $GLOBALS['TL_CONFIG']['addLanguageToUrl'])
 		{
 			$strUri = Environment::get('url') . '/' . Input::get('language') . '/';
 		}
@@ -377,7 +377,7 @@ abstract class Frontend extends Controller
 		// Redirect to the website root or language root (e.g. en/)
 		if (!Environment::get('relativeRequest'))
 		{
-			if (Config::get('addLanguageToUrl'))
+			if ($GLOBALS['TL_CONFIG']['addLanguageToUrl'])
 			{
 				$arrParams = array('_locale' => $objRootPage->language);
 
--- a/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
@@ -129,7 +129,7 @@ class FrontendIndex extends Frontend
 			}
 
 			// Use the first result (see #4872)
-			if (!System::getContainer()->getParameter('contao.legacy_routing') || !$GLOBALS['TL_CONFIG']['addLanguageToUrl'])
+			if (!System::getContainer()->getParameter('contao.legacy_routing') || !System::getContainer()->getParameter('contao.prepend_locale'))
 			{
 				$objNewPage = current($arrLangs);
 			}
--- a/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
@@ -247,7 +247,7 @@ class FrontendIndex extends Frontend
 		}
 
 		// Check wether the language matches the root page language
-		if (isset($_GET['language']) && !$objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
+		if (isset($_GET['language']) && $objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
 		{
 			throw new PageNotFoundException('Page not found: ' . Environment::get('uri'));
 		}
--- a/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
@@ -129,7 +129,7 @@ class FrontendIndex extends Frontend
 			}
 
 			// Use the first result (see #4872)
-			if (!System::getContainer()->getParameter('contao.legacy_routing') || !Config::get('addLanguageToUrl'))
+			if (!System::getContainer()->getParameter('contao.legacy_routing') || !$GLOBALS['TL_CONFIG']['addLanguageToUrl'])
 			{
 				$objNewPage = current($arrLangs);
 			}
@@ -247,7 +247,7 @@ class FrontendIndex extends Frontend
 		}
 
 		// Check wether the language matches the root page language
-		if (isset($_GET['language']) && Config::get('addLanguageToUrl') && Input::get('language') != $objPage->rootLanguage)
+		if (isset($_GET['language']) && !$objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
 		{
 			throw new PageNotFoundException('Page not found: ' . Environment::get('uri'));
 		}
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -501,7 +501,8 @@ class tl_templates extends Backend
 		$objTemplate->title = StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['showDifferences']);
 		$objTemplate->charset = Config::get('characterSet');
 
-		Config::set('debugMode', false);
+		System::getContainer()->setParameter('kernel.debug', false);
+		$GLOBALS['TL_CONFIG']['debugMode'] = false;
 
 		throw new ResponseException($objTemplate->getResponse());
 	}
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -36,6 +36,8 @@ $GLOBALS['TL_DCA']['tl_templates'] = array
 	'config' => array
 	(
 		'dataContainer'               => 'Folder',
+		'uploadPath'                  => 'templates',
+		'editableFileTypes'           => 'html5,twig',
 		'closed'                      => true,
 		'onload_callback' => array
 		(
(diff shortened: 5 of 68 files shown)


ausi:
Depends on https://github.com/contao/test-case/pull/12 ☑️

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
```diff
@@ -344,7 +344,7 @@
 			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['singleSRC'],
 			'exclude'                 => true,
 			'inputType'               => 'fileTree',
-			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>Config::get('validImageTypes'), 'mandatory'=>true),
+			'eval'                    => array('filesOnly'=>true, 'fieldType'=>'radio', 'extensions'=>implode(',', System::getContainer()->getParameter('contao.image.valid_extensions')), 'mandatory'=>true),
```
Toflar: Maybe stupid proposal but because everybody has to do this in their own extensions too. What if we supported parameters in certain options here?

> `'extensions' => '%contao.image.valid_extensions%'

Wouldn't that simplify things a lot. If this file were in YAML and not a PHP file, one couldn't access the container either.

m-vo:
Should we normalize the concatenated paths (`Path::join`) that now contain parameters as well here?

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -48,6 +48,10 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
 
     public function load(array $configs, ContainerBuilder $container): void
     {
+        if ('UTF-8' !== $container->getParameter('kernel.charset')) {
+            trigger_error(sprintf('Using the charset "%s" is not supported by Contao, use "UTF-8" instead.', $container->getParameter('kernel.charset')), E_USER_WARNING);
```
Toflar: Shouldn't we throw an exception here? What am I supposed to do with a warning if nothing works anyway if I don't change it? :D

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
aschempp: Sounds like totally unrelated to this PR? 😬

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
aschempp: Shouldn't we check this only when booting the Contao framework?

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
Toflar: Well, @ausi had to use the container several times here whereas he could've replaced it with a parameter.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
ausi: I’m not sure. I think we should probably trigger a deprecation instead and throw an exception in Contao 5.

core-bundle/tests/Functional/RoutingTest.php
```diff
@@ -1279,7 +1278,7 @@ public function testCorrectPageForUnknownLanguage(): void
         $this->expectDeprecation('Since contao/core-bundle 4.10: Using the "Contao\CoreBundle\Routing\FrontendLoader" class has been deprecated %s.');
 
         Config::set('folderUrl', true);
```
leofeyer, line 1307: I think you have missed this one (also see below).

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
leofeyer: >  if nothing works anyway if I don't change it?

Is this really the case? Because that would be a BC break.

Toflar: @ausi tried it in https://github.com/contao/contao/issues/3050#issuecomment-854714950 - Contao must have never supported anything other than utf-8 for years.

leofeyer: Then I guess it does not matter whether we throw an exception or trigger a deprecation warning. 😄

core-bundle/tests/Functional/RoutingTest.php
ausi: `folderUrl` is not yet deprecated if `legacy_routing` is active I think. If we want to deprecate `legacy_routing` and the accociated config options I can create a seperate PR.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
ausi: Done in 8b209874030e309e9d3eb70be60c9f3858c6ad85.

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
ausi: Like so 60f4e9be2dc45caa5c033349500c3167e904ed62 ?
Which other options do we want to support?

ausi:
> Should we normalize the concatenated paths (`Path::join`) that now contain parameters as well here?

Not necessary I think because the `upload_path` is usually just a single directory like `"files"`.

ausi:
All comments addressed I think, the only remaining question is if we want to support container parameters for more DCA options, https://github.com/contao/contao/pull/3056#discussion_r648264314

Toflar (dismissed):
I think we should but it's not strictly related to this PR. This PR just shows that we'll need it for YAML DCAs or so.

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
leofeyer: This is a very good idea! I like it. 👍🏻

core-bundle/tests/Functional/RoutingTest.php
leofeyer: You are right, the option is still used. My bad.

calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
Toflar: Yeah, that's exactly what I wanted :) Just move it to it's own `protected` method so that any widget can define which properties should be supported for that :)

leofeyer:
> if we want to support container parameters for more DCA options

I guess we cannot abstract this in a way that parameters are supported everywhere automatically?

m-vo:
Well, we could add an event listener that walks through `$GLOBALS['TL_DCA']` and replaces parameters everywhere, but I fear this would potentially introduce unwanted behavior.

leofeyer:
Maybe if we limited parameter replacing to the `eval` section?

m-vo:
Right, if we keep values with no substitutes untouched and limit it to the `eval` section this should probably be fine. Maybe we should allow escaping though?

ausi:
> Maybe if we limited parameter replacing to the `eval` section?

Done in 4b806932ef31827560a39453762a0a8c4900669c

> Maybe we should allow escaping though?

Done in 4b806932ef31827560a39453762a0a8c4900669c

core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -1243,6 +1243,29 @@ public static function getAttributesFromDca($arrData, $strName, $varValue=null,
...
+			{
+				if (!\is_string($varAttrValue) || !preg_match('/%[a-z][a-z0-9_]*\.[a-z0-9_.]+%/i', $varAttrValue, $arrMatches))
+				{
+					continue;
+				}
+
+				$varAttrValue = $objParameterBag->resolveValue($varAttrValue);
+				$varAttrValue = $objParameterBag->unescapeValue($varAttrValue);
+
+				if ($strAttrKey === 'extensions' && \is_array($varAttrValue))
```
leofeyer: Any idea how we could abstract this, too? Maybe we can adjust the `__set()` method to handle both an array and a comma-separated list for the `extension` key?

https://github.com/contao/contao/blob/39e3215c4886350fb65137cff9dc422dbe7de67b/core-bundle/src/Resources/contao/library/Contao/Widget.php#L218-L224

ausi: How about this 22847d5d3d7875b766fd2120e869b10b5934ad59 ?

core-bundle/src/Resources/contao/library/Contao/Config.php
```diff
@@ -77,6 +77,34 @@ class Config
...
+		'urlSuffix'        => 'contao.url_suffix',
+		'uploadPath'       => 'contao.upload_path',
+		'editableFiles'    => 'contao.editable_files',
+		'debugMode'        => 'kernel.debug',
+		'characterSet'     => 'kernel.charset',
+		'enableSearch'     => 'contao.search.default_indexer.enable',
+		'indexProtected'   => 'contao.search.index_protected',
+	);
+
+	private static $arrDeprecated = array
```
leofeyer, line 102: Why did you not add these two to `$arrDeprecatedMap`?

core-bundle/src/Resources/contao/library/Contao/Dbafs.php
```diff
@@ -830,11 +830,6 @@ public static function shouldBeSynchronized($strPath)
 	 */
 	protected static function isFileSyncExclude($strPath)
 	{
-		if (Config::get('uploadPath') == 'templates')
-		{
-			return true;
-		}
-
```
leofeyer, line 837: Is it safe to remove this?

core-bundle/src/Resources/contao/library/Contao/Config.php
ausi: Because `$arrDeprecatedMap` ist used on line 513 and is a direct mapping (without modifications).

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -233,7 +233,7 @@ public static function decodeEntities($strString, $strQuoteStyle=ENT_COMPAT, $st
 
 		if ($strCharset === null)
 		{
-			$strCharset = Config::get('characterSet');
+			$strCharset = $GLOBALS['TL_CONFIG']['characterSet'] ?? 'UTF-8';
```
leofeyer: No `System::getContainer()->getParameter('kernel.charset')` here?

core-bundle/src/Resources/contao/library/Contao/Dbafs.php
ausi: Yes, because the method now uses *contao.upload_path* which is never set to `"templates"` and therefore now works correctly even if `Config::get('uploadPath') == 'templates'` is true.

core-bundle/src/Resources/contao/classes/Frontend.php
```diff
@@ -323,7 +323,7 @@ public static function getRootPageFromUrl()
 		$accept_language = Environment::get('httpAcceptLanguage');
 
 		// Get the language from the URL if it is not set (see #456)
-		if (!isset($_GET['language']) && Config::get('addLanguageToUrl'))
+		if (!isset($_GET['language']) && $GLOBALS['TL_CONFIG']['addLanguageToUrl'])
```
leofeyer: This seems wrong to me. Why are you not using `contao.prepend_locale` here? If we still need the global variable, we should un-deprecate it and continue to use `Config::get()` IMHO.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi: `StringUtil` is so widely used, that `System::getContainer()` would require us to update many many unit-test. And as `$GLOBALS['TL_CONFIG']['characterSet']` is practically always `UTF-8` I thought this is the better compromise.

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -247,7 +247,7 @@ public function renderPage($pageModel)
 		}
 
 		// Check wether the language matches the root page language
-		if (isset($_GET['language']) && Config::get('addLanguageToUrl') && Input::get('language') != $objPage->rootLanguage)
+		if (isset($_GET['language']) && !$objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
```
leofeyer: Are you sure that this is correct? Because `addLanguageToUrl` is the equivalent of `$objPage->urlPrefix` and not `!$objPage->urlPrefix` IMHO.

core-bundle/src/Resources/contao/classes/Frontend.php
ausi: You are right, I think I mixed up `addLanguageToUrl` with legacy routing. Will fix.

core-bundle/src/Resources/contao/dca/tl_templates.php
```diff
@@ -499,9 +502,10 @@ public function compareTemplate(DataContainer $dc)
 		$objTemplate->base = Environment::get('base');
 		$objTemplate->language = $GLOBALS['TL_LANGUAGE'];
 		$objTemplate->title = StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['showDifferences']);
-		$objTemplate->charset = Config::get('characterSet');
+		$objTemplate->charset = System::getContainer()->getParameter('kernel.charset');
 
-		Config::set('debugMode', false);
+		System::getContainer()->setParameter('kernel.debug', false);
+		$GLOBALS['TL_CONFIG']['debugMode'] = false;
```
leofeyer, line 508: Using `$GLOBALS` is worse than using `Config:set()`, isn't it?

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
ausi:
```suggestion
		if (isset($_GET['language']) && $objPage->urlPrefix !== '' && Input::get('language') != $objPage->rootLanguage)
```
This should be correct then?

core-bundle/src/Resources/contao/forms/FormFileUpload.php
```diff
@@ -77,7 +77,7 @@ public function __set($strKey, $varValue)
 			case 'extensions':
 				if ($varValue)
 				{
-					$this->arrAttributes['accept'] = '.' . implode(',.', StringUtil::trimsplit(',', strtolower($varValue)));
+					$this->arrAttributes['accept'] = '.' . implode(',.', StringUtil::trimsplit(',', strtolower(\is_array($varValue) ? implode(',', $varValue) : $varValue)));
```
leofeyer:
```suggestion
					$this->arrAttributes['accept'] = '.' . strtolower(implode(',.', \is_array($varValue) ? $varValue : StringUtil::trimsplit(',', $varValue)));
```

core-bundle/src/Resources/contao/dca/tl_templates.php
ausi: Yes but I wanted to keep it for BC. Our own code uses the `kernel.debug` parameter and doesn’t need the global, but third party code still might use that.

core-bundle/src/Resources/contao/library/Contao/Config.php
leofeyer: And `$GLOBALS['TL_CONFIG']['jpgQuality']` is different from `contao.image.imagine_options.jpeg_quality`?

core-bundle/src/Resources/contao/library/Contao/Dbafs.php
leofeyer: So now the method is executed until line 858 instead of returning early?

core-bundle/src/Resources/contao/classes/Frontend.php
ausi: Fixed in 632af669fd0850f6547e6012d81cdbf5c179780f

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
leofeyer: IMHO we could also just set it to `UTF-8` here.

core-bundle/src/Resources/contao/library/Contao/Config.php
ausi: Yes. `contao.image.imagine_options` is the parameter that is an array. The parameter `contao.image.imagine_options.jpeg_quality` does not exist.

ausi: We could rename to `contao.image.imagine_options[jpeg_quality]` to make the deprecation message more clear.

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
leofeyer: I guess but I haven't tested it. Not sure if the `!== ''` is a good idea, because the value might be `null`.

core-bundle/src/Resources/contao/library/Contao/Dbafs.php
ausi: Yes, and it now works correctly (meaning calling the method on an DBAFS path acually returns `true`).

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
ausi:
```suggestion
		if (isset($_GET['language']) && $objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
```

ausi: Done in aee185e8a07e37885d3edc62deadb5b6a01caa1f

core-bundle/src/Resources/contao/library/Contao/Config.php
leofeyer: Good idea. 👍🏻

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
ausi:
```suggestion
			$strCharset = 'UTF-8';
```

ausi: Done in 1be8e58f2609f30a32cddd2ea5dffdde35cd1cc6

core-bundle/src/Resources/contao/library/Contao/Config.php
ausi: Done in e5ee6ee6880b0ca6ae12c35d9ed8398c7197d9ad

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -247,7 +247,7 @@ public function renderPage($pageModel)
 		}
 
 		// Check wether the language matches the root page language
-		if (isset($_GET['language']) && Config::get('addLanguageToUrl') && Input::get('language') != $objPage->rootLanguage)
+		if (isset($_GET['language']) && $objPage->urlPrefix && Input::get('language') != $objPage->rootLanguage)
```
leofeyer, line 250: I am still not sure about this change in terms of whether it should be `System::getContainer()->getParameter('contao.prepend_locale')` instead of `$objPage->urlPrefix`. The check is only required in legacy routing mode, isn't it? And `$objPage->urlPrefix` is only used in the new routing mode. @aschempp /cc

ausi: But `$objPage->urlPrefix` gets set to the language if legacy routing is active and prepend_locale is set.

leofeyer: I see. Then this is fine. 👍🏻

leofeyer (approved):

core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -1243,6 +1243,24 @@ public static function getAttributesFromDca($arrData, $strName, $varValue=null,
...
+			$objParameterBag = System::getContainer()->getParameterBag();
+
+			foreach ($arrAttributes as $strAttrKey => $varAttrValue)
+			{
+				if (!\is_string($varAttrValue) || !preg_match('/%[a-z][a-z0-9_]*\.[a-z0-9_.]+%/i', $varAttrValue, $arrMatches))
+				{
+					continue;
+				}
+
+				$varAttrValue = $objParameterBag->resolveValue($varAttrValue);
```
Toflar, line 1257: What do you think about adding a `$objParameterBag->has()` check on top of it? Because if you accidentally used `%whatever%` as an `eval` option now, it will resolve to `null` and break your app. With the additional check, this would only happen if this parameter also exists.

ausi: I tried to make the regular expression in line 1252 very strict to limit this risk. `%whatever%` wouldn’t actually match, but `%whatever.whatever%` would.

We would loose the safety for misspelling parameter names if we only resolve existing parameters. I’m not sure.

Toflar: Okay, we could still add it later. Chances are very limited anyway.

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #2305, opened 2020-09-16
Use Locale IDs for tl_page.language

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/1957

Since Contao 4.10, the URL prefix is no longer tied to the page language. With this PR, the page language can be any valid Locale ID. A Locale ID can include language, script, country and additional information (e.g. `de_Latn_CH@currency=EUR`),  see https://github.com/contao/contao/issues/6358.

Locale ID and ICU information can later be used a lot more things like rewriting the number formatting to use the ICU information (decimal point according to language & country).

PR #3057, opened 2021-06-06
Sort routes by region if no preferred language matches

aschempp:
As discussed preparing for https://github.com/contao/contao/pull/2305

### Current situation 1:
- Page languages: `de-CH` and `en-US`
- Preferred languages: `de, en`

**Result:** Language is undefined = ships fallback language because no match of pages and preferred languages

### Current situation 2:
- Page languages: `de-CH` and `en`
- Preferred languages: `de, en`

**Result:** Ships `en` page even though the user prefers german

### Desired situation:

1. The preferred region should be ignored both in the preferred regions as well as the root page settings. If the user accepts `de-CH, en` we should always prefer a german page tree (but still prefer `de-CH` over `de-DE`).

2. If the user prefers `de` and we have `de-CH` and `de-DE`, the sorting should be based on root page sorting. We've discussed relying on "dominant region" but there's no advantage in preferring `en-US` over `en-GB` just because of a larger population; by using root page sorting the same can be achieved but it's more flexible at the same time.

### DIFF
--- a/core-bundle/src/Routing/AbstractPageRouteProvider.php
+++ b/core-bundle/src/Routing/AbstractPageRouteProvider.php
@@ -134,6 +134,28 @@ abstract class AbstractPageRouteProvider implements RouteProviderInterface
         if (null !== $languages && $pageA->rootLanguage !== $pageB->rootLanguage) {
             $langA = $languages[$pageA->rootLanguage] ?? null;
             $langB = $languages[$pageB->rootLanguage] ?? null;
+
+            $primaryA = substr($pageA->rootLanguage, 0, 2);
+            $primaryB = substr($pageB->rootLanguage, 0, 2);
+
+            if ($primaryA !== $primaryB) {
+                // We must not compare by language (without region) if they are the same, otherwise
+                // two pages with same language but different regions might not be sorted by region priority.
+
+                if (null === $langA) {
+                    $langA = $languages[$primaryA] ?? null;
+                }
+
+                if (null === $langB) {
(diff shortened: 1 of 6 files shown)


aschempp:
Failing coverage report seems unrelated.

core-bundle/src/Routing/AbstractPageRouteProvider.php
```diff
@@ -189,19 +211,22 @@ protected function compareRoutes(Route $a, Route $b, array $languages = null): i
...
             $language = str_replace('_', '-', $language);
+            $result[] = $language;
 
             if (5 === \strlen($language)) {
                 $lng = substr($language, 0, 2);
 
                 // Append the language if only language plus dialect is given (see #430)
-                if (!\in_array($lng, $languages, true)) {
-                    $languages[] = $lng;
+                if (!\in_array($lng, $languages, true) && !\in_array($lng, $result, true)) {
```
Toflar: This checks `$result` on every single loop iteration. Wouldn't it be faster to have one `array_unique()` call after collecting them all?

Toflar (commented):
Just a minor nitpick :)

core-bundle/src/Routing/AbstractPageRouteProvider.php
aschempp: improved in f7b7c2e5da7e0abfa712960782a2da3e86bc3233

Toflar (dismissed):

PR #3060, opened 2021-06-07
German country translation of "Isle of Man"

centaurus22:
| Fixed issues | Fix the incomplete German translation of „Isle of Man“ in the language file which contains German translations of country names. The German translation is currently set only to „Isle of“. The associated country code („im“) and the source refers also to the Isle of Man.


aschempp:
Thanks for your report! Translations are managed on transifex.com, but only @leofeyer can modify the german ones, so I assigned this to him 😊

leofeyer:
Fixed on transifex.com. Will be shipped with the next release.

PR #3066, opened 2021-06-08
Rework the response context

Toflar:
Alternative approach for response context as discussed in Core team call.
I guess there's no point in explaining it all here - I'll just document the final iteration once everything is done in the Contao dev docs ^^


calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -132,29 +132,32 @@ protected function compile()
 		// Overwrite the page meta data (see #2853, #4955 and #87)
 		$responseContext = System::getContainer()->get(ResponseContextAccessor::class)->getResponseContext();
 
-		if ($responseContext instanceof WebpageResponseContext)
+		if ($responseContext && $responseContext->has(HtmlHeadBag::class))
```
m-vo, line 135: I'd use an explicit check here (and at the related places):

```suggestion
		if (null !== $responseContext && $responseContext->has(HtmlHeadBag::class))
```

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -26,26 +28,46 @@ public function __construct(ResponseContextAccessor $responseContextAccessor)
...
-    public function createContaoWebpageResponseContext(PageModel $pageModel): ContaoWebpageResponseContext
+    public function createContaoWebpageResponseContext(PageModel $pageModel): ResponseContext
     {
-        $context = new ContaoWebpageResponseContext($pageModel);
-        $this->responseContextAccessor->setResponseContext($context);
+        $context = $this->createWebpageResponseContext();
+
+        if (!$context->has(HtmlHeadBag::class)) {
+            return $context;
+        }
```
m-vo: How could this ever be true?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -12,31 +12,46 @@
...
 
-    public function finalize(Response $response): ResponseContextInterface
+    /**
+     * @return mixed|null
+     */
+    public function get(string $service)
     {
-        foreach ($this->getHeaderBag()->all() as $name => $values) {
-            $response->headers->set($name, $values, false); // Do not replace but add
+        return $this->services[$service] ?? null;
```
m-vo: I'd prefer if we would throw an exception when a service does not exist. Then the result would always be an object and - together with the template annotations above - you could do something like that:

```php
$context->get(HeadBag::class)->setTitle('foo'); // <-- autocompletion here
```

IMO this would closely resemble how `Container#get()` works by default in Sf.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -12,31 +12,46 @@
...
+    }
+
+    public function has(string $service): bool
+    {
+        return null !== $this->get($service);
     }
 
-    public function finalize(Response $response): ResponseContextInterface
+    /**
+     * @return mixed|null
```
m-vo: This wants some generics magic :smile: (also see next comment why I'd also remove `null`):

```suggestion
     * @template T
     * @psalm-param class-string<T> $service
     * @psalm-return T
     *
     * @return object
```

m-vo (changes requested):
IMHO this feels much neater! I like. :heart:

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -204,14 +206,14 @@ protected function prepare($objPage)
 
 		// Set the page title and description AFTER the modules have been generated
 		$this->Template->mainTitle = $objPage->rootPageTitle;
-		$this->Template->pageTitle = htmlspecialchars($responseContext->getTitle());
+		$this->Template->pageTitle = htmlspecialchars($responseContext->get(HtmlHeadBag::class)->getTitle());
```
aschempp, line 209: don't we need a `null`-check here (and the lines below)?

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -26,26 +28,46 @@ public function __construct(ResponseContextAccessor $responseContextAccessor)
...
+
+        /** @var HtmlHeadBag $htmlHeadBag */
+        $htmlHeadBag = $context->get(HtmlHeadBag::class);
+
+        $title = $pageModel->pageTitle ?: StringUtil::inputEncodedToPlainText($pageModel->title ?: '');
```
aschempp, line 55: Why would `pageTitle` not be decoded?

aschempp (changes requested):
We've discussed: services in the `ResponseContext` must not be stored with their class name as array key, but simply as an array of objects. Getting a service then means looping over the objects (in reverse sorting) and returning the first match. This also allows to get a service by interface instead of the class name only. This also means any later added service simply overrides the previously added ones, but not necessary for all interfaces etc.

Toflar:
> Services in the `ResponseContext` must not be stored with their class name as array key, but simply as an array of objects. Getting a service then means looping over the objects (in reverse sorting) and returning the first match. This also allows to get a service by interface instead of the class name only. This also means any later added service simply overrides the previously added ones, but not necessary for all interfaces etc.

I know what we've discussed. I decided against it now during the implementation. There's no point in doing that.
A service like the `HtmlTagBag` will probably never get an interface. And having an additional layer like `HtmlTagBagProvider` with one `getHtmlTagBag(): HtmlTagBag` doesn't provide any useful abastraction because it's still bound to a concrete implementation.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: No, that's also not new code, that's already existing code.

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Well, we set the response context ourselves in this method and we know that it has an `HtmlHeadBag`.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: Defensive coding. If we were to introduce some kind of event in the future or so, who knows what's gonna happen here.

m-vo: But you cannot remove a bag, can you?

Toflar:
Interfaces are now supported and basically work like aliases. You can now find both, concrete implementations or an interface in which case the latest one `add()`ed wins.

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
Toflar: We don't really do that in legacy code.

core-bundle/src/Routing/ResponseContext/HtmlHeadBag/HtmlHeadBag.php
```diff
@@ -10,9 +10,9 @@
  * @license LGPL-3.0-or-later
  */
 
-namespace Contao\CoreBundle\Routing\ResponseContext;
+namespace Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag;
 
-class WebpageResponseContext extends ResponseContext
+class HtmlHeadBag
```
m-vo: As we're composing now instead of inheriting, should this become `final`, too?

leofeyer:
> > This also allows to get a service by interface instead of the class name only. This also means any later added service simply overrides the previously added ones, but not necessary for all interfaces etc.
> 
> I know what we've discussed. I decided against it now during the implementation. There's no point in doing that.
> A service like the `HtmlTagBag` will probably never get an interface.

So why did you add aae109170653ab3d3152819e5afebccb38f92e62 then?

Toflar:
I think having the option can make sense for other use cases but I don't think it does make sense for the `HtmlHeadBag`.
Now, we support both use cases. If you want to have a service that is meant to be replaced, feel free to give it an interface and it will be supported. If you don't, then don't :)

leofeyer:
I see. And how much overhead does using a `Container` object instead of an array add?

Toflar:
Nothing?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -12,31 +12,72 @@
...
-    public function finalize(Response $response): ResponseContextInterface
+    public function add(object $service): self
     {
-        foreach ($this->getHeaderBag()->all() as $name => $values) {
-            $response->headers->set($name, $values, false); // Do not replace but add
+        $this->container->set(\get_class($service), $service);
+
+        // Automatically add aliases for all interfaces (last one added automatically wins by overriding here)
+        foreach ((new \ReflectionClass($service))->getInterfaceNames() as $interfaceName) {
+            $this->container->set($interfaceName, $service);
```
ausi: Should we add parent classes too, so that you can overwrite a service that doesn’t implement an interface?

```suggestion
            $this->container->set($interfaceName, $service);
        }
        
        $parent = new \ReflectionClass($service);
        while ($parent = $parent->getParentClass()) {
            $this->container->set($parent->getName(), $service);
```

ausi (dismissed):
Small suggestion, otherwise very like ❤️

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: Hmmm. What about the parent of the parent? Isn't that an endless discussion?

ausi: > What about the parent of the parent?

My suggested code already handles that ☺️

ausi: This way we add all parents and all interfaces, so it functions exactly the same as type hints everywhere in PHP.

Toflar: Yeah I know, I just don't know if it's a good idea 🙈 Do you have a specific use case in mind and why we would want to support this? I mean, what's the point? If it's a concrete class, there's not two services that can extend and work at the same time. It would be better to use an interface and composition for that case (and even this case is highly unlikely to ever happen in the wild). We can still add this if there's a real use case.

ausi: I just thought it would be more “complete” this way so that all possible `T`s  for `$service instanceof T` get set. As we now store all interfaces, all parent classes interfaces and all parent interfaces of these interfaces, I thought the parent classes are just the last one missing.

But I’m fine too with merging it as is.

Toflar: @contao/developers you decide. I think neither @ausi nor I do have strong preference here. We can add it now, we can also add it later.

m-vo: I would either add parents as well or no aliases at all.

leofeyer: I am with @m-vo.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -26,26 +28,46 @@ public function __construct(ResponseContextAccessor $responseContextAccessor)
...
+
+        $htmlHeadBag
+            ->setTitle($title ?: '')
```
aschempp, line 58: How can `$title` ever be empty / anything else than an empty string?

aschempp (commented):
I don't understand the point of using a `Container` to store the services? The container class does a bazillion other things we don't need? Looks like we can't even reuse aliases or thelike?

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: That's core code, I did not change this.

ausi (dismissed):

Toflar:
Addressed all comments here. Notes:

1. I've removed the usage of `Container` as this seemed to be something people were wondering. I mainly used it because it's simple, already there, provided some levenshtein help if you typed something incorrectly but it doesn't really add any value here so it's gone.
2. I've removed the `remove()` method again. I don't think it makes sense because someone explicitly added support of something to the response context. Why would anyone want to remove that again - sounds wrong.
3. I gave lazy loading some thought (e.g. the response context **would** be capable of some feature but nobody ever accesses it, so no need to intantiate) and there's now also a way to add lazy services.

Toflar:
All green now, I'm happy with this :) Time for a final review.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -12,31 +12,79 @@
...
      * @var PartialResponseHeaderBag|null
      */
     private $headerBag;
 
-    public function getHeaderBag(): ResponseHeaderBag
+    public function add(object $service): self
     {
-        if (null === $this->headerBag) {
-            $this->headerBag = new PartialResponseHeaderBag();
+        return $this->addLazy(\get_class($service), static function () use ($service) { return $service; });
```
aschempp: why would you lazy load the ones that are static? Accessing them would just result in the services key being set?

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -12,31 +12,79 @@
...
+
+        // Automatically add aliases for all interfaces and parents (last one added automatically wins by overriding here)
+        foreach ($ref->getInterfaceNames() as $interfaceName) {
+            $this->services[$interfaceName] = $factory;
         }
 
-        return $this->headerBag;
+        while ($ref = $ref->getParentClass()) {
+            $this->services[$ref->getName()] = $factory;
+        }
```
aschempp: Unfortunately, this isn't going to work, since accessing the same service through multiple classes would call the factory for each of them I think

aschempp (changes requested):
Love the idea of lazy loading services, I was thinking about that as well when I saw the Container being used. Unfortunately, I think there's still a design flaw.

Why do we need to use reflection when setting a service, any reason we can't do an iteration when accessing the property? I know that might be slower in theory, but since not all services are queried that might still be better than doing that upfront?

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -14,8 +14,8 @@
 use Contao\CoreBundle\Exception\PageNotFoundException;
 use Contao\CoreBundle\Exception\RedirectResponseException;
 use Contao\CoreBundle\Image\Studio\Studio;
+use Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag\HtmlHeadBag;
```
leofeyer, line 17: Is the double `\HtmlHeadBag\HtmlHeadBag` intended? It seems like the class `HtmlHeadBag` could also live in the `ResponseContext` namespace.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: Just reusing the logic that's there. No particular reason.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: What doesn't work? The tests work?

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -26,26 +28,42 @@ public function __construct(ResponseContextAccessor $responseContextAccessor)
...
+            ->setMetaDescription(StringUtil::inputEncodedToPlainText($pageModel->description ?: ''))
+        ;
+
+        if ($pageModel->robots) {
+            $htmlHeadBag
+                ->setMetaRobots($pageModel->robots)
+            ;
```
leofeyer:
```suggestion
            $htmlHeadBag->setMetaRobots($pageModel->robots);
```

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
Toflar: It's intended because it will get more classes in the future for head tags.

leofeyer: How will these future classes be called?

Toflar: Who knows? `Script`, `Link`, maybe?

Toflar:
> Unfortunately, I think there's still a design flaw.

I don't think there is. Accessing those details doesn't really cost us anyting (https://3v4l.org/SnrWA). It's correct to do it upfront.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
aschempp: Since you're saving the factory method for multiple service keys, accessing each of the keys will separately call it's factory.

**Example:**
```
$context->addLazy('Foobar extends FooInterface', function () { echo 'test' });

$context->get('Foobar');
$context->get('FooInterface');
```

this would `echo 'test'` twice

Toflar: I don't get the problem, sorry.

Toflar: Maybe you can PR a failing test?

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
leofeyer: That sounds like the following to me then:

```
src/Routing/ResponseContext/HtmlHeadBag.php
src/Routing/ResponseContext/HtmlHeadBag/Script.php
src/Routing/ResponseContext/HtmlHeadBag/Link.php
```

Toflar: I would like to keep the things inside their own namespace. Otherwise I have to add `JsonLdManager` inside `ResponseContext` as well. And then a potential `BodyTagBag` or what not as well and all of a sudden `ResponseContext` becomes cluttered.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
aschempp: Here you go: https://github.com/Toflar/contao/pull/4
Also, please don't resolve an open conversation…

core-bundle/src/Routing/ResponseContext/ResponseContext.php
aschempp: Just looks strange to me, because getting the service then first needs to unwrap the Closure again for no reason…

Toflar: The reason is reusability. That was the quickest way to implement it and it doesn't matter so I just did what I did 🤷‍♂️

core-bundle/src/Routing/ResponseContext/ResponseContext.php
Toflar: Fixed in https://github.com/contao/contao/pull/3066/commits/1d29c95c52158ba7e31347898071e7b5c125bcfc.

Toflar:
@aschempp the unnecessary closure wrapping is now fixed with https://github.com/contao/contao/pull/3066/commits/1d29c95c52158ba7e31347898071e7b5c125bcfc as well.

aschempp (changes requested):
Requesting a service now overrides aliases, which means the priority is changed. The second `add` of a service should override existing ones, including the aliases. See https://github.com/Toflar/contao/pull/5

ausi (dismissed):

Toflar:
> Requesting a service now overrides aliases, which means the priority is changed. The second `add` of a service should override existing ones, including the aliases. See [Toflar#5](https://github.com/Toflar/contao/pull/5)

Fixed in https://github.com/contao/contao/pull/3066/commits/1cb74c7af0fd9c2cdf4eb5150c5dc6943d53daa1.

core-bundle/src/Routing/ResponseContext/ResponseContext.php
```diff
@@ -31,12 +87,34 @@ public function getHeaderBag(): ResponseHeaderBag
...
+    /**
+     * @param \Closure|object $objectOrFactory
+     */
+    private function registerService(string $serviceId, $objectOrFactory): void
+    {
+        $this->services[$serviceId] = $objectOrFactory;
+        $this->current[$serviceId] = $serviceId;
+
+        foreach ($this->getAliases($serviceId) as $alias) {
+            $this->services[$alias] = $objectOrFactory;
```
ausi: Is this line actually needed then? As you always look up `$this->current` when getting a service I think this line can be removed. I tested it, and the unit tests still pass with this line removed.

aschempp: I think so too, this would still be a factory problem otherwise.

Toflar: Ah yes, of course. That was a leftover which I've removed in 84af804d441dd1e0f4663e7b97d7323f7668e52f.

aschempp (dismissed):

ausi (dismissed):

leofeyer:
Thank you @Toflar.

PR #3067, opened 2021-06-09
Lazy-load the `rootFallbackLanguage` property

aschempp:
I noticed the `rootFallbackLanguage` is always loaded for a page, but it is only ever used to fetch (fallback) metadata. I think we can save that SQL query by lazy loading the data.

Also, as far as I understand, this SQL query would be done for EVERY `PageModel::loadDetails()` since it cannot be cached in the registry. That means potentially a lot of calls because Sf routing uses `loadDetails` on all candidates.

### DIFF
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -284,7 +284,8 @@ class PageModel extends Model
 	public function __get($strKey)
 	{
 		// Lazy-load the fallback language (see contao/core#6874)
-		if ($strKey == 'rootFallbackLanguage' && $this->blnDetailsLoaded && !\array_key_exists('rootFallbackLanguage', $this->arrData)) {
+		if ($strKey == 'rootFallbackLanguage' && $this->blnDetailsLoaded && !\array_key_exists('rootFallbackLanguage', $this->arrData))
+		{
 			$this->rootFallbackLanguage = $this->rootIsFallback ? $this->language : null;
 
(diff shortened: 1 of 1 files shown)


ausi:
Should we also append the data in the `row()` method?

aschempp:
> Should we also append the data in the `row()` method?

Added in e3d61e188fac42cf08cbffd8ddf38d76085a7470

Toflar (approved):
We really need to get rid of this on-the-model-inheritance nightmare at some point 🙈

ausi (approved):

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -281,6 +281,40 @@ class PageModel extends Model
...
+			return true;
+		}
+
+		return parent::__isset($strKey);
+	}
+
+	public function row()
+	{
+		// Load lazy properties
+		$this->__get('rootFallbackLanguage');
```
leofeyer, line 313: I am not sure about this change. Now `rootFallbackLanguage` will always be available even if we have not called `loadDetails()`. And it will return `null` anyway, because `$this->rootIsFallback` will not be set without calling `loadDetails()`, therefore we might as well not preload the value.

Toflar: No, it won't:

```php
if ($strKey == 'rootFallbackLanguage' && $this->blnDetailsLoaded && !\array_key_exists('rootFallbackLanguage', $this->arrData))
```

leofeyer: True. NVM.

PR #3068, opened 2021-06-09
Added JSON output format to contao:migrate command

aschempp:
This might finally allow the Contao Manager to integrate migrations at some point.

Since we added the `--migrations--only` option recently, I hope this is fine to add in 4.9 as well 😉 

Not sure if we need to disallow any of the other options? They're currently simply ignored, since `--format=json` means only output the data…

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -77,6 +77,7 @@ class MigrateCommand extends Command
             ->addOption('with-deletes', null, InputOption::VALUE_NONE, 'Execute all database migrations including DROP queries. Can be used together with --no-interaction.')
             ->addOption('schema-only', null, InputOption::VALUE_NONE, 'Only update the database schema.')
             ->addOption('migrations-only', null, InputOption::VALUE_NONE, 'Only execute the migrations.')
+            ->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format', 'text')
             ->setDescription('Executes migrations and updates the database schema.')
         ;
     }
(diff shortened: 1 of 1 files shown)


core-bundle/src/Command/MigrateCommand.php
```diff
@@ -85,6 +86,22 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                'schema' => $this->installer->getCommands(false),
+            ];
+
+            $this->io->write(json_encode($data));
+
+            return 0;
+        }
+
+        if ('text' !== $input->getOption('format')) {
+            throw new \LogicException('Invalid format: '.$input->getOption('format'));
```
Toflar, line 102:
```suggestion
            $this->io->error('Invalid format: '.$input->getOption('format'));
            
            return 1;
```

ausi:
Doesn’t executing the migrations need a JSON format ouput too?

core-bundle/src/Command/MigrateCommand.php
aschempp: Not sure what's right, I've copied this from here: https://github.com/contao/contao/blob/4.x/core-bundle/src/Command/UserListCommand.php#L86

aschempp:
> Doesn’t executing the migrations need a JSON format ouput too?

Good question, does it? Since the command is mostly interactive, not sure what JSON would be output. I would assume the following when using this in the Contao Manager:
1. run `--format=json` to get the necessary changes
2. if there's runonce/migrations, run `--migrations-only --no-interaction`
3. check for schema changes again
4. if there are schema changes, run with `--schema-only --no-interaction`

Since the JSON API cannot be interactive, a JSON output wouldn't be useful there? Or would you think to add JSON to the `--no-interaction` runtime?

ausi:
Yes, I think `--format=json` should only be valid together with `--no-interaction` or `--dry-run`.

The Manager would then execute all with `--format=json`:

1. `--dry-run`
2. `--migrations-only --no-interaction`
3. `--dry-run`
4. `--schema-only --no-interaction`
5. `--dry-run`

And we would have to add `--dry-run`, but that would be a useful option for the regular command too IMO.

aschempp:
what would `--dry-run` do? and what would `--format=json` output when running migrations?

ausi:
> what would `--dry-run` do?

Show the the migrations, runonces and schema updates that would run, but not actually execute anything. I think this pattern with `--dry-run` is pretty common.

> what would `--format=json` output when running migrations?

Returning something like this:

```json
{
	"migrations": [
		{
			"name": "Contao\\CoreBundle\\Migration\\Version411\\RemoveJsHighlightMigration",
			"successful": true,
			"message": "Message from the migration"
		},
		{
			"name": "Contao\\CoreBundle\\Migration\\Version411\\SomeOtherMigration",
			"successful": false,
			"message": "Message from the migration"
		}
	]
}
```

But we could also use LDJSON (Line-delimited JSON) to make it streamable if you want to show the result live in the manager.

leofeyer:
What is necessary to get this PR approved?

ausi:
I can work an a PR that implements the features outlined in https://github.com/contao/contao/pull/3068#issuecomment-857675488

leofeyer:
Closing this in favor of #3310.

PR #3072, opened 2021-06-09
Do not try to add an image if the source is empty

leofeyer:
@m-vo This will prevent an exception if `$this->singleSRC` is empty.

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentAccordion.php
+++ b/core-bundle/src/Resources/contao/elements/ContentAccordion.php
@@ -38,7 +38,7 @@ class ContentAccordion extends ContentElement
 		$this->Template->addBefore = false;
 
 		// Add an image
-		if ($this->addImage)
+		if ($this->addImage && $this->singleSRC)
 		{
 			$figure = System::getContainer()
 				->get(Studio::class)
--- a/core-bundle/src/Resources/contao/elements/ContentText.php
+++ b/core-bundle/src/Resources/contao/elements/ContentText.php
@@ -44,7 +44,7 @@ class ContentText extends ContentElement
 		$this->Template->addBefore = false;
 
 		// Add an image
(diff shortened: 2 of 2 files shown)


fritzmg:
afaik this should not be necessary when using `buildIfResourceExists`.

leofeyer:
Ok, then we must ensure that the method accepts and handles `null`:

```
Contao\CoreBundle\Image\Studio\FigureBuilder::fromPath(): Argument #1 ($path) must be of type string, null given, called in …/vendor/contao/contao/core-bundle/src/Image/Studio/FigureBuilder.php on line 306
```

@m-vo /cc

fritzmg:
I am not sure now actually. @m-vo what do you think? How lenient should the FigureBuilder be? We could change it to 

```php
        return $this->fromPath((string) $identifier);
```

in `from()` and may be add

```php
        if (empty($identifier)) {
            $this->lastException = new InvalidResourceException("Given path is empty.");

            return $this;
        }
```

at the beginning of `fromPath()`, which would cover this case.

m-vo:
Ah, now I see. IMHO we should not weaken the type constraints for any of the more specific "from methods" (like `fromPath(string $path)`) but we could allow `null` in the generic `from(mixed $resource)`.

leofeyer:
Allowing `null` in the `from()` method sounds reasonable, however, we also have to handle the value. The method currently returns `int|string|FilesModel|ImageInterface` and none of these types makes sense for `null`, does it?

m-vo:
Here is how it could look: #3073

m-vo:
> The method currently returns int|string|FilesModel|ImageInterface and none of these types makes sense for null, does it?

They are all fluent interface methods that return `self`. I think you might have confused it with the input argument?

leofeyer:
See #3073

PR #3073, opened 2021-06-10
Allow null in FigureBuilder#from()

m-vo:
Alternative PR for #3072.

/cc @leofeyer, @fritzmg


leofeyer:
I'm not sure if we should throw an exception, because this will make it impossible to delete or edit the content element in the back end. (The back end preview will try to render the image.)

m-vo:
The exception is only thrown if you call `build()`. The content elements will call `buildIfResourceExists()`, which returns `null` instead.

fritzmg (approved):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3075, opened 2021-06-10
Harden `FigureBuilder::from()` against invalid types

m-vo:
This is a followup PR to #3073 that would further enforce type safety in `FigureBuilder#from()`.

With PHP8.0 you could type hint the input argument with `int|string|FilesModel|ImageInterface|null` and would get a `TypeError` if you tried otherwise. This PR introduces a similar forward compatible check. Not sure if we want to go that route but input is welcome.

/cc @ausi

### DIFF
--- a/core-bundle/src/Image/Studio/FigureBuilder.php
+++ b/core-bundle/src/Image/Studio/FigureBuilder.php
@@ -315,7 +315,9 @@ class FigureBuilder
             return $this->fromPath($identifier);
         }
 
-        throw new \TypeError(sprintf('%s(): Argument #1 ($identifier) must be of type FilesModel|ImageInterface|string|int|null, %s given', __METHOD__, \is_object($identifier) ? \get_class($identifier) : \gettype($identifier)));
+        $type = \is_object($identifier) ? \get_class($identifier) : \gettype($identifier);
+
+        throw new \TypeError(sprintf('%s(): Argument #1 ($identifier) must be of type FilesModel|ImageInterface|string|int|null, %s given', __METHOD__, $type));
     }
 
     /**
--- a/core-bundle/src/Image/Studio/FigureBuilder.php
+++ b/core-bundle/src/Image/Studio/FigureBuilder.php
@@ -315,10 +315,7 @@ class FigureBuilder
             return $this->fromPath($identifier);
         }
 
-        $method = __METHOD__;
-        $type = \is_object($identifier) ? \get_class($identifier) : \gettype($identifier);
-
-        throw new \TypeError("$method(): Argument #1 (\$identifier) must be of type FilesModel|ImageInterface|string|int|null, $type given");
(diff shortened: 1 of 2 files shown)


m-vo:
And before you ask why I also added a `string` type check for `Validator::isUuid`: The methods in this class announce they can work with `mixed` input, but in reality this depends on the PHP version and if the calling context has set `strict_types=1`. This is a whole other rabbit hole to go down…

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -301,15 +301,24 @@ public function from($identifier): self
...
 
-        return $this->fromPath($identifier);
+        if ($isString) {
+            return $this->fromPath($identifier);
+        }
+
+        $method = __METHOD__;
+        $type = \is_object($identifier) ? \get_class($identifier) : \gettype($identifier);
+
+        throw new \TypeError("$method(): Argument #1 (\$identifier) must be of type FilesModel|ImageInterface|string|int|null, $type given");
```
leofeyer:
```suggestion
        throw new \TypeError(sprintf(
            '%s(): Argument #1 ($identifier) must be of type FilesModel|ImageInterface|string|int|null, "%s" given',
            ___METHOD__,
            \is_object($identifier) ? \get_class($identifier) : \gettype($identifier)
        ));
```

That is the classic `sprintf()` case IMHO. 😄

core-bundle/tests/Image/Studio/FigureBuilderTest.php
```diff
@@ -345,6 +345,43 @@ public function testFromNullFails(): void
...
+        $framework = $this->mockContaoFramework([
+            Validator::class => new Adapter(Validator::class),
+        ]);
+
+        $figureBuilder = $this->getFigureBuilder(null, $framework);
+
+        $this->expectException(\TypeError::class);
+        $this->expectExceptionMessage(
+            sprintf('Contao\CoreBundle\Image\Studio\FigureBuilder::from(): Argument #1 ($identifier) must be of type FilesModel|ImageInterface|string|int|null, %s given', $typeString)
+        );
```
leofeyer:
```suggestion

        $this->expectExceptionMessage(sprintf(
            'Contao\CoreBundle\Image\Studio\FigureBuilder::from(): Argument #1 ($identifier) must be of type FilesModel|ImageInterface|string|int|null, %s given',
            $typeString
        ));
```

core-bundle/src/Image/Studio/FigureBuilder.php
m-vo: That's right, but then the CS fixer tries to cram everything in a single line. :smile: 
I'll change it.

leofeyer: Ah, I see. Then we might want to keep `$type = \is_object($identifier) ? \get_class($identifier) : \gettype($identifier)` separately?

m-vo: See 55cd2839d, IMO that's a bit neater.

ausi (approved):
❤️

leofeyer:
Thank you @m-vo.

PR #3076, opened 2021-06-10
Add a license field to the file metadata

Toflar:
File/Image licensing information feels like a very common requirement. Let's add this to the core.
My plan is to use that for JSON-LD once the whole response context stuff is settled :)

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_files.php
+++ b/core-bundle/src/Resources/contao/dca/tl_files.php
@@ -252,7 +252,8 @@ $GLOBALS['TL_DCA']['tl_files'] = array
 					'title'           => 'maxlength="255"',
 					'alt'             => 'maxlength="255"',
 					'link'            => array('attributes'=>'maxlength="255"', 'dcaPicker'=>true),
-					'caption'         => array('type'=>'textarea')
+					'caption'         => array('type'=>'textarea'),
+					'license'         => array('attributes'=>'maxlength="255"', 'dcaPicker'=>true)
 				)
 			),
 			'sql'                     => "blob NULL"
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -28,6 +28,7 @@ class Metadata
     public const VALUE_TITLE = 'title';
     public const VALUE_URL = 'link';
     public const VALUE_UUID = 'uuid';
+    public const VALUE_LICENSE = 'license';
 
     /**
      * Key-value pairs of metadata.
@@ -80,6 +81,11 @@ class Metadata
     public function getTitle(): string
     {
         return $this->values[self::VALUE_TITLE] ?? '';
+    }
+
+    public function getLicense(): string
+    {
+        return $this->values[self::VALUE_LICENSE] ?? '';
     }
 
     public function getUrl(): string
--- a/core-bundle/src/File/Metadata.php
+++ b/core-bundle/src/File/Metadata.php
@@ -83,11 +83,6 @@ class Metadata
         return $this->values[self::VALUE_TITLE] ?? '';
     }
 
-    public function getLicense(): string
-    {
-        return $this->values[self::VALUE_LICENSE] ?? '';
-    }
-
     public function getUrl(): string
     {
         return $this->values[self::VALUE_URL] ?? '';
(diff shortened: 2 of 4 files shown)


m-vo:
Nice! Do we also want to output this by default in the image templates?

fritzmg:
We actually decided against this once: https://github.com/contao/core-bundle/issues/1297#issuecomment-373392678

Toflar:
In schema data then, yes. But not im HTML so :)

fritzmg:
I am still very much in favour ;). Just wanted to mention it here. I think originally it was decided against adding a field by default in general, regardless of its output.

fritzmg (dismissed):

m-vo:
> We actually decided against this once: contao/core-bundle#1297 (comment)

Reading this, I think we maybe need some way to differentiate between/handle translated and (likely) non-translated properties like `license` at some point.

Toflar:
I think we decided against it back then because there is no standardized way to display it in HTML. Hence we still don't do it. It's just preparation work for schema.org where it has a specific property :)

m-vo (dismissed):

fritzmg:
> I think we decided against it back then because there is no standardized way to display it in HTML.

There is though: https://github.com/contao/core-bundle/issues/1297#issuecomment-373394876

And I think some stock photo providers are pretty nitpicky anyway about where the license is displayed on the site. ([Apparently](https://community.contao.org/de/showthread.php?81070-Ursprüngliche-Metadaten-werden-im-Sourceset-entfernt) some even require you to have the meta data available in the image itself, which is currently not possible with Contao as these are stripped while processing the image.)

core-bundle/src/Resources/contao/dca/tl_files.php
```diff
@@ -252,7 +252,8 @@
 					'title'           => 'maxlength="255"',
 					'alt'             => 'maxlength="255"',
 					'link'            => array('attributes'=>'maxlength="255"', 'dcaPicker'=>true),
-					'caption'         => array('type'=>'textarea')
+					'caption'         => array('type'=>'textarea'),
+					'license'         => array('attributes'=>'maxlength="255"', 'dcaPicker'=>true)
```
leofeyer, line 256: Why did you add the picker here?

ausi:
> ([Apparently](https://community.contao.org/de/showthread.php?81070-Urspr%C3%BCngliche-Metadaten-werden-im-Sourceset-entfernt) some even require you to have the meta data available in the image itself, which is currently not possible with Contao as these are stripped while processing the image.)

Related: https://github.com/contao/image/issues/80

Toflar:
Most services support schema.org's [`ImageObject->license` property](https://schema.org/license), just like Google does. That's also why I added the picker because it's most likely to be an URL (which can also be a custom license page with your own terms).

core-bundle/src/Resources/contao/dca/tl_files.php
Toflar: https://github.com/contao/contao/pull/3076#issuecomment-858550272

ausi:
> we maybe need some way to differentiate between/handle translated and (likely) non-translated properties like `license` at some point.

How about adding a feature to the `metaWizard` so that it can handle non-language specific fields too? Or how about just adding it as a regular column to `tl_files`?

Toflar:
`license` is a translated property though. So that's for a different discussion and once again, not related to this PR.

ausi (dismissed):

leofeyer:
> So that's for a different discussion and once again, not related to this PR.

I think the discussion is related to this PR, because you are now adding the field as translated property, which e.g. does not make much sense if the entry is always `https://creativecommons.org/licenses/by/4.0/`.

Toflar:
If it is the same URL, then yes. But you can have the same license description in many languages. If there's only one, then only add it to your fallback language.

fritzmg:
> If there's only one, then only add it to your fallback language.

But all the meta fields are always translated together, not individually, aren't they? So you cannot have a fallback just for the `license` field at the moment, whereas `title` should still be translated for example.

m-vo:
Yes, currently the first locale that has data wins. But we could maybe adjust this. Probably not without affecting BC, though.

ausi:
> I think the discussion is related to this PR, because you are now adding the field as translated property, which e.g. does not make much sense if the entry is always `https://creativecommons.org/licenses/by/4.0/`.

The `link` property might also not need to be translated in many cases, so I think the `license` field could be seen the same way as the current `link` field.

> If there's only one, then only add it to your fallback language.

As the fallback language is defined per root page this is probably problematic. But we could extend the `metaWizard` to support entering a fallback explicitly. But that would be not related to this PR 😊

ausi (approved):

fritzmg (approved):

leofeyer:
Thank you @Toflar.

PR #3077, opened 2021-06-10
Correctly render showColumns in the picker widget

aschempp:
If the universal picker widget was used on a table with `showColumns` enabled, the result was simply an "imploded" list of values. This PR fixes the display to render the same table as DC_Table does. It also adds support to drag&drop table rows instead of list items.

This is what it looks like when picking members (ignore the field name 😉):
Bildschirmfoto 2021-06-10 um 17 01 24


leofeyer (changes requested):
I don't like dragging and dropping lines without using the designated icon. It is inconsistent and will interfere with selecting text from these lines.

aschempp:
well that's the same with gallery or page picker, anything in the universal picker?

leofeyer:
No, because items in the universal picker are not sortable at all. The icon is not available.

aschempp:
what icon do you mean? The screenshot has the usual "drag & drop to sort" text above it, same as if you pick multiple pages.

Here's what the page picker in Isotope looks like. There's no hint (seems like its explicitly disabled) but I can still sort the pages by drag&drop.
Bildschirmfoto 2021-06-14 um 07 31 47

leofeyer:
I understand what you mean. It still confuses me to see the list view in the picker widget, but I guess that is just a matter of getting used to it.

PR #3078, opened 2021-06-10
Purge log and undo tables via cron

Toflar:
I've noticed that my `tl_log` table contained entries from 2018 and I was a bit confused because I remembered we once adjusted the `logPeriod` to 7 days for GDPR reasons.

I debugged and quickly found the reason: The cleanup is only ever done when someone actually looks at the log entries in the back end 🤦 

So here we go, this is now handled by an hourly cronjob.

### DIFF
--- a/core-bundle/src/Cron/CleanupCron.php
+++ b/core-bundle/src/Cron/CleanupCron.php
@@ -43,16 +43,20 @@ class CleanupCron
 
         $undoPeriod = (int) $this->framework->getAdapter(Config::class)->get('undoPeriod');
         $logPeriod = (int) $this->framework->getAdapter(Config::class)->get('logPeriod');
+        $versionPeriod = (int) $this->framework->getAdapter(Config::class)->get('versionPeriod');
 
-        // Clean up old tl_undo and tl_log entries
-        if ($undoPeriod > 0) {
-            $stmt = $this->connection->prepare('DELETE FROM tl_undo WHERE tstamp<:tstamp');
-            $stmt->executeStatement(['tstamp' => time() - $undoPeriod]);
+        $this->cleanTable('tl_undo', $undoPeriod);
+        $this->cleanTable('tl_log', $logPeriod);
+        $this->cleanTable('tl_version', $versionPeriod);
+    }
+
+    private function cleanTable(string $table, int $period): void
+    {
+        if ($period <= 0) {
+            return;
         }
 
-        if ($logPeriod > 0) {
-            $stmt = $this->connection->prepare('DELETE FROM tl_log WHERE tstamp<:tstamp');
-            $stmt->executeStatement(['tstamp' => time() - $logPeriod]);
-        }
+        $stmt = $this->connection->prepare(sprintf('DELETE FROM %s WHERE tstamp<:tstamp', $table));
+        $stmt->executeStatement(['tstamp' => time() - $period]);
     }
 }
--- a/core-bundle/src/Cron/CleanupCron.php
+++ b/core-bundle/src/Cron/CleanupCron.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Cron;
+
+use Contao\Config;
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Doctrine\DBAL\Connection;
+
+class CleanupCron
+{
+    /**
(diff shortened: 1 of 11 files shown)


fritzmg (dismissed):

aschempp (commented):
Looks good. Should it maybe also cleanup the version table?

I don't really like the `Cleanup` name but couldn't come up with any better either 🤔

core-bundle/src/Cron/CleanupCron.php
```diff
@@ -0,0 +1,62 @@
...
+    private $connection;
+
+    public function __construct(ContaoFramework $framework, Connection $connection)
+    {
+        $this->framework = $framework;
+        $this->connection = $connection;
+    }
+
+    /**
+     * @\Contao\CoreBundle\ServiceAnnotation\CronJob("daily")
```
ausi: Should we add a use statement instead of using the FQCN?

ausi (dismissed):

fritzmg:
Btw. should the interval may be be `minutely` instead of `daily`? Otherwise there will be entries that are up to less than one day older than they should be.

leofeyer:
I think `hourly` is fine.

leofeyer (changes requested):
I also do not like the name `CleanupCron`. What about `Contao\CoreBundle\Cron\PurgeExpiredData` or `Contao\CoreBundle\Cron\PurgeLogTables`?

fritzmg:
> I think `hourly` is fine.

I mistakenly wrote `hourly`. Currently it is set to `daily`.

Toflar:
I don't think running these queries every minute makes any sense. Whether or not you keep the log entries a bit longer shouldn't matter. We can agree on `hourly`, though.

I named it `CleanupCron` because I thought that we might add other crons there in the future, if they have the same dependencies (framework and connection - quite likely to happen). `PurgeLogTables` sounds wrong. First it's not purging but pruning. Second, it's not only the log tables. I'd go with `PruneExpiredDataCron` if that sounds better to you than `CleanupCron`?

m-vo:
I'd rather have a single service per operation. Yes, there is possibly duplication, but then it's easy to change things if you need to.

Otherwise +1 for `PruneExpiredDataCron` (or `PruneLogTablesCron`) - directly tells you what's in there.

Toflar:
All comments addressed.

fritzmg (dismissed):

ausi (dismissed):

core-bundle/src/Cron/PruneExpiredDataCron.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+namespace Contao\CoreBundle\Cron;
+
+use Contao\Config;
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\ServiceAnnotation\CronJob;
+use Doctrine\DBAL\Connection;
+
+class PruneExpiredDataCron
```
leofeyer: Shouldn't this be `PurgeExpiredDataCron`? I have googled "prune" and it does not seem correct in this context.

core-bundle/src/Cron/PruneExpiredDataCron.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @CronJob("hourly")
+     */
+    public function onHourly(): void
+    {
+        $this->framework->initialize();
+
+        $undoPeriod = (int) $this->framework->getAdapter(Config::class)->get('undoPeriod');
```
leofeyer: Instead of using `$this->framework->getAdapter(Config::class)` three times, why don't we use the following?

```php
/** @var Config $config */
$config = $this->framework->getAdapter(Config::class)
```

This will also make PhpStorm understand the `->get()` call.

core-bundle/src/Cron/PruneExpiredDataCron.php
Toflar: In my world, purging means deleting all data. Pruning means some data. We're purging expired data but we don't delete all the data so. I don't really care. Feel free to update it.

core-bundle/src/Cron/PruneExpiredDataCron.php
Toflar: I don't mind - CS :)

core-bundle/src/Cron/PruneExpiredDataCron.php
leofeyer: Will do. I got ~4,800 results for `"prune expired"` and ~31,100 for `"purge expired"`, so I guess the latter is more commonly used.

leofeyer:
Thank you @Toflar.

PR #3079, opened 2021-06-11
Fix another invalid array access

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

Prevents `Warning: Undefined array key "protected"`.

The code in question is called here: https://github.com/contao/contao/blob/3a8da4a6e638645f0730ebd477f33eac21e92b61/core-bundle/src/Resources/contao/library/Contao/Controller.php#L568-L569


leofeyer:
Thank you @m-vo.

PR #3080, opened 2021-06-11
Reduce memory consumption in search

### LINKED ISSUE #3045: Searching requires a lot of PHP memory
**Affected version(s)**

Contao 4.9+

**Description**

If there is a large number of records in the search index, searching for a keyword that occurs often (such as `contao` on contao.org) will require a lot of PHP memory limit.

@ausi Maybe we can optimize the query?

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3045

In my test with the DB from contao.org it reduced the search for `contao` from 1.528 sec. with 582.0 MiB down to 0.133 sec. with 6.0 MiB. The query for `c*` went down from 3.766 sec. with 590.0 MiB to 1.016 sec. with 4.0 MiB.


Toflar (changes requested):
We're introducing a new API, can we please stop using so many arguments and instead have a `QueryConfig` object or so? Also, we already have a `Document` - maybe we can reuse that for the results/indexing?

ausi:
> Also, we already have a `Document` - maybe we can reuse that for the results/indexing?

I don’t think we can use them very good for the results, indexing is out of scope of this PR IMO.

> We're introducing a new API, can we please stop using so many arguments and instead have a `QueryConfig` object or so?

I did’t want to rewrite all that legacy code. We sould instead invest time in a complete rewrite of the serch IMO.

Toflar:
> I don’t think we can use them very good for the results, indexing is out of scope of this PR IMO.

Agreed.



> I did’t want to rewrite all that legacy code. We sould instead invest time in a complete rewrite of the serch IMO.

It doesn't have to be beautiful with an interface etc. But just the same like you did for `SearchResult`. A concrete implementation for `SearchQuery`. Just to make sure that if we need to add something new, we don't have to add another parameter to the method but can use the object.

leofeyer (dismissed):
Looks good to me. 👍🏻 I don't think that we do need a config object, because the parameters are very unlikely to change and we already plan on using `contao/search` in the future.

Toflar:
`contao/search` died but yes regarding the alternative. I still think an object would make a lot of sense for the time being.

core-bundle/src/Resources/contao/library/Contao/SearchResult.php
```diff
@@ -0,0 +1,130 @@
...
+	public function getResults(int $intCount = PHP_INT_MAX, int $intOffset = 0): array
+	{
+		$arrResults = \array_slice($this->arrResultsById, $intOffset, $intCount, true);
+		$strIds = implode(',', array_keys($arrResults));
+		$strQuery = 'SELECT * FROM tl_search WHERE ID in (' . $strIds . ')';
+		$objResult = Database::getInstance()->prepare($strQuery)->execute();
+
+		while ($arrRow = $objResult->fetchAssoc())
+		{
+			$arrResults[$arrRow['id']] = array_merge($arrRow, $arrResults[$arrRow['id']]);
```
leofeyer: Is it worth trying to optimize this? Instead of

```php
$ids = [];

while (/* something */) {
    $ids = array_merge($ids, $moreIds);
}
```

we could probably do

```php
$tmp = [];

while (/* something */) {
    $tmp[] = $moreIds;
}

$ids = array_merge(...$moreIds);
```

(I haven't tested the code above.)

ausi: every single `array_merge()` merges the (full) data from the database with the fields from the existing data (id, protected, groups, and relevance)...

While writing this answer, it would probably be enough to just copy the `relevance` over.

```suggestion
			$arrRow['relevance'] = $arrResults[$arrRow['id']]['relevance'];
			$arrResults[$arrRow['id']] = $arrRow;
```

ausi: Changed in fd6fa1094a2989d328b2f941406b281402c5876f

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3081, opened 2021-06-12
Fix the unit tests on PHP 8.1

ausi:
Functional tests are still failing, but unit tests are green now 🎉

### DIFF
--- a/manager-bundle/src/ContaoManager/Plugin.php
+++ b/manager-bundle/src/ContaoManager/Plugin.php
@@ -314,7 +314,7 @@ class Plugin implements BundlePluginInterface, ConfigPluginInterface, RoutingPlu
             $connection->connect();
             $connection->query('SHOW TABLES');
             $connection->close();
-        } catch (DriverException $e) {
+        } catch (DriverException | \mysqli_sql_exception $e) {
             $extensionConfigs[] = [
                 'dbal' => [
                     'connections' => [


leofeyer:
Most of the deprecations are `Implicit conversion from float … to int loses precision`.

https://github.com/php/php-src/blob/f5c209989a679e187abfeb6441de47005a927683/Zend/zend_operators.c#L811-L818

It seems we have to cast these values explicitly?

ausi:
> It seems we have to cast these values explicitly?

Yes, `(int)` cast as well as `round()`, `floor()` and `ceil()` seem to work: <https://3v4l.org/0h0O8/rfc#focus=git.master>

But this should only be needed if the float value has digits after the comma <https://3v4l.org/XgBal/rfc#focus=git.master>

m-vo:
> But this should only be needed if the float value has digits after the comma

I think it's more how the number is interpreted: https://3v4l.org/kQXO9/rfc#focus=git.master 

It's probably valid if all bits match the integer representation. I wonder what happens with non-representable whole numbers (like 2^53+1), though. :smile:

leofeyer:
Thank you @ausi.

leofeyer:
On second thought, what if the `mysqli` PHP extension is not installed? Because the change is not in a test class but in a production one. 🙈

ausi:
> On second thought, what if the `mysqli` PHP extension is not installed?

This should not be relevant because non-exisitng interfaces/exceptions just get ignored: https://3v4l.org/UnkTi

leofeyer:
Good. Thanks for checking. ❤️

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -314,7 +314,7 @@ private function addDefaultServerVersion(array $extensionConfigs, ContainerBuild
             $connection->connect();
             $connection->query('SHOW TABLES');
             $connection->close();
-        } catch (DriverException $e) {
+        } catch (DriverException | \mysqli_sql_exception $e) {
```
m-vo, line 317: @ausi on my local 8.1 test machine (yeah, did it…) this error was triggered when executed by `PluginTest#testDoesNotAddDefaultPdoOptionsIfDriverIsNotPdo()`:

```
mysqli_sql_exception: No database selected
```

While I assume the fix is still correct, the test might be testing the wrong thing? Wdyt?

ausi: We could add a test that tests this default server version explicitly I guess?
Or do you mean we should check why there is no database selected in the affected test?

m-vo: As discussed on Slack we should maybe even catch `\Throwable` because an exception should never bubble here.

I looked again at the test. I think it's OK the way it is now. The try block will always fail - if we wanted to prevent this we would need to inject a mocked DBAL connection factory or make it functional. IMHO not worth it.

PR #3084, opened 2021-06-14
Rename the `web` folder to `public`

### LINKED ISSUE #1088: Use/support `public/` folder for Contao Managed Edition
Since Symfony 4.4, in a regular Symfony application you would have support for the following structure:

| App | Bundle |
| --- | --- |
| `config/` | `config/` |
| `public/` | `public/` |
| `src/` | `src/` |
| `templates/` | `templates/` |
| `translations/` | `translations/` |

Basically, everything in the bundle is consistent with the application itself.

After [this PR](https://github.com/contao/contao/pull/1085) is merged, the Contao Managed Edition would support the following:

| App | Bundle |
| --- | --- |
| `config/` | `config/` |
| `contao/` | `contao/` |
| **`web/`** | `public/` |
| `src/` | `src/` |
| `templates/` | `templates/` |
| `translations/` | `translations/` |

So only the `web/` folder would be inconsistent with the regular Symfony application _and_ bundle structure.

I know this has been discussed in the past and it was decided against it - mostly to not cause more confusion within the community (I think?). But due to the recent PR I would like to bring it up and discuss it again :)

### LINKED ISSUE #2461: Setting alternative web/public dir is inconsistent/fails
**Affected version(s)**

Contao 4.9

**Description**

see discussion https://github.com/contao/contao/pull/2183#discussion_r511483927

Contao uses two sources for the configured web/public directory: The config from `extra.symfony-public-dir` / `extra.public-dir` (root `composer.json`) and the config key `contao.web_dir` (`core-bundle`). 

The bundle config defaults to `%kernel.project_dir%/web`. It's used in quite some places (templating, images, combiner, ...). The current composer script handler uses the entry from the root `composer.json` and defaults to `web` (relative path). So you need to keep these things in sync otherwise either one fails.

**Possible solution**

Imo we should only only use the config in our code and its default should be:
1. root `composer.json` &rarr;`extra.public-dir`
1. if it does not exist:  root `composer.json` &rarr;  `extra.symfony-public-dir`
1. if it does not exist: `%kernel.project_dir%/web`

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Implements #1088
| Docs PR or issue | contao/docs#776

This PR renames the `web` folder to `public` as discussed here: https://github.com/contao/contao/issues/1088#issuecomment-598292084

@contao/developers How do we want to implement backwards compatibility? I see two options:

1. Add a check in the `Configuration` class: If `contao.web_dir` is `public` and there is no `public` folder but a `web` folder, change the value to `web` automatically.

2. Make the `GenerateSymlinkCommand` generate a symlink from `public` to `web` if a `web` folder exists.

I am in favor of solution no. 1, because I find having two different public folders confusing. WDYT?


fritzmg:
Don't we need to solve https://github.com/contao/contao/issues/2461 first?

leofeyer:
No:

> As discussed in Mumble on March 11, it should work in Contao 4.9 if you set the web directory in both the composer.json and the app configuration. If this is not the case, we should fix it.

It seems to work though.

aschempp:
Maybe we should change this default (once and for all) in Contao 5 only? That would probably solve a lot of potential headaches (including the Contao Manager).

leofeyer:
We have already discussed this (see https://github.com/contao/contao/issues/1088#issuecomment-598292084).

leofeyer:
This PR is ready to merge. I have thoroughly tested the changes and made sure that the installation commands fall back to `web` if the directory exist. It therefore does not matter when the Contao Manager is adjusted to use `public`.

ausi (approved):

aschempp:
> This PR is ready to merge. I have thoroughly tested the changes and made sure that the installation commands fall back to `web` if the directory exist. It therefore does not matter when the Contao Manager is adjusted to use `public`.

did you also test this **with** the Contao Manager? e.g. if it is possible to manage an installation that uses the `public` dir?

leofeyer:
The Contao Manager does not yet support anything but `web`, because there are hardcoded references to the name e.g. here: https://github.com/contao/contao-manager/blob/8d86a863d585d5d202c9dec85cb29d977f896ab9/api/ApiKernel.php#L261-L262 We will have to adjust this at some point in the future.

aschempp:
Does that mean you cannot manage Contao 4.12 with the Contao Manager?

leofeyer:
Of course not. 😄 You just have to use `web` instead of `public`.

fritzmg:
A problem would arise if you install the Contao Manager _after_ installing Contao via `composer create-project`. But other than that it should be fine, since the Contao Manager requires you to create a web directory.

core-bundle/src/Command/SymlinksCommand.php
```diff
@@ -91,15 +91,23 @@ public function __construct(string $projectDir, string $uploadPath, string $logs
...
     {
         $this->webDir = $input->getArgument('target');
 
+        if (null === $this->webDir) {
+            if ((new Filesystem())->exists($this->projectDir.'/web')) {
+                $this->webDir = 'web'; // backwards compatibility
+            } else {
+                $this->webDir = 'public';
+            }
+        }
```
m-vo, line 109: Now testing this, I'm wondering why we do not just rely on the config value here? IMHO the config should be the single point of truth once configured.

m-vo:
I fear there is still something fishy with this change… :fish: 

In an 4.12 installation where I did not configure anything, `public` was used as a web root but the `contao.web_dir` config key/parameter was still pointing to `web`. This for instance broke the back end as the combiner was getting the wrong directory until I explicitly configured it to `%kernel.project_dir%/public`. 

I think there is another problem with detecting the old default by checking if the `web` directory exists: if you are using a deployment strategy that runs `contao-setup` on installation, no folders might exist beforehand. Your web root will now change. :cry: 

(https://contao.slack.com/archives/CK4J0KNDB/p1629567427484400)

leofeyer:
It does work like a charm on contao.org, therefore I suggest you debug the problem and open a new issue if it turns out to be a Contao problem.

fritzmg:
I am also unable to confirm the problem. After running `composer create-project contao/managed-edition . 4.12.*` a `public` folder was created and `%contao.web_dir%` properly points to that directory.

fritzmg:
Ah I see what you mean now with deployments. That is indeed a problem, since this will break the live site after a deployment (because the deployment would not fail in that case). For our deployments it's not a problem since we usually have versioned files in the `web/` directory anyway (like `.htaccess` files or other public resources). But if you don't need to version anything in `web/` this will for sure break those deployments. Not sure how to solve this though?

m-vo:
> Not sure how to solve this though?

I think we can't auto detect the old state. Instead we IMHO should use a config key in the `composer.json` file that is read and that sets the config. Then we could safely introduce the change in a new version of `contao/managed-edition` without affecting old installations.

PR #3085, opened 2021-06-14
Do not disable search and cache in FAQ and newsletter readers

fritzmg:
In Contao 4.6.0 the reader modules where changed so that they do not dynamically change `$objPage->noSearch` and `$objPage->cache` anymore - which is not really necessary anyway, because a page that contains a reader but does not show anything should still be perfectly search- and cacheable (see https://github.com/contao/contao/commit/65b60a1130af2af4169e63ffac481490ef598765#diff-9daf81af7f0708c0faabb8cb8859d3a1032b6318dc1abbe8d58208a7a0cf2f44L59-R72 for example). 

Another change was to throw an exception if no calendar or archive was selected in the module but still included in the front end.

These changes were still absent in the FAQ and Newsletter reader modules. Probably, because cache tagging for FAQ and Newsletters was added much later in https://github.com/contao/contao/commit/103649e2971e34b8b956c0a645272d07f708a649 - but without the other changes to the respective readers.

On that note though: does it really make sense to pass the module ID as an error code? An error code should be unique to the error itself and should not depend on a random database ID?

https://github.com/contao/contao/blob/4bc1479dab9dce3bf461d46632f380b2958afbcb/news-bundle/src/Resources/contao/modules/ModuleNewsReader.php#L72-L75

### DIFF
--- a/faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php
+++ b/faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php
@@ -68,7 +68,7 @@ class ModuleFaqReader extends Module
 
 		if (empty($this->faq_categories) || !\is_array($this->faq_categories))
 		{
-			throw new InternalServerErrorException('The faq reader ID ' . $this->id . ' has no categories specified.');
+			throw new InternalServerErrorException('The FAQ reader ID ' . $this->id . ' has no categories specified.');
 		}
 
 		return parent::generate();
--- a/faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php
+++ b/faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Contao\CoreBundle\Exception\InternalServerErrorException;
 use Contao\CoreBundle\Exception\PageNotFoundException;
 use Patchwork\Utf8;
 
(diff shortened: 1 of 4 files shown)


Toflar (dismissed):

ausi (dismissed):

ausi:
> On that note though: does it really make sense to pass the module ID as an error code? An error code should be unique to the error itself and should not depend on a random database ID?

Agree, the ID should not be used as an error code.

Toflar:
> Agree, the ID should not be used as an error code.

I second that.

fritzmg:
I have now removed the error code for those exceptions.

Toflar (dismissed):

ausi (dismissed):

faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php
```diff
@@ -57,30 +58,17 @@ public function generate()
...
 		if (empty($this->faq_categories) || !\is_array($this->faq_categories))
 		{
-			/** @var PageModel $objPage */
-			global $objPage;
-
-			$objPage->noSearch = 1;
-			$objPage->cache = 0;
-
-			return '';
+			throw new InternalServerErrorException('The faq reader ID ' . $this->id . ' has no categories specified.');
```
leofeyer:
```suggestion
			throw new InternalServerErrorException('The FAQ reader ID ' . $this->id . ' has no categories specified.');
```

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3086, opened 2021-06-14
Fix wrong dependency name for PurgeExpiredDataCron

fritzmg:
Current `4.9` branch has the following error:

```
Script Contao\ManagerBundle\Composer\ScriptHandler::initializeApplication handling the post-update-cmd event terminated with an exception
                                                                                                                     
  [RuntimeException]                                                                                                  
  An error occurred while executing the "contao:install-web-dir" command:                                             
  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:

    The service "Contao\CoreBundle\Cron\PurgeExpiredDataCron" has a dependency on a non-existent service "connection".
```


leofeyer:
Thank you @fritzmg.

PR #3088, opened 2021-06-14
Add a guests group to the list of allowed groups

### LINKED ISSUE #1725: Show elements to guests if "Protect element" group is assigned
**Affected version(s)**
Contao 4.4 & 4.9

**Description**
If a member group is assigned to a content element with "Protect element/article/page/..." and the option "Show to guests only" is also selected, the element is no longer visible to guests. 

It may sound illogical, but it can happen that you only want to show a certain element to all member groups and guests, but you want to exclude one specific member group.

So, the best ways I think, is to make it possible to use both option at the time, so the element is visible to guests also, if a group is assigned and the "Guests only" option is used.

It still can be achieved to show the elements only guests by not choosing a group in the "Protect element" option.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Implements #1725
| Docs PR or issue | TODO

### TODO

* [x] Migration


aschempp:
I remeber we discussed this, but did we come up with that solution? Why not use the existing checkbox?

leofeyer:
Yes, we came up with this solution (see https://github.com/contao/contao/issues/1725#issuecomment-682024692). This will allow us to show an element/article/page to guests and certain member groups simultaneously. Right now it is either "show to guests only" or "show to certain member groups only".

ausi:
I’m not sure if using `"0"` for the guests group is a good idea as it is a falsy value for PHP. Is it possible to use `"-1"`?

leofeyer:
Yes, however, we would lose the `'foreignKey' => 'tl_member_group.name'` auto-relation if we wanted to add custom options other than `includeBlankOption`. Unless we add a new `extra_options` or `additional_options` mechanism (not sure if we want that).

ausi:
As the auto-relation seems to be only a single select statement https://github.com/contao/contao/blob/4e0c1204af9f39f203eb691f4fd0946e3bdc4833/core-bundle/src/Resources/contao/library/Contao/Widget.php#L1313-L1323 I think I would go with an `options_callaback` and use `-1` for the guests option.

leofeyer:
Changed in be5a1b8f3ca40defd95e17a6613cc24e83a4757b.

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -502,6 +502,7 @@
 			'filter'                  => true,
 			'inputType'               => 'checkbox',
 			'foreignKey'              => 'tl_member_group.name',
+			'options_callback'        => array('tl_page', 'getGroups'),
```
fritzmg: New callbacks should always be implemented as a proper service and not added to the legacy `tl_` classes.

fritzmg (changes requested):

core-bundle/src/Resources/contao/dca/tl_page.php
leofeyer: Again, I really don't see any (or very little) sense in this rule. Having 40 callbacks in the DCA file and 2 in a listener just shatters the code and makes it less comprehensive. IMHO, the rule should be:

* If it is a new DCA file, add the callbacks as event listeners.
* If it is an existing DCA file, add the callbacks there. It most likely already contains dozens of callbacks, so one more does not make any difference.
* If you move callbacks of existing DCA files to event listeners, move them **all at once** and not just selectively.

leofeyer: Nevertheless, changed in f9c8f3ee87897f7159f37dc68d96d8b1171c31e9.

leofeyer:
@contao/developers This PR implies a behavior change, as `tl_page.guests` was not inherited but `tl_page.groups` is. This means that if I mark a page as "guests only", it will also include all its subpages in the future. Is this a problem?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -695,7 +734,7 @@ public static function findPublishedRegularWithoutGuestsByIds($arrIds, array $ar
 
 		if (System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
 		{
-			$arrColumns[] = "$t.guests=''";
+			$arrColumns[] = "$t.guests=''"; # FIXME
```
leofeyer: @contao/developers Any idea how to fix this? We would have to call `->loadDetails()` to know if the page or one of its parent pages is visible to guests only.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -712,22 +751,54 @@ public static function findPublishedRegularWithoutGuestsByIds($arrIds, array $ar
...
 	{
+		trigger_deprecation('contao/core-bundle', '4.12', 'Using PageModel::findPublishedRegularWithoutGuestsByIds() has been deprecated and will no longer work in Contao 5.0. Use PageModel::findPublishedRegularByPid() instead.');
+
 		$t = static::$strTable;
 		$arrColumns = array("$t.pid=? AND $t.type!='root' AND $t.type!='error_401' AND $t.type!='error_403' AND $t.type!='error_404'");
 
 		if (System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
 		{
-			$arrColumns[] = "$t.guests=''";
+			$arrColumns[] = "$t.guests=''"; # FIXME
```
leofeyer: Same here.

core-bundle/src/Resources/contao/modules/ModuleQuicklink.php
```diff
@@ -72,19 +72,41 @@ protected function compile()
...
+		{
+			$user = FrontendUser::getInstance();
+		}
+
 		$items = array();
 
 		/** @var PageModel[] $objPages */
 		foreach ($objPages as $objSubpage)
 		{
+			if ($objSubpage->protected)
```
ausi: Do we have to call `$objSubpage->loadDetails()` before this line?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -713,22 +752,54 @@ public static function findPublishedRegularWithoutGuestsByIds($arrIds, array $ar
...
 	 * @param array   $arrOptions An optional options array
 	 *
 	 * @return Collection|PageModel[]|PageModel|null A collection of models or null if there are no pages
+	 *
+	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5;
+	 *             use PageModel::findPublishedRegularByPid() instead.
 	 */
 	public static function findPublishedRegularWithoutGuestsByPid($intPid, array $arrOptions=array())
 	{
+		trigger_deprecation('contao/core-bundle', '4.12', 'Using PageModel::findPublishedRegularWithoutGuestsByIds() has been deprecated and will no longer work in Contao 5.0. Use PageModel::findPublishedRegularByPid() instead.');
```
```suggestion
		trigger_deprecation('contao/core-bundle', '4.12', 'Using PageModel::findPublishedRegularWithoutGuestsByPid() has been deprecated and will no longer work in Contao 5.0. Use PageModel::findPublishedRegularByPid() instead.');
```

core-bundle/src/Resources/contao/modules/ModuleQuicknav.php
```diff
@@ -126,16 +124,16 @@ protected function getQuicknavPages($pid, $level=1, $host=null)
 
 		foreach ($objSubpages as $objSubpage)
 		{
-			$_groups = StringUtil::deserialize($objSubpage->groups);
```
ausi, line 129: Same here, I think a call to `$objSubpage->loadDetails()` is missing.

core-bundle/src/Resources/contao/models/PageModel.php
ausi: I think calling `->loadDetails()` and checking for the `-1` group would be the correct solution here.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -313,8 +312,10 @@ protected function renderNavigation($pid, $level=1, $host=null, $language=null)
 				$objSubpage->domain = $host;
 			}
 
+			$groups = StringUtil::deserialize($objSubpage->groups, true);
+
 			// Do not show protected pages unless a front end user is logged in
-			if (!$objSubpage->protected || $this->showProtected || ($this instanceof ModuleSitemap && $objSubpage->sitemap == 'map_always') || ($user && $user->isMemberOf(StringUtil::deserialize($objSubpage->groups))))
+			if (!$objSubpage->protected || $this->showProtected || ($this instanceof ModuleSitemap && $objSubpage->sitemap == 'map_always') || (!$user && \in_array(-1, $groups)) || ($user && $user->isMemberOf($groups)))
```
ausi: Same here, I don’t understand how this could have worked before without calling `->loadDetails()` 🤔

ausi (commented):
I’m most confused by the missing `->loadDetails()` calls. If I understand it correctly, everywhere we access `$pageModel->protected` or `$pageModel->groups` we have to `loadDetails()` first.

aschempp:
Honestly, I don't like and understand this change. What if I create a member group called "Guests"? What's the point in changing all of this just for the sake of moving a checkbox from one place to another?

core-bundle/src/Resources/contao/modules/Module.php
leofeyer: Me neither. Could this be a bug that already exists in Contao 4.9?

fritzmg:
> What's the point in changing all of this just for the sake of moving a checkbox from one place to another?

@aschempp it's not about moving the checkbox. As already explained, currently you cannot show an element to guests _and_ to certain member groups. This was the solution we discussed last time.

leofeyer:
See #1725 for the discussion. It is not about removing the checkbox, it is about allowing combinations such as "show this page to guests and to group … only".

core-bundle/src/Resources/contao/modules/ModuleQuicknav.php
leofeyer: Added in 511b137ff74ec28f86d9e59ee13cd348db5fd493.

core-bundle/src/Resources/contao/modules/Module.php
leofeyer: I have added the `loadDetails()` call in 511b137ff74ec28f86d9e59ee13cd348db5fd493. We should still check Contao 4.9 and 4.11.

core-bundle/src/Resources/contao/modules/ModuleQuicklink.php
leofeyer: Added in 511b137ff74ec28f86d9e59ee13cd348db5fd493.

leofeyer:
I have added the missing `loadDetails()` checks in 511b137ff74ec28f86d9e59ee13cd348db5fd493. The quicklink module still behaves differently than all the others, but this will be addressed in #3096.

core-bundle/src/Resources/contao/models/PageModel.php
leofeyer: Fixed in 4de28acf3526481443d8ee6960eb1e0611f01e2f.

core-bundle/src/Resources/contao/models/PageModel.php
leofeyer: Fixed in 4de28acf3526481443d8ee6960eb1e0611f01e2f.

leofeyer:
The PR is ready to merge now.

core-bundle/src/Migration/Version412/MergeGuestsOnlyMigration.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        return true;
+    }
+
+    public function runTable(string $table): void
+    {
+        $rows = $this->connection->fetchAllAssociative("
+            SELECT
+                id, `groups`
```
ausi:
```suggestion
                id, IF(protected = '1', `groups`, NULL) AS `groups`
```

I think we have to ignore the value of `groups` if `protected` is not enabled.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -551,9 +551,9 @@ protected static function getPublishedSubpagesWithoutGuestsByPid($intPid, $blnSh
 		return array_filter(array_map(
 			static function (array $row) use ($blnFeUserLoggedIn): ?array
 			{
-				$page = PageModel::findWithDetails($row['id']);
+				$page = PageModel::findByPk($row['id']);
 
-				if ($blnFeUserLoggedIn && $page->protected && \in_array(-1, StringUtil::deserialize($page->groups)))
+				if (!$blnFeUserLoggedIn && $page->loadDetails()->protected && \in_array(-1, StringUtil::deserialize($page->groups)))
```
ausi:
```suggestion
				if ($blnFeUserLoggedIn && $page->loadDetails()->protected && \in_array(-1, StringUtil::deserialize($page->groups)))
```

The page should only be skipped **if** a user is logged in and the page is only for guests I think.

core-bundle/src/Resources/contao/modules/ModuleQuicknav.php
```diff
@@ -130,6 +130,7 @@ protected function getQuicknavPages($pid, $level=1, $host=null)
 				$objSubpage->domain = $host;
 			}
 
+			$objSubpage->loadDetails();
```
ausi: `->domain` gets overwritten by `loadDetails()` so we have to move this before line 127 IMO.

leofeyer: Good catch. 👍🏻 Fixed in 21f340966cd0f2c06d649b8f6ce7b94e2eee3faa.

ausi (approved):

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -718,23 +718,15 @@ public static function isVisibleElement(Model $objElement)
 		$blnReturn = true;
 
 		// Only apply the restrictions in the front end
-		if (TL_MODE == 'FE')
+		if (TL_MODE == 'FE' && $objElement->protected)
 		{
+			$blnReturn = false;
 			$blnFeUserLoggedIn = System::getContainer()->get('contao.security.token_checker')->hasFrontendUser();
+			$groups = StringUtil::deserialize($objElement->groups);
```
ausi, line 725:
```suggestion
			$groups = StringUtil::deserialize($objElement->groups, true);
```
Otherwise the `in_array` generates a warning.

leofeyer: Good catch. 👍🏻

leofeyer: See #3136.

tsarma:
I think this is not logical, something protected and something openly accessible should be mutually exclusive.

The argument for the use case, why this is implemented as far as I've understood, is that, say we have two groups A and B. And some content Y which is only allowed to view by guests and those in group A only, should not be seen by a person in group B. But the person in group B can also view content Y as a guest.

The use case may be legitimate but it surely is an edge case and the benefit is minimum for the confusion.

PR #3090, opened 2021-06-15
Allow empty content element translation key

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Fixes `Warning: Undefined array key "foo"` if a new `foo` content element controller was created but no translation added yet.

The existing check (elvis) only checks for falsy values but not if the array key actually exists.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -1241,7 +1241,7 @@ class tl_content extends Backend
 	public function addCteType($arrRow)
 	{
 		$key = $arrRow['invisible'] ? 'unpublished' : 'published';
-		$type = ($GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? null) ?: '&nbsp;';
+		$type = $GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? $arrRow['type'];
 		$class = 'limit_height';
 
 		// Remove the class if it is a wrapper element
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -1241,7 +1241,7 @@ class tl_content extends Backend
 	public function addCteType($arrRow)
 	{
 		$key = $arrRow['invisible'] ? 'unpublished' : 'published';
-		$type = $GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?: '&nbsp;';
+		$type = ($GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? null) ?: '&nbsp;';
 		$class = 'limit_height';
 
 		// Remove the class if it is a wrapper element


ausi (dismissed):

aschempp (approved):

fritzmg (approved):

aschempp (changes requested):
I would suggest to use actually fix this to use the actual type string if not label is available?

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -1241,7 +1241,7 @@ public function showJsLibraryHint($dc)
 	public function addCteType($arrRow)
 	{
 		$key = $arrRow['invisible'] ? 'unpublished' : 'published';
-		$type = $GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?: '&nbsp;';
+		$type = ($GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? null) ?: '&nbsp;';
```
fritzmg: I agree with @aschempp 

```suggestion
		$type = ($GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? $arrRow['type']) ?: '&nbsp;';
````
```

fritzmg (changes requested):

core-bundle/src/Resources/contao/dca/tl_content.php
aschempp: `$type = $GLOBALS['TL_LANG']['CTE'][$arrRow['type']][0] ?? $arrRow['type'];` should be enough?

fritzmg: True, yes. I wonder why it wasn't done like that before though?

leofeyer: There was no fallback to `$arrRow['type']` previously, was there? Also, `$arrRow['type']` could theoretically be empty.

aschempp: > There was no fallback to $arrRow['type'] previously, was there?

correct, but it would make sense, and since this is only a label it shouldn't be a problem.

> Also, $arrRow['type'] could theoretically be empty.

Very unlikely, but yes, it would result in an empty string. You mean we should `??` on that property as well?

leofeyer: I don't know, I was just trying to answer @fritzmg's question. 😄

m-vo: How could `$arrRow['type']` be empty/not in the array? 

So do we agree that we want to fall back to the type if not set?

aschempp: I don't think it will ever be empty. It _could_ be empty if you change the default type of your database and add an empty option. But the key would still be there, so I don't think we need to handle that case 😆

m-vo: adjusted in f97349c52

aschempp (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #3091, opened 2021-06-16
Fix the model properties and replace deprecated methods

leofeyer:
This PR fixes a lot of deprecations and "unnecessary type cast" false-positives as well as a couple of other issues found by the PhpStorm code inspector.


manager-bundle/src/Resources/skeleton/config/services.php
```diff
@@ -89,7 +89,6 @@
             ) {
                 $container->removeDefinition($id);
                 --$serviceCount;
-                continue(2);
```
ausi, line 92: Isn’t this a behavior change?

leofeyer: Fixed in #3094

PR #3092, opened 2021-06-16
Remove redundant arguments from private methods

leofeyer:
For example

```php
class FooTest extends TestCase
{
    public function testFoo(): void
    {
        $this->expectArray(['foo', 'bar']);
    }

    public function testBar(): void
    {
        $this->expectArray(['foo', 'bar']);
    }

    private function expectArray(array $arr): void
    {
        // $arr is always ['foo', 'bar'] therefore it does not make sense to make it a variable
    }
}
```

or

```php
class FooTest extends TestCase
{
    public function testFoo(): void
    {
        $this->getMock();
    }

    public function testBar(): void
    {
        $this->getMock();
    }

    private function getMock(boolean $hasResult = true): MockObject
    {
        $mock = $this->generateMock(/* … */);
        $mock
            ->expects($this->once())
            ->method(/* … */)
            ->willReturn($hasResult ? $result : null); // $hasResult is always true therefore the argument makes no sense
    }
}
```

### DIFF
--- a/core-bundle/src/Command/UserCreateCommand.php
+++ b/core-bundle/src/Command/UserCreateCommand.php
@@ -165,12 +165,7 @@ class UserCreateCommand extends Command
         }
 
         if (false === $input->getOption('admin') && ($options = $this->getGroups()) && 0 !== \count($options)) {
-            $answer = $this->askMultipleChoice(
-                'Assign which groups to the user (select multiple comma-separated)?',
-                $options,
-                $input,
-                $output
-            );
+            $answer = $this->askMultipleChoice($options, $input, $output);
 
             $input->setOption('group', array_values(array_intersect_key(array_flip($options), array_flip($answer))));
         }
@@ -245,9 +240,9 @@ class UserCreateCommand extends Command
         return $helper->ask($input, $output, $question);
     }
 
-    private function askMultipleChoice(string $label, array $options, InputInterface $input, OutputInterface $output): array
+    private function askMultipleChoice(array $options, InputInterface $input, OutputInterface $output): array
     {
-        $question = new ChoiceQuestion($label, $options);
+        $question = new ChoiceQuestion('Assign which groups to the user (select multiple comma-separated)?', $options);
         $question->setAutocompleterValues($options);
         $question->setMultiselect(true);
 
--- a/core-bundle/src/Command/UserCreateCommand.php
+++ b/core-bundle/src/Command/UserCreateCommand.php
@@ -165,7 +165,12 @@ class UserCreateCommand extends Command
         }
 
         if (false === $input->getOption('admin') && ($options = $this->getGroups()) && 0 !== \count($options)) {
-            $answer = $this->askMultipleChoice($options, $input, $output);
+            $answer = $this->askMultipleChoice(
+                'Assign which groups to the user (select multiple comma-separated)?',
+                $options,
+                $input,
+                $output
+            );
 
             $input->setOption('group', array_values(array_intersect_key(array_flip($options), array_flip($answer))));
         }
@@ -240,9 +245,9 @@ class UserCreateCommand extends Command
         return $helper->ask($input, $output, $question);
     }
 
-    private function askMultipleChoice(array $options, InputInterface $input, OutputInterface $output): array
+    private function askMultipleChoice(string $label, array $options, InputInterface $input, OutputInterface $output): array
     {
-        $question = new ChoiceQuestion('Assign which groups to the user (select multiple comma-separated)?', $options);
+        $question = new ChoiceQuestion($label, $options);
         $question->setAutocompleterValues($options);
         $question->setMultiselect(true);
 
--- a/core-bundle/tests/Controller/BackendPreviewSwitchControllerTest.php
+++ b/core-bundle/tests/Controller/BackendPreviewSwitchControllerTest.php
@@ -170,17 +170,17 @@ class BackendPreviewSwitchControllerTest extends TestCase
     /**
      * @return TokenChecker&MockObject
      */
-    private function mockTokenChecker(): TokenChecker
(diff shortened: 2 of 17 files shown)


core-bundle/src/Command/UserCreateCommand.php
```diff
@@ -165,12 +165,7 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
         }
 
         if (false === $input->getOption('admin') && ($options = $this->getGroups()) && 0 !== \count($options)) {
-            $answer = $this->askMultipleChoice(
-                'Assign which groups to the user (select multiple comma-separated)?',
-                $options,
-                $input,
-                $output
-            );
+            $answer = $this->askMultipleChoice($options, $input, $output);
```
fritzmg: If this command needs another multiple choice question in the future, this change basically has to be reverted again. Otherwise the method needs to be renamed to `askUserGroup`, since this method is now specifically _only_ to ask about the user groups after this change.

fritzmg (changes requested):

core-bundle/tests/Controller/BackendPreviewSwitchControllerTest.php
```diff
@@ -170,17 +170,17 @@ private function mockRouter(): RouterInterface
     /**
      * @return TokenChecker&MockObject
      */
-    private function mockTokenChecker(?string $frontendUsername = null, bool $previewMode = true): TokenChecker
```
aschempp: Feels like the fact that these arguments are not used means the test is incomplete. I would suggest to revert that and create a ticket/PR to improve the test.

core-bundle/tests/Picker/TablePickerProviderTest.php
```diff
@@ -332,7 +332,7 @@ public function testGetUrlWithPtableAndMultipleTables(): void
         $provider = $this->createTableProvider(
             $this->mockFrameworkWithDcaLoader('tl_article'),
             $this->mockRouterWithExpectedParams($params),
-            $this->mockConnectionForQuery('tl_article', 15, ['pid' => 1])
```
aschempp, line 335: if this argument is not passed, it's not very obvious/not tested that the `mockPickerConfig` and `mockConnectionForQuery` expect the same ID. Removing the argument would not make code more obvious, even though I agree the argument might be kinda unnecessary. But maybe we should use a class constant instead?

core-bundle/src/Command/UserCreateCommand.php
leofeyer: > If this command needs another multiple choice question in the future, this change basically has to be reverted again.

Yes. Is this a problem?

core-bundle/tests/Controller/BackendPreviewSwitchControllerTest.php
leofeyer: > means the test is incomplete

No, it does not. It might as well be a left-over from an older version in which the arguments were used (although it does not seem like that here). Maybe @richardhj can shed some light on this?

core-bundle/tests/Picker/TablePickerProviderTest.php
leofeyer: Yes, we could do this, however, that is not what this PR is about.

core-bundle/src/Command/UserCreateCommand.php
fritzmg: No, but then I vote for changing the name of the method to reflect what it actually does.

core-bundle/tests/EventListener/DataContainer/ContentCompositionListenerTest.php
```diff
@@ -814,7 +814,7 @@ public function testCannotPasteAfterArticleIfPageIsNotFound(): void
 
     public function testCannotPasteAfterArticleIfProviderDoesNotSupportContentComposition(): void
     {
-        $page = $this->expectPageFindByPk(17, $this->pageRecord);
+        $page = $this->expectPageFindByPk($this->pageRecord);
```
ausi:
```suggestion
        $page = $this->expectPageFindByPk();
```

`$this->pageRecord` can be omitted too I think.

core-bundle/src/Command/UserCreateCommand.php
leofeyer: Reverted in 3fa8d4e0e56b089d8476fab4976d0d13729d7f2d.

core-bundle/tests/EventListener/DataContainer/ContentCompositionListenerTest.php
leofeyer: Changed in 3fa8d4e0e56b089d8476fab4976d0d13729d7f2d.

core-bundle/tests/Picker/TablePickerProviderTest.php
leofeyer: I have reverted the changes in 3fa8d4e0e56b089d8476fab4976d0d13729d7f2d and used arbitrary IDs instead of `15` everywhere.

ausi (dismissed):

fritzmg (dismissed):

ausi (dismissed):

core-bundle/tests/Controller/BackendPreviewSwitchControllerTest.php
aschempp: well it obviously does not test what the controller does if there is a username or `isPreviewMode` is not `true`?

richardhj: Yep, looks like a test for `isPreviewMode = false` is missing, if that's a necessary case at all.

leofeyer: Reverted in f1facf7dfe50eee127db070361108801325835ef.

PR #3093, opened 2021-06-17
changed visibility of PaletteManipulator::explode() to public

Defcon0:
According to slack discussion with @aschempp, @Toflar and @m-vo ;-)

### DIFF
--- a/core-bundle/src/DataContainer/PaletteManipulator.php
+++ b/core-bundle/src/DataContainer/PaletteManipulator.php
@@ -158,6 +158,38 @@ class PaletteManipulator
     }
 
     /**
+     * Having the same field in multiple legends is not supported by Contao, so we don't handle that case.
+     *
+     * @return string|false
+     */
+    public function findLegendForField(array $config, string $field)
+    {
+        foreach ($config as $legend => $group) {
+            if (\in_array($field, $group['fields'], true)) {
+                return $legend;
+            }
+        }
+
+        return false;
+    }
+
+    public static function convertConfigToPaletteString(array $config): string
+    {
+        return (new self())->implode($config);
+    }
+
+    public static function isFieldInPalette(string $field, string $palette): bool
+    {
+        foreach ((new self())->explode($palette) as $data) {
+            if (\in_array($field, $data['fields'], true)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
      * @throws PalettePositionException
      */
     private function validatePosition(string $position): void
@@ -179,7 +211,7 @@ class PaletteManipulator
      *
      * @return array<int|string, array<mixed>>
      */
-    public function explode(string $palette): array
+    private function explode(string $palette): array
     {
         if ('' === $palette) {
             return [];
@@ -369,22 +401,6 @@ class PaletteManipulator
                 $config[$legend]['fields'] = array_diff($group['fields'], $remove['fields']);
             }
         }
-    }
-
-    /**
-     * Having the same field in multiple legends is not supported by Contao, so we don't handle that case.
-     *
-     * @return string|false
-     */
-    private function findLegendForField(array $config, string $field)
-    {
-        foreach ($config as $legend => $group) {
-            if (\in_array($field, $group['fields'], true)) {
-                return $legend;
-            }
-        }
-
-        return false;
     }
 
     private function canApplyToParent(array &$config, array $action, string $key, string $position): bool
--- a/core-bundle/src/DataContainer/PaletteManipulator.php
+++ b/core-bundle/src/DataContainer/PaletteManipulator.php
@@ -179,7 +179,7 @@ class PaletteManipulator
      *
      * @return array<int|string, array<mixed>>
      */
-    private function explode(string $palette): array
+    public function explode(string $palette): array
     {
         if ('' === $palette) {
             return [];
--- a/core-bundle/tests/DataContainer/PaletteManipulatorTest.php
+++ b/core-bundle/tests/DataContainer/PaletteManipulatorTest.php
@@ -462,4 +462,36 @@ class PaletteManipulatorTest extends TestCase
             $pm->applyToString('{contact_legend},firstname,lastname')
         );
     }
+
+    public function testExplodesAPalette(): void
+    {
+        $pm = PaletteManipulator::create();
+
+        $this->assertSame(
+            [
+                'contact_legend' => [
+                    'fields' => [
+                        0 => 'title',
+                        1 => 'lastname'
+                    ],
+                    'hide' => false
+                ],
+                'foo_legend' => [
+                    'fields' => [
+                        0 => 'field1',
+                        1 => 'field2'
+                    ],
+                    'hide' => true
+                ],
+                0 => [
+                    'fields' => [
+                        0 => 'field3',
+                        1 => 'field4'
+                    ],
+                    'hide' => false
+                ]
+            ],
+            $pm->explode('{contact_legend},title,lastname;{foo_legend:hide},field1,field2;field3,field4')
+        );
+    }
 }
--- a/core-bundle/tests/DataContainer/PaletteManipulatorTest.php
+++ b/core-bundle/tests/DataContainer/PaletteManipulatorTest.php
@@ -472,24 +472,24 @@ class PaletteManipulatorTest extends TestCase
                 'contact_legend' => [
                     'fields' => [
                         0 => 'title',
-                        1 => 'lastname'
+                        1 => 'lastname',
                     ],
-                    'hide' => false
+                    'hide' => false,
                 ],
                 'foo_legend' => [
                     'fields' => [
                         0 => 'field1',
-                        1 => 'field2'
+                        1 => 'field2',
                     ],
-                    'hide' => true
+                    'hide' => true,
                 ],
                 0 => [
                     'fields' => [
                         0 => 'field3',
-                        1 => 'field4'
+                        1 => 'field4',
                     ],
-                    'hide' => false
-                ]
+                    'hide' => false,
+                ],
             ],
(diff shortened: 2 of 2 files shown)


fritzmg:
Why only explode and not also implode?

Defcon0:
After thinking it over again, I didn't quite get, why we could need `implode()`. I needed explode in order to, well, explode a palette :-) Imploding can always be accomplished by using the methods already public, I guess.

Nevertheless I could also turn it public. Are there any other opinions on that?

After committing and fixing a codestyle issue, the CI still complains about code style but it's not code of mine? Seems to be related with the order of the methods in the test class. But I can't see any order there like alphabetically and can't find the php code style fixer config file in the repo?

m-vo:
> After committing and fixing a codestyle issue, the CI still complains about code style but it's not code of mine?

It is :smile:. The `OrderedClassElementsFixer` wants the public / private methods sorted. Just run with `--fix` and it will resolve itself.

Toflar (changes requested):
I don't think it makes sense to make `explode` public, to be honest. It doesn't say anything, the API as a user of the `PaletteManipulator` would have no meaning.
Methods should be self explaining :-)
As far as I can remember your use case was to check whether `title` is in a palette which is difficult now if your palette looks like `titleA,title,titleFoo` etc.
Why not adding a new method `isFieldInPalette(string $field, string $palette): bool` or so?
It would be super easy to understand and the complexity is kept internal in the class.

aschempp:
That should however be a static method then. The PaletteManipulator is supposed to apply changes to a palette, checking something is not a use case for the instance. 

On the other hand, are you using the check to only add/manipulate if a field is present? Maybe we should add something like `->ifNotHasField()`?

Defcon0:
@aschempp My use case is https://github.com/heimrichhannot/contao-multilingual-fields-bundle . I can mark certain fields as translatable and then have a button to start the language edit mode in an arbitrary entity. In this language edit mode you have the translated field as readonly (in order to have a reference for your translation), a checkbox "en_translate_field123" and the translation field which is basically a copy of the original translated field but the name is prefixed with the destination language. See the README.md for an image visualizing this.

Now in the edit mode in the first draft of the bundle I simply added all translatable fields one after another. But this makes no sense for example in `tl_content`'s text element. Let's say we can translate `text` and `linkTitle`. Both fields belong to different content element palettes. So in the language edit mode I need to check if the translatable field is in the current palette (`$dc->getPalettte()`) and for this I needed `PaletteManipulator::explode()`.

If someone wonders, why we're not using DC_Multilingual for that: The difference to DC_Multilingual is that all translation data is stored in the original record.

@Toflar This would also be ok for me :-)

fritzmg:
> After thinking it over again, I didn't quite get, why we could need `implode()`

I would use it to first explode the palette, then make some custom checks and alterations and then to implode it again to generate the new palette. Having explode public without implode makes no sense to me at least.

Defcon0:
OK, so I would do the following functions:

```
public static function isFieldInPalette(string $field, string $palette): bool {
// use explode()
}
```

and make `implode()` public. Is everone ok with that?

ausi:
> and make `implode()` public. Is everone ok with that?

If we want to make this public, it should be renamed to something like `convertConfigToPaletteString()` IMO.

fritzmg:
Well, no, I think the point would be to either make explode and implode public - or to implement a new function according to your use case ;). I think an `isFieldInPalette` function makes sense in any case. Though may be it should be static?

Toflar:
New methods, the rest is definitely internal.

Defcon0:
OK, then we have:

```
public static function isFieldInPalette(string $field, string $palette): bool {
// use explode()
}
```

and

```
public static function convertConfigToPaletteString(array $config): string {
// use implode()
}
```

OK?

ausi:
And maybe a counter part to `convertConfigToPaletteString()`?  
Something like like `convertPaletteToConfigArray(string $palette): array { /* use explode() */ }`

Defcon0:
@ausi Let's see what the others say.

I developed further with the bundle I need the changes for. In fact I also needed `findLegendForField()` so I would change the visibility of that one also to `public`. The name describes it I guess. I can't think of a name for a new function calling this one so I would just change the visibility.

Is everyone ok with that?

I guess then we're ok and I can continue with the PR.

aschempp:
The palette manipulator works like this:

1. add a field to a position or
2. remove a field
3. then decide what palette(s) this should be applied to

If I understand your use case correctly, you need to do (1) or (2) depending on whether a field exists? So why would you need `implode` or `applyConfig` at all? Shouldn't it be something like:

```
PaletteManipulator::create()
    ->addField('foo', 'bar')
    ->ifFieldExists('bar')
    ->applyToPalette('text', 'tl_content')
;
```

OR 

```
PaletteManipulator::create()
    ->addField('foo', 'some_legend', PaletteManipulator::APPEND)
    ->ifLegendExists('some_legend')
    ->applyToPalette('text', 'tl_content')
;
```

Defcon0:
I added `implode` to the list because of @fritzmg 's comment. I don't need it personally, yet. But I agree that the function is also very useful and shouldn't be reserved for the core only.

In addition to explode, I also needed `findLegendForField` in my code (I experienced that later on).

You can take a look on how I used the methods (simply search for "palettemanipulator" and you find all occurrences):

https://github.com/heimrichhannot/contao-multilingual-fields-bundle/blob/master/src/EventListener/Contao/LoadDataContainerListener.php

Defcon0:
OK, I added the new functions as discussed above. Are you ok with the changes?

aschempp:
@Defcon0 I thoroughly reviewed your bundle again. I really like the general idea of how to translate fields, because it is technically simple and works with the standard driver, models and queries 👍 

As far as I did understand, you are dynamically adding fields _after_ the ones that should be translated (as described in https://github.com/contao/contao/pull/3093#issuecomment-863014049)

> So in the language edit mode I need to check if the translatable field is in the current palette ($dc->getPalettte()) and for this I needed PaletteManipulator::explode().

I did "find" `$dc->getPalette()` as well. Similar to https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Resources/contao/drivers/DC_Table.php#L2355, shouldn't checking that be as simple as this?

```
in_array($field, StringUtil::trimsplit('[;,]', $dc->getPalette()))
```

I see that you then basically re-build the whole palette by adding the original fields and multilingual fields. However, wouldn't that simply be what the PaletteManipulator does by using `addField` with reference to the original field?

Defcon0:
Could be, that this also works, but wouldn't be the new methods be helpful taking a step away from this use case, as well for other users?

aschempp:
> but wouldn't be the new methods be helpful taking a step away from this use case, as well for other users?

keeping internal methods private means we can change them at any time if necessary. As soon as they are public, they become official API that we need to support in our backwards compatibility promise. These methods clearly are internal logic to me, that's why I'm trying to find the best way to support your use case but also limit the maintenance burden on the core devs 😊

Defcon0:
OK, I just tested and

```
in_array($field, StringUtil::trimsplit('[;,]', $dc->getPalette()))
```

works. However, I also need the legend in the context of the current palette situation. I can create a function for that, clear, but in PaletteManipulator there is already one ;-)

I totally understand your argument concerning the burden of maintenance. It's ok if we don't, merge the PR, but I think this particular functions would be helpful to many devs and I guess wouldn't be changed so often. Hence the maintenance effort would not be too large (correct me if I'm wrong).

aschempp:
> However, I also need the legend in the context of the current palette situation. I can create a function for that, clear, but in PaletteManipulator there is already one ;-)

Can you please point me in that direction? What method exactly and what do you use it for?

Defcon0:
Sure:

https://github.com/heimrichhannot/contao-multilingual-fields-bundle/blob/master/src/EventListener/Contao/LoadDataContainerListener.php#L340

I need the method `findLegendForField`:

https://github.com/heimrichhannot/contao-multilingual-fields-bundle/blob/c4bc0853dd3284690ae47d13f3a136fa5b171cd2/src/EventListener/Contao/LoadDataContainerListener.php#L480

I need it to build a palette with the translatable fields in a palette but in their correct legend in the context of the current data container.

aschempp:
> I need it to build a palette with the translatable fields in a palette but in their correct legend in the context of the current data container.

```
PaletteManipulator::create
    ->addField('original_field', 'translated_checkbox,translated_field')
    ->applyToPalette('palettename', 'tl_table')
;
```

Defcon0:
I think you got me wrong (or me ... ^^).

Let's say we have the palettes:

```
'palette1' => '{some_legend},myField1,myField2';
'palette2' => '{some_other_legend},myField2,myField3';
```

Then we have the info that myField2 and myField3 are translatable. In consequence the palettes would be as follows:

```
'palette1' => '{some_legend},myField2,translate_myField2';
'palette2' => '{some_other_legend},myField2,translate_myField2,myField3,translate_myField3';
```

Mind the differing legend in the palettes. A field can appear in different legends depending on the context (dc). Also mind that myField1 is not in the palette anymore which I can't do using your proposed method I think.

aschempp:
FYI I updated your comment to include palette names for clarity (hope that was correct).

```php
PaletteManipulator::create()
    ->addField('translate_myField2', 'myField2')
    ->removeField('myField1')
    ->applyToPalette('palette1', 'tl_table')
;

PaletteManipulator::create()
    ->addField('translate_myField2', 'myField2')
    ->addField('translate_myField3', 'myField3')
    ->applyToPalette('palette2', 'tl_table')
;
```

Defcon0:
OK, it's been quite a headache, but now it works without the private methods ;-) Thanks for your support. Although I still think, the methods would be cool for the public, I'll close the PR now.

PR #3103, opened 2021-06-22
Allow to exclude fields in the file manager

### LINKED ISSUE #2847: Benutzergruppe: Rechteverwaltung: "Von der Synchronisation ausgenommene Ordner" ohne Effekt
**Affected version(s)**

Contao 4.9.12

**Description**

Obwohl ich in der Benutzergruppe die Synchronisation von Ordnern für den Redakteur ausgeschlossen habe, kann dieser trotzdem die Eigenschaft setzten.
Ich habe diesen Bug auf 3 verschiedenen Contao Systemen nachvollziehen können.
Danke für's beheben!

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Implements #2847
| Docs PR or issue | -

What is the best way to add the fields to `alexf` of the existing groups in a migration? Add it to all groups that allow access to the file manager? Or add it to all groups that have file mounts?


ausi:
Add it to all groups that have access to the file manager IMO.

How do we check if the migration needs to run?

leofeyer:
That would have been my next question. 🙈

Toflar:
Imho it would be a great feature to have a `tl_migrations` table or so which just allows to store a key for those cases.
It's always preferrable to have `shouldRun()` do the actual check but in cases where this is not possible, the fact whether it ran once or not could be stored there?

leofeyer:
As discussed with @ausi, there is currently no way to reliably detect whether these settings need to be migrated or already have been. The admin therefore needs to manually adjust the user groups after the update and we have to add a hint to the release announcement and prepare for some "my users cannot edit the file metadata in Contao 4.12 anymore" questions. 🙈

m-vo:
Why don't we go down the `tl_migration` route? IMO that should not be too hard to implement.

fritzmg:
I am also very much in favor of introducing a `tl_migration`(`s`?) table.

leofeyer:
We have got 8 days left to create new PRs for Contao 4.12. If you want to give it a shot, do it. 👍🏻

fritzmg:
I'll work on it :)

m-vo:
@fritzmg Ping me if you need support to get it done.

leofeyer:
If this should not work out, we could still do the following:

* Get all user groups that grant access to the file manager or have file mounts;
* check whether at least one of these groups has a `tl_files::` entry in `alexf`;
* if so, skip the migration. Otherwise run it.

I guess it does not make much sense to grant access to the file manager and not enable any metadata fields, does it?

leofeyer:
As discussed again with @ausi, here is the logic for the migration:

```php
public function shouldRun()
{
    $shouldRun = false;
    $groups = getAllGroups();

    foreach ($groups as $group) {
        if (preg_grep('(^tl_files::)', $groups->alexf)) {
            return false; // the migration ran already
        }

        if (grantsEditPermission($group)) {
            $shouldRun = true;
        }
    }

    return $shouldRun;
}

public function run()
{
    $groups = getAllGroups();

    foreach ($groups as $group) {
        if (grantsEditPermission($group)) {
            $group->alexf = ['tl_files::name', '…'];
            $group->save();
        }
    }
}

private function grantsEditPermission($group)
{
    return (in_array($group->modules, 'files') || !empty($group->filemounts)) && in_array($group->fop, 'f2');
}
```

(The code does not actually work, of course. It is just an outline to explain the logic.)

fritzmg:
`shouldRun` also needs to check if there are actually any groups with access to `files` and non-empty file mounts.

leofeyer:
Migration added in 6a1a4bed321572055f2c240391371982963a5a70.

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
```diff
@@ -0,0 +1,121 @@
...
+        $this->connection = $connection;
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+
+        if (!$schemaManager->tablesExist('tl_user_group')) {
+            return false;
+        }
```
ausi, line 41:
```suggestion
        if (!$schemaManager->tablesExist('tl_user_group')) {
            return false;
        }

        $columns = $schemaManager->listTableColumns('tl_user_group');

        if (!isset($columns['id'], $columns['modules'], $columns['filemounts'], $columns['fop'], $columns['alexf'])) {
            return false;
        }
```

Migrations should be as robust as possible IMO. Like if someone has one of the fields unset in the DCA, the migration should not run into an error.

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
```diff
@@ -0,0 +1,121 @@
...
+
```
ausi, line 42:
```suggestion

        $count = $this->connection->fetchOne("SELECT COUNT(*) FROM tl_user_group WHERE alexf LIKE '%tl_files::%'");

        if ($count > 0) {
            return false;
        }
```

If any `tl_files::` entry exists we should not run the migration IMO.

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
```diff
@@ -0,0 +1,121 @@
...
+        $shouldRun = false;
+
+        foreach ($groups as $group) {
+            if ($this->grantsEditPermission($group)) {
+                $shouldRun = true;
+            }
+        }
+
+        return $shouldRun;
```
ausi:
```suggestion
        foreach ($groups as $group) {
            if ($this->grantsEditPermission($group)) {
                return true;
            }
        }

        return false;
```

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
leofeyer: I guess we do not have to check for `$columns['id']`, because Contao will not work without this column. But otherwise agreed. 👍🏻

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
leofeyer: This check is already present on line 49 and the early return on line 53, isn't it?

ausi: I don’t think so. `0 === \count($groups)` is only true if every single row in `tl_user_group` has a `tl_files::`. But I think we should also skip the migration if a single row has  `tl_files::` in it because that means we are most likeley not in an update to 4.12.

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
leofeyer: Added in 9730741387614720e5e628f53c96dd94457db637.

core-bundle/src/Migration/Version412/AllowedExcludedFieldsMigration.php
leofeyer: Added in 9730741387614720e5e628f53c96dd94457db637.

ausi (approved):

Issue #3097, opened 2021-06-21
Getting rid of TL_NOINDEX_KEYS


Toflar:
Let's discuss how we can get rid of `$GLOBALS['TL_NOINDEX_KEYS']`.

ausi:
I think a solution could be related to a concept for managing `<link rel="canonical">` in Contao. Canonical

leofeyer:
As discussed at the developer's meeting, these are the steps to accomplish the mission:

- [x] Check if Escargot supports canonical URLs (implemented in https://github.com/terminal42/escargot/commit/5e4955739e40d04a8f00ece978283e32e4fadebb)
- [x] Make our indexer ignore pages if the canonical URL does not match the URL (see https://github.com/contao/contao/pull/3617)
- [x] Canonical URLs should be handled in the ResponseContext (with `*` wildcard support) (see https://github.com/contao/contao/pull/3616)
- [x] It should be possible to pass an individual URL to the ResponseContext (see https://github.com/contao/contao/pull/3616)
- [x] It should be possible to enter allowed query parameters and individual URLs in the page settings (see https://github.com/contao/contao/pull/3616)
- [ ] The list modules should have an "add pagination URLs to the search index" checkbox which adds the pagination query variable to the allow list
- [x] It should be possible to disable canonical URLs per root page (see https://github.com/contao/contao/pull/3616)
- [x] Canonical URLs should be disabled for existing pages after the update (see https://github.com/contao/contao/pull/3616)

Toflar:
I'm closing this because 

> The list modules should have an "add pagination URLs to the search index" checkbox which adds the pagination query variable to the allow list

is a totally optional feature and I would even argue, that we shouldn't implement it. The only advantage is that it would automatically add e.g. `page_n23` if you select the correct checkboxes. Which you can also do by configuring `page_n23` or even `page_n*` to your page settings as of now. It's not complicated at all and imho it's even better in the page settings as you have to look at only one place to see the `rel="canonical"` configuration instead of the page and several different module settings.

PR #3105, opened 2021-06-22
Prevent search for pages with alias "index"

eki89:
It's possible to call a page in the frontend in case of using "index" as alias. Actually this should not be possible, because it can lead to duplicate content.

The error is reproducable in Contao 4.4, 4.9 and 4.11.

In my opinion it should never be possible to search for the index alias. So it can be removed completely from collected candidates.

Example:
https://contao.org/de/index.html
https://contao.org/de/

### DIFF
--- a/core-bundle/src/Routing/Candidates/AbstractCandidates.php
+++ b/core-bundle/src/Routing/Candidates/AbstractCandidates.php
@@ -120,6 +120,10 @@ class AbstractCandidates implements CandidatesInterface
                     $withoutSuffix = substr($withoutPrefix, 0, -\strlen($suffix));
                 }
 
+                if ('index' === $withoutSuffix) {
+                    continue;
+                }
+
                 $this->addCandidatesFor($withoutSuffix, $candidates);
             }
         }


eki89:
The described bug is also present in version 4.9. The source code has changed with version 4.10, so I will submit another pull request for 4.9 as soon as it is clear that there is nothing wrong with the change.

From my point of view, the alias "index" must never be dissolved. This is finally used to achieve no further addition in the URL.

aschempp:
Unfortunately, this wont work. Since Contao pages can have any url parameter, the index page is necessary to use them. If you have a reader page on the index page, the resulting URL `example.com/index/somealias.html` wouldn't work anymore. 

However, I do agree its not ideal this URL is accessible. I wonder if it is better to "disallow" it or to add a redirect?

ausi:
> I wonder if it is better to "disallow" it or to add a redirect?

I think it should redirect.

eki89:
I can not confirm that a reader module on the index page would not work. I've tested it right now and everything works well. The change should only affect the identification of page candidates (just for routing), which should have no effect on reader modules.

I'm still testing the change now with an other suffix and the prepend local feature. But yes, a simple redirect should of course also work.

eki89:
Everything works as expected. I cannot detect any undesirable behavior. I've testet it with activated and deactivated legacy routing in all possible combinations.
- URL-Prefix = '' and URL-Suffix = ''
- URL-Prefix = 'de' and URL-Suffix = ''
- URL-Prefix = '' and URL-Suffix = '.htm'
- URL-Prefix = 'de' and URL-Suffix = '.htm'
I've tested all these posibilities with alias 'index', 'start' and 'index/test' with and without reader modules. I could not find any undesirable behavior.

leofeyer:
Isn't this the most basic principle of how web servers work? I doubt that Google rates this as duplicate content, because the same happens on google.com:

* https://www.google.com/
* https://www.google.com/index.html

We should not do anything about it IMHO.

fritzmg:
The URL with `index` (and no other fragments) will never be generated by Contao anyway.

eki89:
Very interesting that Google has the same behavior.

Yes, Contao never generates an index URL. But in case of a relaunch from an other CMS or static Website it's possible that these URLs are already indexed. There is also potential for abuse, as anyone is able to post a link to this duplicate page on the internet (like me in my first post). Because of my link above, Google will probably index the index.html, which will lead to duplicate content on the contao project page.

In my opinion, the "slash" at the end of a URL is removed by Contao for the same reason. I do not see why this is the most basic principle of how web servers work. ^^ By default appache's htaccess file searchs for an index.html or index.php file. But that is not related to the actual "problem".

fritzmg:
> But in case of a relaunch from an other CMS or static Website it's possible that these URLs are already indexed.

In that case you should create a custom redirect.

> There is also potential for abuse, as anyone is able to post a link to this duplicate page on the internet (like me in my first post). Because of my link above, Google will probably index the index.html, which will lead to duplicate content on the contao project page.

In what way can this be abused? Duplicate content like this does not matter.

eki89:
> > But in case of a relaunch from an other CMS or static Website it's possible that these URLs are already indexed.
> 
> In that case you should create a custom redirect.
> 
Yes, that's a possible and working solution.

> In what way can this be abused? Duplicate content like this does not matter.
Unfortunately, I can not answer this question certainty. This topic is quite controversial. In my opinion the question is whether the behavior is desired.

I can understand if the potential disadvantages outweigh the advantages. I've found the described behavior and thought that this is actually not desired.

leofeyer:
As discussed in Mumble on July 1st, we want to keep the current behavior and improve this case in the future with a canonical tag (which we need in other cases as well, see #3097).

@eki89 You might be able to solve your issue with https://github.com/christianbarkowsky/rel-canonical.

PR #3100, opened 2021-06-21
Use HTTPS for all schema.org links

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Implements #2436
| Docs PR or issue | -

PR #3107, opened 2021-06-22
Add the default WebPage schema.org data again

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | https://github.com/contao/contao/pull/3100#pullrequestreview-689561964
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
+++ b/core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
@@ -59,6 +59,7 @@ class CoreResponseContextFactory
         $context = $this->createResponseContext();
         $context->add($this->eventDispatcher);
         $context->addLazy(HtmlHeadBag::class);
+
         $context->addLazy(
             JsonLdManager::class,
             static function () use ($context) {
--- a/core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
+++ b/core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
@@ -18,6 +18,7 @@ use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
 use Contao\CoreBundle\Security\Authentication\Token\TokenChecker;
 use Contao\PageModel;
 use Contao\StringUtil;
+use Spatie\SchemaOrg\WebPage;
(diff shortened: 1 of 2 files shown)


core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -58,7 +59,10 @@ public function createWebpageResponseContext(): ResponseContext
         $context = $this->createResponseContext();
         $context->add($this->eventDispatcher);
         $context->addLazy(HtmlHeadBag::class);
-        $context->addLazy(JsonLdManager::class);
+
+        $jsonLdManager = new JsonLdManager($context);
+        $jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_ORG)->add(new WebPage());
```
Toflar: And now look how easy it would be to add additional data to this type. I'm already loving this new `JsonLdManager` 😊 But let's focus on migrating the existing stuff first.

m-vo: Is there a reason why this does not use `addLazy` with the closure for this?

Toflar: Because the service is built to add the webpage schema?

m-vo: Sorry, I'm sure I'm missing something. Wouldn't the resource only be used when first accessed if done this way:

```php
 $context->addLazy(JsonLdManager::class, static function() use ($context) {
     return (new JsonLdManager($context))
         ->getGraphForSchema(JsonLdManager::SCHEMA_ORG)
         ->add(new WebPage())
     ;
 });
```
What's the difference to the `HtmlHeadBag` where we're doing this?

Toflar: Yeah we could do that if we think it makes sense (which I don't :D).

Toflar: Still added it in cc21b7203deb5605334c0f9a373ffde510a38822. Your code does not work because it would return the `Graph` instead of the `JsonLdManager` ;)

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3108, opened 2021-06-22
Introduce recorded migrations

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/pull/3103#issuecomment-865977632
| Docs PR or issue | https://github.com/contao/docs/pull/782

This introduces a new `AbstractRecordedMigration` that extends the existing `AbstractMigration` and automatically keeps a record on whether the migration was executed or not in a `tl_migration` table, if you use the `createResult` method. In your own `shouldRun` you can either use `hasRun` or call `parent::shouldRun` which also checks for `hasRun`.

The `Migration` entity itself is similar to the `CronJob` entity and will only hold the name of the migration (its FQCN by default) and a timestamp when it was executed.

Your own migration will then look like this at a minimum:

```php
// src/Migration/ExampleMigration.php
namespace App\Migration;

use Contao\CoreBundle\Migration\AbstractRecordedMigration;
use Contao\CoreBundle\Migration\MigrationResult;

class ExampleMigration extends AbstractRecordedMigration
{
    public function run(): MigrationResult
    {
        return $this->createResult(true);
    }
}
```

Or, if you want to check for additional conditions (which will be the case for most):

```php
// src/Migration/ExampleMigration.php
namespace App\Migration;

use Contao\CoreBundle\Migration\AbstractRecordedMigration;
use Contao\CoreBundle\Migration\MigrationResult;

class ExampleMigration extends AbstractRecordedMigration
{
    public function shouldRun(): bool
    {
        if ($this->hasRun()) {
            return false;
        }

        return /* custom condition */;
    }

    public function run(): MigrationResult
    {
        return $this->createResult(true);
    }
}
```

or

```php
// src/Migration/ExampleMigration.php
namespace App\Migration;

use Contao\CoreBundle\Migration\AbstractRecordedMigration;
use Contao\CoreBundle\Migration\MigrationResult;

class ExampleMigration extends AbstractRecordedMigration
{
    public function shouldRun(): bool
    {
        // Custom condition

        return parent::shouldRun();
    }

    public function run(): MigrationResult
    {
        return $this->createResult(true);
    }
}
```


core-bundle/src/Entity/Migration.php
```diff
@@ -0,0 +1,67 @@
...
+use Doctrine\ORM\Mapping\GeneratedValue;
+
+/**
+ * @ORM\Table(
+ *     name="tl_migration",
+ *     indexes={
+ *         @ORM\Index(name="name", columns={"name"})
+ *     }
+ * )
+ * @ORM\Entity(repositoryClass="Contao\CoreBundle\Repository\CronJobRepository")
```
Toflar: Typo

fritzmg: well.. copy & paste 🙈

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,59 @@
...
+
+    protected function hasRun(): ?bool
+    {
+        if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
+            return null;
+        }
+
+        return null !== $this->migrationRepository()->findOneByName($this->getName());
+    }
+
```
m-vo, line 33: What about adding the default implementation that checks for `!hasRun` in `shouldRun` here as well?

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,59 @@
...
+use Contao\CoreBundle\Entity\Migration as MigrationEntity;;
+use Contao\CoreBundle\Repository\MigrationRepository;
+use Doctrine\DBAL\Connection;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Contracts\Service\ServiceSubscriberTrait;
+
+abstract class AbstractRecordedMigration extends AbstractMigration
+{
+    use ServiceSubscriberTrait;
```
m-vo, line 22: I wonder if we should rather use a static service subscriber method here. :thinking: 
We're currently not using this anywhere else in the code base, do we?

I've been using this pattern quite heavily lately but I'm not sure I really like it. It makes you introduce traits for everything and then you'll end up with a lot of magic and more named things that need to be managed.

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: You mean:

```php
abstract public function shouldRun(): bool
{
    return false !== $this->hasRun();
}
```

and in your own code:

```php
public function shouldRun(): bool
{
    // my conditions
    
    return parent::shouldRun();
}
```

?

aschempp: I wonder if we could simply replace the `shouldRun` method? And if you wanna extend it, you can call parent?

core-bundle/src/Repository/MigrationRepository.php
```diff
@@ -0,0 +1,38 @@
...
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\DBAL\Connection;
+use Symfony\Bridge\Doctrine\ManagerRegistry;
+
+/**
+ * @internal
+ *
+ * @method object|null findOneByName(string $name)
+ */
+class MigrationRepository extends ServiceEntityRepository
```
aschempp: I don't think this class is necessary at all. You can get a default repository through the doctrine registry

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: > We're currently not using this anywhere else in the code base, do we?

No, but the Symfony documentation told me about this and this is the first time I have seen it and I wanted to try it 😁

aschempp (changes requested):

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: > I've been using this pattern quite heavily lately but I'm not sure I really like it. It makes you introduce traits for everything and then you'll end up with a lot of magic and more named things that need to be managed.

What intrigued me is that using it this way automatically defines the correct type - compared to `$this->container->get('…')`.

core-bundle/src/Entity/Migration.php
```diff
@@ -0,0 +1,67 @@
...
+     * @var string
+     *
+     * @ORM\Column(type="string", length=255, nullable=false)
+     */
+    protected $name;
+
+    /**
+     * @var \DateTimeInterface
+     *
+     * @ORM\Column(type="datetime", nullable=false)
```
m-vo, line 48: Should we maybe use `\DateTimeImmutable` + `datetime_immutable` ([ref](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#datetime-immutablehere)) here?

m-vo: I'd probably also call the db field `executed_at` to be inline with Doctrine migrations.

core-bundle/src/Repository/MigrationRepository.php
m-vo: I'd still use this repository and introduce an explicit `hasEntry(string $name)` method that checks for existence without the need to fetch the record (like [here](https://stackoverflow.com/questions/36801089/doctorine-check-if-record-exists-in-database#comment94987732_36802084) for example).

core-bundle/src/Entity/Migration.php
fritzmg: Well, we already have a mix here. DCA fields are camel case and in the CronJob entity we also use `lastRun` instead of `last_run` or `last_executed` for example.

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: > I wonder if we could simply replace the shouldRun method? And if you wanna extend it, you can call parent?

Yes that's what I would do. Just remove the `abstract` IMHO.

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: Yes, but you'll get the same experience with just the methods (and their return types) + one additional `getSubscribedServices`. I did this in the image studio classes for example.

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: As discussed in slack, in most cases you will need to implement your own `shouldRun` method anyway - unless your migration has nothing to do with the database at all. Otherwise you will always have to check whether your affected tables and fields are actually present (since the migration could be executed before any of them are there). Otherwise I am fine with providing a default implementation for the `shouldRun` method in the `AbstractRecordedMigration`.

core-bundle/src/Repository/MigrationRepository.php
fritzmg: Hm, couldn't really figure out how. All the documentation and examples point to this being the way to go?

m-vo:
> hasRun will return either true, false or null - the latter will be returned, if the tl_migration table does not exist yet.

In case the table does not exist, this means the migration cannot be run, right? Could/should we create the table in this case and return `false` instead?

core-bundle/src/Repository/MigrationRepository.php
fritzmg: nvm, done ;)

fritzmg:
> In case the table does not exist, this means the migration cannot be run, right? Could/should we create the table in this case and return `false` instead?

Not sure, I'd want to let Contao/Doctrine handle the creation of the table, instead of doing it manually. Kind of defeats the purpose of having an entity otherwise.

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: This would also force you to set `autowiring: true` for your migration service. And then you could just use constructor injection as well which IMO is the easiest to understand. The lazyness could be nice on its own though. Ah, not easy. :slightly_smiling_face:

fritzmg: Since we have a new abstract class anyway, then using a constructor argument to inject both services would be fine. I went this route because I originally extended the existing `AbstractMigration` class and there I did not want to introduce a new constructor.

m-vo:
> Not sure, I'd want to let Contao/Doctrine handle the creation of the table, instead of doing it manually. 

This would still work by filtering the schema I think. But not sure it's worth it.

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: > This would also force you to set `autowiring: true` for your migration service.

Well no, actually all you need is

```yml
_instanceof:
    Contao\CoreBundle\Migration\AbstractRecordedMigration:
        calls:
            - [setContainer, ['@Psr\Container\ContainerInterface']]
```

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: > As discussed in slack, in most cases you will need to implement your own shouldRun method anyway

Makes sense :+1: 

> Otherwise I am fine with providing a default implementation for the shouldRun method

Yeah, because you could *either* have an empty abstract method or a default implementation, I'd go with the default implementation. Then you get documentation how this feature works for free and adding your own stuff is just a quick `CTRL + o` (or similar) away. :slightly_smiling_face:

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: Isn't that more complex to setup or would you ship that in the ME?

fritzmg: The Managed Edition already has autowiring enabled for the `App` namespace. I am not sure we really should introduce a constructor, otherwise you will always have to inject the same arguments for your own constructor and then passt it on to the parent? Which I find more cumbersome. The current solution is more in line with the `AbstractController` from Symfony for example, is it not?

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,59 @@
...
+        if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
+            return null;
+        }
+
+        return null !== $this->entityManager()->getRepository(MigrationEntity::class)->findOneBy(['name' => $this->getName()]);
+    }
+
+    protected function createResult(bool $successful, string $message = null): MigrationResult
+    {
+        if (false === $this->hasRun()) {
```
fritzmg: Initially I also checked for `$successful` here - but that might lead to infinite executions of a migration.

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,59 @@
...
+    public function shouldRun(): bool
+    {
+        return false === $this->hasRun();
+    }
+
+    protected function hasRun(): ?bool
+    {
+        if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
+            return null;
+        }
```
ausi:
```suggestion
    protected function hasRun(): bool
    {
        if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
            return false;
        }
```

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,59 @@
...
+        return null !== $this->entityManager()->getRepository(MigrationEntity::class)->findOneBy(['name' => $this->getName()]);
+    }
+
+    protected function createResult(bool $successful, string $message = null): MigrationResult
+    {
+        if (false === $this->hasRun()) {
+            $migrationEntity = new MigrationEntity($this->getName());
+            $this->entityManager()->persist($migrationEntity);
+            $this->entityManager()->flush();
+        }
```
ausi:
```suggestion
        if (false === $this->hasRun()) {
            if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
                $this->createMigrationTable();
            }
            $migrationEntity = new MigrationEntity($this->getName());
            $this->entityManager()->persist($migrationEntity);
            $this->entityManager()->flush();
        }
```

With the `createMigrationTable()` method (untested):

```php
use Doctrine\ORM\Tools\SchemaTool;
…
private function createMigrationTable(): void
{
	$schemaTool = new SchemaTool($this->entityManager());
	$schemaTool->createSchema([
		$this->entityManager()->getClassMetadata(MigrationEntity::class),
	]);
}
```

ausi (commented):
I think we should create the migration table as soon as we want to write to it the first time. See code comments.

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: I see, thank you, had no idea it was that simple 👍

ausi: Or maybe something like this? 

```php
use Doctrine\ORM\Tools\SchemaTool;
…
private function createMigrationTable(): void
{
	$schemaTool = new SchemaTool($this->entityManager());
	$sql = $schemaTool->getCreateSchemaSql([
		$this->entityManager()->getClassMetadata(MigrationEntity::class),
	]);
	array_filter($sql, /* check for tl_migration */ );
	// Execute SQL...
}
```

m-vo: You probably also to create SQL from the schema and execute it.

Like  `$connection->exec($schemaTool->createSchema(…)->toSql($platform))`.

fritzmg: @m-vo `createSchema` already executes it. The problem is that Contao's DoctrineSchemaListener adds all its tables via event and thus it fails.

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -47,6 +52,21 @@ protected function createResult(bool $successful, string $message = null): Migra
         return parent::createResult($successful, $message);
     }
 
+    private function createMigrationTable(): void
+    {
+        $entityManager = clone $this->entityManager();
```
ausi: If we don’t remove any listeners I guess we don’t have to clone the entity manager?

```suggestion
        $entityManager = $this->entityManager();
```

core-bundle/src/Migration/AbstractRecordedMigration.php
```diff
@@ -0,0 +1,79 @@
...
+        if (!$this->connection()->getSchemaManager()->tablesExist(['tl_migration'])) {
+            return false;
+        }
+
+        return null !== $this->entityManager()->getRepository(MigrationEntity::class)->findOneBy(['name' => $this->getName()]);
+    }
+
+    protected function createResult(bool $successful, string $message = null): MigrationResult
+    {
+        if (!$this->hasRun()) {
```
ausi, line 36: Should we always create an entity if the migration was run, even if it runs multiple times?

ausi (dismissed):
Apart from https://github.com/contao/contao/pull/3108#discussion_r656610641 ✅

core-bundle/src/Migration/AbstractRecordedMigration.php
m-vo: String operations on the SQL are a nasty hack, though. :smile: We could also clone the em and remove all event listeners (because you would not want any probably). But I feel there's a reason why doing this is so clunky. Maybe we should either just *not* do it or not use an entity to begin with and create the schema on the fly.

core-bundle/src/Entity/Migration.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->name = $name;
+        $this->dateRun = $dateRun ?? new \DateTime();
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function getDateRun(): \DateTimeInterface
```
Toflar: Minor nitpick: This migration `ran`, no? It's over so it's past tense. Maybe just name it `executedAt` or so :D

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: I think we should, yes. If your `shouldRun` relies on this information, then it must be set whenever it is run once. Though may be `createResult` is the wrong location for this to happen and it should rather be its own dedicated method, that you have to call yourself separately?

fritzmg:
[This](https://github.com/contao/contao/pull/3108#discussion_r656822681) revealed another problem. I wanted to rename the field's name - but this leads to an error since `hasRun` will now be executed _before_ the field is migrated to its new name.

```
An exception occurred while executing 'SELECT t0.id AS id_1, t0.name AS name_2, t0.executedAt AS executedAt_3 FROM t   
  l_migration t0 WHERE t0.name = ? LIMIT 1' with params ["App\\Migration\\ExampleMigration"]:                    
                                                                                                                 
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 't0.executedAt' in 'field list'
```

Which makes me think we should still go with my original approach and make sure the migration relying on `tl_migration` is not run before the `tl_migration` table is fully migrated itself.

ausi:
> Which makes me think we should still go with my original approach and make sure the migration relying on `tl_migration` is not run before the `tl_migration` table is fully migrated itself.

But we must not run the whole schema migration in this process (only the ones for `tl_migration`). Because most migrations want to run before the schema migrations.

fritzmg:
I have now put the schema update for `tl_migration` into its own separate migration that is tagged with a priority of `1024`, so that the `AbstractRecordedMigration` itself does not have to take care of this. @ausi wdyt?

ausi (dismissed):

core-bundle/src/Migration/AbstractRecordedMigration.php
ausi: `createResult()` is OK for me to store the run information.

core-bundle/src/Migration/AbstractRecordedMigration.php
fritzmg: Well, let's leave the 'nasty hack' for now ;).

core-bundle/src/Entity/Migration.php
fritzmg: It's now renamed to `executedAt` at least, which is consistent with the other Contao entities.

m-vo (dismissed):
My two picks would be:

* I'd use `\DateTimeImmutable` (and `datetime_immutable`) because to me that's the saner default.
* I don't like the `ServiceSubscriberTrait`. :smile_cat: 

But still, very :+1:

fritzmg:
> * I'd use `\DateTimeImmutable` (and `datetime_immutable`) because to me that's the saner default.

Well, I used the same type as in the other Contao entities 🙈. If we want to change that, it should probably be changed for all entities?

> * I don't like the ServiceSubscriberTrait. 😸

Let's see what the other @contao/developers think of it 😁. Otherwise it's of course no problem to implement the `setContainer` & `getSubcribedServices` methods ourselves in `AbstractRecordedMigration`.

fritzmg:
Added some tests. Not really sure if or how I could write a test for `MigrationTableMigration` though.

ausi (approved):

core-bundle/tests/Fixtures/Migration/FooRecordedMigration.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Tests\Fixtures\Migration;
+
+use Contao\CoreBundle\Migration\AbstractRecordedMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+
+class FooRecordedMigration extends AbstractRecordedMigration
```
Toflar, line 18: Thanks to anonymous classes you don't really need this fixture.
You can just use this in your tests:

```php
new class extends AbstractRecordedMigration
{
    public function run(): MigrationResult
    {
        return $this->createResult(true);
    }
}
```

Just as a hint :)

Toflar: Ah you're mocking stuff. That's a different story.

fritzmg: With an anonymous class I get a random string for the class name - but I need to know the exact name for the tests.

Toflar: https://github.com/contao/contao/blob/4.x/core-bundle/tests/Routing/ResponseContext/ResponseContextTest.php#L107-L116 ;)

fritzmg: Right, ok :). I can still change this if needed. Personally I think it's more convenient this way.

fritzmg:
`RoutingMigrationTest::testShouldRun` fails, but I am not really sure what needs to be fixed. @contao/developers any ideas?

m-vo:
> RoutingMigrationTest::testShouldRun fails, but I am not really sure what needs to be fixed. @contao/developers any ideas?

The CI run looks fine to me. Where do you get that error?

fritzmg:
> The CI run looks fine to me. Where do you get that error?

» https://github.com/contao/contao/pull/3108/checks?check_run_id=2903182361

```
 Testing 
PHP Fatal error:  During class fetch: Uncaught ReflectionException: Class "Symfony\Component\Validator\Mapping\Loader\AutoMappingTrait" not found while loading "Symfony\Bridge\Doctrine\Validator\DoctrineLoader". in /home/runner/work/contao/contao/vendor/symfony/error-handler/DebugClassLoader.php:347
Stack trace:
#0 /home/runner/work/contao/contao/vendor/symfony/config/Resource/ClassExistenceResource.php(84): class_exists('...')
#1 /home/runner/work/contao/contao/vendor/symfony/dependency-injection/ContainerBuilder.php(348): Symfony\Component\Config\Resource\ClassExistenceResource->isFresh(0)
#2 /home/runner/work/contao/contao/vendor/terminal42/service-annotation-bundle/src/DependencyInjection/Compiler/ServiceAnnotationPass.php(56): Symfony\Component\DependencyInjection\ContainerBuilder->getReflectionClass('...', false)
#3 /home/runner/work/contao/contao/vendor/symfony/dependency-injection/Compiler/Compiler.php(91): Terminal42\ServiceAnnotationBundle\DependencyInjection\Compiler\ServiceAnnotationPass->process(Object(Symfony\Component\DependencyInjection\ContainerBuilder))
#4 /home/runner/work/contao/contao/vendor/symfony/dependency-injection/ContainerBuilder.php(736): Symfony\Component\DependencyInjection\Compiler\Compiler->compile(Object(Symfony\Component\DependencyInjection\ContainerBuilder))
#5 /home/runner/work/contao/contao/vendor/symfony/http-kernel/Kernel.php(541): Symfony\Component\DependencyInjection\ContainerBuilder->compile()
#6 /home/runner/work/contao/contao/vendor/symfony/http-kernel/Kernel.php(780): Symfony\Component\HttpKernel\Kernel->initializeContainer()
#7 /home/runner/work/contao/contao/vendor/symfony/http-kernel/Kernel.php(121): Symfony\Component\HttpKernel\Kernel->preBoot()
#8 /home/runner/work/contao/contao/vendor/symfony/framework-bundle/Test/KernelTestCase.php(80): Symfony\Component\HttpKernel\Kernel->boot()
#9 /home/runner/work/contao/contao/core-bundle/tests/Functional/Migration/RoutingMigrationTest.php(28): Symfony\Bundle\FrameworkBundle\Test\KernelTestCase::bootKernel()
#10 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestCase.php(1472): Contao\CoreBundle\Tests\Functional\Migration\RoutingMigrationTest->testShouldRun(Array, false)
#11 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestCase.php(1092): PHPUnit\Framework\TestCase->runTest()
#12 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestResult.php(703): PHPUnit\Framework\TestCase->runBare()
#13 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestCase.php(820): PHPUnit\Framework\TestResult->run(Object(Contao\CoreBundle\Tests\Functional\Migration\RoutingMigrationTest))
#14 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestSuite.php(627): PHPUnit\Framework\TestCase->run(Object(PHPUnit\Framework\TestResult))
#15 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestSuite.php(627): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#16 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestSuite.php(627): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#17 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/Framework/TestSuite.php(627): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#18 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/TextUI/TestRunner.php(656): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#19 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/TextUI/Command.php(236): PHPUnit\TextUI\TestRunner->doRun(Object(PHPUnit\Framework\TestSuite), Array, Array, true)
#20 /home/runner/work/contao/contao/vendor/phpunit/phpunit/src/TextUI/Command.php(195): PHPUnit\TextUI\Command->run(Array, true)
#21 /home/runner/work/contao/contao/vendor/phpunit/phpunit/phpunit(76): PHPUnit\TextUI\Command::main()
#22 {main} in /home/runner/work/contao/contao/vendor/symfony/doctrine-bridge/Validator/DoctrineLoader.php on line 31
```

m-vo:
That's beacause the `symfony/validator` component is not installed and thus the trait is missing and autoloading fails. See #3111 (and related issues) where this is handled.

fritzmg:
Ah I see. Failing test not related then 😁

Toflar (approved):

leofeyer:
> > * I don't like the ServiceSubscriberTrait. 😸
> 
> Let's see what the other @contao/developers think of it 😁. Otherwise it's of course no problem to implement the `setContainer` & `getSubcribedServices` methods ourselves in `AbstractRecordedMigration`.

What is wrong with using the trait?

aschempp:
> I have now put the schema update for `tl_migration` into its own separate migration that is tagged with a priority of `1024`, so that the `AbstractRecordedMigration` itself does not have to take care of this. @ausi wdyt?

I see that you changed this now? I don't really like running the migration in `shouldRun` though, shouldn't the migration be executed when running migrations anyway?

fritzmg:
>  don't really like running the migration in `shouldRun` though, shouldn't the migration be executed when running migrations anyway?

I needs to be executed in `shouldRun` because `shouldRun` is executed first for _all_ migrations and thus those that extend from `AbstractRecordedMigration` already need the table to be present in `shouldRun`.

m-vo:
> What is wrong with using the trait?

It introduces a new way of doing something that already works. Following the promoted pattern leads to abstracting the constantly recurring methods into their own traits (like `ConnectionAware`). At first this looks nice, but after a while naming things becomes a problem (e.g. if suddenly there is a `Foo\Connection` ). Also services that cannot be autowired with their FQCN do not work and IIRC there was a problem using it in controllers.

The trait itself is fine, I guess. The implied pattern me no like. 

Maybe also worth a read: https://github.com/symfony/symfony/pull/27077

aschempp:
> I needs to be executed in `shouldRun` because `shouldRun` is executed first for _all_ migrations and thus those that extend from `AbstractRecordedMigration` already need the table to be present in `shouldRun`.

Yeah, but checking for whether a migration needs to be executed should never run a migration. There must be another way. Since the core comes with the table migration in a high priority, can't _recorded migrations_ just safely assume the table will be there if it's not. So the `shouldRun` check should check if either table does not exist or job hasn't run?

fritzmg:
@aschempp this would work if the `tl_migration` table is missing, but it won't work if `tl_migration` already exists and the entity has changed (e.g. new required fields get added, or a field was renamed). `AbstractRecordedMigration` would have to check for _any_ schema changes to `tl_migration` - and then the `MigrationTableMigration` would become obsolete. I'd rather use the current workaround, to keep things simple.

leofeyer:
Isn't it bad UX to handle a few migrations differently than all the others? I'm afraid that this will cause questions such as

> I have restored my `tl_content` and `tl_files` tables from an SQL dump and run `contao:migrate` but Contao only migrated the `tl_content` table.

Should we maybe record all migrations and add a `--rerun-completed` flag (or something similar) to the `contao:migrate` command?

m-vo:
> Should we maybe record all migrations and add a --rerun-completed flag (or something similar) to the contao:migrate command?

How would we handle the existing migrations then that already ran? Still inserting a row if `!shouldRun && notExist`?

fritzmg:
So, by default, all migrations would always execute once and `MigrationCollection` would instead handle the recording of which migrations ran?

aschempp:
> @aschempp this would work if the `tl_migration` table is missing, but it won't work if `tl_migration` already exists and the entity has changed (e.g. new required fields get added, or a field was renamed). `AbstractRecordedMigration` would have to check for _any_ schema changes to `tl_migration` - and then the `MigrationTableMigration` would become obsolete. I'd rather use the current workaround, to keep things simple.

why is that? The `MigrationTableMigration` just needs to be updated if the schema of the `tl_migration` table changes. Since `MigrationTableMigration` is not a recorded migration, we can always just update it instead of adding a new migration for new things.

> Should we maybe record all migrations and add a --rerun-completed flag (or something similar) to the contao:migrate command?

Absolutely not. The beauty of the current concept is that migrations are independent, if I duplicate one old table from my other setup into the current one, I want the respective migration to re-run. Recorded migrations should only be for edge cases where you can't know whether it needs to be executed (again) or not.

fritzmg:
> why is that?

Because `shouldRun` of the `AbstractRecordedMigration` already accesses `tl_migration` and thus might fail, if a new field was added in the entity. The field would be missing, since `shouldRun` is first exectued for _all_ migrations, before `run` is executed.

leofeyer:
> The beauty of the current concept is that migrations are independent, if I duplicate one old table from my other setup into the current one, I want the respective migration to re-run. Recorded migrations should only be for edge cases where you can't know whether it needs to be executed (again) or not.

@aschempp I agree, but this will most certainly confuse people and lead to support requests such as https://github.com/contao/contao/pull/3108#issuecomment-869676427

aschempp:
that's why we should avoid recorded migrations at all costs 🤷 
I honestly think we should never have recorded migrations for anything database related. The only use of recorded migrations is if you e.g. need to move files. As I said, anything you don't know if it has been done before. I know independent migrations can be a pain to write, but thats like unit tests 😂

fritzmg:
> I honestly think we should never have recorded migrations for anything database related. The only use of recorded migrations is if you e.g. need to move files.

The use case for this comes from a database related migration though: https://github.com/contao/contao/pull/3103

leofeyer:
> that's why we should avoid recorded migrations at all costs 🤷

Full ACK!

fritzmg:
> @aschempp I agree, but this will most certainly confuse people and lead to support requests such as [#3108 (comment)](https://github.com/contao/contao/pull/3108#issuecomment-869676427)

I don't really understand your constructed example. Currently the migration for the important part for example is independently checked, so if you restored only your `tl_files` table from a Contao 4.4 backup into your Contao 4.9+ installation,  `tl_files` would still be migrated.

leofeyer:
But https://github.com/contao/contao/pull/3103 would not.

fritzmg:
> But #3103 would not.

Yes, and otherwise we would not be able to write a correct migration anyway.

leofeyer:
Then, as @aschempp stated above, we should rather not have a migration at all.

fritzmg:
I disagree. Also in this particular case (https://github.com/contao/contao/pull/3103 I mean) you can still check (and absolutely have to check) the database independently of `tl_migration` whether the migration needs to run. You need to fall back on `tl_migration` if it is ambiguous.

leofeyer:
That is exactly the point. There is no reliable way to check the database in #3103 (see https://github.com/contao/contao/pull/3103#issuecomment-866022370), which is why we started discussing recorded migrations in the first place. However, the concept does not convince me. 🙈

fritzmg:
Regarding the user experience when re-importing specific tables: I think that's a moot point, as you cannot expect migrations to work when you import just specific tables, regardless of recorded migrations. A migration's `shouldRun` contains arbitrary conditions which might include several tables and/or fields.

If you already specifically _know_ the conditions for a specific migration and thus are able to determine what you need to do so that it will execute again, it again would not matter if it was a recorded migration or not.

And for the use case of https://github.com/contao/contao/pull/3103 specifically: if you re-import `tl_files` and want to run the migration again, you can execute `DELETE FROM tl_migration WHERE name = '…\ExcludedFileFieldsMigration';` afterwards.

Toflar:
Protip: You guys should have a phone call. Saying hello from my vacation :p

fritzmg:
Mumble call tomorrow ;)

ausi:
> Mumble call tomorrow ;)

Tomorrow? Not thursday?

fritzmg:
Thursday 🙈

leofeyer:
As discussed in Mumble on July 1st, we currently do not want to add this to the core. We might release it as separate bundle in case someone needs recorded migrations (now that we have already built it).

PR #3110, opened 2021-06-22
Add the `rawPlainText()` and `rawHtmlToPlainText()` template helper methods

Toflar:
Most of the following PRs for JSON-LD will likely require raw values in the template so I wanted to get this in first.
It makes the templates a lot more concise rather than using e.g. the long `Contao\StringUtil::revertInputEncoding()` calls.

I've also made sure they all start using `raw` so it's very clear that they output raw stuff. And I think the `revertInputEncoding` in the names is not required here as we're explicitly in the template context.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -400,6 +400,46 @@ abstract class Template extends Controller
 	}
 
 	/**
+	 * Helper method to allow quick access in the Contao templates for raw output.
+	 * Be careful when using this. It must NOT be used within regular HTML when $value
+	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
+	 * or JSON-LD arrays.
+	 */
+	public function raw(string $value): string
+	{
+		return StringUtil::revertInputEncoding($value);
+	}
+
+	/**
+	 * Helper method to allow quick access in the Contao templates for raw output.
+	 *
+	 * Compared to $this->raw() it does remove HTML.
+	 *
+	 * Be careful when using this. It must NOT be used within regular HTML when $value
+	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
+	 * or JSON-LD arrays.
+	 */
+	public function rawPlainText(string $value, bool $removeInsertTags = false): string
+	{
+		return StringUtil::inputEncodedToPlainText($value, $removeInsertTags);
+	}
+
+	/**
+	 * Helper method to allow quick access in the Contao templates for raw output.
+	 *
+	 * Compared to $this->rawPlainText() it adds new lines before and after block level HTML elements
+	 * and only then removes the rest of the HTML tags.
+	 *
+	 * Be careful when using this. It must NOT be used within regular HTML when $value
+	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
+	 * or JSON-LD arrays.
+	 */
+	public function rawHtmlToPlainText(string $value, bool $removeInsertTags = false): string
+	{
+		return StringUtil::htmlToPlainText($value, $removeInsertTags);
+	}
+
+	/**
 	 * Adds schema.org JSON-LD data to the current response context
 	 */
 	public function addSchemaOrg(array $jsonLd): void
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -400,20 +400,8 @@ abstract class Template extends Controller
 	}
 
 	/**
-	 * Helper method to allow quick access in the Contao templates for raw output.
-	 * Be careful when using this. It must NOT be used within regular HTML when $value
-	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
-	 * or JSON-LD arrays.
-	 */
-	public function raw(string $value): string
-	{
-		return StringUtil::revertInputEncoding($value);
-	}
-
-	/**
-	 * Helper method to allow quick access in the Contao templates for raw output.
-	 *
-	 * Compared to $this->raw() it does remove HTML.
+	 * Helper method to allow quick access in the Contao templates for safe raw (unencoded) output.
+	 * It replaces (or optionally removes) Contao insert tags and removes all HTML.
 	 *
 	 * Be careful when using this. It must NOT be used within regular HTML when $value
 	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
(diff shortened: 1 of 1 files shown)


m-vo:
I'm just thinking how we are going to transfer these things for Twig templates. If the context itself is already encoded, our "double encoding prevention" would not be enough anymore. Using `|raw` would then output encoded stuff. We would then have to call `revertInputEncoding` on everything which is probably a bad idea or the user would need to do something like `foo|revertInputEncoding|raw` which is terrible UX.

If we can determine the variables that need to be available in raw form before a template gets rendered they could just be added to the context (e.g. under a `raw` array or with a `raw_` prefix) and we could replace the encoded values with these when creating a Twig context or something.

Toflar:
The idea here is to basically support this:

```php
$this->addSchemaOrg([
    'foobar' => $this->raw($value);
]);
```

This is what we need. I don't know for Twig.

Toflar:
But I think Twig is not relevant for this PR and pollutes the discussion. Those methods already exist. We're just making them easily accessible in the template which is a requirement. We need it - no matter what Twig wants :)

m-vo:
> We need it - no matter what Twig wants :)

That is basically killing lots of things in my Twig PR so I strongly disagree. :shrug: 

If we could have something like this…
```php
$template->makeAvailableAsRawValue('foobar');
```

… and this would add the value to an internal "raw array" that is accessed by `Template#raw`, everything could be easily transferred and your use case would still work. But now you are moving the responsibility to the template which Twig would never be able to detect, I think.

Toflar:
I don't understand. The template gets an encoded `foobar` value and it knows it's encoded so it calls `$this->raw()` on it for the JSON-LD stuff. How is that related to Twig?

m-vo:
Because there is no analogy for this in Twig where you assume everything in the context is raw and it gets escaped automatically when not adding `|raw`. We would need to add a similar `raw()` function as function or filter to Twig which would be very confusing to use in my eyes.

Like `foo|contaoRaw` (or `foo|contaoRaw|raw`) or sth. From a Twig UX perspective this does not really make sense. But `foo|raw` would work if `foo` was already put through the revert* when building the context.

Toflar:
But if it is already raw, then you don't need to call any method on the value, you just use it? I still don't understand it at all, sorry.

In Contao templates it would be

```php
$this->addSchemaOrg([
    'foobar' => $this->raw('encoded foobar');
]);
```

and in Twig it's already unencoded so no need for those methods?

```twig
{{ schemaOrg(['foobar' => 'unencoded foobar']) }}
```

m-vo:
Ah maybe we're talking about something different :slightly_smiling_face:. I'm thinking of a Twig template that replaces or extends a Contao template and therefore must work with its context. If you can create your own context, everything's fine of course. But maybe I'm also missing some points. Will look at it again tomorrow with fresh eyes. 

(And I'm sure there is a solution, no hard feelings.)

Toflar:
Yeah sure, probably it'll be best if you PR'ed the required changes for Twig to this PR directly because I'm a bit lost at this point 😄

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -399,6 +399,46 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
...
+	/**
+	 * Helper method to allow quick access in the Contao templates for raw output.
+	 *
+	 * Compared to $this->raw() it does remove HTML.
+	 *
+	 * Be careful when using this. It must NOT be used within regular HTML when $value
+	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
+	 * or JSON-LD arrays.
+	 */
+	public function rawPlainText(string $value, bool $removeInsertTags = false): string
```
aschempp, line 410: I'd prefer `rawText` and `rawHtml` method names

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -399,6 +399,46 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
...
 
+	/**
+	 * Helper method to allow quick access in the Contao templates for raw output.
+	 * Be careful when using this. It must NOT be used within regular HTML when $value
+	 * is uncontrolled user input. It's useful to ensure raw values within e.g. <code> examples
+	 * or JSON-LD arrays.
+	 */
+	public function raw(string $value): string
+	{
+		return StringUtil::revertInputEncoding($value);
```
ausi: I don’t think we should make this method available to templates, it is too dangerous (doesn’t handle insert tags for example).

The other two methods are fine and should be all we need I think.

core-bundle/src/Resources/contao/library/Contao/Template.php
ausi: `rawHtml` would be very confusing IMO as it does not output HTML.

core-bundle/src/Resources/contao/library/Contao/Template.php
Toflar: What's with something like this?

```php
<pre>
<?= $this->raw($this->encodedHtmlExample) ?>
</pre>
```

ausi: Directly printing to the HTML is not safe and could inject basically anything.
But your example would be unsafe with the other two methods too.

m-vo (approved):

Toflar:
Updated description and removed `raw()` after short call with @ausi and @m-vo 😄

m-vo:
Memo to myself: add `contaoInputEncodedToPlainText` + `contaoHtmlToPlainText` filters for Twig (e.g. in #2988).

ausi (approved):

aschempp:
I don't really see the point in just forwarding methods, even tough I understand they make the code "shorter". Unfortunately, there's no autocomplete on them, compared to using the `StringUtil` class.

If we can agree that all/most of the StringUtil methods are useful in a template, why not simply forward all of them? I could then switch to using `$this->specialchars()` as well instead of using `\Contao\StringUtil::specialchars()` when adding a value to an HTML attribute.

```php
class Template
{
    public function __call($method, array $arguments)
    {
        if (\method_exists(StringUtil::class, $method)) {
            return StringUtil::$method(...$arguments);
        }

        throw new \LogicException('Method '.$method.' does not exist.');
    }
}
```

Toflar:
First of all, we already use `__call()` so you can add `$template->foo = function() {};`.
Second, there are methods we don't want to make available such as `revertInputEncoding()`.

Toflar:
Third, if we do this, we also have to do it for Twig. Not sure if @m-vo would be happy with this :D

leofeyer:
Thank you @Toflar.

PR #3111, opened 2021-06-22
Fix the unit and functional tests on PHP 8.1

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

This is an attempt to stop the nightly CI job from failing.

Followup to #3081.


m-vo:
About the synthetic thing: The `AutoMappingTrait` that is used in the `DoctrineLoader` class might not be available. At container build time the class_existence of this class gets checked however which triggers autoloading and thus produces an error. 

This might IMO happens every time (even in prod) it just doesn't show as an error in <8.1.  It's because the service annotation bundle gets the reflection class of each definition what triggers it in the first place.

Not sure how to properly handle this. The last commit would fix it for the tests, but if I get this correctly, this will also appear when building the container for your app. Maybe it needs to be fixed in `DoctrineExtension#loadValidatorLoader` (doctrine-bundle).

m-vo:
for reference: https://github.com/symfony/symfony/issues/41796 / https://github.com/doctrine/DoctrineBundle/issues/1377

aschempp:
how can the MySQLi extension not be enabled? Shouldn't we rather make sure it is always enabled?

m-vo:
I thought this was due to the extension not yet available in 8.1 (fix similar to #3081). The ultimate reason probably is the mysqli default error mode, that is set to exceptions, now. 

See https://wiki.php.net/rfc/mysqli_default_errmode + https://php.watch/versions/8.1/mysqli-error-mode

Any ideas?

ausi (dismissed):
🎉

Can be merged IMO, but we should create a ticket to remind us that we should review *deprecationsBaseline.json* once PHP 8.1 gets released.

leofeyer:
I'm not sure about using a `deprecationsBaseline.json` file. First of all, the file should probably not be in the root directory. And second of all, we now suppress deprecation messages that we could/should handle instead or at least should be reminded to handle them once we increase the minimum PHP or Symfony version.

m-vo:
The deprecation messages (all?) originate from Symfony's/3rd party code, though. That's why they only appear as direct deprecations in the functional tests.

>  First of all, the file should probably not be in the root directory. 

Where should it live then? In `core-bundle/tests/Functional` maybe? I put it in the root because it's the single baseline file for all tests. The deprecation helper does not allow to "namespace" things.

leofeyer:
> The deprecation messages (all?) originate from Symfony's/3rd party code, though. That's why they only appear as direct deprecations in the functional tests.

But direct deprecations should not make the job fail, should they? I figured only self deprecations do.

> Where should it live then? In `core-bundle/tests/Functional` maybe?

I am no sure. Symfony has put a lot of CI related files into the `.github` folder (see https://github.com/symfony/symfony/tree/5.4/.github). Maybe that is a good place?

m-vo:
> But direct deprecations should not make the job fail, should they? I figured only self deprecations do.

I do not really know this, probably need to investigate.

If it helps, here is the full output for the functional tests on php-src@973ae8 without the baseline file:

```
PHPUnit 8.5.16 by Sebastian Bergmann and contributors.

Testing 
...............................................................  63 / 211 ( 29%)
............................................................... 126 / 211 ( 59%)
............................................................... 189 / 211 ( 89%)
......................                                          211 / 211 (100%)

Time: 10.94 minutes, Memory: 94.50 MB

OK (211 tests, 769 assertions)

Unsilenced deprecation notices (14)

  1x: Declaration of Doctrine\DBAL\Driver\PDOConnection::exec($sql) should be compatible with PDO::exec(string $statement): int|false
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOConnection::lastInsertId($name = null) should be compatible with PDO::lastInsertId(?string $name = null): string|false
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOConnection::prepare($sql, $driverOptions = []) should be compatible with PDO::prepare(string $query, array $options = []): PDOStatement|false
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOConnection::quote($value, $type = Doctrine\DBAL\ParameterType::STRING) should be compatible with PDO::quote(string $string, int $type = PDO::PARAM_STR): string|false
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOQueryImplementation::query(?string $query = null, ?int $fetchMode = null, mixed ...$fetchModeArgs) should be compatible with PDO::query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs): PDOStatement|false
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::bindParam($param, &$variable, $type = Doctrine\DBAL\ParameterType::STRING, $length = null, $driverOptions = null) should be compatible with PDOStatement::bindParam(string|int $param, mixed &$var, int $type = PDO::PARAM_STR, int $maxLength = 0, mixed $driverOptions = null): bool
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::bindValue($param, $value, $type = Doctrine\DBAL\ParameterType::STRING) should be compatible with PDOStatement::bindValue(string|int $param, mixed $value, int $type = PDO::PARAM_STR): bool
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::closeCursor() should be compatible with PDOStatement::closeCursor(): bool
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::execute($params = null) should be compatible with PDOStatement::execute(?array $params = null): bool
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEXT, $cursorOffset = 0) should be compatible with PDOStatement::fetch(int $mode = PDO::FETCH_DEFAULT, int $cursorOrientation = PDO::FETCH_ORI_NEXT, int $cursorOffset = 0): mixed
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatement::fetchColumn($columnIndex = 0) should be compatible with PDOStatement::fetchColumn(int $column = 0): mixed
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\DBAL\Driver\PDOStatementImplementations::fetchAll($mode = null, ...$args) should be compatible with PDOStatement::fetchAll(int $mode = PDO::FETCH_DEFAULT, mixed ...$args): array
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\Persistence\Reflection\RuntimePublicReflectionProperty::getValue($object = null) should be compatible with ReflectionProperty::getValue(?object $object = null): mixed
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Declaration of Doctrine\Persistence\Reflection\RuntimePublicReflectionProperty::setValue($object, $value = null) should be compatible with ReflectionProperty::setValue(mixed $objectOrValue, mixed $value = <default>): void
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

Remaining self deprecation notices (108)

  72x: trim(): Passing null to parameter #1 ($string) of type string is deprecated
    24x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional
    23x in RoutingTest::testResolvesAliasesWithLocale from Contao\CoreBundle\Tests\Functional
    16x in RoutingTest::testResolvesAliasesWithoutUrlSuffix from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesTheRootPage from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesTheRootPageWithLocale from Contao\CoreBundle\Tests\Functional
    1x in RoutingTest::testOrdersThePageModelsByCandidates from Contao\CoreBundle\Tests\Functional

  13x: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated
    5x in RoutingTest::testResolvesAliasesWithoutUrlSuffix from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesAliasesWithLocale from Contao\CoreBundle\Tests\Functional

  13x: explode(): Passing null to parameter #2 ($string) of type string is deprecated
    5x in RoutingTest::testResolvesAliasesWithoutUrlSuffix from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional
    4x in RoutingTest::testResolvesAliasesWithLocale from Contao\CoreBundle\Tests\Functional

  8x: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated
    3x in RoutingTest::testResolvesTheRootPage from Contao\CoreBundle\Tests\Functional
    2x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional
    2x in RoutingTest::testResolvesAliasesWithoutUrlSuffix from Contao\CoreBundle\Tests\Functional
    1x in RoutingTest::testResolvesAliasesWithLocale from Contao\CoreBundle\Tests\Functional

  2x: The Serializable interface is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration
    1x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional

Remaining direct deprecation notices (2)

  1x: Declaration of Spatie\SchemaOrg\Graph::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed
    1x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional

  1x: Declaration of Spatie\SchemaOrg\BaseType::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed
    1x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional

Remaining indirect deprecation notices (2)

  1x: Since symfony/event-dispatcher 5.1: Symfony\Component\EventDispatcher\LegacyEventDispatcherProxy is deprecated, use the event dispatcher without the proxy.
    1x in RoutingTest::testResolvesAliases from Contao\CoreBundle\Tests\Functional

  1x: The "Scheb\TwoFactorBundle\Security\Authentication\RememberMe\RememberMeServicesDecorator" class implements "Symfony\Component\Security\Http\Logout\LogoutHandlerInterface" that is deprecated since Symfony 5.1.
    1x in RoutingTest::testResolvesAliases with data set "Renders the page if the alias is "index" and the request is empty" from Contao\CoreBundle\Tests\Functional

```

In red: `Unsilenced deprecation notices (14)`, `Remaining self deprecation notices (108)` and `Remaining direct deprecation notices (2)`.

ausi (dismissed):

m-vo:
Silly me. I used phpunits `--filter` function wrong (double quotes instead of single quotes). 

I identified some self deprecations, will update the PR accordingly.

Toflar:
Just a random input: Shouldn't that go against 4.9?

m-vo:
Will 4.9 be compatible with PHP 8.1?

Toflar:
EOL of PHP 8.0: 26 Nov 2023
EOL of Contao 4.9: 14 Feb 2024

m-vo:
> Just a random input: Shouldn't that go against 4.9?

Changes are still on 4.x for now, I can rebase them if we want to. :+1:  We did not patch any of the other PHP8 related changes (invalid array access …) into the 4.9 codebase, though, did we?

----

**Self deprecations:**

I now fixed all self deprecations. There was a tough one with our `User` class: 
```
The Serializable interface is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)
```

If the `\Serializable` interface is present and no `__*` methods are implemented, PHP will call `serialize()` and `unserialize()`. But that's deprecated now and the magic methods work a bit differently: they are there to prepare/restore the data but they do no serialization/unserialization themselves. I still left the old methods + the interface for BC but they aren't called anymore from within our code.

Note, that this changes the serialized representation for the `BackendUser` class as the parent is now an array and not a string. Does this cause problems? I followed how things are implemented for instance in `Symfony\Component\Security\Core\Authentication\Token\AbstractToken` but still kept our array keys. /cc @bytehead 

-----

**Baseline stuff:**

I still don't understand why the direct/unsilenced deprecations are treated as an error. I updated the baseline file so that they are ignored for now. Many errors will disappear once the fix in https://github.com/symfony/symfony/pull/41557 is released.

m-vo:
The serialize stuff seems to affect the PHP7.3 run - could it be that 7.3 uses the old methods if the itnerface exists? (any help appreciated, I'd be glad to get out of this rabbit hole…)

ausi:
> The serialize stuff seems to affect the PHP7.3 run - could it be that 7.3 uses the old methods if the itnerface exists? (any help appreciated, I'd be glad to get out of this rabbit hole…)

https://github.com/m-vo/contao/pull/4

ausi (dismissed):

ausi:
> Shouldn't that go against 4.9?

We should merge this first IMO, because the CI things are very fragile and most likely need some different changes for 4.9 anyways...

leofeyer:
I am still very unsure about the `deprecationsBaseline.json` file. If we cannot actually fix the deprecations in the code, we should just accept that the test is red. 🤷🏻

m-vo:
> If we cannot actually fix the deprecations in the code

These are all deprecations from outside our codebase, now.

leofeyer:
Yes, I know. They will be handled by the third-party maintainers at some point and until then, the test is red. 🤷🏻

m-vo:
If we cannot adjust phpunit/the deprecation helper to not fail for 3rd party code but only for self deprecations it's IMHO the best way to ignore these things. It doesn't do any harm, does it? But on the other hand the visual feedback of the CI works again.

ausi:
> But on the other hand the visual feedback of the CI works again.

And this has the benefit that it acutally shows us if and when our own code has issues with php 8.1.

m-vo:
> About the synthetic thing: The AutoMappingTrait that is used in the DoctrineLoader class might not be available. At container build time the class_existence of this class gets checked however which triggers autoloading and thus produces an error.

The fix for that one got merged + will probably land in the next 2.4 release: https://github.com/doctrine/DoctrineBundle/pull/1379

core-bundle/src/Resources/contao/classes/BackendUser.php
```diff
@@ -601,28 +601,48 @@ public function getRoles()
...
-	 * {@inheritdoc}
+	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
 	 */
 	public function unserialize($serialized)
 	{
 		$data = unserialize($serialized, array('allowed_classes'=>false));
 
+		if (!isset($data['parent']))
+		{
+			return;
```
leofeyer, line 621: Is this correct? Because it prevents `$this->__unserialize($data);` from being called if `$data['parent']` is not set.

core-bundle/src/Routing/AbstractPageRouteProvider.php
```diff
@@ -44,7 +44,7 @@ public function __construct(ContaoFramework $framework, CandidatesInterface $can
      */
     protected function findCandidatePages(Request $request): array
     {
-        $candidates = $this->candidates->getCandidates($request);
+        $candidates = array_map('strval', $this->candidates->getCandidates($request));
```
leofeyer, line 47: Why are you casting the values to string here? They will be casted to int again in line 58.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -1388,15 +1388,15 @@ public function getSlugOptions()
...
+	 * @param string      $strUrl
+	 * @param string|null $strParams
 	 *
 	 * @return string
 	 */
 	private function applyLegacyLogic($strUrl, $strParams)
 	{
 		// Decode sprintf placeholders
-		if (strpos($strParams, '%') !== false)
+		if (null !== $strParams && strpos($strParams, '%') !== false)
```
```suggestion
		if ($strParams !== null && strpos($strParams, '%') !== false)
```

core-bundle/src/Routing/AbstractPageRouteProvider.php
m-vo: Because `preg_match` expects a string. We could also cast them in line 57 but that wouldn't remove the second typecast in line 58.

core-bundle/src/Resources/contao/classes/BackendUser.php
ausi: IMO yes. If `$data['parent']` is not set, the data is in an unexpected format and it cannot be unserialized.

ausi: It worked the same way before: 

https://github.com/contao/contao/blob/43188094917d7e29242c7bf49418331a2f74f851/core-bundle/src/Resources/contao/classes/BackendUser.php#L611-L614

`parent` was required to be set.

ausi (dismissed):

leofeyer:
As discussed in Mumble on July 1st, we want to rename the file to `deprecations.json` and only use it in the PHP 8.1 job.

ausi (dismissed):

.github/workflows/ci.yml
```diff
@@ -185,7 +185,7 @@ jobs:
               run: vendor/bin/phpunit --colors=always
 
             - name: Run the functional tests
-              run: vendor/bin/phpunit --testsuite=functional --colors=always
+              run: SYMFONY_DEPRECATION_HELPER='max[self]=0&amp;baselineFile=./.github/deprecations.json' vendor/bin/phpunit --testsuite=functional --colors=always
```
leofeyer: I think we can remove the `deprecations.json` file altogether. The variable is called `SYMFONY_DEPRECATIONS_HELPER` and not `SYMFONY_DEPRECATION_HELPER`, which means that the file was never read (check the actions log). 😄 

And the PHP 8.1 test ist green now anyway. 🎉

m-vo: This is super weird, because without it (specified in the `phpunit.xml.dist` like before) it failed in my local 8.1 container yesterday. 

I'll rerun it and then update the PR accordingly.

m-vo: The `Unsilenced deprecation notices` now still show red locally. BUT: the error code is actually `0`. I don't get it. :exploding_head: :question: 

image

So, nevermind, the baseline file is gone in d8a15fcf3.

ausi (approved):

leofeyer:
Thank you @m-vo.

.github/workflows/ci.yml
leofeyer: These are not self deprecations, therefore they do not make the build fail.

PR #3116, opened 2021-06-23
Make all StringUtil methods available in the template

Toflar:
This PR implements what was requested by @aschempp and refers to https://github.com/contao/contao/pull/3110.
He mentioned that we're already using a few `StringUtil` methods in our templates which is a good indicator that making them available quickly might be desirable. I've adjusted all core templates so they would use the new `$this->stringUtil` service. We use `specialchars()`, `standardize()` and `substr()`. All of which make perfect sense.
There are also other use cases such as `encodeEmail()`, `highlight()` etc.

Regarding the argument that there are methods like `revertInputEncoding()` which should not be used in templates: Well, we cannot prevent that, it's changing one line. If `$this->stringUtil->revertInputEncoding()` was prevented, one might just change this to `Contao\StringUtil::revertInputEncoding()` and continue with their day...

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -150,6 +150,22 @@ abstract class Template extends Controller
 	 */
 	public function __get($strKey)
 	{
+		if ('stringUtil' === $strKey)
+		{
+			return new class()
+			{
+				public function __call(string $key, array $params)
+				{
+					if (!method_exists(StringUtil::class, $key))
+					{
+						throw new \InvalidArgumentException("$key is not a valid StringUtil method.");
+					}
+
+					return StringUtil::$key(...$params);
+				}
+			};
+		}
+
 		if (isset($this->arrData[$strKey]))
 		{
 			if (\is_object($this->arrData[$strKey]) && \is_callable($this->arrData[$strKey]))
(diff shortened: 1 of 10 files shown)


core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -150,6 +150,22 @@ public function __set($strKey, $varValue)
 	 */
 	public function __get($strKey)
 	{
+		if ('stringUtil' === $strKey)
+		{
+			return new class()
```
Toflar, line 155: I guess we could ensure we do not create a new instance for every call but let's see for general feedback first.

aschempp:
I like (except for the instance issue) 😊

m-vo:
Regarding the "useful helpers": we could also try to be somewhat forward compatible to Twig and allow the use of any of the https://twig.symfony.com/doc/3.x/filters/u.html methods (which internally use Symfony's `UnicodeString`) instead of the dusty `StringUtil` methods. :wink: 

I don't think we really need to expose the `StringUtil` class more than we're doing now.

aschempp:
> The u filter is part of the StringExtension which is not installed by default.

?

m-vo:
>> The u filter is part of the StringExtension which is not installed by default.
>
>?

Are you questioning this or what do you mean? Twig's components are separated in extensions of which some are included in `twig/twig` and some in their own packages (`twig/intl-extra`, `twig/string-extra`, `twig/cssinliner-extra`, …).

ausi:
I don’t see the benefit here, `$this->stringUtil->` is exactly the same length as `Contao\StringUtil::`. It is “magic” and probably bad for IDE integration.

There might be use cases for methods that we need often (like `specialchars()`). But I think we should handle these separately and maybe come up with better method names. Instead of `specialchars()` I would use something like `$this->escape(…)` that could work the same way as the escape filter of twig does, like `$this->escape(…, 'html')`, `$this->escape(…, 'js')`.

PR #3119, opened 2021-06-24
Use schema.org JSON-LD in the news-bundle

Toflar:
Using the new means for the news-bundle :)

Note for @m-vo: I've considered having some of the data in the `NewsModel` so it can be reused by Twig but there's not really much to be reused here, really. The links require the router, `hasText` queries the DB...so I left it all to the template.

### DIFF
--- a/news-bundle/src/Resources/contao/modules/ModuleNews.php
+++ b/news-bundle/src/Resources/contao/modules/ModuleNews.php
@@ -11,7 +11,11 @@
 namespace Contao;
 
 use Contao\CoreBundle\Image\Studio\Studio;
+use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
+use Contao\CoreBundle\Routing\ResponseContext\ResponseContextAccessor;
 use Contao\Model\Collection;
+use Spatie\SchemaOrg\NewsArticle;
+use Spatie\SchemaOrg\Person;
 
 /**
  * Parent class for news modules.
@@ -230,6 +234,42 @@ abstract class ModuleNews extends Module
 			$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
 			$responseTagger->addTags(array('contao.db.tl_news.' . $objArticle->id));
 		}
+
+		// schema.org information
+		$schemaOrgId = '#/schema/news/' . $objArticle->id;
+		$responseContext = System::getContainer()->get(ResponseContextAccessor::class)->getResponseContext();
+
+		if ($responseContext && $responseContext->has(JsonLdManager::class))
+		{
+			/** @var JsonLdManager $jsonLdManager */
+			$jsonLdManager = $responseContext->get(JsonLdManager::class);
+
+			$article = new NewsArticle();
+			$article->headline(StringUtil::inputEncodedToPlainText($objTemplate->headline));
+			$article->url($objTemplate->link);
+			$article->datePublished(new \DateTime('@' . $objArticle->date));
+
+			if ($objTemplate->hasTeaser)
+			{
+				$article->description(StringUtil::htmlToPlainText($objTemplate->teaser));
+			}
+
+			if ($objTemplate->addImage && $objTemplate->figure)
+			{
+				$article->image($jsonLdManager->createSchemaOrgTypeFromArray($objTemplate->figure->getSchemaOrgData()));
+			}
+
+			if ($objTemplate->authorModel)
+			{
+				$article->author((new Person())->name($objTemplate->authorModel->name));
+			}
+
+			$jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_ORG)->set($article, $schemaOrgId);
+
+			dump($jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_ORG));
+		}
+
+		$objTemplate->schemaOrgId = $schemaOrgId;
 
 		return $objTemplate->parse();
 	}
--- a/news-bundle/src/Resources/contao/modules/ModuleNews.php
+++ b/news-bundle/src/Resources/contao/modules/ModuleNews.php
@@ -11,11 +11,7 @@
 namespace Contao;
 
 use Contao\CoreBundle\Image\Studio\Studio;
-use Contao\CoreBundle\Routing\ResponseContext\JsonLd\JsonLdManager;
-use Contao\CoreBundle\Routing\ResponseContext\ResponseContextAccessor;
 use Contao\Model\Collection;
-use Spatie\SchemaOrg\NewsArticle;
-use Spatie\SchemaOrg\Person;
 
 /**
  * Parent class for news modules.
@@ -236,40 +232,36 @@ abstract class ModuleNews extends Module
 		}
 
 		// schema.org information
-		$schemaOrgId = '#/schema/news/' . $objArticle->id;
-		$responseContext = System::getContainer()->get(ResponseContextAccessor::class)->getResponseContext();
-
-		if ($responseContext && $responseContext->has(JsonLdManager::class))
+		$objTemplate->getBasicSchemaOrgData = static function () use ($objTemplate, $objArticle): array
 		{
-			/** @var JsonLdManager $jsonLdManager */
-			$jsonLdManager = $responseContext->get(JsonLdManager::class);
-
-			$article = new NewsArticle();
-			$article->headline(StringUtil::inputEncodedToPlainText($objTemplate->headline));
-			$article->url($objTemplate->link);
-			$article->datePublished(new \DateTime('@' . $objArticle->date));
+			$jsonLd = array(
+				'@type' => 'NewsArticle',
+				'identifier' => '#/schema/news/' . $objArticle->id,
+				'url' => $objTemplate->link,
+				'headline' => StringUtil::inputEncodedToPlainText($objTemplate->headline),
+				'datePublished' => $objTemplate->datetime,
+			);
 
 			if ($objTemplate->hasTeaser)
 			{
-				$article->description(StringUtil::htmlToPlainText($objTemplate->teaser));
+				$jsonLd['description'] = StringUtil::htmlToPlainText($objTemplate->teaser);
 			}
 
 			if ($objTemplate->addImage && $objTemplate->figure)
 			{
-				$article->image($jsonLdManager->createSchemaOrgTypeFromArray($objTemplate->figure->getSchemaOrgData()));
+				$jsonLd['image'] = $objTemplate->figure->getSchemaOrgData();
 			}
 
 			if ($objTemplate->authorModel)
 			{
-				$article->author((new Person())->name($objTemplate->authorModel->name));
+				$jsonLd['author'] = array(
+					'@type' => 'Person',
+					'name' => $objTemplate->authorModel->name,
+				);
 			}
 
-			$jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_ORG)->set($article, $schemaOrgId);
-
-			dump($jsonLdManager->getGraphForSchema(JsonLdManager::SCHEMA_ORG));
-		}
-
-		$objTemplate->schemaOrgId = $schemaOrgId;
+			return $jsonLd;
+		};
 
 		return $objTemplate->parse();
 	}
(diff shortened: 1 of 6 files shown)


news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -305,7 +305,8 @@ protected function getMetaFields($objArticle)
 					/** @var UserModel $objAuthor */
 					if (($objAuthor = $objArticle->getRelated('author')) instanceof UserModel)
 					{
-						$return['author'] = $GLOBALS['TL_LANG']['MSC']['by'] . ' <span itemprop="author">' . $objAuthor->name . '</span>';
+						$return['author'] = $GLOBALS['TL_LANG']['MSC']['by'] . ' <span class="author">' . $objAuthor->name . '</span>';
```
leofeyer:
```suggestion
						$return['author'] = $GLOBALS['TL_LANG']['MSC']['by'] . ' ' . $objAuthor->name;
```

I don't think that we need the `<span>` at all then.

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -367,11 +368,11 @@ protected function generateLink($strLink, $objArticle, $blnAddArchive=false, $bl
 		$strArticleUrl = News::generateNewsUrl($objArticle, $blnAddArchive);
 
 		return sprintf(
-			'<a href="%s" title="%s"%s itemprop="url">%s%s</a>',
+			'<a href="%s" title="%s"%s>%s%s</a>',
 			$strArticleUrl,
 			StringUtil::specialchars(sprintf($strReadMore, $blnIsInternal ? $objArticle->headline : $strArticleUrl), true),
 			($objArticle->target && !$blnIsInternal ? ' target="_blank" rel="noreferrer noopener"' : ''),
-			($blnIsReadMore ? $strLink : '<span itemprop="headline">' . $strLink . '</span>'),
+			($blnIsReadMore ? $strLink : '<span class="headline">' . $strLink . '</span>'),
```
leofeyer:
```suggestion
			$strLink,
```

leofeyer:
Now the `news_full` template is the only one with schema.org data, which means no more structured data in e.g. the news list. Is this ok?

fritzmg:
> Now the `news_full` template is the only one with schema.org data, which means no more structured data in e.g. the news list. Is this ok?

Hm, not for news where `source` is not `'default'` 🤔 😕

Toflar:
> Is this ok?

Yes, this is okay. `schema.org` data describes entities, not lists. The list microdata was nonsense ever since :)

news-bundle/src/Resources/contao/modules/ModuleNews.php
Toflar: I thought people were maybe using `itemprop="author"` as a CSS selector so I did not want to take away the possibility to still select that.

news-bundle/src/Resources/contao/modules/ModuleNews.php
Toflar: I thought people were maybe using `itemprop="headline"` as a CSS selector so I did not want to take away the possibility to still select that.

fritzmg:
But this means no microdata will be available for news or events that do not have a detail page, i.e. that only link to other URLs. That would be bad especially for events.

news-bundle/src/Resources/contao/modules/ModuleNews.php
leofeyer: I highly doubt that. And even if they did, this markup was never intended to be used for styling, so it is OK to remove it IMHO.

Toflar:
Yes, that's what this means and that should be the default use case. Entities need identifiers which is an URL to a single entity. 
As I said, our defaults in the core currently are nonsense. It does not make sense to output anything if you have a reader page. You should describe the entity only and ideally you do this on the reader page. This should be our default.

You should not add schema.org data of things that are not visible to the user, nor should you describe lists.
Sources:

- https://developers.google.com/search/docs/guides/intro-structured-data

> You should not create blank or empty pages just to hold structured data; nor should you add structured data about information that is not visible to the user, even if the information is accurate. 

- https://developers.google.com/search/docs/data-types/product

> Use markup for a specific product, not a category or list of products. For example, "shoes in our shop" is not a specific product. Currently, product rich results only support pages that focus on a single product. We recommend focusing on adding markup to product pages instead of pages that list products or a category of products.

Toflar:
Except for `news_full` these are all list templates, right? We could use `if (!$this->hasText)` and output basic information in this case for them.

fritzmg:
> Yes, that's what this means and that should be the default use case. Entities need identifiers which is an URL to a single entity.
> As I said, our defaults in the core currently are nonsense. It does not make sense to output anything if you have a reader page. You should describe the entity only and ideally you do this on the reader page. This should be our default.
>
> You should not add schema.org data of things that are not visible to the user, nor should you describe lists.
Sources:

But an event of an event calendar _is_ visible to the user, regardless of whether that event has its own detail page or not. Google pulls this information and shows available events in the future to the user for your website, when applicable for the search query. These events would suddenly be gone. Also the detail page could still be on your own website - the details are just not built within the event itself. Technically, this page would then have to contain the metadata.

Toflar:
As I said, this only makes sense if you do not have a detail page.

fritzmg:
Ah, so you mean we should execute `$this->addSchemaOrg()` in the other templates too, if `$this->source !== 'default'`?

Toflar:
Yes, see https://github.com/contao/contao/pull/3119/commits/8f9484b9e0362c0c1a9f85fde4dbb7966b63ab14 for an example.
There are 2 differences to the reader:

- We do not use the reader URL as identifier because it does not exist, so we use a relative schema URL like we do for our images.
- We do not add text as there is none :)

m-vo:
If I had a news list redirecting to an external source (where the actual content lives) - I wouldn't want metadata for this, would I?

Toflar:
I'd adjust the remaining 2 templates if that's the way to go for you too (I think it's correct like that).

I have a news archive with 2 entries now. One points to an external source. I now have 1 schema.org entry on the list view and the other one which is more detailed is on the detail URL of this page.

Toflar:
> If I had a news list redirecting to an external source (where the actual content lives) - I wouldn't want metadata for this, would I?

Of course? The meta data is the one of your news entry.

fritzmg:
lgtm 👍  🙂 

On that note I was wondering though: what if the module (news, event, etc.) would handle adding the schema metadata instead and in the template you could override it, if you want? This way you would only have to do this in one place (`ModuleNews::parseArticle` for news for example) and we would not have to clutter the templates as much - while still giving the possibility of changing the metadata in templates. Not sure if that's feasible?

Toflar:
I gave it a try in https://github.com/contao/contao/pull/3119/commits/a5a1b12d348170d1a1ea1a36c0b524dfc15ef4a7.

The basic JSON-LD data is now added by default. The list template can remove the ones that do not have a reader page and the reader template can extend an existing entry.

news-bundle/src/Resources/contao/templates/news/news_short.html5
```diff
@@ -16,3 +16,12 @@
...
 
 </div>
+
+<?php
+
+// This template is used as a news list template by default
+// Ensure that there's no JSON-LD for this news article if there is a reader as that one will
+// list all the details.
+if ($this->hasReader) {
+    $this->removeSchemaOrg('NewsArticle', $this->schemaOrgId);
```
leofeyer: Does this not have to be added to `news_latest` and `news_simple` as well? They are news list templates, too.

Toflar: Yes, I said it's not finished?

Toflar: https://github.com/contao/contao/pull/3119#issuecomment-868383748

Toflar:
> The basic JSON-LD data is now added by default. The list template can remove the ones that do not have a reader page and the reader template can extend an existing entry.

I don't like this to be honest, @fritzmg and @m-vo. It introduces a dependency between the business logic and the template itself. As a designer that modifies the template, I now have to know what the module does.

I'd prefer to keep the `addSchmemaOrg()` only and not adding anything in the `ModuleNews` itself. However, we can provide a `getBasicSchemaOrgData()` so all the news templates can reuse that.

Toflar:
Done :) The current state is what I would go with. This provides the most flexibility and is easy to understand for template editors.

fritzmg:
> I don't like this to be honest, @fritzmg and @m-vo. It introduces a dependency between the business logic and the template itself. As a designer that modifies the template, I now have to know what the module does.

Well my thinking is, that the module already sets other meta data like the description and page title and thus I thought it would make sense if the module itself also handles the schema data - while being able to alter it in a template.

Toflar:
Yeah, as I said, this is a bad idea because it requires users to understand what the module does. The current solution is way better.

ausi (dismissed):

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -229,6 +231,38 @@ protected function parseArticle($objArticle, $blnAddArchive=false, $strClass='',
 			$responseTagger->addTags(array('contao.db.tl_news.' . $objArticle->id));
 		}
 
+		// schema.org information
+		$objTemplate->getBasicSchemaOrgData = static function () use ($objTemplate, $objArticle): array
```
leofeyer, line 235: Should we move this above the `parseArticles` hook so people can replace the function if they want?

Toflar: I don't think they should. You can modify the JSON-LD in the central `JsonLdEvent` and modify whatever you want which is exactly why I have designed the JsonLdManager to collect and only then display it :)

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -229,6 +231,38 @@ protected function parseArticle($objArticle, $blnAddArchive=false, $strClass='',
...
+		{
+			$jsonLd = array(
+				'@type' => 'NewsArticle',
+				'identifier' => '#/schema/news/' . $objArticle->id,
+				'url' => $objTemplate->link,
```
leofeyer, line 240: This will be a relative URL such as `en\/news-detail\/new-contao-versions-in-short-intervalls.html`. It this correct or do we have to convert it to an absolute URL instead?

Toflar: I went for relative because we also do this for `ImageObject` currently. For schema.org it doesn't matter.

news-bundle/src/Resources/contao/templates/news/news_latest.html5
```diff
@@ -20,3 +20,11 @@
   <?php endif; ?>
 
 </div>
+
+<?php
+
+// This template is used as a news list template by default, so we only add
+// JSON-LD data in case this is a news article without a reader
+if (!$this->hasReader) {
```
leofeyer, line 28: Why did you not use `if (false === $this->hasText) {`? Then we would not have to add a new variable.

Toflar: Because for external links, `hasText` is `true` too. But for us, only `$objArticle->source === 'default'` is relevant.

leofeyer: Ok, so why are we not checking `if ($this->source != 'default')` then? Or at least let us rename the variable to `$this->isExternal` or something alike, so it becomes more clear what the if condition does.

Toflar: Well, `$this->source` does not exist?

leofeyer: It should exist:

https://github.com/contao/contao/blob/7d70d76d27555041bcf297d42d5737b0b377735e/news-bundle/src/Resources/contao/modules/ModuleNews.php#L78-L79

Toflar: Ah true. Well...yeah. We could. I don't really mind. I find `hasReader` tells me more than "the default source" :)

leofeyer: What about my `$this->isExternal` suggestion above? I find `hasReader` not very clear.

Toflar: But `isExternal` is not true? We want to know which ones have a reader page.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3120, opened 2021-06-24
Beautify the JSON-LD by sorting by key

Toflar:
This will make the generated JSON-LD output a lot more readable. The method looked like it would be a generally useful utility method so I decided to put that into `ArrayUtil`.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/ArrayUtil.php
+++ b/core-bundle/src/Resources/contao/library/Contao/ArrayUtil.php
@@ -44,6 +44,19 @@ class ArrayUtil
 		array_splice($arrCurrent, $intIndex, 0, $arrNew);
 	}
 
+	public static function recursiveKeySort(array &$array): void
+	{
+		foreach ($array as &$value)
+		{
+			if (\is_array($value))
+			{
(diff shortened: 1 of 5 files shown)


core-bundle/src/Resources/contao/library/Contao/ArrayUtil.php
```diff
@@ -44,6 +44,19 @@ public static function arrayInsert(&$arrCurrent, $intIndex, $arrNew): void
 		array_splice($arrCurrent, $intIndex, 0, $arrNew);
 	}
 
+	public static function recursiveKeySort(array &$array): void
```
m-vo, line 50: Could this also replace this?
 https://github.com/contao/contao/blob/3a1c6e8824322d76a05e188ad7fa0f2fc0ea2aaa/core-bundle/tests/Image/Studio/FigureBuilderIntegrationTest.php#L1681-L1689

Toflar: Sure, done :)

fritzmg (dismissed):

leofeyer:
Thank you @Toflar.

PR #3124, opened 2021-06-25
Use a security voter to check the member groups

aschempp:
Instead of using the user object to check group membership, we should use a security voter for the case.

---

Background: I'm setting up a system where users login through Microsoft Identity Services (Microsoft OAuth/OpenID). Authentication works well but I need to protect content based on Contao member groups. This PR allows me to decorate the voter and dynamically add my Microsoft OAuth Security User to have Contao groups 😎


Toflar (dismissed):

fritzmg (dismissed):

leofeyer:
I don't understand the `$security && $security->isGranted('ROLE_MEMBER')`. Does `$container->get('security.helper')` return `null` if no user is logged in?

aschempp:
> I don't understand the `$security && $security->isGranted('ROLE_MEMBER')`. Does `$container->get('security.helper')` return `null` if no user is logged in?

I guess it could be `null` if security is not active. I added these checks because `Container::get()` might return null according to the return type. I don't think it will ever be null in our case.

fritzmg:
Btw. since all the other permission strings are constants, may be add

```php
    /**
     * Access is granted if the current user is a member of any of the given groups.
     */
    public const USER_IN_GROUP = 'contao_group';
```

to `ContaoCorePermissions`.

Toflar (dismissed):

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -232,11 +234,11 @@ public function renderPage($pageModel)
 		// Set the admin e-mail address
 		if ($objPage->adminEmail)
 		{
-			list($GLOBALS['TL_ADMIN_NAME'], $GLOBALS['TL_ADMIN_EMAIL']) = StringUtil::splitFriendlyEmail($objPage->adminEmail);
+			[$GLOBALS['TL_ADMIN_NAME'], $GLOBALS['TL_ADMIN_EMAIL']] = StringUtil::splitFriendlyEmail($objPage->adminEmail);
```
fritzmg: These changes seem unrelated?

aschempp: Fixed in 0eb40934e9773ee44d8604d25570bf918bc8e290. Somehow my IDE started auto-fixing this stuff recently 🙄

fritzmg:
I think we can omit the `!$isMember` check as a lot of services in Contao already have a hard dependency on `@security.helper` and therefore problems would arise elsewhere anyway, if the container does not contain the `security.helper` service.

// wait, when a service does not exist, it would throw an exception anyway. Why would it ever be `null`?

core-bundle/src/Security/Voter/GroupVoter.php
```diff
@@ -0,0 +1,42 @@
...
+        return is_numeric($subject)
+            || (\is_array($subject) && \count($subject) === \count(array_filter($subject, 'is_numeric')));
+    }
+
+    protected function voteOnAttribute($attribute, $subject, TokenInterface $token): bool
+    {
+        $user = $token->getUser();
+
+        if (!$user instanceof User) {
+            return false;
```
ausi: Couldn’t we check the `-1` group here so that we don’t have to do that in all the other places?

```suggestion
            return in_array(-1, array_map('intval', (array) $subject), true);
```

aschempp: Thats not the same at all. Calling `is this user in group 1,2,3` shouldn't return true if I don't have a user (and not checking the groups at all). Being a guest means not having `ROLE_MEMBER`, not _not_ having a Contao user.

ausi: Should it be something like this then?

```php
if (!$user instanceof User || !\in_array('ROLE_MEMBER', $token->getRoleNames(), true)) {
    return in_array(-1, array_map('intval', (array) $subject), true);
}
```

aschempp: Handling a magic group `-1` is wrong imho in the first place, that's what I don't like about the change. As I said in #3088 I feel like that's gonna do more harm than good. But thats not related to this PR at all.

Anyway, calling `isGranted('contao_group', -1)` and expecting the system to return `true` if there's no user is wrong imho. Also, this voter is not supposed to vote on role names, it only votes on the known attribute. Voting on roles is done by the role voter, which you might override for any other reason.

ausi: OK, I see. So this voter is used for both, users (backend) and members (frontend)?

Wouldn’t it make sense to separate it into two voters, so that the voter for members can handle the `-1` case and the voter for users not?

aschempp: The voters we currently have vote on features we currently have. The `BackendAccessVoters` votes on `BackendUser::isAllowed` etc, the `GroupVoter` votes on `User::isMemberOf`.

I think what you're asking for would be a completely separate voter, and I'm not sure that's intended by the Symfony security system. Voters vote on one specific attribute, a voter is not supposed to make a final decision for multiple conditions (e.g. a user is not logged in AND an element is set to be visible to a guest only). It sure would work if we agree that would be necessary, but then it wouldn't be a different voter concept.

ausi: I still think we should split it up in a `BackendGroupVoter` and a `FrontendGroupVoter`, otherwise if someone forgets to vote on the role beforehand, the `GroupVoter` would return true if the IDs of the backend groups accidentaly match with the IDs of the frontend groups.

Also, to know if a frontend group matches, we have to do several checks instead of just one:

1. Check if there is a user
2. If not check for the `-1` group
3. Ask the voter for `ROLE_MEMBER`
4. If not check for the `-1` group
5. Ask the voter for the group

leofeyer: I second this.

m-vo:
> Why would it ever be null?

You can control the behavior (throwing vs. returning `null`) via the second argument.

The type hint just cannot know this.

fritzmg:
> You can control the behavior (throwing vs. returning `null`) via the second argument.
> 
> The type hint just cannot know this.

Which argument? `ContainerInterface::get` only has one argument.

// edit: ah I was looking at the Psr interface, rather than the one from Symfony. But we do not use the second parameter and by default an exception will be thrown, hence we can omit the null check.

aschempp:
> ah I was looking at the Psr interface, rather than the one from Symfony. But we do not use the second parameter and by default an exception will be thrown, hence we can omit the null check.

Yes we could, but my IDE complains about a missing null-check then 😉

core-bundle/src/Security/Voter/GroupVoter.php
aschempp: again, it will no longer be a group voter then. It will be a sort of "has access voter". Also, again, checking if there is user is not done by checking an interface but for the role, and thats not supposed to be in this voter.

ausi: > It will be a sort of "has access voter".

Similar to the `BackendAccessVoters` should we name them `BackendGroupAccessVoter` and `FrontendGroupAccessVoter` then?

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -263,23 +265,23 @@ public function renderPage($pageModel)
...
-			{
-				$this->log('Page ID "' . $objPage->id . '" can only be accessed by groups "' . implode(', ', (array) $objPage->groups) . '" (current user groups: ' . implode(', ', $this->User->groups) . ')', __METHOD__, TL_ERROR);
+				if (!$user instanceof FrontendUser)
+				{
+					throw new AccessDeniedException('Access denied: ' . Environment::get('uri'));
+				}
+
+				$this->log('Page ID "' . $objPage->id . '" can only be accessed by groups "' . implode(', ', $objPage->groups) . '" (current user groups: ' . implode(', ', StringUtil::deserialize($user->groups, true)) . ')', __METHOD__, TL_ERROR);
 
 				throw new AccessDeniedException('Access denied: ' . Environment::get('uri'));
```
ausi:
```suggestion
				if ($user instanceof FrontendUser)
				{
					$this->log('Page ID "' . $objPage->id . '" can only be accessed by groups "' . implode(', ', $objPage->groups) . '" (current user groups: ' . implode(', ', StringUtil::deserialize($user->groups, true)) . ')', __METHOD__, TL_ERROR);
				}

				throw new AccessDeniedException('Access denied: ' . Environment::get('uri'));
```

Would remove the need for the duplicated throw ☺️

ausi (dismissed):

aschempp:
As discussed on Mumble yesterday, I have updated the PR
1. we no longer vote on back end groups, since that never makes sense. There are no permissions applicable to backend groups, permissions are inherited are groups should never be checked.
2. The frontend group voter can now vote on our new permission system, where a group `-1` means guests (`! instanceof FrontendUser`) have access to an element.

We also discussed whether the voter should check for `ROLE_MEMBER` etc. However we agreed that this should never be necessary. If an element is protected for certain groups only, checking the group permission is the only necessary authorization step. We **must** check `ROLE_MEMBER` only when we want to know if a user is logged in (regardless of the groups).

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
aschempp: Changed in 7322e53160c58bab70644a88f6239dbfa8474d08

fritzmg (dismissed):

core-bundle/src/Security/Voter/MemberGroupVoter.php
```diff
@@ -0,0 +1,43 @@
...
+            || (\is_array($subject) && \count($subject) === \count(array_filter($subject, 'is_numeric')));
+    }
+
+    protected function voteOnAttribute($attribute, $subject, TokenInterface $token): bool
+    {
+        $user = $token->getUser();
+        $groups = array_map('intval', (array) $subject);
+
+        if (!$user instanceof FrontendUser) {
+            return \in_array(-1, $groups, true);
```
leofeyer, line 38: This will most likely fail, because it might as well be `'-1'` (string).

leofeyer (changes requested):

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -263,23 +265,21 @@ public function renderPage($pageModel)
...
-			$container = System::getContainer();
-			$token = $container->get('security.token_storage')->getToken();
+			$security = System::getContainer()->get('security.helper');
 
-			if ($container->get('security.authentication.trust_resolver')->isAnonymous($token))
+			if (!$security->isGranted(ContaoCorePermissions::MEMBER_IN_GROUPS, $objPage->groups))
 			{
-				throw new InsufficientAuthenticationException('Not authenticated: ' . Environment::get('uri'));
-			}
+				if ($security->getToken() instanceof AnonymousToken)
```
leofeyer, line 272: This was previously `$container->get('security.authentication.trust_resolver')->isAnonymous($token)`. Guess it does not matter, does it?

core-bundle/src/Security/Voter/MemberGroupVoter.php
fritzmg: It will be an integer. See

```
$groups = array_map('intval', (array) $subject);
```

a few lines above.

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -720,22 +721,17 @@ public static function isVisibleElement(Model $objElement)
...
 			{
 				$groups = StringUtil::deserialize($objElement->groups, true);
-				$blnReturn = (!$user && \in_array(-1, $groups)) || ($user && $user->isMemberOf($groups));
+				$blnReturn = $security->isGranted(ContaoCorePermissions::MEMBER_IN_GROUPS, $groups);
 			}
 			elseif ($objElement->guests)
 			{
 				trigger_deprecation('contao/core-bundle', '4.12', 'Using the "show to guests only" feature has been deprecated an will no longer work in Contao 5.0. Use the "protect page" function instead.');
-				$blnReturn = !$user; // backwards compatibility
+				$blnReturn = $security->getUser() instanceof FrontendUser; // backwards compatibility
```
leofeyer: @aschempp Is this correct?

aschempp: no, never check the object, always check the roles.

leofeyer: Fixed in 9c289eb0a590fb2790fa986f0da1a3d2f2783f47.

leofeyer (approved):

PR #3125, opened 2021-06-28
Check if tstamp exists before hiding unsaved elements

aschempp:
If a content model is manually created to render an element, and the tstamp is not set, it should still be possible to generate the element in the front end. 

Origin PR in Contao: #2994
Fix in Isotope: https://github.com/isotope/core/commit/857c15bda0376c5b03c2d87f7d82e7b81703ac6b

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Hybrid.php
+++ b/core-bundle/src/Resources/contao/classes/Hybrid.php
@@ -264,7 +264,7 @@ abstract class Hybrid extends Frontend
 		}
 
 		// Skip unsaved elements (see #2708)
-		if (!$this->objParent->tstamp)
+		if (isset($this->objParent->tstamp) && !$this->objParent->tstamp)
 		{
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/classes/Hybrid.php
```diff
@@ -264,7 +264,7 @@ protected function isHidden()
 		}
 
 		// Skip unsaved elements (see #2708)
-		if (!$this->objParent->tstamp)
+		if (isset($this->objParent->tstamp) && !$this->objParent->tstamp)
```
leofeyer, line 267:
```suggestion
		if (empty($this->objParent->tstamp))
```

Would this work as well?

aschempp: I don't think so, not having a value would be true, but that's not what we want?

leofeyer:
I don't understand how this could happen. The column has a default value, so it should always be set to `0` as soon as the model is saved:

https://github.com/contao/contao/blob/7d70d76d27555041bcf297d42d5737b0b377735e/core-bundle/src/Resources/contao/dca/tl_content.php#L185

fritzmg:
It happens when you instantiate a content element on the fly and you do not set the `tstamp`. It is something that we used quite a lot across different Contao installations.

aschempp:
@leofeyer see the linked Isotope code to understand the problem 😉

PR #3126, opened 2021-06-28
Give hints how to create suitable SQL-table

ACCakut:
Since the description and official documentation is quite rough and not very clear, I suggest to add the instructions directly where they are needed.

PR for German first to start a discussion.

### DIFF
--- a/core-bundle/src/Resources/contao/languages/de/tl_form.xlf
+++ b/core-bundle/src/Resources/contao/languages/de/tl_form.xlf
@@ -143,7 +143,7 @@
       </trans-unit>
       <trans-unit id="tl_form.targetTable.1">
         <source>The target table must contain a column for every form field.</source>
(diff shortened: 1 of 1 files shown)


fritzmg:
I think this would be better suited for the [official documentation](https://github.com/contao/docs/). Also if I were to create a custom table to save form field data, I would just create a DCA for it and not execute SQL directly.

m-vo:
Maybe the field could get a Contao short url link pointing to the relevant part in the docs, though?

ACCakut:
#3144 #3145 

I made a PR for the docs and these PR to refer to them.

Speaking for me (who is not really into Symfony and Contao development) creating a DCA is too much effort for a little form. The path `contao/dca/` does not even exist on my Contao 4.11 and the examples from the docs are about how to add columns to existing tables or create some fancy tables linked to each other.

fritzmg:
> The path `contao/dca/` does not even exist on my Contao 4.11

See https://docs.contao.org/dev/getting-started/starting-development/#structure

> When customizing your web application, the following files and folders will usually be of interest. __Some of those will need to be created manually:__

m-vo:
> Speaking for me (who is not really into Symfony and Contao development) creating a DCA is too much effort for a little form.

There is the `terminal42/contao-leads` extension you might use instead. Creating tables without entites or dca's or adjusting the schema filtering does not work well together with the automated schema update we're providing.

leofeyer:
Closing this in favor of #3144.

PR #3130, opened 2021-06-29
Add more null checks for DCA lang references

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

We adjusted all `$GLOBALS['TL_LANG']['MSC']['deleteConfirm']` references to include a null check but not the `deleteConfirmFile` ones. 

Is this intended? Just stumbled across this by accident. I did not experience an error yet, but there likely is a way to get one somehow.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_files.php
+++ b/core-bundle/src/Resources/contao/dca/tl_files.php
@@ -116,7 +116,7 @@ $GLOBALS['TL_DCA']['tl_files'] = array
 			(
 				'href'                => 'act=delete',
 				'icon'                => 'delete.svg',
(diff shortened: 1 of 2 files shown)


ausi (dismissed):

bytehead (dismissed):

leofeyer:
Shouldn't this be fixed in Contao 4.11, too?

m-vo:
>Shouldn't this be fixed in Contao 4.11, too?

You are right, I'll change the base.

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #3131, opened 2021-06-29
Add warning if database server is not running in strict mode

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | Todo

Running a MySQL server without strict mode (no `TRADITIONAL`, no  `STRICT_ALL_TABLES` or `STRICT_TRANS_TABLES`) is bad practice and can lead to unnoticed loss of data, e.g. when data is being coerced to other types. 

We should IMHO show a warning and maybe suggest a fix:

image

(Note: Currently implemented with an error that would probably need a config to suppress it, so that it prevents migrations from running, but that's debatable if this is a good idea.)


leofeyer:
There are a lot of different DB configurations out in the wild and not all of them run in strict mode, therefore we should probably not enforce this just because it is "best practice". I can give you 100 more examples of what would be "best practice" but is done differently in practice. 🙈

ausi:
> I can give you 100 more examples of what would be "best practice" but is done differently in practice. 🙈

I think the database is a pretty central part of Contao and we already require some settings for charset, engine, innodb_large_prefix, innodb_file_per_table and innodb_file_format in some cases.

As the SQL mode can be changed only for the session (in contrast to require a global server change) and it can be fixed by changing the doctrine config, I think we should enforce it with Contao 4.12 and show a warning/deprecation in Contao 4.9+.

leofeyer:
What exactly is the risk of not running in strict mode?

installation-bundle/src/Resources/views/configuration_error.html.twig
```diff
@@ -63,6 +63,25 @@
...
+  dbal:
+    connections:
+      default:
+        options:
+          {% if driver == 'mysqli' -%}
+          # Define a MYSQLI_INIT_COMMAND
+          1002: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
+          {% else -%}
+          # Define a MYSQL_ATTR_INIT_COMMAND
+          3: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
```
ausi:
```suggestion
          3: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
          {% else -%}
          # Define a PDO::MYSQL_ATTR_INIT_COMMAND
          1002: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
```
I think this is the other way round?

m-vo: Thanks! You are right of course.

ausi:
> What exactly is the risk of not running in strict mode?

Corrupt, wrong or truncated data. 
Not noticing SQL queries or updates that go totally wrong.

For example, the following works in non-strict mode:

```sql
INSERT INTO tl_log (id, tstamp) VALUES ('not-an-integer', '2020-20-20');
```

What would you guess ends up as data in the `id` and `tstamp` columns? ☺️

Strict mode is also the default since MySQL version 5.7.5.

m-vo:
>  I think we should enforce it with Contao 4.12 and show a warning/deprecation in Contao 4.9+.

Sounds sensible to me. :+1: If we're fine with this, I'll provide a second PR for 4.x.

We could render a note above the `<main>` block in 4.9 I guess. This won't prevent migrations from running before you see the message, but I think we can live with this.

ausi:
>  This won't prevent migrations from running before you see the message, but I think we can live with this.

Agree. IMO the main advantage of adding the warning to 4.9 is, that LTS users get notified about this early and can react to it.

m-vo:
This now reports a warning instead of an error. I had to refactor the error message templates a bit so that they become reusable. Now, they are displayed on top of the main section. Let me know what you think.

image

installation-bundle/src/Resources/views/_configuration_errors.html.twig
```diff
@@ -0,0 +1,78 @@
...
+    connections:
+      default:
+        options:
+          {% if driver == 'mysqli' -%}
+              # Define a MYSQLI_INIT_COMMAND
+              3: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
+          {%- else -%}
+              # Define a PDO::MYSQL_ATTR_INIT_COMMAND
+              1002: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
+          {%- endif %}</pre>
```
ausi: The whitespace is incorrect I think (can also be seen in your screenshot)

```suggestion
        {%- if driver == 'mysqli' %}
          # Define a MYSQLI_INIT_COMMAND
          3: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
        {%- else %}
          # Define a PDO::MYSQL_ATTR_INIT_COMMAND
          1002: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
        {%- endif %}</pre>
```

m-vo: Thanks 😊

core-bundle/src/Resources/contao/themes/flexible/install.css
```diff
@@ -137,6 +142,7 @@ p {
 .tl_info {
 	color:#006494;
 	background:url("icons/show.svg") no-repeat 1px 3px;
+	background:url("icons/show.svg") no-repeat 1px 3px;
```
ausi: Typo?

ausi (dismissed):

core-bundle/src/Resources/contao/themes/flexible/install.css
m-vo: I wanted to make extra sure this CSS is in there, so I doubled the line.


typo for sure  :smile:

ausi (dismissed):

bytehead (dismissed):

installation-bundle/src/Resources/views/_configuration_errors.html.twig
```diff
@@ -0,0 +1,78 @@
...
+  dbal:
+    connections:
+      default:
+        options:
+        {%- if driver == 'mysqli' %}
+          # Define a MYSQLI_INIT_COMMAND
+          3: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
+        {%- else %}
+          # Define a PDO::MYSQL_ATTR_INIT_COMMAND
+          1002: "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
```
m-vo, line 75: We might want to swap out `TRADITIONAL` with `STRICT_TRANS_TABLES` to be less strict (and cause less friction with extensions and the like).

See MySQL docs:
> Before MySQL 5.7.4, and in MySQL 5.7.8 and later, TRADITIONAL is equivalent to STRICT_TRANS_TABLES, STRICT_ALL_TABLES, NO_ZERO_IN_DATE, NO_ZERO_DATE, ERROR_FOR_DIVISION_BY_ZERO, NO_AUTO_CREATE_USER, and NO_ENGINE_SUBSTITUTION.

> From MySQL 5.7.4 though 5.7.7, TRADITIONAL is equivalent to STRICT_TRANS_TABLES, STRICT_ALL_TABLES, NO_AUTO_CREATE_USER, and NO_ENGINE_SUBSTITUTION. The NO_ZERO_IN_DATE, NO_ZERO_DATE, and ERROR_FOR_DIVISION_BY_ZERO modes are not named because in those versions their effects are included in the effects of strict SQL mode (STRICT_ALL_TABLES or STRICT_TRANS_TABLES). Thus, the effects of TRADITIONAL are the same in all MySQL 5.7 versions (and the same as in MySQL 5.6). For additional discussion, see SQL Mode Changes in MySQL 5.7.

ausi: I’d vote for keeping `TRADITIONAL` (as the suggestion), that would make the behavior across installations very equal. And we are only requiring `STRICT_TRANS_TABLES` to not show the warning, right?

installation-bundle/src/Resources/translations/ContaoInstallationBundle.en.xlf
```diff
@@ -266,6 +266,15 @@
       <trans-unit id="88" resname="misconfigured_database_url_explain">
         <source>Could not connect to the database using the connection URL found in the &lt;code&gt;DATABASE_URL&lt;/code&gt; environment variable.</source>
       </trans-unit>
+      <trans-unit id="89" resname="strict_sql_mode">
+        <source>Your database server does not run in strict mode.</source>
+      </trans-unit>
+      <trans-unit id="90" resname="strict_sql_mode_explain">
+        <source>Running without strict mode enabled is bad practice and can lead to unnoticed loss of data. It won't be supported anymore in Contao versions 4.12 and later. Please change your SQL mode globally or add the following command to your connection &lt;code&gt;options&lt;/code&gt; that will enable it on each connection:</source>
```
m-vo:
```suggestion
        <source>Running without strict mode enabled is bad practice and can lead to unnoticed loss of data. It won't be supported anymore in later Contao versions. Please change your SQL mode globally or add the following command to your connection &lt;code&gt;options&lt;/code&gt; that will enable it on each connection:</source>
```

m-vo:
With #3151 targeted against 4.13, this should give us enough time to react if any problems appear in the wild.

ausi:
Feedback from Slack https://contao.slack.com/archives/CJT61P8R1/p1625481391186000 the following code:

```
PDO::MYSQL_ATTR_INIT_COMMAND => "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))"
```

did work on all tested hostings/systems so far:

- hostingwerk
- Anexia
- netcup
- Uberspace
- WebhostOne
- EDIS
- world4you
- ALLINKL
- 1und1
- Strato V-SERVER mit PLESK
- Strato

leofeyer (changes requested):
I would prefer if the check was independent of the `hasConfigurationError()` method and the `configuration_error.html.twig` template. Not only will it require less changes and thus make them much easier to review, but it will also not mix up configuration errors (stopping the script execution) and warnings (not stopping the script execution).

m-vo:
> but it will also not mix up configuration errors (stopping the script execution) and warnings (not stopping the script execution).

True for this PR, but we want it to become an error later on, don't we? I thought this would make it easier to merge (upstream). :thinking:

leofeyer:
I see. Maybe we should merge #3151 first.

m-vo:
With #3151 + #3175 merged, do we still need this PR?

leofeyer:
I don't think we do, although both PRs have only been merged into Contao 4.12.

ausi:
> although both PRs have only been merged into Contao 4.12.

I think backporting the warning to 4.9 would make sense as many users use the LTS version AFAIK.

leofeyer:
Unfortunately, it is more difficult to implement in Contao 4.9 (see https://github.com/contao/contao/pull/3131#pullrequestreview-702912891).

ausi:
> Unfortunately, it is more difficult to implement in Contao 4.9 (see [#3131 (review)](https://github.com/contao/contao/pull/3131#pullrequestreview-702912891)).

But you are in favor of merging a backport if e.g. me or @m-vo provides a PR?

leofeyer:
Sure. 😄

leofeyer:
Closing this in favor of #3303.

PR #3134, opened 2021-06-30
Use the driver from the database URL if there is one

aschempp:
The current implementation has two bugs that are fixed with this PR:

1. when using a database URL, the driver should be taken from that URL, e.g. `mysqli://…`
2. when the parameters are used, the driver in config is not merged into the database URL

Pretty sure no 2. means setting the driver didn't work at all …


manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -403,7 +427,8 @@ private function getDatabaseUrl(ContainerBuilder $container): string
         }
 
         return sprintf(
-            'mysql://%s%s:%s%s',
+            '%s://%s%s:%s%s',
+            $useMysqli ? 'mysqli' : 'mysql',
```
leofeyer: Are these really the only two variants? What is the driver name for PostgreSQL for example?

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -334,25 +334,41 @@ private function addDefaultServerVersion(array $extensionConfigs, ContainerBuild
...
         foreach ($extensionConfigs as $extensionConfig) {
-            // Do not add PDO options if the selected driver is not pdo_mysql
-            if (isset($extensionConfig['dbal']['connections']['default']['driver']) && 'pdo_mysql' !== $extensionConfig['dbal']['connections']['default']['driver']) {
-                return $extensionConfigs;
+            if (isset($extensionConfig['dbal']['connections']['default']['driver'])) {
+                $driver = $extensionConfig['dbal']['connections']['default']['driver'];
+            }
+
+            if (isset($extensionConfig['dbal']['connections']['default']['url'])) {
+                $url = $container->resolveEnvPlaceholders($extensionConfig['dbal']['connections']['default']['url'], true);
```
ausi, line 360:
```suggestion
                $driver = parse_url($container->resolveEnvPlaceholders($extensionConfig['dbal']['connections']['default']['url'], true), PHP_URL_SCHEME) ?? $driver;
```

We should IMO just extract the driver and check for all aliases of `pdo_mysql`:  `mysql`, `mysql2` and `pdo_mysql` isself.

manager-bundle/src/ContaoManager/Plugin.php
aschempp: we don't support anything except mysql as far as I know (we didn't previously either).

manager-bundle/src/ContaoManager/Plugin.php
aschempp: we _could_ use `parse_url` to check the driver a few lines below, but not here. Since both driver and URL _could_ be defined in multiple extension configs, and URL always wins over driver, we need to know which one to use here. Also, not sure if `parse_url` in the loop is efficient (compared to just extracting the scheme once below).

manager-bundle/src/ContaoManager/Plugin.php
leofeyer: Not right now, however, we might in the future. Therefore if there is an easy way to extract the driver from the URL, we should use it.

aschempp: 🤷

leofeyer: IMHO we should just use `$extensionConfig['dbal']['connections']['default']['driver']` as driver. This should work with Doctrine: https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#driver

ausi: Something like this?

```suggestion
            array_reverse(array_filter(array_map(function ($extensionConfig) {
                return $extensionConfig['dbal']['connections']['default']['driver'] ?? null;
            }, $extensionConfigs))[0] ?? 'mysql',
```

aschempp:
So how do we solve this? The current implementation works with all our supported database platforms. To add support for all Doctrine drivers (https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#driver) we would need to know what these URLs look like, and add a lot of unnecessary works since we don't support it anyway? 🤷

ausi:
> we would need to know what these URLs look like

`$driver = parse_url(…, PHP_URL_SCHEME) ?? $driver;`  
and `in_array($driver, ['pdo_mysql', 'mysql', 'mysql2'])`  
should work for all platforms. No need to know how the other URLs would look like exactly I think.

leofeyer:
As discussed in Mumble on September 30th, we want to handle the PDO case explicitly (`in_array($driver, ['pdo_mysql', 'mysql', 'mysql2'])`) instead of assuming PDO whenever the driver is not `mysqli`.

aschempp:
I have updated the PR but now the tests fail. The URL now looks like this, not sure that's valid?
`pdo_mysql://root:…@localhost:3306/contao_test`

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -334,25 +334,41 @@ private function addDefaultServerVersion(array $extensionConfigs, ContainerBuild
...
             }
 
             // Do not add PDO options if custom options have been defined
+            // Since this is merged recursively, we don't need to check other configs
             if (isset($extensionConfig['dbal']['connections']['default']) && \array_key_exists('options', $extensionConfig['dbal']['connections']['default'])) {
```
ausi:
```suggestion
            if (\array_key_exists(\PDO::MYSQL_ATTR_MULTI_STATEMENTS, $extensionConfig['dbal']['connections']['default']['options'] ?? [])) {
```
I think we should check for the multi statements option here directly as many users now set the options to enable the strict mode.

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -403,7 +425,8 @@ private function getDatabaseUrl(ContainerBuilder $container): string
         }
 
         return sprintf(
-            'mysql://%s%s:%s%s',
+            '%s://%s%s:%s%s',
+            $driver,
```
ausi:
```suggestion
            str_replace('_', '-', $driver),
```

See https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url : 

> The scheme names representing the drivers are either the regular driver names (see below) with any underscores in their name replaced with a hyphen (to make them legal in URL scheme names) …

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -334,25 +334,41 @@ private function addDefaultServerVersion(array $extensionConfigs, ContainerBuild
...
             // Do not add PDO options if custom options have been defined
+            // Since this is merged recursively, we don't need to check other configs
             if (isset($extensionConfig['dbal']['connections']['default']) && \array_key_exists('options', $extensionConfig['dbal']['connections']['default'])) {
                 return $extensionConfigs;
             }
         }
 
+        // If URL is set it overrides the driver option
+        if (null !== $url) {
+            $driver = parse_url($url, PHP_URL_SCHEME);
```
ausi:
```suggestion
            $driver = str_replace('-', '_', parse_url($url, PHP_URL_SCHEME));
```

ausi:
> but now the tests fail.

Some of the tests need to be updated to expect `pdo-mysql://` instead of `mysql://`

manager-bundle/src/ContaoManager/Plugin.php
aschempp: if any option is set, we don't add our own options. I think that already solves this?

ausi: > if any option is set, we don't add our own options.

Which is wrong IMO because this now happens on nearly every installation because of the strict mode.

I think we should only skip our own options if the `PDO::MYSQL_ATTR_MULTI_STATEMENTS` is set.

aschempp: added in c3bd64236b7bec212f8758a0d5ef50a9915faa45

manager-bundle/src/ContaoManager/Plugin.php
aschempp: fixed in c3bd64236b7bec212f8758a0d5ef50a9915faa45

manager-bundle/src/ContaoManager/Plugin.php
aschempp: fixed in c3bd64236b7bec212f8758a0d5ef50a9915faa45

ausi (dismissed):
❤️

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -334,25 +334,42 @@ private function addDefaultServerVersion(array $extensionConfigs, ContainerBuild
...
+            }
+
+            if (isset($extensionConfig['dbal']['connections']['default']['url'])) {
+                $url = $container->resolveEnvPlaceholders($extensionConfig['dbal']['connections']['default']['url'], true);
             }
 
             // Do not add PDO options if custom options have been defined
-            if (isset($extensionConfig['dbal']['connections']['default']) && \array_key_exists('options', $extensionConfig['dbal']['connections']['default'])) {
+            // Since this is merged recursively, we don't need to check other configs
+            if (isset($extensionConfig['dbal']['connections']['default']['options'][\PDO::MYSQL_ATTR_MULTI_STATEMENTS])) {
```
leofeyer: Why is this not the first check in the foreach loop? We do not have to set `$driver` and `$url` if we return early anyway.

leofeyer (changes requested):

leofeyer (approved):

ausi (approved):

PR #3136, opened 2021-06-30
Make sure that deserialized groups are always an array

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/pull/3088#discussion_r661375378
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -313,9 +313,9 @@ abstract class Module extends Frontend
 			}
 
 			$subitems = '';
-			$groups = StringUtil::deserialize($objSubpage->groups, true);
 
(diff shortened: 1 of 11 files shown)


ausi (approved):

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -553,7 +553,8 @@ static function (array $row) use ($blnFeUserLoggedIn): ?array
 			{
 				$page = PageModel::findByPk($row['id']);
```
aschempp, line 554: maybe we could use `findWithDetails()` instead and skip the call below?

aschempp (approved):

core-bundle/src/Resources/contao/modules/Module.php
leofeyer: We could, however, this way `loadDetails()` is only called if `$blnFeUserLoggedIn` is true, which might be a tiny little bit more performant (although I doubt that anyone would notice the difference).

PR #3137, opened 2021-06-30
Fix the functional tests

leofeyer:


aschempp:
I've debugged this as well, and the problem is that creating the kernel (before it is booted) somehow tries to load a class that does not exist (most like due to https://github.com/FriendsOfPHP/proxy-manager-lts/blob/1.x/src/ProxyManager/GeneratorStrategy/GeneratorStrategyInterface.php#L9, though that code isn't new). Since the regular autoloader cannot load the class, our autoloader is triggered

Instead of "silently" ignoring errors, it would be better to only access `kernel.project_dir` if we actually have a class (inside the if-conditions where `$projectDir` is currently used). This fixes the problem for me.

leofeyer:
Changed in 568de8869af3d364ae0c57322bc812c0bb094c24.

m-vo (approved):

PR #3138, opened 2021-06-30
Use intl for languages and countries

### LINKED ISSUE #3104: Deprecate getLanguages hook
In many new places we only use the `contao.locales` parameter instead of `System::getLanguages()`. So we should IMO deprecate the `getLanguages` hook as it only affects some parts of Contao.

https://github.com/contao/contao/blob/b14a17ea30c1cf8caa351afade6d4154afe1eeb1/core-bundle/src/Resources/contao/library/Contao/System.php#L654

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3104
| Docs PR or issue | contao/docs#785

Deprecates the hooks `getCountries` and `getLanguages` as well as the `System::getCountries()` method.

I used Symfonys Intl component because with `Countries::getNames` we get a list of “actual” countries instead of other regions (like world, EU,…) and I did not find out if there is an easy way to get this from the `intl` library itself.


aschempp:
Unfortunately, country names are not only a standard but also a geopolitical topic. I have created websites in the past where the client requested `Tibet` to be a country option or `Taiwan` not being called a [`Province of China`](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes). I also fear the list of countries will be outdated (e.g. `South Sudan` missing) if the server is not updated properly. 

I agree Intl would be an ideal source for the country names etc, but we must provide a way to patch the Intl list of countries for specific use cases.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -575,31 +575,37 @@ public static function isInstalledLanguage($strLanguage)
...
-		foreach (array_keys($arrAux) as $strKey)
-		{
-			$return[$strKey] = $GLOBALS['TL_LANG']['CNT'][$strKey] ?? $countries[$strKey];
-		}
+		(new \Collator($locale ?? 'en'))->asort($return);
 
 		// HOOK: add custom logic
 		if (isset($GLOBALS['TL_HOOKS']['getCountries']) && \is_array($GLOBALS['TL_HOOKS']['getCountries']))
 		{
+			trigger_deprecation('contao/core-bundle', '4.12', 'Using the "getCountries" hook has been deprecated and will no longer work in Contao 5.0.');
```
fritzmg: Do we really want to deprecate this fully? I think it's worth keeping the possibility to change the list of countries in a central place.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -618,41 +624,57 @@ public static function getCountries()
...
+			},
+			array_combine($locales, $locales)
+		);
+
+		asort($return);
 
 		// HOOK: add custom logic
 		if (isset($GLOBALS['TL_HOOKS']['getLanguages']) && \is_array($GLOBALS['TL_HOOKS']['getLanguages']))
 		{
+			trigger_deprecation('contao/core-bundle', '4.12', 'Using the "getLanguages" hook has been deprecated and will no longer work in Contao 5.0. Extend the contao.locales parameter instead.');
```
fritzmg: With `contao.locales` you can only change the list of languages available in various places globally. However with the hook you can change this dynamically, depending on the context. Do we really want to remove this possibility?

ausi:
> we must provide a way to patch the Intl list of countries for specific use cases.

Agree. Both for countries as well as for languages. The following should be possible I think:

1. Add an additional country or locale
2. Remove existing country or locale
3. Replace the label

How about this config:

```yaml
contao:
    intl:
        locales:
            add: [bar_AT, gsw_CH]
            remove: [de_IT]
            replace: […] # Ignore intl and hardcode own list
        countries:
            add: [ZZ]
            remove: [CH]
            replace: […] # Ignore intl and hardcode own list
```

And handling overwriting of the labels with the translator as before with `LNG.xx` and `CNT.xx`?

I would then replace the static System methods with a service that can be used to get the list of locales or countries.

aschempp:
I don't like the config syntax, but otherwise 👍 
Maybe we can find out if other systems have/had the same need and what solution they came up with.

As a side note, there are reasons to allow for runtime configuration of countries as well, e.g. when using multiple domains and only some need a new country. Maybe that means we just don't need a configuration at all and since we provide a hook any extension developer can come up with dynamically modifying the countries in the first place?

fritzmg:
If it's a service we can at least decorate it. But what if multiple extensions want to change either languages or countries, not sure if that works with service decoration?

ausi:
> But what if multiple extensions want to change either languages or countries, not sure if that works with service decoration?

Works and the order is even explicitly controllable by `decoration_priority`: https://symfony.com/doc/current/service_container/service_decoration.html#decoration-priority

fritzmg:
I still feel like a hook or event is more appropriate for this.

m-vo:
The benefit of config is, that it's static at runtime. If we can guarantee that this is the case, we should IMHO prefer it over an event.

fritzmg:
> The benefit of config is, that it's static at runtime. If we can guarantee that this is the case, we should IMHO prefer it over an event.

As already said, it must be dynamic, because you might want to have different countries for instance for different domains in a multidomain setup.

ausi:
Alternative config:

```yaml
contao:
    intl:
        locales: ['+bar_AT', '+gsw_CH', '-de_DE']
        countries: ['+EU', '-DE']
        # OR replace
        countries: ['DE', 'AT', 'CH']
```

aschempp (changes requested):
Just requesting changes here to signal that this is not RTM 🙃 
Needs to
 1. use a service that can be decorated
 2. new config syntax
 3. BC for old hooks

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
```diff
@@ -33,6 +34,7 @@ class MakeServicesPublicPass implements CompilerPassInterface
         'security.logout_url_generator',
         'security.helper',
         'uri_signer',
+        Countries::class,
```
m-vo: Why not just declare the service public in `services.yaml`?

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,164 @@
...
+    public function getCountries(string $displayLocale = null): array
+    {
+        if (null === $displayLocale && ($request = $this->requestStack->getCurrentRequest())) {
+            $displayLocale = $request->getLocale();
+        }
+
+        $countries = SymfonyCountries::getNames($displayLocale);
+        $needsResort = false;
+
+        if (\count($this->countriesList)) {
```
m-vo: Personally, I'd go for either…
```suggestion
        if (\count($this->countriesList) > 0) {
```
…or…
```suggestion
        if (!empty($this->countriesList)) {
```
instead, because it IMHO lowers the cognitive load to understand the code.

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->contaoFramework = $contaoFramework;
+        $this->countriesList = $countriesList;
+    }
+
+    /**
+     * @return array<string,string> Translated country names indexed by their uppercase ISO 3166-1 alpha-2 code
+     */
+    public function getCountries(string $displayLocale = null): array
+    {
+        if (null === $displayLocale && ($request = $this->requestStack->getCurrentRequest())) {
```
m-vo:
```suggestion
        if (null === $displayLocale && null !== ($request = $this->requestStack->getCurrentRequest())) {
```
Maybe?

m-vo (commented):
Nitpicks. Love the tests by the way :sweat_smile:

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
ausi: I felt that there was something wrong when I wrote this... ☺️
Fixed in 8181b663df625e1d214440d8535d1f10331490ba

core-bundle/src/Intl/Countries.php
ausi: I always forget what the coding style is here.

Personally I’d favor `if ($this->countriesList) {`

@leofeyer ?

core-bundle/src/Intl/Countries.php
ausi: I think ` && ($request = ` is the common coding style in the rest of the codebase (as long as there is no difference between null and a falsy value).

m-vo: Hmm, I find quite a few occurrences if I search for `null !== (` in the code base.

Examples for `$request`:

 https://github.com/contao/contao/blob/aaf1a3c4e809ff9bbace655505a326855691ede9/core-bundle/src/EventListener/InsertTags/DateListener.php#L151

https://github.com/contao/contao/blob/aaf1a3c4e809ff9bbace655505a326855691ede9/core-bundle/src/Framework/ScopeAwareTrait.php#L66

m-vo (dismissed):

core-bundle/src/Intl/Countries.php
ausi: Fixed in 608d942ae7fe3a25afc3a8afcf7e3645f9ed25f2

core-bundle/src/Intl/Countries.php
ausi: Changed in 608d942ae7fe3a25afc3a8afcf7e3645f9ed25f2

ausi:
Everything should be done now.

1. Hooks `getCountries` and `getLanguages` are deprecated but still work with the new services
2. Translations via `CNT.xx` and `LNG.xx` still work as before
3. Countries and locales are configurable with the syntax as described above https://github.com/contao/contao/pull/3138#issuecomment-873020365
4. ~~`contao.locales` renamed to `contao.backend_locales`. Not sure if this rename is correct or if we can find a better way? `installed_locales`? `translated_locales`?~~ Reverted in 8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba
5. ~~Country codes in `tl_member` get migrated to uppercase.~~ reverted in d35c30c5bc97f11cbe82e6202f184fd207b683fa (But the hook and the translations still use lowercase)

New Service `Intl\Countries`:

- `->getCountries('de')` returns all countries indexed by the alpha-2 code sorted by label in `'de'`.
- `->getCountryCodes()` returns all alpha-2 codes sorted by the codes.

New Service `Intl\Locales`:

- `->getLocales('de')` returns all locales indexed by locale ID sorted by the names in `'de'`
- `->getLocales('de', true)` adds the native names after the translated ones (as `System::getLanguages()` did)
- `->getLanguages(…)` same as `getLocales` but without regions, meaning instead of `de,de_AT,de_CH,en_US` it only returns `de,en`
- `->getLocaleIds()` returns all locale IDs sorted by ID.
- `->getLanguageLocaleIds()` same as `getLocaleIds` without regions
- `->getDisplayNames(['de', 'en_US'], 'de', true)` returns the names of the passed locales indexed by locale ID sorted by the names in `'de'` with the native labels as suffixes.

m-vo (dismissed):

core-bundle/src/Resources/contao/library/Contao/System.php
ausi: This should be adressed now with the `contao.intl.countries` config.

core-bundle/src/Resources/contao/library/Contao/System.php
ausi: The backend locales did not use this method/hook anymore in many places. With the current state of this pull request, you can configure `contao.backend_locales` and `contao.intl.locales` and additionaly decorate the locales service.

fritzmg:
> 5\. Country codes in `tl_member` get migrated to uppercase. (But the hook and the translations still use lowercase)

Is that necessary? We have at least one project where we have comparisons like `$member->country === 'at'`.

ausi:
> > 5. Country codes in `tl_member` get migrated to uppercase. (But the hook and the translations still use lowercase)
> 
> Is that necessary? We have at least one project where we have comparisons like `$member->country === 'at'`.

We could keep it lowercase for `tl_member`, but I would like to use the uppercase version in the services as this is the default for ISO 3166-1 alpha-2.

fritzmg:
Yes, I would like it to be standardised too. And it's fine for us to adjust that Contao instance once updated to Contao 4.12+. However I just fear it might break Contao installations of other Contao users too.

ausi:
> Yes, I would like it to be standardised too.

So, which one would you prefer? Keeping it lower case in `tl_member` until Contao 5.0 or change it as it is now in this pull request?

fritzmg:
I think we should keep it lowercase for BC and change it in Contao 5.

ausi:
> I think we should keep it lowercase for BC and change it in Contao 5.

Done in d35c30c5bc97f11cbe82e6202f184fd207b683fa.

fritzmg (dismissed):

composer.json
```diff
@@ -139,7 +140,7 @@
         "contao-community-alliance/composer-plugin": "<3.0",
         "contao/core": "*",
         "contao/manager-plugin": "<2.0 || >=3.0",
-        "doctrine/annotations": "<1.7",
+        "doctrine/annotations": "<1.9",
```
leofeyer, line 143: Should we add this conflict in the `core-bundle/composer.json` file as well?

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -78,7 +78,8 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('contao.csrf_token_name', $config['csrf_token_name']);
         $container->setParameter('contao.pretty_error_screens', $config['pretty_error_screens']);
         $container->setParameter('contao.error_level', $config['error_level']);
-        $container->setParameter('contao.locales', $config['locales']);
+        $container->setParameter('contao.locales', $config['locales'] ?: $config['backend_locales']); // Deprecated
```
leofeyer:
```suggestion
        $container->setParameter('contao.locales', $config['locales'] ?: $config['backend_locales']); // Backwards compatibility
```

I am always using the comment "backwards compatibility" for BC stuff, so we can find it more easily once we are removing it.

aschempp:
> We could keep it lowercase for `tl_member`, but I would like to use the uppercase version in the services as this is the default for ISO 3166-1 alpha-2.

I don't think we should ever change this. The field contains a locale ID, which users lowercase letters for the language. The languages in a locale ID are based on ISO-3166, but the upper/lowercase representation should be according to locale ID not the ISO standard.

core-bundle/tests/Intl/CountriesTest.php
```diff
@@ -0,0 +1,232 @@
...
+
+        $countryNames = $this->getCountriesService($translator)->getCountries('de');
+
+        $this->assertSame('Schland', $countryNames['DE']);
+
+        $positionDe = array_search('DE', array_keys($countryNames), true);
+        $positionRu = array_search('RU', array_keys($countryNames), true);
+        $positionTr = array_search('TR', array_keys($countryNames), true);
+
+        // "Schland" should be sorted after Russia and before Turkey
```
leofeyer, line 91: 😂

fritzmg:
> I don't think we should ever change this. The field contains a locale ID, which users lowercase letters for the language. The languages in a locale ID are based on ISO-3166, but the upper/lowercase representation should be according to locale ID not the ISO standard.

`tl_member.country` contains country IDs (`at`, `de` etc.), i.e. these keys: https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/config/countries.php

They are unrelated to locales.

composer.json
ausi: I’m not sure, the prefer-lowest test failed because doctrine was confused by the `@template` annotation. Don’t know if this would have an effect outside the tests too.

ausi: https://github.com/contao/contao/runs/3035275702?check_suite_focus=true#step:7:16

leofeyer: It only works because we are not testing the single bundles with `--prefer-lowest`. Otherwise it would fail there, too.

ausi: The same applied for the existing conflict with <1.7 which was added in 74d867306620d8b070183da1a2a0ea36f86cc998
But I’m happy either way, whatever you prefer ☺️

ausi:
> `contao.locales` renamed to `contao.backend_locales`. Not sure if this rename is correct or if we can find a better way? `installed_locales`? `translated_locales`?

I’m very unsure about this point now, is `contao.backend_locales` right? 
Or should we keep it as it is now?

leofeyer:
What was the motivation for renaming the setting in the first place?

composer.json
leofeyer: Let's just leave it as is.

ausi:
That it is a bit misleading because we now have `contao.locales` and `contao.intl.locales`. 
`contao.locales` defaults to all locales with a translation and is mainly used for the locales that are available for the backend. 
`contao.intl.locales` defaults to all locales known to the `intl` extension and is mainly used for `tl_member.language`.

leofeyer:
In my understanding, `contao.intl.locales` is for pages, because Contao allows us to build pages for any language, and `contao.locales` is for users and members, as it defines in which language they want to see the GUI (and it would not make sense to show them languages without a translation).

However, I agree that this is not as straight forward in the front end as it is in the back end, because the front end language is solely defined by the page and not by the user's preference. In fact, the member language only comes into play if an extension uses it, e.g. for issuing invoices or other documents in German even though the member is browsing the English website. And in such a case, a translation will most likely be required again, therefore I see it as follows:

* `contao.locales` → user/member language
* `contao.intl.locales` → page language

ausi:
Then I think we should keep the name `contao.locales`. I will revert the changes.

ausi:
Done in 8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba

aschempp:
> In my understanding, `contao.intl.locales` is for pages, because Contao allows us to build pages for any language, and `contao.locales` is for users and members, as it defines in which language they want to see the GUI (and it would not make sense to show them languages without a translation).
> 
> However, I agree that this is not as straight forward in the front end as it is in the back end, because the front end language is solely defined by the page and not by the user's preference. In fact, the member language only comes into play if an extension uses it, e.g. for issuing invoices or other documents in German even though the member is browsing the English website. And in such a case, a translation will most likely be required again, therefore I see it as follows:
> 
> * `contao.locales` → user/member language
> * `contao.intl.locales` → page language

The page languages are not related to any bundle configuration. The page configuration is a string, so not related to configuration at all. `contao.locales` is only for back end users. The front end members (`tl_member.language`) is a full list of languages, regardless of what is installed (as far as I know).

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,164 @@
...
+                && \is_string($label)
+                && '' !== $label
+            ) {
+                $countries[$countryCode] = $label;
+                $needsResort = true;
+            }
+        }
+
+        if ($needsResort) {
+            (new \Collator($displayLocale ?? 'en'))->asort($countries);
```
aschempp: Shouldn't this use `$GLOBALS['TL_LANGUAGE']` (or `Request::getLocale()`) instead of `en` as fallback?

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * @return array<string> Uppercase ISO 3166-1 alpha-2 codes
+     */
+    public function getCountryCodes(): array
+    {
+        // If the legacy hook is used, it might add or remove countries
+        if (!empty($GLOBALS['TL_HOOKS']['getCountries'])) {
+            $countryCodes = array_keys($this->getCountries());
+            sort($countryCodes);
```
aschempp, line 99: not sure we should sort here?

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,164 @@
...
+    private function filterCountries(array $countries, string $displayLocale): array
+    {
+        $newList = array_filter(
+            $this->countriesList,
+            static function ($country) {
+                return !\in_array($country[0], ['-', '+'], true);
+            }
+        );
+
+        if ($newList) {
```
aschempp, line 119: should this be `!empty()` or `count($newList) > 0`?

core-bundle/src/Intl/Locales.php
```diff
@@ -0,0 +1,273 @@
...
+     */
+    private $backendLocales;
+
+    public function __construct(TranslatorInterface $translator, RequestStack $requestStack, ContaoFramework $contaoFramework, array $localesList, array $backendLocales)
+    {
+        $this->translator = $translator;
+        $this->requestStack = $requestStack;
+        $this->contaoFramework = $contaoFramework;
+        $this->localesList = $localesList;
+        $this->backendLocales = $backendLocales;
```
aschempp: should this be renamed back as well? Also, the variable is currently unused.

core-bundle/src/Intl/Locales.php
```diff
@@ -0,0 +1,273 @@
...
+
+                if ($nativeLabel !== $label) {
+                    $label .= ' - '.$nativeLabel;
+                }
+            }
+
+            $locales[$localeId] = $label;
+        }
+
+        (new \Collator($displayLocale ?? 'en'))->asort($locales);
```
aschempp: Should we use the request locale as fallback?

core-bundle/src/Intl/Locales.php
```diff
@@ -0,0 +1,273 @@
...
+    private function getDisplayNamesWithoutHook(array $localeIds, string $displayLocale = null, bool $addNativeSuffix = false): array
+    {
+        if (null === $displayLocale && null !== ($request = $this->requestStack->getCurrentRequest())) {
+            $displayLocale = $request->getLocale();
+        }
+
+        $locales = [];
+
+        foreach ($localeIds as $localeId) {
+            $langKey = 'LNG.'.$localeId;
```
aschempp, line 225: is `$localeId` lowercase here? Otherwise the translation would fail

core-bundle/src/Resources/contao/controllers/BackendIndex.php
```diff
@@ -107,7 +108,7 @@ public function run()
 		$objTemplate->messages = Message::generate();
 		$objTemplate->base = Environment::get('base');
 		$objTemplate->language = $GLOBALS['TL_LANGUAGE'];
-		$objTemplate->languages = System::getLanguages(true); // backwards compatibility
+		$objTemplate->languages = System::getContainer()->get(Locales::class)->getDisplayNames(System::getContainer()->getParameter('contao.locales'), null, true); // backwards compatibility
```
aschempp: I think we can safely remove this completely, since the login screen does not have a language selection anymore and the variable has only been here for a few years for BC reasons.

core-bundle/src/Intl/Locales.php
```diff
@@ -0,0 +1,273 @@
...
+        $this->requestStack = $requestStack;
+        $this->contaoFramework = $contaoFramework;
+        $this->localesList = $localesList;
+        $this->backendLocales = $backendLocales;
+    }
+
+    /**
+     * @return array<string,string> Translated locales indexed by their ICU locale IDs
+     */
+    public function getLocales(string $displayLocale = null, bool $addNativeSuffix = false): array
```
aschempp, line 65: I'd suggest to have a separate method that returns locales with native suffix.

core-bundle/src/Resources/contao/dca/tl_user.php
```diff
@@ -180,7 +181,7 @@
 			'eval'                    => array('mandatory'=>true, 'tl_class'=>'w50'),
 			'options_callback' => static function ()
 			{
-				return System::getLanguages(true);
+				return System::getContainer()->get(Locales::class)->getDisplayNames(System::getContainer()->getParameter('contao.locales'), null, true);
```
aschempp: I think we should have a `Locales::getInstalledLocales()`. The backend locales are already passed to the service, but not used atm as far as I could see.

core-bundle/src/Intl/Countries.php
ausi: https://github.com/contao/contao/blob/8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba/core-bundle/src/Intl/Countries.php#L57

core-bundle/src/Intl/Countries.php
ausi: I think the country codes should be sorted the same way independent of an existing hook.

core-bundle/src/Intl/Countries.php
ausi: no, see https://github.com/contao/contao/pull/3138#discussion_r667317566

core-bundle/src/Intl/Locales.php
ausi: > should this be renamed back as well?

No.

> Also, the variable is currently unused.

https://github.com/contao/contao/blob/8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba/core-bundle/src/Intl/Locales.php#L209

core-bundle/src/Intl/Locales.php
ausi: We’d have to duplicate all three methods (`getLocales`, `getLanguages` and `getDisplayNames`) then, I don’t think we should do that.

core-bundle/src/Intl/Locales.php
ausi: https://github.com/contao/contao/blob/8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba/core-bundle/src/Resources/contao/languages/de/languages.xlf#L20

core-bundle/src/Intl/Locales.php
ausi: https://github.com/contao/contao/blob/8796a68c24a69e8f7c2e36c0ca205b684fe6d4ba/core-bundle/src/Intl/Locales.php#L178

core-bundle/src/Resources/contao/controllers/BackendIndex.php
ausi: Unrelated to this PR.

core-bundle/src/Resources/contao/dca/tl_user.php
ausi: We would then have two different sources for the “installed” locales. All services currently use the `contao.locales` parameter so I would keep it that way.

leofeyer:
> The front end members (tl_member.language) is a full list of languages, regardless of what is installed (as far as I know).

No, it is not. The explanation is actually in the text that you have quoted. 😉 

> the member language only comes into play if an extension uses it, e.g. for issuing invoices or other documents in German even though the member is browsing the English website. And in such a case, a translation will most likely be required again

fritzmg:
The available member languages definitely need to be independent from the installed languages. When a member registers for the website, the member chooses which language _they_ speak. It would make no sense to limit that to whatever languages are available as back end translations for Contao.

leofeyer:
Disagree for the reasons already explained above.

If the only installed translations are German and English, the user will need to pick in which language they want to receive their invoice. It does not help if they choose "Spanish" if there is no Spanish translation. 🤷🏻

fritzmg:
But that is still independent from the _installed_ languages. You can create Spanish emails in your app regardless of whether you actually have a Spanish back end translation or a Spanish website root.

fritzmg:
See the Notification Center for example: it uses `System::getLanguages()` in order to retrieve a list of _all_ languages (not just the installed ones) so that you can decide in your application in which language your notification is written. And in your application you might then choose to send the notification in the member's language - or in the language of the current page as a fallback.

ausi:
> The available member languages definitely need to be independent from the installed languages.

They are, same as before.

> If the only installed translations are German and English, the user will need to pick in which language they want to receive their invoice.

This is not related to this PR IMO. It was always possible to change the options callback of `tl_member.language` as needed for your project, and the default was to show “all” locales. This hasn’t changed with this pull request.

fritzmg:
> * `contao.locales` → user/member language

That would be an unnecessary behaviour change and would make no sense imho (as explained above). Currently `tl_member.language` is _all_ languages: 

https://github.com/contao/contao/blob/a0523bc35d22fafb0222a69e0b2380f320a70654/core-bundle/src/Resources/contao/dca/tl_member.php#L277

core-bundle/src/Resources/contao/dca/tl_user.php
aschempp: but they _should_ use the service and not a parameter? Because the service would allow us to change the parameter etc. without affecting the API?

ausi: We could add such a method, but is it really worth it?
Also I’m not sure about naming them “installed”, maybe `getTranslatedLocales()` or `getEnabledLocales()` or `getActiveLocales()`?

aschempp:
In conclusion, we agree that https://github.com/contao/contao/pull/3138#issuecomment-878313700 was correct?

`contao.locales` = installed language packs & `tl_user.language`
`contao.intl.locales` = languages/locales known to the system & `tl_member.language`

none of that is related to `tl_page.language`

fritzmg:
> `contao.locales` = installed language packs & `tl_user.language`

Technically `contao.locales` can be arbitrary, since it's a bundle configuration. It just defaults to the installed languages. Currently there is no way of only retrieving the actually installed languages (not sure how relevant that is for the core or any extension developers though) - other than implementing `Configuration::getLocales` yourself of course.

core-bundle/src/Resources/contao/dca/tl_user.php
leofeyer: As briefly discussed in Slack, we want to rename the parameter to `contao.intl.enabled_locales` and add a method `getEnabledLocales()`.

ausi:
Updated the PR. As discussed in https://github.com/contao/contao/pull/3138#discussion_r667985380 `contao.locales` got now moved to `contao.intl.enabled_locales`.

The parameter `contao.locales` is deprecated and all services now use the `getEnabledLocaleIds()` and `getEnabledLocales()` methods of the locales service.

core-bundle/src/Intl/Countries.php
```diff
@@ -0,0 +1,155 @@
...
+     * @var string
+     */
+    private $defaultLocale;
+
+    /**
+     * @var array<string>
+     */
+    private $countries;
+
+    public function __construct(TranslatorInterface $translator, RequestStack $requestStack, ContaoFramework $contaoFramework, array $availableCountries, string $defaultLocale, array $countriesList)
```
leofeyer: Just a minor nitpick: `array $availableCountries, array $countriesList, string $defaultLocale` would be more logical. 🙈

core-bundle/src/Intl/Locales.php
```diff
@@ -0,0 +1,312 @@
...
+     * @var array<string>
+     */
+    private $locales;
+
+    /**
+     * @var array<string>
+     */
+    private $enabledLocales;
+
+    public function __construct(TranslatorInterface $translator, RequestStack $requestStack, ContaoFramework $contaoFramework, array $availableLocales, array $translatedLocales, string $defaultLocale, array $locales, array $enabledLocales)
```
leofeyer: `array $translatedLocales, string $defaultLocale, array $locales, array $enabledLocales`

Really? What is the difference between all these arrays? (`$defaultLocale` could be the last argument here, too.)

leofeyer (dismissed):

core-bundle/src/Intl/Countries.php
ausi: Fixed in 6b26513356bbe038839398a9823a7259669695e1

core-bundle/src/Intl/Locales.php
ausi: Renamed in 3a8c19096ff111e95298cdba81070a9dfb4c51da to:  
```
array $defaultLocales, 
array $defaultEnabledLocales, 
array $configLocales, 
array $configEnabledLocales,
```  
better?

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3139, opened 2021-07-01
Added template flag feature

aschempp:
This is an alternative approach to https://github.com/contao/contao/pull/3116

The idea is that we allow flags similar to Twig or insert tags in Contao templates.

**Before:**
```php
<?= Contao\StringUtil::specialchars($this->foobar) `>
```

**After:**
```php
<?= $this->{'foobar|specialchars'} ?>
```

I see two advantages
 1. the syntax is close to Twig and the existing insert tag flags
 2. we can provide an arbitrary number of flags, and developers can extend them

Unfortunately, PHP does not support `<?= $this->foobar|specialchar ?>` without escaping the string. Also, we sure can discuss about the (duplicate) implementation and whether the hook makes sense, should be an event etc.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -150,6 +150,8 @@ abstract class Template extends Controller
 	 */
 	public function __get($strKey)
 	{
+		list($strKey, ...$flags) = explode('|', $strKey);
+
 		if (isset($this->arrData[$strKey]))
 		{
 			if (\is_object($this->arrData[$strKey]) && \is_callable($this->arrData[$strKey]))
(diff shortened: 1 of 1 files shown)


m-vo:
I don't understand the need to add new features to the `Template` class? Maybe it's my view because I'm trying to build a Twig integration that allows a smooth transition from one engine to the other. Any addition makes it harder. Idk.

Many of these flags do not exist in Twig (or are named + accessed differently) and there is a fundamental difference how escaping works, which usually uses the same `|` syntax. Also, in Twig these things are no flags, but filters, so the order of operations matters - but that's only naming I guess. So what's the tradeoff then? You're loosing autocompletion for saving 10 chars? :upside_down_face:

ausi:
> the syntax is close to Twig and the existing insert tag flags

I think this is actually a disadvantage. If users want to have a “nice” template language, they should use Twig instead and we should endorse that switch IMO. So especially for features that Twig has but our PHP templates do not we should encourage the switch to Twig and not add that feature to our template engine.

Also, Twig has syntax highlighting and auto completion as does regular PHP, but `<?= $this->{'foobar|specialchars'} ?>` does not.

Toflar:
I too think it's time to ring in the time of Twig in Contao and slowly start giving up our PHP templates.

aschempp:
Fine with me, but then we shouldn't merge https://github.com/contao/contao/pull/3116 and remove the helper methods again as well 🙃

PR #3141, opened 2021-07-01
Deprecate using tabindex values greater than 0

### LINKED ISSUE #2949: Remove the tabindex field from the form generator
_Contao 4.9.13_

For the checkbox menu and the radio button menu, please add the options to define a `tabindex`, as it is already possible with other form fields.

**Checkbox menu**

image

**Radio button menu**

image

**Working example for `tabindex`**

image

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2949
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
@@ -237,7 +237,7 @@
         <source>A form field can be focused by pressing the [ALT] or [CTRL] key and the access key simultaneously.</source>
       </trans-unit>
       <trans-unit id="tl_form_field.tabindex.0">
-        <source>Tabindex</source>
+        <source>Tab index</source>
       </trans-unit>
       <trans-unit id="tl_form_field.tabindex.1">
         <source>The position of the form field in the tabbing order.</source>
--- a/core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
@@ -237,7 +237,7 @@
         <source>A form field can be focused by pressing the [ALT] or [CTRL] key and the access key simultaneously.</source>
       </trans-unit>
       <trans-unit id="tl_form_field.tabindex.0">
-        <source>Tab index</source>
+        <source>Tabindex</source>
       </trans-unit>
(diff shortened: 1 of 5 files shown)


aschempp (changes requested):
I don't think we should change the label. That means every translator has to translate that deprecated field yet again just for it to be removed soon …

fritzmg (dismissed):

core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
```diff
@@ -237,7 +237,7 @@
         <source>A form field can be focused by pressing the [ALT] or [CTRL] key and the access key simultaneously.</source>
       </trans-unit>
       <trans-unit id="tl_form_field.tabindex.0">
-        <source>Tab index</source>
+        <source>Tabindex</source>
       </trans-unit>
       <trans-unit id="tl_form_field.tabindex.1">
         <source>The position of the form field in the tabbing order.</source>
```
ausi: Should we add a note that this field is deprecated here too?

```suggestion
        <source>The position of the form field in the tabbing order. (Deprecated)</source>
```

ausi (dismissed):

core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
leofeyer: > That means every translator has to translate that deprecated field yet again just for it to be removed soon …

This would apply here as well then, wouldn't it?

bytehead:
>  just for it to be removed soon …

Define soon, there's ~~4.9 LTS and~~ 4.13 LTS with this kind of change. Not so soon for me (+5 more years at least) 🙃 

Edit: just saw it is tagged to 4.12.

leofeyer:
It will be in Contao 4.13 as well, because we will not remove the field before Contao 5. Therefore "soon" means 4.5 years actually. 😄

core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
ausi: Yes. As I’m not involved in the translation process, I cannot evaluate cost vs. benefit here. I’m happy either way.

aschempp (approved):

PR #3142, opened 2021-07-01
Added |blank flag to open links in new window

aschempp:
This allows to generate a link that opens in a new window

`{{link::123|blank}}` => `<a href="alias-123.html" target="_blank" …>`

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -454,7 +454,7 @@ class InsertTags extends Controller
 						}
 
 						$strName = $objNextPage->title;
-						$strTarget = $objNextPage->target ? ' target="_blank" rel="noreferrer noopener"' : '';
+						$strTarget = $objNextPage->target || \in_array('blank', $flags, true) ? ' target="_blank" rel="noreferrer noopener"' : '';
 						$strClass = $objNextPage->cssClass ? sprintf(' class="%s"', $objNextPage->cssClass) : '';
 						$strTitle = $objNextPage->pageTitle ?: $objNextPage->title;
 					}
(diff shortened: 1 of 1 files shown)


ausi:
Why using an insert tag flag for that? This seems to be against the insert tags concept IMO. `{{link::123::blank}}` would fit better to how the other insert tags work.

And is this not already possible by using `<a href="{{link_url::123}}" target="_blank">` ?

leofeyer:
I'm not sure if that is a use-case that we have to support, but @fritzmg sure has a point here.

aschempp:
My use case is using an insert tag where HTML is not supported. My most used case is a legal checkbox in a form, where I want to link to the legal page in a checkbox label.

fritzmg:
To clarify: an _additional_ flag or attribute for the `{{link::*}}` insert tag is fine - I just meant it cannot be the only way. And I agree with this use case. This also works with `{{link_open::*}}`, correct? (that's were we would need it more often).

ausi:
> `{{link::123::blank}}` would fit better to how the other insert tags work.

Suggested alternative: https://github.com/aschempp/contao/pull/7

aschempp:
I disagree. In your concept, how would I create an absolute URL that opens in a new window? `{{link::42|absolute|blank}}` is exactly right and the order of flags does not matter.

ausi:
> how would I create an absolute URL that opens in a new window?

`{{link::42::absolute::blank}}` or `{{link::42::blank::absolute}}`

leofeyer:
I think we have decided against the `{{tag::value::value::value}}` syntax in favor of `{{tag::value?more=values}}` syntax, so this should either be `{{link::42|absolute|blank}}` or `{{link::42?target=blank|absolute}}`, shouldn't it?

fritzmg:
> I think we have decided against the `{{tag::value::value::value}}` syntax in favor of `{{tag::value?more=values}}` syntax, so this should either be `{{link::42|absolute|blank}}` or `{{link::42?target=blank|absolute}}`, shouldn't it?

We already have `{{trans::*::*::*}}`, `{{asset::*::*}}`, `{{format_date::*::*}}` and `{{convert_date::*::*::*}}`.

ausi:
> in favor of `{{tag::value?more=values}}` syntax,

This is not possible in this case I think because the url itself could already use `?` like `{{link::https://example.com/?query=foo::blank}}`

aschempp:
> > I think we have decided against the `{{tag::value::value::value}}` syntax in favor of `{{tag::value?more=values}}` syntax, so this should either be `{{link::42|absolute|blank}}` or `{{link::42?target=blank|absolute}}`, shouldn't it?
> 
> We already have `{{trans::*::*::*}}`, `{{asset::*::*}}`, `{{format_date::*::*}}` and `{{convert_date::*::*::*}}`.

Yes, but these are value arguments to the insert tag. Flags are boolean, these values are not.

ausi:
> Flags are boolean, these values are not.

What kind of boolen is `|nl2br` or `|strtolower`? Nearly all of them are functions (exception the cache related ones) and are executed in the specified order. They could even take arguments and it would fit the concept imo like `specialchars(double_encode)`, although no one does that currenty AFAIK.

aschempp:
I'm sorry I did mean _flags can be boolean_ 😅 

> They could even take arguments and it would fit the concept imo like specialchars(double_encode), although no one does that currenty AFAIK.

I thought about adding that multiple times, but that's not supported atm.

ausi:
> I did mean _flags can be boolean_ 😅

Not according to our own documentation: _“With flags, you can process insert tags further.”_

leofeyer:
As discussed in Mumble on July 29th, we want to implement `{{link::123::blank}}`.

ausi:
Closed in favor of #3230

PR #3144, opened 2021-07-01
Clarify structure of table and hint to doc

ACCakut:


core-bundle/src/Resources/contao/languages/en/tl_form.xlf
```diff
@@ -108,7 +108,7 @@
         <source>Target table</source>
       </trans-unit>
       <trans-unit id="tl_form.targetTable.1">
-        <source>The target table must contain a column for every form field.</source>
+        <source>The target table must contain a column of the same name for every form field. More in the docs.</source>
```
leofeyer, line 111: I do not like the "More in the docs" part. Can we at least provide a link?

PR #3145, opened 2021-07-01
Clarify structure of table and hint to doc (German)

ACCakut:


leofeyer:
The German translation is managed on transifex.com and cannot be changed in a PR. Please change the english source file instead.

PR #3146, opened 2021-07-01
Deprecate the insert tag flag "|absolute"

ausi:
While discussing https://github.com/contao/contao/pull/3142#issuecomment-872155666 I noticed, that we have an insert tag flag named `absolute`. This is the only flag that is used as an argument AFAIK. All other flags are related to caching or postprocessing the output.

As I understand the insert tag concept this is against the design and IMO very confusing. 
Therefore we should deprecate it IMO.

Is this a feature and still allowed for 4.12 ? Or should I rebase it on 4.9 ?


aschempp:
I don't think this should be deprecated and #3142 is very valid. Yes, insert tag flags transform the output. To the user, it does not matter that it is used differently for different flags (applying a method vs. adjusting the tag itself). Using `{{link::42|absolute}}` adds the base URL, `{{link::42|strip_tags}}` will strip HTML.

On a side note, I have already implemented the `|blank` flag on my client system by simply parsing the HTML and adding the target attribute. Same would be applicable to `|absolute`.

m-vo:
But isn't it a weird design that you have two independent controls (tag / flags) that only work together in very specific combinations? 

Why differentiate between `::` and `|` then in the first place?

(We also have urlencoded strings with query params to denote optional arguments in arbitrary order. See the picture or figure insert tags for instance. Should `enableLightbox` become a flag then as well? And why would you differentiate between boolean options and custom arguments?)

ausi:
The documentation of the insert tags itself says _“With flags, you can process insert tags further. The value can be passed to a PHP function, for example. Any number of flags can be combined with each other”_

The `absolute` flag is the only one behaving different here, as it only works with some insert tags, but not with others like:

- `{{file::1fd067ba-abdd-11ea-b6e3-1d8e749cae97|absolute}}`
- `{{env::request|absolute}}`
- `{{asset::jquery.js::contao-components/jquery|absolute}}`

Also, the order matters for all other flags, but not for `|absolute`.

fritzmg:
The `|flatten` insert tag flag is also a special one, not sure where you would ever use that though.

ausi:
> not sure where you would ever use that though.

See https://github.com/contao/core/pull/8021

aschempp:
Same for `|refresh` and `|uncached`. They add behaviour instead of changing the string. That's a supported use-case for flags imho.

> See the picture or figure insert tags for instance. Should enableLightbox become a flag then as well? 

@m-vo Can you give an example?

ausi:
> > See the picture or figure insert tags for instance. Should enableLightbox become a flag then as well?
> 
> @m-vo Can you give an example?

`{{figure::image.png?size=123&enableLightbox=1}}`
vs.
`{{figure::image.png?size=123|enableLightbox}}`

> Same for `|refresh` and `|uncached`. They add behaviour instead of changing the string.

But they can be used with every insert tag. Like all other flags can be used with every insert tag.

aschempp:
Maybe I fail to see the point of changing a perfectly working system that's widely used and perfectly works 🤷 I don't find it confusing at all.

ausi:
> Maybe I fail to see the point of changing a perfectly working system that's widely used and perfectly works 🤷

I fully agree, the system worked perfectly, was widely used and until Contao 4.5 it was also consistent. I want that consistency back and want to comply with the Contao documentation.

aschempp:
>  want to comply with the Contao documentation

maybe the documentation is just wrong and needs to be updated?

leofeyer:
As discussed in Mumble on July 29th, we want to merge this PR and deprecate the `|absolute` flag.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -444,13 +444,13 @@ protected function doReplace($strBuffer, $blnCache)
 
 								if ($objNext instanceof PageModel)
 								{
-									$strUrl = \in_array('absolute', $flags, true) ? $objNext->getAbsoluteUrl() : $objNext->getFrontendUrl();
+									$strUrl = \in_array('absolute', \array_slice($elements, 2), true) || \in_array('absolute', $flags, true) ? $objNext->getAbsoluteUrl() : $objNext->getFrontendUrl();
```
leofeyer, line 447: Do we really need the `array_slice($elements, 2)` everywhere? We know that `$elements[0]` is the name of the insert tag and `$elements[1]` is a numeric ID, so there should not be any false positives if we just used `in_array('absolute', $elements)`, should there?

ausi: IMO it’s forwards compatibility with #3230

leofeyer: But isn't the `array_slice()` redundant in #3230 as well? IMHO `$elements[0]` and `$elements[1]` will never be `absolute` or `blank`, so we do not have to use `array_slice()` at all.

ausi: > IMHO `$elements[0]` and `$elements[1]` will never be `absolute` or `blank`,

They could if you have a page with the alias `absolute` or `blank`

leofeyer (approved):

leofeyer:
Thank you @ausi.

Issue #3143, opened 2021-07-01
Fix the "guests only" migration


leofeyer:
In Contao <4.12, if an element had both `protected=1` and `guests=1`, the "guests only" setting was ignored. We must therefore only add the `-1` group in the migration, if the target element is not protected.

leofeyer:
We also have to fix the following, because otherwise an exception will be thrown:

https://github.com/contao/contao/blob/43188094917d7e29242c7bf49418331a2f74f851/core-bundle/src/Resources/contao/controllers/FrontendIndex.php#L264-L286

leofeyer:
And we have to keep checking the `guests` field in the `Controller::isVisibleElement()` method:

```php
if (TL_MODE == 'FE')
{
	$user = null;

	if (System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
	{
		$user = FrontendUser::getInstance();
	}

	if ($objElement->protected)
	{
		$groups = StringUtil::deserialize($objElement->groups, true);
		$blnReturn = (!$user && \in_array(-1, $groups)) || ($user && $user->isMemberOf($groups));
	}
	elseif ($objElement->guests && $user)
	{
		trigger_deprecation('…');
		$blnReturn = false; // backwards compatibility
	}
}
```

leofeyer:
Also this:

https://github.com/contao/contao/blob/43188094917d7e29242c7bf49418331a2f74f851/core-bundle/src/Resources/contao/modules/ModuleSearch.php#L173-L187

We should probably review every occurrence of `->isMemberOf()`.

leofeyer:
As discussed in Mumble on July 1st, we want to keep the `guests` checkbox with the previous logic (only hide pages in the navigation menu but completely hide elements etc.)  and also keep the `-1` group to be forward compatible. Then in Contao 5, we will remove the `guests` checkbox.

leofeyer:
All done. 🎉

PR #3148, opened 2021-07-02
Filter "guests only" pages to restore backwards compatibility

leofeyer:
Another follow-up PR to #3143.

Since we are no longer using the `find*WithoutGuests()` methods, we have to filter the "guests only" pages in the menu modules. This also allows us to trigger a deprecation warning, which is desirable.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -280,6 +280,7 @@ abstract class Module extends Frontend
 
 		$items = array();
 		$security = System::getContainer()->get('security.helper');
+		$user = $security->getUser();
 
 		$objTemplate = new FrontendTemplate($this->navigationTpl ?: 'nav_default');
 		$objTemplate->pid = $pid;
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -280,7 +280,7 @@ abstract class Module extends Frontend
 
 		$items = array();
 		$security = System::getContainer()->get('security.helper');
-		$user = $security->getUser();
+		$isMember = $security->isGranted('ROLE_MEMBER');
 
 		$objTemplate = new FrontendTemplate($this->navigationTpl ?: 'nav_default');
 		$objTemplate->pid = $pid;
@@ -315,7 +315,7 @@ abstract class Module extends Frontend
 			}
 
 			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
-			if ($objSubpage->guests && !$objSubpage->protected && $user)
+			if ($objSubpage->guests && !$objSubpage->protected && $isMember)
 			{
 				trigger_deprecation('contao/core-bundle', '4.12', 'Using the "show to guests only" feature has been deprecated an will no longer work in Contao 5.0. Use the "protect page" function instead.');
 				continue;
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -318,7 +318,7 @@ abstract class Module extends Frontend
 			}
 
 			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
-			if ($objSubpage->guests && !$objSubpage->protected && !$user)
+			if ($objSubpage->guests && !$objSubpage->protected && $user)
 			{
 				trigger_deprecation('contao/core-bundle', '4.12', 'Using the "show to guests only" feature has been deprecated an will no longer work in Contao 5.0. Use the "protect page" function instead.');
 				continue;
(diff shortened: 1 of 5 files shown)


core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -317,6 +317,13 @@ protected function renderNavigation($pid, $level=1, $host=null, $language=null)
 				trigger_deprecation('contao/core-bundle', '4.12', 'Using a tabindex value greater than 0 has been deprecated and will no longer work in Contao 5.0.');
 			}
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objSubpage->guests && !$objSubpage->protected && !$user)
```
ausi:
```suggestion
			if ($objSubpage->guests && !$objSubpage->protected && $user)
```

core-bundle/src/Resources/contao/modules/ModuleBooknav.php
```diff
@@ -205,6 +205,13 @@ protected function getBookPages($intParentId, $groups, $time)
 		{
 			$objPage->loadDetails();
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objPage->guests && !$objPage->protected && !$user)
```
ausi:
```suggestion
			if ($objPage->guests && !$objPage->protected && $user)
```

core-bundle/src/Resources/contao/modules/ModuleCustomnav.php
```diff
@@ -95,6 +95,13 @@ protected function compile()
 		{
 			$objModel->loadDetails();
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objModel->guests && !$objModel->protected && !$user)
```
ausi:
```suggestion
			if ($objModel->guests && !$objModel->protected && $user)
```

core-bundle/src/Resources/contao/modules/ModuleQuicklink.php
```diff
@@ -94,6 +94,13 @@ protected function compile()
 		{
 			$objSubpage->loadDetails();
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objSubpage->guests && !$objSubpage->protected && !$user)
```
ausi:
```suggestion
			if ($objSubpage->guests && !$objSubpage->protected && $user)
```

core-bundle/src/Resources/contao/modules/ModuleQuicknav.php
```diff
@@ -132,6 +132,13 @@ protected function getQuicknavPages($pid, $level=1, $host=null)
 				$objSubpage->domain = $host;
 			}
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objSubpage->guests && !$objSubpage->protected && !$user)
```
ausi:
```suggestion
			if ($objSubpage->guests && !$objSubpage->protected && $user)
```

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -313,6 +313,13 @@ protected function renderNavigation($pid, $level=1, $host=null, $language=null)
 				trigger_deprecation('contao/core-bundle', '4.12', 'Using a tabindex value greater than 0 has been deprecated and will no longer work in Contao 5.0.');
 			}
 
+			// Hide the page if it is not protected and only visible to guests (backwards compatibility)
+			if ($objSubpage->guests && !$objSubpage->protected && $user)
```
aschempp: there's no `$user`?

aschempp (changes requested):

core-bundle/src/Resources/contao/modules/Module.php
ausi: Line 281 and 285?

aschempp: not sure what you're looking at, but this has been removed by my frontend group voter and probably a rebase

Bildschirmfoto 2021-07-05 um 08 32 32

ausi: You are right, I looked at the file before the rebase/merge https://github.com/contao/contao/blob/79a8dcff1c6805b15ff3820cf4f6c6c854a83ddd/core-bundle/src/Resources/contao/modules/Module.php#L281-L285

So this needs to be fixed indeed.

leofeyer: Fixed in d3d00c9e709d7de8084d293f16f2ca25f247e1fd. Is it correct this way or should I have checked `ROLE_MEMBER` instead?

aschempp: never check the user, always ask a voter (= check `ROLE_MEMBER`)

aschempp (changes requested):
never check the user, always ask a voter (= check `ROLE_MEMBER`)

leofeyer:
Fixed in d38daaa5098054bdd8a5953db45ccab35a1db2d0.

ausi (approved):

PR #3149, opened 2021-07-02
Update the CI tools

leofeyer:
Version 4.4 of `contao/test-case` finally adds the `request_stack` to the container by default. We can therefore drop a couple of lines in our unit tests. Thanks to @m-vo.


m-vo (approved):
:+1: 

Extra benefit: This also makes some tests succeed when executed individually where we had forgotten to add the service in the past. :slightly_smiling_face:

PR #3150, opened 2021-07-02
Always use the security voter to check member groups

aschempp:
Since we have security voters to check for member group access, we must make sure it is always used and deprecated the old methods.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/FrontendUser.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendUser.php
@@ -251,11 +251,14 @@ class FrontendUser extends User
 	}
 
 	/**
-	 * @param array|int|string $ids
-	 * @return bool
+	 * Return true if the user is member of a particular group
 	 *
-	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
-	 *             Use Symfony security instead.
+	 * @param mixed $ids A single group ID or an array of group IDs
+	 *
+	 * @return boolean True if the user is a member of the group
+	 *
+	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0;
+	 *             use Symfony security instead
 	 */
 	public function isMemberOf($ids)
 	{
--- a/core-bundle/src/Resources/contao/classes/FrontendUser.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendUser.php
@@ -10,6 +10,8 @@
 
 namespace Contao;
 
+use Contao\CoreBundle\Security\ContaoCorePermissions;
+
 /**
  * Provide methods to manage front end users.
  *
@@ -246,6 +248,27 @@ class FrontendUser extends User
 				$this->strLoginPage = $objGroup->jumpTo;
 			}
 		}
+	}
+
+	/**
+	 * @param array|int|string $ids
+	 * @return bool
+	 *
+  	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
+	 *             Use Symfony security instead.
+	 */
+	public function isMemberOf($ids)
+	{
+		$security = System::getContainer()->get('security.helper');
+
+		if ($security->getUser() === $this)
+		{
+			trigger_deprecation('contao/core-bundle', '4.12', 'Using "Contao\FrontendUser::isMemberOf()" has been deprecated and will no longer work in Contao 5.0. Use Symfony security instead.');
+
+			return $security->isGranted(ContaoCorePermissions::MEMBER_IN_GROUPS, $ids);
+		}
+
+		return parent::isMemberOf($ids);
 	}
 
 	/**
--- a/core-bundle/src/Resources/contao/classes/FrontendUser.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendUser.php
@@ -254,7 +254,7 @@ class FrontendUser extends User
 	 * @param array|int|string $ids
 	 * @return bool
 	 *
-  	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
+	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
 	 *             Use Symfony security instead.
 	 */
 	public function isMemberOf($ids)
--- a/core-bundle/src/Security/Voter/MemberGroupVoter.php
+++ b/core-bundle/src/Security/Voter/MemberGroupVoter.php
@@ -22,17 +22,7 @@ class MemberGroupVoter extends Voter
 {
     protected function supports($attribute, $subject): bool
     {
-        if (ContaoCorePermissions::MEMBER_IN_GROUPS !== $attribute) {
-            return false;
-        }
-
-        if (!\is_array($subject)) {
-            return (string) (int) $subject === (string) $subject;
-        }
-
-        $filtered = array_filter($subject, static function ($val) { return (string) (int) $val === (string) $val; });
-
-        return \count($subject) === \count($filtered);
+        return ContaoCorePermissions::MEMBER_IN_GROUPS === $attribute;
     }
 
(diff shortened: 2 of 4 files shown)


core-bundle/src/Resources/contao/classes/FrontendUser.php
```diff
@@ -248,6 +250,27 @@ protected function setUserFromDb()
...
 	}
 
+	/**
+	 * @param array|int|string $ids
+	 * @return bool
+	 *
+  	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
+	 *             Use Symfony security instead.
+	 */
+	public function isMemberOf($ids)
```
leofeyer, line 263: Since you are no longer using the `isMemberOf()` method in the `MemberGroupVoter`, I guess we do not have to add this method at all, do we?

aschempp: huh? That's for anyone doing the same permissions checks we previously did.

aschempp:
Funny code style error …??

core-bundle/src/Resources/contao/classes/FrontendUser.php
```diff
@@ -248,6 +250,27 @@ protected function setUserFromDb()
 		}
 	}
 
+	/**
+	 * @param array|int|string $ids
+	 * @return bool
+	 *
+  	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
```
ausi:
```suggestion
	 * @deprecated Deprecated since Contao 4.12, to be removed in Contao 5.0.
```

ausi (commented):
`Backend::findSearchablePages()` still uses the method and would cause a deprecation warning.

https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Resources/contao/classes/Backend.php#L690

aschempp:
> `Backend::findSearchablePages()` still uses the method and would cause a deprecation warning.
> 
> https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Resources/contao/classes/Backend.php#L690

no it would not. `Backend::findSearchablePages` runs in the back end firewall (I would assume), therefore 1) we can't use the voter and 2) it would not throw a deprecation because the current security user does not equal the user instance.

core-bundle/src/Resources/contao/classes/FrontendUser.php
leofeyer: Why not add a `USER_IN_GROUPS` voter and deprecate the method on the `User` object?

ausi:
>  `Backend::findSearchablePages` runs in the back end firewall (I would assume)

I would assume the opposite:
https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Controller/SitemapController.php#L26
https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Controller/SitemapController.php#L59

core-bundle/src/Security/Voter/MemberGroupVoter.php
```diff
@@ -31,13 +32,26 @@ protected function supports($attribute, $subject): bool
 
     protected function voteOnAttribute($attribute, $subject, TokenInterface $token): bool
     {
+        // Filter non-numeric values
+        $subject = array_filter(array_map('intval', (array) $subject));
```
leofeyer:
```suggestion
        $subject = array_map('intval', array_filter($subject, static function ($val) { return (string)(int) $val === (string) $val; }));
```

Because https://3v4l.org/VupnO

leofeyer (changes requested):

core-bundle/src/Security/Voter/MemberGroupVoter.php
leofeyer: More readable solution:

```php
if (!\is_array($subject)) {
    $subject = [$subject];
}

$subject = array_values(array_filter($subject, static function ($val) { return is_numeric($val); }));

// …

if (!$user instanceof FrontendUser) {
    return \in_array(-1, $subject, false);
}
```

fritzmg:
Historically `Backend::findSearchablePages` was indeed used in the back end context as XML files were previously generated in the back end as well as the cronjob (which is or was also regarded as a back end context).

fritzmg:
I do think though that our `Voters` should always check the instance of the user. Even [Symfony's own example](https://symfony.com/doc/current/security/voters.html#creating-the-custom-voter) checks the user instance (though, granted, it only does so to check whether the user is actually logged in). So the **Member**GroupVoter (`contao_member.groups`) should only vote on `FrontendUser` instances and return false otherwise. Otherwise if a voter happens to get queried with a `BackendUser` instance it might return true - even though the voter is not applicable in the first place.

core-bundle/src/Security/Voter/MemberGroupVoter.php
```diff
@@ -25,19 +26,37 @@ protected function supports($attribute, $subject): bool
...
 
-        return is_numeric($subject)
-            || (\is_array($subject) && \count($subject) === \count(array_filter($subject, 'is_numeric')));
+        if (!\is_array($subject)) {
+            return (string) (int) $subject === (string) $subject;
+        }
+
+        $filtered = array_filter($subject, static function ($val) { return (string) (int) $val === (string) $val; });
+
+        return \count($subject) === \count($filtered);
```
leofeyer: I think we should remove all of this logic, because if `ContaoCorePermissions::MEMBER_IN_GROUPS === $attribute` is true, the voter supports voting on the subject, no matter what the subject is. We will filter non-numeric values later in the `voteOnAttribute()` method.

leofeyer (changes requested):

aschempp:
> So the MemberGroupVoter (contao_member.groups) should only vote on FrontendUser instances and return false otherwise. Otherwise if a voter happens to get queried with a BackendUser instance it might return true - even though the voter is not applicable in the first place.

I think it already does? If you are a backend user (backend firewall context), then we assume you are a guest to the front end (therefore we allow access if group `-1` is in the list).

core-bundle/src/Security/Voter/MemberGroupVoter.php
aschempp: fine with me 😊

fritzmg:
> I think it already does? If you are a backend user (backend firewall context), then we assume you are a guest to the front end (therefore we allow access if group `-1` is in the list).

True, I overlooked that we then only check for `-1`.

leofeyer (approved):

leofeyer:
> `Backend::findSearchablePages()` still uses the method and would cause a deprecation warning.
> 
> https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Resources/contao/classes/Backend.php#L690

That's true. Maybe we should not override the method in the `FrontendUser` class and only use the `User::isMemberOf()` method?

aschempp:
> Maybe we should not override the method in the FrontendUser class and only use the User::isMemberOf() method?

Then there wouldn't be a deprecation warning to use voters? I think the correct solution would be to move `Backend::getSearchablePages` to somewhere that does not "belong" to the backend context. Maybe it should be implemented in the XML controller and the Backend method should be deprecated?

I would however not consider this to be part of the PR. The current implementation works, it just triggers a deprecation warning that should be there, so its a bug we can fix in a separate - well thought-through - PR

leofeyer:
I am not so sure about the "belonging to the back end". We are retrieving the front end user here

https://github.com/contao/contao/blob/d32232fe75993125a9159136d953922cda301436/core-bundle/src/Resources/contao/classes/Backend.php#L678-L681

so why do we not just ask the security helper?

```php
$indexProtected = $security->isGranted(ContaoCorePermissions::MEMBER_IN_GROUPS, $objPage->groups);
```

PR #3151, opened 2021-07-04
Enforce the MySQL strict mode

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | followup to #3131 for 4.x
| Docs PR or issue | todo

This will prevent the installer from running when MySQL strict mode is disabled.

### DIFF
--- a/installation-bundle/src/InstallTool.php
+++ b/installation-bundle/src/InstallTool.php
@@ -242,43 +242,39 @@ class InstallTool
             $row = $this->connection->fetchAssociative("SHOW VARIABLES LIKE 'innodb_large_prefix'");
 
             // The variable no longer exists as of MySQL 8 and MariaDB 10.3
-            if (false === $row || '' === $row['Value']) {
-                return false;
-            }
+            if (false !== $row && '' !== $row['Value']) {
+                // As there is no reliable way to get the vendor (see #84), we are
+                // guessing based on the version number. The check will not be run
+                // as of MySQL 8 and MariaDB 10.3, so this should be safe.
+                $vok = version_compare($version, '10', '>=') ? '10.2.2' : '5.7.7';
 
-            // As there is no reliable way to get the vendor (see #84), we are
-            // guessing based on the version number. The check will not be run
-            // as of MySQL 8 and MariaDB 10.3, so this should be safe.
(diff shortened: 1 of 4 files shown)


m-vo:
This probably  needs to be rebased + adjusted once #3131 is merged upstream.

ausi (dismissed):

aschempp:
I wonder if 
a) people can actually change this on their regular hosting?
b) if we can, shouldn't we just add the option ourselves (like we do with other options)?

also, I remember we had plenty of issues with strict mode before, or what that _another strict mode_ 😅

m-vo:
> also, I remember we had plenty of issues with strict mode before, or what that another strict mode 😅

That likely was `ONLY_FULL_GROUP_BY`?

fritzmg:
Maybe Contao should instead ensure that no data is saved that might get truncated by the database? Users will for sure have no control over the database server configuration.

m-vo:
>I wonder if
a) people can actually change this on their regular hosting?

>  Users will for sure have no control over the database server configuration.

The fix the installer suggests is using `SESSION` config. IMO that should work everywhere. But we should test it for sure.

ausi:
> also, I remember we had plenty of issues with strict mode before

That’s why we add a warning in #3131 so that if someone has problems with strict mode they can notice ahead of time.

> > a) people can actually change this on their regular hosting?
> > Users will for sure have no control over the database server configuration.
> 
> The fix the installer suggests is using `SESSION` config. IMO that should work everywhere. But we should test it for sure.

I couldn’t find any information that suggests you need special permissions for `SET sql_mode`. The MySQL documentation even says “Each client can change its session sql_mode value at any time.”.

bytehead (dismissed):

aschempp:
> I couldn’t find any information that suggests you need special permissions for SET sql_mode. The MySQL documentation even says “Each client can change its session sql_mode value at any time.”.

Yes, we can/could do that through connection options at runtime. I did mean if someone can change the server defaults, which is unlikely I think.

leofeyer:
I have adjusted the explanation text. @m-vo WDYT?

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3153, opened 2021-07-06
Fix the remaining deprecated DBAL methods and some other issues

leofeyer:


core-bundle/src/Migration/Version404/Version447Update.php
```diff
@@ -110,9 +110,9 @@ public function run(): MigrationResult
                     active = '1' DESC, addedOn != '' DESC, id
             ");
 
-            $subscriptions->execute(['pid' => $duplicate['pid'], ':email' => $duplicate['email']]);
```
leofeyer, line 113: I wonder if using `':email'` as key has ever worked?

PR #3155, opened 2021-07-06
Use schema.org JSON-LD in the calendar-bundle

Toflar:
Same same as https://github.com/contao/contao/pull/3119 but for the calendar-bundle :)


calendar-bundle/src/Resources/contao/templates/events/event_list.html5
```diff
@@ -26,9 +26,17 @@
   <?php if ($this->details): ?>
     <?= $this->details ?>
   <?php else: ?>
-    <div class="ce_text block" itemprop="description">
+    <div class="ce_text block" class="description">
```
fritzmg:
```suggestion
    <div class="ce_text description block">
```
The `class` attribute already exists.

calendar-bundle/src/Resources/contao/templates/events/event_teaser.html5
```diff
@@ -5,26 +5,26 @@
...
+      <span class="name"><?= $this->location ?></span>
       <?php if ($this->address): ?>
-        (<span itemprop="address" itemscope itemtype="https://schema.org/PostalAddress"><?= $this->address ?></span>)
+        (<span class="address"><?= $this->address ?></span>)
       <?php endif; ?>
     </p>
   <?php endif; ?>
 
-  <div class="ce_text block" itemprop="description">
+  <div class="ce_text block" class="description">
```
fritzmg:
```suggestion
  <div class="ce_text description block">
```

fritzmg (changes requested):
The `class` attribute was duplicated in some places. However, we did not introduce new classes in #3119, so I think we should not introduce new CSS classes here either.

calendar-bundle/src/Resources/contao/classes/Events.php
```diff
@@ -579,6 +579,35 @@ protected function getDatesFromFormat(Date $objDate, $strFormat)
 				return array(0, time(), $GLOBALS['TL_LANG']['MSC']['cal_empty']);
 		}
 	}
+
+	public static function getBasicSchemaOrgData(CalendarEventsModel $objEvent): array
```
leofeyer, line 583: Why did you add this method but not an equivalent `News::getBasicSchemaOrgData()` method? In the `ModuleNews` class, this logic only exists in the anonymous function:

https://github.com/contao/contao/blob/829cbf481d20962388a5b0fce446c64c2a948e21/news-bundle/src/Resources/contao/modules/ModuleNews.php#L231-L260

Toflar: Because unfortunately the ModuleCalendarReader does not share any logic. In the news bundle we had a central `parseArticle()` which was reused in the reader as well. Here we don't have that, hence the new helper method.

leofeyer: Would you be OK with me renaming the method to `getSchemaOrgDataFromEvent()` and adding an equivalent method `getSchemaOrgDataFromNews()` to the `News` class? The helper methods might be useful in other cases, too.

leofeyer: NVM, I will create a separate PR hereto.

Toflar: Yeah, I think for the sake of consistency that would make sense 👍 But can also be a PR after this one, sure :)

leofeyer (approved):

leofeyer:
Thank you @Toflar.

calendar-bundle/src/Resources/contao/classes/Events.php
leofeyer: See #3161

PR #3156, opened 2021-07-06
Use schema.org JSON-LD in the breadcrumb module

Toflar:
Not sure if people need to be able to change this in the template at all...we could also just add this data in `ModuleBreadcrumb`. Wdyt @contao/developers?

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
@@ -220,6 +220,26 @@ class ModuleBreadcrumb extends Module
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


fritzmg:
If we allow it I would go the same route as in the other places, i.e. provide a `getBasicSchemaOrgData` function instead of building the array directly in the template.

Toflar:
Ah right, yes. Let's do that then.

Toflar:
Done :) Not sure about the coding style here (I thought we're using the old array notation in old modules) but the CS fixer does not require any changes? @leofeyer?

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3160, opened 2021-07-06
Add contao.import_user event

### LINKED ISSUE #3157: contao_import_user event missing
**Affected version(s)**

Contao 4.9

**Description**

According to https://docs.contao.org/dev/reference/hooks/importUser/ we shouldn't use the importUser hook anymore and use the event.

I couldn't find the event anywhere in the code.

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | https://github.com/contao/contao/issues/3157#issuecomment-874945013

Adds the missing `contao.import_user` event as a replacement for the deprecated `importUser` hook.

### DIFF
--- a/core-bundle/src/Event/ContaoCoreEvents.php
+++ b/core-bundle/src/Event/ContaoCoreEvents.php
@@ -76,4 +76,11 @@ final class ContaoCoreEvents
      * @see SlugValidCharactersEvent
      */
     public const SLUG_VALID_CHARACTERS = 'contao.slug_valid_characters';
+
+    /**
+     * The contao.import_user event is triggered when a username cannot be found in the database.
+     *
+     * @see ImportUserEvent
+     */
+    public const IMPORT_USER = 'contao.import_user';
 }
--- a/core-bundle/src/Event/ImportUserEvent.php
+++ b/core-bundle/src/Event/ImportUserEvent.php
@@ -31,6 +31,11 @@ class ImportUserEvent extends Event
      */
     private $table;
 
+    /**
+     * @var bool
+     */
+    private $isLoaded = false;
+
     public function __construct(string $username, string $password, string $table)
     {
         $this->username = $username;
@@ -51,5 +56,17 @@ class ImportUserEvent extends Event
     public function getTable(): string
     {
         return $this->table;
+    }
+
+    public function isLoaded(): bool
+    {
+        return $this->isLoaded;
+    }
+
+    public function setIsLoaded(bool $isLoaded): self
+    {
+        $this->isLoaded = $isLoaded;
+
+        return $this;
     }
 }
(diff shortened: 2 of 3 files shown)


m-vo (dismissed):
btw.: Why is this called `contao.import_user`? It's only a notification, isn't it? I'd probably would expect that the event allows storing a new user or something if it has 'import' in the name. :smile:

bytehead:
Indeed. The behaviour is somehow strange as you read the docs: 

> A record must exist in the database for Contao to load a user. Return true if you added the user to the respective table, or false if not.

https://docs.contao.org/dev/reference/hooks/importUser/#return-values

I'm open for naming proposals 😬

leofeyer (changes requested):
The hook returns `$blnLoaded` therefore the event should do, too. Something like this:

```php
$event = new ImportUserEvent($username, $password, $user->strTable);

System::getContainer()->get('event_dispatcher')->dispatch($event, ContaoCoreEvents::IMPORT_USER);

if (!$event->isLoaded() && self::triggerImportUserHook($username, $password, $user->strTable) === false)
{
    return null;
}
```

bytehead:
Like in 173fe9759?

m-vo:
Or maybe just an `addUser` method on the event? And then the dispatching code performs the db operation + sets loaded accordingly?

leofeyer:
> Like in [173fe97](https://github.com/contao/contao/commit/173fe97591b3eb39680199ceeeef1235b169f25b)?

Yes, I guess so. This at least reproduces the current behavior of the hook.

@contao/developers Do we want to keep it this way or do we want to improve the event logic?

fritzmg:
I am in favor of @m-vo 's suggestion to add an `addUser` or `setUser` method. The last event listener that sets a user would win (whereas wit the hook the first event listener that returns `true` wins).

bytehead:
I like the idea of @m-vo. But I'm not sure, which one should be the "winning" event?

m-vo:
>  But I'm not sure, which one should be the "winning" event?

Good question. I'd say the first one, then it reflects the event priority (edit: or does it actually not? :smile:). Maybe `setUser` can only be called once and throws on subsequent calls?

editedit: nah, I changed my mind. The last one should win. Each event can then decide to read the set user and alter it if not `null` or leave it as it is.

aschempp (changes requested):
I'm against adding this event. We agreed to not add events that simply replace hooks, and there is no advantage in using this event than the current hook. Also, the event is most likely not gonna stay in Contao 5 (with updated Symfony security).

leofeyer:
As discussed in Mumble on July 29th, we do not want to add the event, because it would most likely be removed again in Contao 5. Symfony security has changed massively and we will have to rework the implementation.

PR #3163, opened 2021-07-07
Remove left-over itemprop usages

leofeyer:
Follow-up on #3162

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -195,53 +195,6 @@ template as follows:
 <?php $this->endblock(); ?>
 ```
 
-### Template changes
-
-Adding the schema.org tags required to insert an additional `<span>` element
-into the following templates:
-
-`cal_default.html5`
-
-```php
-<!-- OLD -->
-<div class="event">
-  <a href="<?= $event['href'] ?>"><?= $event['link'] ?></a>
-</div>
-
-<!-- NEW -->
-<div class="event" itemscope itemtype="https://schema.org/Event">
-  <a href="<?= $event['href'] ?>" itemprop="url"><span itemprop="name"><?= $event['link'] ?></span></a>
-</div>
-```
-
-`mod_breadcrumb.html5`
-
-```php
-<!-- OLD -->
-<li>
-  <a href="<?= $item['href'] ?>"><?= $item['link'] ?></a>
-</li>
-
-<!-- NEW -->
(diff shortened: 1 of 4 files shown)


fritzmg (changes requested):
We don't have a JSON-LD solution for the navigation yet, do we?

Toflar:
Nope we don't. It's on my list.

leofeyer:
Do we need one at all? We currently "enrich" navigation menus by saying "this is a list of URLs" – something Google might have figured out from the `<a>` tags already. There is no real added value by using schema.org for navigation menus, is there?

Toflar:
Well that's not really an argument. It could also "guess" that when there's the word `event` in some CSS class that inside it might be an event. schema.org is here to simplify things and provide information in a machine-readable standardized way. I'll continue to provide PRs for 4.12 until we have 100% migrated away from microdata to schema.org. There's not much to do anymore :) Just revert the changes in the template and this PR can be merged :)

leofeyer:
Let's just keep this PR open until your next PR is merged. Then this is a reminder to also adjust the `UPGRADE.md` file.

Toflar:
I've tried to implement it now but while I was doing so, I figured that there's no real good example of how to use `SiteNavigationElement`. We certainly did not use it correctly because the nested microdata lack the description of some kind of list item. So imho that has also been incorrect microdata ever since.
I figured out that Google does not even support it at all. The full list can be found here: https://developers.google.com/search/docs/guides/search-gallery

So imho we can really ignore this. We have to update `mod_navigation` too, though.

leofeyer:
Updated in 2c4c6cfddc4de1f6f7e1cee586d6f443c77f1c37. There is one last occurrence:

https://github.com/contao/contao/blob/472be72880ce1d39723e349f2fc61710a70c9dcf/core-bundle/src/Resources/contao/templates/gallery/gallery_default.html5#L2

Do we still need this now that the image schema.org is handled differently?

Toflar:
I think we can remove that too as this is also not supported by Google as far as I could see (of course if we know of other platforms, we should add it but we should have some kind of RL example).
Looks like we could introduce better support for the search field, FAQ and maybe videos. Not related to this PR, though.

Toflar (approved):

m-vo:
For reference: there are currently also itemprop attributes used in the Twig macros of the image studio. 

As discussed with @Toflar I'm going to provide a PR for that along with the respective Twig function to add JSON-LD data.

PR #3164, opened 2021-07-07
Deprecate the getSearchablePages() method

aschempp:
fixes https://github.com/contao/contao/pull/3150#issuecomment-874158385


fritzmg (changes requested):
This should be a service and not a private function, otherwise we cannot retrieve the searchable pages elsewhere.

aschempp:
Why would you need to retrieve the searchable pages elsewhere? Also, I'll rename the method because these are not searchable pages 😂

Toflar (commented):
Implementation is correct, just requires the CI chain to pass. And no, it does not need to be a service. Nobody needs to access this list :)

fritzmg (dismissed):

core-bundle/src/Controller/SitemapController.php
```diff
@@ -110,4 +122,55 @@ private function callLegacyHook(PageModel $rootPage, array $pages): array
 
         return $pages;
     }
+
+    private function getPageAndArticleUrls(int $pid): array
```
Toflar: Shouldn't we use individual `SitemapEvent` listeners for articles and pages?

leofeyer:
Looks good to me. 👍🏻 Will need to fix the CI chain and answer https://github.com/contao/contao/pull/3164#discussion_r665501221 though.

core-bundle/src/Controller/SitemapController.php
aschempp: This would cause multiple issues, e.g. the pages would no longer be in the legacy hook.

Toflar (dismissed):

Toflar (dismissed):

leofeyer (approved):

fritzmg (approved):

PR #3166, opened 2021-07-07
Revert 'Lazy-load the `rootFallbackLanguage` property'

leofeyer:
This reverts #3067. The root fallback language model is already stored in the registry:

https://github.com/contao/contao/blob/99e04328a3b8eb06baec750aa8142ceb5804981e/core-bundle/src/Resources/contao/models/PageModel.php#L890-L899

Therefore, the changes from #3067 do reduce the number of calls to the `PageMode::findPublishedFallbackByHostname()` method, however, they do **not** reduce the number of DB queries.

Search the code for `contao.dns-fallback` to learn more about the special handling we already have.

### DIFF
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -358,40 +358,6 @@ class PageModel extends Model
 		return parent::__set($strKey, $varValue);
 	}
 
-	public function __get($strKey)
-	{
-		// Lazy-load the fallback language (see contao/core#6874)
-		if ($strKey == 'rootFallbackLanguage' && $this->blnDetailsLoaded && !\array_key_exists('rootFallbackLanguage', $this->arrData))
-		{
-			$this->rootFallbackLanguage = $this->rootIsFallback ? $this->language : null;
-
-			if (!$this->rootIsFallback && ($objFallback = static::findPublishedFallbackByHostname($this->domain)) !== null)
-			{
-				$this->rootFallbackLanguage = $objFallback->language;
-			}
-		}
-
-		return parent::__get($strKey);
-	}
-
-	public function __isset($strKey)
-	{
-		if ($strKey == 'rootFallbackLanguage' && $this->blnDetailsLoaded)
-		{
-			return true;
-		}
-
-		return parent::__isset($strKey);
-	}
-
-	public function row()
-	{
-		// Load lazy properties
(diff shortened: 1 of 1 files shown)


aschempp:
but it is not in the registry if noone requested it? So it will save one DB call?

leofeyer:
Yes, it will save one single DB call. 😂

leofeyer:
And make the code more complicated in exchange.

fritzmg (dismissed):

fritzmg:
Will this make https://github.com/contao/contao/pull/3165 obsolete then as well?

leofeyer:
No. The problem would no longer be triggered, but it would not be solved, either.

Toflar:
Since when is one DB call better than none? I don't understand this.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -1195,6 +1161,21 @@ public function loadDetails()
 			$this->rootIsPublic = ($objParentPage->published && (!$objParentPage->start || $objParentPage->start <= $time) && (!$objParentPage->stop || $objParentPage->stop > $time));
 			$this->rootIsFallback = (bool) $objParentPage->fallback;
 			$this->rootUseSSL = $objParentPage->useSSL;
+			$this->rootFallbackLanguage = $objParentPage->language;
+
+			// Store the fallback language (see #6874)
+			if (!$objParentPage->fallback)
+			{
+				$this->rootIsFallback = false;
```
```suggestion
```

This line can be omitted I think.

ausi (dismissed):

leofeyer:
> Since when is one DB call better than none? I don't understand this.

Saving one DB call is of course a good thing, however, in this case it does not only make the code more complicated but also causes follow-up issues such as #3165. Not really worth it IMHO.

Toflar:
The follow up issue was a bug, so what? I'm against merging this PR.

ausi:
> Since when is one DB call better than none?

I think it’s a tradeoff. More complicated “magic” code in order to save one DB query in a few cases is not worth it IMO.

ausi:
Also, I think in nearly all cases the fallback root page needs to be loaded anyways (if there is an image on the page for example).

ausi:
The initial PR got probalby only merged because of the statement:

> this SQL query would be done for EVERY `PageModel::loadDetails()` since it cannot be cached in the registry.

Which was not correct it seems now.

PR #3167, opened 2021-07-08
Use schema.org JSON-LD in the faq-bundle

Toflar:
Unfortunately, these modules are pretty old and the way the FAQ lists are generated is not even unified. Hence I just duplicated the schema.org stuff as well.

But hey, look where we got!!

Bildschirmfoto 2021-07-08 um 10 05 35


fritzmg (dismissed):

leofeyer (changes requested):
I don't think the changes are correct yet.

1. FAQ list == news list → should be treated the same way
2. FAQ reader == news reader → missing completely
3. Maybe add a helper method `ModuleFaq::getSchemaOrgData()`?

@Toflar I can work on these changes if you want.

Toflar:
I guess you should read https://developers.google.com/search/docs/data-types/faqpage first. FAQ readers do not get any schema.org data :)

leofeyer:
I see. But according to https://developers.google.com/search/docs/data-types/faqpage#content-guidelines, there are two preconditions that are not met in the FAQ list module:

> * Verwende FAQPage nur, wenn deine Seite eine Liste mit Fragen und Antworten enthält.
> * Alle FAQ-Inhalte müssen für die Nutzer auf der Quellseite sichtbar sein.

So must we only add the schema.org data in the FAQ page module?

Toflar:
Yeah I wasn't quite sure about list-reader so I just added it. I figured it's better to list it.

leofeyer:
I feel like we should only add the schema.org data in the FAQ page module, because "Alle FAQ-Inhalte müssen für die Nutzer auf der Quellseite sichtbar sein" is definitely not the case in the FAQ list module.

leofeyer:
I have discussed this with @ausi and according to the docs, the FAQ reader could be a `QAPage` type.

> A QAPage is a WebPage focussed on a specific Question and its Answer(s), e.g. in a question answering site or documenting Frequently Asked Questions (FAQs).

@Toflar WDYT?

Toflar:
I disagree. Which docs did you read? Because Google is pretty clear about it:

https://developers.google.com/search/docs/data-types/qapage

> Users must be able to submit answers to the question. Don't use QAPage markup for content that has only one answer for a given question with no way for users to add alternative answers; instead, use `FAQPage`

Our FAQ module does not support submitting answers. We only ever have one answer so we have to use `FAQPage`.

ausi:
> Which docs did you read?

It’s the text from https://schema.org/QAPage

But I agree, a FAQPage with a single question can probably still be seen as a “list” of one question with it’s answer.

leofeyer:
Ok, then let's add the schema.org data for the FAQ page module (multiple questions with their answers) and the FAQ reader module (single question with its answer). Agreed?

Toflar:
Yeah, that's probably the only option that fulfills

> "Alle FAQ-Inhalte müssen für die Nutzer auf der Quellseite sichtbar sein"

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3169, opened 2021-07-08
Code inspector cleanup

leofeyer:


core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -1195,7 +1195,7 @@ public static function getDecodedHostname()
 	{
 		$host = Environment::get('host');
 
-		if (strpos($host, 'xn--') !== 'false')
+		if (strpos($host, 'xn--') !== false)
```
leofeyer, line 1198: WTF? How can this have worked?

core-bundle/src/Resources/contao/dca/tl_files.php
```diff
@@ -803,7 +803,7 @@ public function editSource($row, $href, $label, $title, $icon, $attributes)
 		/** @var DC_Folder $dc */
 		$dc = (@func_get_arg(12) ?: null);
 
-		if (!in_array($objFile->extension, $dc ? $dc->editableFileTypes : StringUtil::trimsplit(',', strtolower($GLOBALS['TL_DCA'][$table]['config']['editableFileTypes'] ?? $GLOBALS['TL_CONFIG']['editableFiles'] ?? System::getContainer()->getParameter('contao.editable_files')))))
+		if (!in_array($objFile->extension, $dc->editableFileTypes ?? StringUtil::trimsplit(',', strtolower($GLOBALS['TL_DCA']['tl_files']['config']['editableFileTypes'] ?? $GLOBALS['TL_CONFIG']['editableFiles'] ?? System::getContainer()->getParameter('contao.editable_files')))))
```
leofeyer, line 806: We had been using an undefined variable `$table` here. 🤦🏻

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1177,7 +1177,7 @@ public function purgeSearchIndex(DataContainer $dc)
 
 		System::getContainer()
 			->get(PageUrlListener::class)
-			->purgeSearchIndex($dc)
+			->purgeSearchIndex($dc->id)
```
leofeyer: The method expects an ID and gets a `$dc` object. How can this have worked?

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -951,9 +951,9 @@ protected function doReplace($strBuffer, $blnCache)
 							$objFile = new File(rawurldecode($src));
 
 							// Add the image dimensions
-							if (($imgSize = $objFile->imageSize) !== false)
+							if (isset($objFile->imageSize[0], $objFile->imageSize[1]))
```
leofeyer, line 954: It seems that `$objFile->imageSize` cannot be `false` anymore (it could be in the past). @ausi Is my replacement correct?

core-bundle/src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -72,7 +72,7 @@ public function generate()
 		{
 			$this->targetPath = base64_decode($request->request->get('_target_path'));
 		}
-		elseif ($this->redirectBack && $request && $request->query->has('redirect'))
+		elseif ($this->redirectBack && $request->query->has('redirect'))
```
leofeyer: This is a nice one. We do check if `$request` is not `null` in the else condition but not in the if condition. Therefore I have removed the check completely. Would you prefer to add it again?

core-bundle/src/Resources/contao/classes/Backend.php
ausi: I guess because the condition always matched, `Idna::decode()` was always executed and it probably returns the domain as is if there is nothing to decode.

core-bundle/src/Resources/contao/modules/ModuleLogin.php
bytehead: Fine for me, the error when `$request` is `null` would occur on line 71 anyway.

core-bundle/src/Resources/contao/dca/tl_page.php
bytehead: It never worked, but didn't get noticed?

Toflar: I guess it just hasn't :)

core-bundle/src/Resources/contao/modules/ModuleLogin.php
leofeyer: We would have to add the check there as well if we decide to add it again.

leofeyer: Added in bcbd2ad02e32f42a4c41734073ccc59f9c125bf7 again.

core-bundle/src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -63,19 +63,16 @@ public function generate()
 			return $objTemplate->parse();
 		}
 
-		$container = System::getContainer();
-		$request = $container->get('request_stack')->getCurrentRequest();
```
leofeyer, line 67: This is already done in line 52. 🤦🏻

bytehead (dismissed):

comments-bundle/src/Resources/contao/classes/Comments.php
```diff
@@ -247,7 +247,6 @@ protected function renderCommentForm(FrontendTemplate $objTemplate, \stdClass $o
 		// Initialize the widgets
 		foreach ($arrFields as $arrField)
 		{
-			/** @var Widget $strClass */
```
m-vo, line 250: How can the class be inferred here? Shouldn't that be `/** @var Widget|null $strClass */`?

core-bundle/src/Resources/contao/controllers/BackendFile.php
```diff
@@ -84,11 +84,11 @@ public function run()
 		// Set the active record
 		if ($this->Database->tableExists($strTable))
 		{
-			/** @var Model $strModel */
 			$strModel = Model::getClassFromTable($strTable);
 
 			if (class_exists($strModel))
 			{
+				/** @var Model $objModel */
```
m-vo:
```suggestion
				/** @var Model|null $objModel */
```

core-bundle/src/Resources/contao/controllers/BackendPage.php
```diff
@@ -84,11 +84,11 @@ public function run()
 		// Set the active record
 		if ($this->Database->tableExists($strTable))
 		{
-			/** @var Model $strModel */
 			$strModel = Model::getClassFromTable($strTable);
 
 			if (class_exists($strModel))
 			{
+				/** @var Model $objModel */
```
m-vo:
```suggestion
				/** @var Model|null $objModel */
```

comments-bundle/src/Resources/contao/classes/Comments.php
leofeyer: No, `$strClass` is a string, isn't it?

core-bundle/src/Resources/contao/dca/tl_page.php
m-vo:
```suggestion
			->purgeSearchIndex((int) $dc->id)
```

comments-bundle/src/Resources/contao/classes/Comments.php
m-vo: You're right, I looked at the wrong line :see_no_evil:.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: Yes, is correct I think.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Validator.php
```diff
@@ -254,7 +254,7 @@ public static function isUuid($varValue)
 	/**
 	 * Valid binary UUID (version 1)
 	 *
-	 * @param mixed $varValue The value to be validated
+	 * @param string $varValue The value to be validated
```
m-vo: Did you intentionally only fix this one in the `Validator`class?

Nearly all of them only accept strings because they use string functions (like `strlen`). At least under PHP>=8 this produces warnings/errors if not called with a string.

ausi: AFAIK `strlen` can be used with all scalar types without notice or warning in PHP 8.

m-vo: Was it the preg functions then? 🤔 I remember running across this recently in some PR.

m-vo: At least for objects the error message states it must be a string. 🙃

> strlen(): Argument #1 ($str) must be of type string

leofeyer: I only changed it there because the code inspector complained about it. However, I agree that the change is wrong and I have fixed that in 444c3b1277afa488b92e7774128f0cbbacc64c9e.

ausi (approved):

ausi:
Thank you @leofeyer.

PR #3170, opened 2021-07-08
Add JSON-LD support for Twig

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | see https://github.com/contao/contao/pull/3163#issuecomment-875930255
| Docs PR or issue | todo

This refactors the image studio macros to use our new JSON-LD system instead of item* attributes (+ some housekeeping).

I registered the `addSchemaOrg` Twig function in the `ContaoTemplateExtension` extension for now. Note, that this is only temporary. Once #2988 is merged, I'd like to move everything in a single `ContaoExtension`. Having multiple runtimes is a good idea for SOC, but it turned out that multiple extensions doesn't make any sense for our use case (IMHO BC is not relevant in this case).

/cc @Toflar @ausi

### DIFF
--- a/core-bundle/src/Twig/Extension/ContaoTemplateExtension.php
+++ b/core-bundle/src/Twig/Extension/ContaoTemplateExtension.php
@@ -15,6 +15,7 @@ namespace Contao\CoreBundle\Twig\Extension;
 use Contao\BackendCustom;
 use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\CoreBundle\Routing\ScopeMatcher;
+use Contao\CoreBundle\Twig\Runtime\SchemaOrgRuntime;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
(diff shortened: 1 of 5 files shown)


ausi:
> IMHO BC is not relevant in this case

Should we add `@internal` to the affected classes to make that more clear?

m-vo:
> Should we add @internal to the affected classes to make that more clear?

I think the runtime is fine and can stay as it is. I'd remove the other extension classes, though. Even if you technically could have relied on them, I cannot think of a use case for that. Wdyt?

(+ All runtimes should have a private constructor anyway.)

Toflar (dismissed):

core-bundle/src/Twig/Extension/ContaoTemplateExtension.php
```diff
@@ -50,6 +51,7 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('render_contao_backend_template', [$this, 'renderContaoBackendTemplate']),
+            new TwigFunction('addSchemaOrg', [SchemaOrgRuntime::class, 'add']),
```
leofeyer: This should be `add_schema_org`, shouldn't it?

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3171, opened 2021-07-09
Code inspector cleanup

leofeyer:
Backports the fixes from #3169.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -900,7 +900,7 @@ class InsertTags extends Controller
 							// Add the image dimensions
 							if (isset($objFile->imageSize[0], $objFile->imageSize[1]))
 							{
-								$dimensions = ' width="' . StringUtil::specialchars($objFile->imageSize[0]) . '" height="' . StringUtil::specialchars($objFile->imageSize[1]) . '"';
+								$dimensions = ' width="' . $objFile->imageSize[0] . '" height="' . $objFile->imageSize[1] . '"';
 							}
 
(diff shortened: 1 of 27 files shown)


core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -898,9 +898,9 @@ protected function doReplace($strBuffer, $blnCache)
 							$objFile = new File(rawurldecode($src));
 
 							// Add the image dimensions
-							if (($imgSize = $objFile->imageSize) !== false)
+							if (isset($objFile->imageSize[0], $objFile->imageSize[1]))
 							{
-								$dimensions = ' width="' . StringUtil::specialchars($imgSize[0]) . '" height="' . StringUtil::specialchars($imgSize[1]) . '"';
+								$dimensions = ' width="' . StringUtil::specialchars($objFile->imageSize[0]) . '" height="' . StringUtil::specialchars($objFile->imageSize[1]) . '"';
```
leofeyer: @ausi Why are we using `StringUtil::specialchars()` here? Would an `(int)` cast not make more sense?

ausi: Both fine with me.

m-vo: The HTML spec says, `width` and `height` must contain "valid non-negative integers" ([spec](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width)). So I think a cast would be fine here.

leofeyer: Actually, we do not have to cast at all, because `$objFile->imageSize[0]` and `$objFile->imageSize[1]` are always integers. They are set from `ImageDimensions->getSize()->getWidth()` and `->getHeight()` and we use them without a cast or specialchars() in other places, too. Therefore dbd24bda659f1e9aed3190683aa1d2d289fe43c8.

PR #3173, opened 2021-07-09
Fix the PHP 8.1 CI job

m-vo:
Backports #3111

### DIFF
--- a/core-bundle/src/Resources/contao/classes/BackendUser.php
+++ b/core-bundle/src/Resources/contao/classes/BackendUser.php
@@ -601,20 +601,40 @@ class BackendUser extends User
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


m-vo:
With these changes the 4.9 CI should be green again.

I'm not sure how to deal with the deprecation notice we are triggering. In `4.x` we wrote `deprecated since 4.12` but now this is basically `deprecated since 4.9` I guess? @leofeyer Do you handle this when merging upstream?

leofeyer (approved):

leofeyer:
> @leofeyer Do you handle this when merging upstream?

Yes. 👍🏻

leofeyer:
Thank you @m-vo.

PR #3174, opened 2021-07-09
Correctly handle form fields in DC_Folder in "editAll" mode

### LINKED ISSUE #2577: Editing metadata of multiple files causes data loss

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2577
| Docs PR or issue | -

Supersedes #2981

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -37,7 +37,7 @@ abstract class DataContainer extends Backend
 {
 	/**
 	 * Current ID
-	 * @var integer
+	 * @var integer|string
 	 */
 	protected $intId;
 
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -324,7 +324,7 @@ abstract class DataContainer extends Backend
 		// Validate the field
 		if (Input::post('FORM_SUBMIT') == $this->strTable)
 		{
-			$suffix = ($this instanceof DC_Folder ? md5($this->intId) : $this->intId);
+			$suffix = $this->getFormFieldSuffix();
 			$key = (Input::get('act') == 'editAll') ? 'FORM_FIELDS_' . $suffix : 'FORM_FIELDS';
 
 			// Calculate the current palette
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -53,6 +53,12 @@ class DC_Folder extends DataContainer implements \listable, \editable
 	 */
 	protected $strRootDir;
 
+	/**
+	 * Initial ID of the record
+	 * @var integer
+	 */
+	protected $initialId;
```
m-vo, line 60: Is it really really an integer? :upside_down_face:

m-vo (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Folder.php
leofeyer: Probably as much as `$intId` is an integer in the `DC_Folder` class. 🙈

leofeyer: Changed in bc42b3345f7ca3ea6f1df8d514af50f8ee66660d.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1389,6 +1389,11 @@ public function addCtableTags($strTable, $intId, &$tags)
 		}
 	}
 
+	protected function getFormFieldSuffix()
```
m-vo, line 1397: Should this function get a return type hint? Or at least a phpdoc comment? Otherwise you again wouldn't know if you're dealing with an integer or a string.

leofeyer: Added in eb52224ac0a85237dc9e1b936cb47496ddc2387c.

m-vo (approved):

ausi (approved):

Toflar (approved):

PR #3176, opened 2021-07-09
Remove the service definition tests

leofeyer:
Because we mostly forget to add them when we add new services or listener, the extension tests are out of sync. I have now removed most of them and only kept those that test priorities or services that are only added if certain preconditions are met.


leofeyer:
It is probably easier to review if you look at what is left of the `ContaoCoreExtension` class:

https://github.com/contao/contao/blob/343d67039f3c5c97ec06fd3994268ecb2ef62564/core-bundle/tests/DependencyInjection/ContaoCoreExtensionTest.php

PR #3177, opened 2021-07-09
Exclude the migrations from coverage

leofeyer:
We do not require writing unit tests for migrations, because unit tests that only check if the SQL queries are passed as arguments are pointless. If a migration is more complex than just checking if a table and column exist, we **can** still add unit tests (we also have two migration test classes) but thanks to this change, we no longer **have to** add pointless test classes.


m-vo (dismissed):

m-vo:
@leofeyer Why don't we exclude `core-bundle/src/Entity` as well? Testing POPOs doesn't make much sense.

leofeyer:
Even though it will not have the same effect as this change, here you go: #3183

Issue #2808, opened 2021-02-26
Improve custom backend modules


dmolineus:
If I want to add a custom backend module I need to do implement the backend module controller, a backend menu listener and provide custom permission handling. I find it cumbersome and think Contao should provide a better way for it.

1. A custom backend module should automatically be added to the backend menu
2. A custom backend module should automatically be added listed in the backend module permissions list
3. A custom backend controller should also be able to belong to another backend module. The backend menu should then automatically indicate the right active navigation item
4. Permission check should be done automatically

To achieve this goal I propose a new service tag/annotation with following information:

1. `name` Use to identify an active backend module and handle the permissions
2. `category` The category the backend module should be belongs to
3. `parent` Instead of defining a category you might define a parent module. Then no navigation item is added by default

If we want to get it one step further we could also allow to define which backend menu `menu = 'mainMenu'` should be used, for instance to adding a menu to the top menu.

aschempp:
I think this was similar to what we recently discussed with @richardhj (related to https://github.com/contao/contao/pull/2748). To me we certainly can/should improve on the backend controllers. Similar to the fragments, our first draft/implementation did not include abstract classes etc, we only added that after working with it. Unfortunately, I have never worked with backend controllers myself, so I don't have any experience what to improve.

dmolineus:
Thanks for that reference. Missed that @richardhj has proposed a similar approach.

richardhj:
We need to re-think a concept for backend controllers. The approach of using fragments (see #2748) might be a good fit for the dashboard but not for backend modules (i.e., entire pages). If you can think of a concept of an abstract backend controller, that would be great. I'll put effort into this as well soon.

dmolineus:
I have thought about that. Honestly I don't think there is a need for something like fragments for usual backend module. I think a way to integrate it into the backend navigation and the permission handling would as proposed here and in your PR would be enough.

m-vo:
possibly related #3180

PR #3180, opened 2021-07-09
Render be_main with custom back end controller

### LINKED ISSUE #3020: Option to modify mainHeadline from within BackendModule
When building a custom BackendModule there is no way to modify the mainHeadline to anything other than the module name itself which results in a different UX than in the rest of the Contao backend.

The core itself has it's own logic on how the mainHeadline is set up depending on the current table, ptable, key etc. but there seems to be no way to modify the headline in a custom BackendModule.

These are the lines in question

https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Resources/contao/classes/Backend.php#L432-L439

I'd like to propose a minimalistic and clean way (no breaking changes, no warnings) to change that.
How about we just pass on the main template object to the custom modules?

```php
/** @var Module $objCallback */
$objCallback = new $arrModule['callback']($dc);
$objCallback->mainTemplate = $this->Template;
$this->Template->main .= $objCallback->generate();
```

Happy to hear about your thoughts and maybe some alternative solutions :)

m-vo:
With support to overwrite html5 templates from Twig, rendering own back end routes ([currently done like this](https://docs.contao.org/dev/guides/back-end-routes/)) could become this:

```php
/**
 * @Route("/contao/my-backend-route", name=MyBackendController::class, defaults={"_scope" = "backend"})
 */
class MyBackendController extends AbstractCustomBackendController
{
    {
        return $this->render(
            'my_backend_route.html.twig',
            [
                'error' => 'Oh no, an error!',
                'title' => 'My title',
                'version' => 'I can overwrite what I want',
                'foo' => 'bar',
            ]
        );
    }
}
```

```twig
{% extends "@Contao/be_main" %}
```

As you can see, the example template just extends from `be_main` and can overwrite and reuse whatever is in there. For example:

```twig
{% extends "@Contao/be_main" %}

{% block main_headline %}Headline{% endblock %}

{% block main_content %}
    <div class="tl_listing_container">
        Main Content: {{ foo }}
    </div>
{% endblock %}
```

The relevant back end context is provided by the `AbstractCustomBackendController` that enhances the `render` function. This is IMHO much more flexible than the `render_contao_backend_template` function (which I would like to deprecate).


m-vo:
Maybe we could even use the same generic abstract back end controller for our core modules as well. Then there would be no difference between core and custom modules.

core-bundle/src/Controller/AbstractCustomBackendController.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class AbstractCustomBackendController extends AbstractController
+{
+    /**
+     * Render a Twig template with additional context for `@Contao/be_main`.
+     */
+    protected function render(string $view, array $parameters = [], Response $response = null): Response
+    {
+        $backendContext = (new class() extends BackendMain {
```
richardhj: What I don't like is that this PR builds on the `BackendMain` class 🙈 Or isn't this an issue because people extending `AbstractCustomBackendController` don't know about the legacy?
I mean, we could also build something new, no?

m-vo: Yes, it should be decoupled/implementation detail. But this is only a draft for now. I'm pretty sure there's more that can be done. 🙃

aschempp:
Maybe @sheeep has some input on this? Also, we shouldn't name it "Custom", its not `AbstractCustomFragmentController` either 😅

core-bundle/src/Controller/AbstractCustomBackendController.php
```diff
@@ -0,0 +1,50 @@
...
+use Contao\Environment;
+use Contao\Input;
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class AbstractCustomBackendController extends AbstractController
+{
+    /**
+     * Render a Twig template with additional context for `@Contao/be_main`.
+     */
+    protected function render(string $view, array $parameters = [], Response $response = null): Response
```
richardhj: We should implement an `abstract function getResponse()` here as well.
This is why:
- You won't be able to render a Twig template decoupled from the `BackendMain` controller otherwise.
- Implementing `getResponse()` will be part of the contract, making it required to implement for developers. This is important because we must make sure the business logic in `BackendMain::__construct()` will be executed.

richardhj:
After giving it a thought, I like the approach of this PR. What is missing, though, is that backend modules are automatically registered in the backend navigation and permissions (see #2808 for the request and #2748 for how I implemented it).

> Maybe we could even use the same generic abstract back end controller for our core modules as well. Then there would be no difference between core and custom modules.

This is correct!

core-bundle/src/Controller/AbstractCustomBackendController.php
richardhj: Oh wait, this would disallow using multiple controllers per class, which is valid as well.

m-vo:
Thanks for the feedback. I'll incorporate it in this PR soon.

m-vo:
> Maybe we could even use the same generic abstract back end controller for our core modules as well.

There is no real sense in using a common back end controller (which in itself would be useful) for the core modules as long as there is a need to keep BC for the legacy endpoints `BackendMain`, `BackendFile` and so on. It would be a totally different story if we would drop these but that could only be done in Contao 5.

### Why an abstract common thing?
Two things are IMHO helpful to be generalized:
1. The context needed to render the 'back end chrome', i.e. menus, title and so on.
2. A new link in the main menu that points to the route of this controller. 

Surprisingly, the context for 1. wouldn't contain much if we used new templates (and some Twig features): Menus would just become includes, translated/fix strings would just be used directly, things like `Environment::get('base')` would get `{{ app.request.baseUrl }}`, assets would be handled by the asset component… Besides the things you would set in your controller anyway - say title and headline - there would probably be so few things left that we should just move them to a request sensitive template function. Then you could just render any template extending `be_main` in your controller that itself would not need a base class at all.

A helper for 2. could be a service tag together with a new service annotation. We would then have a registry getting the tagged services and add them to menu at once.

### Full page vs fragment/dashboard widget
IMHO we're mainly dealing with one controller per page in the back end (Sf style), there is hardly any composition needed. Exceptions from this are (flash) messages which are already handled and maybe the home dashboard. But as the composition is always in the hand of a developer I'd say you could just alter the template and add your things directly. In case it's a component you could use `{{ render(controller(foo, …)) }}`.

-----

I feel this issue becomes more targeted towards templates than controllers. Maybe we should postpone it after we found a way to transition to Twig and then make the changes in the new templates.

m-vo:
I updated this PR, optimized a bit and added tests, but I kept the initial scope. This solution is not perfect but IMHO a big improvement nevertheless. 

I renamed the controller to `AbstractBackendController`. If this is a big too bold, I'm fine with going back to something like `AbstractCustomBackendController`. Apart from that it basically works like in the initial description.

aschempp:
I don't really understand what is better here compared to using the already-included Twig template by @sheeep ?
https://docs.contao.org/dev/guides/back-end-routes/

I recently used this in https://github.com/terminal42/contao-image-delete/blob/main/src/Controller/ImageDeleteController.php#L88-L96 and it worked pretty flawlessly.

m-vo:
> I don't really understand what is better here compared to using the already-included Twig template by @sheeep ?

See the initial description. You are now not limited anymore to the three predefined blocks. You can change whatever you want.

ausi (dismissed):
❤️

fritzmg:
I must admit, I also do not really get what's now possible that wasn't before 🙃 You are still limited to the blocks defined in the parent template, aren't you? Can you give a more advanced example? The example in the initial description looks pretty much the same as what you already were able to do.

richardhj:
> I kept the initial scope.

IMHO this PR should be able to fix #2808 and handle both backend permissions and navigation (both optional) to have a greater impact 🙂

aschempp:
I agree with @fritzmg. The current support for backend controllers already allows you to do everything – it's just a default template for the back end. You _could_ do anything else. But maybe I (and Fritz) are missing something 🙃

m-vo:
> I must admit, I also do not really get what's now possible that wasn't before upside_down_face You are still limited to the blocks defined in the parent template, aren't you?

The old `@ContaoCore/Backend/be_page.html.twig` does not contain HTML, there is just a function that renders exactly 3 blocks:
```twig
{{ render_contao_backend_template({
    main: block('main'),
    error: block('error'),
    headline: block('headline')
}) | raw }}
```

With the possibility to inherit PHP templates from Twig, we can now use the _real_ `be_main` template. 

Not only can you control every key in the context (`title`, `version`, …), we can also easily add more blocks to the template - I exemplary did so with `main_content`.

fritzmg (dismissed):

core-bundle/tests/Controller/AbstractBackendControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+        ];
+
+        $GLOBALS['TL_LANGUAGE'] = 'en';
+
+        $_SERVER['HTTP_USER_AGENT'] = 'Contao/Foo';
+        $_SERVER['HTTP_HOST'] = 'localhost';
+
+        if (!\defined('TL_FILES_URL')) {
+            \define('TL_FILES_URL', '');
+        }
```
leofeyer, line 59: This seems wrong. Why is it necessary?

core-bundle/tests/Controller/AbstractBackendControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+                ['@ContaoCore/Backend/be_header_menu.html.twig', [], '<header_menu>'],
+                ['custom_be.html.twig', $expectedContext, '<custom_be_main>'],
+            ])
+        ;
+
+        $container->set('security.authorization_checker', $authorizationChecker);
+        $container->set('security.token_storage', $this->createMock(TokenStorageInterface::class));
+        $container->set('contao.security.token_checker', $this->createMock(TokenChecker::class));
+        $container->set('database_connection', $this->createMock(Connection::class));
+        $container->set('session', $this->createMock(Session::class));
```
leofeyer, line 115: Why are you mocking the deprecated `Contao\Session` class instead of the Symfony session?

leofeyer (changes requested):

core-bundle/tests/Controller/AbstractBackendControllerTest.php
m-vo: Good catch, that was just a mistake taking the wrong namespace. Fixed in cb50a0a417eda659af367ffe25f1f75b401c76c8

core-bundle/tests/Controller/AbstractBackendControllerTest.php
m-vo: Otherwise `Controller::setStaticUrls()` wants to set them:

https://github.com/contao/contao/blob/724fd43e33dc8bcaa720a71142bb521688a1c71b/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1881-L1904

In which case you need more setup (files context, …) but the constant will be defined no matter what.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3181, opened 2021-07-10
CS fixes that went unnoticed

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

Currently our legacy and template checks do not work due to an ecs cache issue.

### DIFF
--- a/calendar-bundle/src/Resources/contao/templates/events/event_full.html5
+++ b/calendar-bundle/src/Resources/contao/templates/events/event_full.html5
@@ -48,7 +48,7 @@
 $schemaOrg = $this->getSchemaOrgData();
 
 if ($this->hasDetails()) {
-  $schemaOrg['description'] = $this->rawHtmlToPlainText($this->details);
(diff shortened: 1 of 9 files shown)


leofeyer:
Can we work around the issue somehow? You have mentioned that inserting `rm -rf /tmp/ecs_<container>` after each step.

m-vo:
I'll see if we can do this programmatically in the config files or if this is already to late. I fear it will depend on the inner workings of the library, though.

ausi (dismissed):

m-vo:
There is already a fix https://github.com/symplify/symplify/pull/3399. I did not test it, yet.

m-vo:
See https://github.com/symplify/symplify/pull/3403 - this is now in 9.4.21. I updated the `tools/ecs(composer.lock` file accordingly.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3182, opened 2021-07-11
Simplify Controller::addImageToTemplate tests

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

I initially wrote the `FigureBuilderIntegrationTest`s to make sure the new version of `Controller::addImageToTemplate` (using the `FigureBuilder` internally) behaves identical to the original implementation. The tests were valuable back then but they encapsulate an insane amount of complexity like setting up the legacy framework, hacking the model registry and so on. 

While working on an update to #2518 I realized there are a lot of complication with these tests, therefore this PR.  :slightly_smiling_face: They are now replaced with a much simpler version that mocks things instead.


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1625,7 +1625,7 @@ public static function addImageToTemplate($template, array $rowData, $maxWidth =
 			}
 
 			$size[0] = $maxWidth;
-			$size[1] = floor($maxWidth * ($height / $width));
+			$size[1] = (int) floor($maxWidth * ($height / $width));
```
m-vo, line 1628: I noticed this inconsistency when writing the new tests. Turns out `floor` returns a `float` :facepalm:, so the size array looked like `[100, 75.0, 'crop']`. Apparently did not cause problems but isn't correct either.

ausi (dismissed):

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -24,8 +24,10 @@
  * use the provided legacy helper methods to manually apply the data to them.
  *
  * Wherever possible, the actual data is only requested/built on demand.
+ *
+ * @final
  */
-final class Figure
+class Figure
```
leofeyer, line 30: Perfect example why classes should only be `final` if they implement an interface. 🤓 See https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -24,8 +24,10 @@
  * use the provided legacy helper methods to manually apply the data to them.
  *
  * Wherever possible, the actual data is only requested/built on demand.
+ *
+ * @final
```
leofeyer, line 28: We are not officially supporting the `@final` annotation.

leofeyer (approved):

core-bundle/src/Image/Studio/Figure.php
m-vo: I personally go for the strategy: Use `final` everywhere until you (or your users) have got a use case for it to make the API surface as small as possible. This is where I disagree with the article: I do not need an interface until then and the need for mocking does not justify adding one.

Using `class final` should IMHO be preferred to `/** @final  */` because it's more strict, but it basically says the same: Do not inherit from this class and expect things to work. If mocking is helpful for a test, I do not feel bad to switch to the annotation. :smile: There are concepts in other programming languages ("friends") that would allow to be more specific while still being strict about the API but PHP doesn't have these (yet?).

core-bundle/src/Image/Studio/Figure.php
m-vo: Maybe we should? :smiley_cat:

leofeyer: Yes, however, using `@final` did not have any effect in PhpStorm. I did not even get noticed that I am trying to extend a final class, so I guess the effect will be limited. 🤷🏻

m-vo: > Yes, however, using @final did not have any effect in PhpStorm.

:flushed: That's a bad thing indeed.

The annotation is still used a lot in Symfony and other projects, so I think devs are used to it.

m-vo: We could also use https://github.com/dg/bypass-finals (the lib is basically a stream wrapper)… not sure that's worth it, though.

ausi: > Yes, however, using `@final` did not have any effect in PhpStorm.

I added a comment to https://youtrack.jetbrains.com/issue/WI-31225

leofeyer: https://github.com/contao/contao/pull/3185

leofeyer:
Thank you @m-vo.

Issue #924, opened 2019-11-04
Declaring classes as final


leofeyer:
**Why?**

https://ocramius.github.io/blog/when-to-declare-classes-final/

**If and how?**

That remains to be discussed. 😄 

**Related**

https://github.com/contao/docs/pull/138

aschempp:
Honestly, I don't really agree with this. For the mentioned case, we are using `private` methods in our classes. Making classes `final` does not bring any advantage, because an inheriting class can still only call the public methods. If a method is `protected`, then it is **explicitly** mean to be used in child classes. Also, `final` conflicts a lot with unit tests (as far as I know).

m-vo:
I agree with the reasoning in the linked article. It's imho good practice - not only in PHP ("inheritance breaks encapsulation").

There might be a problem with mocking or using proxies like [Ocramius/ProxyManager](https://github.com/Ocramius/ProxyManager) (how ironically). On the other hand there should probably be **interfaces** in that case (also for mocking) rendering this problem obsolete.

But again, that's also in the article:
>  Final classes only work effectively under following assumptions:
>
>    There is an abstraction (interface) that the final class implements
>   All of the public API of the final class is part of that interface

Maybe also have a look at these:
* https://stackoverflow.com/questions/11730352/circumstances-to-make-a-class-final
* https://stackoverflow.com/questions/9344935/inheritance-breaking-encapsulation

leofeyer:
I also agree with the reasoning in the article.

leofeyer:
We should at least make the `@internal` classes final in Contao 6.

m-vo:
Yes, but only if they implement an interface or do not require mocking. 🙃 Otherwise I'd vote for using the `@final` annotation.

PR #3185, opened 2021-07-12
Allow mocking final classes

leofeyer:
@m-vo and I have discussed the `@final` annotation here https://github.com/contao/contao/pull/3182/files#r667789346, because the class needs to be mocked and therefore cannot be `final` anymore. This PR solves the problem by using [nunomaduro/mock-final-classes](https://github.com/nunomaduro/mock-final-classes), which is based on [dg/bypass-finals](https://github.com/dg/bypass-finals) and simply removes the `final` keyword before running the unit tests.

This would not only solve the `@final` problem but also allow us to declare many of our `@internal` classes (such as listeners etc.) as `final`. @contao/developers WDYT?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -153,6 +153,7 @@
         "bamarni/composer-bin-plugin": "^1.4",
         "composer/composer": "^1.0 || ^2.0",
         "contao/test-case": "^4.4",
+        "nunomaduro/mock-final-classes": "^1.0",
         "phpunit/phpunit": "^8.5",
         "symfony/browser-kit": "4.4.* || 5.2.*",
         "symfony/phpunit-bridge": "^5.2"
--- a/core-bundle/composer.json
+++ b/core-bundle/composer.json
@@ -126,6 +126,7 @@
         "contao/maintenance-bundle-deprecated": "^2.1.5",
         "contao/manager-plugin": "^2.3.1",
         "contao/test-case": "^4.4",
+        "nunomaduro/mock-final-classes": "^1.0",
         "phpunit/phpunit": "^8.5",
         "symfony/browser-kit": "4.4.* || 5.2.*",
         "symfony/http-client": "4.4.* || 5.2.*",


ausi:
I would prefer using `@final`. This is supported by phpstan, psalm and hopefully soon in PHPStorm: https://youtrack.jetbrains.com/issue/WI-31225

aschempp:
we also discussed `final` classes a lot and "agreed" they don't meet our requirements and we don't want to use them 😉

Toflar:
> "agreed" they don't meet our requirements and we don't want to use them 😉

I don't remember agreeing to not using final classes. Because I think, we should use them as much as possible 🤷‍♂️

leofeyer:
Me too. Also, for supposedly agreeing not to use final classes, we have quite a few in the codebase. 😛

Toflar:
I actually like that library. No idea how much it slows down the test run but otherwise 👍

leofeyer:
Not noticeably.

leofeyer:
> how much it slows down the test run

Not noticeably.

aschempp:
Yet another case where people remember different things… I feel like this just happened a lot lately. Maybe we need some sort of protocol of our discussions?

Btw it was the same discussion where we agreed to use `@internal` and our BC promise. A lot of the flexibility of Contao comes from the fact that stuff is extendible and the core can be overridden in a lot of places. Even though I agree that's not necessarily best practice, it is hugely relevant for the flexibility of the system and not-so-experienced developers.

leofeyer:
Related: #924

m-vo:
In my opinion there are two aspects to making things final or not and they do not generalize well. 

1) Some say: use final wherever possible, never mock it because you need interfaces anyway. And by that effectively saying: if you need to introduce an interface just to mock something your classes are coupled too much and you should use an interface in the first place (i.e. test against a contract not an implementation).

2) Then there is: If you introduce a new class make it final (update your IDE defaults, now). If you need inheritance, remove it. If you need to mock it, make it an annotation. If your users need it and you think it's correct or if your intention is to provide a base class, then of course remove it. It's not about about a software design paradigm but about maintainability of an API that has a small surface by default. Using @internal is similarly in this context.

The discussion @leo and I had, started about the `Figure` class. There **is** coupling between this class and the `FigureBuilder` and it's by design. The `Figure` class acts as a value object that people from standpoint 1 will "never mock", "just use". Our legacy logic in there is wrong from this standpoint but it's still correct for this use case. So, the tradeoff is: keep it final (keyword or annotation) but allow mocking it. And it's my pragmatism of standpoint 2 that feels totally comfortable with using an annotation now that I need mocking it.

See the complexity of this example alone? I don't think we can generalize a rule where you should or should not use it. I'd say go for standpoint 2 as long as you understand 1 but I would certainly not add `@final`annotations to all legacy classes for instance.

leofeyer:
Legacy code would not be changed in this regard anyway.

aschempp:
@m-vo sounds about right here. Just keep in mind that the flexibility of Contao comes from the fact that a lot of stuff is not "small surface" but intentionally changeable by anyone, event with limited knowledge about PHP internals. People are starting to use Reflections because we make methods private.

Regardings Figure, I really wonder if that is not supposed to be mocked? The Figure object is passed around I suppose, or is the point that it can & will never be a service argument?

m-vo:
> Regardings Figure, I really wonder if that is not supposed to be mocked? The Figure object is passed around I suppose, or is the point that it can & will never be a service argument?

Normally, you would just set the things you need in the constructor (the object is effectively an immutable value object). The only behavior that may need mocking is the `applyLegacy` stuff. But I'd like to remove that in upcoming Contao versions anyway.

I agree that Contao's flexibility is a big plus! We should IMHO promote composition for that instead of inheritance. 

>  People are starting to use Reflections because we make methods private.

Do you have evidence for this or an example? Because we could likely improve things if we'd knew the use cases, I think. :relaxed:

aschempp:
> Do you have evidence for this or an example? Because we could likely improve things if we'd knew the use cases, I think. ☺️

https://github.com/contao/contao/pull/3093

m-vo:
>> Do you have evidence for this or an example? Because we could likely improve things if we'd knew the use cases, I think. relaxed
>
> #3093

Maybe I'm missing the point, but you wrote in that ticket:

> keeping internal methods private means we can change them at any time if necessary. As soon as they are public, they become official API that we need to support in our backwards compatibility promise. These methods clearly are internal logic to me, that's why I'm trying to find the best way to support your use case but also limit the maintenance burden on the core devs blush

And I fully agree on that.  :smile:

leofeyer:
As discussed in Mumble on July 29th, we want to use `@final` for the `Figure` class as long as the class still contains legacy code. In Contao 5, when we get rid of the legacy code, we want to replace the `@final` with `final class`.

PR #3191, opened 2021-07-13
Use v2 of league/commonmark

Toflar:
Looks like league/commonmark is released as v2 very soon so I think we should ship our own 4.12 RC already with the updated version. Timing looks like we'll be able to use v2 stable soon :)

See https://twitter.com/colinodell/status/1413951540885344261


Toflar:
Oh, it requires PHP 7.4 🙈 Makes sense to me as 7.3 is security fixes only.
What are we doing now @contao/developers? Providing a brand new content element with the latest version of the library sounds desirable somehow...

m-vo:
Can we be compatible to both versions?

Toflar:
Well, then we can also just leave it as is and merge it for 4.13. I guess we'll bump the minimum PHP version there anyway (it's about time).

ausi:
> Oh, it requires PHP 7.4 🙈 Makes sense to me as 7.3 is security fixes only.

+1 for requiring PHP 7.4 for Contao 4.12.

composer.json
```diff
@@ -59,7 +59,7 @@
         "knplabs/knp-menu-bundle": "^2.1 || ^3.0",
         "knplabs/knp-time-bundle": "^1.10",
         "lcobucci/jwt": "^3.4 || ^4.0",
-        "league/commonmark": "^1.6",
+        "league/commonmark": "^2.0@dev",
```
```suggestion
        "league/commonmark": "^2.0@RC",
```
Using `@RC` should be possible here I think, and should be better compatible with installing the Contao RC version.

m-vo:
I just realized it's only 3,5 months after the 4.12 release that PHP7.3 will be EOL. So I'm all for raising the requirement to 7.4 as well. 👍

aschempp (dismissed):
`@dev` requirements are not propagated to the root package. Merging this would mean Contao 4.x can no longer be installed without lowering `minimum-stability` in a root package. I suggest to change this and only merge if the package has been released in a stable version.

ausi:
> `@dev` requirements are not propagated to the root package.

But `@RC` should work I think https://github.com/contao/contao/pull/3191#discussion_r669136391 (as long as Contao is `RC` itself)

But it would be nice if we could get an ETA on the relase date of the stable version of the library.

Toflar:
I've updated the PR to the state it would be once the release is here and converted it to a draft for now.

m-vo:
Looks like there is a stable release by now: https://github.com/thephpleague/commonmark/releases/tag/2.0.0

Toflar:
Ready for review then. Failing CS does not seem to be related.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3195, opened 2021-07-14
Do not prefix our JSON-LD context

ausi:
With the answer from https://github.com/schemaorg/schemaorg/issues/2921#issuecomment-879817283 it should be safe now to use the “correct” `@context` format for our own json-ld data.

### DIFF
--- a/core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
+++ b/core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
@@ -18,48 +18,18 @@ use Spatie\SchemaOrg\BaseType;
 class ContaoPageSchema extends BaseType
 {
     /**
-     * @var string
-     */
-    private $title;
-
-    /**
-     * @var int
-     */
-    private $pageId;
-
-    /**
-     * @var bool
-     */
-    private $noSearch;
-
-    /**
-     * @var bool
-     */
-    private $protected;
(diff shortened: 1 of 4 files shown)


core-bundle/src/Routing/ResponseContext/JsonLd/ContaoPageSchema.php
```diff
@@ -69,22 +69,19 @@ public function getContext(): string
...
 
     public function toArray(): array
     {
-        $this->setProperty('contao:title', $this->title);
-        $this->setProperty('contao:pageId', $this->pageId);
-        $this->setProperty('contao:noSearch', $this->noSearch);
-        $this->setProperty('contao:protected', $this->protected);
-        $this->setProperty('contao:groups', $this->groups);
-        $this->setProperty('contao:fePreview', $this->fePreview);
+        $this->setProperty('title', $this->title);
```
Toflar: Do we even still need to override `toArray()`? We could read and write with `$this->setProperty()` and `$this->getProperty()` in our setters and getters directly now?

ausi: Done in bf0b853e98ab9146a3e709d72c53aeeaf86dd23b

Toflar (dismissed):
<3

leofeyer:
Thank you @ausi.

PR #3196, opened 2021-07-15
Support multiple JSON+LD objects in the search document

qzminski:
The search document does not seem to properly extract the linked data from `<script>` tags that contain multiple JSON+LD elements:

```html
<script type="application/ld+json">
[
    {
        "@context": "https:\/\/contao.org\/",
        "@type": "Page",
        "foobar": true
    },
    {
        "@context": "https:\/\/contao.org\/",
        "@type": "Page",
        "foobar": false
    }
]
</script>
```

/cc @Toflar

### DIFF
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -156,7 +156,7 @@ class Document
             foreach ($jsonLdItems as $jsonLdItem) {
                 if (\is_array($graphs = $jsonLdItem['@graph'] ?? null)) {
                     foreach ($graphs as $graph) {
-                        $this->jsonLds[] = array_merge(['@context' => $jsonLdItem['@context']], $graph);
+                        $this->jsonLds[] = array_merge(array_diff_key($jsonLdItem, ['@graph' => null]), $graph);
                     }
                 } else {
                     $this->jsonLds[] = $jsonLdItem;
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -154,7 +154,7 @@ class Document
 
             // Parsed the grouped values under the @graph within the same context
             foreach ($jsonLdItems as $jsonLdItem) {
-                if (isset($jsonLdItem['@graph']) && is_array($jsonLdItem['@graph'])) {
+                if (isset($jsonLdItem['@graph']) && \is_array($jsonLdItem['@graph'])) {
                     foreach ($jsonLdItem['@graph'] as $graph) {
                         $this->jsonLds[] = array_merge(['@context' => $jsonLdItem['@context']], $graph);
                     }
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -130,7 +130,7 @@ class Document
             return $this->jsonLds;
         }
 
-        $this->jsonLds = $this->getContentCrawler()
+        $jsonLds = $this->getContentCrawler()
             ->filterXPath('descendant-or-self::script[@type = "application/ld+json"]')
             ->each(
                 static function (Crawler $node) {
@@ -146,7 +146,19 @@ class Document
         ;
 
         // Filter invalid (null) values
-        $this->jsonLds = array_filter($this->jsonLds);
+        $jsonLds = array_filter($jsonLds);
+
+        // Go through all data entries and check if any array has numeric keys. If yes, it likely contains multiple data inside it
+        // which should be treated as if coming from separate sources, and thus moved to the root of $jsonLds array.
+        foreach ($jsonLds as $jsonLd) {
+            if (array_keys($jsonLd) === range(0, \count($jsonLd) - 1)) {
+                foreach ($jsonLd as $v) {
+                    $this->jsonLds[] = $v;
+                }
+            } else {
+                $this->jsonLds[] = $jsonLd;
+            }
+        }
 
         return $this->filterJsonLd($this->jsonLds, $context, $type);
     }
--- a/core-bundle/src/Search/Document.php
+++ b/core-bundle/src/Search/Document.php
@@ -154,8 +154,8 @@ class Document
 
             // Parsed the grouped values under the @graph within the same context
             foreach ($jsonLdItems as $jsonLdItem) {
-                if (isset($jsonLdItem['@graph']) && \is_array($jsonLdItem['@graph'])) {
-                    foreach ($jsonLdItem['@graph'] as $graph) {
+                if (\is_array($graphs = $jsonLdItem['@graph'] ?? null)) {
+                    foreach ($graphs as $graph) {
                         $this->jsonLds[] = array_merge(['@context' => $jsonLdItem['@context']], $graph);
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


Toflar:
In that case, imho, one should win. We should not extract both. Either the first or the last must win. Wdyt @ausi ?

Toflar:
Ah no, probably not because you can have multiple `@Article` etc.

ausi:
> In that case, imho, one should win.

But that is the responsitibilty of the indexer I think?

Looks like it already does that: https://github.com/contao/contao/blob/b686423f6e88c3874ebda0bfdc8114d94c3f3162/core-bundle/src/Search/Indexer/DefaultIndexer.php#L160-L161

Toflar:
Yes, that's true. So what problem exactly does this PR fix then? I somehow cannot spot it :D

ausi:
> So what problem exactly does this PR fix then?

If I understand it correctly, this worked already:

```
<script type="application/ld+json">
{
    …
}
</script>
<script type="application/ld+json">
{
    …
}
</script>
```

But the follwing did not:

```
<script type="application/ld+json">
[
    {
        …
    },
    {
        …
    }
]
</script>
```

qzminski:
That's right @ausi.

Toflar:
Ah, now I understand. And we cannot check if it was an object or an array because for PHP it's always an array?

qzminski:
It's always an array – either associative or non-associative. Btw I tried to use `ArrayUtil::isAssoc` method for that but unit tests were failing like if that class was not loaded… not sure about it, so I copied the function logic to `Document` class.

core-bundle/src/Search/Document.php
```diff
@@ -146,7 +146,19 @@ static function (Crawler $node) {
...
+        // Go through all data entries and check if any array has numeric keys. If yes, it likely contains multiple data inside it
+        // which should be treated as if coming from separate sources, and thus moved to the root of $jsonLds array.
+        foreach ($jsonLds as $jsonLd) {
+            if (array_keys($jsonLd) === range(0, \count($jsonLd) - 1)) {
+                foreach ($jsonLd as $v) {
+                    $this->jsonLds[] = $v;
+                }
+            } else {
+                $this->jsonLds[] = $jsonLd;
+            }
```
m-vo:
```suggestion
            $this->jsonLds = array_merge(
                $this->jsonLds,
                ArrayUtil::isAssoc($jsonLd) ? [$jsonLd] : $jsonLd
            );
```

Unit tests work fine for me with `ArrayUtil::isAssoc()`. 
(Suggestion needs additional use statement.)

qzminski:
Just pushed another change that will allow to handle the data grouped under the `@graph` property:

```html
<script type="application/ld+json">
[
    {
        "@context": "https:\/\/contao.org\/",
        "@graph": [
            {
                "@type": "Page",
                "foobar": true
            }
        ]
    },
    {
        "@context": "https:\/\/contao.org\/",
        "@graph": [
            {
                "@type": "Page",
                "foobar": false
            },
            {
                "@type": "Article",
                "foobar": null
            }
        ]
    }
]
</script>
```

Toflar:
Missing `@graph` support currently breaks the search index in 4.12 :)

core-bundle/src/Search/Document.php
qzminski: Thanks, improved in 483d0b8.

Toflar (dismissed):

ausi (dismissed):

core-bundle/src/Search/Document.php
```diff
@@ -145,8 +146,23 @@ static function (Crawler $node) {
...
+        foreach (array_filter($jsonLds) as $jsonLd) {
+            // If array has numeric keys, it likely contains multiple data inside it which should be
+            // treated as if coming from separate sources, and thus moved to the root of an array.
+            $jsonLdItems = ArrayUtil::isAssoc($jsonLd) ? [$jsonLd] : $jsonLd;
+
+            // Parsed the grouped values under the @graph within the same context
+            foreach ($jsonLdItems as $jsonLdItem) {
+                if (isset($jsonLdItem['@graph']) && \is_array($jsonLdItem['@graph'])) {
+                    foreach ($jsonLdItem['@graph'] as $graph) {
+                        $this->jsonLds[] = array_merge(['@context' => $jsonLdItem['@context']], $graph);
```
m-vo: Can we be sure the `@context` key exists?

core-bundle/src/Search/Document.php
```diff
@@ -145,8 +146,23 @@ static function (Crawler $node) {
...
+        // Filter invalid (null) and parse all values
+        foreach (array_filter($jsonLds) as $jsonLd) {
+            // If array has numeric keys, it likely contains multiple data inside it which should be
+            // treated as if coming from separate sources, and thus moved to the root of an array.
+            $jsonLdItems = ArrayUtil::isAssoc($jsonLd) ? [$jsonLd] : $jsonLd;
+
+            // Parsed the grouped values under the @graph within the same context
+            foreach ($jsonLdItems as $jsonLdItem) {
+                if (isset($jsonLdItem['@graph']) && \is_array($jsonLdItem['@graph'])) {
+                    foreach ($jsonLdItem['@graph'] as $graph) {
```
m-vo:
```suggestion
                if (\is_array($graphs = $jsonLdItem['@graph'] ?? null)) {
                    foreach ($graphs as $graph) {
```

core-bundle/src/Search/Document.php
Toflar: No we can't, we need to treat it as optional.

```json
{
    "@context": "https://schema.org",
    "name": "foobar"
}
```

is the same as


```json
{
    "https://schema.org/name": "foobar"
}
```

We already handle it in `expandJsonLdContexts()` so this should be re-used here, you're right!

Toflar (changes requested):
See https://github.com/contao/contao/pull/3196#discussion_r671020559

Toflar:
So basically it's this:

- All of the same @context have to be merged. Whether that's because the property is named "https://schema.org/name" or there's a "name" property inside of "https://schema.org" @context, does not matter.
- @graph is just a way to specify multiple types without repeating @context
- when calling $document->extractJsonLdScripts()  with a schema as parameter, we make sure, the properties are not prefixed in the return value
- when calling  $document->extractJsonLdScripts() without schema parameter, we make sure, the properties are prefixed

So imho in order to merge correctly, you have to expand the context first.
Can you confirm @ausi and @m-vo ?

core-bundle/src/Search/Document.php
qzminski: I changed the logic here in 7ce0abc7 to not rely purely on `@context` but rather on anything that's there, so 

```
{
    "https://schema.org/name": "foobar",
    "@graph" => […],
}
```

is handled the same way as with `@context` thingy.

ausi (approved):

Toflar (approved):
Nevermind. Doublechecked with @ausi directly. LGTM!

fritzmg:
This btw. also fixes the following error, that is currently present in 4.12 and 4.x:

```
request.CRITICAL: Uncaught PHP Exception TypeError: "strncmp() expects parameter 2 to be string, int given" at vendor/contao/contao/core-bundle/src/Search/Document.php line 248
```

leofeyer:
Thank you @qzminski.

PR #3197, opened 2021-07-15
Allow defining entities alongside DCA definitions

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

If you've got an existing DCA (like `tl_page`) you should be able to create a `Page` entity as well. E.g. to define relations to other entities in your application. IMHO the entity definitions should be preferred and DCA fields added if the respective column does not yet exist.

I'll add tests if we're fine with this change.

/cc @bytehead

### DIFF
--- a/core-bundle/src/Doctrine/Schema/DcaSchemaProvider.php
+++ b/core-bundle/src/Doctrine/Schema/DcaSchemaProvider.php
@@ -62,7 +62,7 @@ class DcaSchemaProvider
         $config = $this->getSqlDefinitions();
 
         foreach ($config as $tableName => $definitions) {
-            $table = $schema->createTable($tableName);
+            $table = $schema->hasTable($tableName) ? $schema->getTable($tableName) : $schema->createTable($tableName);
 
             // Parse the table options first
             if (isset($definitions['TABLE_OPTIONS'])) {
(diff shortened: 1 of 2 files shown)


bytehead (dismissed):
Works perfectly! Apart from the missing tests LGTM :)

bytehead (dismissed):

fritzmg (dismissed):

bytehead (dismissed):

aschempp:
I think we intentionally did not support this case because it would only make sense for the "original" author of the DCA to create an entity. What happens in your case if two extensions create a page entity for the same DCA?

bytehead:
The first will win. Fields only get added if not existing. Not supporting this behaviour because of the fact that multiple extensions can extend the entity with the same fields, is not valid to me. _Last will win_ is the behaviour now with traditional DCA.

m-vo:
The primary use case for this is IMO to allow using relations and foreign key constraints from Contao tables (e.g. `tl_page`) to your custom entities inside your application. This is currently very hard to do (and very fragile).

bytehead:
> The primary use case for this is IMO to allow using relations and foreign key constraints from Contao tables (e.g. tl_page) to your custom entities inside your application. 

exactly.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

aschempp:
> The first will win. Fields only get added if not existing. 

What does that mean? Can I no longer override a DCA field if it has been defined before? Or do you mean you can't override a field definition in DCA if it has been defined in an entity? Intentionally or there's no way to support that?

Issue #3200, opened 2021-07-17
Twig bugs


ausi:
I noticed some minor bugs in the twig implementation:

~~`{% extends '@Contao/ce_text' %}` does not work `("Nested blocks are not allowed in child templates") in "@Contao_ContaoCoreBundle/ce_text.html5"`. I think this is probably related to `ce_text` itself extending the `block_searchable` template.~~ fixed in #3203

~~`debug:contao-twig` does not show templates from theme folders.~~ fixed in #3219

_The following are not really bugs and might be future scope:_

~~The follwing usages of Contao html5 templates do not work (and I don’t know if the can or should ☺️)~~ see #3213

```
{% extends ['@Contao/…', '@Contao/…'] %}
{% extends true ? '@Contao/…' : '@Contao/…' %}
{% use '@Contao/…' %}
```

~~We could add support for the `html_attr` escaper in the same way as we did for `html` by disabling double_encode.~~ done in #3204

/cc @m-vo

m-vo:
> {% extends '@Contao/ce_text' %} does not work ("Nested blocks are not allowed in child templates") in "@Contao_ContaoCoreBundle/ce_text.html5". I think this is probably related to ce_text itself extending the block_searchable template.

I noticed this as well. This is IMO a limitation of the Contao template engine, but let's see if we can improve this!

m-vo:
> debug:contao-twig does not show templates from theme folders.

This is actually expected. The theme templates aren't part of the template hierarchy (not tracked) because they can only be the first templates in the chain and we would not know which one to pick in case there are multiple themes (runtime dependent). That's also the reason why you cannot include/embed a theme template with `@Contao/my_template`. 

Maybe we can find a solution for both issues/limitations. :thinking:

ausi:
Maybe `debug:contao-twig` could get an argument to specify the theme folder like `--theme=my-theme/with-subfolder` that could make it more obvious to users of this command that themes are handled separately I think.

m-vo:
With #3203, #3204, #3213 and #3219 now the only thing missing from this ticket is support for `{% use '@Contao/…' %}`.

But as the Twig documentation states…

> Horizontal reuse is an advanced Twig feature that is hardly ever needed in regular templates. It is mainly used by projects that need to make template blocks reusable without using inheritance.

… I think it's fine to implement this at a later point in time.

ausi:
Closed in favor of #3203, #3204, #3213 and #3219

PR #3204, opened 2021-07-17
Implement a contao_ variant of the html_attr escaper

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | see #3200 
| Docs PR or issue | -

This introduces a new `contao_html_attr` strategy that replaces `html_attr`. 

Even if we do not want this change we should port the fix in `ContaoEscaperNodeVisitor#isEscaperFilterExpressionWithHtmlStrategy` (support `e` or `escape` as name).

/cc @ausi

### DIFF
--- a/core-bundle/src/Twig/Interop/ContaoEscaper.php
+++ b/core-bundle/src/Twig/Interop/ContaoEscaper.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Twig\Interop;
 
 use Contao\Controller;
-use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\StringUtil;
 use Twig\Environment;
 use Twig\Error\RuntimeError;
@@ -29,16 +28,6 @@ use Twig\Error\RuntimeError;
  */
 final class ContaoEscaper
 {
-    /**
-     * @var ContaoFramework
-     */
-    private $framework;
-
-    public function __construct(ContaoFramework $framework)
-    {
-        $this->framework = $framework;
-    }
-
     /**
      * This implementation is a clone of Twig's html escape strategy but calls
      * htmlspecialchars with the double_encode parameter set to false.
@@ -78,7 +67,7 @@ final class ContaoEscaper
         $string = (string) $string;
 
         // Replace insert tags before '{' and '}' get encoded
-        $string = $this->framework->getAdapter(Controller::class)->replaceInsertTags($string, false);
+        $string = Controller::replaceInsertTags($string, false);
         $string = StringUtil::decodeEntities($string);
 
         // Original logic
--- a/core-bundle/src/Twig/Interop/ContaoEscaper.php
+++ b/core-bundle/src/Twig/Interop/ContaoEscaper.php
@@ -19,8 +19,8 @@ use Twig\Environment;
 use Twig\Error\RuntimeError;
 
 /**
- * The ContaoEscaper mimics Twig's default escape filters but prevents
- * double encoding. It must therefore ONLY be applied to templates with already
+ * The ContaoEscaper mimics Twig's default escape filters but prevents double
+ * encoding. It must therefore ONLY be applied to templates with already
  * encoded context (input encoding)!
  *
  * This strategy will get dropped once we move to output encoding.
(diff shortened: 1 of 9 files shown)


core-bundle/tests/Twig/Interop/ContaoEscaperNodeVisitorTest.php
```diff
@@ -97,16 +99,47 @@ public function testUppercaseEntities(): void
...
     {
         $loader = new ArrayLoader([
             'modern.html.twig' => $templateContent,
             'legacy.html.twig' => $templateContent,
         ]);
 
+        $controller = $this->getMockBuilder(Adapter::class)
+            ->disableOriginalConstructor()
+            ->addMethods(['replaceInsertTags'])
+            ->setMockClassName(Controller::class)
```
m-vo: I have no idea why this is necessary. A normal mock works (but has the wrong type in `getAdapter`) and setting the mock class name works. If I don't do either, the adapter is `null` when multiple tests are run.

IMHO looks a bit like a PHPUnit caching issue.

core-bundle/src/Twig/Interop/ContaoEscaper.php
```diff
@@ -43,10 +60,85 @@ public function __invoke(Environment $environment, $string, ?string $charset): s
...
         if (null !== $charset && 'UTF-8' !== strtoupper($charset)) {
-            throw new RuntimeError(sprintf('The "contao_html" escape filter does not support the %s charset, use UTF-8 instead.', $charset));
+            throw new RuntimeError(sprintf('The "contao_html_attr" escape filter does not support the %s charset, use UTF-8 instead.', $charset));
         }
 
-        return htmlspecialchars($string, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8', false);
+        $string = (string) $string;
+
+        // Replace insert tags before '{' and '}' get encoded
+        $string = $this->framework->getAdapter(Controller::class)->replaceInsertTags($string);
```
ausi:
```suggestion
        $string = $this->framework->getAdapter(Controller::class)->replaceInsertTags($string, false);
```

`$blnCache` needs to be `false` I think, otherwise uncachable insert tags will not get replaced.

ausi (dismissed):

core-bundle/tests/Twig/Interop/ContaoEscaperNodeVisitorTest.php
leofeyer: This triggers the following warnings on my system:

```
There were 5 warnings:

1) Contao\CoreBundle\Tests\Twig\Interop\ContaoEscaperNodeVisitorTest::testEscapesEntities
Class "Contao\Controller" already exists.

/Users/leofeyer/Sites/contao/managed-edition/vendor/contao/contao/vendor/phpunit/phpunit/phpunit:76

2) Contao\CoreBundle\Tests\Twig\Interop\ContaoEscaperNodeVisitorTest::testDoesNotDoubleEncode
Class "Contao\Controller" already exists.

/Users/leofeyer/Sites/contao/managed-edition/vendor/contao/contao/vendor/phpunit/phpunit/phpunit:76

3) Contao\CoreBundle\Tests\Twig\Interop\ContaoEscaperNodeVisitorTest::testHandlesFiltersAndFunctions
Class "Contao\Controller" already exists.

/Users/leofeyer/Sites/contao/managed-edition/vendor/contao/contao/vendor/phpunit/phpunit/phpunit:76

4) Contao\CoreBundle\Tests\Twig\Interop\ContaoEscaperNodeVisitorTest::testUppercaseEntities
Class "Contao\Controller" already exists.

/Users/leofeyer/Sites/contao/managed-edition/vendor/contao/contao/vendor/phpunit/phpunit/phpunit:76

5) Contao\CoreBundle\Tests\Twig\Interop\ContaoEscaperNodeVisitorTest::testHtmlAttrFilter
Class "Contao\Controller" already exists.
```

m-vo: Ah, then this might be an autoloading issue. :thinking:

m-vo: This was a tough one. Fixed in 4dd61be7b.

Generating and executing Twig sources did not play well with PHPUnits reference handling. I rewrote the code to use old school `Controller::replaceInsertTags()` instead and the tests to mock and set a container with `System::getContainer` and real insert tag listeners.

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #3205, opened 2021-07-18
Use "user groups" instead of "groups" where applicable

Mynyx:
Adjustment of the explanation ``tl_settings.defaultGroup.1`` in respect of the title of the back end module "User groups".

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/tl_page.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/tl_page.xlf
@@ -234,7 +234,7 @@
         <source>Group</source>
       </trans-unit>
       <trans-unit id="tl_page.cgroup.1">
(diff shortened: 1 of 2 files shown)


fritzmg (changes requested):
If this really needs to be changed, then this needs to be changed in multiple places, not just this one. Further examples:

* `tl_article.groups.1`
* `tl_content.groups.1`
* `tl_module.groups.1`
* `tl_page.groups.1`
* `tl_page.cgroup.1`

Mynyx:
Thank you for the hint. I have also amended ``tl_page.cgroup.1`` now. 

However, the other places refer to member groups, which is already evident from the label there.

aschempp (approved):

leofeyer:
Thank you @Mynyx.

PR #3207, opened 2021-07-19
Relay statement parameters to doctrine dbal

ausi:
This rewires our legacy database statements to doctrine to use real prepared statements.

AFAIK, we only supported `?` for parameters and `%s` for `INSERT`/`UPDATE` sets, is this correct?


aschempp:
Didn't we have issues with the `set` method when we previouly tried that?

ausi:
> Didn't we have issues with the `set` method when we previouly tried that?

I could only find https://github.com/contao/core-bundle/pull/182 and https://github.com/contao/core-bundle/pull/31 and I didn’t see problems with `set()` there.

aschempp:
Whats with the condition in the `set()` method that is not supported (anymore) with this PR? Was `set()` previously supported for other statements or was the message just missing?

ausi:
> Was `set()` previously supported for other statements or was the message just missing?

See https://github.com/contao/contao/blob/722d397e3429e3fc4a8d05263fe2f834aa8f2122/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php#L167

core-bundle/tests/Contao/Database/StatementTest.php
```diff
@@ -0,0 +1,346 @@
...
+
+        $this->expectDeprecation('%sUsing "%s" with an array parameter has been deprecated%s');
+
+        $statement->execute($params ?? []);
+    }
+
+    /**
+     * @dataProvider getQueriesWithParametersAndSets
+     */
+    /*
```
leofeyer: Since this is commented out, I guess it can be removed?

leofeyer:
> AFAIK, we only supported `?` for parameters and `%s` for `INSERT`/`UPDATE` sets, is this correct?

Yes.

core-bundle/tests/Contao/Database/StatementTest.php
ausi: I’d wait for the reviews. This shows how the old implementation worked. Once approved we should remove it before the PR gets merged.

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -270,8 +263,44 @@ public function query($strQuery='')
...
+					!\is_string($varParam)
+					&& !\is_bool($varParam)
+					&& !\is_float($varParam)
+					&& !\is_int($varParam)
+					&& $varParam !== null
+				) {
+					return serialize($varParam);
+				}
+
+				return $varParam;
```
m-vo: Maybe better comprehensible when de-morganified? :slightly_smiling_face: 
```suggestion
				if (
					\is_string($varParam)
					|| \is_bool($varParam)
					|| \is_float($varParam)
					|| \is_int($varParam)
					|| $varParam === null
				) {
					return $varParam;
				}

				return serialize($varParam);
```

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3152, opened 2021-07-05
Fix the CI

leofeyer:

PR #3208, opened 2021-07-20
Always concatenate the 'image_container' class in figure.html.twig

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #-
| Docs PR or issue | -

We must explicitly concatenate the `image_container` class in the `figure.html.twig` template instead of just setting it. Because these are template options that always win over `Figure` options, you would otherwise not be able to set your own class via `FigureBuilder#setOptions(['attr' => ['class' => 'foo]]);`.

If you really need the class to be gone, you can simply overwrite the template with:

```twig
{% import "@ContaoCore/Image/Studio/_macros.html.twig" as studio %}

{{- studio.figure(figure) -}}
```

/cc @fritzmg

### DIFF
--- a/core-bundle/src/Resources/views/Image/Studio/figure.html.twig
+++ b/core-bundle/src/Resources/views/Image/Studio/figure.html.twig
@@ -1,3 +1,5 @@
 {% import "@ContaoCore/Image/Studio/_macros.html.twig" as studio %}
 
-{{- studio.figure(figure, { attr: { class: 'image_container' }}) -}}
+{{- studio.figure(figure, figure.options|default({})|merge({
+    attr: { class: ('image_container ' ~ figure.options.attr.class|default(''))|trim }
+})) -}}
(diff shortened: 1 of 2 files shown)


fritzmg (dismissed):
Looks like the most sensible option to me for the default `figure.html.twig` template 👍 (in order to be in line with the default `image.html5` template).

ausi (dismissed):

Toflar (dismissed):

leofeyer (approved):

m-vo:
Failing tests look unrelated. 

Didn't we fix those already? Maybe they are not merged upstream to 4.11, yet.

leofeyer:
Maybe we have only fixed it in the 4.x branch?

leofeyer:
Thank you @m-vo.

m-vo:
> Maybe we have only fixed it in the 4.x branch?

:+1: 

We likely need to backport https://github.com/contao/contao/pull/3152 + https://github.com/contao/contao/pull/3137.

PR #3210, opened 2021-07-22
Ensure the numberOfItems label does refer to items only

Toflar:
I'm reusing this field quite a lot in custom elements (and I see others doing the same) and I don't think it makes sense to default that to "images". It's still perfectly valid for the gallery element like this as well :)


fritzmg (approved):

bytehead (approved):

aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #3214, opened 2021-07-25
Support request tokens in Symfony forms

### LINKED ISSUE #3211: CSRF-Token is removed while using Symfony forms in a content element
**Affected version(s)**
Contao 4.9 (possible 4.11, not tested)

**Description**
While using Symfony forms in a Contao content element, the CSRF token gets stripped away from the response by https://github.com/contao/contao/blob/4.9/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php#L148. 

The `CsrfTokenManager` is injected via DI.

```php
// my content element
$form = $this->formFactory->create(MyFormType::class, $myForm, [
    'csrf_token_manager' => $this->csrfTokenManager,
]);
```

```php
// my form type
    // ...
	public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'data_class' => MyFormType::class,
            'translation_domain' => 'forms',
            'csrf_token_id' => 'contao_csrf_token',
        ]);
    }
    // ...
```

When rendering the form, the token value gets removed from the mentioned listener.

I don't want to disable the CSRF check on the route, as there possible other forms (e.g. Contao forms) on the same route/page.

/cc @ausi

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3211
| Docs PR or issue | contao/docs#795

This should make it possible to use request tokens in Symfony forms with the following configuration:

```php
$this->createFormBuilder(…, [
    'csrf_field_name' => 'REQUEST_TOKEN',
    'csrf_token_manager' => $manager, // @contao.csrf.token_manager
    'csrf_token_id' => $tokenName, // %contao.csrf_token_name%
]);
```

**UPDATE 2021-07-26:** Can now be used like this:

```php
$this->createFormBuilder($data, $this->getCsrfFormOptions());
```
or like this:
```php
$this->createFormBuilder($data, array_merge(
    $this->getCsrfFormOptions(),
    [ /* custom options */ ]
));
```

### DIFF
--- a/core-bundle/src/Controller/FrontendController.php
+++ b/core-bundle/src/Controller/FrontendController.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Controller;
 
 use Contao\CoreBundle\Cron\Cron;
+use Contao\CoreBundle\Csrf\ContaoCsrfTokenManager;
 use Contao\FrontendIndex;
 use Contao\FrontendShare;
 use Symfony\Component\HttpFoundation\RedirectResponse;
@@ -20,7 +21,6 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Core\Exception\LogoutException;
-use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 /**
  * @Route(defaults={"_scope" = "frontend", "_token_check" = true})
@@ -107,7 +107,7 @@ class FrontendController extends AbstractController
     public function requestTokenScriptAction(): Response
     {
         $token = $this
-            ->get('contao.csrf.token_manager')
+            ->get(ContaoCsrfTokenManager::class)
             ->getToken($this->getParameter('contao.csrf_token_name'))
             ->getValue()
         ;
@@ -115,7 +115,7 @@ class FrontendController extends AbstractController
         $token = json_encode($token);
 
         $response = new Response();
-        $response->setContent('document.querySelectorAll("input[name=REQUEST_TOKEN]").forEach(function(i){i.value='.$token.'})');
+        $response->setContent('document.querySelectorAll(\'input[name=REQUEST_TOKEN],input[name$="[REQUEST_TOKEN]"]\').forEach(function(i){i.value='.$token.'})');
         $response->headers->set('Content-Type', 'application/javascript; charset=UTF-8');
         $response->headers->addCacheControlDirective('no-store');
         $response->headers->addCacheControlDirective('must-revalidate');
@@ -127,7 +127,7 @@ class FrontendController extends AbstractController
     {
         $services = parent::getSubscribedServices();
 
-        $services['contao.csrf.token_manager'] = CsrfTokenManagerInterface::class;
+        $services[] = ContaoCsrfTokenManager::class;
 
         return $services;
     }
--- a/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
+++ b/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Csrf;
+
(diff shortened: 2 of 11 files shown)


m-vo:
Would it make sense to always adjust these defaults when using the `AbstractFragmentController`? Then we could maybe overwrite the method and set them (if not yet present in the options array).

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
```diff
@@ -0,0 +1,60 @@
...
+            && 'POST' === $request->getRealMethod()
+            && !$request->getUserInfo()
+            && (
+                0 === $request->cookies->count()
+                || [$this->csrfCookiePrefix.$token->getId()] === $request->cookies->keys()
+            )
+            && !($request->hasSession() && $request->getSession()->isStarted())
+        ) {
+            return true;
+        }
```
fritzmg: Should we reuse part of this logic in the `RequestTokenListener`?

ausi:
> Would it make sense to always adjust these defaults when using the `AbstractFragmentController`? Then we could maybe overwrite the method and set them (if not yet present in the options array).

That would be great I think, but AFAIK `createFormBuilder()` is only one of serveral ways to create forms in Symfony. If only one of these ways has the correct CSRF defaults this might be confusing to the developer?

bytehead:
For me it's fine to set the csrf-stuff on the form. 👍

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
ausi: Done in 49c2a74fd6971bef9acc63033ee23f32d9a9239b

fritzmg: Nice :). On that note though: the `RequestTokenListener` still checks for more stuff. I am wondering why the _Contao_ Csrf Token Manager does not check for the same things as the `RequestTokenListener` does?

ausi: The listener checks if it should interfere the current request while the token manager checks if it can skip the token validation.

If `isTokenValid()` is called in a non-POST request or in other contexts it should still validate correctly I think, whereas the listener should not interfere with the request in this case.

Toflar:
We could provide a `createContaoFormBuilder` as a shortcut in our AbstractFragmentController? That would cover the 90% use case and for the other ways to create a form, at least as a dev you'd have immediate docs/access to what you need to configure :)

ausi:
How about something like this?

```php
$this->createFormBuilder($data, array_merge(
	$this->getCsrfFormOptions(),
	[
		// Your own options…
	]
));
```

m-vo:
+1 for `$this->getCsrfFormOptions()` - then you do not need to care about injecting/subscribing to the Contao token manager service + token name parameter. And it would also work for `AbstractController#createForm()` as well as building a form from scratch with the form builder.

ausi:
> +1 for `$this->getCsrfFormOptions()`

Done in 094727d51cbdb0a96b1f317dad790bc0bf51745d

aschempp:
Can you explain what's the point in using request tokens in Symfony forms within a Contao request. I remember we had discussion about that before (like around Contao 4.0), but since the tokens are always checked on framework initialisation, it does not make sense to double-check them in forms?

bytehead:
> Can you explain what's the point in using request tokens in Symfony forms within a Contao request.

Just imagine a content element containing a Symfony form.

core-bundle/src/Controller/FrontendController.php
```diff
@@ -107,15 +107,15 @@ public function checkCookiesAction(): Response
...
+            ->get(ContaoCsrfTokenManager::class)
             ->getToken($this->getParameter('contao.csrf_token_name'))
             ->getValue()
         ;
 
         $token = json_encode($token);
 
         $response = new Response();
-        $response->setContent('document.querySelectorAll("input[name=REQUEST_TOKEN]").forEach(function(i){i.value='.$token.'})');
+        $response->setContent('document.querySelectorAll(\'input[name=REQUEST_TOKEN],input[name$="[REQUEST_TOKEN]"]\').forEach(function(i){i.value='.$token.'})');
```
leofeyer, line 118: Why this change?

ausi: Symfony by default creates named forms so instead of `name="REQUEST_TOKEN"` it would create something like `name="form[REQUEST_TOKEN]"` or `name="product[REQUEST_TOKEN]"`

leofeyer:
Thank you @ausi.

PR #3215, opened 2021-07-26
Make Contao compatible with Symfony 5.3

ausi:
As Symfony 5.2 is EOL now we should add compatibility for Symfony 5.3

Todo:

~~- [ ] Fix deprecations~~


ausi:
Regarding the deprecations: Fixing them would probably be a lot of work (if possible at all) because we still support Symfony 4.4 in Contao 4.12 and being compatible with 4.4, 5.4 and 6.0 is overkill IMO.

I think we should keep the Symfony deprecations as they are now in the 4.12 branch (as 4.12 will never be compatible with Sf 6.0 anyway) and move to Symfony 5.4/6.0 in Contao 4.13. This way fixing the deprecations will be way easier as all the 6.0 methods and classes are already available in version 5.4.

m-vo (dismissed):

Toflar (dismissed):

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
```diff
@@ -63,7 +63,7 @@ protected function attemptAuthentication(Request $request): ?TokenInterface
         $username = $request->request->get('username');
         $password = $request->request->get('password');
 
-        if (!\is_string($username) && (!\is_object($username) || !method_exists($username, '__toString'))) {
```
aschempp, line 66: Are you sure this can be changed since we still support Symfony 4.4?

core-bundle/tests/Security/Authentication/AuthenticationFailureHandlerTest.php
```diff
@@ -130,12 +131,17 @@ private function getRequest(): Request
...
-        $request->request
-            ->expects($this->never())
-            ->method('get')
-            ->with('username')
-        ;
+        $params = [
+            'username' => new class() implements \Stringable {
+                public function __toString(): string
+                {
+                    AuthenticationFailureHandlerTest::fail('Unexpected call to get("username")');
```
aschempp: This will not lead to an assertion that the test was successful if the method is never called.

core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
```diff
@@ -292,7 +292,7 @@ public function testHandlesContaoUsers(): void
 
     public function testBeginsTwoFactorAuthenticationForContaoUsers(): void
     {
-        $user = $this->createPartialMock(BackendUser::class, ['save']);
+        $user = $this->createPartialMock(BackendUser::class, ['save', 'getUserIdentifier']);
```
aschempp, line 295: can you explain why we need to mock the method to return null instead of a property?

core-bundle/src/Resources/contao/library/Contao/User.php
```diff
@@ -509,6 +509,14 @@ public function setUsername($username)
 		return $this;
 	}
 
+	/**
+	 * {@inheritdoc}
+	 */
+	public function getUserIdentifier(): string
+	{
+		return $this->username;
```
aschempp, line 527:
```suggestion
		return (string) $this;
```

The `toString` method handles empty username cases. We should either use that method or copy it here and use `getUserIdentifier` in `__toString()`.

calendar-bundle/composer.json
```diff
@@ -19,12 +19,12 @@
         "contao/core-bundle": "self.version",
         "friendsofsymfony/http-cache": "^2.4",
         "patchwork/utf8": "^1.2",
-        "symfony/config": "4.4.* || 5.2.*",
-        "symfony/dependency-injection": "4.4.* || 5.2.*",
+        "symfony/config": "4.4.* || 5.3.*",
```
aschempp:
```suggestion
        "symfony/config": "4.4.* || 5.2.* || 5.3.*",
```

There is no reason to drop Symfony 5.2 support afaik? Also see https://github.com/contao/contao/blob/4.8/composer.json#L79

aschempp (changes requested):

calendar-bundle/composer.json
ausi: I don’t see a reason to support Symfony 5.2 as it is EOL. Why don’t we support 5.0 and 5.1?

core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
ausi: Otherwise the tests fail.

core-bundle/tests/Security/Authentication/AuthenticationFailureHandlerTest.php
ausi: If the method was never called the test is successful. What do you want to change?

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
ausi: Yes.

core-bundle/src/Resources/contao/library/Contao/User.php
ausi: Fixed in 510ca63882b794d4077ec46697608bd4171ed21e

calendar-bundle/composer.json
fritzmg: Symfony 5 support was introduced with Contao `4.11.0`, starting with Symfony `5.2.*`, since that was the active Symfony version at the time. `5.2.*` should definitely stay supported, otherwise a downgrade path of Symfony to 4.4 is necessary in some circumstances where a dependency still only requires `5.2.*`.

ausi: But this pull request is for Contao `4.12.0` can’t we move up the dependecy to `5.3.*` the same way we moved it up from `4.3.*` to `4.4.*` in Contao `4.9`?

m-vo: Running Symfony 5.2 isn't a viable option, same as Contao 4.11 (as soon as 4.12 is released), is it? All extensions blocking the update have the same problem IMHO.

I think we should consider using open requirements (`^` instead of `.*`) again in the future, though.

ausi: I think we made the decision to use `X.X.*` before we had the `contao/conflicts` package. With `contao/conflicts` we can IMO move back to using `^X.X` for the Symfony packages because if incompatibilities occur with a new Symfony version we can quickly add conflicts for the affected packages.

@contao/developers WDYT?

I would update this PR then to use `^5.2`.

fritzmg: > But this pull request is for Contao `4.12.0` can’t we move up the dependecy to `5.3.*` the same way we moved it up from `4.3.*` to `4.4.*` in Contao `4.9`?

It's different there as `4.4.*` is the minimum requirement - and still is in Contao `4.12.0` and `4.13.0`. Only an upgrade path would be allowed, not a downgrade path. However a downgrade because of this is probably more an edge case as I don't expect many other dependencies to also limit the Symfony version to a bugfix version.

ausi: > I would update this PR then to use `^5.2`.

Done in cfafcf4995665f400cb5c8c4428ff4823ba4ed46

bytehead (dismissed):

m-vo (dismissed):

core-bundle/tests/Security/Authentication/AuthenticationFailureHandlerTest.php
aschempp: The test does not fail, but it does not assert a successful test. The test suite does not know whether you actually tested something successfully.

PS: you shouldn't resolve a conversation if you ask me a question 😆

m-vo: > The test does not fail, but it does not assert a successful test

There are still 3 assertions in this test. Besides that triggering `self::fail()` is equivalent to`self::never()` in a mocked method AFAIK. 

There is an edge case where you do neither have any positive assertions, nor any uses of `never()` - there PHPUnit cannot interpret the call to `fail()` and would (falsely) complain that the test doesn't make assertions. But that's a limitation of PHPUnit and IMHO not the case here.

ausi: > PS: you shouldn't resolve a conversation if you ask me a question 😆

I probably sensed that you will not answer it anyways 😜

> The test does not fail, but it does not assert a successful test.

Unable to reproduce 
Bildschirmfoto 2021-07-28 um 13 01 09

aschempp: yeah, those are three other assertions. There is no assertion that username is never retrieved. I think there must be another way to write a better test 🤷

ausi: > I think there must be another way to write a better test

Feel free to create a PR to replace all `fail()` calls with something else, but I don’t see how this is a bad test.

aschempp: A simple solution in general would be to check a variable outside the callback: https://stackoverflow.com/questions/9296529/phpunit-how-to-test-if-callback-gets-called

However, I checked the code, and there seems to be absolutely not reason for this check at all (most likely became obsolete). The request is never queried for any property, and the username is not relevant at all. The `logException` method does not even use `$request`, so I'd suggest to simply remove the method argument and the test 🙃

ausi: Still unrelated to this PR IMO. Anyway, removed in 544ddfd5a712b802fba996d4c952f861b4ead024

m-vo (dismissed):

Toflar (dismissed):

aschempp:
I wonder if the change to use carrets is correct. Remember, if we already had carrets, Contao 4.11 would now receive Symfony 5.3 and the fixes that are in this PR would not be in 4.11 (especially not in few weeks, which might break 4.11 completely). Be aware the the conflicts package currently also only works for all Contao versions, not individual minors (though that might be possible but a pita).

ausi:
> Remember, if we already had carrets, Contao 4.11 would now receive Symfony 5.3 and the fixes that are in this PR would not be in 4.11

There are no fixes in this PR. Contao 4.11 would be 5.3 compatible already a long time ago if we already used carrets.

> Be aware the the conflicts package currently also only works for all Contao versions

We can split the conflicts package for every contao version, we especially designed it that way (using `*@dev`) AFAIR.

calendar-bundle/composer.json
```diff
@@ -34,13 +34,13 @@
...
         "phpunit/phpunit": "^8.5",
-        "symfony/http-client": "4.4.* || 5.2.*",
+        "symfony/http-client": "^4.4 || ^5.2",
         "symfony/phpunit-bridge": "^5.2"
     },
     "extra": {
         "contao-manager-plugin": "Contao\\CalendarBundle\\ContaoManager\\Plugin",
         "symfony": {
-            "require": "4.4.* || ^5.2"
+            "require": "^4.4 || ^5.2"
```
leofeyer, line 43: I do not like these changes, either. We agreed to pin the Symfony minor versions (e.g. `4.4.* || 5.3.*`), because in the past, there were too many (partly unintended) breaking changes. Using `^5.2` not only implies that we have to continue to support Symfony 5.2, it also means that our users will get Symfony 5.4 as soon as it has been released.

leofeyer (changes requested):

calendar-bundle/composer.json
ausi: > it also means that our users will get Symfony 5.4 as soon as it has been released.

I think we made the decision to use `X.X.*` before we had the `contao/conflicts` package. So this should not be an issue anymore. We can very quickly add a conflict for the component that causes issues and have enough time to fix it.

bytehead: > it also means that our users will get Symfony 5.4 as soon as it has been released.

What's wrong with this? In the case, 5.4 does not work as expected, I can easily adjust the constraint.

leofeyer: I am fine with changing our policy here. @contao/developers Any objections?

aschempp: As described above, I'm unsure about this. If Contao 4.10 hat `^5.1` set, it would get Sf 5.4 as well and it wouldn't surprise me if that wouldn't work. Yes, you're using an outdated version, but that doesn't mean it should break on an update. Yes, we _could_ manage that through conflicts, but we don't really want to manage conflicts for any previous Contao version? Otherwise we'll need to add conflicts for Contao 4.10 in 4 years when Symfony 5.4 adds an unintended break?

m-vo: > Otherwise we'll need to add conflicts for Contao 4.10 in 4 years when Symfony 5.4 adds an unintended break?

I don't think we ever added conflicts for EOL versions, did we?

bytehead: I personally would never update EOL packages with the expectation that everything will be fine 🙃

fritzmg: Yeah, Contao 4.10 will be EOL at that point and the only way to fix it (as with any problem, no matter if it's direct or indirect) is to update to a currently supported version. There are currently also numerous other problems in 4.5, 4.6, 4.7, 4.8 for example that occur through varies updated dependencies. (These issues come up in the community because sometimes people try to update their Contao 4.4 instance to 4.5, 4.6, 4.7, 4.8 first instead of going directly to 4.9 because they think that's what you are supposed to do.)

leofeyer (approved):

Toflar (approved):

leofeyer:
Thank you @ausi.

m-vo:
Deprecation handling will need some love now. :upside_down_face: 
These things are output all over the place when using the console with 5.3 components:

```
[info] User Deprecated: Since symfony/framework-bundle 5.3: The "session.storage.factory.service" service is deprecated, use "session.storage.factory.native", "session.storage.factory.php_bridge" or "session.storage.factory.mock_file" instead.
[info] User Deprecated: Since symfony/framework-bundle 5.3: The "session.storage.native" service is deprecated, use "session.storage.factory.native" instead.
[info] User Deprecated: Since symfony/framework-bundle 5.3: The "session.storage.metadata_bag" service is deprecated, create your own "session.storage.factory" instead.
[info] User Deprecated: Since symfony/http-foundation 5.3: "Symfony\Component\HttpFoundation\RequestStack::getMasterRequest()" is deprecated, use "getMainRequest()" instead.
[info] User Deprecated: Since symfony/cache 5.4: "Symfony\Component\Cache\DoctrineProvider" is deprecated, use "Doctrine\Common\Cache\Psr6\DoctrineProvider" instead.
[info] User Deprecated: Since symfony/cache 5.4: "Symfony\Component\Cache\DoctrineProvider" is deprecated, use "Doctrine\Common\Cache\Psr6\DoctrineProvider" instead.
[info] User Deprecated: Since symfony/security-bundle 5.3: The "security.encoder_factory" service is deprecated, use "security.password_hasher_factory" instead.
[info] User Deprecated: Since symfony/security-core 5.3: The "Symfony\Component\Security\Core\Encoder\EncoderFactory" class is deprecated, use "Symfony\Component\PasswordHasher\Hasher\PasswordHasherFactory" instead.
[info] User Deprecated: Since symfony/security-core 5.3: The "Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface" class is deprecated, use "Symfony\Component\PasswordHasher\Hasher\PasswordHasherFactoryInterface" instead.
```

bytehead:
@m-vo which error level? I don't get these deprecations on the console with 5.3 components and error level 8191.

m-vo:
> @m-vo which error level? 

I didn't adjust the error level. :thinking: 
The … default one? :smile: 

Try running a command that does not exist or is ambiguous, like `contao-console contao:` or `contao-console -vvv`. This is where most errors get reported for me as more commands are instantiated.

PR #3217, opened 2021-07-27
Allow robots setting for redirect pages

fritzmg:
Currently the `robots` field is only available in the `regular` and `error_*` palettes. The original reasoning behind this is that these are the only pages that actually produce HTML output where the robots meta tag can be shown.

However, the `robots` setting does not just control the robots meta tag anymore. It also controls the following things:

* Whether or not the page is included in the `sitemap.xml`.
* Whether to add `nofollow` to the `rel` attribute in the navigation (see [here](https://github.com/contao/contao/blob/f0a6afc23de389f6342af6a67db0ebc90122ea89/core-bundle/src/Resources/contao/modules/Module.php#L438-L441)).

These things currently cannot be controlled for redirect pages.

This also leads to unwanted behavior. If you first create a `regular` page, configure its `robots` setting and then change the page type to `internal` or `external`, the `robots` setting will still be in effect. E.g. if you set it to `noindex,nofollow` the `rel="nofollow"` attribute will be added to the link of the page in the navigation menu, but you cannot change it anymore (without first changing the page type and then back again).

Discussion points:

* Should we introduce separate settings for these things after all?
* Should the same be applied to the `logout` page?


aschempp (approved):
I would not add a separate setting, this seems the best option to me.

ausi (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #3218, opened 2021-07-27
Do not start the session in the login module

ausi:
Placing a login module on a page currently starts the session and therefore prevents caching.

This pull request should fix that. (Successfully tested in a 4.11.5 setup).

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleLogin.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleLogin.php
@@ -93,9 +93,20 @@ class ModuleLogin extends Module
 		global $objPage;
 
 		$container = System::getContainer();
+		$request = $container->get('request_stack')->getCurrentRequest();
 
 		/** @var AuthenticationException|null $exception */
-		$exception = $container->get('security.authentication_utils')->getLastAuthenticationError();
+		$exception = null;
+		$lastUsername = '';
+
+		// Only call the authentication utils if there is an active session to prevent starting an empty session
+		if ($request && $request->hasSession() && ($request->hasPreviousSession() || $request->getSession()->isStarted()))
+		{
+			$authUtils = $container->get('security.authentication_utils');
+			$exception = $authUtils->getLastAuthenticationError();
+			$lastUsername = $authUtils->getLastUsername();
+		}
+
 		$authorizationChecker = $container->get('security.authorization_checker');
 
 		if ($authorizationChecker->isGranted('ROLE_MEMBER'))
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

fritzmg:
Hm, what starts the session in `AuthenticationUtils::getLastAuthenticationError`/`getLastUsername` in the first place? Seems unintended.

ausi:
I would guess it is this line that starts the session by calling `->has(…)` https://github.com/symfony/symfony/blob/20cd469e9fab51737abed5dccb22cade9318fa86/src/Symfony/Component/Security/Http/Authentication/AuthenticationUtils.php#L43

aschempp:
Looks like something that should be fixed in Symfony (as well) ?

fritzmg:
Yes, may be Symfony needs to add `->isStarted()` too? Or may be it is expected that a request should not have a session at all when not needed - and thus we erroneously set the session some where? (edit: no, the session will always be set on the request by Symfony via `session_listener` (`Symfony\Component\HttpKernel\EventListener\SessionListener`))

ausi:
> Yes, may be Symfony needs to add `->isStarted()` too?

I’m not sure. Symfony does not send a session cookie for empty sessions so maybe it doesn’t create a problem there?

We could also try to check if the session is empty everywhere we check for `isStarted()` now. But I’m not sure if this is easily possible as `$session->isEmpty()` is marked as `@internal`.

aschempp:
> Symfony does not send a session cookie for empty sessions so maybe it doesn’t create a problem there?

do we send a session cookie on empty session?

ausi:
> > Symfony does not send a session cookie for empty sessions so maybe it doesn’t create a problem there?
> 
> do we send a session cookie on empty session?

No, Symfony handles that.

aschempp:
Can you please be a bit more specific? If I understand correctly, the problem is that our `MakeResponsePrivateListener` makes the response private, because `Session::isStarted()` returns true. Even though the session is empty and the cookie is unset by Symfony?

ausi:
> If I understand correctly, the problem is that our `MakeResponsePrivateListener` makes the response private, because `Session::isStarted()` returns true. Even though the session is empty and the cookie is unset by Symfony?

Correct.

leofeyer (approved):

leofeyer:
Thank you @ausi.

aschempp:
> > If I understand correctly, the problem is that our `MakeResponsePrivateListener` makes the response private, because `Session::isStarted()` returns true. Even though the session is empty and the cookie is unset by Symfony?
> 
> Correct.

So the behaviour of `MakeResponsePrivateListener` is wrong and should be fixed instead of this?

leofeyer:
@Toflar /cc

ausi:
> > > If I understand correctly, the problem is that our `MakeResponsePrivateListener` makes the response private, because `Session::isStarted()` returns true. Even though the session is empty and the cookie is unset by Symfony?
> > 
> > 
> > Correct.
> 
> So the behaviour of `MakeResponsePrivateListener` is wrong and should be fixed instead of this?

See https://github.com/contao/contao/pull/3218#issuecomment-887818448

> We could also try to check if the session is empty everywhere we check for `isStarted()` now. But I’m not sure if this is easily possible as `$session->isEmpty()` is marked as `@internal`.

Toflar:
> @Toflar /cc

I can't really follow. A failing unit test for the `MakeResponsePrivateListener` would be good for illustration purposes.

PR #3219, opened 2021-07-28
Track theme templates in the template hierarchy

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | See #3200 
| Docs PR or issue | -

This is a rather naive approach that does the following:

- also track templates from theme folders (we didn't do that before: theme templates are always the first ones, never a parent)
- add `$themeAlias` option to the template hierarchy methods
- when the hierarchy is requested, filter out any `Contao_Theme_*` namespaced template that does not match the alias

This way the template hierarchy can easily be queried in the command (see new option). Downside is the filtering and the additional paths that are tracked.

- [x] Decide if we want to do this.
- [x] Wait for other Twig related PRs are merged + rebase/refactor.
- [x] Add some tests.

/cc @ausi


ausi:
> Downside is the filtering and the additional paths that are tracked.

I think this performance downside is negligible.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #1411, opened 2020-02-26
Skip orphan pages in the route provider

aschempp:
fixes #1408

PR #3220, opened 2021-07-28
Correctly clean up left-over records in DCA mode 5

### LINKED ISSUE #1408: Endless 404 page loop due to orphan pages
**Affected version(s)**
4.8.8, probably 4.9 as well

**Description**
The `contao.routing.page_404_router` service introduced in 16355846ae8029800aaacc5382cb1876fb4e811c can cause an infinite 404 page loop when rendering any frontend page in `prod` environment. This happens only if there are some orphan `tl_page` records (i.e. subpages without a page root)  in the database:

```sql
SELECT * FROM tl_page WHERE pid > 0 AND pid NOT IN (SELECT id FROM tl_page)
```

The problem is that Contao generates routes for these pages (https://github.com/contao/contao/blob/4.8/core-bundle/src/Routing/RouteProvider.php#L249) and if they do not have an associated website root, they will simply have no `_locale` requirement that is one of the requirements in `Route` class. This also causes the `vendor/bin/contao-console debug:router` command to throw an `\InvalidArgumentException: Routing requirement for "_locale" must be a string.`.

It is highly unlikely that such records will appear in the database but is certainly possible (not sure how to reproduce though). The `DC_Table::reviseTable()` does not check against such records and perhaps it should.

**How to reproduce**
Have orphan pages non-related to any website root.

**Proposals**
1. Make `DC_Table::reviseTable()` wipe out the orphans described above.
2. Ensure the `_locale` route requirement is present before adding a page route to collection.

/cc @Toflar

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1408, #1411

Since we yet-again had issues the router generating `NoRootPageFoundException`s, we looked into why there are orphan pages in the page tree at all. Looks like a DCA in tree mode (5) was never correctly cleaned up!

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3301,6 +3301,11 @@ class DC_Table extends DataContainer implements \listable, \editable
 		$ptable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ptable'];
 		$ctable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ctable'];
 
+		if ($ptable === null && $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 5)
+		{
+			$ptable = $this->strTable;
+		}
+
 		/** @var AttributeBagInterface $objSessionBag */
 		$objSessionBag = System::getContainer()->get('session')->getBag('contao_backend');
 
@@ -3393,6 +3398,10 @@ class DC_Table extends DataContainer implements \listable, \editable
 			if ($GLOBALS['TL_DCA'][$this->strTable]['config']['dynamicPtable'])
 			{
 				$objIds = $this->Database->execute("SELECT c.id FROM " . $this->strTable . " c LEFT JOIN " . $ptable . " p ON c.pid=p.id WHERE c.ptable='" . $ptable . "' AND p.id IS NULL");
+			}
+			elseif ($ptable == $this->strTable)
+			{
+				$objIds = $this->Database->execute('SELECT c.id FROM '.$this->strTable.' c LEFT JOIN '.$ptable.' p ON c.pid=p.id WHERE p.id IS NULL AND c.pid > 0');
 			}
 			else
 			{
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3301,7 +3301,7 @@ class DC_Table extends DataContainer implements \listable, \editable
 		$ptable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ptable'];
 		$ctable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ctable'];
 
-		if ($ptable === null && $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 5)
+		if ($ptable === null && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) == 5)
 		{
 			$ptable = $this->strTable;
 		}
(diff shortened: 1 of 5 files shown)


aschempp:
I also decided to remove the exception handling in the routing system, since this case should not longer happen. If it does happen, we want it to throw exceptions to further investigate the issue.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3301,6 +3301,11 @@ protected function reviseTable()
 		$ptable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ptable'];
 		$ctable = $GLOBALS['TL_DCA'][$this->strTable]['config']['ctable'];
 
+		if ($ptable === null && $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 5)
```
```suggestion
		if ($ptable === null && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) == 5)
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3394,6 +3399,10 @@ protected function reviseTable()
 			{
 				$objIds = $this->Database->execute("SELECT c.id FROM " . $this->strTable . " c LEFT JOIN " . $ptable . " p ON c.pid=p.id WHERE c.ptable='" . $ptable . "' AND p.id IS NULL");
 			}
+			elseif ($ptable == $this->strTable)
+			{
+				$objIds = $this->Database->execute('SELECT c.id FROM ' . $this->strTable . ' c LEFT JOIN ' . $ptable . ' p ON c.pid=p.id WHERE p.id IS NULL AND c.pid > 0');
+			}
 			else
 			{
 				$objIds = $this->Database->execute("SELECT c.id FROM " . $this->strTable . " c LEFT JOIN " . $ptable . " p ON c.pid=p.id WHERE p.id IS NULL");
```
ausi, line 3408: Can’t we always check for `c.pid > 0` regardless of the mode?
```suggestion
			else
			{
				$objIds = $this->Database->execute('SELECT c.id FROM ' . $this->strTable . ' c LEFT JOIN ' . $ptable . ' p ON c.pid=p.id WHERE p.id IS NULL AND c.pid > 0');
```

aschempp: no, if a `tl_content` record has pid=0 it should be deleted

ausi: Why? The core never creates `tl_content` rows with `pid=0`, or does it?

aschempp: Probably not, but incorrectly imported data for example could lead to these invalid records. `reviseTable` should clean up everything that is incorrect, not just our core cases.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3394,6 +3399,10 @@ protected function reviseTable()
 			{
 				$objIds = $this->Database->execute("SELECT c.id FROM " . $this->strTable . " c LEFT JOIN " . $ptable . " p ON c.pid=p.id WHERE c.ptable='" . $ptable . "' AND p.id IS NULL");
 			}
+			elseif ($ptable == $this->strTable)
+			{
+				$objIds = $this->Database->execute('SELECT c.id FROM ' . $this->strTable . ' c LEFT JOIN ' . $ptable . ' p ON c.pid=p.id WHERE p.id IS NULL AND c.pid > 0');
```
leofeyer, line 3404: I do not understand the `c.pid > 0` part. How does it solve the issue?

ausi: The root pages have `c.pid=0` and therefore no parent node, but `reviseTable()` should not delete the root pages but only pages with a set `pid` that points to a non-existing parent.

leofeyer (approved):

PR #3227, opened 2021-07-29
Use ClockMock for time sensitive test

### LINKED ISSUE #3226: Use ClockMock in AuthenticationProviderTest
For reference, see this failing CI job: https://github.com/contao/contao/pull/3219/checks?check_run_id=3191509596

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3226
| Docs PR or issue | n/a

Use ClockMock for this time sensitive test.

### DIFF
--- a/core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
+++ b/core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
@@ -121,7 +121,7 @@ class AuthenticationProviderTest extends TestCase
         $provider = $this->createTwoFactorProvider($authProvider, $userChecker);
         $hasException = false;
 
-        ClockMock::withClockMock(1142164800);
+        ClockMock::withClockMock(true);
 
         try {
             $provider->authenticate($token);
--- a/core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
+++ b/core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
@@ -26,6 +26,7 @@ use Scheb\TwoFactorBundle\Security\TwoFactor\AuthenticationContext;
 use Scheb\TwoFactorBundle\Security\TwoFactor\AuthenticationContextFactoryInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Handler\AuthenticationHandlerInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Trusted\TrustedDeviceManagerInterface;
+use Symfony\Bridge\PhpUnit\ClockMock;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Security\Core\Authentication\Provider\AuthenticationProviderInterface;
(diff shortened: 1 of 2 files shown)


core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
```diff
@@ -120,6 +121,8 @@ public function testLocksUserOnInvalidTwoFactorCode(int $initialAttempts, int $l
         $provider = $this->createTwoFactorProvider($authProvider, $userChecker);
         $hasException = false;
 
+        ClockMock::withClockMock(true);
```
m-vo, line 124: I think this also needs to be registered - either by calling `ClockMock::register()` or by adding the `@time-sensitive` annotation on the method/in the `phpunit.xml`. You test can with `sleep(1000);` for instance - this should pass instantly.

m-vo (changes requested):

core-bundle/tests/Security/Authentication/Provider/AuthenticationProviderTest.php
m-vo: Just noticed, that we have already got some classes in the `phpunit.xml`:

https://github.com/contao/contao/blob/2ca2e45476e5a98aba8644093551a6576fa37261/phpunit.xml.dist#L63-L70

bytehead: This are whole directories. I'd rather register the `ClockMock` in the test, what do you think?

m-vo: Sounds sensible to me!

I'm personally in favor of the method annotation, but not sure what @leofeyer thinks about this?

bytehead: annotation is also fine for me :)

ausi: It’s not about directories but about namespaces, ClockMock works by creating a function in the registered namespaces so that the classes in these namespaces use the moked functions instead of the native php functions.

We have to register the namespaces in the phpunit xml file because registering it in the tests can be too late (if the class was already loaded by a different test).

bytehead: Thanks for the clarification. I'll adjust it.

bytehead: Added to the list in b5486f939.

phpunit.xml.dist
```diff
@@ -63,6 +63,8 @@
                             <element key="1"><string>Contao\CoreBundle\Tests\Command</string></element>
                             <element key="2"><string>Contao\CoreBundle\Cron\Command</string></element>
                             <element key="3"><string>Contao\CoreBundle\Tests\Cron</string></element>
+                            <element key="4"><string>Contao\CoreBundle\Security\Authentication\Provider</string></element>
+                            <element key="4"><string>Contao\CoreBundle\Tests\Security\Authentication\Provider</string></element>
```
ausi:
```suggestion
                            <element key="5"><string>Contao\CoreBundle\Tests\Security\Authentication\Provider</string></element>
```

bytehead: ups 🙈

ausi (approved):

m-vo (approved):
:heart:

leofeyer:
Thank you @bytehead.

PR #3228, opened 2021-07-29
Require twig/extra-bundle in the managed edition

m-vo:
There are several Twig "core" extensions for various tasks like string manipulation (`string-extra`), converting markdown (`markdown-extra`), email template stuff (`inky-extra`) and so on… See the [Twig monorepo](https://github.com/twigphp/Twig/tree/3.x/extra) for all of them.

These are all config-less extensions but you need to load them in your kernel (and have auto configuration enabled). Lucky for us there is already a solution that doesn't require this step: With `twig/extra-bundle` installed you can simply require extra packages and they are made available. No Contao Manager plugin or config needed. IMO this would make using these extensions much easier for our users.

So I think we should ship this bundle by default. As a bonus, this also suggests installing the relevant extensions in case you are using the Twig filters/functions but don't have them installed (see `MissingExtensionSuggestorPass`).


ausi (approved):

aschempp:
Unfortunately, the Twig extensions won't be installable through the Contao Manager. An alternative approach would be to have a (maybe official?) bundle that has suggestions what could be installed through the Manager. Not sure what that bundle would contain though, and I guess requiring this by default does not add any/much overhead…?

m-vo:
> Unfortunately, the Twig extensions won't be installable through the Contao Manager. 

Does the manager forbid requiring anything that is not in the search index/not suggested by any package or is there no way to enter a custom requirement? 

If there is no way to enter a package without searching it, would this mean I have to search for the `contao/suggestions` bundle and can then install suggestions or do I need to install the bundle first?

m-vo:
@ausi and me had a brief discussion about this in slack. What about this:

- We add a `twig-suggestions-bundle`, that can be found in the Contao Manager and that suggests all 'twig-extra' bundles.
- We merge this PR as is.
- Bonus: The manager could allow adding suggested packages without the need to install the package suggesting it.

manager-bundle/composer.json
```diff
@@ -55,6 +55,7 @@
         "symfony/web-profiler-bundle": "^4.4 || ^5.2",
         "symfony/yaml": "^4.4 || ^5.2",
         "toflar/psr6-symfony-http-cache-store": "^2.1 || ^3.0",
+        "twig/extra-bundle": "^2.11.3",
```
leofeyer, line 58: Isn't it a little odd that the core bundle requires `twig/twig` but the manager bundle requires `twig/extra-bundle`?

m-vo: My reasoning was: in a non-managed edition you would install `twig/extra-bundle` and register it in the kernel yourself because you need it for your app. Twig however is needed by the core as well, so it must be required there.

leofeyer:
Thank you @m-vo.

aschempp:
> Does the manager forbid requiring anything that is not in the search index/not suggested by any package or is there no way to enter a custom requirement?

You can only install what is found in the search results.

> If there is no way to enter a package without searching it, would this mean I have to search for the `contao/suggestions` bundle and can then install suggestions or do I need to install the bundle first?

You can only install suggestions of an installed package. Suggestions are here to add features to the mentioned package, not provide a means to find packages.

> * We add a `twig-suggestions-bundle`, that can be found in the Contao Manager and that suggests all 'twig-extra' bundles.
> * We merge this PR as is.
> * Bonus: The manager could allow adding suggested packages without the need to install the package suggesting it.

That will not be the case. It is meant to be exactly the opposite: If you want twig extras, install the contao-twig-extras bundle, then you can install suggestions from that package. So either
 1) we should revert this commit and create the extra package. That would mean we don't add "overhead" if people dont use any extra package
 2) we add the suggestions to contao/manager-bundle, but that seems pretty wrong to me, since we could add a million suggestions that way …

m-vo:
The idea of the `twig-extra` bundle is that it is very lightweight and it gives you a very nice DX: If you for instance look up something in the Twig docs and put it in your template which is a feature of an not yet installed "extra bundle", you'll get a friendly error message:

```twig
{% apply markdown_to_html %}
# Foo
## Bar
{% endapply %}
```

```text
The "markdown_to_html" filter is part of the MarkdownExtension, which is not installed/enabled; try running "composer require twig/markdown-extra" […].
```

Then you run `composer require twig/markdown-extra` like stated in the error message and you're good to go. 

This PR enables users to do this which is a good thing IMO (independent of the manager). But I'd like to make it as simple as possible for manager users as well.

aschempp:
That's a nice **D**eveloper E**x**perience, but our template editors are not developers. Can we somehow influence that message? How should we solve that in the Manager?

ausi:
> but our template editors are not developers

Not all template editors are developers, but there is surely a great amout of developers who also edit templates therefore the message should stay this way IMO.

> How should we solve that in the Manager?

How about an empty `contao/twig-extra-bundle` that a user can install and afterwards install one of it’s suggestions?

m-vo:
> Can we somehow influence that message? 

We could do this in the back end once we improve the template editing there. I'd like to parse + output syntax errors there anyway (on save). But that's a whole new topic, IMO.

aschempp:
> How about an empty contao/twig-extra-bundle that a user can install and afterwards install one of it’s suggestions?

Yes that would work, but would we still need to require the package in the core (this PR)? Would we need that for the nice message?

We can also add the twig extra packages to the manager search result, but I'm unsure if that would be good **U**ser E**x**perience? 😆 It would for example suggest them for Contao 4.9 as well where it does not make much sense.

m-vo:
> would we still need to require the package in the core (this PR)? Would we need that for the nice message?

IMO yes, the point of this PR is that you can require something and you're ready to go. Without it you would either have to require the `twig/extra-bundle` as well or add config yourself.

aschempp:
We could also have a `contao/html-twig-bundle` that requires both, but that also seems pretty overkill 🤔 
Maybe we should discuss the best approach in a call?

m-vo:
> Maybe we should discuss the best approach in a call?

:+1:

aschempp:
@leofeyer can we label this up-for-discussion or do we need to create a separate issue?

m-vo:
…we could also use this open issue for that: https://github.com/contao/contao-manager/issues/653

leofeyer:
Please open a new issue as we cannot reopen this one.

PR #3251, opened 2021-08-02
Harden literal insert tag replacement

### LINKED ISSUE #3249: Literal insert tag replacement kills Symfony HtmlDumper output
The `HtmlDumper` injects JS code like this ([ref](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L436)):

```js
if (/[\[{]$/.test(elt.previousSibling.nodeValue)) {
    // …
}
```

Our `FrontendTemplate` class however does the following on the whole buffer:
https://github.com/contao/contao/blob/8468881d30b54c448310bc2456920d4ec328e088/core-bundle/src/Resources/contao/classes/FrontendTemplate.php#L155-L156

Which means, it effectively destroys the regular expression. 

To reproduce:

* create `fe_page.html.twig`
* enter `{{ dump(app) }}`
* open a page in the browser (the dumper JS is broken)
* now comment out the above line in `FrontendTemplate` (it's working again)

m-vo:
This is an attempt at fixing #3249.

This way only `[{]the_tag[}]` will be transformed to `{{the_tag}}`, arbitrary single delimiters stay the same.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
@@ -152,7 +152,7 @@ class FrontendTemplate extends Template
 			$this->strBuffer = $this->minifyHtml($this->strBuffer);
 		}
 
-		// Replace literal insert tags (see #670)
+		// Replace literal insert tags (see #670, #3249)
 		$this->strBuffer = preg_replace_callback(
 			'/<script[^>]*>.*?<\/script[^>]*>|\[[{}]]/is',
 			static function ($matches)
--- a/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
@@ -154,7 +154,7 @@ class FrontendTemplate extends Template
 
 		// Replace literal insert tags (see #670)
 		$this->strBuffer = preg_replace_callback(
-			'/<script[^>]*>.*?<\/script[^>]*>|\[[{}]\]/is',
+			'/<script[^>]*>.*?<\/script[^>]*>|\[[{}]]/is',
 			static function ($matches)
 			{
 				return $matches[0][0] === '<' ? $matches[0] : $matches[0][1] . $matches[0][1];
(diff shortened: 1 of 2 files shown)


bytehead (dismissed):

leofeyer:
I doubt that this will work reliably. E.g. `[{]foo[}] bar [{]foo[}]` would be replaced with `{{foo[}] bar [{]foo}}`.

bytehead:
> E.g. [{]foo[}] bar [{]foo[}] would be replaced with {{foo[}] bar [{]foo}}.

You're right. Didn't think about that case. Regex is not made for things like this 🙈

core-bundle/src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -153,7 +153,7 @@ protected function compile()
 		}
 
 		// Replace literal insert tags (see #670)
-		$this->strBuffer = str_replace(array('[{]', '[}]'), array('{{', '}}'), $this->strBuffer);
+		$this->strBuffer = preg_replace('/\[{](.*)\[}]/', '{{$1}}', $this->strBuffer);
```
ausi:
```suggestion
		$this->strBuffer = preg_replace_callback('/<script[^>]*>.*?<\/script[^>]*>|\[[{}]\]/is', function ($matches) {
			return $matches[0][0] === '<' ? $matches[0] : $matches[0][1] . $matches[0][1];
		}, $this->strBuffer);
```

This should exclude script tags (untested). Maybe this is a better “workaround” for now?

m-vo:
> I doubt that this will work reliably. E.g. [{]foo[}] bar [{]foo[}] would be replaced with {{foo[}] bar [{]foo}}.

You're right, I forgot the `/U` (ungreedy) modifier.

leofeyer:
LGTM now. 👍🏻

m-vo:
This uses @ausi's suggestion now and has some tests added.

ausi (approved):

bytehead (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #3252, opened 2021-08-03
Allow overwriting form settings in includes

bezin:
Hi there, 

I miss some things from the core forms functionality for a long time already. I guess we all use the Notification Center for everything that needs more attention to detail, but for a very simple contact form it seems a little overkill to me. 

What I miss from the core for my simplest use-cases is:

1. ~Set the from address and name from the back-end~ Can and should be done via mailer configuration
2. Overwrite some config parameters, when I include a form several times (element or module)
3. (Use simple token in the subject to show the users email address or name)

This PR currently targets 1 and 2 (elements). 3 might follow in another PR.

~I remember discussing the `from address` topic before, but I could not find the thread or issue. I think the issue back then was, that some feared, more email-related issues as you can enter any email address in that field, but need to take several measures to actually make that work. On the other hand, the default admin email address is my companies address in most cases, but the clients' server is not allowed to send emails from this address either. Hence I think, this should be doable with additional documentation.~

Happy to hear your thoughts :-)


fritzmg:
The from address can already be customised by using a specific mailer configuration. In my opinion it should not be configurable directly in the back end, in order to prevent mistakes by non-admin users.

bezin:
Ahh, I did not know that, but that makes sense. Thanks for the hint. Would only procede with overwriting values then.

cliffparnitzky:
Nice initiative @bezin ... I would also suggest to add the `jumpTo` (redirect page) to the overwritable parameters. This will make the form usable in different page trees.

bezin:
That is already included in the last commits. I will continue working on another branch for now, so I will stop spamming everybody else with notifications :-)

aschempp:
I like the idea of being able to override the `jumpTo` of a form. Not sure about the recipient etc. though. Also, the fields are still just in one language, so how would that form be useful across multiple domains/languages?

bezin:
We have a project currently, where this would have been helpful, because we manage six websites (different subdivisions of one company, so to say) in one system, but all pages are in German. However we need a different recipient per root page / subdivision. 

It certainly does not solve every issue, but it provides some more features than the original form include. For everything else, there is the Notification Center :-)

ausi:
> Also, the fields are still just in one language, so how would that form be useful across multiple domains/languages?

How about using something like `{{label::…}}`?

core-bundle/src/Resources/contao/classes/Hybrid.php
```diff
@@ -296,6 +298,27 @@ protected function isHidden()
 		return true;
 	}
 
+	private function overwriteFormConfiguration(): void
+	{
+		if ($this->objParent && $this->objParent->form_overwriteConfig === '1')
```
fritzmg:
```suggestion
		if ($this->objParent->form_overwriteConfig ?? false)
```

fritzmg (changes requested):

core-bundle/src/EventListener/DataContainer/FormOverwriteListener.php
```diff
@@ -33,35 +41,12 @@ public function addOverwritableFields(string $table): void
         System::loadLanguageFile('tl_form');
```
bezin: Is there any way load the language file via a service? I know I could adapt the `\Contao\DcaLoader` for the data container, but I did not find a suitable replacement for `System::loadLanguageFile`

m-vo: If you need the `$GLOBALS` array to be populated that's the only way to do AFAIK. For everything else you can use the translator service.

ausi: The translator service would also populate the `$GLOBALS` array if you use something like `->trans(…, 'contao_tl_form')` but I’m not sure if you should rely on that.

An alternative would be to inject the framework and use `$this->framework->getAdapter(System::class)->loadLanguageFile('tl_form');`

bezin: Thanks for your hints, I'll use the adapter pattern for the language files then 👍

bezin:
I cleaned this mess up a little. Don't really know what Psalm complains about, as explicitly declared the criticized array. Anyhow, feedback much appreciated. Will target this after my holidays again :-)

core-bundle/src/EventListener/DataContainer/FormOverwriteListener.php
```diff
@@ -0,0 +1,104 @@
...
+
+        if (!isset($GLOBALS['TL_DCA'][$table]['fields'][$targetField]['load_callback'])) {
+            $GLOBALS['TL_DCA'][$table]['fields'][$targetField]['load_callback'] = [];
+        }
+
+        $GLOBALS['TL_DCA'][$table]['fields'][$targetField]['load_callback'][] = [
+            self::class, 'getPlaceholderFromForm',
+        ];
+    }
+}
```
fritzmg: Personally I would make the following adjustments:

* Use `__invoke` instead of `addOverwritableFields`.
* Put the annotation on the class, instead of the method.
* Rename the class to `AddOverwritableFormSettingsListener`.

bezin: Yep, that sounds more easily comprehensible 👍

bezin: After discussing it on Slack and trying out a couple things I cannot come up with a solution which is better. I added a comment to the actual callback implementation to clarify that it is registered dynamically and where this happens.

fritzmg (dismissed):

fritzmg:
@bezin could you please resolve the conflicts 🙃

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -132,7 +132,8 @@
 		'defineRoot'                  => 'rootPage',
 		'protected'                   => 'groups',
 		'reg_assignDir'               => 'reg_homeDir',
-		'reg_activate'                => 'reg_jumpTo,reg_text'
+		'reg_activate'                => 'reg_jumpTo,reg_text',
+		'form_overwriteSettings'        => '',
```
fritzmg:
```suggestion
		'form_overwriteSettings'      => '',
```

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -159,7 +159,8 @@
 		'protected'                   => 'groups',
 		'splashImage'                 => 'singleSRC,size',
 		'markdownSource_sourceText'   => 'code',
-		'markdownSource_sourceFile'   => 'singleSRC'
+		'markdownSource_sourceFile'   => 'singleSRC',
+		'form_overwriteSettings'        => '',
```
fritzmg:
```suggestion
		'form_overwriteSettings'      => '',
```

fritzmg (changes requested):

fritzmg (dismissed):

Toflar (dismissed):
First of all, I can totally see how this can make sense and I appreciate the work on this @bezin!
I'm just a bit hesitant with this one because it will dynamically extend the `tl_content` and `tl_module` tables with new columns even more. However, you don't really have a lot of options here as there's still no other solution for that.

bezin:
> I'm just a bit hesitant with this one because it will dynamically extend the tl_content and tl_module tables with new columns even more. However, you don't really have a lot of options here as there's still no other solution for that.

I understand that. I went with the established pattern, as there was no obvious alternative. What might be possible in the future? Store module config as an serialized array? As JSON?

fritzmg:
> I'm just a bit hesitant with this one because it will dynamically extend the tl_content and tl_module tables with new columns even more. However, you don't really have a lot of options here as there's still no other solution for that.

This is still something I want to create a draft PR on - but it likely won't happen for `4.13` anymore 🙃

leofeyer (changes requested):
We are not using a `form_` prefix for any form fields, therefore I would prefer to rename `form_overwriteSettings` to `overwriteFormSettings` (or maybe something shorter).

core-bundle/src/Resources/config/listener.yml
```diff
@@ -81,6 +81,10 @@ services:
             - '@database_connection'
             - '@translator'
 
+    Contao\CoreBundle\EventListener\DataContainer\OverwriteFormSettingsListener:
```
leofeyer: This should be `contao.listener.data_container.overwrite_form_settings`.

bezin:
Failing Windows CI test seems unrelated 🤔

bezin:
Failing Windows CI test was unrelated :-)

core-bundle/src/Resources/contao/classes/Hybrid.php
```diff
@@ -296,6 +298,27 @@ protected function isHidden()
...
+					continue;
+				}
+
+				if ($value)
+				{
+					$field = str_replace('form_', '', $key);
+					$this->{$field} = $this->objParent->{$key};
+				}
+			}
+		}
```
leofeyer:
```suggestion
		if (!$this->objParent->formSettings)
		{
			return;
		}

		// Overwrite the form configuration
		foreach ($this->objParent->row() as $key => $value)
		{
			if (!str_starts_with($key, 'form_'))
			{
				continue;
			}

			if ($value)
			{
				$field = str_replace('form_', '', $key);
				$this->{$field} = $this->objParent->{$key};
			}
		}
```

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -808,6 +809,13 @@
 			),
 			'sql'                     => "int(10) unsigned NOT NULL default 0"
 		),
+		'formSettings' => array
+		(
+			'exclude'                 => true,
+			'inputType'               => 'checkbox',
+			'eval'                    => array('tl_class'=>'w50 clr', 'submitOnChange'=>true),
```
leofeyer: Why do we need `clr` here? The widget is the first one.

core-bundle/src/Resources/contao/classes/Hybrid.php
```diff
@@ -296,6 +298,27 @@ protected function isHidden()
...
 	}
 
+	private function overwriteConfigFromParent(): void
+	{
+		if ($this->objParent->formSettings ?? false)
+		{
+			// Overwrite form configuration
+			foreach ($this->objParent->row() as $key => $value)
+			{
+				if (!str_starts_with($key, 'form_'))
```
aschempp: shouldn't this check the DCA config for overwrite-fields instead?

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -808,6 +809,13 @@
 			),
 			'sql'                     => "int(10) unsigned NOT NULL default 0"
 		),
+		'formSettings' => array
```
aschempp:
```suggestion
		'formOverride' => array
```

aschempp (changes requested):
Generally looks good, the concept of dynamically merging other DCA fields is interesting and I hope it will not be problematic 😅 
The term `overwrite` sounds wrong to my non-native english ears. I would prefer `override`, also according to https://www.grammar.com/override_vs_overwrite

fritzmg:
> The term overwrite sounds wrong to my non-native english ears. I would prefer override, also according to https://www.grammar.com/override_vs_overwrite

The functionality actually overwrites parameters in Hybrid [here](https://github.com/contao/contao/pull/3252/files#diff-2ccc6cc25abaf2bfbf8e34324dc4cae9802090822e36219c1a166a2ae1591dd5R316), so I think overwrite is correct?

aschempp:
https://www.grammar.com/override_vs_overwrite

> Override means to add to or enhance something apart from its existing behaviour. It is used very commonly in programming languages.

> Overwriting something is to completely remove the old version and then add a new version. The previous version is then lost.

We are **overriding** the form configuration in this specific module or element. We are not _overwriting_ the original configuration, it is not lost.

fritzmg:
> We are not _overwriting_ the original configuration, it is not lost.

Depends on how you look at it. Within the object instance it is lost ;)

core-bundle/src/Resources/contao/classes/Hybrid.php
bezin: Yes, you are right. Fixed in https://github.com/contao/contao/pull/3252/commits/38faeb9e25e2f2d2e90f5d74831384a68d36e8cb

bezin:
After reading the grammar article I kinda lean towards overriding, but:

In `tl_content` we already have `overwriteMeta` and `overwriteLink` which do the exact same thing: They toggle a subpalette that allow you to `[override|overwrite]` some fields. 

Given that, I would like to keep overwrite and also revert the renaming of the field `overwriteFormSettings` to `formSettings` in order to keep naming consistent.

core-bundle/src/Resources/contao/dca/tl_content.php
bezin: Fixed in 9e580618c

core-bundle/src/Resources/contao/classes/Hybrid.php
bezin: Fixed in 38faeb9e2

leofeyer:
> Given that, I would like to keep overwrite and also revert the renaming of the field `overwriteFormSettings` to `formSettings` in order to keep naming consistent.

Yes, go for it. 👍

frontendschlampe:
We included this feature in our extension https://github.com/hofff/contao-formtools since many years ;-)

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -133,7 +133,8 @@
 		'defineRoot'                  => 'rootPage',
 		'protected'                   => 'groups',
 		'reg_assignDir'               => 'reg_homeDir',
-		'reg_activate'                => 'reg_jumpTo,reg_text'
+		'reg_activate'                => 'reg_jumpTo,reg_text',
+		'formSettings'                => '',
```
m-vo: Judging from the failing functional test these are currently expected to be called `overwriteFormSettings`.

bezin: Thanks! Fixed in 4487b7e

leofeyer (changes requested):
Are we really sure that we want to add this to the core? The PR adds quite a few columns to two tables that already have a lot of columns, so IMHO we should only do this if the majority of users really needs the feature.

m-vo:
That's probably the perfect use case for a JSON column that combines all values.

bezin:
>Are we really sure that we want to add this to the core? The PR adds quite a few columns to two tables that already have a lot of columns, so IMHO we should only do this if the majority of users really needs the feature.

I see your point. However this is currently the only way to do this. The Form include needs these columns and that is the way the core handles it 🤷  

> That's probably the perfect use case for a JSON column that combines all values.

I agree, I and @fritzmg discussed this briefly already.

fritzmg:
Alternatively we would have to implement new events/hooks in `Hybrid` so that this could be done as an extension... but that would need to be added as a bugfix for `4.13` 🤷

bezin:
We could also do not have overwritable fields by default and anybody who would want to use this, could set the DCA flag `formOverwritable` manually? This way, the columns would only be added, if necessary.

leofeyer:
As discussed in the Contao call, we would prefer if this feature was implemented as an extension. We would add additional extension points if necessary.

Regardless, we want to address the "too many columns" problem on the next developers meeting (probably with using JSON columns in the database and merging #4179).

leofeyer:
Thank you @bezin and sorry for eventually not merging this. 🙈

Issue #2862, opened 2021-03-10
Deprecated classes in Symfony 5.1


ausi:
Deprecated since Symfony version 5.1

Interfaces:

- `Symfony\Component\Security\Http\Logout\LogoutHandlerInterface`
- `Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface`

Classes:
- `Symfony\Component\Security\Http\Logout\DefaultLogoutSuccessHandler`

We should check if we can update/remove the usage of those classes (without breaking compatibility with other Symfony versions), or otherwise keep this issue active until we increase the Symfony requirement to 5.1 or higher.

Related: #2857

**Update 2021-07-19:** Also check `Symfony\Component\EventDispatcher\LegacyEventDispatcherProxy` from #3206

fritzmg:
Also related to https://github.com/contao/contao/pull/2818. That bug would not occur with the new security components.

aschempp:
I will most likely have a client budget (and time) to upgrade Symfony security this year 😎

ausi:
I think `LogoutSuccessHandlerInterface` was already removed. 

`LegacyEventDispatcherProxy` only needs to be removed from the tests and from the `DeprecatedClasses`-phpunit-extension it seems.

`LogoutHandlerInterface` and `DefaultLogoutSuccessHandler` are still in use I think.

@bytehead can you confirm that?

bytehead:
Yes, I can confirm that. I've added the `LogoutSuccessListener` to deprecate the `LogoutSuccessHandler`. I'm not sure how to replace the `LogoutHandler` - maybe we should integrate the logic to the `LogoutSuccessListener` and deprecate it as well?

leofeyer:
Yes, if that works. 👍 Although, since we will switch to the new authentication system in Contao 5, maybe we should not spend too much time on this.

bytehead:
The documentation says so (there is only one event in the new authentication system): https://symfony.com/doc/current/security.html#customizing-logout. I'll provide a PR.

leofeyer:
@ausi Can I move this issue to the `5.0` milestone or is there anything left to do for Contao 4.13?

ausi:
> I'm not sure how to replace the `LogoutHandler` - maybe we should integrate the logic to the `LogoutSuccessListener` and deprecate it as well?

Don’t know if the PR @bytehead wants to provide should go into 4.13 or 5.0?

bytehead:
[...]

PR #3257, opened 2021-08-03
Fix the Symfony 5.3 security deprecations

bytehead:
See https://github.com/contao/contao/pull/3215#issuecomment-891731721

### DIFF
--- a/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
+++ b/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
@@ -30,8 +30,8 @@ use Symfony\Component\Security\Http\Session\SessionAuthenticationStrategyInterfa
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 /**
- * @deprecated since Contao 4.13, to be removed in Contao 5.0.
- *             Use the new authenticator system instead
+ * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0.
+ *             Use the new authenticator system instead.
  */
 class ContaoLoginAuthenticationListener extends AbstractAuthenticationListener
 {
--- a/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
+++ b/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
@@ -52,13 +52,13 @@ class ContaoLoginAuthenticationListener extends AbstractAuthenticationListener
         $currentToken = $this->tokenStorage->getToken();
 
         if ($currentToken instanceof TwoFactorTokenInterface) {
-            $authCode = $request->request->all()['verify'] ?? '';
+            $authCode = $request->request->get('verify');
 
             return $this->authenticationManager->authenticate($currentToken->createWithCredentials($authCode));
         }
 
-        $username = $request->request->all()['username'] ?? null;
-        $password = $request->request->all()['password'] ?? null;
+        $username = $request->request->get('username');
+        $password = $request->request->get('password');
 
         if (!\is_string($username)) {
             throw new BadRequestHttpException(sprintf('The key "username" must be a string, "%s" given.', \gettype($username)));
--- a/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
+++ b/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
@@ -29,6 +29,10 @@ use Symfony\Component\Security\Http\HttpUtils;
 use Symfony\Component\Security\Http\Session\SessionAuthenticationStrategyInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
+/**
+ * @deprecated since Contao 4.13, to be removed in Contao 5.0.
+ *             Use the new authenticator system instead
+ */
 class ContaoLoginAuthenticationListener extends AbstractAuthenticationListener
 {
     private TokenStorageInterface $tokenStorage;
--- a/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
+++ b/core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
@@ -52,13 +52,13 @@ class ContaoLoginAuthenticationListener extends AbstractAuthenticationListener
         $currentToken = $this->tokenStorage->getToken();
 
         if ($currentToken instanceof TwoFactorTokenInterface) {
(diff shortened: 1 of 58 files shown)


ausi:
You can increase all Symfony requirements from `^4.4 || ^5.2` to `^5.3`. This should make fixing this easier.

ausi:
Also related: #2862

bytehead:
All security related deprecations are gone except the ones we only can replace with the new authentication system (Contao 5.x):

`... is deprecated since Symfony 5.3, use the new authenticator system instead.`

core-bundle/src/Security/User/ContaoUserProvider.php
```diff
@@ -50,18 +50,23 @@ public function __construct(ContaoFramework $framework, SessionInterface $sessio
     }
 
     public function loadUserByUsername($username): User
+    {
+        return $this->loadUserByIdentifier($username);
```
leofeyer, line 60: Do we have to deprecate the `loadUserByUsername()` method as well?

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
```diff
@@ -55,13 +55,13 @@ protected function attemptAuthentication(Request $request): ?TokenInterface
         $currentToken = $this->tokenStorage->getToken();
 
         if ($currentToken instanceof TwoFactorTokenInterface) {
-            $authCode = $request->request->get('verify');
+            $authCode = $request->request->all()['verify'] ?? '';
```
leofeyer: Why this change?

core-bundle/src/Security/User/ContaoUserProvider.php
bytehead: Fixed in ab36ce407.

core-bundle/src/Security/Authentication/ContaoLoginAuthenticationListener.php
bytehead: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpFoundation/InputBag.php#L39

bytehead: But this (`non-string value`) is more a theoretical case during tests.

bytehead: Fixed in cb47aa7d3.

leofeyer (dismissed):

leofeyer:
There are still some security-related deprecations when running the functional tests:

```
  2x: Since symfony/security-bundle 5.1: The "success_handler" at path "security.firewalls.contao_backend.logout" is deprecated, register a listener on the "Symfony\Component\Security\Http\Event\LogoutEvent" event instead.
    2x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  2x: Since symfony/security-bundle 5.1: The "success_handler" at path "security.firewalls.contao_frontend.logout" is deprecated, register a listener on the "Symfony\Component\Security\Http\Event\LogoutEvent" event instead.
    2x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  2x: Since symfony/security-bundle 5.3: The child node "encoders" at path "security" is deprecated, use "password_hashers" instead.
    2x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration
```

@bytehead Do you want to fix these here or in a separate PR?

bytehead:
I'll fix them in this PR. 👍

bytehead:
@leofeyer I've replaced the depreacted `LogoutSuccessHandler` with a corresponding listener. Somehow there is still some deprecations, but I don't know where they come from as I've adapted the security configuration for the functional tests as well.

```
   1x: Since symfony/security-http 5.1: The "Symfony\Component\Security\Http\Logout\DefaultLogoutSuccessHandler" class is deprecated, use "Symfony\Component\Security\Http\EventListener\DefaultLogoutListener" instead.
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Since symfony/security-http 5.1: The "Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface" interface is deprecated, create a listener for the "Symfony\Component\Security\Http\Event\LogoutEvent" event instead.
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration
```

bytehead:
I've catched some more deprecations, the one mentioned [before](https://github.com/contao/contao/pull/3257#issuecomment-967229687) still exist. I have no clue what triggers them.

core-bundle/tests/Security/Logout/LogoutSuccessHandlerTest.php
```diff
@@ -41,6 +47,8 @@ public function testRedirectsToAGivenUrl(): void
             ->willReturn(false)
         ;
 
+        $this->expectDeprecation('Since symfony/security-http 5.1: The "%s" class is deprecated, use "Symfony\Component\Security\Http\EventListener\DefaultLogoutListener" instead.');
```
ausi: Deprecations that occur only once when the php file is autoloaded need to be handled in the *tests/PhpunitExtension/DeprecatedClasses.php* file, because the file could have already been autoloaded somewhere else and the deprecation would then not be triggered again.

bytehead:
These won't be triggered while unit testing, only during the functional tests. I can't add them to 
`Contao\CoreBundle\Tests\PhpunitExtension\DeprecatedClasses`, as the unit test will fail then.

```
1x: Since symfony/security-http 5.1: The "Symfony\Component\Security\Http\Logout\DefaultLogoutSuccessHandler" class is deprecated, use "Symfony\Component\Security\Http\EventListener\DefaultLogoutListener" instead.
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration

  1x: Since symfony/security-http 5.1: The "Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface" interface is deprecated, create a listener for the "Symfony\Component\Security\Http\Event\LogoutEvent" event instead.
    1x in RoutingMigrationTest::testShouldRun from Contao\CoreBundle\Tests\Functional\Migration
```

bytehead:
And I don't know, if we should deprecate most of the security related classes as they won't exist in Contao 5 anymore?

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #622, opened 2019-08-07
Automatically generate response from template in fragment controllers

aschempp:
This PR fixes and improves two things:

1. The `$template->getResponse()` method will add cache headers for the current page, because it _assumes_ it needs to generate the response for a page type. Therefore we should **never** use that method.

2. The PR also simplifies the controller, because I no longer need to generate anything.

Due to the BC promise, we cannot change the `getResponse` method name. But we can *weaken* the response type, because existing child classes will *enforce* a response return, which is fine.

PR #3258, opened 2021-08-04
Deprecated the return of null response in fragments

aschempp:
This should never have made it into the code base. When I suggested this in https://github.com/contao/contao/pull/622 I actually thought you don't need to write `return …` in the `getResponse()` method. But writing `return null` does not make sense at all…

The idea is that we can change the return value of the abstract method again in Contao 5


m-vo:
Funny, I was just in the process of writing an issue that covers the exact same method. :smile: 

IMO we should return an empty response instead of rendering the template when `null` was returned in `getReponse()`. 

:+1: for deprecating that `null` could be returned in the first place.

m-vo:
If we're at it: Should anyone ever call `return $template->getResponse()` in their fragment controllers? Shouldn't that be `return new Response($template->parse());`? 

If so, we should fix the docs and maybe make it more clear how to use this somehow…

leofeyer:
You have assigned the PR to the `4.12` milestone but the base branch is `4.x`. Can you rebase the PR please?

aschempp:
> IMO we should return an empty response instead of rendering the template when `null` was returned in `getReponse()`.

No, we can't change that. We shouldn't allow `null` at all. Just return an empty response yourself.


> If we're at it: Should anyone ever call `return $template->getResponse()` in their fragment controllers? Shouldn't that be `return new Response($template->parse());`?

This was an issue in earlier Contao 4 versions. It should be fixed so that `$template->getResponse()` works correctly and is the correct way to do it. It will for example generate correct caching headers.

m-vo:
> It should be fixed so that $template->getResponse() works correctly and is the correct way to do it. 

It will also currently execute `FrontendTemplate#compile()` which cannot be right.

m-vo:
If we are setting cache headers, why not in the `AbstractFragmentController`? 

There are multiple ways how users will return a response in a fragment's `getReponse()` method:

- `return $template->getResponse();`
- `return new Response('foo');`
- `return $this->render('foo.html.twig');`

For the last two cases I would have to set headers myself while the first one would set them for me?

leofeyer (approved):

leofeyer:
I am merging this PR as it is correct to add the deprecation.

@m-vo Can you open a separate issue regarding `$template->getResponse()` vs. `new Response($template->parse())`? I think you have a point here.

m-vo:
> @m-vo Can you open a separate issue regarding $template->getResponse() vs. new Response($template->parse())? I think you have a point here.

see https://github.com/contao/contao/issues/3263

PR #3267, opened 2021-08-04
Do not add the native suffix in System::getLanguages()

leofeyer:
The native suffix has not been added in Contao <4.12, either, therefore we should not change the behavior.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -606,10 +606,10 @@ abstract class System
 
 		if ($blnInstalledOnly)
 		{
-			return self::getContainer()->get(Locales::class)->getEnabledLocales(null, true);
+			return self::getContainer()->get(Locales::class)->getEnabledLocales();
 		}
 
-		return self::getContainer()->get(Locales::class)->getLocales(null, true);
+		return self::getContainer()->get(Locales::class)->getLocales();
 	}
 
 	/**


aschempp:
Where is that dropdown from? This is Contao 4.9.16 demo.contao.org and there sure are native suffixes.

Bildschirmfoto 2021-08-04 um 16 20 02

leofeyer:
Congrats, you have found the one place where we actually do have native suffixes. 😄 But if you have `'options' => System::getLanguages(),` in your DCA, we did not add native suffixes and as you can see in the screenshot, they are pretty confusing there.

leofeyer:
Ok, it seems that you are right. However, now that we are using Intl, the labels have become a lot longer, which is pretty confusing IMHO.

aschempp:
is it because we now really have all the options available and previously it was just simpler because not everything was there?

fritzmg:
I think `System::getLanguages()` should continue to only provide ISO 639-1 two-letter codes, not the full list of IETF language tags, (ISO 639-1 + ISO 3166-1) (by default).

leofeyer:
Yes, that would be much better in terms of backwards compatibility.

aschempp:
> … should **continue** to only provide two-letter codes …

Unfortunately it already did display locales 😉 

This is the member language selection

Bildschirmfoto 2021-08-04 um 16 37 05

fritzmg:
True. Using intl just added more variants - which unfortunately now have a long label, like the different Chinese ones. Contao 4.9 only has 4 Chinese variants, while 4.12 now has 10.

Imho we should keep everything as is, since the list itself as well as their labels are correct.

ausi:
I think the native suffix can be pretty helpful if the dropdown is used by the user or member themself to set their own language.

See for example the following screenshot, without the native suffix I would not be able to spot my language:

screenshot-wrong-language

How about showing the suffix dependig on the context?
Pull request for this approach: #3269

leofeyer:
Closing this in favor of #3269.

PR #3268, opened 2021-08-04
Fix a type error in the event list module

leofeyer:
`Events::getSchemaOrgData()` expects a `CalendarEventsModel` and not an array.

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -401,7 +401,7 @@ class ModuleEventlist extends Events
 			// schema.org information
 			$objTemplate->getSchemaOrgData = static function () use ($objTemplate, $event): array
 			{
-				$jsonLd = Events::getSchemaOrgData($event);
+				$jsonLd = Events::getSchemaOrgData((new CalendarEventsModel())->setRow($event));
 
 				if ($objTemplate->addImage && $objTemplate->figure)
 				{
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -1526,11 +1526,8 @@ abstract class Controller extends System
 		{
 			if ($interpretAsContentModel)
 			{
-				/** @var ContentModel $contentModel */
-				$contentModel = (new \ReflectionClass(ContentModel::class))->newInstanceWithoutConstructor();
-
 				// This will be null if "overwriteMeta" is not set
-				return $contentModel->setRow($rowData)->getOverwriteMetadata();
+				return (new ContentModel())->setRow($rowData)->getOverwriteMetadata();
 			}
 
 			// Manually create metadata that always contains certain properties (BC)


bytehead (dismissed):

calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
```diff
@@ -401,7 +401,7 @@ protected function compile()
 			// schema.org information
 			$objTemplate->getSchemaOrgData = static function () use ($objTemplate, $event): array
 			{
-				$jsonLd = Events::getSchemaOrgData($event);
+				$jsonLd = Events::getSchemaOrgData((new CalendarEventsModel())->setRow($event));
```
Toflar, line 404: Don't we have the model instance available?

bytehead: There is, L354 https://github.com/contao/contao/pull/3268/files#diff-7ab0b7a52b7e174853ae8b92164755ee9ecbd4461608e13ed3d5aa8431e98507L354

leofeyer: No, unfortunately not. `Events::getAllEvents()` returns the events as arrays.

bytehead: ah, only if an image is added.

Toflar: Not sure if it's a good idea to expect a model then? Technically apparently it would have been possible to add an event that is not a `CalendarEventsModel` using the hook then? In which case this code here might possibly break because it might possibly contain data the model does not know of? Not sure though but I hope you get my point.

leofeyer: The problem is not that the method requires a model. The other methods such as `Event::addEvent()` or `Event::generateEventUrl()` require a model, too, which is way better than allowing an arbitrary array that might or might not include the required fields.

An alternative solution would be to add the event in the `Events::addEvent()` method somewhere around [line 355](https://github.com/contao/contao/blob/4.12/calendar-bundle/src/Resources/contao/classes/Events.php#L355). We do have the model at this point.

leofeyer: Ah, no, we do not have the template object there. 😢

Toflar: I know. In which case the schema.org would not work if you did not add the model instance yourself. Although to add them, technically speaking, you don't need a model. The data is part of the array. If we change this, schema.org data would not work for events added via hook without a model instance.

leofeyer: You cannot add events that do not have a model. Otherwise this would fail:

https://github.com/contao/contao/blob/ec8b7ee1b6377aee8448ea77b150ee5886f6f2e8/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php#L353-L354

@m-vo /cc

m-vo: The metadata overwrite for the image would also work with the model created from an array. I used the same "hack" in `Controller::addImageToTemplate`: 

https://github.com/contao/contao/blob/722d397e3429e3fc4a8d05263fe2f834aa8f2122/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1529-L1533

leofeyer: I think you could have simply used `new ContentModel()` instead of the reflection workaround. The constructor does not do much if you do not pass a DB result.

m-vo: True, I think I didn't want the `DcaExtractor` to interfere back then, so that you could test these things more in isolation. But we could probably change this now.

ausi (approved):

PR #3269, opened 2021-08-04
Show native language suffix only in user profile

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3267

Alternative to #3267

Shows the native suffix only in the user profile ~~or the member personal data module~~.


aschempp:
I wonder whats the point of showing the native language suffix? I think that only ever made sense on the back end login screen, so you could switch to your local language if you didn't know the current. It does not make sense in the front end to me?

fritzmg:
> It does not make sense in the front end to me?

I agree, but that would be a behaviour change.

ausi:
> I think that only ever made sense on the back end login screen, so you could switch to your local language if you didn't know the current.

I think it also makes sense in the user profile (see https://github.com/contao/contao/pull/3267#issuecomment-892877521 ) as there is no language selection in the login form anymore.

> It does not make sense in the front end to me?

Agree, removed in 5a0b033a96e3c02b965f9103bbcdd99636806017

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #3270, opened 2021-08-04
Always refresh the template hierarchy in the debug:contao-twig command

m-vo:
We should always reset the loader cache + rebuild the template hierarchy when a user runs `debug:contao-twig`. 

The old implementation had a stupid issue where the reset part was run _after_ outputting results, but I could also create scenarios* where the cache wasn't up to date and you would need to reset it manually and _only_ on the command line. Removing the option and always resetting has a much better DX and is quick enough that you won't notice it.

*) e.g. running `contao-console cache:pool:clear cache.system` = clearing the cache pool, no warming

/cc @ausi

### DIFF
--- a/core-bundle/src/Command/DebugContaoTwigCommand.php
+++ b/core-bundle/src/Command/DebugContaoTwigCommand.php
@@ -52,7 +52,6 @@ class DebugContaoTwigCommand extends Command
     {
         $this
             ->setDescription('Displays the Contao template hierarchy.')
-            ->addOption('refresh', 'r', InputOption::VALUE_NONE, 'Refresh the cache.')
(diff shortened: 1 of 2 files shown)


fritzmg:
This will always warm the cache, even in `dev`. But I am assuming that's ok, since it's only for the Twig `ContaoFilesystemLoaderWarmer`?

m-vo:
Yes, and it's only the paths and hierarchy entries from the system cache pool. The `dev` environment was the problem by the way - there you did not expect an outdated cache but could get one.

ausi (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #3275, opened 2021-08-06
Fix missing tooltips for widget help text

ausi:
With the fixes from #2958 the tooltips didn’t show up for the help texts of the widgets anymore.

This PR should fix this I think, @rabauss can you verify?


rabauss:
Oh damn it - I did not test this case of tooltip 😞 
But the fix is looking good to me!

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #3278, opened 2021-08-09
Check if stopwatch is started before stopping

### LINKED ISSUE #3272: Stopwatch not started
**Affected version(s)**

Contao 4.9.17

**Description**

Since the update to 4.9.17 I get a lot of this `LogicExceptions`:

```
Event "contao.frontend_module.header (ID 10)" is not started.
Event "contao.frontend_module.newsreader (ID 16)" is not started.
```

Stacktrace:
```
LogicException:
Event "contao.frontend_module.newsreader (ID 16)" is not started.

  at vendor/symfony/stopwatch/Section.php:142
  at Symfony\Component\Stopwatch\Section->stopEvent('contao.frontend_module.newsreader (ID 16)')
     (vendor/symfony/stopwatch/Stopwatch.php:126)
  at Symfony\Component\Stopwatch\Stopwatch->stop('contao.frontend_module.newsreader (ID 16)')
     (vendor/contao/core-bundle/src/Resources/contao/elements/ContentModule.php:103)
  at Contao\ContentModule->generate()
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:607)
  at Contao\Controller::getContentElement(object(ContentModel), 'main')
     (vendor/contao/core-bundle/src/Resources/contao/modules/ModuleArticle.php:218)
  at Contao\ModuleArticle->compile()
     (vendor/contao/core-bundle/src/Resources/contao/modules/Module.php:214)
  at Contao\Module->generate()
     (vendor/contao/core-bundle/src/Resources/contao/modules/ModuleArticle.php:71)
  at Contao\ModuleArticle->generate(false)
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:536)
  at Contao\Controller::getArticle(object(ArticleModel), false, false, 'main')
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:377)
  at Contao\Controller::getFrontendModule('0', 'main')
     (vendor/contao/core-bundle/src/Resources/contao/pages/PageRegular.php:168)
  at Contao\PageRegular->prepare(object(PageModel))
     (vendor/contao/core-bundle/src/Resources/contao/pages/PageRegular.php:48)
  at Contao\PageRegular->getResponse(object(PageModel), true)
     (vendor/contao/core-bundle/src/Resources/contao/controllers/FrontendIndex.php:337)
     (public/index.php:43)
```

Possibly related to #2940

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3272
| Docs PR or issue | n/a

This could be a fix for Contao 4.9.20+

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentModule.php
+++ b/core-bundle/src/Resources/contao/elements/ContentModule.php
@@ -100,7 +100,7 @@ class ContentModule extends ContentElement
 
 		if (isset($objStopwatch) && $objStopwatch->isStarted('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')'))
 		{
-		    $objStopwatch->stop('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')');
+			$objStopwatch->stop('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')');
 		}
 
 		return $strBuffer;
--- a/core-bundle/src/Resources/contao/elements/ContentModule.php
+++ b/core-bundle/src/Resources/contao/elements/ContentModule.php
@@ -98,9 +98,9 @@ class ContentModule extends ContentElement
 
 		$strBuffer = $objModule->generate();
 
-		if (isset($objStopwatch))
+		if (isset($objStopwatch) && $objStopwatch->isStarted('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')'))
 		{
-			$objStopwatch->stop('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')');
+		    $objStopwatch->stop('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')');
 		}
 
 		return $strBuffer;
--- a/core-bundle/src/Resources/contao/elements/ContentModule.php
+++ b/core-bundle/src/Resources/contao/elements/ContentModule.php
@@ -80,10 +80,12 @@ class ContentModule extends ContentElement
 		$objModel->cssID = $cssID;
 		$objModel->typePrefix = 'ce_';
 
+		$strStopWatchId = 'contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')';
+
 		if (System::getContainer()->getParameter('kernel.debug'))
 		{
 			$objStopwatch = System::getContainer()->get('debug.stopwatch');
-			$objStopwatch->start('contao.frontend_module.' . $objModel->type . ' (ID ' . $objModel->id . ')', 'contao.layout');
+			$objStopwatch->start($strStopWatchId, 'contao.layout');
 		}
 
 		/** @var Module $objModule */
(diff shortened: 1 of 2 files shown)


fritzmg (dismissed):

core-bundle/src/Resources/contao/elements/ContentModule.php
```diff
@@ -98,9 +100,9 @@ public function generate()
 
 		$strBuffer = $objModule->generate();
 
-		if (isset($objStopwatch))
+		if (isset($objStopwatch) && $objStopwatch->isStarted($strStopWatchId))
```
m-vo, line 103: Did you find out why the stopwatch isn't started at this point? Is something during `$objModule->generate();` stopping it?

bytehead: No, but as I understand it: different requests do have different stopwatch sections and there we get the little mess: https://github.com/symfony/symfony/issues/36623

bytehead: In my real life demo all stopwatches get started, but after the first subrequest all following `$stopwatch->stop()` will fail and `objStopwatch->isStarted($strStopWatchId)` will return `false`.

bytehead: I can follow this: https://github.com/symfony/symfony/issues/36762#issuecomment-626352054

m-vo: Ah yes, missed the linked ticket. Thanks!

m-vo (dismissed):

fritzmg:
Btw. I think this should get merged regardless of the original issue. It makes sense anyway to always check whether the stopwatch has actually started before trying to stop it. Similar to always checking whether a session is present and started etc. 
Also the stop watch could have been stopped by a third party in between.

bytehead:
> Also the stop watch could be stopped by a third party in between.

Yes. 

Symfony itself checks in [their code using the stopwatch](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Messenger/Middleware/TraceableMiddleware.php#L74)  if it's started upfront.

Toflar (dismissed):

aschempp (dismissed):
I agree we might need to check if a stopwatch was started, but I dislike the fact that it would be silently ignored. If we do start a stopwatch, shouldn't we be the only ones allowed to stop it? Otherwise the stopwatch will not give useful results and no-one will notice 🤷

bytehead:
If we check if it's started upfront, no annoying exception will be thrown in case of an error (the stopwatch is just for debugging purposes anyway). You would also notice the error introduced from Symfony in the profiler, as the timing information will not show up for these modules/content elements which are rendered as fragments. So first people don't get annoying errors and you would notice it if something is wrong when you check the performance tab in the profiler and search for the data you want.

bytehead:
Status quo is now unusable in dev mode (and the information is not _more correctly_ than if it was checked upfront and not shown) and that's a problem 🙃

aschempp:
> You would also notice the error introduced from Symfony in the profiler, as the timing information will not show up for these modules/content elements which are rendered as fragments.

Is that true? Isn't the stopwatch just stopped incorrectly at another point (on kernel.response in the current issue)?

bytehead:
Only if somebody stopps it. Otherwise it gets omitted (without an error anymore in this case).

aschempp:
> Only if somebody stopps it. Otherwise it gets omitted (without an error anymore in this case).

No idea why that should be possible with a stopwatch, but apparently I don't know enough about how it works 😂
I trust you on this 👍

aschempp (approved):

Toflar (approved):

leofeyer:
Thank you @bytehead.

PR #3279, opened 2021-08-09
Run the CI tools and the code inspector

leofeyer:


core-bundle/src/EventListener/RequestTokenListener.php
```diff
@@ -83,8 +83,8 @@ public function __invoke(RequestEvent $event): void
             'POST' !== $request->getRealMethod()
             || $request->isXmlHttpRequest()
             || false === $request->attributes->get('_token_check')
-            || (!$request->attributes->has('_token_check') && !$this->scopeMatcher->isContaoRequest($request))
             || $this->csrfTokenManager->canSkipTokenValidation($request, $this->csrfCookiePrefix.$this->csrfTokenName)
+            || (!$request->attributes->has('_token_check') && !$this->scopeMatcher->isContaoRequest($request))
```
m-vo, line 87: Isn't the `isContaoRequest` call cheaper than `canSkipTokenValidation`?

m-vo (dismissed):

core-bundle/src/EventListener/RequestTokenListener.php
leofeyer: I guess the code inspector only counts the number of function calls. Not sure though.

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -213,7 +214,7 @@ public function fromUuid(string $uuid): self
     {
         $this->lastException = null;
 
-        $filesModel = $this->filesModelAdapter()->findByUuid($uuid);
+        $filesModel = $this->getFilesModelAdapter()->findByUuid($uuid);
```
m-vo, line 217: In case you wondered: this naming scheme was taken from Symfony. :slightly_smiling_face: 
(https://symfony.com/doc/current/service_container/service_subscribers_locators.html#service-subscriber-trait)

core-bundle/src/EventListener/RequestTokenListener.php
ausi: I find this version now harder to read as the two `_token_check` conditions are seperated by a line that checks something different.

Also the performance could be worse now for every non-Contao request I think. (But probably doesn’t really matter).

PR #3281, opened 2021-08-10
Fix compatibility with PHP 8.1

ausi:
💪


bytehead (dismissed):

m-vo (dismissed):

.github/workflows/ci.yml
```diff
@@ -219,7 +219,9 @@ jobs:
               uses: actions/checkout@v1
 
             - name: Install the dependencies
-              run: composer install --ignore-platform-reqs --no-interaction --no-progress
+              run: |
+                  composer install --ignore-platform-req=php --no-interaction --no-progress
+                  composer require "symfony/http-foundation:5.3.*@dev" --ignore-platform-req=php --no-interaction --no-progress
```
Toflar, line 225: Can you explain this change? Or add a comment for anyone to understand?

ausi: Deprecations fail currently with the stable 5.3.6, this will most probably be fixed with 5.3.7. But I would merge it as is so that we see errors with PHP 8.1 in our own codebase.

Toflar: Then we should add a TODO to remove that again once it's fixed, no? I just want to make sure the test setup is fixed afterwards as we're probably testing a wrong combination otherwise :)

ausi:
```suggestion
                  # TODO: Remove the next line once Symfony 5.3.7 was released
                  composer require "symfony/http-foundation:5.3.*@dev" --ignore-platform-req=php --no-interaction --no-progress
```
Like so?

Toflar: Yessa :)

Toflar (dismissed):

leofeyer:
Thank you @ausi.

PR #3283, opened 2021-08-11
Check if the indexer service exists when purging search tables

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | n/a
| Docs PR or issue | n/a

When the default search indexer is disabled and you click on purge search index you'll get a `ServiceNotFoundException`: `You have requested a non-existent service "contao.search.indexer".`

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -35,13 +35,13 @@ class Automator extends System
 	 */
 	public function purgeSearchTables()
 	{
-		$searchIndexer = System::getContainer()->get('contao.search.indexer');
-
 		// The search indexer is disabled
-		if (null === $searchIndexer)
+		if (!System::getContainer()->has('contao.search.indexer'))
 		{
 			return;
 		}
+
+		$searchIndexer = System::getContainer()->get('contao.search.indexer');
 
 		// Clear the index
 		$searchIndexer->clear();
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -43,6 +43,11 @@ class Automator extends System
 
 		$searchIndexer = System::getContainer()->get('contao.search.indexer');
 
+		if (null === $searchIndexer)
+		{
+		    return;
+		}
+
 		// Clear the index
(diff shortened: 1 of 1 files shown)


ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Automator.php
```diff
@@ -35,9 +35,14 @@ public function __construct()
...
+		// The search indexer is disabled
+		if (!System::getContainer()->has('contao.search.indexer'))
+		{
+			return;
+		}
+
 		$searchIndexer = System::getContainer()->get('contao.search.indexer');
 
-		// The search indexer is disabled
 		if (null === $searchIndexer)
```
Toflar: This can be removed as well then, no?

Toflar (changes requested):

core-bundle/src/Resources/contao/library/Contao/Automator.php
bytehead: The `Container::get()` could still return `null`, no?

Toflar: If you check for `has()` previously, I don't think that can work, no :)

bytehead: Removed in ecc8f31db.

Toflar (dismissed):

ausi (approved):

leofeyer:
Thank you @bytehead.

aschempp:
Shouldn't we just not offer this option instead in the back end or automator CLI ?

bytehead:
We could indeed. I've recognized that truncating the search tables needs the indexer service. As you have deactivated the search indexer, you're not able to clear the search tables anymore. Wouldn't make it sense to truncate the tables directly in the automator?
https://github.com/contao/contao/blob/f76879682c1280fbc11e6a20497213368e0939ae/core-bundle/src/Resources/contao/library/Contao/Automator.php#L47

bytehead:
Like here: https://github.com/bytehead/contao/commit/d68a3eb06e401f86e73862af0e00658bb49851fa

leofeyer:
We should open a new issue for that and optionally add the "up for discussion" flag.

bytehead:
I'll do :)

bytehead:
See #3326

PR #3292, opened 2021-08-11
Improve compatibility with JSON in attributes

ausi:
**Affected version(s)**

Contao 4.4.56, 4.9.18, 4.11.7

**Description**

With the security fix some attributes that contain JSON no longer work

```html
<span data-json="{foo:{bar:123}}"></span>
```

gets replaced with

```html
<span data-json="{foo:{bar:123|attr}}"></span>
```

This should be fixed with this pull request (without loosing the safety).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -1210,10 +1210,10 @@ class InsertTags extends Controller
 			$tag = $matches[0][0];
 
 			// Encode insert tags
-			$tag = preg_replace('/(?:\|attr)?}}/', '|attr}}', $tag);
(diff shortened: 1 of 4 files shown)


ausi:
We should fix this in https://github.com/contao/contao/blob/6a676204294a2613013bfb88a1be8d4247ac2254/core-bundle/src/Resources/contao/library/Contao/StringUtil.php#L973-L979 for `specialcharsAttribute()` too.

ausi:
> We should fix this in … for `specialcharsAttribute()` too.

Done in 79e367610bdf7c4ae10e2405f10f6aea3c053bb7

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #3293, opened 2021-08-11
Do not show the urlattr flag for rgxp url fields

ausi:
When saving `rgxp => url` fields we add the `urlattr` flag in https://github.com/contao/contao/blob/6a676204294a2613013bfb88a1be8d4247ac2254/core-bundle/src/Resources/contao/library/Contao/Widget.php#L983-L984

This pull request does the counterpart by removing it when the value is shown to the user.

Related: https://community.contao.org/de/showthread.php?81418


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #3294, opened 2021-08-11
Enable putenv for Dotenv

fritzmg:
While developing a Google Firestore connection within Contao I noticed that defining the project ID and service credentials could not be set via the respective environment variables when using the `.env` files.

This is because `google/cloud-core` for instance reads these environment variables only via `getenv()`. The `ContaoKernel` of the manager bundle however specifically disables the usage of `putenv()` for `Dotenv`.

Is there any specific reason for that? Otherwise I think it would be helpful to enable this for the Managed Edition.

### DIFF
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -406,7 +406,7 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
                 $_ENV[$k] = $_ENV[$k] ?? (isset($_SERVER[$k]) && 0 !== strpos($k, 'HTTP_') ? $_SERVER[$k] : $v);
             }
         } elseif (file_exists($projectDir.'/.env')) {
-            (new Dotenv(false))->loadEnv($projectDir.'/.env', 'APP_ENV', $defaultEnv);
+            (new Dotenv(true))->loadEnv($projectDir.'/.env', 'APP_ENV', $defaultEnv);
         }
 
         $_SERVER += $_ENV;


ausi:
> Is there any specific reason for that?

I think this might have been a mistake in d1b9f2e049649de3ff685b221ee826bd636f5275 because of the message from https://github.com/symfony/symfony/blob/8e45fc043ec74ebc2d112b1ce3196b452a11b4a3/src/Symfony/Component/Dotenv/Dotenv.php#L48

fritzmg:
Ah yes, then it worked before Contao 4.8.0 because in Symfony 4.4 the default is `true`.

_Note:_ with Symfony 5.0 the default was changed to `false` and with Symfony 5.1 the parameter is deprecated and you should use `->usePutenv(true)` instead.

ausi:
@fritzmg do we need to upate these places too?

https://github.com/contao/contao/blob/6a676204294a2613013bfb88a1be8d4247ac2254/manager-bundle/src/ContaoManager/ApiCommand/GetDotEnvCommand.php#L58

https://github.com/contao/contao/blob/6a676204294a2613013bfb88a1be8d4247ac2254/manager-bundle/src/Dotenv/DotenvDumper.php#L44

fritzmg:
> @fritzmg do we need to upate these places too?

I think it does not matter for these places, because they are only used to display the environment variables set via `.env` and they don't retrieve them via `getenv()`. They just use `Dotenv::parse` directly.

ausi:
> Is there any specific reason for that?

Symfony writes in the method description of `usePutenv()` _“Beware that `putenv()` is not thread safe, that's why this setting defaults to false”_

But the thread safety is not really relevant for us I guess? 
Shoud we additionaly create an issue at `google/cloud-store` to tell them about the problems with `getenv()`?

ausi (approved):

fritzmg:
> Symfony writes in the method description of `usePutenv()` _“Beware that `putenv()` is not thread safe, that's why this setting defaults to false”_
> 
> But the thread safety is not really relevant for us I guess?

I am not even sure if that's actually true? The PHP doc do not mention any thread safety issues with `putenv()`. And the PHP doc for `getenv()` mentions the following:

> If PHP is running in a SAPI such as Fast CGI, this function will always return the value of an environment variable set by the SAPI, even if putenv() has been used to set a local environment variable of the same name. Use the local_only parameter to return the value of locally-set environment variables.

leofeyer:
Is this really a Contao issue? Sounds more like `google/cloud-store` should adjust their .env handling, doesn't it?

ausi:
As the PHP doc doesn’t say anything about `getenv()` being deprecated or shouldn’t be used I think I would be in favor of enabling `putenv` in Contao.

From the docs:
Bildschirmfoto 2021-08-12 um 09 07 15

ausi:
From nicolas-grekas in https://github.com/symfony/recipes/pull/571#issuecomment-481711573

> Alternatively, you should ensure to always use `$_SERVER` and never `getenv()`. It's broken anyway.

From nicolas-grekas in https://github.com/symfony/recipes/issues/532#issuecomment-466290599

> My recommendation is to not use `getenv()`: it's not thread safe so using it can break randomly on some machines, it's slower, and DI configuration is more flexible...

fritzmg:
Threads like [this one](https://stackoverflow.com/questions/8798294/getenv-vs-env-in-php) actually overall recommend using `getenv()` over `$_SERVER` or `$_ENV`. Though those are old.

aschempp:
I worry `setenv` can be a security issue on a shared hosting. Why would Google read the parameters from a **running** PHP environment?

Toflar (changes requested):
👎  on this PR for the mentioned reasons.

fritzmg:
> I worry `setenv` can be a security issue on a shared hosting.

In what way?

> Why would Google read the parameters from a **running** PHP environment?

What do you mean by that?

> 👎 on this PR for the mentioned reasons.

Which reasons would that be? So far haven't found any actual references regarding thread safety or security issues when using `putenv()` and `getenv()`. When creating an issue in Google's repository I would need to have some basis for why it should be changed, otherwise there is no real reason for them to change it.

Toflar:
https://mattallan.me/posts/how-php-environment-variables-actually-work/ would be a good read as a start :)

fritzmg:
Thank you :). But that article actually suggests to use `putenv`, as you cannot make sure that `$_SERVER` or `$_ENV` are actually available. Instead you should use `putenv` - but make sure that `putenv` is not used in worker threads. Which I would argue is the case when using something like `Dotenv` in a Symfony application, because you execute that once in the beginning of your script (i.e. the Kernel), before potentially any sub-threads are created.

Toflar:
No, that article suggests NOT to use `putenv()`. It explains the differences and why it's not safe to use `putenv()`. You cannot rely on `$_SERVER` as it is dependent on the SAPI but in the SAPIs we support, we know it's **always** there. If it wasn't there, nothing would work. You couldn't even access Contao at all.

fritzmg:
I was referring to this section of the article:

> How do we avoid the segmentation fault? Some developers have begun suggesting that you use $_SERVER or $_ENV instead of getenv to read environment variables. This certainly avoids the problem but it’s not as easy as you might think.
> 
> As mentioned above, if you don’t control the server you can’t guarantee $_SERVER and $_ENV will be enabled. $_ENV is disabled by default. It’s unlikely that $_SERVER is disabled but if you use $_SERVER your app won’t work with PHP’s built-in web server.
> 
> Secondly, it’s very difficult to guarantee that all the C libraries you depend on will avoid getenv. For example, the finfo_open calls getenv if you don’t specify the $magic_file. Even $_SERVER and $_ENV call getenv when they are initialized. You > would need to audit all of libc, PHP, every PHP library, and every extension to guarantee getenv wasn’t being called.
> 
> Thirdly, if you are using pthreads the superglobals are empty in the worker thread. The only way to access the main thread’s environment variables is to call getenv.
> 
> A much simpler solution is to avoid calling putenv in worker threads. If you are using putenv to populate environment variables you only need to do that once. Each worker thread will inherit the environment variables of the parent thread, so you don’t need to populate them again. Get the bootstrapping out of the way before you create a thread and you won’t have any issues.

Specifically the last paragraph.

> You cannot rely on `$_SERVER` as it is dependent on the SAPI but in the SAPIs we support, we know it's **always** there. If it wasn't there, nothing would work. You couldn't even access Contao at all.

But that would be application specific and thus not really an argument for an independent library to not use `getenv()`.

Toflar:
We just cannot use `putenv()` by default on a CMS that's used on thousands on different server settings. No way.

fritzmg:
I tend to agree now. It just means that the convenience of `.env` will not work for every library. You'll either have to use true environment variables - or none.

For anyone stumbling over the same: I ended up defining the Firestore client service like this, so that I can still use `.env`:

```yaml
services:
    Google\Cloud\Firestore\FirestoreClient:
        arguments: [{keyFilePath: '%env(GOOGLE_APPLICATION_CREDENTIALS)%', projectId: '%env(GOOGLE_CLOUD_PROJECT)%'}]
```

PR #3291, opened 2021-08-11
Fix the "iflng" and "ifnlng" insert tags

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3290

PR #3295, opened 2021-08-11
Fix the "iflng" and "ifnlng" insert tags

ausi:
Same as #3291 but for Contao 4.12 because the language handling changed in version 4.12.


leofeyer:
Thank you @ausi.

PR #3303, opened 2021-08-12
Warn if the DB server is not running in strict mode

ausi:
Backport of #3151 + #3175 as discussed in https://github.com/contao/contao/pull/3131#issuecomment-891887659


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3310, opened 2021-08-15
Add the "dry-run" and "format=ndjson" options to the migrate command

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3068

As discussed in https://github.com/contao/contao/pull/3068#issuecomment-877072227

Example output for `contao:migrate --dry-run --format=ndjson`

```
{"type":"migration-pending","name":"Contao 4.8.0 Update"}
{"type":"migration-pending","name":"Runonce file: contao\/config\/runonce.php"}
{"type":"schema-pending","commands":["ALTER TABLE tl_files ADD meta BLOB DEFAULT NULL"]}
```

Example output for `contao:migrate --format=ndjson --no-interaction`

```
{"type":"migration-pending","name":"Contao 4.8.0 Update"}
{"type":"migration-pending","name":"Runonce file: contao\/config\/runonce.php"}
{"type":"migration-result","message":"Deleted invalid important part [10,10,50,50] from image \"files\/contaodemo\/media\/content-images\".","isSuccessful":true}
{"type":"migration-result","message":"Executed runonce file: contao\/config\/runonce.php","isSuccessful":true}
{"type":"schema-pending","commands":["ALTER TABLE tl_files ADD meta BLOB DEFAULT NULL"]}
{"type":"schema-execute","command":"ALTER TABLE tl_files ADD meta BLOB DEFAULT NULL"}
{"type":"schema-result","command":"ALTER TABLE tl_files ADD meta BLOB DEFAULT NULL","isSuccessful":true}
```

ndjson format: http://ndjson.org/ (alternative: https://jsonlines.org/ not sure what the correct “standard” is here).
This format is used so that the manager can stream the output (every single line is a valid JSON) of the potentially long running command.

@aschempp is this something you can work with in the manager?

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -78,6 +78,7 @@ class MigrateCommand extends Command
             ->addOption('schema-only', null, InputOption::VALUE_NONE, 'Only update the database schema.')
             ->addOption('migrations-only', null, InputOption::VALUE_NONE, 'Only execute the migrations.')
             ->addOption('dry-run', null, InputOption::VALUE_NONE, 'Show pending migrations and schema updates without executing them.')
+            ->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format (txt, ndjson)', 'txt')
             ->setDescription('Executes migrations and updates the database schema.')
         ;
     }
@@ -85,7 +86,38 @@ class MigrateCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $this->io = new SymfonyStyle($input, $output);
+
+        if ('ndjson' !== $input->getOption('format')) {
+            return $this->executeCommand($input);
+        }
+
+        try {
+            return $this->executeCommand($input);
+        } catch (\Throwable $exception) {
+            $this->writeNdjson('error', [
+                'message' => $exception->getMessage(),
+                'code' => $exception->getCode(),
+                'file' => $exception->getFile(),
+                'line' => $exception->getLine(),
+                'trace' => $exception->getTraceAsString(),
+            ]);
+        }
+
+        return 1;
+    }
+
+    private function executeCommand(InputInterface $input): int
+    {
         $dryRun = (bool) $input->getOption('dry-run');
+        $asJson = 'ndjson' === $input->getOption('format');
+
+        if (!\in_array($input->getOption('format'), ['txt', 'ndjson'], true)) {
+            throw new InvalidOptionException(sprintf('Unsupported format "%s".', $input->getOption('format')));
+        }
+
+        if ($asJson && !$dryRun && $input->isInteractive()) {
+            throw new InvalidOptionException('Use --no-interaction or --dry-run together with --format=ndjson');
+        }
 
         if ($input->getOption('migrations-only')) {
             if ($input->getOption('schema-only')) {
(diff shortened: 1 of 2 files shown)


aschempp:
I'm not sure `ndjson` is useful/necessary at all. As explained in https://github.com/contao/contao/pull/3068#issuecomment-857666263, the Manager would first need a list of pending migrations. If there are migrations pending, we must not run the schema changes, right? AFTER migrations, I would need a list of schema changes if the user should be able to confirm that.

Since the manager _polls_ the task information, streaming the output does not really have any advantage, especially not when asking the user for a confirmation. Streaming would theoretically only be useful when showing the result, but we only "stream" the composer log at the moment.

I have no idea how the schema update interface would look like, but I very much doubt streaming would be useful at all?

ausi:
> I have no idea how the schema update interface would look like

```sh
$ contao:migrate --schema-only --format=ndjson --no-interaction
{"type":"schema-pending","commands":["ALTER TABLE tl_files DROP someExtraCol","ALTER TABLE tl_page DROP someExtraCol"]}
{"type":"schema-execute","command":"ALTER TABLE tl_files DROP someExtraCol"}
{"type":"schema-result","command":"ALTER TABLE tl_files DROP someExtraCol","isSuccessful":true}
{"type":"schema-execute","command":"ALTER TABLE tl_page DROP someExtraCol"}
{"type":"schema-result","command":"ALTER TABLE tl_page DROP someExtraCol","isSuccessful":true}
```

As every single SQL query could potentially take up some time I think it would be useful to show these results “live” to the user.

aschempp:
I was referring to the Manager UI, sorry for not being clear. A streaming interface is just 10x more complicate to me without a real benefit? I mean we never had anything similar in the install tool and it was always just fine?

aschempp:
Maybe to have a better discussion starting point, here's what I would imagine in the UI:

1. go to the migration view
2. see a list of pending migrations
3. execute pending migrations, see task window with console output (similar to on composer updates)
4. return back to migration view
5. see a list of pending SQL updates
6. execute pending SQL updates, see task window with console output

That's the most simple implementation to me, thats why I only implemented a "list" functionality in #3068. I agree this PR might allow to implement something like the (old) search index update in the back end, where you have 100s of lines of SQL updates that are executed sequentially. But I really doubt adding something like that is worth the effort?

ausi:
> A streaming interface is just 10x more complicate to me without a real benefit?

I think it would be a benefit as it would provide better feedback to the user, but I agree that the amout of work is probably way higher. I’d like to keep the ndjson output to be prepared for the future (and to be able to parse partial output of killed processes). Parsing ndjson compared to regular JSON should be straight forward as you can see in the tests.

ausi:
> Maybe to have a better discussion starting point, here's what I would imagine in the UI:

1. go to the migration view
2. see a list of pending migrations => `contao:migrate --format=ndjson --dry-run --migrations-only`
3. execute pending migrations, see task window with console output (similar to on composer updates) => `contao:migrate --format=txt --no-interaction --migrations-only`
4. return back to migration view
5. see a list of pending SQL updates => `contao:migrate --format=ndjson --dry-run --schema-only`
6. execute pending SQL updates, see task window with console output => `contao:migrate --format=txt --no-interaction --schema-only` (including `--with-deletes` if the user enables it in the manager UI I think)

Added the command lines I think we could use for this.

This solution seems very reasonable to me!

Toflar:
LGTM but at the end of the day, I guess it's @aschempp's call.
However, if we can have it simple enough for the Manager (only a few additional lines) and yet more powerful for potential other use cases, I'd go with the proposed ndjson format as well :)

aschempp:
LGTM as well. ndjson isn't really a problem for the Contao Manager. I just have never seen ndjson before and I was unsure if we want to add (and ultimately keep up with) that additional format. I just figured since I've never seen it in the wild it might be a niche standard we might not want to deal with. But ultimately it does not really matter, we can always add regular JSON should there be a need (which I doubt very much).

aschempp:
One more thing following up on the original links. Should it be `ndjson` or `jsonl` (due to two "conflicting" "standards")?

ausi:
> Should it be `ndjson` or `jsonl` (due to two "conflicting" "standards")?

I think we should go with `ndjson` because it is the only one having a “spec” <http://dataprotocols.org/ndjson/> it seems.

For a comparision of the three types (ndjson, jsonl, ldjson) see https://github.com/ndjson/ndjson.github.io/issues/1#issuecomment-109935996

As we are compatible with all three variants, renaming `ndjson` afterwards to something else should be easy.

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #1446, opened 2020-03-01
Determine ptable dynamically

fritzmg:
Currently you have to do something like

```php
// src/Resources/contao/dca/tl_content.php
if (\Contao\Input::get('do') == 'foo')
{
	$GLOBALS['TL_DCA']['tl_content']['config']['ptable'] = 'tl_foo';
}
```

in order to be able to use content elements for your own back end module's table. With this PR the parent table will be determined dynamically according to the back end module configurations.

PR #3311, opened 2021-08-15
Fix version creation for entries with dynamic ptable

fritzmg:
If you try to create a new version for `tl_content` somewhere that is not in the back end in either of the `article` or `page` module, an error will occur. For example if you update content elements somewhere in a custom back end module or in a command etc, and you still want to properly create new versions with

```php
$version = new Versions('tl_content', $id);
$version->initialize();
// …
$version->create();
```

The following error will occur:

```
Exception:
The table name must not be empty

  at vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\DcaLoader.php:50
  at Contao\DcaLoader->__construct('')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Controller.php:1394)
  at Contao\Controller::loadDataContainer('')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Controller.php:1478)
  at Contao\Controller->getParentEntries('', '33')
     (vendor\contao\contao\core-bundle\src\Resources\contao\classes\Versions.php:246)
  at Contao\Versions->create(true)
     (vendor\contao\contao\core-bundle\src\Resources\contao\classes\Versions.php:146)
  at Contao\Versions->initialize()
```

This is actually already fixed in Contao 4.10 and up, because the same fix was already incorporated in [my PR](https://github.com/contao/contao/pull/1446) back then. This PR backports this fix to Contao 4.9 (sans the actual feature for automatic dynamic ptables).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -1448,7 +1448,7 @@ abstract class Controller extends System
 	protected function getParentEntries($strTable, $intId)
 	{
 		// No parent table
-		if (!isset($GLOBALS['TL_DCA'][$strTable]['config']['ptable']))
+		if (empty($GLOBALS['TL_DCA'][$strTable]['config']['ptable']))
 		{
 			return '';
 		}


ausi (approved):

aschempp (approved):

Toflar (approved):

bytehead (approved):

fritzmg:
Btw. we could also fix this function in such a way that it would query the database whether the table has a `ptable` field and use that if the DCA has `dynamicPtable` enabled. Currently this function would not collect parent records for `tl_content`, even though it technically has one. But it may not really be worth the effort.

leofeyer:
Thank you @fritzmg.

PR #3321, opened 2021-08-16
Correctly explode foreign keys in the DCA

### LINKED ISSUE #3318: DCA: Additional dots in foreignKey or list.label.fields causing errors
**Affected version(s)**
Contao 4.11

**Description**

I am getting an error, when I try to use additional dots inside a table reference in foreignKey or list.label.fields.

```
...
  'list' => [
    'fields' => ['member:tl_member.CONCAT(lastname,",",firstname)']
  ],
...
```

Works fine.

```
...
  'list' => [
    'fields' => ['member:tl_member.CONCAT(lastname,".",firstname)']
  ],
...
```

Caused the following error.

`
An exception occurred while executing 'SELECT CONCAT(lastname," FROM tl_member WHERE id='2528' LIMIT 0,1'
`

Greetings
Dennis

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3318
| Docs PR or issue | -

When defining a `foreignKey` for a database relation, or a field label in `list.label.fields` for your DCA, you are able to write the following syntax: `tl_member.CONCAT("firstname", " ", "lastname")` (or `member:tl_member.CONCAT("firstname", " ", "lastname")` for `list.label.fields`). This string is then split in two by using the first occurrence of `.` as the delimiter. However, some places of the `explode()` call were missing the third parameter, causing the string to be split into more than two parts, if there happen to be dots in other places. This PR fixes that by adding the necessary third parameter to the remaining `explode()` calls.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
+++ b/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
@@ -418,7 +418,7 @@ class DcaExtractor extends Controller
 
 					if (isset($config['foreignKey']))
 					{
-						$table = substr($config['foreignKey'], 0, strrpos($config['foreignKey'], '.'));
+						$table = substr($config['foreignKey'], 0, strpos($config['foreignKey'], '.'));
 					}
 
 					$arrRelations[$field] = array_merge(array('table'=>$table, 'field'=>'id'), $config['relation']);
--- a/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
+++ b/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
@@ -418,7 +418,7 @@ class DcaExtractor extends Controller
 
 					if (isset($config['foreignKey']))
 					{
-						$table = substr($config['foreignKey'], 0, strpos($config['foreignKey'], '.'));
+						$table = explode('.', $config['foreignKey'])[0];
 					}
 
 					$arrRelations[$field] = array_merge(array('table'=>$table, 'field'=>'id'), $config['relation']);
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -340,7 +340,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 
 			if (isset($GLOBALS['TL_DCA'][$this->strTable]['fields']['name']['foreignKey']))
 			{
-				list($t, $f) = explode('.', $GLOBALS['TL_DCA'][$this->strTable]['fields']['name']['foreignKey']);
+				list($t, $f) = explode('.', $GLOBALS['TL_DCA'][$this->strTable]['fields']['name']['foreignKey'], 2);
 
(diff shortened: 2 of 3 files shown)


core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -418,7 +418,7 @@ protected function createExtract()
 
 					if (isset($config['foreignKey']))
 					{
-						$table = substr($config['foreignKey'], 0, strrpos($config['foreignKey'], '.'));
+						$table = substr($config['foreignKey'], 0, strpos($config['foreignKey'], '.'));
```
fritzmg: I did not notice any bug with this. However, if you for example use the following DCA:

```php
// contao/dca/tl_comments.php
$GLOBALS['TL_DCA']['tl_comments']['fields']['member']['foreignKey'] = 'tl_member.CONCAT(firstname,".",lastname)';
```

then the `DcaExtractor` would create the following relation:

```
  "member" => array:4 [
    "table" => "tl_member.CONCAT(firstname,""
    "field" => "id"
    "type" => "belongsTo"
    "load" => "lazy"
  ]
```

which seems wrong of course. As with the other places, the _first_ occurrence of a `.` should split the string.

ausi: Maybe use a shorter version here?
```suggestion
						$table = explode('.', $config['foreignKey'])[0];
```

ausi (dismissed):

ausi (approved):

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -418,7 +418,7 @@ protected function createExtract()
 
 					if (isset($config['foreignKey']))
 					{
-						$table = substr($config['foreignKey'], 0, strrpos($config['foreignKey'], '.'));
+						$table = explode('.', $config['foreignKey'])[0];
```
leofeyer, line 421: This is a behavior change. If a foreign key is `foo.bar.baz`, the table name was previously `foo.bar` and now it is only `foo`. What is the motivation for this change?

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
fritzmg: See the comment here: https://github.com/contao/contao/pull/3321#discussion_r689708367

Also you cannot create a DCA called `tl_foo.bar` for example, as this would lead to the following exception:

```
Execute database migrations
---------------------------

 * CREATE TABLE tl_foo.bar (id INT UNSIGNED AUTO_INCREMENT NOT NULL, tstamp INT UNSIGNED DEFAULT 0 NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB ROW_FORMAT = DYNAMIC......FAILED


 [OK] Executed 0 SQL queries.



 [ERROR] An exception occurred while executing 'CREATE TABLE tl_foo.bar (id INT UNSIGNED AUTO_INCREMENT
         NOT NULL, tstamp INT UNSIGNED DEFAULT 0 NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4
         COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB ROW_FORMAT = DYNAMIC':

SQLSTATE[42000]: Syntax
         error or access violation: 1049 Unknown database 'tl_foo'
```

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3325, opened 2021-08-17
Truncate tl_search* tables independently from indexer service

bytehead:
Follow up for #3283:
Purging search tables after disabling the indexer is now only possible directly via database.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Automator.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Automator.php
@@ -35,16 +35,11 @@ class Automator extends System
 	 */
 	public function purgeSearchTables()
 	{
-		// The search indexer is disabled
-		if (!System::getContainer()->has('contao.search.indexer'))
-		{
-			return;
-		}
+		$objDatabase = Database::getInstance();
 
-		$searchIndexer = System::getContainer()->get('contao.search.indexer');
-
-		// Clear the index
-		$searchIndexer->clear();
+		// Truncate the tables
+		$objDatabase->execute("TRUNCATE TABLE tl_search");
+		$objDatabase->execute("TRUNCATE TABLE tl_search_index");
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

leofeyer:
Let's see what @Toflar thinks about this. I recall he had changed this on purpose.

Toflar:
Yeah that's definitely wrong. The search indexer service can be anything. It can be ES or MeiliSearch.

bytehead:
I see, missed that 😬

aschempp:
But then clearing the tables does not make sense anyway? Should we check the instance of the search indexer?

bytehead:
It only does make sense in case of the default indexer then. Maybe to discuss in #3326?

Toflar:
I have no clue what you guys are discussing, sorry. Every indexer has to implement our `IndexerInterface` and we call `clear()` on it. The indexers are then responsible to clear the stuff which is all implemented: https://github.com/contao/contao/blob/af45f19fa34e833ddec862fd0ae7dd05ad6ad801/core-bundle/src/Search/Indexer/DefaultIndexer.php#L127

So no idea what you're trying to do.

fritzmg:
Unless someone overrides the DCA, the tables `tl_search`, `tl_search_index` and `tl_search_term` always exist, regardless of the `IndexerInterface` implementation. But if you switch the indexer implementation somewhere down the road (or disable it), I think it's reasonable to assume that you will clear these tables yourself manually, if you want to.

bytehead:
It's just confusing when you disable the search at some point and you can't clear the tables anymore in the backend. 🤷‍♂️

Toflar:
I don't think that's confusing. If you disable a service, any functionality with it becomes deactivated.
You don't need to clear it. If you want to improve anything here I guess that would be to only add the search tables if it is enabled. That way they would get dropped after you disabled it :)

PR #3329, opened 2021-08-17
Make sure X-Forwarded-Host is trusted if trusted hosts are configured

Toflar:
This PR contains two fixes:

1. `Request::HEADER_X_FORWARDED_ALL` is deprecated in favor of `Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_HOST`
2. While fixing the deprecation I noticed that our previous `Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST` did not make a lot of sense together with `$_SERVER['TRUSTED_HOSTS']`. You could configure the trusted hosts using `$_SERVER['TRUSTED_HOSTS']` but it would've never worked as we excluded `X-Forwarded-Host` from the trusted headers. Now that's automatically trusted if you configure `$_SERVER['TRUSTED_HOSTS']` which then makes sense.

### DIFF
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -252,13 +252,18 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
     {
         self::loadEnv($projectDir, 'jwt');
 
-        // See https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/public/index.php
-        if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? null) {
-            Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
-        }
-
         if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? null) {
             Request::setTrustedHosts(explode(',', $trustedHosts));
+        }
+
+        if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? null) {
+            $trustedHeaderSet = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO;
+
+            if ($trustedHosts) {
+                $trustedHeaderSet |= Request::HEADER_X_FORWARDED_HOST;
+            }
+
+            Request::setTrustedProxies(explode(',', $trustedProxies), $trustedHeaderSet);
         }
 
         Request::enableHttpMethodParameterOverride();
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -259,6 +259,7 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
         if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? null) {
             $trustedHeaderSet = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO;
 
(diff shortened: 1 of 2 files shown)


leofeyer:
> 2. While fixing the deprecation I noticed that our previous `Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST` did not make a lot of sense together with `$_SERVER['TRUSTED_HOSTS']`.

We have copied the code from [Symfony](https://github.com/symfony/recipes/commit/875018cd2dc74aed699a9634060f29e3539a01ec#diff-c7178c4268edcdc10b5fd504db864941cfc703782ef1ecc5598bfa46e7a77646R27-R33), therefore I guess it did make sense at some point.

leofeyer:
Did you make sure that the changes work with Symfony 4.4?

Toflar:
Yes, the constants have existed ever since (https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/HttpFoundation/Request.php#L42-L45), just the "all" one was deprecated. Guess they want you to be specific :)

manager-bundle/src/HttpKernel/ContaoKernel.php
```diff
@@ -252,15 +252,20 @@ public static function fromRequest(string $projectDir, Request $request): HttpKe
...
-
         if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? null) {
             Request::setTrustedHosts(explode(',', $trustedHosts));
         }
 
+        if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? null) {
+            $trustedHeaderSet = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO;
+
+            if ($trustedHosts) {
+                $trustedHeaderSet |= Request::HEADER_X_FORWARDED_HOST;
```
leofeyer, line 264: Why are we doing this and [Symfony is not](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/public/index.php#L15-L21)?

Toflar: Because I think we can be smarter? :D Or in other words, tell me how am I supposed to configure my `.env` now to trust a certain host? You can't because it's never read from `X-Forwarded-Host`.
If you want to trust a certain host by setting `TRUSTED_HOSTS` in your `.env` I think you want to read it from there?

leofeyer:
As discussed in Mumble on August 26th, we want to merge the first fix and check with Symfony if the second fix is correct and should be added in Symfony as well.

ausi (dismissed):

manager-bundle/tests/HttpKernel/ContaoKernelTest.php
```diff
@@ -312,11 +312,13 @@ public function testSetsRequestTrustedHostsFromEnvVars(): void
     {
         $this->assertSame([], Request::getTrustedHosts());
 
+        $_SERVER['TRUSTED_PROXIES'] = '1.1.1.1,2.2.2.2';
         $_SERVER['TRUSTED_HOSTS'] = '1.1.1.1,2.2.2.2';
```
ausi:
```suggestion
        $_SERVER['TRUSTED_HOSTS'] = '1.1.1.1,2.2.2.2,example.com';
```
This would maybe be helpful for better understanding what “hosts” means in this context?

Toflar:
So I checked and it's irrelevant for us, what Symfony does.
Symfony doesn't base it on environment variables in the latest versions anymore due to changes with the Symfony Runtime component.
So their values are coming from the container (https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/HttpKernel/Kernel.php#L790-L796) now and the default values cannot be dependent on each other. So there's no point to open an issue or ask anything there because for Symfony, you'll always have to configure both values. They cannot automatically adust their default like we can.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3339, opened 2021-08-18
Allow Twig v3

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

We should loosen our type constraints so that installing Twig in version 3 is possible.

IMHO no surprises with that, see https://github.com/twigphp/Twig/blob/3.x/CHANGELOG


m-vo:
As the `IncludeTokenParser` was made internal in v3, I pulled in the implementation into our `DynamicIncludeTokenParser` (similar to the `ExtendsTokenParser` in `DynamicExtendsTokenParser`) and added tests for it.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3343, opened 2021-08-19
Make the ContaoFilesystemLoader compatible with Symfony 4.4

### LINKED ISSUE #3342: Argument #1 ($name) must be of type string, Symfony\Bundle\FrameworkBundle\Templating\TemplateReference given
**Affected version(s)**

Contao 4.12-RC3 / PHP 8.0.9

**Description**

```
$ php bin/console cache:warmup --env=dev

// Warming up the cache for the dev environment with debug true

In ContaoFilesystemLoader.php line 404:
                                                                                                                                                                                                                                                                                                           
  [Symfony\Component\Debug\Exception\FatalThrowableError]                                                                                                                                                                                                                                                  
  Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader::getThemeTemplateName(): Argument #1 ($name) must be of type string, Symfony\Bundle\FrameworkBundle\Templating\TemplateReference given, called in /vendor/contao/core-bundle/src/Twig/Loader  
  /ContaoFilesystemLoader.php on line 256                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                           

Exception trace:
  at /vendor/contao/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php:404
 Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader->getThemeTemplateName() at /vendor/contao/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php:256
 Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader->exists() at /vendor/twig/twig/src/Loader/ChainLoader.php:87
 Twig\Loader\ChainLoader->getCacheKey() at /vendor/twig/twig/src/Environment.php:299
 Twig\Environment->getTemplateClass() at /vendor/twig/twig/src/Environment.php:381
 Twig\Environment->loadTemplate() at /vendor/twig/twig/src/Environment.php:359
 Twig\Environment->load() at /vendor/symfony/twig-bundle/CacheWarmer/TemplateCacheCacheWarmer.php:76
 Symfony\Bundle\TwigBundle\CacheWarmer\TemplateCacheCacheWarmer->warmUp() at /vendor/symfony/http-kernel/CacheWarmer/CacheWarmerAggregate.php:97
 Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerAggregate->warmUp() at /vendor/symfony/framework-bundle/Command/CacheWarmupCommand.php:80
 Symfony\Bundle\FrameworkBundle\Command\CacheWarmupCommand->execute() at /vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /vendor/symfony/console/Application.php:1027
 Symfony\Component\Console\Application->doRunCommand() at /vendor/symfony/framework-bundle/Console/Application.php:97
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /vendor/symfony/console/Application.php:273
 Symfony\Component\Console\Application->doRun() at /vendor/symfony/framework-bundle/Console/Application.php:83
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /vendor/symfony/console/Application.php:149
 Symfony\Component\Console\Application->run() at /bin/console:46

```

bytehead:
Try to fix #3342.

Edit: Simply cast to string (@m-vo pointed me to https://github.com/twigphp/Twig/pull/1183#issuecomment-25272767)

### DIFF
--- a/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
+++ b/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
@@ -16,6 +16,7 @@ use Contao\CoreBundle\Twig\ContaoTwigUtil;
 use Contao\CoreBundle\Twig\Inheritance\TemplateHierarchyInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Contracts\Service\ResetInterface;
+use Twig\Error\Error;
 use Twig\Error\LoaderError;
 use Twig\Loader\FilesystemLoader;
 use Twig\Source;
@@ -185,7 +186,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
      */
     public function getCacheKey($name): string
     {
-        $templateName = $this->getThemeTemplateName($name) ?? $name;
+        $templateName = $this->getThemeTemplateName((string) $name) ?? $name;
 
         return parent::getCacheKey($templateName);
     }
@@ -202,7 +203,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
      */
     public function getSourceContext($name): Source
     {
-        $templateName = $this->getThemeTemplateName($name) ?? $name;
+        $templateName = $this->getThemeTemplateName((string) $name) ?? $name;
         $source = parent::getSourceContext($templateName);
 
         // The Contao PHP templates will still be rendered by the Contao
--- a/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
+++ b/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
@@ -15,7 +15,7 @@ namespace Contao\CoreBundle\Twig\Loader;
 use Contao\CoreBundle\Twig\ContaoTwigUtil;
 use Contao\CoreBundle\Twig\Inheritance\TemplateHierarchyInterface;
 use Psr\Cache\CacheItemPoolInterface;
-use Symfony\Bundle\FrameworkBundle\Templating\TemplateReference;
+use Symfony\Component\Templating\TemplateReferenceInterface;
 use Symfony\Contracts\Service\ResetInterface;
 use Twig\Error\LoaderError;
 use Twig\Loader\FilesystemLoader;
@@ -252,11 +252,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
             return true;
         }
 
-        if ($name instanceof TemplateReference) {
-            $name = $name->getLogicalName();
-        }
-
-        if (null !== ($themeTemplate = $this->getThemeTemplateName($name))) {
+        if (null !== ($themeTemplate = $this->getThemeTemplateName($name instanceof TemplateReferenceInterface ? $name->getLogicalName() : $name))) {
             return parent::exists($themeTemplate);
         }
 
--- a/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
+++ b/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
@@ -185,6 +185,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
      */
     public function getCacheKey($name): string
     {
+        // string cast to be removed with Symfony 5 (see #3343)
         $templateName = $this->getThemeTemplateName((string) $name) ?? $name;
 
         return parent::getCacheKey($templateName);
@@ -202,6 +203,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
      */
     public function getSourceContext($name): Source
     {
+        // string cast to be removed with Symfony 5 (see #3343)
         $templateName = $this->getThemeTemplateName((string) $name) ?? $name;
         $source = parent::getSourceContext($templateName);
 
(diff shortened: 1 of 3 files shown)


core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -251,6 +252,10 @@ public function exists($name): bool
             return true;
         }
 
+        if ($name instanceof TemplateReference) {
```
m-vo: The current version of the `LoaderInterface` is strictly typed with `string $name` ([ref](https://github.com/twigphp/Twig/blob/3.x/src/Loader/LoaderInterface.php#L48)).

Could you find out your used Twig version?

bytehead: v2.14.6

m-vo: IMHO `string` is the correct type even there:
https://github.com/twigphp/Twig/blob/27e5cf2b05e3744accf39d4c68a3235d9966d260/src/Loader/LoaderInterface.php#L62

ausi (dismissed):

m-vo:
We probably need to add the string cast to the other places where `$name` gets passed as well, do we?

https://github.com/twigphp/Twig/blob/3.x/src/Loader/LoaderInterface.php#L29-L48

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -251,7 +251,7 @@ public function exists($name): bool
             return true;
         }
 
-        if (null !== ($themeTemplate = $this->getThemeTemplateName($name))) {
+        if (null !== ($themeTemplate = $this->getThemeTemplateName((string) $name))) {
```
leofeyer: I am not sure about this at all. According to the phpDoc, `$name` cannot be anything but a string, so why does the underlying error occur in the first place?

leofeyer (changes requested):

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
m-vo: I think it's a Symfony bug that cannot really be fixed without breaking BC…

https://github.com/symfony/symfony/pull/26243#issuecomment-367304955
https://github.com/symfony/symfony/issues/25712
https://github.com/symfony/symfony/pull/15970

I could not find the parts where the old `TemplateReference` is used, yet, but according to @bytehead there seem quite a few Symfony bundles affected.

m-vo: And yes: I fully agree that this is a lot of nonsense. The interface should be the only thing that matters.

leofeyer: Can we explicitly ignore non-string inputs?

```php
if (!\is_string($name)) {
    return false;
}
```

m-vo: IMO we could, yes.

ausi:
```suggestion
        // is_string check to be removed with Symfony 5
        if (\is_string($name) && null !== ($themeTemplate = $this->getThemeTemplateName($name))) {
```

m-vo: This check should probably go in L250.

leofeyer: IMHO we should not make assumptions for the parent class and only check the type in line 255.

m-vo: Our loader will probably never yield anything but `false` if a `TemplateReference` was input. :thinking: 

Btw.: if we go down this route, we also need to adjust `getSourceContext()`, `getCacheKey()` and `isFresh()`. They all call `normalizeName()` in the original `FilesystemLoader` that has a `(string)` cast.

bytehead: Added in b1bbd59cf.

m-vo (dismissed):

ausi (dismissed):

leofeyer (changes requested):
Sorry for the fuss but I just realized that the problem is caused by our own private method. I therefore suggest that we remove the type hint from the `getThemeTemplateName()` method  and return `null` if `$name` is not a string.

bytehead:
> Sorry for the fuss but I just realized that the problem is caused by our own private method. I therefore suggest that we remove the type hint from the getThemeTemplateName() method and return null if $name is not a string.

I've always knew it is because of the private method but @m-vo's intention was probably to follow the interface.

bytehead:
Like in 7111c8d9a?

leofeyer:
Yes, exactly. And you could add a `@todo` to the phpDoc comment that reminds us to add the type cast again in Contao 4.13.

bytehead:
Done in 1401c0911.

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #3346, opened 2021-08-19
Fix versioning condition in DC_Folder edit action

### LINKED ISSUE #3301: Creating a new folder in dev env, PHP 8 causes an ErrorException
**Affected version(s)**

 Contao 4.11 (PHP 8.0.9)

**Description**
When **creating a new folde**r in the backend in **APP_ENV=dev** a warning occurs:
```Attempt to read property "id" on null```
This leads to an ErrorException in vendor/contao/core-bundle/src/Resources/contao/drivers/DC_Folder.php (line 1324)

**Solution**
Check, if $objModel is not null:
```$objVersions = new Versions($this->strTable, $objModel->id ?? null);```

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3301

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -1307,30 +1307,30 @@ class DC_Folder extends DataContainer implements \listable, \editable
 				}
 
 				$this->objActiveRecord = $objModel;
-			}
 
-			$this->blnCreateNewVersion = false;
+				$this->blnCreateNewVersion = false;
 
-			/** @var FilesModel $objModel */
-			$objVersions = new Versions($this->strTable, $objModel->id);
+				/** @var FilesModel $objModel */
+				$objVersions = new Versions($this->strTable, $objModel->id);
 
-			if (!$GLOBALS['TL_DCA'][$this->strTable]['config']['hideVersionMenu'])
-			{
-				// Compare versions
-				if (Input::get('versions'))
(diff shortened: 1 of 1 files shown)


bytehead (approved):

Toflar (approved):

bytehead:
LGTM (tested in Contao 4.11.x-dev with patch / PHP 8.0.9)

bytehead:
I'm not sure if this should go against 4.11 as 4.9 should not be run with warnings enabled under PHP 8?

ausi:
> I'm not sure if this should go against 4.11 as 4.9 should not be run with warnings enabled under PHP 8?

IMO this was not only a missing `null` check but the wrong logic. But I would be fine merging it in 4.11 only too.

fritzmg:
Since `$objVersions` is now not always initialized, are we sure lines 

https://github.com/contao/contao/blob/866e4c31ab16f4d0e1169da5283872c02fca0b1a/core-bundle/src/Resources/contao/drivers/DC_Folder.php#L1514

and 

https://github.com/contao/contao/blob/866e4c31ab16f4d0e1169da5283872c02fca0b1a/core-bundle/src/Resources/contao/drivers/DC_Folder.php#L1547

won't get executed now?

ausi:
Both check for `$objModel !== null` so this should be fine.

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #3353, opened 2021-08-20
Fix dynamic debug mode for templates

### LINKED ISSUE #3351: Cannot disable debug output for templates anymore
**Affected version(s)**

Contao 4.12

**Description**

In https://github.com/contao/contao/commit/4e0c1204af9f39f203eb691f4fd0946e3bdc4833 `TemplateInheritance::inherit` was changed, so that it uses `kernel.debug` instead of `Config::get('debugMode')`. However, as noticed via https://github.com/contao/contao/issues/3349, this now makes it impossible to dynamically disable the debug output when parsing a template.

This will certainly break a lot of extensions in the `dev` environment that rely on this functionality.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3351

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -504,9 +504,6 @@ class tl_templates extends Backend
 		$objTemplate->title = StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['showDifferences']);
 		$objTemplate->charset = System::getContainer()->getParameter('kernel.charset');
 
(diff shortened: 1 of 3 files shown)


m-vo (approved):
Nice side effect with this solution: We can now disable the debug output when rendering via Twig (proxy) and implement some sort of "debug output node visitor". Previously this would yield duplicate wrappers when extending PHP templates.

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #3359, opened 2021-08-20
Fix the media element migration

### LINKED ISSUE #2984: Contao 4.8 migration removing custom j_mediaelement from layout
**Affected version(s)**
Contao 4.9+

**Description**
With Contao 4.8 the `j_mediaelement` was removed.
In a customer-installation I had to add the `j_mediaelement` back again.

To do this, I manually added the file `j_mediaelement.html5` in the templates folder and then activated it in the layout.

Unfortunately, every time the migration is called (console or install-tool), the Contao 4.8 migration is executed and the script template is removed from the layout again.

See:
https://github.com/contao/contao/blob/8eb97ae0fcb8d611be318417e273c5e427aa082e/core-bundle/src/Migration/Version408/Version480Update.php#L151

As a workaround, I have now simply renamed my template and given it a custom name.

**Question**
Is this a bug and needs to be fixed or have I simply made an unsupported use-case?

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2984

### DIFF
--- a/core-bundle/src/Migration/Version408/Version480Update.php
+++ b/core-bundle/src/Migration/Version408/Version480Update.php
@@ -149,12 +149,7 @@ class Version480Update extends AbstractMigration
         $controller = $this->framework->getAdapter(Controller::class);
 
         foreach (['jquery' => 'j_mediaelement', 'scripts' => 'js_mediaelement'] as $column => $templateName) {
-            if (
-                \array_key_exists(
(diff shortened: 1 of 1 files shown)


fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3360, opened 2021-08-21
Use intl for text direction

fritzmg:
Currently the text-direction information for a language is stored in the Contao translations themselves. This is a problem however if the used front end language is not an installed language - or when a translation gets removed from Contao because it is not sufficiently translated anymore. For example in Contao 4.12 Farsi got removed (again) and then the front end will suddenly display the text left-to-right instead of right-to-left. See https://github.com/contao/core-bundle/commit/4073033679391d66c67aa08b9892da6bd8b70c00#commitcomment-55223764 for instance.

PHP intl already provides this information so I think we should not store this information in our own translations and rather extract it from intl.

This PR uses ~a lightweight library~ PHP intl to extract the direction for the current language. It also removes the `MSC.textDirection` translation "label" and adds it again in `System::loadLanguageFile` for BC (in case anyone else still accesses `MSC.textDirection`).

As I consider this a bugfix, I have implemented this for Contao 4.9.

/cc @hamidabbaszadeh

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -459,7 +459,6 @@ abstract class System
 		}
 
 		$strCacheKey = $strLanguage;
-		$blnLoaded = isset(static::$arrLanguageFiles[$strName][$strLanguage]);
 
 		// Make sure the language exists
 		if ($strLanguage != 'en' && !static::isInstalledLanguage($strLanguage))
@@ -488,12 +487,6 @@ abstract class System
 		// Prepare the XLIFF loader
 		$xlfLoader = new XliffFileLoader(static::getContainer()->getParameter('kernel.project_dir'), true);
 		$strCacheDir = static::getContainer()->getParameter('kernel.cache_dir');
-
-		// Always unset MSC.textDirection before loading the language (see #3360)
-		if ('default' === $strName)
-		{
-			unset($GLOBALS['TL_LANG']['MSC']['textDirection']);
-		}
 
 		// Load the language(s)
 		foreach ($arrCreateLangs as $strCreateLang)
@@ -531,14 +524,7 @@ abstract class System
 		// Set MSC.textDirection (see #3360)
 		if ('default' === $strName)
 		{
-			if (empty($GLOBALS['TL_LANG']['MSC']['textDirection']))
-			{
-				$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
-			}
-			elseif (!$blnLoaded)
-			{
-				@trigger_error('Using "MSC.textDirection" to set the text direction will no longer work in Contao 5.0. To override the text direction create a custom "fe_page" template instead.', E_USER_DEPRECATED);
-			}
+			$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
 		}
 
 		// HOOK: allow to load custom labels
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -18,6 +18,7 @@ use Contao\Database\Updater;
 use League\Uri\Components\Query;
 use Patchwork\Utf8;
 use Psr\Log\LogLevel;
+use Salarmehr\Cosmopolitan\Cosmo;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
@@ -519,6 +520,12 @@ abstract class System
 					}
 				}
 			}
+		}
+
+		// Set MSC.textDirection
+		if ('default' === $strName)
+		{
+			$GLOBALS['TL_LANG']['MSC']['textDirection'] = Cosmo::create($strLanguage)->direction();
 		}
 
 		// HOOK: allow to load custom labels
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -522,7 +522,7 @@ abstract class System
 			}
 		}
 
-		// Set MSC.textDirection
+		// Set MSC.textDirection (see #3360)
 		if ('default' === $strName)
 		{
 			$GLOBALS['TL_LANG']['MSC']['textDirection'] = Cosmo::create($strLanguage)->direction();
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -525,7 +525,7 @@ abstract class System
 		// Set MSC.textDirection (see #3360)
 		if ('default' === $strName && empty($GLOBALS['TL_LANG']['MSC']['textDirection']))
 		{
-			$GLOBALS['TL_LANG']['MSC']['textDirection'] = Cosmo::create($strLanguage)->direction();
+			$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
 		}
 
 		// HOOK: allow to load custom labels
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -523,7 +523,7 @@ abstract class System
 		}
 
 		// Set MSC.textDirection (see #3360)
-		if ('default' === $strName)
+		if ('default' === $strName && empty($GLOBALS['TL_LANG']['MSC']['textDirection']))
 		{
 			$GLOBALS['TL_LANG']['MSC']['textDirection'] = Cosmo::create($strLanguage)->direction();
 		}
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -528,7 +528,7 @@ abstract class System
 			{
 				$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
 			}
-			else
+			elseif (!isset(static::$arrLanguageFiles[$strName][$strLanguage]))
 			{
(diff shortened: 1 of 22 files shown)


core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -176,11 +177,8 @@ protected function prepare($objPage)
 
 		$this->Template->sections = $arrCustomSections;
 
-		// Mark RTL languages (see #7171)
-		if ($GLOBALS['TL_LANG']['MSC']['textDirection'] == 'rtl')
-		{
-			$this->Template->isRTL = true;
-		}
+		// Mark RTL languages (see #7171, #3360)
+		$this->Template->isRTL = 'rtl' === ($GLOBALS['TL_LANG']['MSC']['textDirection'] ?? Cosmo::create($GLOBALS['TL_LANGUAGE'])->direction());
```
fritzmg: _Note:_ Intl is used as a fallback, in case anyone is forcing the `MSC.textDirection` via `contao/languages/…/default.php`.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -521,6 +522,12 @@ public static function loadLanguageFile($strName, $strLanguage=null, $blnNoCache
 			}
 		}
 
+		// Set MSC.textDirection (see #3360)
+		if ('default' === $strName && empty($GLOBALS['TL_LANG']['MSC']['textDirection']))
+		{
+			$GLOBALS['TL_LANG']['MSC']['textDirection'] = Cosmo::create($strLanguage)->direction();
```
ausi:
```suggestion
			$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
```

I would prefer not using a third party library for that use case.

core-bundle/src/Resources/contao/pages/PageRegular.php
ausi:
```suggestion
		$this->Template->isRTL = 'rtl' === ($GLOBALS['TL_LANG']['MSC']['textDirection'] ?? ((\ResourceBundle::create($GLOBALS['TL_LANGUAGE'], 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr'));
```

ausi (changes requested):

core-bundle/src/Resources/contao/library/Contao/System.php
fritzmg: Agreed, though the `create` and `direction` methods of the library also have error handling which would then be missing here.

fritzmg:
I have also added some deprecations. @ausi wdyt?

We can't really provide deprecation warnings for accessing `$GLOBALS['TL_LANG']['MSC']['textDirection']` directly though.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -521,6 +521,19 @@ public static function loadLanguageFile($strName, $strLanguage=null, $blnNoCache
...
+			if (empty($GLOBALS['TL_LANG']['MSC']['textDirection']))
+			{
+				$GLOBALS['TL_LANG']['MSC']['textDirection'] = (\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';
+			}
+			else
+			{
+				@trigger_error('Using "MSC.textDirection" to set the text direction will no longer work in Contao 5.0. To override the text direction create a custom "fe_page" template instead.', E_USER_DEPRECATED);
+			}
+		}
+
```
ausi: I think this would trigger the deprecation now if two different languages get loaded one after the other as the first one sets `$GLOBALS['TL_LANG']['MSC']['textDirection']`.

We should only trigger the deprecation if `empty($GLOBALS['TL_LANG']['MSC']['textDirection'])` is true in the beginning of the `loadLanuageFile` method I think.

fritzmg: It needs to be triggered after the language files are actually loaded, otherwise it will always be empty. 

When another language gets loaded, the whole `$GLOBALS['TL_LANG']` array is replaced anyway, so this should be fine. Also I have updated the code so that the deprecation is only triggered on first load of the language, i.e. when it is not cached yet. Otherwise `$GLOBALS['TL_LANG']['MSC']['textDirection']` will not empty, since it's already loaded for the current language.

fritzmg: So for example:

```php
System::loadLanguageFile('default', 'fa'); // MSC.textDirection is now available
System::loadLanguageFile('default', 'fa', true); // no deprecation triggered now
```

fritzmg: Oh wait, now the deprecation is never triggered 🙈

ausi: > When another language gets loaded, the whole `$GLOBALS['TL_LANG']` array is replaced anyway

I don’t think it does. `$GLOBALS['TL_LANG']` never gets cleared and the language files write to the label keys directly.

fritzmg: Yes you are right. But then the only way to test whether `MSC.textDirection` is set by a custom translation is to always unset `MSC.textDirection` first before loading the language. Which is fine since we will always set it afterwards anyway. I will adjust the PR accordingly.

fritzmg: Adjusted in 25d394004

ausi (dismissed):
Besides the failing functional tests. ✅

fritzmg:
> Besides the failing functional tests. ✅

Hm, I think this is because `MSC.textDirection` is still set for all the other languages besides English. Should I remove the translation for all languages directly in this PR instead? Usually these translations would only get removed once @leofeyer updates the translations.

ausi (dismissed):
The PHP 8.1 error `Undefined global variable $TL_LANG` is a bug in PHP itself I think.

fritzmg:
Bug reported here: https://bugs.php.net/bug.php?id=81377

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -488,6 +489,12 @@ public static function loadLanguageFile($strName, $strLanguage=null, $blnNoCache
 		$xlfLoader = new XliffFileLoader(static::getContainer()->getParameter('kernel.project_dir'), true);
 		$strCacheDir = static::getContainer()->getParameter('kernel.cache_dir');
 
+		// Always unset MSC.textDirection before loading the language (see #3360)
```
leofeyer: Why?

leofeyer (changes requested):
Please only adjust the source language file (en/default.xlf). The rest will be adjusted automatically through Transifex.

fritzmg:
> Please only adjust the source language file (en/default.xlf). The rest will be adjusted automatically through Transifex.

As previously mentioned, this will break the unit tests, since the unit tests will still then have `MSC.textDirection` set for all languages other than English. If you we want to merge this with passing unit tests, then all `MSC.textDirection` instances need to be removed in this PR, not just from the fallback language (`en`).

core-bundle/src/Resources/contao/library/Contao/System.php
fritzmg: Because otherwise `MSC.textDirection` will already be populated by a previous `System::loadLanguageFile` call. e.g. when you have a German front end, `System::loadLanguageFile` will be called for `en` (always, because it's the fallback language) and then again for `de`. `MSC.textDirection` will at that point be filled with the value from `en` and thus checking whether `MSC.textDirection` had been set via a language file will not work.

leofeyer:
Why does this have to be so complicated? We do not have to provide BC for `$GLOBALS['TL_LANG']['MSC']['textDirection']`, because it is very unlikely that someone has used the setting to tag a LTR language as RTL or vice versa. We should just rely on Intl and always set `$GLOBALS['TL_LANG']['MSC']['textDirection']` to `(\ResourceBundle::create($strLanguage, 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr';` IMHO.

fritzmg:
Whether this is needed or not I cannot say, as I have no experiences with such use-cases. I am just thinking of those who _might_ 😉. I can remove the BC part of the code again, if you do not think it is worth it. For @hamidabbaszadeh 's case it would not matter for example, as he wants the text direction to be what it is supposed to be anyway (and not have to force it).

leofeyer:
Exactly. And I don't think that anyone would ever want something else.

fritzmg:
Done. Do you want the deprecation notice in the translator also removed?

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -176,11 +176,8 @@ protected function prepare($objPage)
 
 		$this->Template->sections = $arrCustomSections;
 
-		// Mark RTL languages (see #7171)
-		if ($GLOBALS['TL_LANG']['MSC']['textDirection'] == 'rtl')
-		{
-			$this->Template->isRTL = true;
-		}
+		// Mark RTL languages (see #7171, #3360)
+		$this->Template->isRTL = 'rtl' === ((\ResourceBundle::create($GLOBALS['TL_LANGUAGE'], 'ICUDATA', true)['layout']['characters'] ?? null) === 'right-to-left' ? 'rtl' : 'ltr');
```
leofeyer:
```suggestion
		// Mark RTL languages (see #7171, #3360)
 		if ((\ResourceBundle::create($GLOBALS['TL_LANGUAGE'], 'ICUDATA', true)['layout']['characters'] ?? null) == 'right-to-left')
 		{
 			$this->Template->isRTL = true;
 		}
```

core-bundle/src/Translation/Translator.php
```diff
@@ -58,6 +58,10 @@ public function trans($id, array $parameters = [], $domain = null, $locale = nul
             $locale = $this->translator->getLocale();
         }
 
+        if ('MSC.textDirection' === $id && 'contao_default' === $domain) {
+            @trigger_error('Using "MSC.textDirection" to retrieve the locale\'s text direction will no longer work in Contao 5.0. Use the PHP intl library instead.', E_USER_DEPRECATED);
+        }
+
```
leofeyer:
```suggestion
```

leofeyer (changes requested):

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Accessing `$GLOBALS['TL_LANGUAGE']` here seems odd? Why not the request locale?

fritzmg: Actually we already have the appropriate locale available in a variable at this point. Changed in c48467c.

core-bundle/src/Translation/Translator.php
fritzmg: Removed in 79d8e3e

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3362, opened 2021-08-22
Allow extending both front end and back end templates with Twig

### LINKED ISSUE #3361: Fix extending back end templates with Twig
**Affected version(s)**

4.12

**Description**

Extending a back end PHP template with Twig currently produces the following error:

```
An exception has been thrown during the rendering of a template ("getLocaleString is not set or not a callable") in "@Contao_ContaoCoreBundle/be_main.html5".
```

This happens because we're rendering the proxy/surrogate via the `FrontendTemplate` class but also need the methods from the  `BackendTemplate` class. This is already fixed in the (currently draft) PR #3180 where I extracted the template methods into traits and am using both in the proxy.

We should probably extract the fixes as a new PR.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3361 
| Docs PR or issue | -

This PR extracts the methods in `FrontendTemplate` and `BackendTemplate` that are supposed to be used from with the templates themselves into traits. This way both can be used when rendering via the Twig template proxy.


ausi (dismissed):

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -172,7 +174,10 @@ public function renderLegacyTemplate(string $name, array $blocks, array $context
     {
         $template = Path::getFilenameWithoutExtension($name);
 
-        $partialTemplate = new class($template) extends FrontendTemplate {
+        $partialTemplate = new class($template) extends Template {
+            use FrontendTemplateTrait;
+            use BackendTemplateTrait;
```
leofeyer, line 179: I don't like that we are mixing back end and front end methods here. Can we improve this somehow (e.g. by checking the template prefix for `be_`)?

m-vo: I don't think you can reliably find out the scope by the name. This might work for the core, but not in general.

leofeyer: That was just an example. We could also add a `renderLegacyBackendTemplate()` and a `renderLegacyFrontendTemplate()` method to prevent mixing up the contexts.

m-vo: Not sure I understand. The template files themselves don't say anything about their scope - how would the proxy code know which one to call?

ausi: AFAIK we would need to know if the template is a backend or frontend template at compile time to be able to call the correct function in https://github.com/contao/contao/blob/7bd09fae31188fb3d3a38c36e527e0ba5f6dfbf3/core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php#L32-L40

I think extending from both traits is the best solution here.

m-vo: > AFAIK we would need to know if the template is a backend or frontend template at compile time

Even dynamically wouldn't work btw because the context doesn't tell you anything about the template class it was used/intended to use with.

core-bundle/src/Resources/contao/classes/FrontendTemplateTrait.php
```diff
@@ -0,0 +1,222 @@
...
+ * @property string  $keywords
+ * @property string  $content
+ * @property array   $sections
+ * @property array   $positions
+ * @property array   $matches
+ * @method   string  getTemplate(string $strTemplate)
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ *
+ * @internal
```
leofeyer, line 24: Not sure if we can make this `@internal`, because they methods have been part of the public API before.

core-bundle/src/Resources/contao/classes/BackendTemplateTrait.php
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	protected function getLocaleString()
+	{
+		$container = System::getContainer();
+
+		return
+			'var Contao={'
+			. 'theme:"' . Backend::getTheme() . '",'
+			. 'lang:{'
+			. 'close:"' . $GLOBALS['TL_LANG']['MSC']['close'] . '",'
```
leofeyer: Can you please restore the indentation here and in the method below?

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/FrontendTemplateTrait.php
m-vo: I think this is fine because it only disallows using the trait, not the methods:

Screenshot_20210824_145043

core-bundle/src/Resources/contao/classes/BackendTemplateTrait.php
m-vo: Oops, auto formatting. Fixed in 9c6dc45830523700c850d89bb07efe83dab9ca1f.

core-bundle/src/Resources/contao/classes/FrontendTemplateTrait.php
leofeyer: @ausi WDYT?

m-vo: Psalm seems happy, too: https://psalm.dev/r/a343df1b5c

ausi: Symfony seems to be using it this way too, for example here: https://github.com/symfony/symfony/blob/9a2b25dbfd0af2af50147a3d69b0d3ba93f638b0/src/Symfony/Bridge/PhpUnit/Legacy/ExpectDeprecationTraitForV8_4.php the method `expectDeprecation()` is allowed to be used but not the trait itself via `use`.

So I think it is correct this way.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3369, opened 2021-08-23
Fix using insert tags in the page title

ausi:
Allowing HTML in the page title and/or using insert tags like `{{br}}` is broken with `{{page::pageTitle}}`.

This pull request should fix this.

Related: https://github.com/contao/core/issues/4687

/cc @aschempp

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -236,7 +236,7 @@ $GLOBALS['TL_DCA']['tl_page'] = array
 			'exclude'                 => true,
 			'search'                  => true,
 			'inputType'               => 'text',
-			'eval'                    => array('maxlength'=>255, 'decodeEntities'=>true, 'tl_class'=>'w50'),
+			'eval'                    => array('maxlength'=>255, 'tl_class'=>'w50'),
 			'sql'                     => "varchar(255) NOT NULL default ''"
 		),
 		'language' => array
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -697,8 +697,6 @@ class InsertTags extends Controller
 
 				// Page
 				case 'page':
-					$flags[] = 'attr';
-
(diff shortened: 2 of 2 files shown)


core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -713,6 +711,11 @@ protected function doReplace($strBuffer, $blnCache)
 					}
 
 					// Do not use StringUtil::specialchars() here (see #4687)
```
leofeyer, line 713: Did you add your changes after the comment on purpose? Because the comment actually belongs to line 719.

ausi: Yes because this if now skips encoding for the title properties.

leofeyer (approved):

leofeyer:
Thank you @ausi.

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -236,7 +236,7 @@
 			'exclude'                 => true,
 			'search'                  => true,
 			'inputType'               => 'text',
-			'eval'                    => array('maxlength'=>255, 'decodeEntities'=>true, 'tl_class'=>'w50'),
```
aschempp, line 239: @Toflar I think we discussed this might cause issues for the response context?

Toflar: I guess @ausi is the first one who would think of this because he authored the decoding stuff there ;)

ausi: I think @aschempp is right, we probably should adjust https://github.com/contao/contao/blob/7bd09fae31188fb3d3a38c36e527e0ba5f6dfbf3/core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php#L86

leofeyer: See #3380

aschempp: Or maybe actually rely on the DCA config so this wouldn't happen in the first place?

PR #3375, opened 2021-08-24
Merge branch '4.11' into 4.12

leofeyer:


core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -728,6 +726,11 @@ protected function doReplace($strBuffer, $blnCache)
 					else
 					{
 						// Do not use StringUtil::specialchars() here (see #4687)
+						if (!\in_array($elements[1], array('title', 'parentTitle', 'mainTitle', 'rootTitle', 'pageTitle', 'parentPageTitle', 'mainPageTitle', 'rootPageTitle'), true))
+						{
+							$flags[] = 'attr';
+						}
```
leofeyer, line 732: @ausi Can you please check if we have to add this inside the `if` condition above as well?

ausi: No, we don’t have to, because encoding is already handled correctly there.

PR #3379, opened 2021-08-24
Add "Do Not Track" option to the Vimeo content element

### LINKED ISSUE #1657: Add "do not track" option to the Vimeo content element
By default, Vimeo sets a cookie to track the performance of an embedded player. By adding "dnt=1" to the URL, you can disable tracking and prevent Vimeo from setting a cookie: https://vimeo.zendesk.com/hc/en-us/articles/360001494447-Using-Player-Parameters (search for "dnt").

As of Contao 4.9.2, this can only be done by editing the ce_vimeo template. I think it would be nice to have a checkbox for this option in the Vimeo content element, along with autoplay, looping etc.

MarkejN:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1657
| Docs PR or issue | -

This adds an option to the Vimeo content element to set the url parameter `dnt` to 1.

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/tl_content.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/tl_content.xlf
@@ -641,6 +641,9 @@
       <trans-unit id="tl_content.vimeo_byline">
         <source>Hide the author</source>
       </trans-unit>
+      <trans-unit id="tl_content.vimeo_dnt">
+        <source>Set to "Do Not Track"</source>
+      </trans-unit>
       <trans-unit id="tl_content.new.0">
         <source>New</source>
       </trans-unit>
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/languages/en/tl_content.xlf
```diff
@@ -641,6 +641,9 @@
       <trans-unit id="tl_content.vimeo_byline">
         <source>Hide the author</source>
       </trans-unit>
+      <trans-unit id="tl_content.vimeo_dnt">
+        <source>Set to "Do Not Track"</source>
```
ausi: How about something like that?
```suggestion
        <source>Prevent session data tracking (dnt)</source>
```

From the Vimeo docs <https://developer.vimeo.com/player/sdk/embed#table-1>
> Whether to prevent the player from tracking session data, including cookies. Keep in mind that setting this argument to true also blocks video stats.

MarkejN: Yes, this is more meaningful.

ausi (dismissed):

m-vo (dismissed):

bytehead (dismissed):

leofeyer (approved):

leofeyer:
Thank you @MarkejN.

PR #3391, opened 2021-08-27
Always show the parent trails in the tree view

### LINKED ISSUE #3242: Issues with page mounts
**Affected version(s)**

All

**Description**

The way the current page mounts are handled causes a few issues:

1. As an editor, I cannot pick a page I do not have access to. This is nonsense because I just because I cannot edit the content of a page does not mean, I cannot link to a certain page.
2. When I add a new page, I may have issues with the alias because it already exists which is puzzling for a user as they cannot see the other pages.
3. The order of the pages in the site structure is totally random. E.g. as an admin I can sort the pages like so

- Root page A
- Root page B
- Root page C

and a user can get

- Root page C
- Root page A

I think the only real solution to this problem is to always show the whole page tree structure to all users but disallow any operations on the pages that were not allowed by the root settings.

### LINKED ISSUE #3428: Feature: Startseite bei gefilterten Seiten anzeigen
Ich fände sehr hilfreich, wenn beim Seitenfilter auch die Startpunkte angezeigt wären, damit man weiß, zu welcher Sprachversion die aufgelisteten Seiten gehören. 

Wenn man sich vorstellt, dass man mehrere Sprachen mit den Seiten namens Download hat, kann man gleich diesen vorgeschlagen Vorteil nachvollziehen (siehe beigefügtes Bild)
contao-feature_seitenfilter-mit-startpunkten_1

Toflar:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/3242

This PR is a first shot at trying to display the root elements of page mounts (or any other tree view).

Before (until Contao 4.12.x):

Bildschirmfoto 2021-08-27 um 14 10 00

With this PR:

Bildschirmfoto 2021-08-27 um 14 09 05

It not only shows the trail but also ensures a consistent sorting 😎


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6204,6 +6180,75 @@ protected function formatGroupHeader($field, $value, $mode, $row)
...
+		foreach ($this->root as $id)
+		{
+			$trail = array_map('\intval', $this->Database->getParentRecords($id, $table));
+
+			$this->rootTrails[$id] = $trail;
+			$topMostPerRoot[$id] = end($trail) ?: $id;
+		}
+
+		// Sort the root based on the sorting of the root trails if sorting is enabled
+		if ($this->Database->fieldExists('sorting', $table))
```
Toflar: This here ensures, the tree views are consistent according to the sorting of the topmost tree element.

core-bundle/src/Resources/contao/library/Contao/Database.php
```diff
@@ -572,6 +572,12 @@ public function getParentRecords($intId, $strTable)
 
 		while ($objPages->next())
 		{
+			// TODO: Find a better solution for this? Why does this method include the ID itself?
```
Toflar: I'm not sure about this but why does `getParentRecords()` return an array including itself? If it says parent records? Can we change that?

fritzmg: This appears to be out of convenience, as it is mostly used to check whether a specific node (or list of nodes) is part of a specific sub tree, including itself. While the method's name is misleading, we cannot change it for BC reasons. We could introduce a third parameter `$blnIncludeSelf = true` …

Toflar: Implemented in b559a04e4362af8ee7271d6cb10b97c3756e6c84.

Toflar:
This is ready to be reviewed. While I was on it, I also fixed `Database::getParentRecords()` to be limited to 10 levels. While it might be enough in most cases, I don't see why we should not add recursion here to fix this properly :)

leofeyer:
> While it might be enough in most cases, I don't see why we should not add recursion here to fix this properly :)

It seems that it has been enough in **all** cases in the last 15 years – at least it was never a (reported) issue. So I doubt that it is necessary to change it now. 🤷‍♂️ But since the way you have implemented it does not hurt performance, I don't mind.

leofeyer:
The paste buttons should not be visible, should they?

Toflar:
Fixed. There are now no operations anymore for parent trails.
The PR is also updated:

* The feature is now disabled by default for BC, other existing trees. The setting is currently named `showRootTrails`.
* I've fixed the breadcrumb to also show the parent trail
* The node filtering (limiting the tree view) is now only enabled on links which are not part of the trail
* `showRootTrails` is dynamically disabled if you use the breadcrumb now because having a breadcrumb and still showing the trail in the tree makes no sense.

core-bundle/src/Resources/contao/library/Contao/Database.php
```diff
@@ -559,23 +559,35 @@ public function getChildRecords($arrParentIds, $strTable, $blnSorting=false, $ar
...
 
-		while ($objPages->next())
+		$ids = $objPages->fetchEach('id');
+		$ids = array_map('\intval', $ids);
+
+		// Trigger recursion in case our query returned exactly 10 IDs in which case we might have higher parent records
+		if (\count($ids) === 10)
 		{
-			$arrReturn[] = $objPages->id;
+			$ids = array_merge($ids, $this->getParentRecords(end($ids), $strTable, $skipId));
```
ausi:
```suggestion
			$ids = array_merge($ids, $this->getParentRecords(end($ids), $strTable, true));
```

As `end($ids)` is already in the `$ids` we should skip it in the recursion call I think, otherwise we could end up with duplicated ids.

core-bundle/src/Resources/contao/library/Contao/Database.php
```diff
@@ -559,23 +559,35 @@ public function getChildRecords($arrParentIds, $strTable, $blnSorting=false, $ar
...
 
-		return $arrReturn;
+		if ($skipId)
+		{
+			$ids = array_flip($ids);
+			unset($ids[$intId]);
+			$ids = array_values(array_flip($ids));
+		}
+
+		return $ids;
```
ausi:
```suggestion
			unset($ids[array_search($intId, $ids)]);
		}

		return array_values($ids);
```
Nitpick ☺️

ausi (commented):
Apart from the duplicate IDs ✅

core-bundle/src/Resources/contao/library/Contao/Database.php
Toflar: Yeah I just did the flipping because of performance reasons but the array is so small, it's probably irrelevant ;)

core-bundle/src/Resources/contao/library/Contao/Database.php
Toflar: Ah right :)

ausi (dismissed):

leofeyer (changes requested):
Something does not work as expected. These are the pagemounts:



However, the user only sees one of the three selected pages:



When I remove one of the page mounts, the user sees another page but still not all pages:





Any idea what could be causing this?

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -873,8 +873,8 @@ public static function addPagesBreadcrumb($strKey='tl_page_node')
 
 				$arrIds[] = $intId;
 
-				// No link for the active page
-				if ($objPage->id == $intNode)
+				// No link for the active page or pages in the trail
+				if ($objPage->id == $intNode || (!$objUser->isAdmin && !$objUser->hasAccess($objPage->id, 'pagemounts')))
```
leofeyer:
```suggestion
				if ($objPage->id == $intNode || !$objUser->hasAccess($objPage->id, 'pagemounts'))
```

`User::hasAccess()` already checks `isAdmin`, therefore this condition can possibly be simplified.

Toflar: Right, I just copied the original code ;)

Toflar:
Fixed and found another issue with root nodes having the same parent which is also fixed now

Toflar:
There's still a bug with rendering though. On it.

Toflar:
Now you can review again :)

leofeyer:
Almost there. Here's the admin view:



And here is the user view:



The order is not yet correct. Or does it depend on the order of the pagemounts?

leofeyer:
Or maybe it is correct, because that is how I have set the order in the user group settings.



Since the pages are sortable, I guess we should not mess with it.

Toflar:
The order matters for routing so it should be the same (hence this PR in the first place, so you can actually see, which root page is first and it must be consistent with the admin view). I'll try to reproduce your issue.

leofeyer:
Thank you.

Toflar:
Completely reworked the concept. It's now going top down and skipping stuff that should not be visible instead of going up from the visible elements which makes sorting a lot easier :)

asaage:
If the resulting site-structure is a tree anyway then the ordering of pagemounts in user/groups should in my opinion be totaly omited.

Toflar:
It's not ordered so nothing to change.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -209,42 +208,7 @@ public function __construct($strTable, $arrModule=array())
...
-				}
-			}
-		}
-
-		// Get the IDs of all root records (list view or parent view)
-		elseif (\is_array($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'] ?? null))
-		{
-			$this->root = array_unique($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root']);
-		}
+		$this->initRoots();
```
leofeyer, line 211: Why did you make this a separate method? It is not used anywhere else.

Toflar: Because it's nicer. We keep talking about refactoring the DC_Table so here's one step: Move the logic of deterimining the roots into its own method.

leofeyer (changes requested):
I think adding type safety might be a BC break, because we do not know for sure how third-party extensions use the hooks or extend the `DC_Table` class. I would therefore prefer not to require type safety.

Great job otherwise! 💪

Toflar:
I think it's a must because otherwise, proper `in_array()` checks are almost impossible. I don't think it's a BC break, existing comparisons should continue to work just fine.

leofeyer:
I am not worried about existing comparisons but rather that a child class sets `$this->root` with string IDs. Then the strict `in_array()` checks would fail (DC_Table line 3844).

leofeyer:
I have found another issue with the implementation: We now see the trail pages but we cannot click on them to limit the site structure to a parent node.



Wouldn't it be handy if we could click on "Modules" to only show this subtree?

**Edit:** This is a different issue that we should handle in a separate PR. Never mind.

leofeyer (dismissed):

leofeyer:
Thank you @Toflar.

PR #3394, opened 2021-08-28
Show table name in model relation exception

### LINKED ISSUE #3132: die aktuelle Tabelle mit in der Fehlermeldung ausgeben
https://github.com/contao/contao/blob/f277abad6b6a66ace61804b4017e7e2bf8f6c99d/core-bundle/src/Resources/contao/library/Contao/Model.php#L584

die Tabelle mit in der Fehlermeldung ausgeben wo das jeweilige Feld nicht related werden kann.

`throw new \Exception("Field $strKey does not seem to be related in ".self::getTable());`

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3132

This PR adds the model's table name to the Exception message for a missing relation.

Old exception message:
> Field bar does not seem to be related

New message:
> Field tl_foo.bar does not seem to be related

This makes debugging a missing relation easier.

Note: While this may be a feature, it should easily be mergeable to the 4.9 branch.


SeverinGloeckle:
Fixed both the coding standard and the failing unit-test for PHP 7.2.
Can you rerun the workflow please?

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -583,7 +583,8 @@ public function getRelated($strKey, array $arrOptions=array())
 		// The relation does not exist
 		if (!isset($this->arrRelations[$strKey]))
 		{
-			throw new \Exception("Field $strKey does not seem to be related");
+			$table = static::getTable();
```
bytehead, line 586: CS needs a empty line after `$table = static::getTable();`

SeverinGloeckle: Thanks, added.

Any reason why this doesn't get fixed or flagged by:

````
tools/ecs/vendor/bin/ecs check core-bundle/src/Resources/contao/library/Contao/Model.php --config tools/ecs/vendor/contao/easy-coding-standard/config/default.php --no-progress-bar --ansi
````

As I understand, this is the same check that is run by the workflow. 
Is there any different configuration?

Comment (bytehead) [reply core-bundle/src/Resources/contao/library/Contao/Model.php:586]: did you miss the `--fix` flag?

File: core-bundle/tests/Contao/ModelTest.php [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessageMatches('/'.preg_quote($expectedExceptionContent).'/');
+
+        // Sadly, we cannot use a mock of \Contao\Model for abstract testing
+        // since PHPUnit mocks cannot call or mock static methods (in this case,
+        // static::getTable() within the Model), so use a real model
+        // class and dynamically set the table name via reflection.
+
+        // ContentModel could be any other Model class extending from \Contao\Model
+        $mock = new \Contao\ContentModel();
```
bytehead: Instead of using reflection here you can use the `Adapter` class: https://github.com/contao/contao/blob/4.9/core-bundle/src/Framework/Adapter.php

bytehead (changes requested):

core-bundle/tests/Contao/ModelTest.php
SeverinGloeckle: I don't see how this would be possible with the Adapter class since the problematic call to `static::getTable()` is right within the class instance.

While the `Adapter` would help for an external call via `$adapter->getTable()`, we need to mock the static method within the subject itself – the Model instance we call `getRelation` on, which then  calls `static::getTable()` internally. 

Sadly, mocking a static method is not possible with PHPUnit mocks (anymore), so I think reflection is the easiest way here.

On the other hand, we could use the `Adapter` within `Model::getRelated()` to resolve the table name in a testable way, but this doesn't feel like the right way either looking at the amount of static calls within the `Model` class to itself (`static::getTable`, `static::getClassFromTable` etc.).

Am I missing something?

core-bundle/src/Resources/contao/library/Contao/Model.php
SeverinGloeckle: I was running ecs with the default configuration:

````
--config tools/ecs/vendor/contao/easy-coding-standard/config/default.php
````

but this skips the legacy code in `Resources/contao/`, which has to be checked with the legacy configuration:

````
--config tools/ecs/vendor/contao/easy-coding-standard/config/legacy.php
````

My mistake.

bytehead: 👍

core-bundle/tests/Contao/ModelTest.php
bytehead: Hmm I see. I don't think we ever tested legacy code because of this... Let's see what the others think about.

leofeyer: I don't think we need a test for a modified exception message. Just delete the `ModelTest` class again.

SeverinGloeckle: Sounds reasonable. I have updated the pull request and removed the unit-test.

bytehead (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3396, opened 2021-08-30
Evaluate robots setting for sitemap

### LINKED ISSUE #3395: bug sitemap.xml Contao 4.12 RC2 / 4.12.1
Contao 4.9.12.RC2 or Contao 4.12.1

There is a bug in the above Contao versions. All pages are included in the sitemap.xml, even if they are marked with noindex / nofollow.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3395
| Docs PR or issue | -

This bug was introduced in https://github.com/contao/contao/commit/485bc50898acf8baca7b4a147692ca8407adcfe7. Previously the sitemap controller used `Backend::findSearchablePages` which checked the `robots` setting of each page - but it's missing in the current implementation.

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -140,9 +140,9 @@ class SitemapController extends AbstractController
                 continue;
             }
 
-            $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time()) && 'noindex,nofollow' !== $pageModel->robots;
+            $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time());
 
-            if ($isPublished && !$pageModel->requireItem && $this->pageRegistry->supportsContentComposition($pageModel)) {
+            if ($isPublished && !$pageModel->requireItem && $this->pageRegistry->supportsContentComposition($pageModel) && 'noindex,nofollow' !== $pageModel->robots) {
                 $urls = [$pageModel->getAbsoluteUrl()];
 
                 // Get articles with teaser
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -142,7 +142,7 @@ class SitemapController extends AbstractController
 
             $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time());
 
-            if ($isPublished && !$pageModel->requireItem && $this->pageRegistry->supportsContentComposition($pageModel) && 'noindex,nofollow' !== $pageModel->robots) {
+            if ($isPublished && !$pageModel->requireItem && 'noindex,nofollow' !== $pageModel->robots && $this->pageRegistry->supportsContentComposition($pageModel)) {
                 $urls = [$pageModel->getAbsoluteUrl()];
 
                 // Get articles with teaser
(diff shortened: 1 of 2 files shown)


fritzmg:
I'll also add a test.

fritzmg:
@aschempp @Toflar I am having trouble understanding which data I need to feed to `SitemapControllerTest::mockFrameworkWithPages`. It keeps failing for me with

```
1) Contao\CoreBundle\Tests\Controller\SitemapControllerTest::testSkipsNoindexNofollowPages
Expectation failed for method name is "findPublishedWithTeaserByPid" when invoked 2 time(s).
Method was expected to be called 2 times, actually called 1 times.
```

even though I am defining in the second array, that neither of my test pages have articles (`[43 => null, 44 => null]`). What am I missing here?

Toflar:
I guess `findPublishedWithTeaserByPid()` is not called because your page is skipped entirely due to your new comparison? So `$articles` should only be one element?

fritzmg:
Ah, yes, thank you :). RTM

Toflar (dismissed):

bytehead (dismissed):

ausi (dismissed):

core-bundle/src/Controller/SitemapController.php
```diff
@@ -140,7 +140,7 @@ private function getPageAndArticleUrls(int $parentPageId): array
                 continue;
             }
 
-            $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time());
+            $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time()) && 'noindex,nofollow' !== $pageModel->robots;
```
leofeyer: It is wrong to consider the meta robots setting to determine the `isPublished` status, isn't it?

leofeyer (changes requested):

core-bundle/src/Controller/SitemapController.php
ausi: Should we rename the Variable to something like `$isPublishedForSitemap` to make it more clear?

fritzmg: True. I can move it down to line 145 instead. Then it would be more or less in the "same" spot as before: 

https://github.com/contao/contao/blob/bdc6e30cf50421bf3ec98d67b0b21ddd5ffa950d/core-bundle/src/Resources/contao/classes/Backend.php#L695-L698

core-bundle/src/Controller/SitemapController.php
```diff
@@ -142,7 +142,7 @@ private function getPageAndArticleUrls(int $parentPageId): array
 
             $isPublished = $pageModel->published && (!$pageModel->start || $pageModel->start <= time()) && (!$pageModel->stop || $pageModel->stop > time());
 
-            if ($isPublished && !$pageModel->requireItem && $this->pageRegistry->supportsContentComposition($pageModel)) {
+            if ($isPublished && !$pageModel->requireItem && $this->pageRegistry->supportsContentComposition($pageModel) && 'noindex,nofollow' !== $pageModel->robots) {
```
ausi:
```suggestion
            if ($isPublished && !$pageModel->requireItem && 'noindex,nofollow' !== $pageModel->robots && $this->pageRegistry->supportsContentComposition($pageModel)) {
```
Maybe more performant this way?

ausi (dismissed):

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3400, opened 2021-08-30
Feature/bearer scope

xprojects-de:
| Q                | A
| -----------------| ---
| Docs PR or issue | https://github.com/contao/contao/issues/3024

Hello,

this request is a prototype for authentication using the JWT Bearer method which can be a feature for future version. 

I need this very often and wanted to discuss with you guys whether it might be something for the core :-)

**Goal:**
- There are two new scopes called "bearerFrontend" and "bearerBackend"
- Controller/Routes can be annotated with this scope and so the Controllers/Routes are only avalible in the bearer scope
- There is a new Controller "BearerController" where some basic features (Routes) are implemented (Login, UserInfo, RefreshToken, Logout) 

Currently only the basic features are included, which can be expanded.

**Note:**
Currently the deprecated AbstractGuardAuthenticator (https://symfony.com/doc/current/security/guard_authentication.html) is used because in don´t really know if the currenty Security of Contao is compatible to the new Authenticator-based Security (https://symfony.com/doc/current/security/authenticator_manager.html). The goal should be to use this but the deprecated  AbstractGuardAuthenticator still works in 5.4. So we have to discuss if the new Security can be used here...

**More features for future:**
- generate token from Backend
- make possibility to generate alternative token without TTL (e.g. to use it in machines without username/password)
- maybe show a list of active tokens in Backend
- ...

Would be great if you take a look at it and give your opinion on it :-)

Thank a lot and greetings

### DIFF
--- a/core-bundle/src/Controller/Bearer/BearerController.php
+++ b/core-bundle/src/Controller/Bearer/BearerController.php
@@ -156,7 +156,7 @@ class BearerController extends AbstractController
     }
 
     /**
-     * @Route("/bearerFrontend/logout", name="contao_bearer_logout", defaults={"_scope" = "bearerFrontend", "_token_check" = false}, methods="GET")
+     * @Route("/bearerFrontend/logout", name="contao_bearerFrontend_logout", defaults={"_scope" = "bearerFrontend", "_token_check" = false}, methods="GET")
      *
      * @param Request $request
      * @param UserInterface $user
@@ -165,12 +165,28 @@ class BearerController extends AbstractController
      */
     public function logoutMember(Request $request, UserInterface $user): JsonResponse
     {
-        // For future use - but therefore the tokens has to be stored in database and auth should compare to existing database entries
-        throw new \Exception('not supported');
+        try {
+
+            $this->framework->initialize();
+
+            $user = MemberModel::findByUsername($user->getUsername());
+            if ($user === null) {
+                throw new \Exception('user not found');
+            }
+
+            $user->jwt = null;
+            $user->save();
+
+            return (new JsonResponse(true));
+
+        } catch (\Exception $exception) {
+            return (new JsonResponse($exception->getMessage(), 400));
+        }
+
     }
 
     /**
-     * @Route("/bearerBackend/logout", name="contao_bearer_logout", defaults={"_scope" = "bearerBackend", "_token_check" = false}, methods="GET")
+     * @Route("/bearerBackend/logout", name="contao_bearerBackend_logout", defaults={"_scope" = "bearerBackend", "_token_check" = false}, methods="GET")
      *
      * @param Request $request
      * @param UserInterface $user
@@ -179,8 +195,23 @@ class BearerController extends AbstractController
      */
     public function logoutUser(Request $request, UserInterface $user): JsonResponse
     {
-        // For future use - but therefore the tokens has to be stored in database and auth should compare to existing database entries
-        throw new \Exception('not supported');
+        try {
+
+            $this->framework->initialize();
+
+            $user = UserModel::findByUsername($user->getUsername());
+            if ($user === null) {
+                throw new \Exception('user not found');
+            }
+
+            $user->jwt = null;
+            $user->save();
+
+            return (new JsonResponse(true));
+
+        } catch (\Exception $exception) {
+            return (new JsonResponse($exception->getMessage(), 400));
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


fritzmg:
> Currently the deprecated AbstractGuardAuthenticator (https://symfony.com/doc/current/security/guard_authentication.html) is used because in don´t really know if the currenty Security of Contao is compatible to the new Authenticator-based Security (https://symfony.com/doc/current/security/authenticator_manager.html). The goal should be to use this but the deprecated AbstractGuardAuthenticator still works in 5.4. So we have to discuss if the new Security can be used here...

I think for Contao 4.13 we want to go with Symfony 5 only, so no backwards compatibility with anything prior to 5.1 or even 5.3 is really necessary.

aschempp:
Generally, I would like to switch to the new Authentication system in Contao 5, since changing this in Contao 4 would be sort of a BC break. Regardless of that, I'm not sure if we want to add a new firewall scope.

@xprojects-de can you explain the use-case to differentiate between "bearer frontend authentication" and "regular frontend authentication"? Apart from the problem that the Symfony firewall does not support "stateful" and "stateless" in the same firewall, shouldn't it just be `frontend` or `backend`?

xprojects-de:
ok fine, but honestly said i dont´really know what effect it has if the" enable_authenticator_manager" will be enabled at the security!?! I don't know enough about the existing structure ... Otherwise the new Authenticator Manager seem not to be working!?

xprojects-de:
@aschempp Definitely it would be nice to add the bearer Auth to the existing scopes. But as you said i think there is a general problem with the stateless topic! 
Is it also possible to do the complete Guard-Things on a different way with existing scopes?

xprojects-de:
It would be also nice to use the UserChecker also at the BearerLogin. This is a nice thing and i will implement it soon.

@aschempp 

Looking at the AuthenticationProvider and also UserProvider i think it´s not really good to mix it especially because of the statefull firewall. In my opinion this would be not a clean implementation. Furthermore backend-Scope Requests are currently prepared to render visual BackendRoutes. I don´t really know how much effort it would be to support also e.g. Json-Content-Types and other instead of text/html!?

But that's just my opinion. You definitely have a lot more experience there !!! :-)

Thanks

xprojects-de:
added Userchecker and attributes (anonymous and lazy). Lazy needs "symfony/security-bundle": "^5.1". If it´s a problem we can also remove it!

xprojects-de:
I also took a closer look to the current Auth system and i still think that there should be a different when using classical form or using any stateless token based system. Using the guard (in future the new custom Authenticator providers) the Auth process is sepeareted cleanly without the current LoginFactory and AuthenticationProviderListener which are very session focused.

if better it´s possible to use just one scope for both processes e.g. "bearer" or "api" using a own UserProvider which handles both types of user instances. That would be no problem.

Furthermore it would be interesting :-) how the future structure of the API should look like!? Is that also done using jwt/Bearer or do you have other ideas? Maybe the process can build  that i can help for the future API!?

Greetings

xprojects-de:
After further tests and i think last tests on it:

It would be possible to add the gaurd to existing firewalls (or say scopes) like e.g. 

```yaml 
...
contao_backend:
            request_matcher: contao.routing.backend_matcher
            provider: contao.security.backend_user_provider
            user_checker: contao.security.user_checker
            anonymous: ~
            switch_user: true

            contao_login:
                remember_me: false

            logout:
                path: contao_backend_logout
                handlers:
                    - contao_manager.security.logout_handler
                success_handler: contao.security.logout_success_handler

            guard:
                authenticators:
                    - contao.security.bearer_token_authenticator

        contao_frontend:
            request_matcher: contao.routing.frontend_matcher
            provider: contao.security.frontend_user_provider
            user_checker: contao.security.user_checker
            anonymous: ~
            switch_user: false

            contao_login:
                remember_me: true

            remember_me:
                secret: '%kernel.secret%'
                remember_me_parameter: autologin

            logout:
                path: contao_frontend_logout
                handlers:
                success_handler: contao.security.logout_success_handler

            guard:
                authenticators:
                    - contao.security.bearer_token_authenticator
     ...
 ```

and check the if matched in the supports() e.g.

```php
...
public function supports(Request $request): ?bool
    {
        if ($request->headers->has('Authorization')) {
            return true;
        }

        return false;
    }
...
```
But the stateful attribute is a problem because always the session is checked/initialized and so on and in my opinion this is not clean code and not a clear implementation.

i also tried to use the new Authenticator-based Security using a custom Authenticator (https://symfony.com/doc/current/security/authenticator_manager.html#creating-a-custom-authenticator) but it seems it´s not compatible with current implementation (there is an error with the contao_login...).

Currently i think it´s very difficult to merge such types of Auth with the current Auth implementation without using separate firewalls (or say scopes).

I really understand that this is maybe a special usecase and honestly i don´t know if i would approve that either :-)

But I just want to show that it would be nice to have something so simple for Contao... :-) Maybe it´s easier when using the new Auth System in future or someone else has another good idea to maybe get it into 4.13 !

Still thanks to everyone 
Greetings

Toflar:
@xprojects-de did you re-visit this topic for Contao 5 that uses Symfony 6 security stuff only? Maybe things got easier? Either way, please close the PR if you don't plan on working on it (or update to 5.x if things are easier now 😊 ).

xprojects-de:
I am still interested in a general bearer authentication, but I don't know if it is currently desirable to have a separate firewall that is without a session and exists only for a API? Perhaps Core Team will release something like that soon :-) 

Currently, I always implement the feature directly within the bundle. You can define a separate firewall there, too.
Of course, this works much better with Symfony 6/7 :-)

Therefore, I think I will close the pull request.

PR #3402, opened 2021-08-30
Fix the `Adapter::__call()` method signature

SeverinGloeckle:
Currently it is impossible to mock Contao's `Adapter` class with the [Mockery framework](https://github.com/mockery/mockery):

````php
Mockery::mock(\Contao\CoreBundle\Framework\Adapter::class);

// Declaration of Mockery_X_Contao_CoreBundle_Framework_Adapter::__call(string $method, array $args) 
// should be compatible with Contao\CoreBundle\Framework\Adapter::__call(string $name, array $arguments = Array)
````

This is due to the method signature of the `Adapter` defining `$arguments` as an optional parameter with a default value:

````php
public function __call(string $name, array $arguments = [])
````

This differs from the standard method signature of [__call](https://www.php.net/manual/en/language.oop5.overloading.php#object.call):

````php
public __call(string $name, array $arguments): mixed
````

Even though Contao isn't using Mockery itself, it would be nice if we could use Mockery in our projects.

There are multiple similar (old) issues in mockery/mockery#263

### DIFF
--- a/core-bundle/src/Framework/Adapter.php
+++ b/core-bundle/src/Framework/Adapter.php
@@ -39,7 +39,7 @@ class Adapter
     /**
      * Calls a method of the adapted class.
      */
-    public function __call(string $name, array $arguments = [])
+    public function __call(string $name, array $arguments)
     {
         return \call_user_func_array([$this->class, $name], $arguments);
     }


aschempp:
I don't see a way to remove a default argument, this would break with any existing code using the adapter without a second argument?

ausi:
> this would break with any existing code using the adapter without a second argument?

Unable to reproduce: https://3v4l.org/rDRvh

ausi (approved):

bytehead (approved):

SeverinGloeckle:
This change does not break the regular use of an adapter since PHP adds the `$arguments` array when calling the magic `__call()` method (a behaviour that doesn't seem to be documented well, but can be seen in the example from @ausi).

But @aschempp  still has a point: Removing the default parameter would break subclasses of the adapter that rely on the default argument by calling `parent::__call()` directly:

````php
parent::__call('foo'); // throws ArgumentCountError
````

However, since the `Adapter` class is flagged as `@internal`, this change doesn't break the backwards compatibility promise in my opinion – but makes the `Adapter` more/easier testable.

ausi:
Also, the [official method signature of `__call`](https://www.php.net/manual/en/language.oop5.overloading.php#object.call) is 
`public __call(string $name, array $arguments): mixed`

So I think it is correct to change that. We should also add `: mixed` once we switch to PHP 8.0.

leofeyer:
Thank you @SeverinGloeckle.

PR #3405, opened 2021-08-31
Harden the version 4.5.0 migration

### LINKED ISSUE #3403: Improve 4.5.0 Migration
**Affected version(s)**

Contao 4.9+

**Description**

If the `Version450Update` migration fails somewhere before the `ALTER TABLE tl_layout ADD externalJs` execution for whatever reason, the migration will be offered again, but it will definitely fail then, because some of the fields have already been created by then. To fix this, the migration should check the existence for each field, before trying to create the field.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3403
| Docs PR or issue | -

This hardens the Contao 4.5.0 migration in case errors happened during the migration - or in case of an otherwise inconsistent database state.

### DIFF
--- a/core-bundle/src/Migration/Version405/Version450Update.php
+++ b/core-bundle/src/Migration/Version405/Version450Update.php
@@ -51,44 +51,62 @@ class Version450Update extends AbstractMigration
 
     public function run(): MigrationResult
     {
-        $this->connection->query('
-            ALTER TABLE
-                tl_content
-            ADD
-                youtubeOptions text NULL
-        ');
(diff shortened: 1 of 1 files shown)


ausi:
~~Should we add the same check for `tl_layout.externalJs` on line 140 too?~~

nevermind, this is already checked be the `shouldRun()` I see.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #3406, opened 2021-08-31
Remove the "generateSitemap" cronjob

fritzmg:
Currently the `SitemapController` defines that its response can be cached up to 30 days. However there is also a cronjob that will invalidate all sitemaps every 24 hours. This cronjob existed historically before the current sitemap implementation as a controller, in order to purge and regenerate the static sitemap XML files. This is not needed anymore.


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #3410, opened 2021-09-01
Use PHP functions to modify query strings

### LINKED ISSUE #3392: Controller::addToUrl() method too long runtime
Contao 4.9

Controller :: addToUrl() method is more than 20 times slower in 4.9 than in 3.5. As a result, e.g. the page administration can become slow. In a specific example of a multisite installation, TFB takes about 5 seconds in 3.5, but more than 20 seconds in 4.9. The addToUrl() method is used for almost all action buttons in the backend. According to this, all backend lists are slower than in Contao 3.5 due to this method.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3392
| Docs PR or issue | -

Tested with the page structure of the official demo.

### After

Needs to be backported to Contao 4.9 if we agree that changing the major version is OK.


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1070,29 +1075,26 @@ public static function generateMargin($arrValues, $strType='margin')
 	 */
 	public static function addToUrl($strRequest, $blnAddRef=true, $arrUnset=array())
 	{
-		$query = new Query(Environment::get('queryString'));
+		if (static::$objQuery === null)
+		{
+			static::$objQuery = Query::createFromRFC3986(Environment::get('queryString'));
```
Toflar: Why is this not from `Request` like the `_contao_referer_id` as well?

fritzmg:
I am wondering in general why we use `league/uri-components` for `Backend::addToUrl`, but not for `Frontend::addToUrl` as well (or vice versa).

leofeyer:
> The problem isn't the component but the ridiculous amount of calls to it. Likely unwanted recursion. 
> _Originally posted by @m-vo in https://github.com/contao/contao/issues/3392#issuecomment-910156791_

I don't think so. 35 pages with 8 buttons each will trigger 280 calls alone. Together with module links and global buttons, 302 calls are very likely.

leofeyer:
> I am wondering in general why we use `league/uri-components` for `Backend::addToUrl`, but not for `Frontend::addToUrl` as well (or vice versa).

Unrelated discussion. Please open a separate issue.

fritzmg:
> Unrelated discussion. Please open a separate issue.

It's not unrelated if the custom implementation in `Frontend::addToUrl` without `league/uri-components` is faster than the one in `Backend::addToUrl`, that's why I brought it up.

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: Because the `_contao_referer_id` part is new and the rest is legacy code? I am fine with changing this, though.

Maybe we can get rid of `leagure/uri-components` completely if we use a combination of Symfony's `HeaderUtils::parseQuery()`, some custom logic and `http_build_query()` (like Symfony does in `Request::normalizeQueryString()`)?

core-bundle/tests/Contao/ControllerTest.php
```diff
@@ -348,4 +349,94 @@ public function testAddImageToTemplateLogsAndCreatesFallbackDataWhenNoFigureIsBu
 
         $this->assertSame($expectedTemplateData, get_object_vars($template));
     }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testAddToUrlWithoutQueryString(): void
+    {
+        define('TL_SCRIPT', '');
```
Toflar: If we're already working on this method we could also get rid of this deprecated constant, no?

leofeyer:
I have added a variant without `league/uri-components` in 16395a221e69fe55cf6f5f68eb5059c1a42a0a9e. The performance is only slightly better than with `league/uri-components` in version 2, , therefore we should ask us if we rather want to use a well-tested third-party library or a custom function (I have added unit tests for it, too). @contao/developers WDYT?

core-bundle/tests/Contao/ControllerTest.php
leofeyer: That is also not related to this PR. Maybe create a separate PR and fix it everywhere instead of only in one place?

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: See 6d1f50dd22e02bc20f038b4a56b639b7352f146d

composer.json
```diff
@@ -60,8 +60,7 @@
         "knplabs/knp-time-bundle": "^1.10",
         "lcobucci/jwt": "^3.4 || ^4.0",
         "league/commonmark": "^1.6",
-        "league/uri-components": "^1.8",
-        "league/uri-schemes": "^1.2",
+        "league/uri-components": "^2.4",
```
leofeyer, line 63: We are also using this in other places, so if we decide to go without, we have to adjust those as well.

ausi:
> if we rather want to use a well-tested third-party library or a custom function

In this case I would use the custom function as the problem it solves is not complex enough to justify the use of a library IMO.

core-bundle/src/Resources/contao/pages/PageLogout.php
```diff
@@ -56,14 +55,21 @@ public function getResponse($objPage)
...
+		{
+			parse_str($request->server->get('QUERY_STRING'), $pairs);
+		}
 
 		// Add the redirect= parameter to the logout URL
-		$query = new Query($uri->getQuery());
-		$query = $query->merge('redirect=' . $strRedirect);
+		$pairs['redirect'] = $strRedirect;
+
+		$request->server->set('QUERY_STRING', http_build_query($pairs, '', '&', PHP_QUERY_RFC3986));
```
leofeyer: Not sure whether modifying the request object is the best solution here?

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1070,29 +1074,53 @@ public static function generateMargin($arrValues, $strType='margin')
...
 		// Merge the request string to be added
-		$query = $query->merge(str_replace('&amp;', '&', $strRequest));
+		if ($strRequest)
+		{
+			parse_str(str_replace('&amp;', '&', $strRequest), $newPairs);
+			$pairs = array_merge($pairs, $newPairs);
+		}
 
 		// Add the referer ID
 		if (isset($_GET['ref']) || ($strRequest && $blnAddRef))
```
ausi: Should we ask the request here instead of `$_GET`?

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -400,10 +399,14 @@ public static function getReferer($blnEncodeAmpersands=false, $strTable=null)
...
-				$queryObj = new Query($query);
-				$queryObj = $queryObj->withoutPairs($params);
+				parse_str($query, $pairs);
 
-				return $path . $queryObj->getUriComponent();
+				foreach ($params as $param)
+				{
+					unset($pairs[$param]);
+				}
+
```
ausi, line 408: Should we return without the `?` if the whole query string was unset?
```suggestion

				if (!$pairs) 
				{
					return $path;
				}

```

core-bundle/src/Resources/contao/pages/PageLogout.php
ausi: Maybe we could use `parse_url(…, PHP_URL_QUERY)` instead of the Request class and string replace the query afterwards?

leofeyer: The "replacing the query afterwards" is the problematic part here. The Symfony `Request` class is able to create a URI from the `parse_url()` parts, that's why I have used it here. Do you know a better way?

core-bundle/src/Resources/contao/library/Contao/System.php
leofeyer: Merged in cbae1469f9fa05aa3672f68e518ad86cdd77d435.

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: Yes. Changed in f013fab6a704077a2e9c6890fa979fc6ed084c7c.

core-bundle/src/Resources/contao/pages/PageLogout.php
ausi: Something like this? (untested)
```php
$url = preg_replace('(\?'.preg_quote($oldQuery).'(?=#|$))', '?'.$newQuery, $url);
```

leofeyer: I don't think that this is "better" than using the request object. 🙈 @contao/developers WDYT?

Toflar: `Request::create()` accepts `$_SERVER` like arguments, so you could pass it in there.

leofeyer: I have already tried this with `$request->duplicate(…, ['QUERY_STRING' => '…'])`, however, this will replace the `$server` variable instead of merging my array.

ausi: > I don't think that this is "better" than using the request object.

You’re right, it is probably not better ☺️

The current solution is fine then I think. 👍

ausi (dismissed):

Toflar (dismissed):
Thanks for the effort ❤️

core-bundle/src/Resources/contao/pages/PageLogout.php
leofeyer: I have made a small adjustment in 8da7cde8cadf4a44a2959adb164060027bfb605c, so we no longer have to overwrite the `QUERY_STRING` variable.

ausi (approved):

Toflar (approved):

core-bundle/tests/Contao/ControllerTest.php
aschempp: I have added it to the Contao 5 project ideas.

leofeyer: We don't have to wait for Contao 5. There is already an alternative to using `TL_SCRIPT` IIRC, so we could change this right away. It was just out of the scope of this PR.

PR #3411, opened 2021-09-01
Fix includeBlankOption with value 0 as next value in select

### LINKED ISSUE #1022: includeBlankOption does not work at select if first value 0

zonky2:
implement https://github.com/contao/contao/issues/1022#issuecomment-575210705 with cast option values to string if values int - see https://github.com/menatwork/contao-multicolumnwizard-bundle/issues/63

| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1022

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -1155,7 +1155,7 @@ abstract class Widget extends Controller
 			return '';
 		}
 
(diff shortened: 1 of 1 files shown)


ausi:
The comments of https://github.com/contao/contao/pull/2350#issuecomment-698518211 are still not addressed I think.

PR #3412, opened 2021-09-02
Do not dynamically disable HttpCache but rely on env vars instead

Toflar:
Bug description:

* Right now, you can circumvent the HttpCache by simply sending `Surrogate-Capability: ESI/1.0` with your request.

Reproducing is easy:

```
# This response will contain the "Contao-Cache" headers
curl "https://contao.org/de/" \
     -H 'Accept: */*' \
     -H 'Connection: keep-alive'

# This will not contain any "Contao-Cache" header but the X-Cache-Tags
curl "https://contao.org/de/" \
     -H 'Accept: */*' \
     -H 'Connection: keep-alive' \
     -H 'Surrogate-Capability: ESI/1.0'
```

Possible solutions:

1. Accept `Surrogate-Capability` only from trusted proxies.
2. Replace current dynamic logic with an explicit env var config

I went for number 2 because:

* No matter which solution we choose, both pose a possible BC break and need adjustments in your .env config. Either the trusted proxies or the new environment variable. But it's a bug, so we need to fix this. So BC is not relevant here.
* The dynamic logic does not allow for having a cache proxy in front of Contao without Surrogate Capabilities (ESI). However, this is totally supported by Symfony as it will just not generate any `<esi...` tags in this case.
* it's way more explicit and clear.

### DIFF
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -289,8 +289,8 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
         if (!$kernel->isDebug()) {
             $cache = $kernel->getHttpCache();
 
-            // Enable the Symfony reverse proxy if request has no surrogate capability
-            if (($surrogate = $cache->getSurrogate()) && !$surrogate->hasSurrogateCapability($request)) {
+            // Enable HTTP Cache if not disabled explicitly
+            if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
                 return $cache;
             }
         }
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -287,11 +287,9 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
         }
 
         if (!$kernel->isDebug()) {
-            $cache = $kernel->getHttpCache();
-
             // Enable HTTP Cache if not disabled explicitly
             if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
-                return $cache;
+                return $kernel->getHttpCache();
             }
         }
 
--- a/manager-bundle/src/HttpKernel/ContaoKernel.php
+++ b/manager-bundle/src/HttpKernel/ContaoKernel.php
@@ -286,11 +286,9 @@ class ContaoKernel extends Kernel implements HttpCacheProvider
             $kernel->setJwtManager($jwtManager);
         }
 
-        if (!$kernel->isDebug()) {
-            // Enable HTTP Cache if not disabled explicitly
-            if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
-                return $kernel->getHttpCache();
-            }
+        // Enable the Symfony reverse proxy if not disabled explicitly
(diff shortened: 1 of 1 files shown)


fritzmg:
Sounds reasonable to me. I think I remember one or two cases in the community where they tried to get their own cache proxy working but couldn't, because it did in fact not send the appropriate `Surrogate-Capability` header. Having to specifically disable the internal reverse proxy seems more straight forward 👍

manager-bundle/src/HttpKernel/ContaoKernel.php
```diff
@@ -289,8 +289,8 @@ public static function fromRequest(string $projectDir, Request $request): HttpKe
         if (!$kernel->isDebug()) {
             $cache = $kernel->getHttpCache();
 
-            // Enable the Symfony reverse proxy if request has no surrogate capability
-            if (($surrogate = $cache->getSurrogate()) && !$surrogate->hasSurrogateCapability($request)) {
+            // Enable HTTP Cache if not disabled explicitly
+            if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
                 return $cache;
```
ausi:
```suggestion
            // Enable HTTP Cache if not disabled explicitly
            if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
                return $kernel->getHttpCache();
```

ausi (dismissed):

fritzmg:
Regarding BC: I am not sure if we really should change this in a a bugfix release. Those community members I mentioned did end up getting their proxy configured so that it would send the appropriate `Surrogate-Capability` header. With this change the Symfony reverse proxy would automatically be enabled again after updating Contao to the recent bugfix version. Though I would assume this does not really break anything?

Toflar:
Imho it is a bug that you can disable the cache by simply sending this header. So we have to fix it and there's no way in doing it without a potential BC break. And it affects like 0.00000000001% of users anyway ^^

fritzmg (dismissed):
👍 Can you also please prepare a PR for contao/docs once merged :) https://docs.contao.org/dev/framework/caching/ and https://docs.contao.org/dev/reference/config/#environment-variables-for-the-contao-managed-edition

manager-bundle/src/HttpKernel/ContaoKernel.php
leofeyer: What about this suggestion?

manager-bundle/src/HttpKernel/ContaoKernel.php
```diff
@@ -287,11 +287,9 @@ public static function fromRequest(string $projectDir, Request $request): HttpKe
...
 
         if (!$kernel->isDebug()) {
-            $cache = $kernel->getHttpCache();
-
-            // Enable the Symfony reverse proxy if request has no surrogate capability
-            if (($surrogate = $cache->getSurrogate()) && !$surrogate->hasSurrogateCapability($request)) {
-                return $cache;
+            // Enable HTTP Cache if not disabled explicitly
+            if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? false)) {
+                return $kernel->getHttpCache();
```
leofeyer, line 292:
```suggestion
        // Enable the Symfony reverse proxy if not disabled explicitly
        if (!($_SERVER['DISABLE_HTTP_CACHE'] ?? null) && !$kernel->isDebug()) {
            return $kernel->getHttpCache();
        }
```

leofeyer (approved):

aschempp (approved):

ausi (approved):

leofeyer:
Thank you @Toflar.

Issue #3271, opened 2021-08-05
Register Twig namespaces for themes outside of /templates


richardhj:
**Affected version(s)**

Contao 4.12

**Description**

Currently, in the 4.12 RC, all theme templates folders under `/templates` get registered, i.e., https://github.com/contao/contao/blob/4.x/core-bundle/src/Twig/Loader/ContaoFilesystemLoaderWarmer.php#L54-L59

Prior to 4.12, it was supported to set a theme template folder to a path outside of `/templates`.
```sql
UPDATE tl_theme SET `templates` = '/themes/my_theme/templates' WHERE id=1
```

With an eye on #2781, this behavior will also become necessary. This behavior is consequently required by my [theme-framework](https://github.com/richardhj/contao-theme-framework) extension.

m-vo:
Currently the template locator returns all subdirectories of `/template` as theme paths and each is added:

https://github.com/contao/contao/blob/722d397e3429e3fc4a8d05263fe2f834aa8f2122/core-bundle/src/Twig/Loader/TemplateLocator.php#L46-L70

https://github.com/contao/contao/blob/722d397e3429e3fc4a8d05263fe2f834aa8f2122/core-bundle/src/Twig/Loader/ContaoFilesystemLoaderWarmer.php#L54-L59

We could maybe return registered paths from `tl_theme` instead and refresh the loader cache whenever a new folder gets picked in the back end.

But how would the slugs (used for namespaces) look for arbitrary paths? 
Currently they are `str_replace('/', '_', Path::normalize($relativePath))`.

richardhj:
> But how would the slugs (used for namespaces) look for arbitrary paths?
> Currently they are `str_replace('/', '_', Path::normalize($relativePath))`.

Has to be the full path.
But if we decide whether #2781 will be integrated into Contao one day, we can also strip the `/\/themes/(.*?)/templates/` part.

ausi:
In order to prevent duplicates we could use `@Contao_Theme_foo` for /templates/foo and `@Contao_Theme__foo` for /foo

ausi:
Would be something like `str_replace(['../', '/'], '_', Path::normalize($relativePath))` then.

m-vo:
> In order to prevent duplicates we could use @Contao_Theme_foo for /templates/foo and @Contao_Theme__foo for /foo

Or we always use the project-relative path:  `@Contao_Theme_templates_foo` and `@Contao_Theme_foo`.
[...]

PR #3414, opened 2021-09-03
Improve the Twig theme path handling

### LINKED ISSUE #3354: Folder names with "." not possible in templates folder
**Affected version(s)**

Contao 4.12

**Description**

After updating from 4.11.x to 4.12 I got an error in Contao Manager and after access to install tool:

```html
Twig\Error\LoaderError:
Tried to register an invalid Contao namespace 'Contao_Theme_XXX.XX_DEFAULT_DE'.

  at /htdocs/vendor/contao/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php:108
  at Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader->addPath('/htdocs/templates/XXX.XX_DEFAULT_DE', 'Contao_Theme_XXX.XX_DEFAULT_DE', true)
```

I renamed the folder `templates/XXX.XX_DEFAULT_DE` to `templates/XXX_XX_DEFAULT_DE` and it works. I think there's a problem with `.` in folder names.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3354, #3271
| Docs PR or issue | todo

This PR improves the Twig theme path handling:

- we now check for invalid characters and throw an `InvalidThemePathException` if the name contains anything else than `a-zA-Z0-9-`
- until Contao 5.0 we handle the exception and trigger a deprecation warning (except on CLI)
- when saving via the back end, paths with invalid characters are now rejected<sup>*)</sup>
- we allow relative paths outside the template directory. so `../../foo/bar` will get the slug  `__foo_bar`.
- theme paths aren't searched on the file system anymore, instead we ask the theme models for configured paths and refresh the cache when a user changes a path.

<sup>*)</sup> This is a small BC break but IMHO makes for a better UX and only affects changes made after the update.

A word of warning: This is untested, yet (in a real project).

### DIFF
--- a/core-bundle/src/Twig/Loader/Theme.php
+++ b/core-bundle/src/Twig/Loader/Theme.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Loader;
+
+use Contao\CoreBundle\Exception\InvalidThemePathException;
+use Webmozart\PathUtil\Path;
+
+class Theme
+{
+    /**
+     * Generates a theme slug from a relative path.
+     *
+     * @throws InvalidThemePathException if the path contains invalid characters
+     */
+    public function generateSlug(string $relativePath): string
+    {
+        if (!Path::isRelative($relativePath)) {
+            throw new \InvalidArgumentException("Path '$relativePath' must be relative.");
+        }
+
+        $path = Path::normalize($relativePath);
+        $invalidCharacters = [];
+
+        $slug = implode('_', array_map(
+            static function (string $chunk) use (&$invalidCharacters) {
+                // Allow paths outside the template directory (see #3271)
+                if ('..' === $chunk) {
+                    return '';
+                }
+
+                // Check for invalid characters (see #3354)
+                if (0 !== preg_match_all('%[^a-zA-Z0-9-]%', $chunk, $matches)) {
+                    $invalidCharacters = array_merge($invalidCharacters, $matches[0]);
+                }
+
+                return $chunk;
+            },
+            explode('/', $path)
+        ));
+
+        if (!empty($invalidCharacters)) {
+            throw new InvalidThemePathException($path, $invalidCharacters);
+        }
+
+        return $slug;
+    }
+
+    /**
+     * Builds the namespace for a certain theme slug.
+     */
+    public function getThemeNamespace(string $slug): string
+    {
+        return "@Contao_Theme_$slug";
+    }
+
+    /**
+     * Extracts a theme slug from a given logical name.
+     *
+     * @return string the theme slug or null if not a theme namespace
+     */
+    public function matchThemeNamespace(string $logicalName): ?string
+    {
+        if (1 === preg_match('%^@Contao_Theme_([a-zA-Z0-9_-]+)/%', $logicalName, $matches)) {
+            return $matches[1];
+        }
+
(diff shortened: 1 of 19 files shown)


ausi (dismissed):
Code LGTM, but someone testing this patch in a real world project would be great before merging IMO.

richardhj:
I'll provide feedback within the next 36 hours.

richardhj:
The changes in this PR work out to be great. I think it is a perfect idea to ask the theme models for registered template folders -- instead of registering the whole /templates dir.

m-vo:
Should we refactor `TemplateLocator::createDirectorySlug()` to become a service instead? 

* No static context, we can safely use parameters (in the future).
* SOC, because this doesn't have much to do with locating templates.
* People can decorate it if they want a custom mapping. E.g. if your slug is `_themes_foo_bar_templates` for your "main theme" you could instead return an empty string which would give you `@Contao_Theme/baz.html.twig` instead of `@Contao_Theme__themes_foo_bar_templates/baz.html.twig`.

@ausi, @richardhj  Wdyt?

richardhj:
Being able to alter the template slug is a major concern for me, and to refactor the slugging to a service that can be decorated sounds reasonable for me 🙂

m-vo:
I refactored all theme slug/path related things into a ~~`Twig/Loader/Theme`~~ `Twig/Loader/ThemeNamespace` service. I think it's much cleaner this way and you should get full control over slugs and namespaces when decorating. But let's hear what you think first. :slightly_smiling_face: 

I also made sure the term `slug` is consistently used (and not `alias`).

core-bundle/src/Twig/Loader/Theme.php
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Loader;
+
+use Contao\CoreBundle\Exception\InvalidThemePathException;
+use Webmozart\PathUtil\Path;
+
+class Theme
```
ausi: Is there maybe a better name that is not so generic? Maybe `ThemeNamespace` or `ThemeNamespaces`?

core-bundle/src/Twig/Loader/Theme.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return "@Contao_Theme_$slug";
+    }
+
+    /**
+     * Extracts a theme slug from a given logical name.
+     *
+     * @return string the theme slug or null if not a theme namespace
+     */
+    public function matchThemeNamespace(string $logicalName): ?string
```
ausi:
```suggestion
    public function match(string $logicalName): ?string
```

This could be renamed then maybe if we change the class name?

ausi (dismissed):
LGTM besides naming the `Theme` class.

core-bundle/src/Twig/Loader/Theme.php
m-vo: renamed in 2e75fbfebfcc368d5362536522cd1849d1177b27

core-bundle/src/Twig/Loader/Theme.php
m-vo: renamed in 2e75fbfebfcc368d5362536522cd1849d1177b27

m-vo:
Sorry about the force push noise. I forgot to check in the changes to the service declarations. Twice. :facepalm:

ausi (dismissed):

richardhj (dismissed):

richardhj (dismissed):

richardhj (approved):

leofeyer:
Thank you @m-vo.

PR #3416, opened 2021-09-03
Dispatch the PreviewUrlCreateEvent even with an empty ID

### LINKED ISSUE #2580: Preview link not correctly set when opening the backend record page (act=edit) directly after login
**Affected version(s)**

4.4, 4.9 (probably 4.10 also)

**Description**

After opening e.g. an event edit page in contao backend in the usual way, i.e. by clicking "Events" and going into the archive and then in the edit page of the event, the preview link is correctly set (`BackendMain` -> `frontendFile` property).

Now some of the our customers have very short session times leading to the record page being reloaded, logged out having to log in again. After this login the `CURRENT_ID` is not set so the following if is false:

```
// vendor/contao/core-bundle/src/Resources/contao/controllers/BackendMain.php

// Front end preview links
if (\defined('CURRENT_ID') && CURRENT_ID != '')
{
	if (\Input::get('do') == 'page')
	{
		$this->Template->frontendFile = '?page=' . CURRENT_ID;
	}
	elseif (\Input::get('do') == 'article' && ($objArticle = \ArticleModel::findByPk(CURRENT_ID)) !== null)
	{
		$this->Template->frontendFile = '?page=' . $objArticle->pid;
	}
	elseif (\Input::get('do') != '')
	{
		$event = new PreviewUrlCreateEvent(\Input::get('do'), CURRENT_ID);
		\System::getContainer()->get('event_dispatcher')->dispatch(ContaoCoreEvents::PREVIEW_URL_CREATE, $event);

		if (($strQuery = $event->getQuery()) !== null)
		{
			$this->Template->frontendFile = '?' . $strQuery;
		}
	}
}
```

and the `frontendFile` is not being set.

Currently for Contao 4.4 I built a rather ugly workaround using the parseTemplate hook ;-) But I don't quite know what a good solution could be. Do you have ideas?

Thanks!

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2580

This ensures scripts like the frontend preview can use the CURRENT_ID session parameter even in edge cases, e.g. after a login-redirect to an edit view as described in #2580

It is enough to update the `$id` variable which is passed to the session as `'CURRENT_ID'`.
The legacy constant `CURRENT_ID` always uses the GET parameter:

````php
\define('CURRENT_ID', (Input::get('table') ? $id : Input::get('id')));
````

An alternative solution is to update the `CURRENT_ID` in the session the same way as the constant: Always update the value if `Input::get('table')` has a value. I’m not sure about any undesirable side effects in this case though, especially in custom backend extensions.

Since this fix only updates the CURRENT_ID if there is no value in the session yet, this should not have any undesirable side effects while making sure there is a CURRENT_ID after a re-login.

### CODE REVIEW

**Comment by ausi** [comment]
Related issues: #2201, #2046

I think we should try to get rid of the `session.CURRENT_ID` completely as it causes problems, see linked issues above.

Setting `session.CURRENT_ID` to the `id=` parameter most probably sets the wrong ID (the one of the child table instead of the expected parent table).

**Comment by SeverinGloeckle** [comment]
Thanks for linking the other issues.

However, I'm confused about the idea of `session.current_ID` now:
- Is there any difference between `session.CURRENT_ID` and the `CURRENT_ID` constant (besides backwards compatibility)?
- Should the `CURRENT_ID` always point to the parent ID of the current record in any action views (e.g. edit view)?

I guess removing `session.CURRENT_ID` (and the `CURRENT_ID` constant in framework code) would be a good idea. 

Maybe there should be some backend or data container service that one could ask for:
- the current entity/table (returns the data container name)
- the ID of the current record (returns an ID)
- the ID of the current record in relation to an entity/table (returns either the ID of the current record if the entity is the current one or the ID of a parent if the entity is the parent)

The service could get this information from the query string and the DCA data.

Besides removing the `CURRENT_ID`, a lot of `Input::get('do')`, `Input::get('table')`, `Input::get('act')` could be standardized this way.

**Comment by ausi** [comment]
> Is there any difference between `session.CURRENT_ID` and the `CURRENT_ID` constant (besides backwards compatibility)?

I think so, yes. If the query does not contain `table=` https://github.com/contao/contao/blob/465dc7457491fa9066806def5f73e6dbdfac2b99/core-bundle/src/Resources/contao/classes/Backend.php#L346-L349

**Comment by SeverinGloeckle** [comment]
I have debugged the flow of the `PreviewUrlCreateEvent` and while the initial change fixed the issue described in #2580, setting `session.CURRENT_ID` does not seem to be the proper solution.

The problem with #2580 is:
- After a re-login there is no value in `session.CURRENT_ID`
- The `BackendPreviewListener` checks if an ID is available and **does not dispatch a preview event if there is no ID**.
- Thus, the (correctly working even without `session.CURRENT_ID`) preview listener of the calendar-bundle is not run.

BackendPreviewListener:
````php
if (!$id || !$do = $request->query->get('do')) {
    return $url;
}

...

$event = new PreviewUrlCreateEvent($do, $id);
$this->eventDispatcher->dispatch($event, ContaoCoreEvents::PREVIEW_URL_CREATE);
````

**Solution**
A fix for only #2580 – without touching `session.CURRENT_ID` – is to dispatch the `PreviewUrlCreateEvent` even with an ID of 0 since the calendar-bundle is creating the preview url correctly from the request.

**Fix for news-bundle**
In addition to events, the same bug is happening for news as well, since they also do not receive the `PreviewUrlCreateEvent` if there is an empty ID after a re-login in the edit view..

**Comment by aschempp** [comment]
thank you @SeverinGloeckle! We sure should dispatch the event regardless of any input ID. Maybe the best option would be to refactor the page and article handling into their own event listener anyway, wdyt @contao/developers?

**Comment by SeverinGloeckle** [comment]
> Maybe the best option would be to refactor the page and article handling into their own event listener anyway

Good idea.

I refactored the preview url generation from `BackendPreviewListener` to a new `PreviewUrlCreateListener`.

The `BackendPreviewListenerTest` got simpler this way too since the generation of the correct url is now tested in `PreviewUrlCreateListenerTest`.

File: core-bundle/src/Resources/config/listener.yml [hunk 1]
```diff
@@ -196,6 +196,14 @@ services:
         tags:
             - kernel.event_listener
 
+    contao.listener.preview_url_create:
```
aschempp: please use the class name as service name for any new services

aschempp (commented):
Looks very good to me! Just a minor nitpick 🙃

ausi (dismissed):
❤️

core-bundle/src/Resources/config/listener.yml
SeverinGloeckle: I have adjusted the defintion.

ausi (dismissed):

core-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Adds the page ID to the front end preview URL.
+     *
+     * @throws \RuntimeException
+     */
+    public function __invoke(PreviewUrlCreateEvent $event): void
+    {
+        if (!$this->framework->isInitialized() || !$event->getId() || !\in_array($event->getKey(), ['page', 'article'], true)) {
```
leofeyer: Why the `!$event->getId()` check here? We are not checking this in the other listeners, either, and the event is always initialized with an ID.

leofeyer:
@SeverinGloeckle Can you please enable the "Allow edits by maintainers" option?

core-bundle/src/EventListener/PreviewUrlCreateListener.php
SeverinGloeckle: The idea is to exit early in order to prevent the unnecessary database call to `PageModel` or `ArticleModel` via `$adapter->findByPk` further down if there is no ID.

SeverinGloeckle:
> Can you please enable the "Allow edits by maintainers" option?

I would like to do that. Unfortunately I sent the pull request via an organization repo. It seems that the function is not possible in this case (see https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/9).

I will send future pull requests via my personal account in order to avoid this.

core-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -0,0 +1,78 @@
...
+            $adapter = $this->framework->getAdapter(ArticleModel::class);
+
+            if (!$article = $adapter->findByPk($event->getId())) {
+                return;
+            }
+
+            $id = $article->pid;
+        }
+
+        $adapter = $this->framework->getAdapter(PageModel::class);
```
leofeyer, line 71:
```suggestion
        /** @var PageModel $adapter */
        $adapter = $this->framework->getAdapter(PageModel::class);
```

core-bundle/tests/EventListener/PreviewUrlCreateListenerTest.php
```diff
@@ -0,0 +1,160 @@
...
+        $event = new PreviewUrlCreateEvent('page', 42);
+
+        /** @var PageModel&MockObject $pageModel */
+        $pageModel = $this->mockClassWithProperties(PageModel::class);
+
+        $adapters = [
+            PageModel::class => $this->mockConfiguredAdapter(['findByPk' => $pageModel]),
+        ];
+
+        $framework = $this->mockContaoFramework($adapters);
```
leofeyer, line 41:
```suggestion
        $framework = $this->mockContaoFramework($adapters);

```

core-bundle/tests/EventListener/PreviewUrlCreateListenerTest.php
```diff
@@ -0,0 +1,160 @@
...
+
+        /** @var PageModel&MockObject $pageModel */
+        $pageModel = $this->mockClassWithProperties(PageModel::class);
+
+        $adapters = [
+            ArticleModel::class => $this->mockConfiguredAdapter(['findByPk' => $articleModel]),
+            PageModel::class => $this->mockConfiguredAdapter(['findByPk' => $pageModel]),
+        ];
+
+        $framework = $this->mockContaoFramework($adapters);
```
leofeyer, line 67:
```suggestion
        $framework = $this->mockContaoFramework($adapters);

```

leofeyer (changes requested):
Besides the suggestions I have made, I would also ask you to remove the `private $framework;` property and the `ContaoFramework $framework` constructor argument from the `BackendPreviewListener` class and the `services.yml` file.

core-bundle/src/EventListener/PreviewUrlCreateListener.php
m-vo: You should IMHO inline the assignment then `… || !($id = $event->getId()) || …` and reuse it further down. :slightly_smiling_face:

leofeyer: IMHO, we should remove the check completely, because the event is always initialized with an ID. Or at least add it to the other listeners as well.

SeverinGloeckle: The `PreviewUrlCreateListener` in the news and calendar bundle try to get the ID from the request themselves. Thus, they do not necessarily rely on the ID of the event - on the contrary, they have to work with an empty ID (that's what the original issue was about).

https://github.com/contao/contao/blob/5b719bf9ee5191d4b87389398386b0da38b512bd/news-bundle/src/EventListener/PreviewUrlCreateListener.php#L67-L75

The generation of a preview url for a page or article however has no custom logic to retrieve the ID. With an empty ID or an ID of 0 the listener will not be able to generate an url and will just create `WHERE tl_page.id=0` requests to the database. That's why I assume an early exit is appropriate here.

Inlining the `$id` variable seems like a good idea.

core-bundle/tests/EventListener/PreviewUrlCreateListenerTest.php
SeverinGloeckle: The blank line should add a separation between the "arrange" part and the "set"/"act" part of the test, correct?

core-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        $request = $this->requestStack->getCurrentRequest();
+
+        if (null === $request) {
+            throw new \RuntimeException('The request stack did not contain a request');
+        }
+
+        if ('article' === $event->getKey()) {
```
leofeyer, line 59: I would move the condition here

```suggestion
        if (!$id = $event->getId()) {
            return;
        }

        if ('article' === $event->getKey()) {
```

and also replace

```php
        if (null === ($newsModel = $this->getNewsModel($this->getId($event, $request)))) {
            return;
        }
```

with

```php
        if ((!$id = $this->getId($event, $request)) || (!$newsModel = $this->getNewsModel($id))) {
            return;
        }
```

in the other `PreviewUrlCreateListener` classes. Then we are consistently handling the case that `$id` is empty.

SeverinGloeckle: In my opinion, the ID should be checked before handling the request since nothing in the method will yield a result without an ID.

Maybe I am missing something and there is a reason to throw the `RuntimeException` if there is no request during an event with a key of `page` or `article ` but without an ID - instead of just ignoring the listener?

SeverinGloeckle: > in the other `PreviewUrlCreateListener` classes. Then we are consistently handling the case that `$id` is empty.

I like this, saving a useless call to the databse if there is no ID. I will update the request.

leofeyer: > In my opinion, the ID should be checked before handling the request since nothing in the method will yield a result without an ID.

You are right, checking the ID before retrieving the request makes more sense here. 👍

calendar-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -61,7 +61,7 @@ public function __invoke(PreviewUrlCreateEvent $event): void
             return;
         }
 
-        if (null === ($eventModel = $this->getEventModel($this->getId($event, $request)))) {
+        if ((!$id = $this->getId($event, $request)) || (!$eventModel = $this->getEventModel($id))) {
```
ausi, line 64:
```suggestion
        if (!($id = $this->getId($event, $request)) || !($eventModel = $this->getEventModel($id))) {
```
I think our coding style is to put the exclamation mark before the braces. (But I’m not 100% sure).

news-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -61,7 +61,7 @@ public function __invoke(PreviewUrlCreateEvent $event): void
             return;
         }
 
-        if (null === ($newsModel = $this->getNewsModel($this->getId($event, $request)))) {
+        if ((!$id = $this->getId($event, $request)) || (!$newsModel = $this->getNewsModel($id))) {
```
ausi, line 64:
```suggestion
        if (!($id = $this->getId($event, $request)) || !($newsModel = $this->getNewsModel($id))) {
```

ausi (approved):

calendar-bundle/src/EventListener/PreviewUrlCreateListener.php
leofeyer: No, it depends.

```php
// Before the braces if the ! denies more than just what's inside the braces (it refers to instanceof in this case)
if (!($foo = $this->getFooPage()) instance of PageModel)

// Before the braces if the ! denies two or more conditions inside the braces
if (!(foo() && bar()))

// But inside the braces if the ! denies a variable assignment
if ((!$id = $this->getId()) || foo())
```

ausi: K, sorry for the interruption ☺️

ausi (dismissed):

core-bundle/src/EventListener/PreviewUrlCreateListener.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Adds the page ID to the front end preview URL.
+     *
+     * @throws \RuntimeException
+     */
+    public function __invoke(PreviewUrlCreateEvent $event): void
+    {
+        if (!$this->framework->isInitialized() || !($id = $event->getId()) || !\in_array($event->getKey(), ['page', 'article'], true)) {
```
SeverinGloeckle:
```suggestion
        if (!$this->framework->isInitialized() || (!$id = $event->getId()) || !\in_array($event->getKey(), ['page', 'article'], true)) {
```

This should be changed as well I guess?

See https://github.com/contao/contao/pull/3416/files#r709073556

leofeyer: Yes please.

leofeyer (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3419, opened 2021-09-04
Add options to customize the layout inheritance for pages

SeverinGloeckle:
Add an option to configure the template inheritance from parent pages to child pages.

Possible options are:
- use layout on subpages (current behaviour, default & fallback behaviour)
- use layout on this page only
- use a custom layout for subpages

> contao-layout-inheritance

It is much easier to implement special overview pages or more complex layout structures via the page structure in this way:
- If a single overview page requires a special layout, define `use layout on this page only` – subpages then use the layout of the overview page’s parents (the current page is skipped when walking through the parents looking for a layout).
- If all subpages of a page should have the same layout which is not the layout of the parent page, set `use a custom layout for subpages` and select the layout for the subpages.

Overall, this decouples layout inheritance somewhat from the page tree (which is primarily used for menu structure).
Another advantage is that the default inheritance of the layout to subpages is displayed directly to editors by the default  option `use layout on subpages`.

**Handling missing configuration:**
This new implementation still does not resolve a bad/missing configuration:
- If a parent page has `includeLayout` set but has no `layout`, the empty `layout` is inherited by the current page.
- If a parent page has `layoutInheritance` set to `use a custom layout for subpages` but has no `subpagesLayout`, the empty `subpagesLayout` is inherited by the current page.

Since the layout select widgets do not have a blank option, both cases should not happen when editing a page mindfully, but could happen after a `submitOnChange` without a manual save after.

One could argue that in case of a missing layout value, the layout should not be inherited from this parent.

**Target Contao version:**
As a feature, this should probably be added to the latest Contao version. But since it is a progressive enhancement which defaults to the current behavious, it could be added to Contao 4.9 rather easily. The current pull request is based on Contao 4.9.

Issues on this topic:
- #1797
- https://github.com/contao/core/issues/1270 (very old)
- https://github.com/contao/core/issues/4868 (very old)


ausi:
How about moving the options from `layoutInheritance` to the `subpagesLayout` field?

The options of it would then be something like:

1. Use this layout on subpages (the default)
2. Use the parent layout on subpages
3. Layout A
4. Layout B
5. …

You won’t need a subpalette then and the usability would be greater this way too I think.

SeverinGloeckle:
Good suggestion. I like the better usability for editors and the possibility to go without the `submitOnChange` on the select widget.

However, from a developer's point of view I don't like storing both arbitrary strings (`inherit`, `disable`/`parent`) and layout IDs in the same field. We would lose the relation information of the DCA field as well.

Maybe we can make things even easier:
- Only add one new select widget `subpagesLayout`.
- The widget only provides layouts as options, but a blank value is allowed.
- If the value is empty, no layout will be inherited from the current page.
- If a `subpagesLayout` is selected, this layout will be inherited by subpages.

This way the `selectLayout` field remains technically clean while giving the editor multiple possibilities to adjust the layout inheritance:
- Use the same layout as the parent page by selecting the same layout for subpages.
- Ignore the current page layout for subpages by selecting no layout for subpages.
- Define a custom layout for subpages by selecting another layout.

However, this makes adding the feature to an existing installation more difficult since the `subpagesLayout` is not hidden behind a flag anymore. 

The feature should then come along with a migration that prefills the new `subpagesLayout` field with the value of `layout` for all pages with `includeLayout` to make sure the previous inheritance is kept intact.

I guess the feature should also be moved to Contao 4.13 then.

What do you think?

ausi:
> I guess the feature should also be moved to Contao 4.13 then.

Yes. Any version of this feature would be targeted against 4.13 anyways as we only add bugfixes to 4.9 and 4.12.

> * Only add one new select widget `subpagesLayout`.
> * The widget only provides layouts as options, but a blank value is allowed.

I like that Idea. @contao/developers what do you think?

> The feature should then come along with a migration that prefills the new `subpagesLayout` field with the value of `layout` for all pages with `includeLayout` to make sure the previous inheritance is kept intact.

Agree.

leofeyer:
> > * Only add one new select widget `subpagesLayout`.
> > * The widget only provides layouts as options, but a blank value is allowed.
> 
> I like that Idea. @contao/developers what do you think?
 
Sounds good. 👍 

> > The feature should then come along with a migration that prefills the new `subpagesLayout` field with the value of `layout` for all pages with `includeLayout` to make sure the previous inheritance is kept intact.
> 
> Agree.

Why is this necessary? Can we not keep the current behavior if the `subpagesLayout` field is empty?

ausi:
> Can we not keep the current behavior if the `subpagesLayout` field is empty?

If we use the blank option for “Ignore the current page layout for subpages by selecting no layout for subpages.” we have to use a migration. But this migration would IMO be straight forward and schould be very robust, so not causing any problems.

leofeyer:
> If we use the blank option for “Ignore the current page layout for subpages by selecting no layout for subpages.”

This is not very intuitive, is it? I would expect not selecting anything to mean "no special treatment for subpages" and not "do not apply the page layout to subpages".

ausi:
> This is not very intuitive, is it?

The blank value would have an appropriate label, so it should be clear what it does. Making the empty value meaning “inherit from the parent” is also conceptionally meaningful I think, as it is the same way with other fields.

m-vo:
> However, from a developer's point of view I don't like storing both arbitrary strings (inherit, disable/parent) and layout IDs in the same field. 

I agree. It's the same with magical numbers like "0", "-1". You cannot use foreign keys for these…

We should IMHO add a boolean `propagateLayout` (default: `true`) to the model that basically says: should this layout be used for my child pages (feature 1). Then a new `subpagesLayout` field (feature 2) could become a real nullable foreign key field in the future.

The combination "don't propagate", "use a layout for sub pages" could either mean this only affects direct children or the sub page layout is ignored.

Could be something like this:

| `layout` | `subpagesLayout` | `propagateLayout` | Effect                                                                       |
|----------|------------------|-------------------|------------------------------------------------------------------------------|
| ''/null  | null             | true              | old behavior, inherit layout (e.g. 1)                                        |
| 2        | null             | true              | old behavior, this page + sub pages will get layout 2                        |
| 2        | 3                | true              | this page will get layout 2, all sub pages layout 3                          |
| 2        | null             | false             | this page will get layout 2, sub pages will inherit from further up (e.g. 1) |
| 2        | 3                | false             | this page will get layout 2, only direct sub pages will get layout 3, …      |

I'd still use two selects in the UI with the additional "virtual options" and just handle storing/loading them in DCA callbacks.

asaage:
Why not keep it simple?
The ```subpagesLayout``` field could just be a checkbox/boolvalue wich decides if the current layout should be passed to subpages (default) or if it is a onetime-skip-inheritance situation here subpages would then inherit their layout from the parentpage instead of the page itself.

SeverinGloeckle:
> This is not very intuitive, is it? I would expect not selecting anything to mean "no special treatment for subpages" and not "do not apply the page layout to subpages".

I think it is comprehensible with the correct labeling:
- The help tip text of `Assign a layout` gets changed from _Assign a page layout to the page and its subpages._ to _Assign a page layout to the page and **optionally** to its subpages._.
- The new `subpagesLayout` is labeled as `Layout for subpages` with the tip `This layout will be assigned to the subpages of this page`.

If no layout for subpages is selected, the user should not rightfully assume that the subpages inherit the layout from the current page. However, since it was done this way for years, I see how the new behaviour could confuse experienced users.

However, I think this hits the sweet spot between _easy configuration_ and _intuitive behaviour_.

SeverinGloeckle:
> Why not keep it simple?
> The `subpagesLayout` field could just be a checkbox/boolvalue wich decides if the current layout should be passed to subpages (default) or if it is a onetime-skip-inheritance situation here subpages would then inherit their layout from the parentpage instead of the page itself.

This would remove the option to define a parent page with `Layout A` and pre-defining `Layout B` for all subpages (see https://github.com/contao/core/issues/4868).

asaage:
> This would remove the option...

@SeverinGloeckle agree

SeverinGloeckle:
> The combination "don't propagate", "use a layout for sub pages" could either mean this only affects direct children or the sub page layout is ignored.

This would be a very flexible solution that also adds the possiblility to target only direct subpages of a page.

However, I think this may be a little bit too complex and error-prone for editors. Moving pages around in the page tree could lead to unexpected layout choices that are hard to spot from the page tree alone (without some kind of special "layout inheritance view").

Without the feature to propagate a layout only one level down, we are back to:
1. use the same layout for subpages
2. do not set a layout for subpages (inherit from a parent further up)
3. use a custom layout for subpages

Since option 1 is the same as option 3 with the same layout selected for both `layout` and `subpagesLayout`, a single new widget would suffice.

I can work on this feature and:
- refactor the proposal to a single `subpagesLayout` select with a meaningful label.
- adjust the layout inheritance in `PageModel::loadDetails()`.
- add a migration to prefill the `subpagesLayout` to mimic the current inheritance.
- rebase to Contao 4.13.

core-bundle/src/Migration/Version413/PageSubpagesLayoutMigration.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+
+        if (!$schemaManager->tablesExist(['tl_page'])) {
+            return false;
+        }
+
+        $pageColumns = $schemaManager->listTableColumns('tl_page');
+
+        return !isset($pageColumns['subpageslayout']);
```
ausi:
```suggestion
        return !isset($pageColumns['subpageslayout']) && isset($pageColumns['layout'], $pageColumns['includeLayout']);
```

core-bundle/tests/Contao/PageModelTest.php
```diff
@@ -63,6 +65,7 @@ protected function tearDown(): void
         parent::tearDown();
 
         Registry::getInstance()->reset();
+        unset($GLOBALS['TL_MODELS']['tl_page']);
```
ausi:
```suggestion
        unset($GLOBALS['TL_MODELS']);
```

m-vo:
Thinking about this again, I guess these are the most common usage scenarios (correct me if I'm wrong):

1. My page has a `layout_default` layout applied to its root page, so all pages get it by default. I want to apply `layout_special` to a certain page without affecting any other page (including its children).
1. I want to make sure every new child page of a page "things" gets the `layout_thing` layout. The "things" page should either just inherit the layout, set its own or also use the children's layout.

If I see this correctly, the first one (effectively suppressing inheritance) would now work but I'd have to explicitly select the originally inherited layout again for the sub pages? Not sure I like this mechanic (while I do for the other case!).

ausi:
> If I see this correctly, the first one (effectively suppressing inheritance) would now work but I'd have to explicitly select the originally inherited layout again for the sub pages?

I don’t think so. You would only select the layout and leave the subpageLayout empty in this case.

SeverinGloeckle:
You can change the layout for a single page whithout affecting the subpages if you leave the `subpagesLayout` empty.

The actual logic in `PageModel` is simple:
````php
if ($this->layout === false && $objParentPage->includeLayout && $objParentPage->subpagesLayout)
{
    $this->layout = $objParentPage->subpagesLayout;
}
````
- Only change the layout if the current page has no layout yet (same as before)
- Only change the layout if the parent page has the `includeLayout` flag (same as before)
- Only change the layout if the parent page has a `subpagesLayout` (new)
- Add the `subpagesLayout` to the page (new, was `layout` before)

You only have to select both layouts if:
- the subpages should inherit the layout from the current page (new behaviour, supported by migration)
- you want to use a custom layout for subpages (new feature)

I can still see how experienced users may stumble on the need to select the same layout for subpages again, however the added flexibility should be worth the tradeoff.

**Special case for root pages**
Thinking about it, maybe we should add a special treatment for root pages since a root page does not need its own layout and only provides layouts for subpages.

So the palettes of `root` and `rootfallback` don’t need the `includeLayout` option but only the `subpagesLayout` select, (which allows a blank option).

In fact, one could even argue that a root page **should always** provide a fallback layout for its subpages, but that’s a rather opinionated discussion.

**Layout inheritance for logout page**
For the sake of completeness, we should probably also add the `subpagesLayout` for the `logout` palette, since there is no way to define layouts for subpages of a logout page (for whatever reason).

--

What do you think?

**Comment by leofeyer** [comment]
I agree with almost everything above except this:

> You only have to select both layouts if the subpages should inherit the layout from the current page (new behaviour, supported by migration)

Here is an alternative approach:

| Subpages select | Meaning |
| --- | --- |
| Empty | Inherit the page layout to the subpages (just like it is now) |
| Do not inherit | Special option `-1` → use the page layout for the page only and do not inherit it to the subpages |
| Layout ID | Use the page layout for the page only and inherit the selected layout to the subpages |

This would preserve the UX that our users are used to and it would not require a migration.

WDYT?

**Comment by m-vo** [comment]
> You can change the layout for a single page whithout affecting the subpages if you leave the subpagesLayout empty.

I see. From how Contao worked ever since I would have expected that not selecting anything in the select (for sub pages) would mean they would inherit the layout setting of the current page. :upside_down_face: 

>  Special option -1 → use the page layout for the page only and do not inherit it to the subpages

:+1: for the labels + behavior but please no magic numbers. This won't allow FKs in the future.

**Comment by leofeyer** [comment]
We have already done exactly this in Contao 4.12 with the "guests group". 🤷‍♂️

**Comment by ausi** [comment]
> This would preserve the UX that our users are used to and it would not require a migration.
> 
> WDYT?

IMO most of the time the current UX is not optimal because in many use cases you want to change only the layout and not the subpage-layout.

I would prefer how it currently works in this pull request.

**Comment by ausi** [comment]
> So the palettes of `root` and `rootfallback` don’t need the `includeLayout` option but only the `subpagesLayout` select, (which allows a blank option).

Agree.

> For the sake of completeness, we should probably also add the `subpagesLayout` for the `logout` palette

Agree.

**Comment by aschempp** [comment]
Please also consider the current user permissions and existing custom page palettes. If we add a new dropdown for the subpage, this field will most likely be invisible in both these cases. If we add a checkbox/radio/select to change the inheritance flow, not having it should not affect existing behaviour.

**Comment by ausi** [comment]
> Please also consider the current user permissions…

Yes, we should add that to the migration. Every user that currently has access to the `layout` should get access to the `subpageLayout` as well.

> …and existing custom page palettes.

IMO we do not have a BC promise for custom palettes.

**Comment by asaage** [comment]
Just thinking: Which layout would a subPage get if the parentPage has ```subpageLayout``` set but the subPage has set it's own layout? The order of validation needs to be in the manual then and i am pretty sure it will cause issues with permissions no matter how it's implemented.

**Comment by SeverinGloeckle** [comment]
> Please also consider the current user permissions...

Good point. I will adjust the migration accordingly.

> ... and existing custom page palettes

It would be pretty difficult to add the new field to custom palettes without knowing their intended use. However, an additional note in `UPGRADE.md` about the new field and the potential need to update custom palettes for `tl_page` sounds like a good idea.

> Just thinking: Which layout would a subPage get if the parentPage has subpageLayout set but the subPage has set it's own layout?

Any (sub-)page that defines a custom layout via `includeLayout` will use its own layout, regardless of parent pages – the same way it has been for now. I think this behavious is really important when moving pages around in the page tree.

**Comment by SeverinGloeckle** [comment]
From what I understand, the question of how to properly manage subpage layouts in the backend has not yet been resolved to everyone's satisfaction.

The variants are:

### (1) `layout` & `subpagesLayout`

Good:
- Simple administration in the backend
- Simple technical implementation

Bad:
- Effect of an empty `subpagesLayout` contradicts the old way of inheritance
- Migration for `tl_page` necessary

### (2) `layout` & `subpagesLayout` with additional options `propagate` & `do not propagate`

Good:
- Simple administration in the backend
- Clearly defined effect through `propagate` and `do not propagate` options

Bad:
- Unclean data in `subpagesLayout` field

### (3) `layout`, `propagateLayout` & `subpagesLayout`

Good:
- Explicit explanation of template logic

Bad:
- More complex management in backend (3 widgets)
- Needs a subpalette to prevent setting a `subpagesLayout` together with `do not propagate` 
---

One could also imagine a mixed implementation of the second variant with the individual options `propagate` and `do not propagate`, which are however not stored in the field `subpagesLayout` via save and load callbacks, but in a field `propagateLayout` which is not visible in the palette (something like @m-vo  suggested above).

However, the corresponding callbacks would then have to be maintained in the long run.

I still prefer variant 1 as the cleanest solution but have no hard opinion on this.

**Comment by ausi** [comment]
> I still prefer variant 1 as the cleanest solution

Fully agree.
I would say that *“Effect of an empty `subpagesLayout` contradicts the old way of inheritance”* might even be a benefit because the current/old way of inheritance is a problem for many users.

**Comment by SeverinGloeckle** [comment]
If we think about the select options for the subpages layout dropdown in isolation, I guess we can all agree that these would be the most understandable labels:
- Inherit page layout
- Do not change
- Layout A
- Layout B
- ...

This way, the user doesn’t have to guess the result of an empty value and the whole inheritance logic is somewhat explained by the options.

I tried adding custom logic to handle these custom options in addition to the layout IDs in `subpagesLayouts` without adding magic values to the database field, but this requires three callbacks:
1. options_callback to get the layout IDs merged with the custom options
2. save_callback to store the custom options in another field
3. load_callback to restore the current value for custom options

The callbacks included a lot of boilerplate code and manual option comparisons, which did not feel right.

### Proposal: new DCA option `foreignOptions`
I extended the `Widget` class to allow merging of custom options from another DCA field and extracted all the logic for reading and saving the options to `Widget`.

The backend configuration of subpages layouts now can look like this:

> contao-foreign-options-layout

The logic in `PageModel::loadDetails()` is still simple and just checks the `$objParentPage->layoutPropagation` value. While a boolean check would be enough for now, using real options for `layoutPropagation` provides the benefit of adding more options in the future (e.g. propagate a single level, force propagation, propagate under conditions, ...).

The two custom options for the subpages layout are stored in an additional field `tl_page.layoutPropagation`.

The default value is `layoutPropagation=propagate` which is the current behaviour

The definition in tl_page looks like this (note the `foreignOptions` configuration):
````php
'subpagesLayout' => array
(
	...
	'options_callback'        => array('tl_page', 'getPageLayouts'),
	'foreignOptions'          => array(
		'layoutPropagation' => array('prepend' => true, 'reference' => &$GLOBALS['TL_LANG']['tl_page']['subpagesLayout'])
	),
	...
),
'layoutPropagation'       => array
(
	'options'                 => array('propagate', 'disable'),
	'reference'               => &$GLOBALS['TL_LANG']['tl_page']['layoutPropagation'],
	'sql'                     => "varchar(32) NOT NULL default 'propagate'"
),
````

Notes:
- The additional options are read from the foreign field properly following option_callback, foreignKey and reference.
- The option merging works for both associative options (e.g. foreign keys or custom option callbacks) and flat option arrays.
- The foreign options can be prepended or appended to the original options of the field.
- The field can specify custom labels via `reference` for the included foreign options (useful if the context changes, e.g. from `layoutPropagation` to `subpagesLayout`).
- If a foreign option is used, its value is saved to the corresponding field. The original field gets an empty value.
- Changing the value will add the foreign option field in the version diff view (e.g. `layoutPropagation` instead of / additionally to `subpagesLayout`).
- It is possible to add multiple `foreignOptions` to a field while the value would always be saved to the correct field.
- `foreignOptions` must be defined within the same DCA as the original field.
- Unknown options will always be saved to the original field.

In fact I would prefer if the modification of the foreign field value could be done in the  `save` method of a `DataContainer`. Maybe this can be optimized for Contao 5 if the `foreignOptions` are well accepted.


### Next steps

This is a first draft to enhance option widgets (select, checkboxes, radio) with custom options from another field. In the future, this could be used for other DCA fields as well (image sizes come to mind).

Since all new logic is tied to the `foreignOptions` setting, the changes do not interfer with any current widget logic.

If this turns out to be the way to go, the next steps would be:
- Add tests for the `foreignOptions` logic to `WidgetTest`
- Remove the migration to prefill `subpagesLayout`
- Add the missing migration for user access rights
- Adjust special page palettes
- Add the `foreignOptions` field to the developer documentation


Any input is welcome!

**Comment by SeverinGloeckle** [comment]
It would be nice if this feature can make it into Contao 4.13.

The last comment above provides a good overview of the current state as well as open tasks. I can take care of the necessary adaptations if it is clear that the basic implementation is what we want.

If the concept or the implementation of the `foreignOptions` is debatable, we may not add it to the developer documentation yet and just use it internally for the page layout inheritance.

File: core-bundle/src/Migration/Version413/PageSubpagesLayoutMigration.php [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+
+        if (!$schemaManager->tablesExist(['tl_page'])) {
+            return false;
+        }
+
+        $pageColumns = $schemaManager->listTableColumns('tl_page');
+
+        return !isset($pageColumns['subpageslayout']) && isset($pageColumns['layout'], $pageColumns['includeLayout']);
```
aschempp: Schema manager column names are all lowercase.

```suggestion
        return !isset($pageColumns['subpageslayout']) && isset($pageColumns['layout'], $pageColumns['includelayout']);
```

core-bundle/src/Migration/Version413/PageSubpagesLayoutMigration.php
```diff
@@ -0,0 +1,67 @@
...
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->getSchemaManager();
```
aschempp:
```suggestion
        $schemaManager = $this->connection->createSchemaManager();
```

aschempp (changes requested):
I really like the feature this PR tries to add. But I think we're making it harder than necessary. In all my time building Contao pages, the **only** thing I ever wanted was to assign a different layout to multiple subpages than the current layout. Sure there are exceptions where I know it's gonna be the same layout as the parent page – but there's really no point in adding that much complexity imho.

The whole feature could be very much simplified by adding a new field (as proposed). Blank option and blank value would be the current behaviour, or select a layout for subpages that will be inherited by `PageModel::loadDetails()` instead of the current field.
Using that would be BC on the database, BC on the user behaviour and BC on the user permissions.

PS: I don't see the point in adding a "adopt" feature to the widget class, isn't it just a regular callback with values?

SeverinGloeckle:
Thanks for the feedback @aschempp.

> In all my time building Contao pages, the only thing I ever wanted was to assign a different layout to multiple subpages than the current layout. Sure there are exceptions where I know it's gonna be the same layout as the parent page – but there's really no point in adding that much complexity imho.

I have a different view. Imagine this example:
- A website uses a generic layout A with a sidebar.
- The news overview page of the website displays a grid of news without a sidebar, using layout B.
- The news section has additional subpages (e.g. an archive, press releases, ...) which should use the sidebar layout A.

Without a "Do not change layout of subpages" option, you'd have to either re-set the layout A for each subpage or define layout A as a child layout on the news overview page. The latter sounds tempting, but if you change the base layout of the website from layout A to layout C, you'd have to remember to adjust the child layout on the news overview page as well.

----

In my opinion, there are multiple issues to this pull request:

## Make layout inheritance more flexible
In the course of this topic, the various possibilities have already been discussed, and the mapping originally proposed by @m-vo in https://github.com/contao/contao/pull/3419#issuecomment-913441710 via `layout`, `subpagesLayout` and `propagateLayout` has proven to be the best solution in my opinion.

This way there is a clean data structure without magic numbers and full flexibility in the configuration.

## Interface for inheritance options
The configuration of the options could be mapped accordingly via three widgets. However, since certain combinations make very little sense and in order to enable an interface that is as streamlined as possible (without subpalettes), the special options "inherit page layout" and "do not change" should be integrated into the selection of a layout for subpages.

## Concept of "foreign options"
Merging several database fields into one widget is, in my opinion, a very good tool in the DCA toolbox in order to avoid magic numbers in the future and to be able to use actual foreign keys on a database level more easily.

## Implementation of "foreign options"
The technical implementation of the "foreign options", on the other hand, is rather tricky.

For me, the biggest weakness of the current implementation is the overloading of `$_POST` when changing data in a "foreign option field" to notify the DataContainer that the value of a field has changed (so that a corresponding version is generated).

Maybe we can come up with a better solution here.

> PS: I don't see the point in adding a "adopt" feature to the widget class, isn't it just a regular callback with values?

As described in https://github.com/contao/contao/pull/3419#issuecomment-917958673, this would require three callbacks with a lot of repeated code. An integrated solution within the widget seems more useful to re-use `foreignOptions` in other places as well (as long as the concept is accepted).

core-bundle/src/Migration/Version413/PageSubpagesLayoutMigration.php
SeverinGloeckle: Good point. However, prefilling the `subpagesLayout` will not be necessary with the current concept of `foreignOptions` and this migration would be removed. 

I will keep this in mind for possibly adding a migration to adjust the access rights for users and user groups.

core-bundle/src/Migration/Version413/PageSubpagesLayoutMigration.php
SeverinGloeckle: Good point. However, prefilling the `subpagesLayout` will not be necessary with the current concept of `foreignOptions` and the migration could be deleted. 

I will keep this in mind for adding a migration to adjust the access rights for users and user groups.

aschempp:
> but if you change the base layout of the website from layout A to layout C, you'd have to remember to adjust the child layout on the news overview page as well.

yes, and I'm totally fine with that. I don't think the parent layout is changed that often, and changing one more field sounds good to me. And it will make this implementation very simple and not unnecessarily complex. Also, you currently have to select the supage layout **for every new subpage**, which is the real painpoint to me that should be solved (and does with this).

SeverinGloeckle:
So there's two ways to go:

1. only allow to define a different layout for child pages or no layout for child pages
	- if no layout is defined, the current page's layout will be used for subpages (as in the past)
	- advantage: simple change that covers a lot of use-cases
	- ambiguous UI: an empty "subpages layout" field implicitly means "propagate the current layout to subpages"
	- missing feature: you can not set subpages to keep using the regular layout from the current page's parent page

2. same as above but also add options "use current layout" and "do not change layout" for the subpages layout (see screenshot and description in https://github.com/contao/contao/pull/3419#issuecomment-917958673)
	- advantage: all inheritance options are clearly defined
	- additional feature: the special use-case of "do not change" the layout of subpages is covered
	- disadvantage: requires either some way to merge the two additional options with the available layout options (e.g. `foreignOptions`) or multiple widgets

Note on migrations: In both cases, a migration for configured layouts or for access rights should not be necessary as the current behaviour is the fallback and thus non-admin user don't miss a feature without the new subpages layout field.

-----

I'd still argue for the second way, but definitly prefer a simpler feature over no feature.

Maybe we can get some votes on this:
- 👍  for the first way
-  🎉  for the second way

leofeyer:
As discussed in the Contao call, we only want to add the second select menu (which would be the first of the two ways mentioned above).

SeverinGloeckle:
As discussed in the Contao Call, the first way will be implemented:

> only allow to define a different layout for child pages or no layout for child pages
> - if no layout is defined, the current page's layout will be used for subpages (as in the past)
advantage: simple change that covers a lot of use-cases

The blank label of the `Subpages layout` widget should be labelled `Inherit page layout`.

I will adjust this pull request accordingly.

SeverinGloeckle:
I have simplified the changes by removing `tl_page.layoutInheritance` and the foreign options concept.

A migration to extend the access rights for users with access to `tl_page.layout` to include the new `tl_page.subpagesLayout` would be possible, but is probably not necessary, since these users can still edit page layouts as before and only miss the new option to explicitly define a layout for subpages.

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -191,7 +191,8 @@
 	(
 		'autoforward'                 => 'jumpTo',
 		'protected'                   => 'groups',
-		'includeLayout'               => 'layout',
+		'createSitemap'               => 'sitemapName',
```
ausi:
```suggestion
```

This looks like a left over from Contao 4.9?

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -590,6 +591,17 @@ static function ($value)
 			'sql'                     => "int(10) unsigned NOT NULL default 0",
 			'relation'                => array('type'=>'hasOne', 'load'=>'lazy')
 		),
+		'subpagesLayout' => array
+		(
+			'exclude'                 => true,
+			'search'                  => true,
+			'inputType'               => 'select',
+			'foreignKey'              => 'tl_layout.name',
+			'options_callback'        => array('tl_page', 'getPageLayouts'),
```
fritzmg: This should be a service.

fritzmg (changes requested):

core-bundle/src/Resources/contao/dca/tl_page.php
ausi: But the callback already existed?

fritzmg: Ah yes, but without caching the results.

core-bundle/src/Resources/contao/dca/tl_page.php
SeverinGloeckle: Good catch, thank you.

core-bundle/src/Resources/contao/dca/tl_page.php
SeverinGloeckle: Since the existing callback cannot be removed for backwards compatibility, I assume that reusing this callback is preferable to adding a new service that does the same thing.

The caching is added to avoid querying the database twice for the options of both `layout` and `subpagesLayout`.

Both calls could be replaced by a `LayoutOptionsListener`, while the existing `tl_page::getPageOptions` is kept for backwards compatibility. Please let me know if that is what is preferred.

ausi (dismissed):

core-bundle/src/Resources/contao/dca/tl_page.php
ausi: > Please let me know if that is what is preferred.

I’d be happy either way. IMO you could integrate the caching into the `getPageLayouts()` method so that we don’t create a new public `loadPageLayouts()` method that needs BC in the future.

fritzmg: These are not covered by our BC promise. We have already removed such methods and even whole `tl_` classes in the past.

aschempp (commented):
I don't like the new `loadPageLayouts` method either. A callback would be nice, but if not then the code should be inside the existing method. Apart from that thats now ✅ and very clean and simple 😎

core-bundle/src/Resources/contao/dca/tl_page.php
ausi: The the `LayoutOptionsListener` would probably be the best option then, as all core developers seem to be fine with that ☺️

SeverinGloeckle:
I have added a new `LayoutOptionsListener` in https://github.com/contao/contao/pull/3419/commits/080503ce7c9ca7e0628fbc8a3ca98516c96f3bb9.

The listener still caches the layout options for subsequent calls so both `layout` and `subpagesLayout` get their options through a single database query.

Following the advice by @fritzmg I also removed the existing `tl_page::getPageLayout` method.

ausi (dismissed):
❤️

core-bundle/src/EventListener/DataContainer/LayoutOptionsListener.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Contao\CoreBundle\EventListener\DataContainer;
+
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @Callback(table="tl_page", target="fields.layout.options")
+ * @Callback(table="tl_page", target="fields.subpagesLayout.options")
+ */
+class LayoutOptionsListener
```
aschempp: This should implent `ResetInterface` (and be tagged accordingly) to reset the layouts on kernel shutdown.

aschempp (changes requested):
Just one minor nitpick before ✅ 😊

core-bundle/src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -431,6 +437,9 @@
       <trans-unit id="tl_page.map_never">
         <source>Show never</source>
       </trans-unit>
+      <trans-unit id="tl_page.layout_inherit">
```
m-vo, line 440: This should probably be camelCase.

m-vo (dismissed):
LGTM besides the [comment](https://github.com/contao/contao/pull/3419/files#r785416557) about the `ResetInterface`. :+1:

core-bundle/src/EventListener/DataContainer/LayoutOptionsListener.php
SeverinGloeckle: Good point, implemented in https://github.com/contao/contao/pull/3419/commits/ad9dbb73cba25a9576274e0d4a1eadbe43797e19.

core-bundle/src/Resources/contao/languages/en/tl_page.xlf
SeverinGloeckle: I went with snake case to be in line with labels like `map_never` (the line above), somewhat indicating the `inherit` part is an option of `layout`.

But either is fine with me.

ausi (dismissed):

fritzmg (dismissed):

aschempp (dismissed):

m-vo (dismissed):

leofeyer:
@SeverinGloeckle Since you did not allow maintainers to edit the PR, please review and merge https://github.com/pixlscript/contao/pull/1.

SeverinGloeckle:
This rather old request unfortunately still belongs to an organisation repository, which is why an "edit by maintainers" is not possible.

I am fine with renaming `subpagesLayout` to `subpageLayout` and have merged your request.

m-vo (approved):

ausi (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3423, opened 2021-09-05
Convert allowed attributes to lowercase

### LINKED ISSUE #3409: HTML attributes - case sensitivity
The registration of _allowed HTML attributes_ allows lowercase and uppercase input. The output ist just lowercase. Trying to use the HTML attribute _viewBox_ (contains the upper case letter _B_) was confusing, because it wasn't found in the output.

Or the other way round, the HTML attributes should only be entered and saved in lower case in the settings, because the HTML Attributes are case-insensitive as @ausi  mentioned.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3409

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_settings.php
+++ b/core-bundle/src/Resources/contao/dca/tl_settings.php
@@ -123,6 +123,21 @@ $GLOBALS['TL_DCA']['tl_settings'] = array
 					return $varValue;
 				},
 			),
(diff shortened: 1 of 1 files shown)


Toflar (approved):

bytehead (approved):

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #3424, opened 2021-09-05
Fix link_url back insert tag

### LINKED ISSUE #3364: Add warning in system log for {{link_url::back}}
`{{link_url::back}}` doesn’t work anymore as it gets encoded as `javascript%3Ahistory.go(-1)`, we should probably add a warning to the system log if it gets used somewhere.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3364

Implemented suggestion from https://github.com/contao/contao/issues/3364#issuecomment-903572563

But I’m not sure if this is a good idea 🤔

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -386,13 +386,22 @@ class InsertTags extends Controller
 					// Back link
 					if ($elements[1] == 'back')
 					{
-						$strUrl = 'javascript:history.go(-1)';
+						$strUrl = '#_back';
 						$strTitle = $GLOBALS['TL_LANG']['MSC']['goBack'];
 
 						// No language files if the page is cached
 						if (!$strTitle)
 						{
 							$strTitle = 'Go back';
+						}
+
+						$strBackScript = "<script>document.documentElement.addEventListener('click',function(e){"
+							. "e&&e.target&&e.target.matches('a[href$=\"#_back\"],a[href$=\"#_back\"] *')&&(history.go(-1),e.preventDefault&&e.preventDefault())"
+						. "},true)</script>";
+
+						if (!\in_array($strBackScript, $GLOBALS['TL_HEAD'] ?? array(), true))
+						{
+							$GLOBALS['TL_HEAD'][] = $strBackScript;
 						}
 
 						$strName = $strTitle;
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -395,14 +395,9 @@ class InsertTags extends Controller
 							$strTitle = 'Go back';
 						}
 
-						$strBackScript = "<script>document.documentElement.addEventListener('click',function(e){"
+						$GLOBALS['TL_HEAD']['contao_back_insert_tag'] = "<script>document.documentElement.addEventListener('click',function(e){"
 							. "e&&e.target&&e.target.matches('a[href$=\"#_back\"],a[href$=\"#_back\"] *')&&(history.go(-1),e.preventDefault&&e.preventDefault())"
 						. "},true)</script>";
-
-						if (!\in_array($strBackScript, $GLOBALS['TL_HEAD'] ?? array(), true))
-						{
-							$GLOBALS['TL_HEAD'][] = $strBackScript;
-						}
 
 						$strName = $strTitle;
 					}


leofeyer:
Me neither. 🤷‍♂️

m-vo:
I like this solution way more than hard coding an exception into the sanitizer but I'm also not sure about it… Do we want to keep this functionality alive in the core in the future? If not, let's at least trigger a deprecation warning.

IMHO users could just use their browser's back button for that. If a page adds history items while using it (e.g. with a JS filter list or something) you would maybe want to provide a static back link (with an actual URL). So I don't really see the benefit of this insert tag.

Fun fact: History back/back button behavior isn't that consistent across browsers and has lots of intricacies.  https://www.youtube.com/watch?v=W6lvQvdeF8U :slightly_smiling_face:

SeverinGloeckle:
I think this is a good fix:
- It fixes the JavaScript-only backlink with a JavaScript solution.
- It only adds the global JavaScript (once) if there is a `{{link_url::back}}` instert tag on the current page.
- The JavaScript seems robust and catches the regular use case of a back link with various markup.
- The global event listener is attached to the document so links in applications are handled properly (e.g. a Vue enhanced page).
- If there are more complex event listeners on anchors, there could be problems with the listener. But the same was true for the inline `javascript:` call, so there's no change for the worse here.

Just two comments:
1. The selector `a[href$="#_back"]` matches links like `frontpage.html#_back` and possibly even external links. I'd go with the specific `a[href="#_back"]` since the insert tag creates `href="#_back` exactly.
2. Maybe the href `#_back` could be changed to something more specific to reduce possible collisions with user code (e.g. some link to a layout section with an ID of `_back`). I'd suggest `#_goback`.

--

In my opinion, a back link as in _a link with the label `go back`_, isn't the best concept in most cases anyway, since more often than not you actually want to say _go to overview_ or _show all entries_.

A future removal and a deprecation sound good to me.

aschempp:
does this really add an event listener for every click on the page, not tied to the back elements?

My alternative solution proposed in Slack would be something like this: `javascript:history.back(-1);return;`. The `return` should prevent any further scripts from being executed.

ausi:
> does this really add an event listener for every click on the page, not tied to the back elements?

What exactly do you mean by that? Only one event listener gets added by the code.

> My alternative solution proposed in Slack would be something like this: `javascript:history.back(-1);return;`.

But this would still be converted to `javascript%3Ahistory.back(-1);return;` by the `urlattr` flag,

SeverinGloeckle:
> does this really add an event listener for every click on the page, not tied to the back elements?

There is only one listener, but it's listening to all clicks on the page.

I think this is a good thing since it will catch clicks on back links in dynamic content (e.g. after loading content via AJAX) and in re-mounted markup (e.g. vue applications).

However, maybe the global script should be disableable since it _could_ interfer with userland JavaScript.

You could add some configurable flag (which is not set by default) to prevent adding the global script to `$GLOBALS['TL_HEAD']` and leave it to the developer to add a handler for `href="_goback"`.

In most cases, the global handler should still be a valid drop-in replacement for the inline `javascript:history.back(-1);`.

SeverinGloeckle:
> I think this is a good thing since it will catch clicks on back links in dynamic content (e.g. after loading content via AJAX) and in re-mounted markup (e.g. vue applications).

Thinking about this, if a back link is loaded dynamically while the current page does not contain a back link, there will be no script to handle the click on the new back link and it will not work – while the `javascript:` solution would still work.

To fix this, the back link handler would have to be loaded on every page (e.g. via a JavaScript template in the page layout).

Maybe @m-vo is right ([comment](https://github.com/contao/contao/issues/3364#issuecomment-904762293)):

> Keep in mind, that if it's for security reasons, BC breaks might be a necessity.

And the fix would be:
- replace the inline JavaScript with something like `href="_goback"`
- prepare a javascript template js_backlink with the global handler to restore the behaviour
- let users decide if they need the global handler
- inform users about the requirements to make `{{link_url::back}}` work (manual, maybe even a news on the Contao website)

This way Contao provides an opt-in solution to keep back links working.

fritzmg:
> But this would still be converted to `javascript%3Ahistory.back(-1);return;` by the `urlattr` flag,

We could hardcode an exception for this specific string. Not very pretty of course.

aschempp:
Maybe a stupid alternative idea, can we not just use the referrer? If there's none, then there's no back link 🤷

fritzmg:
Well we already have an insert tag for the referrer. And a link to the referrer is functionally different to `javascript:history.back(-1);` of course.

Toflar:
The referer is a very bad idea. First of all, it might not even be there for security reasons. Second, it might be prone to injection -> security reasons. And more: https://developer.mozilla.org/en-US/docs/Web/Security/Referer_header:_privacy_and_security_concerns

We also removed all our session based referer stuff from the FE for caching reasons.

JS is the only correct way to go and imho this is solved beautifully here. I would've done exactly the same.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -395,6 +395,15 @@ protected function doReplace($strBuffer, $blnCache)
 							$strTitle = 'Go back';
 						}
 
+						$strBackScript = "<script>document.documentElement.addEventListener('click',function(e){"
+							. "e&&e.target&&e.target.matches('a[href$=\"#_back\"],a[href$=\"#_back\"] *')&&(history.go(-1),e.preventDefault&&e.preventDefault())"
+						. "},true)</script>";
+
+						if (!\in_array($strBackScript, $GLOBALS['TL_HEAD'] ?? array(), true))
```
Toflar: Can't we just use a key here? Like `$GLOBALS['TL_HEAD']['contao_back_insert_tag'] = $strBackScript;`?

Toflar (approved):

fritzmg (approved):

m-vo (approved):

leofeyer:
As discussed in Mumble on September 30th, we want to allow list the exact string `javascript:history.go(-1)` to restore backwards compatibility and deprecate the `{{link::back}}` insert tag for removal in Contao 5.0.

ausi:
We also want to remove all usages of `javascript:history.go(-1)` and other back links from templates (e.g. reader modules).

aschempp (changes requested):
just requesting changes to not get this merged 😆

ausi:
> we want to allow list the exact string `javascript:history.go(-1)` to restore backwards compatibility and deprecate the `{{link::back}}` insert tag for removal in Contao 5.0.

See #3585 

> We also want to remove all usages of `javascript:history.go(-1)` and other back links from templates (e.g. reader modules).

See #3586

PR #3426, opened 2021-09-06
Migrate the newsletter module keys in version 4.0

aschempp:
According to https://github.com/contao/contao/blob/4.x/UPGRADE.md#front-end-module-keys these modules have been renamed in Contao 4. I noticed my RSS feed modules were automatically migrated, but the newsletter modules were not.

Since the newsletter-bundle does not have any migrations, and migrating this in core does not harm if they just don't exist, I would propose to add these migrations here 🙃

### DIFF
--- a/newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
+++ b/newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
@@ -39,6 +39,12 @@ class NewsletterModuleMigration extends AbstractMigration
             return false;
         }
 
+        $columns = $schemaManager->listTableColumns('tl_module');
+
+        if (!isset($columns['type']) {
+            return false;
+        }
+
         return $this->connection->fetchOne("SELECT COUNT(*) FROM tl_module WHERE type='nl_list' OR type='nl_reader'") > 0;
     }
 
--- a/newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
+++ b/newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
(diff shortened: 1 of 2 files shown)


fritzmg:
Since Contao 4.10 we are already using Migrations directly in the newsletter-bundle: https://github.com/contao/contao/tree/4.10/newsletter-bundle/src/Migration

May be this migration should be added there as well instead of in the core-bundle?

aschempp:
this is for Contao 4.9 though …

leofeyer:
Just put the migration into the corresponding bundle. 😄 It will fall into place once merged upstream.

fritzmg:
> this is for Contao 4.9 though …

Yes. And Migrations are already available in Contao 4.9. So new bundle-specific migrations should also be put in the respective bundle.

aschempp:
Here you go

fritzmg (dismissed):

m-vo (dismissed):

newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+
+        if (!$schemaManager->tablesExist(['tl_module'])) {
+            return false;
+        }
+
```
ausi, line 41:
```suggestion

        $columns = $schemaManager->listTableColumns('tl_module');

        if (!isset($columns['type']) {
            return false;
        }

```

newsletter-bundle/src/Migration/Version400/NewsletterModuleMigration.php
```diff
@@ -0,0 +1,73 @@
...
+        $columns = $schemaManager->listTableColumns('tl_module');
+
+        if (!isset($columns['type']) {
```
leofeyer:
```suggestion
        if (!isset($columns['type'])) {
```

PR #3434, opened 2021-09-07
Update the dependencies and remove the BC layers

leofeyer:
* [x] Make PHP 7.4 the minimum PHP version
* [x] Remove old minor versions that were still allowed due to BC
* [x] Remove the BC layers
* [x] Upgrade to Symfony 5.3 only (will be changed later to Symfony 5.4 only)
* [x] Use the current major versions where possible
* [ ] Use `doctrine/dbal` in version 3


m-vo:
There are also still some conditional statements checking for `PHP_VERSION_ID` in the codebase that we can drop.

Should we also refactor code? With PHP7.4 we could finally use…
* typed properties
* arrow functions (`fn(…) => …`)  without the need for `use` 
* shorter null coalescing assignment (`$a ??= $b`)
* array unpacking (`['a', 'b', ...$items]`) instead of `array_merge`

leofeyer:
Easy, tiger, this PR is still a draft. 😄

m-vo:
It's the excitement :smile:

leofeyer:
I feel you. 😄

leofeyer:
Since the PR is already pretty huge, let's merge it as is and add the other changes (doctrine/dbal v3, PHP 7.4 features etc.) later in separate PRs.

core-bundle/src/Controller/BackendController.php
```diff
@@ -182,12 +181,7 @@ public function pickerAction(Request $request): RedirectResponse
...
         if ($request->query->has('extras')) {
-            if ($request->query instanceof InputBag) {
-                $extras = $request->query->all('extras');
-            } else { /** @phpstan-ignore-line */
-                // Backwards compatibility with symfony/http-foundation <5.0
-                $extras = $request->query->get('extras');
-            }
+            $extras = $request->query->all('extras');
 
             if (empty($extras) || !\is_array($extras)) {
```
m-vo: `InputBag#all()` always returns an array, so we can IMHO drop the `is_array` check.

bytehead (dismissed):
LGTM.

m-vo (changes requested):

core-bundle/src/Controller/BackendController.php
leofeyer: Removed in ad0ef7b3b5a99aea231805324931b5da23f59315.

m-vo (approved):

bytehead (approved):

PR #3436, opened 2021-09-08
Drop the contao/polyfill-symfony package

leofeyer:
Now that we are on Symfony 5 only, we no longer need `contao/polyfill-symfony`.

### DIFF
--- a/core-bundle/src/Image/Studio/Figure.php
+++ b/core-bundle/src/Image/Studio/Figure.php
@@ -60,7 +60,6 @@ class Figure
      * All arguments but the main image result can also be set via a Closure
      * that only returns the value on demand.
      *
-     * @param ImageResult                                                                 $image          Main image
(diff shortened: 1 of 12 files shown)


aschempp:
Do we need it again with Symfony 6 maybe? Should we keep it or re-add it then?

leofeyer:
Add it again if necessary.

aschempp (approved):

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -60,7 +60,6 @@ class Figure
      * All arguments but the main image result can also be set via a Closure
      * that only returns the value on demand.
      *
-     * @param ImageResult                                                                 $image          Main image
```
bytehead, line 63: How is this related?

leofeyer: This change was made by the CS fixer.

PR #3437, opened 2021-09-08
Fix issues found by the code inspector

leofeyer:


core-bundle/src/DataContainer/PaletteManipulator.php
```diff
@@ -45,17 +45,16 @@ public static function create(): self
      * If the legend already exists, nothing will be changed.
      *
      * @param string|array $parent
-     * @param bool         $hide
      */
-    public function addLegend(string $name, $parent, string $position = self::POSITION_AFTER, $hide = false): self
+    public function addLegend(string $name, $parent, string $position = self::POSITION_AFTER, bool $hide = false): self
```
fritzmg: Wouldn't the same arguments apply here as in https://github.com/contao/contao/issues/3001 ? If this is possible after all, then we should have just changed the method signature of `FilesModel::findByPath`.

Otherwise we cannot allow this change, since it changes the method signature of a public method?

leofeyer: I seriously doubt that anyone has ever passed a non-boolean value here (in contrast to #3001 where `$path` could be `null`), but I have reverted the change in 7f1606206d692382796dc9f30ee23ece18acf9fb, so that we do not have to waste time with unnecessary discussions.

aschempp:
I get a lot of compile errors after this merge because our child classes have stricter arguments than their parent:

```
Compile Error: Declaration of Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader::getSourceContext(string $name): Twig\Source must be compatible with Twig\Loader\FilesystemLoader::getSourceContext($name)

Warning: Declaration of Contao\CoreBundle\Twig\Loader\FailTolerantFilesystemLoader::addPath(string $path, string $namespace = self::MAIN_NAMESPACE): void should be compatible with Twig\Loader\FilesystemLoader::addPath($path, $namespace = self::MAIN_NAMESPACE)
```

PR #764, opened 2019-09-16
Fix annotation definition

fritzmg:
As shown to @aschempp at the developer meeting, the current type definitions within the annotations can cause the following error:
```
  [RuntimeException]
  An error occurred while executing the "contao:install-web-dir" command:
  In AnnotationException.php line 83:

    [Type Error] Attribute "priority" of @Hook declared on method App\EventListener\ParseTemplateListener::onParseTemplate() expects a(n) int|null, but got integer.
```
The error only occurs if you actually use the attribute in your annotation, like `priority` in this case:
```
/**
 * @Hook("parseTemplate", priority=-10)
 */
```

PR #3439, opened 2021-09-08
Use the PHP 7.4 syntax

leofeyer:
Now that we are on PHP 7.4+, let's use the syntax.

### DIFF
--- a/core-bundle/src/Routing/Page/PageRoute.php
+++ b/core-bundle/src/Routing/Page/PageRoute.php
@@ -20,23 +20,12 @@ use Symfony\Component\Routing\Route;
 
 class PageRoute extends Route implements RouteObjectInterface
 {
-    /**
-     * @var PageModel
-     */
-    private $pageModel;
+    private PageModel $pageModel;
+    private ?string $urlPrefix;
+    private ?string $urlSuffix;
 
     /**
-     * @var string
-     */
-    private $urlPrefix;
-
-    /**
-     * @var string
-     */
-    private $urlSuffix;
-
-    /**
-     * The referenced content object.
+     * The referenced content object (can be anything).
      */
     private $content;
 
--- a/core-bundle/src/ServiceAnnotation/Callback.php
+++ b/core-bundle/src/ServiceAnnotation/Callback.php
@@ -30,20 +30,9 @@ use Terminal42\ServiceAnnotationBundle\Annotation\ServiceTagInterface;
  */
 final class Callback implements ServiceTagInterface
 {
-    /**
-     * @var string
-     */
-    public $table;
-
-    /**
-     * @var string
-     */
-    public $target;
-
-    /**
-     * @var int
-     */
-    public $priority;
+    public string $table;
+    public string $target;
+    public ?int $priority = null;
 
     public function getName(): string
     {
@@ -57,7 +46,7 @@ final class Callback implements ServiceTagInterface
             'target' => $this->target,
         ];
 
-        if ($this->priority) {
+        if (null !== $this->priority) {
             $attributes['priority'] = $this->priority;
         }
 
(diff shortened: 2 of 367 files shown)


core-bundle/src/Routing/Page/PageRoute.php
```diff
@@ -20,25 +20,14 @@
...
-    private $urlSuffix;
+    private PageModel $pageModel;
+    private ?string $urlPrefix;
+    private ?string $urlSuffix;
 
     /**
      * The referenced content object.
      */
-    private $content;
+    private ?object $content = null;
```
leofeyer: @aschempp Is this correct? Is `$content` either an object or `null`?

core-bundle/src/ServiceAnnotation/Callback.php
```diff
@@ -30,20 +30,9 @@
...
-     */
-    public $target;
-
-    /**
-     * @var int
-     */
-    public $priority;
+    public string $table;
+    public string $target;
+    public int $priority = 0;
```
leofeyer: Is this correct or does the default value have to be `null`?

core-bundle/src/ServiceAnnotation/Hook.php
```diff
@@ -29,15 +29,8 @@
...
-     * @var string
-     */
-    public $value;
-
-    /**
-     * @var int
-     */
-    public $priority;
+    public string $value;
+    public int $priority = 0;
```
leofeyer: Is this correct or does the default value have to be `null`?

core-bundle/src/ServiceAnnotation/PickerProvider.php
```diff
@@ -28,10 +28,7 @@
  */
 final class PickerProvider implements ServiceTagInterface
 {
-    /**
-     * @var int
-     */
-    public $priority;
+    public int $priority = 0;
```
leofeyer: Is this correct or does the default value have to be `null`?

core-bundle/src/ServiceAnnotation/Hook.php
fritzmg: The doc used to be `int|null`, but that lead to an `AnnotationException`: https://github.com/contao/contao/pull/764
But it is supposed to be `null` if not defined otherwise. So I think `public int $priority;` is correct.

But I think you will need to test if hook registrations with

```php
/**
 * @Hook(…)
 */
```

and

```php
/**
 * @Hook(…, priority=-10)
 */
```

still work after the change.

leofeyer: > So I think `public int $priority;` is correct.

No, it is not. It will lead to an `Error: Typed property Contao\CoreBundle\ServiceAnnotation\Hook::$priority must not be accessed before initialization` error if there is no default value, therefore the question is does the default value have to be `null` or can it also be `0`? @contao/developers

Toflar: Well, if you say it must be an `int`, then it must be an `int` :D It cannot be `null`. That would only be valid if you specified `?int`.

fritzmg: Yes, that's what I meant to write actually 🙈. I meant it should be tested with `public ?int $priority;` since the phpdoc type definition `int|null` lead to an error with the annotation parser in the past.

leofeyer: > Well, if you say it must be an `int`, then it must be an `int` :D It cannot be `null`.

If we need the default to be `null`, we would have to change the type to `public ?int $priority = null`. But I think `null` is equivalent to `0` anyway, because any service without a priority will implicitly have `0`, won't it?

core-bundle/src/Routing/Page/PageRoute.php
leofeyer: As briefly discussed, `$content` can be anything and can therefore not be type hinted.

core-bundle/src/ServiceAnnotation/Hook.php
leofeyer: As discussed with @aschempp, it has to be `public ?int $priority = null` (see #3444).

PR #3442, opened 2021-09-08
Change all occurrences of master (request) to main

aschempp:
Related to https://github.com/contao/contao/pull/3436

PR #3443, opened 2021-09-08
Renamed method to check for main scope

aschempp:
We could also consider changing/adding this in Contao 4.9.x to allow future-proof extensions already 🙃
I would rebase the PR if you agree.


Toflar:
I would keep it for 4.x. The forward compatibility argument could be used always. This PR is no different from any other :)

fritzmg (approved):

aschempp:
I have incorporated these changes into https://github.com/contao/contao/pull/3442

PR #3445, opened 2021-09-09
Allow to pass the actual 40x page to the page type

aschempp:
The current implementation renders all `TL_PTY` page types through `FrontendIndex::renderPage`. `FrontendIndex::renderPage` handles `error_40x` pages types differently from all other page types and does not forward the page object, which has already been passed to it. It then unnecessarily looks up the root page and error page again, instead of generating the object that has already been passed to `FrontendIndex::renderPage`.

### DIFF
--- a/core-bundle/src/Resources/contao/pages/PageError404.php
+++ b/core-bundle/src/Resources/contao/pages/PageError404.php
@@ -24,13 +24,15 @@ class PageError404 extends Frontend
 {
 	/**
 	 * Generate an error 404 page
+	 *
+	 * @param PageModel|null $page
 	 */
-	public function generate()
+	public function generate($page=null)
 	{
 		/** @var PageModel $objPage */
 		global $objPage;
 
-		$obj404 = $this->prepare();
+		$obj404 = $this->prepare($page);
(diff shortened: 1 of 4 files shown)


core-bundle/src/Resources/contao/pages/PageError404.php
```diff
@@ -76,14 +80,23 @@ public function getResponse()
...
 	 * @return PageModel
 	 *
 	 * @internal Do not call this method in your code. It will be made private in Contao 5.0.
 	 */
-	protected function prepare()
+	protected function prepare($page=null)
 	{
-		// Find the matching root page
-		$objRootPage = $this->getRootPageFromUrl();
+		if ($page instanceof PageModel && $page->type === 'error_404')
```
leofeyer, line 91:
```suggestion
		$obj404 =  null;

		if ($page instanceof PageModel && $page->type === 'error_404')
```

Otherwise you will get a `variable $obj404 might not be defined` warning further down.

leofeyer (dismissed):
It seems that your changes break the PHP 8 tests. 🤔

leofeyer (approved):

PR #3449, opened 2021-09-10
Handle invalid array access in tl_page

m-vo:
Screenshot_20210910_101434

Here is another one. The session array may not contain the key `CURRENT`.


Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #3454, opened 2021-09-11
Fix the event endTime adjustment

### LINKED ISSUE #3376: Event endtime will not be saved
**Affected version(s)**

Contao 4.9 
Bildschirmfoto 2021-08-24 um 11 12 00

**Description**

I found a bug in the events module (backend). If you set the time for "Date and Time" for a new or existing entry to 01:00 (end time), this time will not be saved. Any other value (01:01, 02:00, etc.) works correctly. I've tested many combinations in different systems (times, start and end dates, etc.).

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3376
| Docs PR or issue | -

`endTime` can be zero for certain times and thus `?:` would not work here. The `endTime` is not set if it is `null`.


ausi (approved):

m-vo (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #3455, opened 2021-09-11
Handle flags in insertTagToSrc()

### LINKED ISSUE #3430: {{file::uuid}} insert tag gets converted to uuid|urlattr
**Affected version(s)**

Contao `4.9.20`, `4.9.x-dev`

**Description**

If you use a regular `text` content element and insert a link to a file via the file picker in the TinyMCE, e.g.

```html
<p>Lorem <a href="{{file::c7869fac-ede7-11eb-9279-d8bbc1145c99}}">ipsum</a> dolor.</p>
```

then the content will be converted to

```html
<p>Lorem <a href="c7869fac-ede7-11eb-9279-d8bbc1145c99|urlattr">ipsum</a> dolor.</p>
```

after saving, which of course breaks the file reference.

/cc @ausi

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3430

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
+++ b/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
@@ -805,7 +805,7 @@ class StringUtil
 	public static function insertTagToSrc($data)
 	{
 		$return = '';
-		$paths = preg_split('/((src|href)="([^"]*){{file::([^"}]+)}}")/i', $data, -1, PREG_SPLIT_DELIM_CAPTURE);
+		$paths = preg_split('/((src|href)="([^"]*){{file::([^"}|]+)[^"}]*}}")/i', $data, -1, PREG_SPLIT_DELIM_CAPTURE);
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

leofeyer:
Thank you @ausi.

Defcon0:
I think we've had this discussion earlier, but I can't find the ticket :-) Why are file links replaced by the verbose URL path anyway? When I change the file's location, the link is broken then (unlike page urls which are insert tags as well).

ausi:
> Why are file links replaced by the verbose URL path anyway?

Otherwise images would not be visible in the tinyMCE editor.

> When I change the file's location, the link is broken then

Why? In the database the URL gets stored as `{{file::…}}` so renames should not be a problem.

Defcon0:
Aah, I see. I thought, it would also get saved to the actual path. So no problem here :-) Thanks for clearifying!

PR #3456, opened 2021-09-12
Set memory limit for sub-processes when running contao-setup

m-vo:
Running `php -dmemory_limit=<limit> contao-setup` won't make the sub processes run with a memory limit of `<limit>`. In fact this is true for all arguments because the binary cannot know what the php process was called with. 

We should IMHO at least handle the case for the memory limit, though. This PR does this by reading the value via `ini_get` and adding it to the call if not empty.

/cc @rabauss

### DIFF
--- a/manager-bundle/tests/Command/ContaoSetupCommandTest.php
+++ b/manager-bundle/tests/Command/ContaoSetupCommandTest.php
@@ -46,9 +46,10 @@ class ContaoSetupCommandTest extends ContaoTestCase
 
         $phpPath = (new PhpExecutableFinder())->find();
 
-        array_unshift($phpFlags, '-dmemory_limit=1G');
-
         $this->assertStringContainsString('php', $phpPath);
+
+        ini_set('memory_limit', '1G');
+        array_unshift($phpFlags, '-dmemory_limit=1G');
 
         $commandFilePath = (new \ReflectionClass(ContaoSetupCommand::class))->getFileName();
         $consolePath = Path::join(Path::getDirectory($commandFilePath), '../../bin/contao-console');
@@ -65,7 +66,7 @@ class ContaoSetupCommandTest extends ContaoTestCase
 
         $createProcessHandler = $this->getCreateProcessHandler($processes, $commandArguments, $invocationCount);
 
-        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler, '1G');
+        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler);
 
         (new CommandTester($command))->execute([], $options);
 
--- a/manager-bundle/tests/Command/ContaoSetupCommandTest.php
+++ b/manager-bundle/tests/Command/ContaoSetupCommandTest.php
@@ -63,7 +63,7 @@ class ContaoSetupCommandTest extends ContaoTestCase
 
         $createProcessHandler = $this->getCreateProcessHandler($processes, $commandArguments, $invocationCount);
 
-        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler);
+        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler, '');
 
         (new CommandTester($command))->execute([], $options);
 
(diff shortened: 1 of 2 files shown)


fritzmg (dismissed):

aschempp:
Here's what the Contao Manager does in the latest version. We did try to now override if not necessary in the first versions, but realised setting all these "recommended" flags does not harm in any case, so we just always set them now without any issues. https://github.com/contao/contao-manager/blob/1.x/api/Process/ConsoleProcessFactory.php#L228-L232

m-vo:
> but realised setting all these "recommended" flags does not harm in any case

This was my first idea as well but I figured it's a good thing you can explicitly set the memory limit. If you're server for instance only allows for 128MB you might want to run your CI/locally with this limit as well so that you won't get any surprises when deploying. And debugging is easier.

Just my assumptions, though. :smile:

manager-bundle/tests/Command/ContaoSetupCommandTest.php
```diff
@@ -63,7 +63,7 @@ public function testExecutesCommands(array $options, array $flags, array $phpFla
 
         $createProcessHandler = $this->getCreateProcessHandler($processes, $commandArguments, $invocationCount);
 
-        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler);
+        $command = new ContaoSetupCommand('project/dir', 'project/dir/public', $createProcessHandler, '');
```
leofeyer: Why `''` instead of `null`?

m-vo: This makes the test ignore the setting, while `null` would make it read from the current configuration (fallback, see constructor).

leofeyer: Can't you set a real value such as `-1` or `1G` instead?

m-vo: This test basically tests the `empty` statement (see below) while `testSetsMemoryLimitForSubProcesses()` sets + tests a value. We could use 1G there instead of `<memory-limit>` if you prefer?

```php
if (!empty($this->memoryLimit)) {
    $php[] = "-dmemory_limit=$this->memoryLimit";
}
```

leofeyer: My point is that `''` is not a valid (or at least reasonable) value. It is just a workaround for the unit test, isn't it?

m-vo: I think you can set anything: `php -dmemory_limit='' -r "echo ini_get('memory_limit') . PHP_EOL;"` outputs an empty string but I don't know what this would do effectively.

I think the unit test is correct as long as the `empty()` check is in there. But I am tempted to remove it altogether. Then we can drop the other test and just set it to`1G`.  Wdyt?

leofeyer: Like it. 👍 

Are you sure that `1G` is enough? The Contao Manager will try to increase the memory limit to 1.5G:

https://github.com/contao/contao-manager/blob/981e239354ef2e230b0e1e045add7acbb59a8ef4/api/console#L52-L55

m-vo: The `1G` setting is only for the test that only creates mocks, so this is fine. 
Hopefully, we're talking about the same thing. :see_no_evil: 

Here are the changes, now quite minimal: a623be39930c351cad0ecfd92e34fea45463f64d

leofeyer (changes requested):
The concept does not yet make sense to me.

A custom memory limit will only be used if passed as constructor argument aka if defined in the container, so most people will get `ini_get('memory_limit')`. If the memory limit is too small, the sub-process will still fail.

IMHO, we should always add `-dmemory_limit=` and set it to `max('1G', ini_get('memory_limit))` aka "at least 1G".

m-vo:
> A custom memory limit will only be used if passed as constructor argument aka if defined in the container, so most people will get ini_get('memory_limit').

That's the idea. The constructor allows mocking the value.

The sub processes will get the server's default value if not explicitly defined. This PR makes sure you'll get a different value if you deliberately defined one in the main process. This value is read via `ini_get()` and then turned into a `-d` flag for the sub process. And as it cannot change during runtime it's IMO fine to set it in the constructor.

m-vo:
> IMHO, we should always add -dmemory_limit= and set it to max('1G', ini_get('memory_limit)) aka "at least 1G".

That's IMHO a different topic, that we could also have a look into. We would probably need to use something like `filter_var` to get the exact value and handle cases like `-1`.

leofeyer:
> > A custom memory limit will only be used if passed as constructor argument aka if defined in the container, so most people will get ini_get('memory_limit').
> 
> That's the idea. The constructor allows mocking the value.

I see. Then why don't we remove the argument and "mock" the value using `ini_set()`?


>> IMHO, we should always add -dmemory_limit= and set it to max('1G', ini_get('memory_limit)) aka "at least 1G".
>
> That's IMHO a different topic, 

Agreed.

m-vo:
> I see. Then why don't we remove the argument and "mock" the value using ini_set()?

Makes sense! Changed in 1b9400a3ee6c035fa077dde780617830e1c38045.

leofeyer:
I have suggested changes in https://github.com/m-vo/contao/pull/6.

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

m-vo:
@leofeyer  Thanks to you as well. :relaxed:

PR #3457, opened 2021-09-12
Fix the picker for child tables without PID

### LINKED ISSUE #2046: Content element selector doesn't let you select elements, only pages and articles
**Affected version(s)**

Contao 4.9.2

**Description**

The content element selector in backend doesn't let you pick an element. Only lets you pick pages or articles.

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2046

A picker that picks elements from a child table like the one used in the include content element currently creates links like `?do=article&table=tl_content` which show an empty page because the pid (`id=`) is missing.

If there is a `CURRENT_ID` in the session, an arbitrary page might show up instead of an empty page, or it causes a permission error as in #2046.

This pull request fixes this by only adding the `table=` parameter if there is a valid pid.

### DIFF
--- a/core-bundle/src/Picker/AbstractTablePickerProvider.php
+++ b/core-bundle/src/Picker/AbstractTablePickerProvider.php
@@ -241,6 +241,8 @@ abstract class AbstractTablePickerProvider implements PickerProviderInterface, D
             return [null, null];
         }
 
+        $data = false;
+
         if ($id) {
             $qb = $this->connection->createQueryBuilder();
             $qb->select('pid')->from($table)->where($qb->expr()->eq('id', $id));
@@ -250,8 +252,6 @@ abstract class AbstractTablePickerProvider implements PickerProviderInterface, D
             }
 
             $data = $qb->execute()->fetch();
-        } else {
-            $data = false;
         }
 
         if ($dynamicPtable) {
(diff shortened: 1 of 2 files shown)


fritzmg (dismissed):
Tested and confirmed 👍

core-bundle/src/Picker/AbstractTablePickerProvider.php
```diff
@@ -226,36 +231,29 @@ protected function getPtableAndPid(string $table, string $value): array
...
 
-        if (false === $data) {
-            return [null, null];
+            $data = $qb->execute()->fetch();
         }
 
-        $pid = (int) $data['pid'];
-
         if ($dynamicPtable) {
             $ptable = $data['ptable'] ?: $ptable;
```
leofeyer: IIRC, `$data` can be `false` at this point. Should we replace `?:` with `??`?

fritzmg: In Contao 4.9, `ptable` can be an empty string: https://github.com/contao/contao/blob/2bc452ebc042a476285243a3c2371c1fa0203a9b/core-bundle/src/Resources/contao/dca/tl_content.php#L18

ausi:
```suggestion
            if (!empty($data['ptable'])) {
                $ptable = $data['ptable'];
            }
```

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3458, opened 2021-09-12
Add deprecations

ausi:
I added some deprecations messages for things that I think are deprecated.

Open Questions and ToDos: 

- [x] `System::getTimeZones()` should be replaced like we did with countries and languages 
      => #3661
- [x] `BackendPage` and `PageSelector` can be deprecated?
- [x] `BackendFile` and `FileSelector` can be deprecated?
- [x] `UnusedArgumentsException` is the last exception from our *exceptions.php* file. Should we move it to the `Contao` namespace or replace and deprecate it?
- [x] The interfaces `listable`, `editable`, `executable`, `uploadable` should probably be moved to the `Contao` namespace and be renamed?
- [x] ~~Can we deprecate `ModuleRandomImage`? It is bad for caching IMO as the randomness should be done on the client side I think.~~ Keep it.
- [x] ~~Modules quicknav and quicklink can be deprecated?~~ No
- [x] `TextStore` widget can be deprecated?
- [x] The globals `$GLOBALS['TL_CROP']`, `$GLOBALS['TL_CRON']`
- [x] `$GLOBALS['TL_AUTO_ITEM']`
      => #3662
- [x] `$GLOBALS['TL_CSS_UNITS']`
      => #3663
- [x] ~~Deprecate the CDN options for jQuery and Mootools.~~ Was already done in cc815eac2ac0b3d247e0984c7a89ead176ea80f4
- [x] Todo: Fix the tests ☺️


fritzmg:
> * Can we deprecate `ModuleRandomImage`? It is bad for caching IMO as the randomness should be done on the client side I think.

Modules that generate random output are not inherently "bad" for caching. You just need to be aware of the implications when enabling the cache (documentation issue). Depending on your use-case it might be ok if you see a random image only every 5 minutes and otherwise it will stay the same image (for everyone) - then you can simply set the cache time to 5 minutes for the affected page or pages.

And if this module was implemented as a proper fragment, you could even set optional separate cache times within the module's settings itself, so that you can configure that independently from the page caches times of the pages that this module will end up being displayed on.

And you could even implement it to be rendered as an ESI fragment (just generally speaking, whether it makes sense or not is a different story and also dependent on the use-case).

Toflar:
I would, however, still deprecate it. To be honest, I haven't had that use case in quite a while because "random images" are usually also combined with slides, swiping, mobile optimization etc. so you need JS anyway. You can sort the images randomly in JS then. 



>  `UnusedArgumentsException` is the last exception from our _exceptions.php_ file. Should we move it to the `Contao` namespace or replace and deprecate it?

I think this can be deprecated as soon as @aschempp finished his work on pages with aliases and placeholders. Or well, it just needs to be deprecated together with the auto_item logic in favor of proper routing is what I'm trying to say.

fritzmg:
> I would, however, still deprecate it. To be honest, I haven't had that use case in quite a while because "random images" are usually also combined with slides, swiping, mobile optimization etc. so you need JS anyway. You can sort the images randomly in JS then.

We only used it once in a project so far, but we used it for a random header image. No JavaScript necessary for that. A JavaScript solution would require you to output all possible image URLs, which is not really ideal if the random image is supposed to be chosen out of hundreds or thousands of images.

m-vo:
> We only used it once in a project so far, but we used it for a random header image.

Had that once, too, but used a custom route + a JS fetch request (with a random seed) that then loaded the image. Caching not affected, only a few lines of code. I think I'd implement the module this way, today.

fritzmg:
Just depends on your needs. If the image only needs to be random once a day for example, you can still configure the cache accordingly (if you want to cache at all in the first place) and then no JavaScript overhead is necessary.

core-bundle/src/Resources/contao/library/Contao/Config.php
```diff
@@ -105,6 +105,27 @@ class Config
 		'jpgQuality'      => 'contao.image.imagine_options[jpeg_quality]',
 	);
 
+	private static $arrToBeRemoved = array
+	(
+		'os'                    => true,
+		'browser'               => true,
+		'dbCharset'             => true,
+		'dbCollation'           => true,
+		'disableRefererCheck'   => true,
```
fritzmg, line 115: We should provide a bundle config for this first, before deprecating it.

ausi: I don’t know. It can be disabled by either setting the route attribute `_token_check=false` or by removing the listener. `disableRefererCheck` is already deprecated since Contao 4.7.3 AFAIK.

leofeyer:
> * [ ]  `System::getTimeZones()` should be replaced like we did with countries and languages

Yes

> * [ ]  `BackendPage` and `PageSelector` can be deprecated?
> * [ ]  `BackendFile` and `FileSelector` can be deprecated?

Yes, I think so, too. We have the picker now.

> * [ ]  `UnusedArgumentsException` is the last exception from our _exceptions.php_ file. Should we move it to the `Contao` namespace or replace and deprecate it?

Move it and deprecate it for removal in Contao 5 (see `UnresolvableDependenciesException`).

> * [ ]  The interfaces `listable`, `editable`, `executable`, `uploadable` should probably be moved to the `Contao` namespace and be renamed?

We cannot move them without breaking BC, but we should at least introduce successor interfaces and flag them for removal in Contao 5.

> * [x]  ~Can we deprecate `ModuleRandomImage`? It is bad for caching IMO as the randomness should be done on the client side I think.~ Keep it.

I am in favor of deprecating it, because then we could also get rid of some fields in `tl_module` that we had to add specifically for this one module (e.g. `imgSize`, `fullsize`, `useCaption`).

> * [ ]  Modules quicknav and quicklink can be deprecated?

No, we have to keep them. My initial idea to provide quicknav/quicklink templates for the regular navigation module turned out to be more complicated than anticipated.

> * [ ]  `TextStore` widget can be deprecated?

Yes, I guess so. We should be able to use the password widget now that we have removed the confirmation field.

> * [ ]  The globals `$GLOBALS['TL_CROP']`, `$GLOBALS['TL_CRON']`, `$GLOBALS['TL_AUTO_ITEM']` and `$GLOBALS['TL_CSS_UNITS']` ce be deprecated?

Definitely.

ausi:
> > * [ ]   The interfaces `listable`, `editable`, `executable`, `uploadable` should probably be moved to the `Contao` namespace and be renamed?
> 
> We cannot move them without breaking BC, but we should at least introduce successor interfaces and flag them for removal in Contao 5.

Note to myself:
- `executable` => `Contao\MaintenanceModuleInterface`
- `listable` => `Contao\ListableDataContainerInterface`
- `editable` => `Contao\EditableDataContainerInterface`
- `uploadable` => `Contao\UploadableWidgetInterface`

fritzmg:
> Note to myself:
> 
>   * `executable` => `Contao\MaintenanceModuleInterface`
>   * `listable` => `Contao\ListableDataContainerInterface`
>   * `editable` => `Contao\EditableDataContainerInterface`
>   * `uploadable` => `Contao\UploadableWidgetInterface`

In https://github.com/contao/contao/pull/3600 a new `Contao\FinalizableWidget` interface would be introduced. Should I add the `Interface` suffix there too?

ausi:
> Should I add the `Interface` suffix there too?

IMO yes.

ausi:
This should be ready to review. I’ve created new issues for the open topics.

@leofeyer should I create a ticket for deprecating `ModuleRandomImage` too?

core-bundle/src/Image/ImageSizes.php
```diff
@@ -108,8 +108,18 @@ private function loadOptions(): void
...
         $this->framework->initialize();
 
+        // Backwards compatibility
         $this->options = $GLOBALS['TL_CROP'] ?? [];
 
+        if (
+            3 !== \count($this->options)
+            || 0 !== \count($this->options['image_sizes'] ?? [])
+            || 2 !== \count($this->options['relative'] ?? [])
+            || 10 !== \count($this->options['exact'] ?? [])
```
m-vo, line 118: Why do we need these checks? Shouldn't `$this->options` always be an empty array if the deprecated `TL_CROP` is not used?

ausi: `$GLOBALS['TL_CROP']` is used for reading and modifying the image size options. These checks ensure that the default value was not changed. We cannot remove the default value for `$GLOBALS['TL_CROP']` as some users will rely on it. The default is set in: https://github.com/contao/contao/blob/f2e179360dd7e1c5a10d657ca92f0746e18a0508/core-bundle/src/Resources/contao/config/config.php#L474-L491

m-vo: I see. 

So if I got you correctly, this is basically saying…
```php
$default = [
    'image_sizes' => [],
    'relative' => ['proportional', 'box'],
    'exact' => [
        'crop', 
        'left_top', 'center_top', 'right_top',
        'left_center', 'center_center', 'right_center',
        'left_bottom', 'center_bottom', 'right_bottom',
    ],
 ];

if($GLOBALS['TL_CROP'] !== $default) {
  // trigger deprecation
}`
```
…, right?

ausi: Yes but I did not find a great way to compare them while ignoring numeric keys and the order.

m-vo (approved):

core-bundle/src/Resources/contao/config/constants.php
```diff
@@ -8,16 +8,19 @@
...
 define('BUILD', '0-dev');
 define('LONG_TERM_SUPPORT', false);
 
+// Backwards compatibility
 // Link constants
 define('LINK_BLUR', ' onclick="this.blur()"');
 define('LINK_NEW_WINDOW', ' onclick="return !window.open(this.href)"');
 define('LINK_NEW_WINDOW_BLUR', ' onclick="this.blur();return !window.open(this.href)"');
 
+// Backwards compatibility
```
fritzmg, line 23: May be we should put these constants in a class somewhere? e.g. `ContaoContext` - since at least some of those are used when pushing entries into the system log for example. Then we can deprecate them and remove them from Contao 5.0.

core-bundle/src/Resources/contao/library/Contao/Config.php
fritzmg: Well, a bundle config is much easier than implementing a compiler class and manager plugin that removes the listener from the container. I think this setting is actually used a lot.

core-bundle/src/Resources/contao/library/Contao/EditableDataContainerInterface.php
```diff
@@ -0,0 +1,26 @@
...
+	public function cut();
+
+	public function copy();
+
+	public function move();
+
+	public function edit();
+}
+
+class_alias(EditableDataContainerInterface::class, 'editable');
```
fritzmg, line 26: Should we also add a comment here, that we can remove this in 5.0?

ausi: Probably not needed as we will remove all `class_alias()` calls from all Contao classes.

core-bundle/src/Resources/contao/config/constants.php
ausi: I would rather replace them with the string values directly.

fritzmg: But they have special meaning which could be documented in the code. e.g. if you use `TL_ERROR` (`'ERROR'`) a system log entry will be coloured red. `TL_CRON` (`'CRON'`) entries will be coloured green etc.

ausi: Ah, now I understand I think. The class `Contao\CoreBundle\Monolog\ContaoContext` already exists and has all these these constants already. I will change the usage everywhere in the core then I think. Is that what you meant? ☺️

fritzmg: Right, I wasn't even really aware that these constants are already present there 😁. Yes, that is what I meant 👍

fritzmg: On that note: since `System::log` is also deprecated - you would proabably need to change any `$this->log` or `System::log` occurrence to:

```php
System::getContainer()->get('logger')->log('…', ['contao' => new ContaoContext(__METHOD__, ContaoContext::GENERAL)]);
```

However, I personally always find this rather cumbersome and still often use just `System::log`, since it's more convenient. May be we could introduce a `SystemLogger` helper service to improve DX here? Follow up PR of course, since out of scope.

ausi: Done in c67d824ebb32d37d2df2ba7f622db59f494470af

> Follow up PR of course, since out of scope.

Created issue #3701

m-vo (approved):

core-bundle/src/Resources/contao/helper/exception.php
```diff
@@ -8,13 +8,15 @@
...
- * Class UnusedArgumentsException
- *
- * @author Leo Feyer <https://github.com/leofeyer>
- */
-class UnusedArgumentsException extends Exception
+// Register alias in the global namespace for backwards compatibility
+class_exists(\Contao\UnusedArgumentsException::class);
+
+// Let composer find the deprecated class for autoload backwards compatibility
+if (!class_exists('UnusedArgumentsException', false))
```
leofeyer, line 15: How can this condition ever be `true`? The `class_exists()` call above will register the global alias.

ausi: Never, but we need composer to register this file for the class map of `UnusedArgumentsException`

ausi: We could try if we can use `if (false)` instead?

fritzmg (approved):
As discussed on Slack we can deprecate `disableRefererCheck` and if the need arises in Contao 5.0, we can still introduce a new bundle configuration there. I can also provide a guide in the dev docs where you could disable the token check via your own event listener (where you can also implement your own conditions).

core-bundle/src/Resources/contao/helper/exception.php
leofeyer: That's not necessary IMHO.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3461, opened 2021-09-13
Do not reset the entire flash bag when resetting the message system

### LINKED ISSUE #3393: Session flash bag is cleared upon "save and close"
I don't understand, why contao is clearing the whole session flash bag in contao backend mode, when pressing the saveNclose button? Wouldn't it be better to clear 

_symfony_flashes.contao.BE.info,
_symfony_flashes.contao.BE.error,
_symfony_flashes.contao.BE.confirm

only?

https://github.com/contao/contao/blob/bdc6e30cf50421bf3ec98d67b0b21ddd5ffa950d/core-bundle/src/Resources/contao/drivers/DC_File.php#L336

As it is now, it is not possible to third party backend extensions to properly use the symfony session flash bag.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3393
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Message.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Message.php
@@ -178,7 +178,15 @@ class Message
 			return;
 		}
 
-		$session->getFlashBag()->clear();
+		$flashBag = $session->getFlashBag();
+
+		// Find all contao.BE. keys (see #3393)
+		$keys = preg_grep('(^contao\.BE\.)', $flashBag->keys());
+
+		foreach ($keys as $key)
+		{
+			$flashBag->get($key); // clears the message
+		}
 	}
 
 	/**
--- a/core-bundle/src/Resources/contao/library/Contao/Message.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Message.php
@@ -180,8 +180,8 @@ class Message
 
 		$flashBag = $session->getFlashBag();
 
-		// Find all contao.BE. keys (see #3393)
-		$keys = preg_grep('(^contao\.BE\.)', $flashBag->keys());
+		// Find all contao. keys (see #3393)
+		$keys = preg_grep('(^contao\.)', $flashBag->keys());
 
 		foreach ($keys as $key)
 		{


ausi (dismissed):

m-vo:
Why do we need to clear the flash bag at all?

m-vo (dismissed):

bytehead (dismissed):

core-bundle/src/Resources/contao/library/Contao/Message.php
```diff
@@ -178,7 +178,15 @@ public static function reset()
 			return;
 		}
 
-		$session->getFlashBag()->clear();
+		$flashBag = $session->getFlashBag();
+
+		// Find all contao.BE. keys (see #3393)
+		$keys = preg_grep('(^contao\.BE\.)', $flashBag->keys());
```
Toflar: `FE` is possible too, right? Probably not used by the core but if you use `Message::add()` from within the FE, it would now not get cleared anymore which is the most important part due to the session/caching.

Toflar (changes requested):

core-bundle/src/Resources/contao/library/Contao/Message.php
leofeyer: Agreed. Should we maybe remove all keys starting with `contao.`?

Toflar: I think that makes sense, yes. We use the prefix already in the `getFlashBagKey()` method. Make that prefix a class constant and use it in both methods, then that should be fine :)

fritzmg: May be only depending on current scope? An action that clears the flash bag for the back end should not also clear the flash bag for the front end (or vice versa), in case you are logged into the back end and also visit the front end with the same session.

leofeyer: The scope can be anything, because all `add*()` and `generate*()` methods allow to set a custom scope. Therefore, removing all keys starting with `contao.` is the only way to reliably reset the messages.

m-vo (approved):

bytehead (approved):

Toflar (approved):

PR #3464, opened 2021-09-13
Redirect if the back end is called via the front end preview entry point

### LINKED ISSUE #3099: preview.php allows to create xml files with wrong urls.
**Affected version(s)**

Contao 4.9.15 and Contao 4.11.4

**Description**

The frontend preview allows you to call the backend via preview.php. For changes that lead to the generation of the XML files, the URLs are generated as preview.php URLs.

This happens easily when using the extension Frontend-Helper from Rocksolid. See also [https://github.com/madeyourday/contao-rocksolid-frontend-helper/issues/45](url)

Without the extension the behavior can be adjusted by: domain.tld/preview.php/contao?do=maintenance -> Rewrite XML files.

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3099
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/EventListener/BackendPreviewRedirectListener.php
+++ b/core-bundle/src/EventListener/BackendPreviewRedirectListener.php
@@ -39,9 +39,10 @@ class BackendPreviewRedirectListener
 
         $request = $event->getRequest();
 
-        if (!$request->attributes->has('_preview')
+        if (
+            !$request->attributes->has('_preview')
             || true !== $request->attributes->get('_preview')
-            || !$this->scopeMatcher->isBackendRequest($request)
+            || $this->scopeMatcher->isFrontendRequest($request)
             || ($request->attributes->has('_allow_preview') && true === $request->attributes->get('_allow_preview'))
         ) {
             return;
--- a/core-bundle/src/EventListener/BackendPreviewRedirectListener.php
+++ b/core-bundle/src/EventListener/BackendPreviewRedirectListener.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Routing\ScopeMatcher;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+
+/**
+ * @internal
+ */
+class BackendPreviewRedirectListener
+{
+    /**
+     * @var ScopeMatcher
+     */
+    private $scopeMatcher;
+
+    public function __construct(ScopeMatcher $scopeMatcher)
+    {
+        $this->scopeMatcher = $scopeMatcher;
+    }
+
+    public function __invoke(RequestEvent $event): void
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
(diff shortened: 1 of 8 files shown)


ausi (dismissed):

bytehead (dismissed):

leofeyer:
What I dislike about the PR, is that the redirect only happens for the `contao_backend` route and not for the other `contao_*` routes such as `contao_backend_login` or `contao_backend_password`. @contao/developers WDYT?

Toflar:
Imho the redirect should happen based on the scope? If it's not `frontend` then redirect?

aschempp (changes requested):
I agree with @Toflar. I would only redirect backend scope though.

leofeyer:
New concept added in fa97955ef8ad69f96267aa3882648b25c96e73ce. I will add the tests if we agree to go this way.

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
```diff
@@ -0,0 +1,52 @@
...
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        if (!$request->attributes->has('_preview')
+            || true !== $request->attributes->get('_preview')
+            || !$this->scopeMatcher->isBackendRequest($request)
+            || ($request->attributes->has('_allow_preview') && true === $request->attributes->has('_allow_preview'))
```
ausi:
```suggestion
            || ($request->attributes->has('_allow_preview') && true === $request->attributes->get('_allow_preview'))
```

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        if (!$request->attributes->has('_preview')
+            || true !== $request->attributes->get('_preview')
+            || !$this->scopeMatcher->isBackendRequest($request)
```
ausi: Would it make sense to check for the frontend instead?
```suggestion
            || $this->scopeMatcher->isFrontendRequest($request)
```

ausi:
> New concept added in [fa97955](https://github.com/contao/contao/commit/fa97955ef8ad69f96267aa3882648b25c96e73ce).

I’m not 100% sure if I like the new `_allow_preview`, on the other hand hardcoding the two routes would make it inflexible for extension developers...

The rest LGTM! 🎉

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
leofeyer: You mean in case someone has added their own scope?

leofeyer:
> I’m not 100% sure if I like the new `_allow_preview`, on the other hand hardcoding the two routes would make it inflexible for extension developers...

Same.

Toflar:
I like the concept, I have to admit 🤔

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
ausi: Yes, or their own route without any scope.

ausi: Idea © by @Toflar https://github.com/contao/contao/pull/3464#issuecomment-918095113

leofeyer: Changed in cc58b21e612a18924c370a9cbc4ab83d68fb5278.

ausi (approved):

PR #3466, opened 2021-09-13
Implemented a contao:database:import command

Toflar:
In our onging quest to get rid of the installation bundle at some point, there's only the import DB command missing which this PR should implement.

PS: Not sure if this is really the last missing piece since @m-vo added the db strict mode stuff to the installation bundle again in the latest versions so maybe we also need this information as a command.

To do:

- [ ] Decide on whether that should be core-bundle or manager-bundle (and then we should also move `contao:migrate` maybe)
- [ ] Decide on the `tl_` filter for the truncate queries
- [ ] Tests

### DIFF
--- a/core-bundle/src/Command/DatabaseImportCommand.php
+++ b/core-bundle/src/Command/DatabaseImportCommand.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Command;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Imports.
+ *
+ * @internal
+ */
+class DatabaseImportCommand extends Command
+{
+    protected static $defaultName = 'contao:database:import';
+
+    private Connection $connection;
+
+    public function __construct(Connection $connection)
+    {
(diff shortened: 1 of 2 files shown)


core-bundle/src/Command/DatabaseImportCommand.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @throws Exception
+     */
+    private function truncate(): void
+    {
+        $tables = $this->connection->getSchemaManager()->listTableNames();
+
+        foreach ($tables as $table) {
+            if (0 === strncmp($table, 'tl_', 3)) {
```
Toflar, line 129: This is copied from the current installation bundle. Not sure we should keep this?

core-bundle/src/Command/DatabaseImportCommand.php
```diff
@@ -0,0 +1,134 @@
...
+                $this->connection->executeStatement('TRUNCATE TABLE '.$this->connection->quoteIdentifier($table));
```
m-vo, line 130: IMHO truncating won't work reliably if there are FKs in place. A typical workaround is to disable FK checks beforehand and re-enable them afterwards.

Toflar: The installation bundle doesn't do this either at the moment...

m-vo:
> PS: Not sure if this is really the last missing piece since @m-vo added the db strict mode stuff to the installation bundle again in the latest versions so maybe we also need this information as a command.

 This should definitely go into a command in the future. I added it there because there are currently already some other helpful checks that too should get migrated.

core-bundle/src/Command/DatabaseImportCommand.php
m-vo: The prefix doesn't make much sense to me. But I wonder if we should rather drop the whole truncation logic and let the SQL files handle that themselves? They can include the schema and arbitrary commands anyway.

Maybe we could pair the command with a `contao:database:dump` command that creates compatible files?

Toflar: Yes I will provide one but it's not related.

Toflar:
Review requested regarding the TODOs so we can decide on those.

Toflar:
Replaced by https://github.com/contao/contao/pull/3594.

PR #3467, opened 2021-09-13
Prevent image overflow for selected content elements

fritzmg:
When you reference a content element with an image in the `alias` content element, the image might overflow the container:

Screenshot 2021-09-13 172450

This  PR fixes that by applying the respective CSS also to the output of the selected element:

image


aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #3468, opened 2021-09-14
Make sure redirects are at least path-absolute

### LINKED ISSUE #3065: Externe Weiterleitung mit News-Inserttag funktioniert nicht (mehr?)
**Affected version(s)**

Contao 4.9

**Description**

Hallo,

wir haben gerade von einem Kunden die Info bekommen, dass die Weiterleitungen auf Nachrichten (bspw.) als externe Weiterleitung nicht mehr funktionieren. Dieses System ist 4.4, ich kann es aber auch lokal reproduzieren mit 4.9. In der Online-Demo ließ es sich auch reproduzieren.

Im Feld "url" steht `{{news_url::11}}`. Damit landet man bei `/de/de/news-alias` (doppeltes locale). Setzt man ein Slash davor (/`{{news_url::11}}`), geht es und man landet bei `/de/news-alias`. Laut Kunde war das bisher nicht nötig. Wurde diesbzgl. etwas verändert in letzter Zeit (könnte auch ca. 1 Jahr betreffen)?

Das wäre grundsätzlich nicht problematisch, aber in Contao 4.9 können Redakteure ja per DCA Picker News dort auswählen und da landet automatisch kein Slash davor, was zum Fehler führt.

Ciao Defcon0

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3065 
| Docs PR or issue | -

This PR fixes the relative path issue for redirect responses by implementing a response listener (instead of changing every instance where a `RedirectResponse` object is created) which checks whether this was a Contao frontend request and makes sure that the redirect response is at least path-absolute.

@contao/developers wdyt?

TODO: small test for the listener (to make sure it does not change the URL when it's not a Contao frontend request or already an absolute URL).

### DIFF
--- a/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
+++ b/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
@@ -41,7 +41,7 @@ class MakeRedirectResponseUrlAbsoluteListener
 
         $url = $response->getTargetUrl();
 
-        if (preg_match('@^(http|/)@i', $url)) {
+        if ('/' === $url[0] || null !== parse_url($url, PHP_URL_SCHEME)) {
             return;
         }
 
--- a/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
+++ b/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Routing\ScopeMatcher;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Terminal42\ServiceAnnotationBundle\Annotation\ServiceTag;
+
+/**
+ * This makes sure that any redirect response in the Contao frontend scope is at least path-absolute (see #3065).
+ * 
+ * @ServiceTag("kernel.event_listener")
+ */
+class MakeRedirectResponseUrlAbsoluteListener
+{
+    private $scopeMatcher;
+
+    public function __construct(ScopeMatcher $scopeMatcher)
+    {
+        $this->scopeMatcher = $scopeMatcher;
+    }
+
+    public function __invoke(ResponseEvent $event): void
+    {
+        $response = $event->getResponse();
+
+        if (!$this->scopeMatcher->isFrontendMasterRequest($event) || !$response instanceof RedirectResponse) {
+            return;
+        }
+
+        $url = $response->getTargetUrl();
+
+        if (preg_match('@^(http|/)@i', $url)) {
+            return;
+        }
+
+        $response->setTargetUrl('/'.$url);
+    }
+}
--- a/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
+++ b/core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
@@ -1,5 +1,15 @@
 <?php
 
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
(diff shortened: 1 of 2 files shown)


core-bundle/src/EventListener/MakeRedirectResponseUrlAbsoluteListener.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $response = $event->getResponse();
+
+        if (!$this->scopeMatcher->isFrontendMasterRequest($event) || !$response instanceof RedirectResponse) {
+            return;
+        }
+
+        $url = $response->getTargetUrl();
+
+        if (preg_match('@^(http|/)@i', $url)) {
```
ausi: This check could result in false negatives I think, like for `httpisgreat.html`.

```suggestion
        if ($url[0] === '/' || parse_url($url, PHP_URL_SCHEME) !== null) {
```

fritzmg: True, changed in 7f7a238ee

m-vo: Can the url be empty?

fritzmg: > Can the url be empty?

No, that would lead to

```
InvalidArgumentException:
Cannot redirect to an empty URL.
```

ausi: We should proably add `'' === $url ||` for that case. Or do we want to convert the empty string to `/`?

fritzmg: You cannot create a `RedirectResponse` object with an empty string (or null) and you cannot call `setTargetUrl` with an empty string. It would lead to aforementioned exception.

ausi:
+1 from me. But I would still try to fix all instances of `RedirectResponse` and probably trigger a deprecation warning in the listener so that we can remove it sometime in the future.

fritzmg:
> But I would still try to fix all instances of `RedirectResponse` and probably trigger a deprecation warning in the listener so that we can remove it sometime in the future.

To be honest I did not feel like checking all 55 instances where redirect responses are created or the 100ish `Controller::redirect` calls in the core 😁. And we would need some sort of helper function/service in order to not copy the same code so many times. But even that seems a bit tedious. To be fair though, in many instances the URL is already either an absolute URL returned by `PageModel::getAbsoluteUrl` or an absolute path generated by the router (though in many instances the leading slash will have been stripped again by Contao).

At a minimum we would need to process the URLs here:

* https://github.com/contao/contao/blob/1525618c3b8aea3c1aec97c3c1629f72475d93bd/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1194
* https://github.com/contao/contao/blob/5b719bf9ee5191d4b87389398386b0da38b512bd/news-bundle/src/Resources/contao/modules/ModuleNewsReader.php#L121
* https://github.com/contao/contao/blob/5b719bf9ee5191d4b87389398386b0da38b512bd/calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php#L126

And then each extension would need to do the same in their code.

aschempp:
I understand the motivation for this solution, but I really dislike that we're creating a listener to patch code that should be fixed in the original place. I would be fine with adjusting something like this in `Controller::redirect()`, but if a "real" redirect exception is thrown that place should make sure the URL is absolute. I get that this is complicated, but only because our system generates relative requests in the first place - which it should not.

maybe... just maybe this would be a temporary solution for Contao 4 and we change the requests to absolute (with or without host) in Contao 5?

leofeyer:
As discussed in the Contao call, we want to add a warning if that is not too much work. In Contao 5, we want to switch to path absolute URLs.

leofeyer:

fritzmg:
Closing in favor of #6217

PR #3470, opened 2021-09-14
Make sure .public is a file and not a directory

### LINKED ISSUE #3465: ce_image 404 in 4.12 / files directory not symlinked to public directory
**Affected version(s)**

Contao 4.12

**Description**

`files` directory is no longer symlinked running the `contao:symlinks` command, which result in broken images using default `ce_image` content elements

I manually added `$this->symlink('files', Path::join($this->webDir, 'files'));` inside `contao:symlinks` command to solve that issue,
am I doing anything wrong ?

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3465
| Docs PR or issue | -

If you happen to have a _directory_ called `.public` inside one of your directories in `files/`, Contao will think that it is already public, because it uses `file_exists` - which also returns `true` for directories. However, such a directory will then never actually be public (the symlink command filters for files only), nor can it be made public via the file manager

This PR changes the respective `file_exists` calls to `is_file`. It also adds new exceptions to the `Folder` class when trying to protect or unprotect a directory that contains a `.public` directory, so that the `.public` directory does not get accidentally deleted.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -22,6 +22,7 @@ use Imagine\Gd\Imagine;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpFoundation\Session\Session;
+use Webmozart\PathUtil\Path;
 
 /**
  * Provide methods to modify the file system.
@@ -2156,7 +2157,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 		}
 
 		// Protect or unprotect the folder
-		if (file_exists($this->strRootDir . '/' . $this->intId . '/.public'))
+		if (is_file($this->strRootDir . '/' . $this->intId . '/.public'))
 		{
 			$objFolder = new Folder($this->intId);
 			$objFolder->protect();
@@ -2692,7 +2693,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			$protected = $blnProtected;
 
 			// Check whether the folder is public
-			if ($protected === true && \in_array('.public', $content))
+			if ($protected === true && is_file(Path::join($folders[$f], '.public')))
 			{
 				$protected = false;
 			}
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2693,7 +2693,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			$protected = $blnProtected;
 
 			// Check whether the folder is public
-			if ($protected === true && is_file(Path::join($folders[$f], '.public')))
+			if ($protected === true && \in_array('.public', $content) && !is_dir(Path::join($folders[$f], '.public')))
 			{
 				$protected = false;
 			}
--- a/core-bundle/src/Resources/contao/dca/tl_files.php
+++ b/core-bundle/src/Resources/contao/dca/tl_files.php
@@ -823,7 +823,7 @@ class tl_files extends Contao\Backend
 		$blnUnprotected = $objFolder->isUnprotected();
 
 		// Disable the checkbox if a parent folder is public (see #712)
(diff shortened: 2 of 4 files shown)


core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2692,7 +2693,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 			$protected = $blnProtected;
 
 			// Check whether the folder is public
-			if ($protected === true && \in_array('.public', $content))
+			if ($protected === true && is_file(Path::join($folders[$f], '.public')))
```
fritzmg: Note that this will increase the amount of file system access operations though.

m-vo:
> It also adds new exceptions to the Folder class when trying to protect or unprotect a directory that contains a .public directory

I don't understand why we care about the name of the directory? IMHO it should be possible to name a directory `.public` and it should just behave the same as `foo`.

fritzmg:
> I don't understand why we care about the name of the directory? IMHO it should be possible to name a directory `.public` and it should just behave the same as `foo`.

You cannot have a file and a directory called `.public` in the same folder. So suppose you have a `foo/.public` directory and then you call `(new Folder('foo'))->protect()`, Contao would delete the `.public` folder.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
ausi: Changing it to use `!is_dir()` could help in this case because that call should already be cached from the code above.

m-vo:
> You cannot have a file and a directory called .public in the same folder.

:man_facepalming: Lol, ok I didn't think of the simplest thing.

All good then. :+1:

core-bundle/src/Resources/contao/drivers/DC_Folder.php
fritzmg: Good point! Changed in 3cc318964

m-vo: Alternatively, we could disallow naming a folder `.public` from the UI, make it a known limitation and just keep the exceptions in case someone perfomed the operation on the filesystem nevertheless.

m-vo: OR we say it's not a bug, it's a feature! (and silently adjust the symlink command to also scan for directories)

fritzmg: But the system still needs to know whether it is a file or not, otherwise the folder will still be displayed as being public, when it is not.

m-vo: Yes, this would only work if a `.public` folder was implicitly considered public.

asaage:
Maybe we can evaluate for a feature version of contao if it makes sense to ditch the ```.public```-file-mechanism and symlink everything without a dot in front of the name - that would probably also allow for protected folders inside public folders - which currently and unfortunately is not possible...

fritzmg:
> Maybe we can evaluate for a feature version of contao if it makes sense to ditch the `.public`-file-mechanism and symlink everything without a dot in front of the name

I do not think that is a good idea. Hidden files and folders are different from protected files and folders. Furthermore some FTP servers are set up to hide hidden files and folders - so if you had to hide a folder this way in order to protect the folder for Contao, anyone with only FTP access would not be able to access that folder anymore.


>  that would probably also allow for protected folders inside public folders - which currently and unfortunately is not possible...

That would still be the same.

asaage:
Agree -the FTP-thing would be anoing.
But still a little more fine-grained control in tl_files would be nice like:

- ability to have protected folders inside public folders
- ability to see trail of parentfolders even if the user has only access to a subset of filemounts (https://github.com/contao/contao/issues/3242#issuecomment-891042301)
- other properties like "read-only" for certain groups/users in BE for example

Can't the routing decide what to do with files based on dbfs-fields instead of relying on the correct symlinks?

fritzmg:
@asaage this is all beyond the scope of this PR. You need to open a separate feature request.

ausi (approved):

aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #3472, opened 2021-09-15
Make the backend path configurable

richardhj:
| Q                | A
| -----------------| ---
| Fixed issues     | 
| Docs PR or issue | contao/docs#873

This allows setting a custom path for the Contao backend, i.e., `https://demo.contao.org/gondao`.

```yaml
# config/config.yaml

contao:
  backend:
    route_prefix: '/gondao'
```

I don't know yet, how to write a test for this. I already added a config_.yml in the Functional folder, but I don't know when these files are loaded.


leofeyer:
Why?

richardhj:
> Why?

Different motivations:
- "Security by obscurity"
- To satisfy requirements of enterprise applications

m-vo:
> Why?

Tbh, I had this request numerous times in the past. And using `/admin` for instance is IMO a sensible choice, especially for users that do not care about the type of system.

manager-bundle/src/Resources/skeleton/config/security.yml
```diff
@@ -52,6 +52,7 @@ security:
                     - contao.security.logout_handler
                 success_handler: contao.security.logout_success_handler
 
+    # Here for BC considerations
     access_control:
```
fritzmg, line 56: So what does `access_control` actually do then? I thought for the back end firewall to work properly, the route _must_ start with `/contao`, due to the `access_control` configuration. Doesn't this need to be changed as well?

richardhj: Nope, the firewall works because of the route matcher -- the route matcher matches all routes with scope=backend attribute.
The access_control ist just in case someone forgets to set the attribute in the route.

fritzmg: Ah I see, thank you for the clarification :). Need to update that in the docs too ;)

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -513,6 +513,11 @@ private function addBackendNode(): NodeDefinition
             ->getRootNode()
             ->addDefaultsIfNotSet()
             ->children()
+                ->scalarNode('route_prefix')
+                    ->info('Define the path of the Contao backend.')
+                    ->example('/admin')
+                    ->defaultValue('/contao')
```
fritzmg: Should we validate for a leading slash here?

m-vo: Something like `/^\/\S*[^\s\/]$/` maybe? :upside_down_face:

richardhj: Validation added in 35ace90 :-)

fritzmg: idk, I was just wondering. What would even happen if you defined a route (prefix) without a leading slash in Symfony? 🤔 May be it does not matter anyway?

aschempp:
I think this came up numerous times in the last 15 years. The general go-to answer I can remember is that we don't support that because ultimately it should not matter. If it does matter, it can never be for security, because there are plenty of other ways to find out a site is built in Contao. If you do want to "obfuscate" that you're using Contao, its most likely you want to present the application as "your own" to a client, which is highly discouraged from my Open Source view.

I'm up for discussing this again though. We could allow it as a new feature in Contao 5, since I agree its sometimes asked for and ultimately does not harm technically, it only harms the spirit 🙃

richardhj:
> If you do want to "obfuscate" that you're using Contao, its most likely you want to present the application as "your own" to a client, which is highly discouraged from my Open Source view.

Contradictory, for me, allowing to customize the application is actually the spirit ✨ of non-proprietary software.

For me, this question is about whether Contao wants to become a CMS that can be tailored to different project's needs (it's just a small change for sure).

I see it the way that "opening" the system allows more use cases for Contao -- that in turn results in more users and contributors.

fritzmg (dismissed):
Generally I am fine with this. Other systems like Magento also allow this and we occasionally get this request too (though it was never that important to any of our customers).

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -80,6 +80,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('contao.security.two_factor.enforce_backend', $config['security']['two_factor']['enforce_backend']);
         $container->setParameter('contao.localconfig', $config['localconfig'] ?? []);
         $container->setParameter('contao.backend', $config['backend']);
+        $container->setParameter('contao.backend.route_prefix', $config['backend']['route_prefix']);
```
leofeyer, line 90: Why is this necessary? The value should already be set by the line above, shouldn't it?

aschempp: Pretty sure it needs to be a string parameter if you want to use it in a route config.

leofeyer: It is a string parameter already?

aschempp: `contao.backend` is not a string parameter I would say 😅

aschempp: also, this might be a problem since the configuration of `contao.backend` is used to generate the toolbar info icon, isn't it?

richardhj: I will check if `$config['backend']['route_prefix']` gets normalized to `contao.backend.route_prefix` correctly :-)

richardhj: > also, this might be a problem since the configuration of contao.backend is used to generate the toolbar info icon, isn't it?

Hm? I am just using the same configuration node that Ralf introduced in #1999

aschempp: Looking at the PR it seems that shouldn't be an issue 😊

leofeyer: I think we should discuss this on a more general level. Take lines 84 to 86 for example:

```php
$container->setParameter('contao.intl.locales', $config['intl']['locales']);
$container->setParameter('contao.intl.enabled_locales', $config['intl']['enabled_locales']);
$container->setParameter('contao.intl.countries', $config['intl']['countries']);
```

Do we explicitly want to set it like this or would it be enough to do the following?

```php
$container->setParameter('contao.intl', $config['intl']);
```

@contao/developers /cc

aschempp: I think you're confusing something. Most of these parameters are used in service configuration. So if your service argument requires the countries only, you must have them as separate parameters. This especially applies to the route prefix, because the parameter is used in route config you **cannot** use an array parameters.

richardhj: Yep, the parameter `contao.backend.route_prefix` is required. There is no normalization process in the extension that "unpacks" the `contao.backend` parameter, as one could think.

Without defining `contao.backend.route_prefix`, the following issue occurs:

Screenshot 2021-11-05 at 19 50 06

Closing this.

richardhj:
What's still missing is a functional test. Someone got a hint on setting it up? 🤗

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
leofeyer: Unfortunately, this problem will not disappear by resolving the conversation. It shows us that the changes from #1999 are incorrect, because they take the entire `contao.backend` namespace instead of separate sub-namespaces. We will have to fix this first.

leofeyer: See https://github.com/contao/contao/pull/3653

richardhj: To be honest, I don't see a problem or a necessity for #3653 since the `contao.backend` "namespace" is not blocked.
You were able to fetch the route prefix by `$container->get('contao.backend')['route_prefix']` or `$container->get('contao.backend.route_prefix')`. The latter one was only required because it is used directly as a service parameter.

leofeyer (approved):

bytehead (approved):

leofeyer:
Thank you @richardhj.

manager-bundle/src/Resources/skeleton/config/security.yml
aschempp: However, this "security" measure would no longer work if we the path is reconfigured, right? Can't we use the parameter in this path as well?

richardhj: In security.yml, the container parameters are not available (yet).
I consider changing the backend path as "you better know what you do". The documentation uses the _scope=backend since ever.

richardhj: This would also only affect custom backend routes that usually still have the /contao prefix in the path until one manually updates to the new config parameter.

aschempp: Sorry I might be unclear. The **firewall** uses the scope to run, but it does not use the scope to protected the routes! If you do not have an `isGranted` check, and the current `access_control` no longer takes place, you will have a backend firewall but access will still be granted!

fritzmg: So in the current state of `4.x` there will be no access control check anymore, if the back end path is changed?

richardhj: > If you do not have an `isGranted` check, and the current `access_control` no longer takes place, you will have a backend firewall but access will still be granted!

Ok, I guess that's correct. But I want to add that this is not the case when you extend `BackendCustom` or `BackendMain`.

fritzmg: But it would be for custom back end controllers. Currently we only mention the following in the docs:

> If the scope is `backend`, the route is automatically protected by the `contao_backend` firewall.

richardhj: While using a custom route prefix, I don't use `isGranted()` in my custom controllers and get redirected to /contao/login for unauthenticated requests. So there must be a different security mechanism in place already.

aschempp: see https://github.com/contao/contao/pull/3720

raynor84:
Hi,
I've tested it in 4.13.18.
A minor issue: 
After Logging in, the path should redirect to the new backend-url instead of contao.

kind regards

Stefan

PR #3475, opened 2021-09-15
Allow WebP to PNG/JPG conversion

fritzmg:
Currently you can only let Contao convert from JPG/PNG _to_ WEBP via database image size definitions - but not the other way around. The only way to implement seamless JPG/PNG to WEBP conversion _and vice versa_ is via the bundle configuration, e.g.:

```yaml
contao:
    image:
        sizes:
            example:
                width: 100
                formats:
                    jpg: [webp, jpg]
                    png: [webp, png]
                    webp: [webp, jpg]
```

With this PR the same will also be possible via the image size configuration in the database/backend:

This allows you to seamlessly convert selected JPGs to WEBP and selected WEBP images to JPG (for the fallback) - which in turn means it will not matter whether your editors select a JPG, PNG or WEBP image in the backend. The output will always be

```html
<picture>
  <source srcset="foobar.webp" type="image/webp" …>
  <img src="foobar.jpg" …><!-- or .png -->
</picture>
```

for example.

As discussed with @ausi previously we decided to implement this as a bugfix.

_Note:_ this will allow you to select both _WEBP to JPG_ and _WEBP to PNG_ at the same time, though as far as I tested it does not matter. Contao will always pick _WEBP to JPG_ in such a case. @ausi should we implement some special handling or error message when both are selected? The bundle configuration would of course give you an error, if you tried to use two `webp` keys.

_Question:_ should we also allow _WEBP to GIF_ there?

### DIFF
--- a/core-bundle/src/Image/PictureFactory.php
+++ b/core-bundle/src/Image/PictureFactory.php
@@ -28,6 +28,7 @@ use Contao\StringUtil;
 class PictureFactory implements PictureFactoryInterface
 {
     private const ASPECT_RATIO_THRESHOLD = 0.05;
+
     private const FORMATS_ORDER = [
         'jxl' => 1,
         'avif' => 2,
--- a/core-bundle/src/Image/PictureFactory.php
+++ b/core-bundle/src/Image/PictureFactory.php
@@ -28,6 +28,16 @@ use Contao\StringUtil;
 class PictureFactory implements PictureFactoryInterface
 {
     private const ASPECT_RATIO_THRESHOLD = 0.05;
+    private const FORMATS_ORDER = [
+        'jxl' => 1,
+        'avif' => 2,
+        'heic' => 3,
+        'webp' => 4,
+        'png' => 5,
+        'jpg' => 6,
+        'jpeg' => 7,
+        'gif' => 8,
+    ];
 
     /**
      * @var array
@@ -168,22 +178,29 @@ class PictureFactory implements PictureFactoryInterface
                 if (null !== $imageSizes) {
                     $options->setSkipIfDimensionsMatch((bool) $imageSizes->skipIfDimensionsMatch);
 
-                    $config->setFormats(array_merge(
-                        [],
-                        ...array_map(
-                            static function ($formatsString) {
-                                $formats = [];
+                    $formatsString = implode(';', StringUtil::deserialize($imageSizes->formats, true));
+                    $formats = [];
 
-                                foreach (explode(';', $formatsString) as $format) {
-                                    [$source, $targets] = explode(':', $format, 2);
-                                    $formats[$source] = explode(',', $targets);
-                                }
+                    foreach (explode(';', $formatsString) as $format) {
+                        [$source, $targets] = explode(':', $format, 2);
+                        $targets = explode(',', $targets);
 
-                                return $formats;
-                            },
-                            StringUtil::deserialize($imageSizes->formats, true)
-                        )
-                    ));
+                        if (!isset($formats[$source])) {
+                            $formats[$source] = $targets;
+                            continue;
+                        }
+
+                        $formats[$source] = array_unique(array_merge($formats[$source], $targets));
+
+                        usort(
+                            $formats[$source],
+                            static function ($a, $b) {
+                                return (self::FORMATS_ORDER[$a] ?? $a) <=> (self::FORMATS_ORDER[$b] ?? $b);
+                            }
+                        );
+                    }
+
+                    $config->setFormats($formats);
                 }
 
                 if ($imageSizes) {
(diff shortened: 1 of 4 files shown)


ausi:
> should we implement some special handling or error message when both are selected?

I think we should handle it when reading the values so that if both are selected we use `webp:webp,png,jpg`. We would need this logic anyways for future additions like AVIF, HEIC and JXL.

ausi:
> should we also allow _WEBP to GIF_ there?

I would not allow that as I can not think of a valid use case that would not need special tweaks anyways.

fritzmg:
> I think we should handle it when reading the values so that if both are selected we use `webp:webp,png,jpg`

But does that make sense? The output would then be

```html
<picture>
  <source srcset="foobar.webp" type="image/webp" …>
  <source srcset="foobar.png" type="image/png" …> 
  <img src="foobar.jpg" …>
</picture>
```

When would a client ever choose the `image/png` `source` in this case?

// edit: but I see what you mean about other formats. Conceivably we can even allow the same for JPG and PNG (i.e. convert JPG to PNG and WEBP and JPX or whatever).

ausi:
> When would a client ever choose the `image/png` `source` in this case?

~~Always, but~~ I get your point, ~~the JPEG version would never be used by any browser.~~ _(actually old browsers not supporting `<source>` would show the JPEG 🙈)_

I think we should predefine a list of always available formats (jpg, gif, png) and stop at the first occurence of them. So `webp:webp,png,jpg` would get changed to `webp:webp,png`, but `jpg:jxl,jpg;jpg:avif,jpg;` should get merged to `jpg:jxl,avif,jpg` IMO.

Should I provide a different pull request for that as it is not directly related or needed for this PR?

fritzmg:
> Should I provide a different pull request for that as it is not directly related or needed for this PR?

I think it's related enough and this PR is pretty small anyway. PR for the PR? ;)

fritzmg:
Btw. if anyone needs this right way, you can add these new options via the following application customisations:

```php
// src/EventListener/DataContainer/AddImageSizeConversionFormatsListener.php
namespace App\EventListener\DataContainer;

use Contao\DataContainer;

/**
 * @Callback(table="tl_image_size", target="fields.formats.options")
 */
class AddImageSizeConversionFormatsListener
{
    public function __invoke(DataContainer $dc = null): array
    {
        $formats = (new \tl_image_size())->getFormats($dc);

        if (!\in_array('png:webp,png', $formats, true) || \in_array('webp:webp,png', $formats, true)) {
            return $formats;
        }

        return array_merge($formats, ['webp:webp,png', 'webp:webp,jpg']);
    }
}
```
```php
// contao/languages/en/tl_image_size.php
$GLOBALS['TL_LANG']['tl_image_size']['webp:webp,png'] = 'WEBP to PNG';
$GLOBALS['TL_LANG']['tl_image_size']['webp:webp,jpg'] = 'WEBP to JPG';
```

ausi (dismissed):

core-bundle/src/Image/PictureFactory.php
```diff
@@ -168,22 +178,29 @@ private function createConfig($size): array
...
-
-                                return $formats;
-                            },
-                            StringUtil::deserialize($imageSizes->formats, true)
-                        )
-                    ));
+                    $formatsString = implode(';', StringUtil::deserialize($imageSizes->formats, true));
+                    $formats = [];
+
+                    foreach (explode(';', $formatsString) as $format) {
```
leofeyer, line 185: If you remove the `implode(';'` above, you do not have to use `explode(';'` here, do you?

core-bundle/src/Image/PictureFactory.php
```diff
@@ -28,6 +28,16 @@
 class PictureFactory implements PictureFactoryInterface
 {
     private const ASPECT_RATIO_THRESHOLD = 0.05;
+    private const FORMATS_ORDER = [
```
leofeyer, line 32:
```suggestion

    private const FORMATS_ORDER = [
```

leofeyer (changes requested):

core-bundle/src/Image/PictureFactory.php
ausi: I do, the strings themself can include `;` like the JPEG one does. (`jpg:webp,jpg;jpeg:webp,jpeg`)

ausi (approved):

core-bundle/src/Image/PictureFactory.php
leofeyer: I see. 👍

leofeyer:
Thank you @fritzmg.

fritzmg:
> Thank you @fritzmg.

and @ausi ;)

PR #3481, opened 2021-09-16
Fix inconsistencies in the search field rendering

leofeyer:


fritzmg (approved):
Also tested in Firefox on Windows.

PR #3483, opened 2021-09-18
Render the Welcome Screen in the Backend using Fragments

richardhj:
| Q                | A
| -----------------| ---
| Fixed issues     | 
| Docs PR or issue | tba.

This PR allows adding new widgets/fragments to the dashboard. Often, extension developers used to use the `getSystemMessages` hook to show information to the user on the home screen. This always has been a hack.

```php
// src/Controller/Dashboard/HelloWidgetController.php

<?php

namespace App\Controller\Dashboard;

use Contao\CoreBundle\Controller\BackendDashboard\AbstractDashboardWidgetController;
use Contao\CoreBundle\ServiceAnnotation\DashboardWidget;

/**
 * @DashboardWidget()
 */
class HelloWidgetController extends AbstractDashboardWidgetController
{
    {
        return new Response('This section will be added on the backend dashboard.');
    }
}
```

Screenshot 2021-09-18 at 19 09 56

Supersedes #2748.

### DIFF
--- a/core-bundle/src/ContaoCoreBundle.php
+++ b/core-bundle/src/ContaoCoreBundle.php
@@ -42,6 +42,7 @@ use Contao\CoreBundle\Event\PreviewUrlCreateEvent;
 use Contao\CoreBundle\Event\RobotsTxtEvent;
 use Contao\CoreBundle\Event\SlugValidCharactersEvent;
 use Contao\CoreBundle\Fragment\Reference\ContentElementReference;
+use Contao\CoreBundle\Fragment\Reference\DashboardWidgetReference;
 use Contao\CoreBundle\Fragment\Reference\FrontendModuleReference;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\SecurityExtension;
 use Symfony\Cmf\Component\Routing\DependencyInjection\Compiler\RegisterRouteEnhancersPass;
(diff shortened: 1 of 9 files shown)


m-vo:
Maybe we should base this on a generic abstract back end controller, compare to the draft in #3180.

aschempp (commented):
Fragments are not the right thing for this case in my opinion. Fragments are useful if each part its their own response, possible with cache time or even as an ESI request. In the back end case, rendering fragments through subrequests seems a lot of overhead with no benefit to me.

richardhj:
What is the alternative, then?

> Fragments are useful if each part its their own response, possible with cache time or even as an ESI request.

The dashboard looks like this case for me.

aschempp:
The back end does not cache at all. Would you cache individual parts of the home screen?

m-vo:
IMHO widgets could be a great use case for [embedded controllers](https://symfony.com/doc/current/templates.html#templates-embed-controllers):

```twig
{% block widgets %}
  {{ render(controller('App\\Controller\\FooBarWidgetController::baz')) }}
{% endblock %}
```

Every extension or the app could just extend the respective block and add their controller. Maybe a bit less control than a registry but near zero overhead and an easily customizable solution (from an app's perspective).

aschempp:
Embedded controllers are exactly ~fragments~ subrequests referenced by class name. 🙃

On the concept of the home screen: fragments are meant to have a type and be configured and repeatable. On the home screen, it would mean _every existing fragment_ would always be there and only ever be there once.

Maybe we can extend that concept though, or come up with something similar. After all, all we need is a list of home screen "widgets", right? So maybe priority tagged services would be sufficient?

richardhj:
> Maybe we can extend that concept though, or come up with something similar. After all, all we need is a list of home screen "widgets", right? So maybe priority tagged services would be sufficient?

In that each service returns a response? Then all we skip is the fragment registry, no? Or how does it differ exactly? :)

PR #3488, opened 2021-09-21
Add a configuration for allowed URL protocols

### LINKED ISSUE #3473: Make list of allowed protocols configurable
One of our customers uses Skype-Links `href="skype:name?call"` on his website, which are now encoded. The only way to allow them is to make this list of allowed protocols configurable, right? What do you think?
https://github.com/contao/contao/blob/a7628093a5710bbbb8f0c97950a70e89f3727ba7/core-bundle/src/Resources/contao/library/Contao/StringUtil.php#L874-L876

MarkejN:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3473
| Docs PR or issue | contao/docs#840

This adds a new configuration option:

```yaml
contao:
    sanitizer:
        allowed_url_protocols:

            # Defaults:
            - https
            - http
            - ftp
            - mailto
            - tel
            - data
            - skype
            - whatsapp
```
However, I'm not sure if it is a good idea to directly use the input in the `preg_match` pattern? To help with that, I added a validation based on this [specification](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1):

>Scheme names consist of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").  Although schemes are case- insensitive, the canonical form is lowercase and documents that specify schemes must do so with lowercase letters.  An implementation should accept uppercase letters as equivalent to lowercase in scheme names (e.g., allow "HTTP" as well as "http") for the sake of robustness but should only produce lowercase scheme names for consistency.
>
>scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -426,7 +426,7 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->arrayNode('allowed_url_protocols')
                     ->prototype('scalar')->end()
-                    ->defaultValue(['http', 'https', 'ftp', 'mailto', 'tel', 'data'])
+                    ->defaultValue(['http', 'https', 'ftp', 'mailto', 'tel', 'data', 'skype', 'whatsapp'])
                     ->validate()
                         ->always(
                             static function (array $protocols): array {
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -121,6 +121,23 @@ class Configuration implements ConfigurationInterface
                         )
                     ->end()
                 ->end()
+                ->arrayNode('allowed_protocols')
+                    ->prototype('scalar')->end()
+                    ->defaultValue(['http', 'https', 'ftp', 'mailto', 'tel', 'data'])
+                    ->validate()
+                        ->always(
+                            static function (array $protocols): array {
+                                foreach ($protocols as $protocol) {
+                                    if (!preg_match('/^[a-z][a-z0-9\-+.]*$/i', (string) $protocol)) {
+                                        throw new \InvalidArgumentException(sprintf('The protocol name "%s" must be a valid URI scheme.', $protocol));
+                                    }
+                                }
+
+                                return $protocols;
+                            }
+                        )
+                    ->end()
+                ->end()
                 ->append($this->addImageNode())
                 ->append($this->addSecurityNode())
                 ->append($this->addSearchNode())
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -121,27 +121,11 @@ class Configuration implements ConfigurationInterface
                         )
                     ->end()
                 ->end()
-                ->arrayNode('allowed_protocols')
-                    ->prototype('scalar')->end()
-                    ->defaultValue(['http', 'https', 'ftp', 'mailto', 'tel', 'data'])
-                    ->validate()
-                        ->always(
-                            static function (array $protocols): array {
-                                foreach ($protocols as $protocol) {
-                                    if (!preg_match('/^[a-z][a-z0-9\-+.]*$/i', (string) $protocol)) {
-                                        throw new \InvalidArgumentException(sprintf('The protocol name "%s" must be a valid URI scheme.', $protocol));
-                                    }
-                                }
-
-                                return $protocols;
-                            }
-                        )
-                    ->end()
-                ->end()
                 ->append($this->addImageNode())
                 ->append($this->addSecurityNode())
                 ->append($this->addSearchNode())
                 ->append($this->addCrawlNode())
+                ->append($this->addSanitizerNode())
             ->end()
         ;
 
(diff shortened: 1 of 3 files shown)


core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -871,9 +871,11 @@ static function ($match)
 			. ');?'               // Optional semicolon
 		. ')i';
 
+		$arrAllowedProtocols = System::getContainer()->getParameter('contao.allowed_protocols');
+
 		// URL-encode colon to prevent disallowed protocols
 		if (
-			!preg_match('@^(?:https?|ftp|mailto|tel|data):@i', self::decodeEntities($strString))
+			!preg_match('@^(?:' . implode('|', array_map('preg_quote', $arrAllowedProtocols)) . '):@i', self::decodeEntities($strString))
```
ausi:
```suggestion
			!preg_match('(^(?:' . implode('|', array_map('preg_quote', $arrAllowedProtocols)) . '):)i', self::decodeEntities($strString))
```

Default delimiters need to be used together with `preg_quote()`

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -109,6 +109,23 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->always(static fn (string $value): string => Path::canonicalize($value))
                     ->end()
                 ->end()
+                ->arrayNode('allowed_protocols')
```
ausi: Allowed Protocols is a bit too generic I think.

Maybe add something like url, link, html attribute to the name to make it more clear?

Or how about moving it to its own section `html_filter`?
The allowed attributes and tags can be moved here too then in the future.

m-vo: > Allowed Protocols is a bit too generic I think

Yes, that sounds like Contao would disallow others in general.

Maybe `contao.sanitizer.allowed_url_protocols`?

The problem is, this targets a helper function where the usage or promoted usage is unclear (yet).

ausi: +1 for `contao.sanitizer.allowed_url_protocols`

richardhj:
From a practical perspective, I'd vote to just allow "skype://" (and similar) as a protocol (see #3473). There just aren't that many use-cases for making this configurable IMHO.

MarkejN:
I'd like to have a configuration, because then I could always just quickly add the protocol I need. But it's true, my only use case right now would actually be skype.

aschempp:
I kinda agree with @richardhj there shouldn't be a need to configure this. We should allow what we know is safe (or block what is unsafe) and quickly add something if we forget it. 

I also dislike the `contao.sanitizer` namespace since Contao does not have anything we call a `sanitizer`. The implementation is part of `StringUtil` or `Input` if I'm right...?

leofeyer:
@MarkejN This looks like a bugfix that should be merged into Contao 4.9. Can you please rebase this PR on the `4.9` branch?

MarkejN:
> I kinda agree with @richardhj there shouldn't be a need to configure this. We should allow what we know is safe (or block what is unsafe) and quickly add something if we forget it.

I'm fine with either way. Should I revert my commit and just add skype to the list?

ausi:
> Should I revert my commit and just add skype to the list?

No, I think we agreed on adding a configuration option as you did in this pull request.

The open question is how we should name the config option, I’m in favor of `contao.sanitizer.allowed_url_protocols`.

leofeyer:
That's the one then. 😄

Total-Reality:
Sounds good, you already know in which contao version this feature will be added? 

Btw. you could also add a new option to the dca/eval as I wrote here:
https://github.com/contao/contao/issues/3828

MarkejN:
I didn't have a lot of time recently, but I will finish the PR this weekend.

MarkejN:
I remember some failed tests when this PR was targeted at the 4.x branch. Anything else I need to add here?

ausi (dismissed):

ausi:
> Anything else I need to add here?

I don’t think so, no.

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -417,6 +418,33 @@ static function (array $uris): bool {
...
 
+    private function addSanitizerNode(): NodeDefinition
+    {
+        return (new TreeBuilder('sanitizer'))
+            ->getRootNode()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('allowed_url_protocols')
+                    ->prototype('scalar')->end()
+                    ->defaultValue(['http', 'https', 'ftp', 'mailto', 'tel', 'data'])
```
fritzmg: I would may be also add `skype` and `whatsapp` to the defaults, but otherwise 👍

fritzmg (dismissed):

aschempp (dismissed):
Code looks good to me, but sounds like a feature for 4.13 and not a bugfix for 4.9?

fritzmg:
I think it was implemented as a bugfix because in 4.9 there is currently no way of preventing the encoding of these protocols otherwise.

ausi:
> sounds like a feature for 4.13 and not a bugfix for 4.9?

IMO bugfix, because the encoding of the protocols was added in a security release.

core-bundle/src/DependencyInjection/Configuration.php
MarkejN: I agree 👍
Added in https://github.com/contao/contao/commit/15985ec5ac2b3b4540fb2ef0f6ecb1e1c3fa30f2.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @MarkejN.

PR #3489, opened 2021-09-21
Add the "send newsletter" event

### LINKED ISSUE #2891: Add newsletter hook before sending to customize HTML
At the moment you can't manipulate the newsletter before it is sent. With a hook you could e.g. get the HTML from an article or from content elements and don't have to work with insert tags. This should be an easy pick and maybe i can send a PR.

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2891
| Docs PR or issue | - todo -

The hook is executed right before sending each email of a newsletter. It passes the email object (with the parsed content), the newsletter data, the recipient data and – if HTML is not disabled – the newsletter template object.

This way, one can adjust the email data for each newsletter email as requested in #2891.

Additionally, the hook adds the possibility to skip sending a newsletter to a specific recipient by returning `false` from the callback.

If we agree on the hook, I can submit an update to [contao/docs](https://github.com/contao/docs).


fritzmg (changes requested):
@SeverinGloeckle thank you for your contribution 👍 

We do not want to add new hooks as a new feature though (only for bugfixes). Instead it should be an event.

ausi:
> Instead it should be an event.

👍 But we also want events to use as few legacy classes as possible. So we should check if we could replace some of the hook arguments with something different. For example the email object should probably be replaced with an object from the Symfony mailer (if possible in a BC way).

SeverinGloeckle:
I see the reasons for events instead of more proprietary hooks.

Though, in my opinion it would be confusing to have an event `PrepareNewsletter` and a hook `sendNewsletter` within the same (legacy) class doing very similar things. 

I added #3492 to discuss the issue in detail so we can decide what to do here.

ausi:
> This way, one can adjust the email data for each newsletter email as requested in #2891.

Would decorating the `mailer` service maybe suffice to solve that use case?

fritzmg:
> Would decorating the `mailer` service maybe suffice to solve that use case?

There you would lack the context, i.e. you would have no access to the newsletter object, the template, etc. The hook/event could be useful to enrich the newsletter template with additional data for instance, among other things.

SeverinGloeckle:
I have re-implemented the hook via a new `SendNewsletterEvent`.

The event knows neither the `Email` object nor the `BackendTemplate` of the newsletter to keep it independent from any legacy logic.

However, it allows to pass around arrays of newsletter data and recipient data. This way, a consumer can access custom data from the newsletter (e.g. the uncompiled HTML content) and from the recipient (e.g. a custom field like `tl_newsletter_recipients.company`).

The event uses the parsed newsletter template as its HTML by default. The HTML can be changed through the event, but a developer probably has to parse a custom template (or reparse the original template) to get the desired HTML.


Example listener logic:

````php
if ($event->allowsHtml() && $html = $event->getNewsletterValue('content')) {
  $html = '<strong>' . $html . '</strong>';
  $event->setHtml($html);
}

if (preg_match('/@contao\.org/i', $event->getRecipientAddress())) {
  $event->setPreventSubmission(true);
}

$event->setText('New text');
$event->setRecipientAddress('foo@bar.baz');
````

Contrary to the current `sendNewsletter` hook, the new `SendNewsletterEvent` allows to modify each newsletter email **before** it is sent and also allows to prevent the submission of a newsletter mail. 

From what I see, almost all use cases of the hook can also be implemented via the event.
Therefore, it should be possible to deprecate the `sendNewsletter` hook.

Notes:
- If the HTML or newsletter text are modified through the event, the parameters `$html` and `$text` for the `sendNewsletter` hook are not updated. They still contain the unparsed HTML and text of the newsletter like before (no parsed template for the HTML, no parsed tokens for the text). The updated data is available via `$objEmail->text` and `$objEmail->html`.

Open questions:
- Should the `sendNewsletter` hook be deprecated in favor of the event?
- Is the test setup for the newsletter-bundle complete and accurate? (there has been no testsuite for the bundle so far)
- How to add `"Contao\\NewsletterBundle\\Tests\\": "newsletter-bundle/tests/"` to the monorepo tools?

**Comment by fritzmg** [comment]
Personally I would prefer it, if the email instance can be modified directly, so that you could control _any_ parameter of the newsletter email and not just the ones that the event implements.

**Comment by SeverinGloeckle** [comment]
Thank you for the feedback @fritzmg.

I fully agree that it would be more flexible and convenient to be able to edit the email instance directly.

However, my intention was to keep the event as abstract and detached from the current logic as possible and in particular independent from Contao's email object. This way, the event and corresponding listeners can be used in the future, even if the logic for sending the newsletter changes.

File: newsletter-bundle/src/Event/SendNewsletterEvent.php [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public function setHtml(string $html): self
+    {
+        $this->html = $html;
+
+        return $this;
+    }
+
+    public function allowsHtml(): bool
```
leofeyer: IMHO we should adopt one of the following naming strategies here:

### 1. `is()` and `set()`

```
public function isAllowHtml(): bool;
public function setAllowHtml(bool $allowHtml);
```

As e.g. done in Composer with the `isClassMapAuthoritative()` and `setClassMapAuthoritative()` methods.

### 2. `is()` and `enable()` + `disable()`

```
public function isHtml(): bool;
public function allowHtml();
public function disallowHtml();
```

As e.g. done in Symfony with the `isDebug()`, `enableDebug()` and `disableDebug()` methods.

### 3. Default setter and getter

```
public function getAllowHtml(): bool;
public function setAllowHtml(bool $allowHtml);
```

As e.g. done in Doctrine entities.

### 4. Speaking methods

```
public function allowsHtml(): bool;
public function allowHtml(bool $allowHtml);
```

The latter seems to be a rather uncommon approach though, because setter methods usually start with `set` and getter methods with `get` (if they return a value) or `is`, `has` or `can` (if they return a boolean).

leofeyer (changes requested):

newsletter-bundle/src/Event/SendNewsletterEvent.php
m-vo: In this specific case I think it's not about allowing something but sending a HTML version along with the text or not? So why not remove this extra property and just allow a nullable parameter in `setHtml(?string $html)`?

SeverinGloeckle: Thanks for the detailed overview!

I see your point and have adjusted the naming in https://github.com/contao/contao/pull/3489/commits/f1af37d81ae3d7f0d9d755b03f20131f36e29a75 to:
* `isSubmissionPrevented` and `setSubmissionPrevented`
* `isHtmlAllowed` and `setHtmlAllowed`

We could get away with saying `setHtml(null)` is the same as `setHtmlAllowed(false)` for most use cases. However, it wouldn't be possible for the newsletter to be initialized without HTML (`html = null`) but allowing HTML to be added by listeners – and the other way around: for an event to inform all following listeners that HTML is disallowed. The latter case could be handled via proper listener priority (last listener calls `setHtml(null)` but in my opinion, an additional flag is more meaningful here (and prevents possible questions like is `setHtml(null)` the same as `setHtml('')`).

newsletter-bundle/src/Event/SendNewsletterEvent.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public function setRecipientAddress(string $recipientAddress): self
+    {
+        $this->recipientAddress = $recipientAddress;
+
+        return $this;
+    }
+
+    public function isSubmissionPrevented(): bool
```
leofeyer: How do you like `isSkipSending()` and `setSkipSending()` instead?

newsletter-bundle/src/Resources/contao/classes/Newsletter.php
```diff
@@ -396,6 +415,22 @@ protected function sendNewsletter(Email $objEmail, Result $objNewsletter, $arrRe
 			$objEmail->imageDir = System::getContainer()->getParameter('kernel.project_dir') . '/';
 		}
 
+		$event = (new SendNewsletterEvent($arrRecipient['email'], $objEmail->text, $objEmail->html))
+			->setHtmlAllowed($objNewsletter->sendText != 1)
```
leofeyer:
```suggestion
			->setHtmlAllowed(!$objNewsletter->sendText)
```

newsletter-bundle/src/Event/SendNewsletterEvent.php
aschempp: How about `discardNewsletter` or `preventSending`? Similar to `stopPropagation()`. Also, this should probably call `stopPropagation()`?

SeverinGloeckle: `isPreventSending` and `setPreventSending` sounds good to me. I have renamed the property and methods in https://github.com/contao/contao/pull/3489/commits/69e6b717477f6d09728b16c09b884ec14ea52e20

I don't think the event should call `stopPropagation()` automatically – this should be a conscious decision by the developer. There might be cases where one listener prevents the sending while another listener wants to explicitly re-enable some prevented newsletters.

ausi (dismissed):
I think `newsletter-bundle` needs to be added here too: https://github.com/contao/contao/blob/6a815a472a46c44c4dbf3885f44c229a26fb8510/.github/workflows/ci.yml#L320

> From what I see, almost all use cases of the hook can also be implemented via the event.
> Therefore, it should be possible to deprecate the `sendNewsletter` hook.

We should then add something like this:

```php
trigger_deprecation('contao/core-bundle', '4.13', 'Using the "sendNewsletter" hook has been deprecated and will no longer work in Contao 5.0. Use the SendNewsletterEvent instead.');
```

to here I think: https://github.com/contao/contao/blob/69e6b717477f6d09728b16c09b884ec14ea52e20/newsletter-bundle/src/Resources/contao/classes/Newsletter.php#L450-L452

Apart from that, 👍❤️🎉

leofeyer (approved):

leofeyer:
Thank you @SeverinGloeckle.

SeverinGloeckle:
Thanks for providing the missing parts @leofeyer!

SeverinGloeckle:
There already is a documentation prepared for the event, see https://github.com/contao/docs/pull/929

PR #3496, opened 2021-09-23
Handle edge cases in StringUtil::convertEncoding()

### LINKED ISSUE #3493: Folder name "0" is not displayed
**Affected version(s)**

Contao 4.9+

**Description**

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3493

There is an additional check via `is_string` and `is_numeric` to only allow encoding for simple stringable parameters and to prevent errors from the `mb_` functions when called with e.g. an array. For invalid parameters, an empty string is returned.

This also makes sure the method returns an empty string when called with a non-string parameter and the same `$to` and `$from` encoding. Until now, the method has returned the unchanged parameter (which could be an array, object etc.).

Additionally, the return value is always cast to a string if the encodings match to make sure integers and floats are returned as strings (which they were for different input & output encodings but not for matching encodings).

**Test case**
This also adds a simple unit test for `StringUtil::convertEncoding`.

One could add a lot more test cases for the concrete encodings, but since the `StringUtil` only uses `mb_encode`, `mb_decode`, `mb_detect_encoding` and `mb_convert_encoding` internally, these tests should not be necessary.


SeverinGloeckle:
I guess some of my assumptions about the correct return value for edge cases are debatable. I tried to keep in line with what I understood as the primary use-case of `StringUtil::convertEncoding`.

Throwing an `InvalidArgumentException` sounds like a good alternative if anything that's not a string is passed to the conversion.

With PHP 8 we could check the argument for `Stringable` as well in the future.

leofeyer:
What about an early return if `$str` is not a string?

```php
if (!\is_string($str)) {
    return $str;
}
```

Casting the return value to string would be a behavior change that might break third-party extensions.

SeverinGloeckle:
You are right about the behaviour change. However, depending on how highly you value the `@return string` annotation, it was buggy to assume a return value other than a string anyway.

The method could keep returning non-string values for non-stringable arguments but that would imply that the method cannot get a proper return type `:string` in the future.

At least there should be checks for both `is_string` and `is_numeric` and if those pass, the return value should be cast to a string.

Otherwise this (current) behaviour is highly confusing and dangerous (see test cases _Integer argument..._ and _Float argument..._):

````php
StringUtil::convertEncoding(42, 'ASCII', 'UTF-8') // String '42'
StringUtil::convertEncoding(42, 'ASCII') // Integer 42
StringUtil::convertEncoding(42, 'ASCII', 'ASCII') // Integer 42
StringUtil::convertEncoding(42, 'UTF-8', 'UTF-8') // Integer 42
StringUtil::convertEncoding(42, 'UTF-8') // String '42'
````

In my opinion, all calls should return the string `'42'`.

**Comment by leofeyer** [comment]
We should add a deprecation notice if a non-string argument is passed:

```php
if (!\is_string($str)) {
    trigger_deprecation('contao/core-bundle', '4.9', 'Passing a non-string argument to StringUtil::convertEncoding() has been deprecated an will no longer work in Contao 5.0');

    return $str;
}
```

Then we can add a `string` type hint and a `: string` return type in Contao 5.

SeverinGloeckle:
I like the deprecation.

To be completely accurate, the return type would then have to be adjusted to `mixed` as well.

However, a simple `is_string` check would also bring a behaviour change for numeric values since the method would always return an `int` or `float` in this case (while it was returning a string or a numeric value based on the encodings, see examples above).

Thinking about this, we probably should consider stringable arguments as well do get this right for the long term.

If we have this:

````php
// Any class implementing __toString()
$context = new ContaoContext('foo');
$encoded = StringUtil::convertEncoding($context, 'ASCII', 'UTF-8');
````

Should `$encoded` be the `ContaoContext` or its encoded string representation?
Should a deprecation be thrown for a stringable (numeric or `\Stringable`) argument?

**Comment by leofeyer** [comment]
@contao/developers WDYT?

**Comment by ausi** [comment]
IMO always return string. `StringUtil::convertEncoding(…, 'UTF-8')` ist the most common usage and this one already returned a string.

**Comment by m-vo** [comment]
> IMO always return string

+1 on that one. The class is called **String**Util after all and encoding literals doesn't make much sense.

**Comment by SeverinGloeckle** [comment]
I added two utility methods:
- `StringUtil::isStringable` to check whether a given value can be cast to a string.
- `StringUtil::toString` to cast a value to a string if possible.

Currently, these utility methods are used within `StringUtil::convertEncoding` only but could be used to improve other methods of `StringUtil` as well to handle various stringable arguments. This should probably be another pull request.

The utility methods could be made private to be used internally only, but in my opinion they can actually provide value for the whole system to ensure a consistent handling of strings (and stringables).

Note:
The two new utility methods have their return type defined (while the other methods in StringUtil do not for 4.9, but some do in later versions). I'm not sure if there is a standard for when and why methods should declare a return type – maybe the return type should be added for later Contao versions only.

### The actual `StringUtil::convertEncoding` method
Using the new utility methods, `StringUtil::convertEncoding` now always returns a string for any stringable argument. If the argument is not stringable, a deprecation is triggered and *an empty string is returned* **<--❗️ I'm not sure if we agreed on this or only on _return a string for literals_.❗️**

With PHP 8 and Contao 5, the deprecation could be dropped and the method can be adjusted to:

````php
public static function convertEncoding(string|\Stringable $str, string $to, ?string $from=null): string
````

**Comment by leofeyer** [comment]
As discussed in Mumble on September 30th, we want to cast the return value to string.

File: core-bundle/src/Resources/contao/library/Contao/StringUtil.php [hunk 1]
```diff
@@ -917,11 +917,20 @@ public static function resolveFlaggedUrl(&$url)
...
 	{
-		if (!$str)
+		if (!static::isStringable($str))
 		{
+			@trigger_error('Passing a non-stringable argument to StringUtil::convertEncoding() has been deprecated an will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+
 			return '';
 		}
 
+		$str = static::toString($str);
```
leofeyer:
```suggestion
		$str = (string) $str;
```

We already know that `$str` is stringable at this point, therefore we do not need a private method for the cast.

core-bundle/tests/Contao/StringUtilTest.php
```diff
@@ -597,4 +598,118 @@ public function trimsplitProvider(): \Generator
...
+        ];
+
+        yield 'String "0"' => [
+            '0',
+            'UTF-8',
+            '0',
+        ];
+
+        yield 'Stringable argument' => [
+            new Stringable('foobar'),
```
leofeyer: Can we use an anonymous class here instead?

leofeyer (changes requested):
I still think having two private methods to encapsulate code that is only used once is over-engineered. Especially since we do not even need the second method at all.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
SeverinGloeckle: The idea was to provide the utility methods `toString` and `isStringable` for other use cases. However, as they are private now and only used in `convertEncoding`, they might be just inlined.

The  `isStringable` check should be inlined as well then (and may get extracted to a utility method in the future if any other method requires a similar check).

I will adjust the code accordingly.

leofeyer (approved):
Looks good. Thank you for your patience. 😄

leofeyer:
Thank you @SeverinGloeckle.

PR #495, opened 2019-05-19
Improve undo section

bezin:
Hello there, 

I worked on the new restore functionality during the camp and discussed some options with fellow friends. I would like to discuss the general approach with this PR. 

**What it does already (some of it core-only)**
* Display an editor friendly description rather than a SQL delete statement
* Display which user deleted something, 
* Let users filter by user (all deletions by john.doe) or source table (all deleted articles)
* It supports defining fields and a format in $GLOBALS['TL_DCA'][$table]['list']['restore'] like in the label section
* It actually gets away without defining anything quite often (in the core-bundle) rather than falling back to a couple of generic fields used often (title, headline, name, etc.)
* It is event based and extendible

**What is still missing**
* A way to determine descriptors for content elements (and other DataContainers whose content changes depending of a field like `type`)
*A way to include a parents' descriptor (e.g. to display `field name (form name)`. The screen shot shows some rows like this, but this was tested with another implementation (with old callbacks ;-)). This still needs to be figured out...

It works quite well for tables with either one of the defined generic fields or by manually defining one or more fields:
https://github.com/contao/contao/pull/495/commits/8b50295e51f10e4149bb7ab5527d7a28bb6bf835#diff-e87e5cf73a979d4358a51845475500eb

One can also implement another listener and implement a custom way to determine a tables' descriptor:
https://github.com/contao/contao/pull/495/commits/8b50295e51f10e4149bb7ab5527d7a28bb6bf835#diff-d7aee13190f8689ab803a06f48049e3e

However this would still lead to a messy listener for tl_content with many methods and a giant switch statement or alike or A LOT OF LISTENERS, for each individual content element:
https://github.com/contao/contao/pull/495/commits/29c0be2aefab73bc6c3c42ceab690388109b8985

What I like so far is that some decent default fields already yield pleasant generic results for extensions without the need for developers to configure something. Adding a new listener to handle special cases in ones' extension ensures extensibility, however I don't know how many developers would actually go all the way for such a tiny yet important feature for editors... 🤔

I appreciate any comments and ideas for a better architecture. Feel free to comment on implementation and naming $things, but that should not be the main discussion :-)

Cheerio 

2019-05-19 19_30_26-Restore _ contao test

PR #3498, opened 2021-09-24
Improve the undo module for better editor experience

bezin:
Hello folks, 

after much tinkering I finally gave up the approach we followed in #495 and followed another lead, that was easier to implement and actually even better interface-wise. Here is what this PR implements at the moment:

* Editors get a decent preview of the content they or somebody else deleted, rather than an ugly SQL query
* Editors may filter by `user` or `type` aka source table
* The list renders the deleted dataset as it was rendered in their respective backend modules list view. 
This means you get the full content element previewed
* It hooks into the existing DCA list label config (`fields`, `format`, `label_callback`)

As a side effect: 
* Outputs the table name as CSS class on the `tl_listing_container` which was necessary to style everything as needed. I wished I had this plenty times already in other list views. 

The implementation of the `renderLabel` method is copied from the Picker widget by @aschempp with minor adjustments. Thanks, Andy :) 

I would have prefered the `restore` and `show` icons and the undo meta data (who, what, when) in a single line, but this is not possible with the two column table the DC_Table driver generates 🤷 So the interface is a compromise of what is possible with the given structure. 

What's left to do?

- [x]  Translate more table names (i. e. `tl_content` = `Content element`, `tl_article` = `Article`)
- [x]  Implement a function to safely use table names as CSS classes.
- [x]  Test this implementation against some more extensions
- [x]  Add a proper icon for jump-to-parent functionality
- [] Update documentation
~- Refactor the `renderLabel` method and the Picker widget to use the same function?~

Happy to hear your comments :-)

Cheerio


bezin:
Feedback in Slack mentioned the lack of context for the deleted entries hence I now output the parent type, too. It would be better if we also provide a name or title of the parent dataset, too, but then we have the same issue as with the original proposal: the lack of a descriptor. 
 
Two other suggestions were:
* Provide an filter option for parents: I deleted an custom element of article XYZ, but I don't remember what the content of this element was > Filter by article XYZ to only show undo entries related to this article
* Add a jump To Parent functionality: e.g. as a third button, that takes you to the content elements parent news entry or article. I like this one, but don't know how to determine where to jump to in a universal way that works with every DCA. > **This is implemented now**

Would be happy to discuss these and the whole PR in the upcoming Mumble @leofeyer

aschempp:
> Add a jump To Parent functionality: e.g. as a third button, that takes you to the content elements parent news entry or article. I like this one, but don't know how to determine where to jump to in a universal way that works with every DCA.

I really like that idea as well. It should be possible to determine the parent view/module by searching the (parent!) table in `$GLOBALS['BE_MOD']`.

bezin:
Thanks @aschempp I thought about this approach but did not think it will work. However, it does and I implemented it. There is now an additional button to jump to the parent entry. The parent in the header entry is also clickable. I used the `forward` icon for now, but another one should be added. 



I also added the last missing translations.

bezin:
What do you think about extracting the method that renders the label so the picker widget and the undo view use the same method? 

~I would also need some input on my table name class, that I added to `tl_listing_container`, e.g. `.tl_listing_container.tl_undo` in the undo view. Is this safe to do or should we put in some safeguards? Drupal has a [clean_css_identfier](https://github.com/drupal/drupal/blob/515d10367bbe5cc158153a90e7960f92c2862745/core/lib/Drupal/Component/Utility/Html.php#L95-L131) method, which would be interesting.~

Better safe and sorry, table name classes are standardized and prefixed now.

aschempp:
> What do you think about extracting the method that renders the label so the picker widget and the undo view use the same method?

That sounds useful, but I have no idea where that should go … and it should probably be used in the DC driver as well? I suggest to create a separate ticket for that to keep this one focused.

bezin:
Yes, I agree – it would make most sense, if DC_Table uses the new method as well. For now I leave it, where it is. I made some minor adjustments to your implementation for my case anyways.

bezin:
Hello all,

I updated this PR according to what we discussed in the call last week. 



1. If an entry has no parent, the jump to parent icon is present, but disabled. 
2. The jump to parent link hidden in the label header is removed because it was not very Contao-like and barely recognizable anyway. 
3. I use the new DataContainer constants 😎 

I have some more ideas concerning the handling of parent-child relations, but I would rather open a new PR for that after this is merged. I await @frontendschlampe's contribution for a jump to parent icon now to finish here :-)

bezin:
Hi folks, 

here we have three alternatives for the jump to parent icon by @frontendschlampe and me. What do you think?

### Variant 1
undo_v1

### Variant 2
undo_v2

### Variant 3
undo_v3

Let's have a vote!

**Variant 1** React with 🎉 
**Variant 2** React with 🚀 
**Variant 3** React with ❤️ 
**None of the above** React with 👎

leofeyer:
Since the link takes you to the overview page where there are multiple children, I find  more suitable.

frontendschlampe:
> Since the link takes you to the overview page where there are multiple children, I find  more suitable.

This was our first idea, but we think, that it's too detailed, that's why we try Variant 1 ;-)

aschempp:
Also, the icon is referring to the parent **of the deleted record**, not multiple deleted records.

bezin:
> Also, the icon is referring to the parent **of the deleted record**, not multiple deleted records.

Yes! It really is about the relation of the single record you (or somebody else) deleted. That this entry may have siblings is IMO not relevant here.

bezin:
For those reacting with 👎 You are encouraged to show us your take on an icon that suits the task better than the provided alternatives  :-)

Ernestopheles:
As far as I can see, the situation shows up like this:

An element has been deleted and can be restored. In the restore panel a jump to the parent of this element shall be made possible.

If so, I disagree witrh the coloring:
Red rather means deleted or to be deleted, so the element should better be colored red and not its parent. I suggest to color the parent green and the element red. 

This of course is not so well suiteable to choice number 2, as of positioning of the red child becomes difficult.

Therefore I suggest to use this coloring with choice number 1 or 3.

bezin:
> As far as I can see, the situation shows up like this: An element has been deleted and can be restored. In the restore panel a jump to the parent of this element shall be made possible.

Correct!

> If so, I disagree witrh the coloring: Red rather means deleted or to be deleted, so the element should better be colored red and not its parent. I suggest to color the parent green and the element red.

I partially agree and disagree: The red rectangle might indeed be misleading. I would prefer blue than, because we use blue for information already. I disagree with coloring the record itself, especially green: First it would give it to much emphasis (its all about the parent) and second we should not mix red and green for accessibility.

Ernestopheles:
> I partially agree and disagree: The red rectangle might indeed be misleading. I would prefer blue than, because we use blue for information already.<
I agree to this

 >I disagree with coloring the record itself, especially green: First it would give it to much emphasis (its all about the parent) and second we should not mix red and green for accessibility.<
My suggestion was to color the ement red only in the icon. red and blue will go for accessibility?

Maybe instead of coloting the child in the icon might be a red border

frontendschlampe:
I agree with @bezin … the parent is more important, that's why it should have the main color. I think blue would be the best, too. The children are not important, that's why I would let them be gray.

frontendschlampe:
> Maybe instead of coloting the child in the icon might be a red border

I think, in this size, it wouldn't be visible enough.

Ernestopheles:
I think grey is well

bezin:
The vote was quite unambiguous 😂 @frontendschlampe updated the icon and added an proper inactive vesion. This is what it looks right now:

bezin:
IMO this is RTM

core-bundle/src/EventListener/DataContainer/UndoListener.php
```diff
@@ -0,0 +1,347 @@
...
+        $controller = $framework->getAdapter(Controller::class);
+        $this->controller = $controller;
+
+        /** @var Backend $backend */
+        $backend = $framework->getAdapter(Backend::class);
+        $this->backend = $backend;
+
+        /** @var Image $image */
+        $image = $this->framework->getAdapter(Image::class);
+        $this->image = $image;
```
m-vo: The framework will need to be initialized at this point. It's probably better to move these calls out of the constructor + call `$framework->initialize()` before the first use.

core-bundle/src/EventListener/DataContainer/UndoListener.php
```diff
@@ -0,0 +1,347 @@
...
+
+        $params = $this->getParentLinkParameters($parent, $table);
+
+        return sprintf(
+            '<a href="%s" title="%s" onclick="Backend.openModalIframe({\'title\':\'%s\',\'url\': this.href });return false">%s</a> ',
+            $this->backend->addToUrl($params),
+            StringUtil::specialchars($title),
+            $title,
+            $this->image->getHtml($icon, $label)
+        );
```
m-vo: Two remarks that do not specifically apply to this method:

1.  If you move variable declarations down to where they are needed (basically as far down as possible), they often do not need to be computed in the first place if an early return or guard is hit. And if they are used only once, you can easily inline them.

2. Please make sure not to overwrite arguments with same name (`$title` in this case). If you're using PHPStorm the "PHP Inspections (EA Extended)" plugin will easily help you find those things. The inspection is called "Suspicious assignments" and is enabled by default.

With this in mind, here is how this could get refactored:

```suggestion
        $table = $row['fromTable'];
        $originalRow = StringUtil::deserialize($row['data'])[$table][0];
        $parent = $this->getParentTableForRow($table, $originalRow);

        if (!$parent || !$this->checkIfParentExists($parent)) {
            return $this->image->getHtml('parent_.svg', $label).' ';
        }

        $newTitle = sprintf(
            $GLOBALS['TL_LANG']['tl_undo']['parent_modal'], 
            $this->getTypeFromTable($table), 
            $originalRow['id']
        );

        return sprintf(
            '<a href="%s" title="%s" onclick="Backend.openModalIframe({\'title\':\'%s\',\'url\': this.href });return false">%s</a> ',
            $this->backend->addToUrl($this->getParentLinkParameters($parent, $table)),
            StringUtil::specialchars($newTitle),
            $newTitle,
            $this->image->getHtml($icon, $label)
        );
```

core-bundle/src/EventListener/DataContainer/UndoListener.php
```diff
@@ -0,0 +1,347 @@
...
+    {
+        $this->controller->loadLanguageFile($table);
+
+        return isset($GLOBALS['TL_LANG'][$table]['_table']) ? $GLOBALS['TL_LANG'][$table]['_table'][0] : $table;
+    }
+
+    private function checkIfParentExists(array $parent): bool
+    {
+        $count = $this->connection->fetchOne(
+            'SELECT COUNT(*) FROM '.$parent['table'].' WHERE id = :id',
```
m-vo:
```suggestion
            "SELECT COUNT(*) FROM {$this->connection->quoteIdentifier($parent['table'])} WHERE id = :id",
```

bytehead (commented):
Why do the tests still fail?

core-bundle/src/Resources/config/listener.yml
```diff
@@ -450,3 +450,8 @@ services:
             package: contao/core-bundle
             version: 4.13
             message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "contao.listener.data_container.page_url" instead.
+
+    Contao\CoreBundle\EventListener\DataContainer\UndoListener:
```
m-vo: This should now be `contao.listener.data_container.undo`, see #3686.

core-bundle/src/Resources/contao/dca/tl_undo.php
```diff
@@ -217,16 +225,23 @@ public function showDeletedRecords($data, $row)
...
-	 * @param array  $row
-	 * @param string $label
-	 *
-	 * @return string
-	 */
-	public function ellipsis($row, $label)
+	public function getFromTableOptions(DataContainer $dc)
 	{
-		return '<div class="ellipsis">' . $label . '</div>';
+		$tables = $this->Database->execute('SELECT DISTINCT ' . Database::quoteIdentifier('fromTable') . ' FROM tl_undo');
```
m-vo: Any reason why you're using the old `Database` class here instead of `System::getContainer()->get('connection')`?

m-vo (changes requested):
Thanks a lot for your work! I did not go through everything, here are just some first remarks looking at the code.

bezin:
> Why do the tests still fail?

Because I forgot to update them 🙄 

@m-vo thanks for your comments, will look into them after the weekend 👍

core-bundle/src/EventListener/DataContainer/UndoListener.php
bezin: I see your point; I chose this approach as it is used in other DataContainer listeners, too, e.g. `TemplateOptionsListener`, `ContentCompositionListener`. Should probably be updated there as well.

core-bundle/src/Resources/contao/dca/tl_undo.php
bezin: Nope, except that I just went with the legacy code flow ;-)

bezin:
@m-vo Thanks for your comments, I hope I addressed them all appropriately.

With #3633 in mind, I would like to create a sub-namespace `Undo` and split up everything in separate listeners: 

* `LabelListener` for the rendering of the label
* `JumpToParentOperationListener` for the jumpToParent operation
* `UndoOperationListener` for restoring parent records first (not part of this PR but #3633 )

I will introduce a Trait for the parent record related logic. 

What do you think?

bezin:
I implemented the listeners separately in https://github.com/contao/contao/pull/3498/commits/2f8a4cdaba576237bbcd73fca5f9933fba282b93 I think this is better structured and improves readability

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,229 @@
...
+        $newLabel = $this->renderLabel($originalRow, $fromTableDc);
+
+        if ($GLOBALS['TL_DCA'][$table]['list']['label']['showColumns'] ?? false) {
+            $newLabel = $this->renderColumns($newLabel);
+        }
+
+        return $header.'<div class="tl_undo_preview">'.$newLabel.'</div>';
+    }
+
+    private function renderHeader(string $table, array $row, array $originalRow): string
```
m-vo: I don't like rendering HTML in PHP. I know it's done in the DC files but we should IMHO not introduce more code with this pattern.

All of this can be done in templates instead. There you already have all the tools for date formatting, translations and encoding (which is missing here as well). Plus: everyone can easily adjust it to their needs.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,229 @@
...
+                [$strKey, $strTable] = explode(':', $v);
+                [$strTable, $strField] = explode('.', $strTable);
+
+                $objRef = $this->connection
+                    ->prepare('SELECT '.$strField.' FROM '.$strTable.' WHERE id=? LIMIT 1')
+                    ->executeQuery($arrRow[$strKey])
+                ;
+
+                $labelValues[$k] = $objRef->rowCount() ? $objRef->{$strField} : '';
+            } elseif (\in_array($GLOBALS['TL_DCA'][$dc->table]['fields'][$v]['flag'], [DataContainer::SORT_DAY_ASC, DataContainer::SORT_DAY_DESC, DataContainer::SORT_MONTH_ASC, DataContainer::SORT_MONTH_DESC, DataContainer::SORT_YEAR_ASC, DataContainer::SORT_YEAR_DESC], true)) {
```
m-vo: Please use early returns in favor of `elseif`/`else` clauses for better readability.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6331,6 +6331,11 @@ public function initPicker(PickerInterface $picker)
 
 		return $attributes;
 	}
+
+	private function addTableClass(string $table)
```
Toflar:
```suggestion
	private function getTableClass(string $table)
```

It doesn't really add anything, does it? ;)

core-bundle/src/EventListener/DataContainer/Undo/UndoListenerTrait.php
```diff
@@ -0,0 +1,60 @@
...
+        return null;
+    }
+
+    private function getTranslatedTypeFromTable(string $table): string
+    {
+        /** @var Controller $controller */
+        $controller = $this->framework->getAdapter(Controller::class);
+        $controller->loadLanguageFile($table);
+
+        return isset($GLOBALS['TL_LANG'][$table]['_type']) ? $GLOBALS['TL_LANG'][$table]['_type'][0] : $table;
```
aschempp: `['_type'][1]` is defined in the language files but seems unused? I have two thoughts about that

1. this basically is a label for each table, right? I wonder if we don't have that already, if we really need it and if we could use that in other places. How do we for example generate the breadcrumb?
2. By having these labels in the default language file (e.g. `$GLOBALS['TL_LANG']['MODEL'][...]`) we could prevent the system from needing to load all table translation files just for this.

SeverinGloeckle: Sadly, there are no translations for the table names yet.

I agree that these would be useful in multiple places! We often customize the version overview on the backend dashboard to show translated table names instead of `tl_content`, `tl_module` etc. to the editor.

The translations should go into `modules.xlf` or alternatively into `default.xlf`, possibly as something like `tl_lang.table.[table]` ( e.g. `tl_lang.table.tl_content = 'Content element'`).

Since I assume there'll be some discussion about the correct names for some tables, maybe these should get their own pull request to keep the focus. What do you think @bezin?

If we have these labels, I'll happily provide another pull request to use them in the version overview.

bezin: Hello all, we discussed this in a Mumble call two and a half years ago :D We decided we want to introduce singular and plural translations for tables/models/types which is what I implemented here. 

I agree that it might be an issue when we put it into each individual DCA translation file. `modules.xlf` is IMHO the wrong place (a module, back-end or front-end can deal with many models),, but `default.xlf` sounds suitable.

@aschempp You are right, they do exist for some tables, but not in a universal way. Have a look at the form or comments backend module in the demo: If you edit a form, the German breadcrumb is `Formulare > Felder`, but I would need `Formular` and `Feld`. In the comments module, you only have `Kommentare`, but no singular either. Also, `tl_calendar` is translated with `Events` as is `tl_calendar_events`.

That said, I still think it is a good idea and we should introduce these in a universal way.

I think, for the core elements the translation is a no brainer, as we use singular and plural labels in many places; we simply lack these labels as separately translated strings. As one of my goals with this PR was **to make the undo module less technically**, I would not like to merge this without translated table names (for the core types), but if this causes discussion I am happy to provide a separate PR to merge the translations first.

SeverinGloeckle: Thanks for the background information! 

I hadn't thought of the plural forms - they would also come in handy in many places. Nice!

If the translations are collected in a single file, they could probably also be discussed here if necessary.

bezin: I moved all translations to `contao_default.TABLES` Documentation needs an udpate here then: https://docs.contao.org/dev/framework/translations/#categories

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: I agree this is ugly legacy style in a listener (I copied it from the Picker widget class), but before I start refactoring this, I wait for #3644 and its `generateRecordLabel ` method to settle, because that already would handle some of this. 

Also I wonder if [Haste\ModelData](https://github.com/codefog/contao-haste/blob/main/library/Haste/Generator/ModelData.php#L35-L90) is not something we might add to the Core bundle? Right now, every feature/implementation has to carefully handle a DCA field on its own – decoding, encoding, formatting… However this would be something for another PR ;-)

calendar-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -65,6 +65,24 @@
       <trans-unit id="MSC.eventPicker">
         <source>Events</source>
       </trans-unit>
+      <trans-unit id="TABLES.tl_calendar.0">
+        <source>Calendar</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_calendar.1">
+        <source>Calendar</source>
```
SeverinGloeckle:
```suggestion
        <source>Calendars</source>
```

news-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -35,6 +35,24 @@
       <trans-unit id="MSC.newsPicker">
         <source>News</source>
       </trans-unit>
+      <trans-unit id="TABLES.tl_news.0">
+        <source>News</source>
```
SeverinGloeckle:
```suggestion
        <source>News item</source>
```

news-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -35,6 +35,24 @@
...
+      </trans-unit>
+      <trans-unit id="TABLES.tl_news.1">
+        <source>News</source>
```
SeverinGloeckle:
```suggestion
        <source>News items</source>
```

calendar-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -65,6 +65,24 @@
...
+        <source>Event</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_calendar_events.1">
+        <source>Events</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_calendar_feed.0">
+        <source>Calendar feed</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_calendar_feed.1">
+        <source>Calendar feed</source>
```
SeverinGloeckle:
```suggestion
        <source>Calendar feeds</source>
```

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -2210,6 +2213,126 @@
...
+        <source>Form field</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_form_field.1">
+        <source>Form fields</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_image_size.0">
+        <source>Image size</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_image_size.1">
+        <source>Image size</source>
```
SeverinGloeckle:
```suggestion
        <source>Image sizes</source>
```

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -2210,6 +2213,126 @@
...
+        <source>Layout</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_layout.1">
+        <source>Layouts</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_log.0">
+        <source>Log entry</source>
+      </trans-unit>
+      <trans-unit id="TABLES.tl_log.1">
+        <source>Log entrys</source>
```
SeverinGloeckle:
```suggestion
        <source>Log entries</source>
```

faq-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -8,6 +8,18 @@
       <trans-unit id="MSC.faqPicker">
         <source>FAQ</source>
       </trans-unit>
+      <trans-unit id="TABLES.tl_faq.0">
+        <source>FAQ</source>
```
SeverinGloeckle:
```suggestion
        <source>FAQ item</source>
```

faq-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -8,6 +8,18 @@
...
+      </trans-unit>
+      <trans-unit id="TABLES.tl_faq.1">
+        <source>FAQs</source>
```
SeverinGloeckle:
```suggestion
        <source>FAQ items</source>
```

SeverinGloeckle (commented):
I have just added some missing pluralizations and would suggest some more explicit translations for terms like news and FAQ.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: I rewrote this part to avoid elseif / else. Still think, there is room for utilities to help with formatting DCA data, but this should be done separately.

core-bundle/src/EventListener/DataContainer/Undo/UndoListenerTrait.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+trait UndoListenerTrait
+{
+    private Connection $connection;
+    private ContaoFramework $framework;
+    private TranslatorInterface $translator;
+
+    private function getParentTableForRow(string $table, array $row): ?array
+    {
+        if (isset($GLOBALS['TL_DCA'][$table]['config']['dynamicPtable']) && true === $GLOBALS['TL_DCA'][$table]['config']['dynamicPtable']) {
```
leofeyer:
```suggestion
        if (true === ($GLOBALS['TL_DCA'][$table]['config']['dynamicPtable'] ?? null)) {
```

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
```diff
@@ -0,0 +1,41 @@
...
+    <div class="tl_undo_header_item">
+        <span class="tl_undo_header_label">{{ 'tl_undo.pid.0'|trans({}, 'contao_tl_undo') }}</span>
+        <strong>{% if user %}{{ user.username }}{% else %}{{ row.pid }}{% endif %}</strong>
+    </div>
+    <div class="tl_undo_header_item">
+        <span class="tl_undo_header_label">{{ 'tl_undo.fromTable.0'|trans({}, 'contao_tl_undo') }}</span>
+        <strong>{{ translatedFromTable }}</strong>
+    </div>
+    <div class="tl_undo_header_item">
+        <span class="tl_undo_header_label">ID</span> {{ originalRow.id }}
```
leofeyer: `{{ originalRow.id }}` is the only string that is now wrapped in a `<span>` or `<strong>` tag. Intended?

leofeyer (changes requested):
I like this PR a lot, however, there are a few things that bother me:



The "Parent" column is always the same as the "Type" column. I assume there are some cases in which they differ, however, in my case the column did not provide any valuable information. Can we just omit it?

I also don't like that the link to the parent element opens in a modal window.

And what bothers me most is the large amount of new labels, which partly duplicate existing labels from the `MOD` namespace. That implies a lot of redundant work for our translators. Can we maybe show the table names instead?

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
bezin: Yes, it is not necessary. No styling needed ATM.

bezin:
Thanks for your comments!

> The "Parent" column is always the same as the "Type" column.

This is not true and I wonder what has happened in your Screenshot. I will look into this. It should just be shown if a recrod has a parent and then it should display the parent tables name. One could think about if we only show it in case of dynamic parent tables? Maybe it is not important to show that a `form field` has a parent `form` but I like that it tells me that a `content element` is from a `news item` rather than an `article`

> I also don't like that the link to the parent element opens in a modal window.

My thoughts were: I as an editor do not want to be carried away from my actual task (i. e. "restore a row") just because I want to verify, where this record originated from. To me this is like the info modal, that shows me more information in detail. A modal is not the best solution UI-wise, but felt like an established pattern in the Contao back-end and is hence a compromise. What would you suggest?

> large amount of new labels, which partly duplicate existing labels from the MOD namespace.

I disagree. The MOD namespace translates back-end modules. The TABLES namespace contains singular and plural translations for the actual entities or content that editors have to work with. My main goal was to remove anything technical from the restore view as it is for editors. They should not be required to know about implementation details – what are tables to an editor anyways? Table names are also somewhat cryptic (`tl_` what?) and most of the time in English. 

I added these translations many times myself already and would love to see these provided for all Core tables / models. It is work, yes, but it improves the UI.

bezin:
> The "Parent" column is always the same as the "Type" column. I assume there are some cases in which they differ, however, in my case the column did not provide any valuable information. Can we just omit it?

I fixed this. 

As for the deprecation hint: I cannot reproduce this.

bezin:
> One could think about if we only show it in case of dynamic parent tables? Maybe it is not important to show that a `form field` has a parent `form` but I like that it tells me that a `content element` is from a `news item` rather than an `article`

The more I think about it, the better I like it. It would remove a little clutter from the UI and would only be displayed, if it actually contributes to clarity. WDYT?

bezin:
@leofeyer `up-for-discussion` later?

bezin:
I removed the genereal parent label in the row header and only show it in case the parent table is dynamic

(Shown for the first row, because it is a content element that has a dynamic parent table. Not shown for the event, because its parent is always tl_calendar)

bezin:
@leofeyer As the 31st is coming closer, how do we go from here? If only the additional labels are the issue I am happy to update this PR to remove these, for the sake of merging this to 4.13. Maybe we can treat the missing labels als a bug, so we may fix this for 4.13 later ;-)

leofeyer:
No worries, your PR does not have to be merged by December 31. It just needs to be ready to be merged, which means

1. the PR must no longer be a draft;
2. all features must be implemented;
3. all unit tests must be in place;
4. the CI checks must not fail (except maybe codecov/project).

Reviewing and merging will be done until around January 15, so we will have enough time to discuss the additional labels. Maybe the other @contao/developers can give their opinion as well?

leofeyer:
I am still for using the table names. It would save us A LOT of translation work.



@contao/developers WDYT?

leofeyer:
I just noticed that there is no unit test for the `LabelListener` class. @bezin Did you deliberately not add that?

m-vo:
> I am still for using the table names. It would save us A LOT of translation work.

:+1: We can still improve this at some point and also include other places (like the allowed fields config).

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
```diff
@@ -0,0 +1,115 @@
...
+        $controller = $this->framework->getAdapter(Controller::class);
+        $controller->loadDataContainer($parent['table']);
+
+        $module = $this->getModuleForTable($parent['table']);
+
+        if (!$module) {
+            return $params;
+        }
+
+        $params = 'do='.$module['_module_name'];
```
m-vo: Can we use `http_build_query` here, so that we do not need to care about encoding and the like?

core-bundle/src/Resources/contao/dca/tl_undo.php
```diff
@@ -95,6 +98,8 @@
 		'fromTable' => array
 		(
 			'sorting'                 => true,
+			'filter'                  => true,
+			'options_callback'        => array('tl_undo', 'getFromTableOptions'),
```
m-vo: Why aren't we using an event listener (with nice constructor injection and tests) here as well?

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,195 @@
...
+
+        if (true === ($GLOBALS['TL_DCA'][$table]['config']['dynamicPtable'] ?? null)) {
+            $parent = ['table' => $originalRow['ptable'], 'id' => $originalRow['pid']];
+        }
+
+        return [
+            'preview' => $this->renderPreview($originalRow, $originalTableDc),
+            'user' => $user,
+            'row' => $row,
+            'translatedFromTable' => $this->getTranslatedTypeFromTable($table),
```
m-vo: Please apply translations in the template using `|trans`.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,195 @@
...
+            'parent' => $parent,
+            'translatedParentTable' => $parent ? $this->getTranslatedTypeFromTable($parent['table']) : null,
+            'originalRow' => $originalRow,
+            'dateFormat' => $config->get('dateFormat'),
+            'timeFormat' => $config->get('timeFormat'),
```
m-vo, line 84: (Sidenote: We should probably set this as Twig globals: https://symfony.com/doc/current/templating/global_variables.html instead of passing it into each template. I can do this in a followup PR, though.)

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            $labelValues[$k] = $arrRow[$v];
+        }
+
+        $label = $labelConfig['format'] ? vsprintf($labelConfig['format'], $labelValues) : implode(', ', $labelValues);
+
+        // Remove empty brackets (), [], {}, <> and empty tags from the label
+        $label = preg_replace('/\( *\) ?|\[ *] ?|{ *} ?|< *> ?/', '', $label);
+        $label = preg_replace('/<[^>]+>\s*<\/[^>]+>/', '', $label);
```
m-vo: What is this doing?

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,195 @@
...
+
+        if (\is_callable($labelConfig['label_callback'] ?? null)) {
+            if (\in_array($mode, [DataContainer::MODE_TREE, DataContainer::MODE_TREE_EXTENDED], true)) {
+                return $labelConfig['label_callback']($arrRow, $label, $dc, '', false, null);
+            }
+
+            return $labelConfig['label_callback']($arrRow, $label, $dc, $labelValues);
+        }
+
+        return $label ?: (string) $arrRow['id'];
```
m-vo: Can't we always return an array? Having to check the type isn't the best DX. :see_no_evil:

m-vo (changes requested):

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
fritzmg: See https://github.com/contao/contao/pull/3644 - it's part of the label creation in the DataContainer.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
fritzmg: Depending on the data container mode the labels are either an array (for showing multiple columns) or a single string. See also https://github.com/contao/contao/pull/3644

bezin: What @fritzmg said. Also, I need the raw column values to render my own table in the preview. I could wrap everything in a function that always returns a string, but then I would render HTML (the table) in the class itself again.

core-bundle/src/Resources/contao/dca/tl_undo.php
bezin: I did not do this here for the same reason I did not do this in the [feed image size pr](https://github.com/contao/contao/pull/3630#pullrequestreview-824357882) I find a select field without clearly defined options / options callback anoying. If however this is the way the core wants to go, I will update this of course.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: AFAIK this is not possible because I cannot define a fallback in the template. If a table is not translated (e.g. an extension is not yet updated), I want to display the table name as fallback, but the Twig filter would use the translation key as fallback (e.g. `TABLES.tl_content.0` rather than `tl_content`)

fritzmg: We probably want to get rid of the translation anyway 🙃

core-bundle/src/Resources/contao/dca/tl_undo.php
Toflar: Proper listeners are the way to go. Always :)

bezin: Alright, updated it https://github.com/contao/contao/pull/3498/commits/1278ffc56725e70fcfd94524344f7ab144d9398e

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
m-vo: > AFAIK this is not possible

Why? You could use the same fall back in the template as you currently did in `getTranslatedTypeFromTable()`. But getting rid of them would of course solve this as well. :smile:

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
m-vo: If we need to duplicate this magic, should we rather put it into `StringUtil` and make it available for anyone?

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,195 @@
...
+        $labelConfig = &$GLOBALS['TL_DCA'][$dc->table]['list']['label'];
+        $labelValues = [];
+
+        foreach ($labelConfig['fields'] as $k => $v) {
+            if (false !== strpos($v, ':')) {
+                [$strKey, $strTable] = explode(':', $v);
+                [$strTable, $strField] = explode('.', $strTable);
+
+                $objRef = $this->connection
+                    ->prepare('SELECT '.$strField.' FROM '.$strTable.' WHERE id=? LIMIT 1')
```
m-vo: Identifiers should be quoted.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
fritzmg: If #3644 gets merged `DataContainer::generateRecordLabel` can be used in this PR as well and nothing needs to be duplicated.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: I removed the translations.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
m-vo: If I get this correctly, the Twig template simply renders all availabe values for `preview` side by side. Why can't `preview` always be an array, then? One element is fine as well. Or am I missing something?

bezin:
> I am still for using the table names. It would save us A LOT of translation work.

I removed the translations – with a heavy heart. For the future: Which languages must be provided in order to add new translations to the Core? English, German, French, Spanish?

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
```diff
@@ -0,0 +1,41 @@
...
+    {% if preview is not iterable %}
+        {{ preview|raw }}
+    {% else %}
+        <table style="width: 100%;">
+            <tr>
+            {% for value in preview %}
+                <td>{{ value|raw }}</td>
+            {% endfor %}
+            </tr>
+        </table>
```
m-vo: Should we use grid/flex here instead of a table?

m-vo (changes requested):

fritzmg:
> Which languages must be provided in order to add new translations to the Core? English, German, French, Spanish?

Only English. The translations are then managed on [Transifex](https://www.transifex.com/contao/core-bundle/) by each language's translators.

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
aschempp: please don't. The backend uses tables everywhere, so this is consistent.

bezin:
> I just noticed that there is no unit test for the `LabelListener` class. @bezin Did you deliberately not add that?

I did not add one on purpose, but I added a basic one now. What to test depends on #3644 being merged, as I could reuse this to simplify my listener by a lot

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
bezin: The table is used here a) because tables are used every and b) this markup is used, whenever a data container uses `showColumns = true`, so it actually is tabular data (even though just one row). 

I use my own table markup here to not have specific CSS classes, that somebody might have set or used to style the table in the DC list view, e.g. in the member module, each table cell has a class like `col_firstname`, `col_lastname` and so on. There might be custom ones in custom data containers, too. This could potentially break the Undo view, e.g. you set custom widths for table cells.

leofeyer:
I see. I'll try to get #3644 merged before this one then.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: The difference here is, that with an array it is tabular data (member list, user list, ...). When it is only a string, it is a label like article, page, content element etc. I could always return an array and use `count()`. But let's first wait for #3644 to be implemented as this makes refactoring necessary anyways.

leofeyer:
#3644 has been merged. 🎉

leofeyer:
#3644 has been merged (this time for real). 😄

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
bezin: I now use the  `generateRecordLabel` method 👍

core-bundle/src/Resources/contao/themes/flexible/main.css
```diff
@@ -2139,6 +2139,73 @@ fieldset.collapsed > legend {
...
+	border-left-color: #fffce1;
+}
+.listing__tl_undo.list_view .odd .limit_toggler {
+	background-color: #fafafc;
+}
+.tl_undo_header .tl_undo_header_item:not(:first-child) {
+	margin-left: 32px;
+}
+.tl_undo_header_item.align-right {
+	align-self: right;
```
leofeyer: This definition seems to be wrong:

core-bundle/src/Resources/contao/languages/en/tl_undo.xlf
```diff
@@ -36,7 +36,7 @@
         <source>Restore element ID %s</source>
       </trans-unit>
       <trans-unit id="tl_undo.parent_modal">
-        <source>Show parent of %s ID %s</source>
+        <source>Show parent element of %s ID %s</source>
```
bezin: Shouldn't this be "parent **record**"?

leofeyer (changes requested):

core-bundle/src/Resources/contao/themes/flexible/main.css
bezin: Seems to be a left-over. I removed it in 65e8a177dfae88315905dbc83adbee5cfb396961

core-bundle/src/Resources/contao/languages/en/tl_undo.xlf
leofeyer: To me, "show parent record" is this:



So in my opinion, the link should say "Go to the parent […]" instead of "Show the parent […]". Im not sure what to use for […], though.

leofeyer: To me, "show parent record" is this:



So in my opinion, the link should say "Go to the parent […]" instead of "Show the parent […]". Im not sure what to use for […], though.

bezin: I see your point with `go` vs `show`. Maybe we should use `origin` rather than parent? In the end we show the view where the deleted record would be shown if it had not been removed. 
So `Go to origin of %s ID %s`?

leofeyer:
I have simplified the CSS in 54a5d232804d3c038a384817373cc79fef32297d. I hope you don't mind.

**Before**



**After**



**Before mobile**



**After mobile**



@bezin WDYT?

core-bundle/src/Resources/contao/languages/en/tl_undo.xlf
leofeyer: That's good. 👍

core-bundle/src/EventListener/DataContainer/Undo/FromTableOptionsListener.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        $options = [];
+
+        foreach ($tables->fetchFirstColumn() as $table) {
+            $options[] = $table;
+        }
+
+        return $options;
+    }
```
m-vo: Could this be shortened? :upside_down_face: 
```suggestion
        return $this->connection
            ->executeQuery('SELECT DISTINCT '.$this->connection->quoteIdentifier('fromTable').' FROM tl_undo')
            ->fetchFirstColumn()
        ;
```

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
```diff
@@ -0,0 +1,116 @@
...
+            StringUtil::specialchars($newTitle),
+            StringUtil::specialchars($newTitle),
+            $image->getHtml($icon, $label)
+        );
+    }
+
+    private function getParentLinkParameters(array $parent, string $table): string
+    {
+        $params = [];
+
```
m-vo:
```suggestion
```

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
```diff
@@ -0,0 +1,116 @@
...
+        $controller = $this->framework->getAdapter(Controller::class);
+        $controller->loadDataContainer($parent['table']);
+
+        $module = $this->getModuleForTable($parent['table']);
+
+        if (!$module) {
+            return '';
+        }
+
+        $params['do'] = $module['_module_name'];
```
m-vo:
```suggestion
        $params = ['do' => $module['_module_name']];
```

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
```diff
@@ -0,0 +1,116 @@
...
+
+    private function getModuleForTable(string $table): array
+    {
+        $module = null;
+
+        foreach ($GLOBALS['BE_MOD'] as $group) {
+            foreach ($group as $name => $config) {
+                if (\is_array($config['tables'] ?? null) && \in_array($table, $config['tables'], true)) {
+                    $module = $config;
+                    $module['_module_name'] = $name;
```
m-vo: Should there be a `break` or an early return? Module gets overwritten in each iteration.

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -0,0 +1,107 @@
...
+            if (\is_array($callback)) {
+                return System::importStatic($callback[0])->{$callback[1]}($row);
+            }
+
+            if (\is_callable($callback)) {
+                return $callback($row);
+            }
+        }
+
+        return $dc->generateRecordLabel($row, $dc->table);
```
m-vo: The diff looks ugly, but I'd basically switch around logic, so that there is only 1 level of indentation.
```suggestion
        if (DataContainer::MODE_PARENT !== $GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['mode']) {
            return $dc->generateRecordLabel($row, $dc->table);
        }
        
        $callback = $GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback'] ?? null;

        if (\is_array($callback)) {
            return System::importStatic($callback[0])->{$callback[1]}($row);
        }

        if (\is_callable($callback)) {
            return $callback($row);
        }
```

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
```diff
@@ -0,0 +1,35 @@
+{% trans_default_domain 'contao_default' %}
```
m-vo: Would't it make more sense to use `contao_tl_undo` as the default domain? :upside_down_face:

m-vo (changes requested):

core-bundle/src/EventListener/DataContainer/Undo/FromTableOptionsListener.php
leofeyer: While trying to answer your question, I noticed that  the listener is actually never used. 🤔 Since undo records cannot be edited, I wonder what the listener is used for?

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
leofeyer: Interesting catch. With `break` it would be "first match wins" and without it would be "last match wins".

m-vo: If it should be "last match", we could terminate early as well, we would just need to iterate backwards. :slightly_smiling_face:

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
leofeyer: See 9f4534f4e182fa59f6c16172dd8bf38b1273e90f

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
leofeyer: See 9f4534f4e182fa59f6c16172dd8bf38b1273e90f

core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
leofeyer: See 9f4534f4e182fa59f6c16172dd8bf38b1273e90f. I have also added `?? null` in the if condition.

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
leofeyer: See 9f4534f4e182fa59f6c16172dd8bf38b1273e90f

core-bundle/src/Resources/contao/languages/en/tl_undo.xlf
leofeyer: Added in 5ed440f7ddb5e81eef45f41119f5319359975b03.

core-bundle/src/EventListener/DataContainer/Undo/FromTableOptionsListener.php
bezin: I just realised this is obsolete now. It is from when we had translations for tables and these were spread out across multiple translation files. I used the `options_callback` to gather the translations and use them for the filter menu. 

Removed in fe4f685c9

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
bezin: Yes, good catch. I would say, "first match wins" is fine. But is having the same table in more than one backend module a thing?

bezin:
> I have simplified the CSS in 54a5d23. I hope you don't mind.

I don't mind simplifying the CSS (and the markup): I like the grid usage, because it is so much cleaner in the template.

I do mind the actual changes you made to the view, though:

IMO the visual hierarchy is now lost completely. I found it necessary to have a visual separation of the deleted record preview and the undo record with meta data, to make it more easy to skim through the list and find the right record. 

I do agree that the mobile view needed some more work though.

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
m-vo: There might be use cases, but I think I'd go for "first match wins" for now. We can still change that later on to showing all or a different one should the need arise.

leofeyer:
> I found it necessary to have a visual separation of the deleted record preview and the undo record with meta data, to make it more easy to skim through the list and find the right record.

How do you like this?

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
bezin: Fixed in baf9e75

leofeyer:
Or maybe we should just vertical align the buttons at the top? Then they would provide additional guidance on where a new record starts.

bezin:
> Or maybe we should just vertical align the buttons at the top? Then they would provide additional guidance on where a new record starts.

I was about to propose the same 😂 I went with `padding-top: 8px`, so it would not look too dense. 


However, I am also missing the separation between the meta data labels and values. Having a different shade of gray here helps with skimming through the list, too.


Last not least: The tables should not break here on a desktop computer at least for the core table names.

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
```diff
@@ -0,0 +1,115 @@
...
+                    break 2;
+                }
+            }
+        }
+
+        return $module;
```
m-vo:
```suggestion
        foreach ($GLOBALS['BE_MOD'] as $group) {
            foreach ($group as $name => $config) {
                if (\is_array($config['tables'] ?? null) && \in_array($table, $config['tables'], true)) {
                    return array_merge($config, ['_module_name' => $name]);
                }
            }
        }

        return null;
```

leofeyer:
I have fixed this in c1c274187c9083c6e9cf5ec18a99f107548f6732:



We could also not decrease the font size of the meta fields to put more emphasize on them. However, in my opinion this decreases the overall readability.

core-bundle/src/Resources/views/Backend/be_undo_label.html.twig
```diff
@@ -0,0 +1,35 @@
...
+    <div>
+        {{ 'tl_undo.pid.0'|trans }}: <strong>{% if user %}{{ user.username }}{% else %}{{ row.pid }}{% endif %}</strong>
+    </div>
+    <div>
+        {{ 'tl_undo.fromTable.0'|trans }}: <strong>{{ fromTable }}</strong>
+    </div>
+    <div>
+        ID: <strong>{{ originalRow.id }}</strong>
+    </div>
+    {% if (parent) %}
```
m-vo:
```suggestion
    {% if parent %}
```

m-vo (changes requested):
final nitpicks :see_no_evil:

leofeyer:
We could also increase the margin around the preview element:

core-bundle/src/EventListener/DataContainer/Undo/JumpToParentButtonListener.php
leofeyer: This is nice. Does it really work?

m-vo: IMHO this does exactly the same as before: if any module is found, return it (instead of break + return). The last row otherwise returns null every time as the inner condition wasn't hit.

bezin: Yip, works as needed, implemented in2a38b2fb1

m-vo (dismissed):

bezin:
I changed the label to `Show origin of ...` because it is also used as modal title and `go to` does not make sense there.

leofeyer (approved):

leofeyer:
Thank you @bezin for your work and your patience. 😉🙏

PR #3500, opened 2021-09-24
Move TL_MODE parameter default to method body

### LINKED ISSUE #3495: Impossible to mock classes extending `Contao\System` with Mockery due to TL_MODE parameter default
**Affected version(s)**

Contao 4.9 and up

**Description**
It is currently not possible to mock a class that extends from `Contao\System` with Mockery (...) due to the usage of `TL_MODE` as a default parameter value in `getMessages`:

````php
/**
 * Return all messages as HTML
 *
 * @param string $strScope An optional message scope
 *
 * @return string The messages HTML markup
 *
 * @deprecated Deprecated since Contao 4.0, to be removed in Contao 5.0.
 *             Use Message::generate() instead.
 */
protected function getMessages($strScope=TL_MODE)
{
  @trigger_error('Using System::getMessages() has been deprecated and will no longer work in Contao 5.0. Use Message::generate() instead.', E_USER_DEPRECATED);

  return Message::generate($strScope);
}
````

Since there is no constant `TL_MODE` in the test environment, this throws an exception:

````php
Mockery::mock(Widget::class);

// Throws: Use of undefined constant TL_MODE - assumed 'TL_MODE' (this will throw an Error in a future version of PHP)
````

This is because Mockery by default is creating `MethodDefinitions` for all methods on a mocked class.

It would be nice if the legacy Contao classes would be mockable via Mockery. 

**Fix**
The fix is simple by moving the parameter default into the method body - or in this case even simpler by removing the default value altogether since it is set in `Message::generate()` again, which is the only thing called within `System::getMessages()`.

````php
protected function getMessages($strScope=null)
{
  // Not really necessary, since $strScope = null will be set to TL_MODE in Message::generate().
  $strScope = $strScope ?? TL_MODE;

  trigger_deprecation('contao/core-bundle', '4.0', 'Using "Contao\System::getMessages()" has been deprecated and will no longer work in Contao 5.0. Use "Contao\Message::generate()" instead.');

  return Message::generate($strScope);
}
````

Note:
This seems to be the only case besides `Contao\Messages` where a legacy constant is used as a default parameter value.

If we agree that the default value can be dropped, I can submit a pull request. 
I'd suggest to target Contao 4.9.

### PR DESCRIPTION
**Pull request by SeverinGloeckle**
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3495 

Make it easier to mock the `System` class and its subclasses in unit-tests.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/System.php
+++ b/core-bundle/src/Resources/contao/library/Contao/System.php
@@ -1087,8 +1087,10 @@ abstract class System
 	 * @deprecated Deprecated since Contao 4.0, to be removed in Contao 5.0.
 	 *             Use Message::generate() instead.
 	 */
-	protected function getMessages($strScope=TL_MODE)
+	protected function getMessages($strScope=null)
 	{
+		$strScope = $strScope ?? TL_MODE;
+
 		@trigger_error('Using System::getMessages() has been deprecated and will no longer work in Contao 5.0. Use Message::generate() instead.', E_USER_DEPRECATED);
 
 		return Message::generate($strScope);

### CODE REVIEW

File: core-bundle/src/Resources/contao/library/Contao/System.php [hunk 1]
```diff
@@ -1087,8 +1087,10 @@ protected function addMessage($strMessage, $strType)
 	 * @deprecated Deprecated since Contao 4.0, to be removed in Contao 5.0.
 	 *             Use Message::generate() instead.
 	 */
-	protected function getMessages($strScope=TL_MODE)
+	protected function getMessages($strScope=null)
 	{
+		$strScope = $strScope ?? TL_MODE;
```
fritzmg, line 1092:
```suggestion
		// Set default scope in method body (see #3495)
		$strScope = $strScope ?? TL_MODE;
```

SeverinGloeckle: I relly like the comments in Contao's code base pointing to issues with background information on why some specific logic is required.

However, is this comment really helpful if you are not aware of the context that TL_MODE was previously set as the default parameter in the method?

fritzmg: I think it would be helpful to know for future reference (or deprecation) why we set a default in the method body here, as we don't usually do that (I think?).

leofeyer:
Shouldn't we use the `ScopeMatcher` instead of the `TL_MODE` constant?

SeverinGloeckle:
Yes, that would be the proper long-term approach.

Since `System::getMessages()` is already deprecated since Contao 4.0 though, it's probably better to change as little code as possible to avoid any side effects.

The `ScopeMatcher` would probably be a good addition for the `Message` class.

leofeyer (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3503, opened 2021-09-27
Correctly handle empty image size formats

aschempp:
The changes in #3475 break on my system if `$imageSizes->formats` is empty (no image conversion is enabled). It previously did not cause any problem because `array_map()` was not executed on an empty array.

### DIFF
--- a/core-bundle/src/Image/PictureFactory.php
+++ b/core-bundle/src/Image/PictureFactory.php
@@ -182,23 +182,25 @@ class PictureFactory implements PictureFactoryInterface
                     $formatsString = implode(';', StringUtil::deserialize($imageSizes->formats, true));
                     $formats = [];
 
-                    foreach (explode(';', $formatsString) as $format) {
-                        [$source, $targets] = explode(':', $format, 2);
-                        $targets = explode(',', $targets);
+                    if ('' !== $formatsString) {
+                        foreach (explode(';', $formatsString) as $format) {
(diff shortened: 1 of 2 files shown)


core-bundle/src/Image/PictureFactory.php
```diff
@@ -182,23 +182,25 @@ private function createConfig($size): array
...
-                        [$source, $targets] = explode(':', $format, 2);
-                        $targets = explode(',', $targets);
-
-                        if (!isset($formats[$source])) {
-                            $formats[$source] = $targets;
-                            continue;
-                        }
+                    if ('' !== $formatsString) {
+                        foreach (explode(';', $formatsString) as $format) {
+                            [$source, $targets] = explode(':', $format, 2);
```
ausi, line 188: How about adding the check here? Trailing `;` would then not fail either.

```suggestion
                            if ($format === '') {
                                continue;
                            }
                            [$source, $targets] = explode(':', $format, 2);
```

aschempp: That would never make a difference, only create a useless loop. I don't see how there could be a trailing `;`?

m-vo:
```php
foreach (array_filter(explode(';', $formatsString)) as $format {
    // …
}
```
? :grin:

fritzmg (approved):

PR #3504, opened 2021-09-27
Add assertions to the picture factory tests

aschempp:
While implementing https://github.com/contao/contao/pull/3503 I noticed that the `PictureFactoryTest` contains a lot of assertions in method callbacks, but there is not assertion that the method (and therefore the callbacks) are actually executed.

Unfortunately, now one test fails because apparently there's an invalid assertion. @qzminski @ausi maybe you can tell whether the assertions should not be there or something else is wrong?

### DIFF
--- a/core-bundle/tests/Image/PictureFactoryTest.php
+++ b/core-bundle/tests/Image/PictureFactoryTest.php
@@ -251,30 +251,7 @@ class PictureFactoryTest extends TestCase
             ->willReturn($pictureMock)
         ;
 
-        $imageFactory = $this->createMock(ImageFactoryInterface::class);
-        $imageFactory
-            ->expects($this->once())
-            ->method('create')
-            ->with(
-                $this->callback(
-                    function (string $imagePath) use ($path): bool {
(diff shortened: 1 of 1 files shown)


core-bundle/tests/Image/PictureFactoryTest.php
```diff
@@ -250,6 +253,7 @@ function (ResizeOptions $options): bool {
 
         $imageFactory = $this->createMock(ImageFactoryInterface::class);
         $imageFactory
+            ->expects($this->once())
```
ausi: `$imageFactory` Can be removed completely here.
It would only be needed if `$pictureFactory->create()` is called with a path instead of an image object which is not the case here.

aschempp: 👍  removed in ea2ab19dfe8fa433393adb03143c4def46a69ce4

ausi (approved):

qzminski (approved):

Toflar (approved):

PR #3506, opened 2021-09-27
Add the missing stuff from the 4.9-4.12 upstream merge

### LINKED ISSUE #3477: Route defaults inconsistencies
**Affected version(s)**

Contao 4.12

**Description**

https://github.com/contao/contao/blob/4cf0a3ce0b5eeb52a35037648e23b4d29cf8c9de/core-bundle/src/Routing/Route404Provider.php#L158

@ausi Does this need to be wrapped in `LocaleUtil::formatAsLanguageTag()` like in the `PageRoute` class?

https://github.com/contao/contao/blob/4cf0a3ce0b5eeb52a35037648e23b4d29cf8c9de/core-bundle/src/Routing/Page/PageRoute.php#L55

@aschempp Do we need to add `'_canonical_route' => 'tl_page.'.$page->id,` after this line like in the `Route404Provider` class?

leofeyer:


core-bundle/src/Routing/Page/PageRoute.php
```diff
@@ -47,17 +47,21 @@ public function __construct(PageModel $pageModel, string $path = '', array $defa
...
-        );
+        $initialDefaults = [
+            '_token_check' => true,
+            '_controller' => 'Contao\FrontendIndex::renderPage',
+            '_scope' => ContaoCoreBundle::SCOPE_FRONTEND,
+            '_format' => 'html',
+            '_canonical_route' => 'tl_page.'.$pageModel->id,
+        ];
+
+        if ($pageModel->rootLanguage) {
```
aschempp, line 58: why do we need an if-statement here? The page model always has a `rootLanguage`

leofeyer: Maybe the unit test does not set it everywhere?

aschempp: most likely, yes. But we're always calling `loadDetails()` on PageModel so it must be there. I'll do a PR to fix this.

PR #3507, opened 2021-09-27
Handle pages from PageRegistry in DebugPagesCommand

### LINKED ISSUE #3505: DebugPagesCommand handles legacy page types only
**Affected version(s)**

Contao 4.11 and up

**Description**
The command `contao-console debug:pages` only shows pages registered in `$GLOBALS['TL_PTY']`:

https://github.com/contao/contao/blob/c88eddcf7eb71916036512d21ebbe53dfdcb1eed/core-bundle/src/Command/DebugPagesCommand.php#L71

**Expected behaviour**
It should show custom page types from the `PageRegistry` as well.

**Possible fix**
If the expected behaviour is correct, I can submit a fix and maybe add a unit-test for `DebugPagesCommand`.

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3505 

The `DebugPagesCommand` now lists both pages from `$GLOBALS['TL_PTY']` and from the `PageRegistry`.

Additionally, the output of the command is optimized for empty values:
- Show token `-` for empty arrays in requirements, defaults & options instead of empty space
- Show token `-` for empty strings in path & url suffix instead of empty space


core-bundle/src/Command/DebugPagesCommand.php
```diff
@@ -86,13 +86,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
             $rows[] = [
                 $type,
-                $config ? $config->getPath() : '*',
-                $config ? $config->getUrlSuffix() : '*',
+                $config ? $config->getPath() ?: '-' : '*',
+                $config ? $config->getUrlSuffix() ?: '-' : '*',
```
SeverinGloeckle: I'm not sure whether an empty path & suffix should be indicated with `-` (clear that the config doesn't have these) or `*` (default value with no config).

m-vo:
@aschempp I guess there is no way that the page registry also contains the records for `$GLOBALS['TL_PTY']`, no? I find it kind of strange that the consumer needs to merge entries from these two sources.

SeverinGloeckle:
Since `PageRegistry::keys()` returns the keys for all registered page/route configurations, each page type from `$GLOBALS['TL_PTY']` would have to be registered, presumably with an empty `RouteConfig`.

On a quick glance this should work without any side effects since `PageRegistry::getRoute()` already falls back to an empty route config for unknown (= legacy) page types:

https://github.com/contao/contao/blob/3a7dd11b134fb473cb070598a249c28a93933ef2/core-bundle/src/Routing/Page/PageRegistry.php#L65-L69

Adding multiple empty configurations would just lead to some additional iterations when checking values of registered configurations (e.g. url suffixes).

However, while the tagged page services can be added in `RegisterPagesPass`, it is not possible to add the legacy page types at this point since `$GLOBALS['TL_PTY']` is not available yet. The legacy pages are available only after loading the config in `ContaoFramework::initializeFramework`.

So the registration could happen:
- via a custom call in `ContaoFramework::initializeFramework`
- via an initializeSystem hook (with a low priority to catch as many other callbacks as possible)

This would still not catch custom page types that are added to `$GLOBALS['TL_PTY']` at runtime after the framework initialisation.

If @aschempp agrees, I may adjust this pull request. 
Or maybe there should be a follow up request to this one.

aschempp:
I don't think adding the legacy pages to the registry should be part of this bug fix PR, and I'm not sure we should actually change that in a hotfix release.

SeverinGloeckle:
Sounds reasonable to me. So it's just this bugfix to add the custom page types to `DebugPagesCommand`.

core-bundle/src/Command/DebugPagesCommand.php
leofeyer: I'd vote for `-`.

aschempp:
```suggestion
                ($config && $config->getPath()) ? $config->getPath() : '*',
                ($config && $config->getUrlSuffix()) ? $config->getUrlSuffix() : '*',
```

core-bundle/src/Command/DebugPagesCommand.php
```diff
@@ -86,13 +86,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
                 $this->pageRegistry->supportsContentComposition($page) ? 'yes' : 'no',
                 isset($this->routeEnhancers[$type]) ? \get_class($this->routeEnhancers[$type]) : '-',
-                $config ? $this->generateArray($config->getRequirements()) : '-',
-                $config ? $this->generateArray($config->getDefaults()) : '-',
-                $config ? $this->generateArray($config->getOptions()) : '-',
+                $config ? $this->generateArray($config->getRequirements()) ?: '-' : '-',
```
aschempp: we should return `-` in `generateArray` of the array is empty instead of handling it here.

aschempp (changes requested):

core-bundle/src/Command/DebugPagesCommand.php
SeverinGloeckle: I infer that you prefer to continue showing a `*` for configurations without a path / url suffix (instead of a `-` as for other cells)?

core-bundle/src/Command/DebugPagesCommand.php
SeverinGloeckle: Yes, I like this.

SeverinGloeckle:
I have adjusted the generation of empty values.

Can you please review whether the generated overview is as intended? Especially path and suffix `*` for the custom type `baz` (see test case `'With custom pages'`):

Custom page types:
````php
['root', new RouteConfig('foo', null, '.php', [], [], ['_controller' => RootPageController::class]), null],
['bar', new RouteConfig('foo/bar', null, '.html', [], [], ['_controller' => TestPageController::class]), null],
['baz', new RouteConfig(null, null, null, ['page' => '\d+'], ['utf8' => false], []), null]
````

Legacy types:
````php
'regular' => PageRegular::class,
'root' => PageRoot::class
````

Output:
````
Contao Pages
============
 --------- --------- ------------ --------------------- ---------------- -------------- ----------------------------------------------------------------------------- -------------- 
  Type      Path      URL Suffix   Content Composition   Route Enhancer   Requirements   Defaults                                                                      Options       
 --------- --------- ------------ --------------------- ---------------- -------------- ----------------------------------------------------------------------------- -------------- 
  bar       foo/bar   .html        no                    -                -              _controller : Contao\CoreBundle\Fixtures\Controller\Page\TestPageController   -             
  baz       *         *            no                    -                page : \d+     -                                                                             utf8 : false  
  regular   *         *            yes                   -                -              -                                                                             -             
  root      foo       .php         no                    -                -              _controller : Contao\CoreBundle\Controller\Page\RootPageController            -             
 --------- --------- ------------ --------------------- ---------------- -------------- ----------------------------------------------------------------------------- -------------- 

````

aschempp:
It looks correct to me. `*` means "anything is allowed" and `-` means "not available"/"no information" to me. We _could_ extend the information for unknown/legacy page types (e.g. we know the controller), but I'm not sure that's worth the effort?

SeverinGloeckle:
I'm fine with `*` for path and URL suffix. 

Adding additional debug information is outside the scope of this bug fix and could be done as a follow up.

If everything looks fine to you, please resolve the pending review.

aschempp (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3516, opened 2021-09-29
Fix the PHP 8.1 job

leofeyer:


core-bundle/src/Image/Studio/Figure.php
```diff
@@ -60,7 +60,6 @@ class Figure
      * All arguments but the main image result can also be set via a Closure
      * that only returns the value on demand.
      *
-     * @param ImageResult                                                                 $image          Main image
```
m-vo, line 63: Why this? :upside_down_face: 

Isn't the description helpful? Or if not, why keep it with the other parameters?

leofeyer: I don't know exactly why the CS fixer has removed it (probably because there is a type hint already), but no, I don't consider the description "Main image" very helpful. Especially as there is no secondary image or something alike that could make people misinterpret the parameter called `$image`. No hard feelings. 🙈

m-vo: Well there is a secondary image (lightbox) but it is derived. Fine for me anyways. :relaxed:

PR #3517, opened 2021-09-29
Merge duplicate tags in HTML attributes config

### LINKED ISSUE #3485: Duplicate allowed attribute tags should throw an error or be merged
Currently if you set allowed attributes for the same tag twice, the second row overwrites the first one.

We should either not allow to save such configurations, or merge the two attributes into one.

Related: https://community.contao.org/de/showthread.php?81673#post549518

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3485


leofeyer:
Thank you @ausi.

PR #3518, opened 2021-09-29
Implemented a contao:database:dump command

Toflar:
This is the complementary draft PR for https://github.com/contao/contao/pull/3466.

I've researched the use on the dump library (or none at all) quite a bit before going for `webfactory/slimdump`.
Here are the reasons for my decision

* Requiring the `mysqldump` command line tool seemed like an option but this has 2 serious disadvantages:
    1. We all know, not everybody has access to `mysqldump` with their hosting provider
    2. Actually, even if they had access, we could have an issue with making the dump process compatible with the import process as we have no control over the version of `mysqldump` installed and thus, we might run into compatibility issues.
* https://packagist.org/packages/ifsnop/mysqldump-php is the most popular library but there does not seem to be a lot of activity. Especially PRs regarding PHP 8 are being ignored.
* https://packagist.org/packages/dg/mysql-dump seems popular too but uses `mysqli` directly and thus does not seem to be the right fit for us as we use Doctrine/PDO.
* https://packagist.org/packages/rah/danpu is also very popular but kind of has the same issue as it requires to pass the DSN.
* I then decided for https://github.com/webfactory/slimdump because it seems maintained, I like the code, it's based on Doctrine and I know at least one maintainer which is also active in the Symfony environment :) 

To do:

- [ ] Decide on whether that should be core-bundle or manager-bundle (and then we should also move contao:migrate maybe)
- [x] Should I merge both PR's (so this one with https://github.com/contao/contao/pull/3466) into one so we have everything db related in one PR? I could add functional tests to ensure the import command can handle the dumps generated by the dump command.
- [ ] Tests

### DIFF
--- a/core-bundle/src/Command/DatabaseDumpCommand.php
+++ b/core-bundle/src/Command/DatabaseDumpCommand.php
@@ -1,0 +1,160 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Command;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Schema\Table;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Formatter\OutputFormatterInterface;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Output\StreamOutput;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Webfactory\Slimdump\Config\Config;
+use Webfactory\Slimdump\Config\ConfigBuilder;
+use Webfactory\Slimdump\DumpTask;
+
+/**
+ * Dumps the database.
+ *
+ * @internal
+ */
+class DatabaseDumpCommand extends Command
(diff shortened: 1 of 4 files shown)


m-vo:
Nice!

Just adding some use cases I had in the past when implementing database tooling for Contao projects:

* ignore a table completely
* ignore the data (i.e. just dump the structure)
* allow to alter the data (either before dumping or before/when importing)
* dump each table in a separate file and omit date information (allow versioning)

> Should I merge both PR's (so this one with Implemented a contao:database:import command #3466) into one so we have everything db related in one PR? I could add functional tests to ensure the import command can handle the dumps generated by the dump command.

:+1:  I'd say yes. To me these two things belong together (and must function together).

core-bundle/src/Command/DatabaseDumpCommand.php
```diff
@@ -0,0 +1,160 @@
...
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setAliases(['contao:db:dump'])
+            ->addArgument('file', InputArgument::REQUIRED, 'The path to the SQL dump.')
+            ->addOption('buffer-size', 'b', InputOption::VALUE_OPTIONAL, 'Maximum length of a single SQL statement generated. Requires said amount of RAM. Defaults to "100MB".')
+            ->addOption('ignore-tables', 'i', InputOption::VALUE_OPTIONAL, 'A comma-separated list of database tables to ignore. Defaults to the Contao configuration (contao.db.dump.ignoreTables).')
```
aschempp, line 52: don't we already have an option to tell Contao to ignore certain tables?

Toflar:
Replaced by https://github.com/contao/contao/pull/3594.

PR #3519, opened 2021-09-29
Do not encode option values

### LINKED ISSUE #3421: Don’t use input encoding for fields with predefined options
Encoding the values for fields with `options` like `select` or `radio` doesn’t make much sense I think because the values are predefined and can therefore not be used to inject bad code.

See https://github.com/isotope/core/issues/2249#issuecomment-911553712

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3421


Toflar (approved):

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #3523, opened 2021-09-29
Upgrade to Doctrine 3

### LINKED ISSUE #2750: Changing a field's sql type from blob to mediumblob isn't recognized by migrate script
**Affected version(s)**

Contao 4.9

**Description**

I need a larger field "tokens" in tl_nc_queue.tokens in the notification center. Therefor I created a DCA file:

```
<?php

$dca = &$GLOBALS['TL_DCA']['tl_nc_queue'];

$dca['fields']['tokens']['sql'] = 'MEDIUMBLOB NULL';
```

In addition I added `notification_center` to the load after array in the `Plugin` class.

Now if I remove the field in the database and run contao:migrate, it works. But if there's already a field with the type blob of the same name, it doesn't alter the field.

Is it a bug? Or is it a quirk in mysql?

Thanks in advance!

Bye Defcon0

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2750

~~First draft for~~ compatibility with doctrine 3.


leofeyer:
Yeah! 🎉

Do we want to support both Doctrine 2 and 3? How long will Doctrine 2 be maintained?

Toflar:
I'd be fine with going for 3 only for the next LTS.

leofeyer:
Do you happen to know when `doctrine/orm` version 2.10 will be released?

ausi:
No, I couldn’t find any information about it. We would have to wait with merging this pull request until `doctrine/orm` 2.10 is released.

ausi:
> Do you happen to know when `doctrine/orm` version 2.10 will be released?

Yesterday 🎉 https://github.com/doctrine/orm/releases/tag/2.10.0

fritzmg (dismissed):

Toflar (dismissed):

m-vo (dismissed):

aschempp:
How can extensions work with major BC breaks like that? I often use `MySqlPlatform::LENGTH_LIMIT_BLOB` in my DCA to set the field length, but that would break with this update right?

fritzmg:
@aschempp your extension wouldn't break - it just would not be installable in Contao 4.13+. Assuming your extension defined a dependency on `"doctrine/dbal": "^2.0"` for example, since it is using `Doctrine\DBAL\Platforms\MySqlPlatform::LENGTH_LIMIT_BLOB`.

aschempp:
yeah… apart from the fact that it most likely does not, I wouldn't really be able to actually support doctrine 2 and 3? I mean should I do a `class_exists` in the DCA in all the places? 🤔

m-vo:
> I wouldn't really be able to actually support doctrine 2 and 3?

At least for the stupid renaming from `MySqlPlatform` to `MySQLPlatform` there could be a BC shim package that adds something like this:

```php
<?php

declare(strict_types=1);

namespace Doctrine\DBAL\Platforms;

if (class_exists(MySQLPlatform::class)) {
    class MySqlPlatform extends MySQLPlatform {}
}

**Comment by ausi** [comment]
> At least for the stupid renaming from `MySqlPlatform` to `MySQLPlatform` there could be a BC shim package that adds something like this

As class names are case insensitive in PHP this will not work I think.
`Cannot declare class MySqlPlatform, because the name is already in use in …`

Triggering the autoloader for both case variations should be all you need I think:

```
class_exists('Doctrine\DBAL\Platforms\MySQLPlatform') || class_exists('Doctrine\DBAL\Platforms\MySqlPlatform');
```

But all this is unrelated to this pull request IMO.

leofeyer (dismissed):

leofeyer:
Thank you @ausi.

PR #3524, opened 2021-09-29
Add an event count to the event list

### LINKED ISSUE #74: eventCount in event_ templates
**Description**  
Since 4.6 we have the function eventCount, but we cannot use it in _event__ templates. [Originally we asked to have the same logic as in news](https://github.com/contao/core/issues/7825). Unfortunately, this is not so. Could you make it as in news, please? 
**Example**  
Then we could influence the _event__ templates like [in this example](http://help.rocksolidthemes.com/discussions/contao/12212-verwendung-newsnachrichten-in-rocksolid-columns).

leofeyer:
| Q                | A
| -----------------| ---
| Implemented issue     | #74
| Docs PR or issue | -

The implementation behaves exactly like in the news module. However, I have noticed that counting starts at `1` and not at `0`. @ausi Is this something we want to change in the future?

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
@@ -248,6 +248,7 @@ class ModuleEventReader extends Events
 		$objTemplate->until = $until;
 		$objTemplate->locationLabel = $GLOBALS['TL_LANG']['MSC']['location'];
 		$objTemplate->calendar = $objEvent->getRelated('pid');
+		$objTemplate->count = 0; // see #74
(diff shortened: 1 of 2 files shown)


ausi:
> However, I have noticed that counting starts at `1` and not at `0`. @ausi Is this something we want to change in the future?

I don’t think we should change that. If I’m reading the code correctly, the reader module sets the count to `0` so the template could actually use this to know if there is a count at all.

leofeyer:
I am explicitly setting the counter to `0`, because that is how the news module behaves. It was `null` before.

ausi (approved):

fritzmg (approved):

PR #3527, opened 2021-09-30
Update the Facebook share link

aschempp:
I noticed that the Facebook share link no longer works. The issue can be reproduced on https://contao.org/en/news/contao-4_12_0.html by clicking the Facebook icon at the bottom. 

Unfortunately, I couldn't find any official documentation since the ? `sharer.php` is deprecated in favor of their JS library, which is not an option for us. Searching through the net I found https://stackoverflow.com/questions/20956229/has-facebook-sharer-php-changed-to-no-longer-accept-detailed-parameters/22969530#22969530 and using `p[url]=` instead of `u=` seems to make it work again.

### DIFF
--- a/core-bundle/src/Resources/contao/controllers/FrontendShare.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendShare.php
@@ -31,7 +31,7 @@ class FrontendShare extends Frontend
 			case 'facebook':
 				return new RedirectResponse(
 					'https://www.facebook.com/sharer/sharer.php'
-						. '?u=' . rawurlencode(Input::get('u', true))
+						. '?p[url]=' . rawurlencode(Input::get('u', true))
 				);
 
 			case 'twitter':


ausi:
Unable to reproduce

Bildschirmfoto 2021-09-30 um 10 24 32

Toflar:
Unable to reproduce.

leofeyer:
I'm also not able to reproduce this. I guess your ad blocker or something alike prevents sharing the link on Facebook.

aschempp:
Bildschirmfoto 2021-10-05 um 09 29 07

Here's what I get. This is stock Safari 14.1.2 on MacOS 11.5.2 without adblock extensions (only 1Password is installed). I don't get that error with the other URL attribute.

fritzmg:
@aschempp that error message is related to the Facebook account's setting. You need to go to https://www.facebook.com/settings?tab=applications&ref=settings and klick on "Turn On" for "Apps, websites and games". It's not related to the share URL.

aschempp:
Thanks for the hint! Here's my setting
Bildschirmfoto 2021-10-05 um 09 47 46

I obviously don't want to login with my Facebook account on an app if I just want to share something. So maybe we should actually update the URL so sharing works for people that have this disabled?

leofeyer (approved):

PR #3530, opened 2021-09-30
Replace patchwork/utf8 with symfony/string

### LINKED ISSUE #2630: Replace patchwork/utf8 with symfony/string
We should switch to symfony/string as soon as Contao is Symfony 5 only: `Package patchwork/utf8 is abandoned, you should avoid using it. Use symfony/polyfill-mbstring or symfony/string instead.`

leofeyer:
| Q                | A
| -----------------| ---
| Implemented issue     | #2630
| Docs PR or issue | -

Contao now uses `mb_` functions (optionally made available by `symfony/polyfill-mbstring`) for casing and length and `symfony/string` for transliterating.


calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
```diff
@@ -55,7 +54,7 @@ public function generate()
 		if ($request && System::getContainer()->get('contao.routing.scope_matcher')->isBackendRequest($request))
 		{
 			$objTemplate = new BackendTemplate('be_wildcard');
-			$objTemplate->wildcard = '### ' . Utf8::strtoupper($GLOBALS['TL_LANG']['FMD']['calendar'][0]) . ' ###';
+			$objTemplate->wildcard = '### ' . $GLOBALS['TL_LANG']['FMD']['calendar'][0] . ' ###';
```
m-vo, line 57: Did you remove the `strtoupper` calls intentionally?

leofeyer: Yes. This is now done via CSS.

core-bundle/composer.json
```diff
@@ -96,6 +95,7 @@
         "symfony/process": "^5.3",
         "symfony/routing": "^5.3",
         "symfony/security-bundle": "^5.3",
+        "symfony/string": "^5.3",
```
m-vo, line 99: As the core bundle now is using  lots of`mb_*` functions, don't we need to add `"symfony/polyfill-mbstring": "^1.0"` here as well?

m-vo (changes requested):

core-bundle/composer.json
leofeyer: You are absolutely right. Added in 17882bf720ca06945124c88d8b827ed2bcb2f06a.

m-vo (approved):

PR #3534, opened 2021-10-01
Translate all languages in the file meta data

### LINKED ISSUE #3533: Croatian and Romanian language abbreviations are not recognized correctly
**Affected version(s)**

Contao 4.9.x-dev

**Description**

In the site structure, I can fill in the language abbreviations "ro" (Romanian) and "hr" (Croatian) from ISO Standard 639-1.
image

In the file management, when you edit meta data of an image, all languages occur in their full notation, e.g. "Französisch - Français" - except Croatian and Romanian, they show the  abbreviations.

image

The following languages (abbreviations) do not have the problem:
- de
- en
- fr
- nl
- sl
- sv
- cs
- it

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/issues/3533

The first parameter to `System::getLanguages()` says to only return installed languages - which can't be right in this context.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/MetaWizard.php
+++ b/core-bundle/src/Resources/contao/widgets/MetaWizard.php
@@ -133,7 +133,7 @@ class MetaWizard extends Widget
 			return '<p class="tl_info">' . $GLOBALS['TL_LANG']['MSC']['metaNoLanguages'] . '</p>';
 		}
 
-		$languages = $this->getLanguages(true);
+		$languages = $this->getLanguages();
 
 		// Add the existing entries
 		if (!empty($this->varValue))


Toflar (approved):

ausi:
This would be a behavior change. Are we sure we want to change that in a bugfix release?

ausi:
> which can't be right in this context.

@aschempp why?

fritzmg:
> This would be a behavior change

Why? The meta wizard always translated the language codes, regardless of whether it's an installed language or not (at least in prior Contao versions - 4.4 for instance), since this regards the front end, not the back end.

ausi:
This is not about translating the language codes but about the “add language” select menu. We removed this select in https://github.com/contao/contao/commit/96b9d4dea12d2edff6ccf161711f236caa22f146 so I don’t think we should change its values for Contao 4.9 now.

fritzmg:
But this issue is not about the select. It's about the label in the top left corner, which is currently untranslated for languages that are not "installed languages".

ausi:
Before: Bildschirmfoto 2021-10-04 um 16 08 54

After: Bildschirmfoto 2021-10-04 um 16 09 40

fritzmg:
Ah, I see, this also affects the select. Then yes, this should be reverted and a different solution should be found to translate the language label.

ausi:
> But this issue is not about the select.

Then it should have not changed the values of the select menu.

leofeyer:
Here is the correct fix: #3543

PR #3535, opened 2021-10-01
Set the contao.web_dir parameter from composer.json

### LINKED ISSUE #3363: Issues with switching to public as the default web root
**Affected version(s)**

Contao 4.12?

**Description**

Detecting the use of `web` as public directory by checking if the resource exists on the filesystem might lead to unwanted behavior, see https://github.com/contao/contao/pull/3084#issuecomment-903163906.

I noticed the problem when creating a new project from a template/skeleton (now with 4.12): 

* the project assumes `web` as the web root
* `web`, however, does not exist initially &rarr; `public` gets created on install
* webpack is run and outputs stuff into the now created `web` directory
* the symfony server is started and picks up `public` as the web root
* the cache is cleared after some work, now `web` exists and things break<sup>*)</sup>

<sup>*)</sup> because the `contao.web_dir` config setting now differs from the actual web root

The same thing will happen when deploying by copying a bare minimum and installing, i.e. without explicitly creating the web directory. There the website will stop working as no entry points will be at the expected location.

IMHO we should neither rely on the existence nor absence of the web directory but only on the config. I've also seen projects using `web` as their js/layout root. :see_no_evil: 

What about this strategy instead?

- use `extra.public-dir` in the app's `composer.json` to set the public dir [like Symfony does](https://symfony.com/doc/current/configuration/override_dir_structure.html#override-the-public-directory)
- default config to `web` unless configured otherwise via `composer.json` (or config)
- ship the switch to `public` via the 4.12 managed edition `composer.json`, so that only new projects are affected

We could then ultimately change the default in Contao 5 and remove the `composer.json` config key again (as a BC break).

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3363
| Docs PR or issue | -

https://github.com/contao/contao/issues/3363#issuecomment-931469709

> As discussed in the Contao call on September 30th, we want to set contao.web_dir from extra.public-dir in the composer.json file and deprecate the parameter in the configuration tree.


core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -63,7 +65,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $loader->load('services.yml');
         $loader->load('migrations.yml');
 
-        $container->setParameter('contao.web_dir', $config['web_dir']);
+        $container->setParameter('contao.web_dir', $this->getComposerPublicDir($projectDir) ?? $config['web_dir']);
```
m-vo, line 70: This is a forward compatible implementation. It should allow us to remove the `contao.web_dir` config entry in Contao 5 and simply make `public` the fallback here:
```suggestion
        $container->setParameter('contao.web_dir', $this->getComposerPublicDir($projectDir) ?? 'public');
```

fritzmg (dismissed):

ausi (dismissed):

aschempp (dismissed):

leofeyer (dismissed):

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -103,6 +103,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->end()
                 ->scalarNode('web_dir')
                     ->info('Absolute path to the web directory. Defaults to %kernel.project_dir%/public.')
+                    ->setDeprecated('contao/core-bundle', '4.13', 'Setting the web directory in a config file is deprecated since Contao 4.13. Use the "extra.public-dir" config key in your root composer.json instead.')
                     ->cannotBeEmpty()
                     ->defaultValue($this->getDefaultWebDir())
```
leofeyer, line 108: I wonder if we should reset this to `Path::join($this->projectDir, 'web')` now?

m-vo: I don't know. Then everyone that is using `public` now (without having configured anything) would get `web` again, right?

leofeyer: You are right, that is a bad idea.

leofeyer: So how do we proceed in Contao 5? Do we remove the entire `->scalarNode('web_dir')`?

m-vo: IMHO yes. We should only keep the parameter.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
leofeyer: You mean `Path::join($projectDir, 'public')`, don't you?

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3537, opened 2021-10-01
Harden more time-sensitive tests

m-vo:
The tests that I adjusted in this PR failed in one of the last CI runs because they're using `time()` and the output was off by one.

Unfortunately you cannot see the CI results after rerunning the jobs (which fixed the momentary issue).

### DIFF
--- a/core-bundle/phpunit.xml.dist
+++ b/core-bundle/phpunit.xml.dist
@@ -58,6 +58,8 @@
                             <element key="1"><string>Contao\CoreBundle\Tests\Command</string></element>
                             <element key="2"><string>Contao\CoreBundle\Cron\Command</string></element>
                             <element key="3"><string>Contao\CoreBundle\Tests\Cron</string></element>
+                            <element key="4"><string>Contao\CoreBundle\EventListener</string></element>
+                            <element key="5"><string>Contao\CoreBundle\Tests\EventListener</string></element>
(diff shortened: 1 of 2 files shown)


leofeyer:
We might as well hardcode the timestamp, as it is does not seem to be relevant for the tests, does it?

m-vo:
> We might as well hardcode the timestamp, as it is does not seem to be relevant for the tests, does it?

I think it's set here (L172), which causes the trouble in the test:

https://github.com/contao/contao/blob/a03017dc5d3a17425f31abf9200680459812ae84/core-bundle/src/EventListener/DataContainer/ContentCompositionListener.php#L168-L180

Or do you mean replacing the `with` assertion with a callback that ignores `tstamp`?

leofeyer:
I am talking about replacing `time()` with `1633346605` or any other integer.

m-vo:
My referenced code above (where `time()` gets set) isn't from the tests. We cannot replace it there. :stuck_out_tongue:

leofeyer:
<del>Yes, we can. As I said above, it is irrelevant for the test whether `tstamp` is a fixed timestamp or not.</del>

No, you are right, of course. Sorry.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

Issue #3387, opened 2017-12-02
Twig Templates


Fruchuxs:
In alten Feature-Requests und Co habe ich des öfteren gelesen, dass es wohl geplant ist, die Templates mit Twig zu realisieren. Zurück in der Zukunft bei Contao 4, sehe ich immer noch diese "ätzenden" PHP-Templates.
Ist das noch geplant, oder eher nicht?

m-vo:
@leofeyer Shouldn't this issue be closed or which are the 'incompatible features' that are missing? :upside_down_face:

leofeyer:
I figured we would replace all `.html5` templates with `.html.twig` templates in Contao 5. 🙈

m-vo:
Ah yes. We could actually do this once some yet missing BE features are implemented. :+1:

leofeyer:
What exactly is missing?

m-vo:
Most important for the BE is…

* showing the template in the selects if the base template is a Twig template (#3274)
* allow to create new templates in the BE
* better support for editing templates (ideally with support for blocks) and appropriate cache invalidation when saving

We probably still need to allow users to overwrite/extend core templates with `.html5` templates in Contao 5. There are two ways we could achieve this: either by enhancing the old template engine, so that extending Twig templates is possible<sup>*)</sup> - we could go fully Twig then. Or by getting it to work that both variants can be provided - we would then need to maintain both for some time.

<sup>*)</sup> I thought this wasn't feasible but I just had another idea. I'll look into this in the upcoming weeks.

ausi:
> Or by getting it to work that both variants can be provided - we would then need to maintain both for some time.

I’d vote for this solution. Maintaining the html5 templates is not needed for most cases I think, because they are only present for BC and are not used in a new setup.

Also, this would force users to use twig for extending new templates (because there is no BC html5 version). This would be a strong signal to every Contao user that they have to switch to Twig at some point.

leofeyer:
Twig templates will be the default in Contao 5, even though `.html5` templates are still supported.

PR #3538, opened 2021-10-03
Allow providing Twig templates alongside PHP templates

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | One step towards #3387
| Docs PR or issue | todo

This PR aims at allowing Twig and PHP Templates side by side. This way we could add Twig versions of our own templates and deprecate the PHP ones. Users could still use PHP templates for the existing stuff but we could introduce new features for Twig only. This would also allow us to provide the correct content when creating Twig templates from within the BE 

Note: It still won't be possible to inherit a Twig template from a PHP template. So if an extension decides to overwrite/extend a Twig template from the core, a user will only be able to use/extend it if when using Twig as well. Otherwise the core PHP template would be used as a base.

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -944,6 +944,7 @@ services:
             - '@Contao\CoreBundle\Twig\Loader\TemplateLocator'
             - '@Contao\CoreBundle\Twig\Loader\ThemeNamespace'
             - '%kernel.project_dir%'
+        public: true
         tags:
             - { name: twig.loader, priority: 2 }
             - { name: kernel.reset, method: reset }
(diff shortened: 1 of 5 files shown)


m-vo:
Not sure about the milestone. If this will also ship Twig templates for the core this might need to go into 4.13.

leofeyer:
So with this change we could already replace all default .html5 templates with .twig templates?

m-vo:
> So with this change we could already replace all default .html5 templates with .twig templates?

Hopefully :slightly_smiling_face:
 For BC we would need to keep the `.html5` ones as well for some time, though.

aschempp:
I guess we are (only) allowed to add a new method to an interface because Twig is still experimental, right?

m-vo:
Kind of superseded by the new naming scheme (`content_element/text`, …), where collisions (old/new) won't happen anymore. We can still revive the PR should the need arise, still.

PR #3540, opened 2021-10-04
Improve handling of pages with `tl_page.requireItem`

### LINKED ISSUE #3450: Pages with `tl_page.requireItem` throw exceptions in ModuleBreadcrumb
**Affected version**

 Contao 4.12

**Description**

Using `ModuleBreadcrumb` on a page with `tl_page.requireItem` enabled ("Require an item"-option) throws:

````php
Contao\CoreBundle\Exception\RouteParametersException

"Parameter "parameters" for route "cmf_routing_object" must match "/.+" ("" given) 
to generate a corresponding URL."
````

The exception is caught within `ModuleBreadcrumb` and the error is just printed to the system log:
> Unable to generate URL for page ID X: Parameter "parameters" for route "cmf_routing_object" must match "/.+" ("" given) to generate a corresponding URL.

**Details**
This is because the `ModuleBreadcrumb` does not pass the parameters to the router:

https://github.com/contao/contao/blob/50079d9fe09ad802ca3bdde6d7ccac6b2338739d/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php#L202

https://github.com/contao/contao/blob/50079d9fe09ad802ca3bdde6d7ccac6b2338739d/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php#L250-L262

**Possible fix**
The current auto-item should be resolved by `ModuleBreadcrumb` and passed to the router – either by default or just if the current page has `tl_page.requireItem` set to true.

As an alternative, there could not be generated any url since the breadcrumb menu does not require the url of the current page.

### Bug 2: Parent pages which require an item
The exception is thrown as well if any other page in the breadcrumb trail has `tl_page.requireItem` set to true. While I can hardly think of a reason to have a child page on a page with an auto-item, I still think this is a bug.

**Possible fix**
In my opinion, a parent page which requires an item should not be included in a breadcrumb trail since no correct url can be generated. Thus, it could be ignored as well in:

https://github.com/contao/contao/blob/50079d9fe09ad802ca3bdde6d7ccac6b2338739d/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php#L100-L103

### LINKED ISSUE #3525: Frontend preview throws exception for pages with `tl_page.requireItem`
**Affected version(s)**

Contao 4.12, 4.11 (likely Contao 4.10 as well)

**Description**

Starting the frontend preview for a page with `tl_page.requireItem` enabled ("Require an item"-option) throws:

````php
Contao\CoreBundle\Exception\RouteParametersException

"Parameter "parameters" for route "cmf_routing_object" must match "/.+" ("" given) 
to generate a corresponding URL."
````

This exception is not caught.

**Details**
The exception is thrown since the router cannot generate the URL for a page which requires an item without an item parameter (somewhat similar to #3450).

**Possible fix**
Generating a proper URL is not possible in this case since there is no way to guess the correct item for the page. 

So there are two options left:
1. Generate the base URL of the page without the item parameter. This will likely show a 404 page in the frontend preview as a result of calling the page without an item parameter (similar to the old behaviour where modules like news readers would trigger the 404 on such pages).
2. Disable the frontend preview for pages which require an item and redirect to the default root page (that's how it is done for non-previewable content already)

Neither option is perfect, but each is preferable to the exception.

### PR DESCRIPTION
**Pull request by SeverinGloeckle**
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3450 , #3525 

With the new routing, trying to generate a URL for a page with `tl_page.requireItem=1` throws a `RouteParametersException`. This pull request prevents or improves the handling of this exception in breadcrumbs and navigation modules as well as in the frontend preview.

Improved cases:

**1. Breadcrumbs, current item has `requireItem`**
Generate the URL from the current request instead of re-generating it from the taget page.

Note: The URL is generated via `strtok(Environment::get('request'), '?')` so any query string is not added to the breadcrumbs URL. 

**2. Breadcrumbs, item in parent trail has `requireItem`**
Skip pages with `tl_page.requireItem=1` in the parent trail of the breadcrumbs, even if `showHidden` is enabled.

Note: These pages are skipped explicitly in the breadcrumbs (and not handled by the future proof fix below) to prevent spamming the log with error messages for failed URL generations.

**3. Future proof breadcrumbs**
Only add items with a non-empty URL (`$href`) to the breadcrumbs (except root page).

This way, if some page types will not generate a proper URL (after an exception or for pages with no public route in the future), they will neither get added to the breadcrumbs nor to the JSON-LD data.

**4. Frontend preview**
If a `RouteParametersException` is thrown for a page when converting the frontend preview URL and the reason for the exception is an `InvalidParameterException` on a page with `tl_page.requireItem`, catch the exception and do not set a URL for the preview, thus starting the preview on the root page.

### DIFF
--- a/core-bundle/src/EventListener/PreviewUrlConvertListener.php
+++ b/core-bundle/src/EventListener/PreviewUrlConvertListener.php
@@ -70,7 +70,7 @@ class PreviewUrlConvertListener
             } catch (RouteParametersException $e) {
                 $route = $e->getRoute();
 
-                if (!$e->getPrevious() instanceof InvalidParameterException || !$route instanceof PageRoute || !$route->getPageModel()->requireItem) {
+                if (!$route instanceof PageRoute || !$route->getPageModel()->requireItem) {
                     $event->setUrl($this->getFragmentUrl($request, $page));
                 }
 
--- a/core-bundle/src/EventListener/PreviewUrlConvertListener.php
+++ b/core-bundle/src/EventListener/PreviewUrlConvertListener.php
@@ -24,7 +24,6 @@ use Symfony\Component\HttpKernel\Controller\ControllerReference;
 use Symfony\Component\HttpKernel\Fragment\FragmentUriGenerator;
 use Symfony\Component\HttpKernel\UriSigner;
 use Symfony\Component\Routing\Exception\ExceptionInterface;
-use Symfony\Component\Routing\Exception\InvalidParameterException;
 
 /**
  * @internal
--- a/core-bundle/src/EventListener/PreviewUrlConvertListener.php
+++ b/core-bundle/src/EventListener/PreviewUrlConvertListener.php
@@ -73,7 +73,7 @@ class PreviewUrlConvertListener
                     $event->setUrl($this->getFragmentUrl($request, $page));
                 }
 
-                // Swallow exception and set no url for pages with requireItem (see #3525)
+                // Ignore the exception and set no URL for pages with requireItem (see #3525)
             } catch (ExceptionInterface $e) {
                 $event->setUrl($this->getFragmentUrl($request, $page));
             }
--- a/core-bundle/src/EventListener/PreviewUrlConvertListener.php
+++ b/core-bundle/src/EventListener/PreviewUrlConvertListener.php
@@ -14,14 +14,17 @@ namespace Contao\CoreBundle\EventListener;
 
 use Contao\ArticleModel;
 use Contao\CoreBundle\Event\PreviewUrlConvertEvent;
+use Contao\CoreBundle\Exception\RouteParametersException;
 use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\CoreBundle\Routing\Page\PageRegistry;
+use Contao\CoreBundle\Routing\Page\PageRoute;
 use Contao\PageModel;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Controller\ControllerReference;
 use Symfony\Component\HttpKernel\Fragment\FragmentUriGenerator;
 use Symfony\Component\HttpKernel\UriSigner;
 use Symfony\Component\Routing\Exception\ExceptionInterface;
+use Symfony\Component\Routing\Exception\InvalidParameterException;
 
 /**
  * @internal
@@ -64,6 +67,14 @@ class PreviewUrlConvertListener
 
             try {
                 $event->setUrl($page->getPreviewUrl($this->getParams($request)));
+            } catch (RouteParametersException $e) {
+                $route = $e->getRoute();
+
+                if (!$e->getPrevious() instanceof InvalidParameterException || !$route instanceof PageRoute || !$route->getPageModel()->requireItem) {
+                    $event->setUrl($this->getFragmentUrl($request, $page));
+                }
+
+                // Swallow exception and set no url for pages with requireItem (see #3525)
             } catch (ExceptionInterface $e) {
                 $event->setUrl($this->getFragmentUrl($request, $page));
             }
--- a/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
@@ -96,8 +96,8 @@ class ModuleBreadcrumb extends Module
 
 		for ($i=(\count($pages)-1); $i>0; $i--)
 		{
-			// Skip pages that require an item (see #3450)
-			if (($pages[$i]->hide && !$this->showHidden) || $pages[$i]->requireItem || (!$pages[$i]->published && !$blnShowUnpublished))
+			// Skip pages that require an item (see #3450) and hidden or unpublished pages
+			if ($pages[$i]->requireItem || ($pages[$i]->hide && !$this->showHidden) || (!$pages[$i]->published && !$blnShowUnpublished))
 			{
 				continue;
 			}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by SeverinGloeckle** [comment]
There is still one case unhandled: If a page with `tl_page.requireItem=1` is visible in any navigation, the `RouteParametersException` is thrown, but handled within `Module::renderNavigation` resulting in the page not being added to the navigation and an error message written to the system log.

In my opinion, there is no need to explicitly skip pages with `tl_page.requireItem=1` in `Module::renderNavigation` since the current handling seems valid. To get rid of the error messages, an editor should flag these pages with `tl_page.hide`.

However, maybe the option / checkbox `tl_page.hide` should be automatically activated (and assigned with the `disabled` attribute) when a page is saved with `tl_page.requireItem=1`. What do you think?

File: core-bundle/src/EventListener/PreviewUrlConvertListener.php [hunk 1]
```diff
@@ -55,7 +58,17 @@ public function __invoke(PreviewUrlConvertEvent $event): void
                 return;
             }
 
-            $event->setUrl($page->getPreviewUrl($this->getParams($request)));
+            try {
+                $event->setUrl($page->getPreviewUrl($this->getParams($request)));
+            } catch (RouteParametersException $e) {
+                $route = $e->getRoute();
+
+                if (!$e->getPrevious() instanceof InvalidParameterException || !$route instanceof PageRoute || !$route->getPageModel()->requireItem) {
```
SeverinGloeckle: This is debatable. 

Swallowing any `RouteParametersException` should definitely be too loose since it catches exceptions like `RouteNotFoundException` as well.

However, swallowing any `InvalidParameterException` could be a more useful default here.
Maybe @aschempp  can share his opinion.

leofeyer: I think catching the `InvalidParameterException` is correct here. Not 100% sure though.

core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
```diff
@@ -199,7 +204,8 @@ protected function compile()
 			(
 				'isRoot'   => false,
 				'isActive' => true,
-				'href'     => $this->getPageFrontendUrl($pages[0]),
+				// Use the current request without query string for the current page (see #3450)
+				'href'     => strtok(Environment::get('request'), '?'),
```
leofeyer, line 207: We should use the Symfony request here. `$request->getPathInfo()` might already have the path without the query string.

SeverinGloeckle: It would be nice to use the Symfony request here.

However, I tried to keep the generated URL in line with the rest of the URLs in the breadcrumbs:
- The regular URLs contain the script name (e.g. `preview.php` or `app.php`) while `getPathInfo()` does not.
- The regular URLs are relative paths while `getPathInfo()` generates an absolute path reference (`page/subpage.html` vs. `/page/subpage.html`).

Whether the regular URLs are perfect surely is another good question.

I'm not quite convinced that manually building a fitting URL (trim leading `/`, add script name) is worth using the Symfony request:

````php
ltrim($request->getScriptName() . $request->getPathInfo(), '/')

// or

strtok(Environment::get('request'), '?')
````

What do you prefer?

leofeyer: I guess you are right. Let's stick with `strtok(Environment::get('request'), '?')`.

core-bundle/src/EventListener/PreviewUrlConvertListener.php
aschempp: I think we can safely swallow any routing exception. If we can't generate the URL, there's just no URL for that page in the frontend, so the preview event cannot generate one. I don't think it ever makes sense to display an exception message?

SeverinGloeckle: Since the `PreviewUrlListener` only handles the preview for pages, we could get away with swallowing any `RouteParametersException`. 

However, I think this could lead to some hard to debug issues when generating the frontend preview for custom pages (no preview URL gets generated and the developer is left without an exception/trace on what went wrong). 

That's why I decided to only swallow the specific case for pages with `tl_page.requireItem` since this could occur in a correctly configured standard system.

fritzmg:
Related: #3605

SeverinGloeckle:
Thanks for the update @fritzmg!

This pull request fixes the handling of pages with `tl_page.requireItem`. These pages are routable, but only with a proper parameter which is not available in the frontend preview and in the breadcrumbs module (hence this fix).

Your pull request will improve the handling of unroutable pages like error pages, which could also end up in a breadcrumbs trail.

Can we get this fix merged for Contao 4.12 so the adjustments from @fritzmg  may be used to also improve the breadcrumbs for Contao 4.13?

leofeyer:
@SeverinGloeckle @aschempp How do we proceed here?

aschempp (dismissed):
I have rebased the PR and I think this can finally be merged. Two things are changed:
 1. the breadcrumb menu no longer shows pages if no URL can be generated
 2. pages with `requireItem` no longer have a preview link. The icon in the page list is no longer clickable. If the page URL (`...preview?page=17`) is called, it now redirects to the home page (fallback for any "unknown" preview link) instead of throwing an exception.

leofeyer (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3544, opened 2021-10-04
Only add regular pages to the sitemap

### LINKED ISSUE #3415: Improve the error page handling in the sitemap controller
**Affected version(s)**

Contao 4.12

**Description**

Currently the `/sitemap.xml` route also shows URLs to the 404, 403 and 401 pages (which then the Contao Crawler for example will complain about, since they do not respond with 200).

The `supportsContentComposition` check is insufficient since these page types do support content composition, but should never show up in a sitemap (nor should they be accessible via their alias at all in my opinion). @aschempp any idea how to filter page types for this purpose in a general way? The old `Backend::findSearchablePages` only returned pages with type `regular`, but that is not sufficient due to custom page types / page controllers. Should we filter the page types based on a hard coded list (or a service parameter)?

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3415
| Docs PR or issue | -

This is the temporary fix for #3415 as discussed in the last Contao call.

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -163,6 +163,11 @@ class SitemapController extends AbstractController
             $result[] = $this->getPageAndArticleUrls((int) $pageModel->id);
         }
 
+        // Backwards compatibility for PHP 7.3
+        if (empty($result)) {
+            return [];
+        }
+
         return array_merge(...$result);
     }
 }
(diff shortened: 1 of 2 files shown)


aschempp (changes requested):
the discussion was to skip everything except `regular` pages like we previously did when generating sitemaps.

fritzmg:
Indeed. We discussed that, while this might be overly restrictive initially, users that implemented custom page controllers (or custom page types in general) can still inject their pages' URLs via the sitemap event. Until we implement the proposed solution where every page is skipped for which an URL cannot be generated.

leofeyer:
My bad. Changed in 042584593230d20d78c502c2b34a1ac2233091b7.

fritzmg (approved):

PR #3548, opened 2021-10-05
Fix pagination item calculation for an even number of links

### LINKED ISSUE #3539: Error when calculating maxPaginationLinks if number is even
the default setting of maxPaginationLinks with 7 is an odd number - here the output is correct and we see 7 links

you can change the setting (Contao 4.9.20) e.g.
```
contao:
  localconfig:
    maxPaginationLinks: 4
```
then the display is as follows

FireShot Pro Webpage Screenshot #618 - 'News   Newsletter - GUTcert' - www gut-cert de

so one too many

with maxPaginationLinks 2 then 3 or maxPaginationLinks 8 then 9

I do not understand the calculation at this point :(
https://github.com/contao/contao/blob/dccba73da017cdffe74d0604ec67e6f8d4fbf137/core-bundle/src/Resources/contao/library/Contao/Pagination.php#L333

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3539 

Improve the calculation of pagination links to ensure the pagination never has more items than defined by the `$intNumberOfLinks` parameter (usually the `maxPaginationLinks` configuration value). Add one more link after the current page if an even distribution of links is not possible.

Three notes:
1. To make the `Pagination` class testable, the calculation of `$this->intTotalPages` has been moved from `Pagination::generate` to its constructor. This is possible since its calculated value never changes.

2. `Pagination::getItemAsArray` now returns an empty array if no pagination is required (total pages < 2). Previously an array with a single item would have been generated – but the method was never executed since `Pagination::generate` already returned early in this case. 

3. For Contao 4.10+, the annotations `@runTestsInSeparateProcesses` and `@preserveGlobalState disabled` as well as the `include_once` of `functions.php` can be removed from `PaginationTest`. For Contao 4.9, they are necessary since the pagination relies on the `ampersand` function (which was moved to `StringUtil` in Contao 4.10).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Pagination.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Pagination.php
@@ -135,6 +135,7 @@ class Pagination
 		$this->intRows = (int) $intRows;
 		$this->intRowsPerPage = (int) $intPerPage;
 		$this->intNumberOfLinks = (int) $intNumberOfLinks;
+		$this->intTotalPages = $this->intRows > 0 && $this->intRowsPerPage > 0 ? (int) ceil($this->intRows / $this->intRowsPerPage) : 0;
 
 		// Initialize default labels
 		$this->lblFirst = $GLOBALS['TL_LANG']['MSC']['first'];
@@ -227,7 +228,6 @@ class Pagination
 		}
 
 		$this->strVarConnector = $blnQuery ? '&amp;' : '?';
-		$this->intTotalPages = ceil($this->intRows / $this->intRowsPerPage);
 
 		// Return if there is only one page
 		if ($this->intTotalPages < 2 || $this->intRows < 1)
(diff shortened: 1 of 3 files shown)


aschempp (dismissed):

leofeyer:
To make testing and upstream merging easier, I would backport the `StringUtil::ampersand()` method (see https://github.com/SeverinGloeckle/contao/pull/1).

SeverinGloeckle:
Backporting the `StringUtil::ampersand()` is a great way to keep the unit test simpler, very nice!

The one failing test for PHP 7.3 is a time-sensitive test that failed by unfortunate coincidence.

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3552, opened 2021-10-06
Always set the _locale attribute from the page language

aschempp:
see https://github.com/contao/contao/pull/3506#pullrequestreview-771748070

FYI the `xx` is the official short code for an unknown language. The tests now also actually test the locale is correctly converted 😬


fritzmg (dismissed):

leofeyer (changes requested):
Please also fix the `Route404Provider`:

https://github.com/contao/contao/blob/9dea31c07577b926547e23a464de1b5f8fb04c92/core-bundle/src/Routing/Route404Provider.php#L164-L166

leofeyer (approved):

PR #3553, opened 2021-10-06
Enable parallel execution in ECS

leofeyer:
See https://github.com/contao/easy-coding-standard/issues/11


Toflar:
Is it faster? :)

leofeyer:
Yes, I think so. Big files such as `DC_Table` are still slow of course, but overall 👍.

PR #3554, opened 2021-10-06
Adjust the MySQL strict mode explanation

Ernestopheles:
clearer

### DIFF
--- a/installation-bundle/src/Resources/translations/ContaoInstallationBundle.en.xlf
+++ b/installation-bundle/src/Resources/translations/ContaoInstallationBundle.en.xlf
@@ -270,7 +270,7 @@
         <source>Your database server is not running in strict mode!</source>
       </trans-unit>
       <trans-unit id="90" resname="strict_sql_mode_explain">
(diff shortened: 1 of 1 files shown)


m-vo:
I don't see the purpose why you changed the wording from

     Please enable the strict mode either in your my.cnf file or configure the connection options…

to

     Please enable the strict mode either in your my.cnf file or by configuring the connection options…

?

Ernestopheles:
In my version it becomes clear that traditional also brings strict Mode. Some were confused see slack. One might think traditional only is a not so good way to handle the situation.

Ernst
Am 7. Okt. 2021, 15:58 +0200 schrieb M. Vondano ***@***.***>:

leofeyer:
Not sure about this. Please keep in mind that changing the source file will automatically remove the translation in Transifex, too. Each translation team then has to translate the string again. Is this really worth the effort?

m-vo:
> One might think traditional only is a not so good way to handle the situation.

So rather?

     Please enable a strict MySQL mode (like `TRADITIONAL`) …

But I'm with @leofeyer - I'm not sure that's worth it.

fritzmg:
imho there is no need to change anything in Contao itself. However, I think we can add a `{{% notice "tip" %}}` to https://docs.contao.org/manual/de/installation/systemvoraussetzungen/#mysql-mindestanforderungen with a short explanation what this actually does and why it's called `TRADITIONAL`, or what the setting actually is, so that any confusion can be cleared up. I'll create an according PR.

fritzmg:
See https://github.com/contao/docs/pull/856

PR #3560, opened 2021-10-07
Harden the template locator against failing DB connections

### LINKED ISSUE #3557: TemplateLocator queries database even if not fully installed
**Affected version(s)**

Contao 4.12

**Description**

The `TemplateLocator::findThemeDirectories()` tries to find all themes even if the Contao installation has not been fully installed.

You can reproduce this with `composer create-project contao/managed-edition contao 4.12` on the command line.

@m-vo How do we fix this? By catching more exceptions in the try-catch block?

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3557
| Docs PR or issue | -

In this PR we're now also catching the `ConnectionException` so that failing database connections won't throw anymore.

### DIFF
--- a/core-bundle/src/Twig/Loader/TemplateLocator.php
+++ b/core-bundle/src/Twig/Loader/TemplateLocator.php
@@ -15,6 +15,7 @@ namespace Contao\CoreBundle\Twig\Loader;
 use Contao\CoreBundle\Exception\InvalidThemePathException;
 use Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception\ConnectionException;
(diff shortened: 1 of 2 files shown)


aschempp:
Are there any DBAL exceptions you don't want to catch? Or why not catch the generic one?

m-vo:
> Are there any DBAL exceptions you don't want to catch? Or why not catch the generic one?

I thought it would be a bad idea to catch things like `DeadlockException`, `ConstraintViolationException`, …

~But I figured the error in #3557 will also appear if there *is* a database configured (just not with a `.parameters` file). We need to fix this problem as well, I guess, because otherwise the theme directories won't be added to the hierarchy.~

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3563, opened 2021-10-08
Correctly filter the root nodes in the extended tree view

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3391
| Docs PR or issue | -

I have noticed that the changes from #3391 make some articles disappear:

Here is how it should look:

@Toflar Can you please confirm that the fix is correct? And could we not omit the check entirely for admins?

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -83,7 +83,7 @@ $GLOBALS['TL_DCA']['tl_page'] = array
 		'sorting' => array
 		(
 			'mode'                    => 5,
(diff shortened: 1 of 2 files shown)


Toflar (changes requested):
Changes are not correct, will check after my vacation :)

Toflar:
So imho the problem here is that this also filters on `tl_article` IDs whereas it should only on `tl_page`. Is that correct?

leofeyer:
Yes.

Toflar:
Then I think https://github.com/contao/contao/pull/3596 is the correct fix.

PR #3574, opened 2021-10-13
Correctly format the locale for legacy requests

aschempp:
When parsing a path like `de-DE/foo/bar`, the `$locale` variable is used to re-build a path from hook fragments. If a request can be matched, the `_locale` parameter is incorrectly formatted and re-builds the path as `de_DE/foo/bar` without this fix.


fritzmg (approved):

hh-com:
**Use case:**
Created a fresh Contao installation with the following sitemap:

pagetree

Without using the Hook getPageIdFromUrl everything works fine.

After activating the Hook i got the 404 Error "No route found for GET...." on the Subsites from "de-DE" and "de-AT".
AT and DE works fine.

Return from the Hook ($arrFragments):
hookreturn


PR live-test seems to work!

PR #3577, opened 2021-10-14
Handle invalid accept-language headers

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

If the accept language header contains an invalid locale, e.g. `en,en-US;q=0.9,pt-BR;q=0.8,pt;q=0.7,und;q=0.6`, `Request::getLanguages()` will return:

```php
[
  0 => "de"
  1 => "de_DE"
  2 => "pt_BR"
  3 => "pt"
  4 => "und"
]
```

This array will be passed to our `AbstractPageRouteProvider::convertLanguagesForSorting()` method and the method then calls `LocaleUtil::getFallbacks()` with every given locale. The latter returns an empty array for `und`, which triggers an `Warning: Undefined array key "language"` warning here:

https://github.com/contao/contao/blob/cbead7bbe42c47206f05cd237ee25fc79140d0a1/core-bundle/src/Util/LocaleUtil.php#L59-L60

The `$result` variable of the `convertLanguagesForSorting()` method is then compiled as:

```php
[
  0 => "de"
  1 => "de_DE"
  2 => "pt_BR"
  3 => "pt"
  4 => "pt"
  5 => null
]
```

This triggers a `Warning: array_flip(): Can only flip string and integer values, entry skipped` warning here:

https://github.com/contao/contao/blob/cbead7bbe42c47206f05cd237ee25fc79140d0a1/core-bundle/src/Routing/AbstractPageRouteProvider.php#L216

The PR should fix both, although I am not 100% sure if the changes are correct. @ausi /cc

### DIFF
--- a/core-bundle/src/Util/LocaleUtil.php
+++ b/core-bundle/src/Util/LocaleUtil.php
@@ -61,6 +61,8 @@ class LocaleUtil
 
         if (isset($data[\Locale::LANG_TAG])) {
             $result[] = $data[\Locale::LANG_TAG];
+        } else {
+            $data[\Locale::LANG_TAG] = '';
         }
 
         if (isset($data[\Locale::REGION_TAG])) {
--- a/core-bundle/src/Util/LocaleUtil.php
+++ b/core-bundle/src/Util/LocaleUtil.php
@@ -56,6 +56,10 @@ class LocaleUtil
             return self::$fallbacks[$locale];
         }
 
+        if ('' === $locale) {
+            return [];
+        }
+
         $result = [];
         $data = \Locale::parseLocale($locale);
 
--- a/core-bundle/src/Util/LocaleUtil.php
+++ b/core-bundle/src/Util/LocaleUtil.php
@@ -56,8 +56,12 @@ class LocaleUtil
             return self::$fallbacks[$locale];
         }
 
+        $result = [];
         $data = \Locale::parseLocale($locale);
-        $result = [$data[\Locale::LANG_TAG]];
+
+        if (isset($data[\Locale::LANG_TAG])) {
+            $result[] = $data[\Locale::LANG_TAG];
+        }
 
         if (isset($data[\Locale::REGION_TAG])) {
             $result[] = $data[\Locale::LANG_TAG].'_'.$data[\Locale::REGION_TAG];
--- a/core-bundle/src/Routing/AbstractPageRouteProvider.php
+++ b/core-bundle/src/Routing/AbstractPageRouteProvider.php
@@ -202,7 +202,10 @@ abstract class AbstractPageRouteProvider implements RouteProviderInterface
         $result = [];
 
         foreach ($languages as $language) {
-            $locales = LocaleUtil::getFallbacks($language);
+            if (!$locales = LocaleUtil::getFallbacks($language)) {
+                continue;
+            }
+
             $language = array_pop($locales);
             $result[] = $language;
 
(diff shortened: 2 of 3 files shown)


core-bundle/src/Util/LocaleUtil.php
```diff
@@ -56,8 +56,12 @@ public static function getFallbacks(string $locale): array
             return self::$fallbacks[$locale];
         }
 
+        $result = [];
         $data = \Locale::parseLocale($locale);
```
ausi, line 64:
```suggestion
        $data = \Locale::parseLocale($locale) + [\Locale::LANG_TAG => 'und'];
```

ausi (commented):
I think we should fallback to the `und` language instead.

The current state of this pull request still fails for languages like `_DE` I believe.

core-bundle/src/Util/LocaleUtil.php
leofeyer: This would set an invalid language as default, wouldn't it?

ausi: No, `und` is the “undetermined” language. So `und_DE` and `_DE` are semantically the same AFAIK.

ausi: My suggestion is wrong I think now. See https://github.com/contao/contao/pull/3577#discussion_r729130414

core-bundle/src/Util/LocaleUtil.php
```diff
@@ -56,8 +56,12 @@ public static function getFallbacks(string $locale): array
...
-        $result = [$data[\Locale::LANG_TAG]];
+
+        if (isset($data[\Locale::LANG_TAG])) {
+            $result[] = $data[\Locale::LANG_TAG];
+        }
```
ausi, line 70: Instead we should add this here:
```suggestion
        } else {
            $data[\Locale::LANG_TAG] = '';
        }
```

core-bundle/src/Util/LocaleUtil.php
```diff
@@ -56,8 +56,12 @@ public static function getFallbacks(string $locale): array
             return self::$fallbacks[$locale];
         }
```
ausi, line 58:
```suggestion

        if ('' === $locale) {
            return [];
        }

```

ausi (commented):
Added unit tests in b657c2905832799b5faf53fea71e7acaaea9ea8b with the code suggestions added these would no longer fail.

core-bundle/src/Util/LocaleUtil.php
leofeyer: I initially had it this way, however, the result will not be added to `self::$fallbacks[$locale]` then.

core-bundle/src/Util/LocaleUtil.php
leofeyer: You mean like this?

```php
$data = \Locale::parseLocale($locale);

if (!isset($data[\Locale::LANG_TAG])) {
    $data[\Locale::LANG_TAG] = '';
}

$result = [$data[\Locale::LANG_TAG]];
```

ausi: No, if `$data[\Locale::LANG_TAG]` is not set we should not add the empty string to the result array, but we need to set `$data[\Locale::LANG_TAG] = ''` in order for the other candidates to work correctly like `_DE`.

core-bundle/src/Util/LocaleUtil.php
ausi: The empty string has a special meaning for locales so the alternative to this code would be to use something like `$data = \Locale::parseLocale('' === $locale ? 'und' : $locale);` two lines below.
Both solutions would be fine for me.

ausi (approved):

leofeyer:
Thank you @ausi.

PR #3594, opened 2021-10-21
Implemented contao:database:dump and contao:database:import

Toflar:
Merging the PRs https://github.com/contao/contao/pull/3518 and https://github.com/contao/contao/pull/3466 here as they belong together and can then be functional tested against each other.
Repeating my comments here for consistency:

## contao:database:dump

I've researched the use on the dump library (or none at all) quite a bit before going for `webfactory/slimdump`.
Here are the reasons for my decision

* Requiring the `mysqldump` command line tool seemed like an option but this has 2 serious disadvantages:
    1. We all know, not everybody has access to `mysqldump` with their hosting provider
    2. Actually, even if they had access, we could have an issue with making the dump process compatible with the import process as we have no control over the version of `mysqldump` installed and thus, we might run into compatibility issues.
* https://packagist.org/packages/ifsnop/mysqldump-php is the most popular library but there does not seem to be a lot of activity. Especially PRs regarding PHP 8 are being ignored.
* https://packagist.org/packages/dg/mysql-dump seems popular too but uses `mysqli` directly and thus does not seem to be the right fit for us as we use Doctrine/PDO.
* https://packagist.org/packages/rah/danpu is also very popular but kind of has the same issue as it requires to pass the DSN.
* I then decided for https://github.com/webfactory/slimdump because it seems maintained, I like the code, it's based on Doctrine and I know at least one maintainer which is also active in the Symfony environment :) 

## contao:database:import

In our onging quest to get rid of the installation bundle at some point, there's only the import DB command missing which this PR should implement.

PS: Not sure if this is really the last missing piece since @m-vo added the db strict mode stuff to the installation bundle again in the latest versions so maybe we also need this information as a command.

To do:

- [x] Add a dump to `contao:migrate` by default
- [ ] Add support for Doctrine DBAL 3.0 to slimdump (https://github.com/webfactory/slimdump/pull/83)
- [ ] Tests


Toflar:
@contao/developers this is now ready for a basic review of the concept. Remarks:

* Both, the import as well as the dump command share common settings and use the same service which I named `Dumper`. I did not want to name it `DumpOrImportHandler` or anything like that. That service is responsible for both, importing and dumping.
* I've added a check so that only dumps that have been generated with our dump command can be used for the import command. I went with an internal version on top of it which we only ever need to change if we make incompatible changes which I cannot think of just yet but who knows.
* I deliberately went with `var/db_dump.sql.gz` as default target and did not include any time information in order to prevent flooding the `var` directory with backups.
* `contao:migrate` now by default creates a backup with the default settings as well. You can skip it by passing `--no-backup`. If you want to have a backup with different options than the default ones. Just run `contao:database:dump` with your custom options first and then `contao:migrate --no-backup`.

aschempp:
> I've added a check so that only dumps that have been generated with our dump command can be used for the import command. I went with an internal version on top of it which we only ever need to change if we make incompatible changes which I cannot think of just yet but who knows.

That sounds problematic to me, how do I import an existing theme SQL file or a dump created from a phpMyAdmin I have on a remote system?

core-bundle/src/Command/AbstractDatabaseCommand.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this
+            ->addArgument('file', InputArgument::OPTIONAL, 'The path to the SQL dump.')
+            ->addOption('ignore-tables', 'i', InputOption::VALUE_OPTIONAL, 'A comma-separated list of database tables to ignore. Defaults to the Contao configuration (contao.db.dump.ignoreTables).')
+        ;
+    }
+
+    protected function handleCommonConfig(InputInterface $input, AbstractConfig $config): AbstractConfig
+    {
+        if ($file = $input->getArgument('file')) {
```
aschempp: feels like this should be an option `--config` (and `-c` short name similar to PHP) or the argument should at least be called `config-file`

Toflar:
> That sounds problematic to me, how do I import an existing theme SQL file or a dump created from a phpMyAdmin I have on a remote system?

That's not the scope of this PR. This PR provides ways to dump a database in a given format and import from that format. Any other dumps are simply not supported.

core-bundle/src/Command/DatabaseDumpCommand.php
```diff
@@ -0,0 +1,94 @@
...
+            $io->error($e->getMessage());
+
+            return 1;
+        }
+
+        $io->success(
+            sprintf(
+                'Successfully created an SQL dump at "%s" while ignoring following tables: %s.',
+                $config->getFilePath(),
+                implode(', ', $config->getTablesToIgnore())
```
aschempp: what if I didn't set any tables to be ignored?

core-bundle/src/Command/DatabaseImportCommand.php
```diff
@@ -0,0 +1,63 @@
...
+class DatabaseImportCommand extends AbstractDatabaseCommand
+{
+    protected static $defaultName = 'contao:database:import';
+
+    protected function configure(): void
+    {
+        parent::configure();
+
+        $this
+            ->addOption('truncate', 't', InputOption::VALUE_NONE, 'Truncate the existing database')
```
aschempp: `truncate` usually refers to emptying a database table, and not deleting all tables inside a database. Maybe we can come up with a better name (`drop-tables` maybe). If I understand correctly that it actually deletes all tables (to re-add them from schema) and not truncates them?

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -57,6 +60,7 @@ protected function configure(): void
             ->addOption('migrations-only', null, InputOption::VALUE_NONE, 'Only execute the migrations.')
             ->addOption('dry-run', null, InputOption::VALUE_NONE, 'Show pending migrations and schema updates without executing them.')
             ->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format (txt, ndjson)', 'txt')
+            ->addOption('no-backup', null, InputOption::VALUE_NONE, 'Disable the database backup which is created by default before executing the migrations.')
```
aschempp, line 63: I would actually prefer a `--backup=` option where I can specify where to backup instead of being forced to disable the option 🤷

core-bundle/src/Command/AbstractDatabaseCommand.php
Toflar: It's not a config file....it's the dump.

core-bundle/src/Command/DatabaseDumpCommand.php
Toflar: Then the default ones are used. Configuring it to an empty array would then just output nothing.

core-bundle/src/Command/DatabaseImportCommand.php
Toflar: Yeah...it truncates and does not delete. That's why it's named truncate.

core-bundle/src/Command/MigrateCommand.php
Toflar: We agreed on having it enabled by default.

mpdude:
👋🏻 One of the maintainers for webfactory/slimdump here.

It's a surprising and unexpected honor to be chosen for an integration like this 😉. Thank you for your trust, I hope we can live up to your expectations.

A word of ⚠️ caution, though: `slimdump` was written as a CLI tool, not as a library. We do not have an explicit policy stating what comprises our API. But from what I can tell, the general mindset of everybody who worked on it is that as long as you can use the same configuration files and command line switches as before, internal (potentially huge) refactorings of the code base _do not_ warrant corresponding (semver) version number bumps and/or deprecation provisions.

Toflar:
Hey Matthias, I'll get back to your comment soon :-)
Notes to myself from the developers meeting:

- We cannot use this to replace the theme template import from the installation-bundle.
- [x] Allow to disable the check for the dump version check
- [x] `contao:backup:create`, `contao:backup:restore`, `contao:backup:list` (make sure we have `--json` as option)
- [x] `create` and `restore` have to support file paths somewhere else
- [x] Configurable `var/backups` directory
- [x] Configurable retention number (default: 5)

Toflar:
I've closed this in favor of https://github.com/contao/contao/pull/3631 as we went from a simple dump command to a fully blown backup management solution.

@mpdude I kind of expected that. I've checked quite a few libraries and none of them seemed to fit our needs. For us, there's always the option to pin it to a certain version or extract/copy the logic. Maybe you're interested in making it a library though? :) You already have some benefits now by having me working on DBAL 3 compatiblity, maybe there would be even more traction if everything but the commands were extracted into a library on your part?
You know where you can find me if you want to talk about that :)

PR #3595, opened 2021-10-21
Fix the base path in the asset context

### LINKED ISSUE #3448: Make Template::asset static
Is there any particular reason why methods like `generateStyleTag`, `generateInlineStyle` etc. are static but the method `asset` is not?

Currently I'm building a form widget where I seem to have no access to an instance of `Template` or rather `FrontendTemplate`. In order to make use of the `asset` method I need to create a new `FrontendTemplate` instance.

Adding the assets from within the template, rather than adding them to `TL_HEAD` or `TL_JAVASCRIPT`, wouldn't be a clean solution in this case because a form could have the same type of widget multiple times which would result in multiple loads of the same asset.

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3448


m-vo:
Memo to myself: Is `getStaticUrl()` unaffected by this or are we using it correctly in https://github.com/contao/contao/blob/4.x/core-bundle/src/Image/Studio/ImageResult.php#L230? (/cc @ausi)

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -400,6 +401,12 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
...
 		$url = System::getContainer()->get('assets.packages')->getUrl($path, $packageName);
+		$request = System::getContainer()->get('request_stack')->getMainRequest();
+
+		if ($request !== null)
+		{
+			return Url::makeRelative($url, $request->getBasePath());
+		}
 
 		// Contao paths are relative to the <base> tag, so remove leading slashes
 		return ltrim($url, '/');
```
leofeyer: Is this still required now that we are using `Url::makeRelative()`?

aschempp: It is only applied if there is no request (e.g. on the command line).

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -400,6 +401,12 @@ public function trans($strId, array $arrParams=array(), $strDomain='contao_defau
 	public function asset($path, $packageName = null)
 	{
 		$url = System::getContainer()->get('assets.packages')->getUrl($path, $packageName);
+		$request = System::getContainer()->get('request_stack')->getMainRequest();
```
bennyborn: I guess that should also say `getMasterRequest`

leofeyer (dismissed):

bennyborn:
Still not quite working for me

image

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -401,8 +401,21 @@ public function asset($path, $packageName = null)
...
+
+		if ($request !== null)
+		{
+			$basePath = $request->getBasePath() . '/';
+		}
+
+		if (0 === strncmp($url, $basePath, \strlen($basePath)))
+		{
+			return substr($url, \strlen($basePath));
+		}
```
fritzmg, line 415: I'll test if this will work under Windows.

fritzmg (changes requested):

core-bundle/src/Resources/contao/library/Contao/Template.php
aschempp: i think it should since it is a URL and not a file path, but please test it

fritzmg: Yep, looks fine. `('assets.packages')->getUrl('js/swipe.min.js')` will return 

```
/subfolder/web/assets/js/swipe.min.js
```

(when Contao is accessed via `http://localhost/subfolder/web` for example) and `$basePath` will be 

```
/subfolder/web/
```

and thus the condition will be true and `assets/swipe/js/swipe.min.js` is returned 👍

fritzmg (approved):

fritzmg:
> Still not quite working for me

@bennyborn should be fine now, can you test again? :)

bennyborn:
Yep, works now! 🎉

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -401,8 +401,21 @@ public function asset($path, $packageName = null)
...
+
 		// Contao paths are relative to the <base> tag, so remove leading slashes
-		return ltrim($url, '/');
+		return $url;
```
leofeyer, line 418: So now we do not need the `ltrim()` anymore? Haven't you just explained to me that it is still required if there is no request? https://github.com/contao/contao/pull/3595/files/46f536900cda60b46dc8cd8a7e993d75919a8fc8#r733842318

aschempp: `$basePath` is `/` by default so will be stripped if available. The unit tests confirm that 😉

Issue #3263, opened 2021-08-04
Make sure Template#getResponse() is used correctly


m-vo:
**Affected version(s)**

Contao 4.9 and up

**Description**

see https://github.com/contao/contao/pull/3258#issuecomment-892492357

Currently, `Template#getResponse()` sets headers and executes `compile()` which will make the following code run (that is only meant to run once for `fe_page` or any other 'global wrapper' template):
https://github.com/contao/contao/blob/4cf71a9b3cf787093c9cf9e076d9c1fc6104f2f5/core-bundle/src/Resources/contao/classes/FrontendTemplate.php#L102-L159

- Should users in fragment controllers use `$template->getResponse()` or `new Response($template->parse())`?
- If so: Should we move things out to listeners?

m-vo:
taken from https://github.com/contao/contao/pull/3258#issuecomment-892493203:

> If we are setting cache headers, why not in the `AbstractFragmentController`? 
> 
> There are multiple ways how users will return a response in a fragment's `getReponse()` method:
> 
> - `return $template->getResponse();`
> - `return new Response('foo');`
> - `return $this->render('foo.html.twig');`
>
> For the last two cases I would have to set headers myself while the first one would set them for me?

leofeyer:
I guess they should use `new Response($template->parse())`.

fritzmg:
> I guess they should use `new Response($template->parse())`.

But then the appropriate response headers would be missing if the fragment is rendered as ESI.

m-vo:
> But then the appropriate response headers would be missing if the fragment is rendered as ESI.

They are missing anyway if I don't use Contao templates. Can't we add the headers in the controller?

fritzmg:
True. The only header that makes sense to me that maybe should be added by the template itself is `Content-Type`, since the template knows its content type. But even that depends on `Config::get('characterSet')` currently - which is always `UTF-8`.

m-vo:
I think there are two places where we should split/reorganize responsibilities. Here are some thoughts:

### Template#getResponse()

Here is how this could potentially look when dissected:

* `Template#getResponse()` is simplified to only set the content, content-type and charset:

    ```php
    public function getResponse()
    {
    	$response = new Response($this->compile());
    	$response->headers->set('Content-Type', $this->strContentType);
    	$response->setCharset(System::getContainer()->getParameter('kernel.charset'));  // drop this as its's always UTF-8?
    
    	return $response;
    }
    ```
[...]

PR #3599, opened 2021-10-25
Move interest-cohort header to response listener

aschempp:
related to #3263


fritzmg (changes requested):
The service is not registered in the `listeners.yml` 🙃

bytehead (approved):
Looks good apart from missing test 😬

fritzmg (approved):

PR #3600, opened 2021-10-25
Introduce finalize method for Widget

### LINKED ISSUE #1185: Residual file via frontend upload when server-side form validation fails because of other field
**Affected version(s)**
4.4, 4.8

**Description**
If a file is uploaded via a frontend form, the upload is completed and the file is retained even when server-side form validation fails because of another field in the form.

**How to reproduce**
* Create a new form in the form generator.
* Create a file upload field. Check "Store uploaded files" and select a directory.
* Create a text field. For "Input validation", select "Alphabetic characters".
* Add a send field.
* In the frontend, select a valid file, but type `123` into the text field. Send the form.

**Result:** You get an error message for the "Alphabetic characters" field, but the file from the other field is uploaded anyway.

This produces a greater issue if you also select "Preserve existing files" in the file upload field settings: A user initially selects a file, but enters something invalid in another field. They send the form. Once they get the validation error, they enter something correct and re-send the form, again with the file. Their initially-validated and already-uploaded file is now uploaded a second time and given a suffix.

**Expected result:** Ideally, the file would not be uploaded or not be moved to the selected directory if the form does not pass validation.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1185
| Docs PR or issue | https://github.com/contao/docs/pull/862

This PR introduces a `finalize` method for (front end) widgets, which will be executed by the form generator after all widgets were validated. This way the `FormFileUpload` widget will be able to process the uploaded file and move it to its final destination only if the form was submitted successfully.

_Note:_ the `method_exists` call is necessary since any registered form widgets do not really have to extend from `\Contao\Widget`.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -755,7 +755,7 @@ abstract class Widget extends Controller
 		{
 			@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
 
-			if ($this instanceof FinalizableWidget)
+			if ($this instanceof FinalizableWidgetInterface)
 			{
 				$this->finalize();
 			}
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -753,10 +753,10 @@ abstract class Widget extends Controller
 
 		if ($blnFinalize)
 		{
-			@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
-
 			if ($this instanceof FinalizableWidgetInterface)
 			{
+				@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0. Use Widget::validate(false) and Widget::finalize() instead.', E_USER_DEPRECATED);
+
 				$this->finalize();
 			}
 		}
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -730,8 +730,10 @@ abstract class Widget extends Controller
 
 	/**
 	 * Validate the user input and set the value
+	 *
+	 * @param bool $blnFinalize Whether to finalize a finalizable widget during validate
 	 */
-	public function validate()
+	public function validate(/* $blnFinalize = true */)
 	{
 		$varValue = $this->validator($this->getPost($this->strName));
 
@@ -741,6 +743,23 @@ abstract class Widget extends Controller
 		}
 
 		$this->varValue = $varValue;
+
+		$blnFinalize = true;
+
+		if (\func_num_args() > 0)
+		{
+			$blnFinalize = func_get_arg(0);
+		}
+
+		if ($blnFinalize)
+		{
+			@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+
+			if ($this instanceof FinalizableWidget)
+			{
+				$this->finalize();
+			}
+		}
 	}
 
 	/**
--- a/core-bundle/src/Resources/contao/modules/ModuleChangePassword.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleChangePassword.php
@@ -149,7 +149,7 @@ class ModuleChangePassword extends Module
 			// Validate the widget
 			if (Input::post('FORM_SUBMIT') == $strFormId)
 			{
-				$objWidget->validate();
+				$objWidget->validate(false);
 
 				// Validate the old password
 				if ($strKey == 'oldPassword')
--- a/core-bundle/src/Resources/contao/modules/ModuleChangePassword.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleChangePassword.php
@@ -113,6 +113,7 @@ class ModuleChangePassword extends Module
 
 		/** @var FormPassword $objNewPassword */
 		$objNewPassword = null;
+		$arrWidgets = [];
 
 		// Initialize the widgets
 		foreach ($arrFields as $strKey=>$arrField)
@@ -169,6 +170,19 @@ class ModuleChangePassword extends Module
 			}
 
 			$strFields .= $objWidget->parse();
+			$arrWidgets[] = $objWidget;
+		}
+
+		// Finalize form widgets (#1185)
+		if (!$doNotSubmit && Input::post('FORM_SUBMIT') == $strFormId)
+		{
+			foreach ($arrWidgets as $objWidget)
+			{
+				if ($objWidget instanceof FinalizableWidgetInterface)
+				{
+					$objWidget->finalize();
+				}
+			}
 		}
 
 		$this->Template->fields = $strFields;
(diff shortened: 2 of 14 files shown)


SeverinGloeckle:
Great addition to core forms!

> _Note:_ the `method_exists` call is necessary since any registered form widgets do not really have to extend from `\Contao\Widget`.

In my opinion, an interface like `FinalizableWidget` would provide a more reliable way to check for the `finalize` method.

Since the pull requests targets Contao 4.9, it should be avoided to trigger any side effects (e.g. with a custom widget that randomly has the `finalize` method).

````php
if ($objWidget instanceof FinalizableWidget)
{
  ...
}
````

The toughest question then, however, is where such an interface for a legacy class should be defined?

**Comment by fritzmg** [comment]
Originally I did not want to introduce a new interface for the legacy framework. However we decided to use one now and I'll adjust the PR accordingly 👍

File: core-bundle/src/Resources/contao/helper/interface.php [hunk 1]
```diff
@@ -62,3 +62,13 @@ public function isActive();
 interface uploadable
 {
 }
+
+/**
+ * Interface finalizable
+ *
+ * @author Fritz Michael Gschwantner <https://github.com/fritzmg>
+ */
+interface finalizable
```
Toflar: Can we put that into the `Contao` namespace, please? I know you wanted to remain consistent with the rest but everytime we add something to the global namespace, we risk more conflicts with future PHP versions.
I know it's "legacy" but I would still just register this as `Contao\finalizable`.

ausi: Should we maybe store it as a regular autoloaded class in *src/Resources/contao/library/Contao/FinalizableWidget.php* instead?

Toflar (changes requested):
This is a BC break. Just leaving this here so it does not get merged as is :)

fritzmg:
To keep backwards compatibility we introduced an optional parameter for `Widget::validate()`. If no parameter or `true` is passed, then the widget still finalizes during `Widget::validate()`. This way the `FormFileUpload` widget with its new `FinalizableWidget` interface implementation will still move the file when calling `->validate()` on it, just as it did before. 

The `FormFileUpload` widget now does not implement `validate()` on its own anymore. It only implements `validator()` and `finalize()` now. The former validates the uploaded file while the latter stores the file to the session and moves the file, if applicable.

fritzmg:
Failing test ~~is~~ was unrelated.

fritzmg:
There are still places where Contao uses `$objWidget->validate()`, these still need to be adjusted as well. I am not sure anymore, if the deprecation is the correct route.

SeverinGloeckle:
Maybe it would be better to conceptually split the changes into two groups:

1. Add a `finalize` method.
2. Fix `FormFileUpload`.

Since the `finalize` method is opt-in, there is no need to run it automatically in `validate`.

This is only required as a fallback if logic gets moved within a widget from `validate` to `finalize`.

To prevent this, the current `FormFileUpload` should not be changed. Instead, a new `FinalizableFormFileUpload` could replace it in `$GLOBALS['TL_FFL']`. This way, standard forms with a regular upload element benefit from the fix while custom forms and widgets inheriting from `FormFileUplaod` are left unchanged.

On the other hand, one can also simply argue with _"A bugfix justifies breaking changes"_. In this case, the current implementation with a flag in `validate` seems to be a reasonable solution - however, as @Toflar mentioned, this can lead to BC breaks in individual cases.

fritzmg:
> To prevent this, the current `FormFileUpload` should not be changed. Instead, a new `FinalizableFormFileUpload` could replace it in `$GLOBALS['TL_FFL']`. This way, standard forms with a regular upload element benefit from the fix while custom forms and widgets inheriting from `FormFileUplaod` are left unchanged.

I am not sure about that. This would be a new widget class, just to fix a bug in the other one, which we will then have to support "forever".

SeverinGloeckle:
You are right. A new widget would be too much to just fix a bug in the upload widget. For the introduction of a new finalize feature, on the other hand, I think it is a possible way to avoid any effects on the existing upload element (and possible extensions of it).

The alternative is to introduce a BC break in some cases. Since your pull request already tries to minimize any problems with existing code, this should be fine for a new version like 4.13, together with a note in `UPGRADE.md`.

As a bug fix for 4.9, the changes look rather risky to me.

fritzmg:
@contao/developers wdyt? Should we only fix it in `4.13`?

fritzmg:
As discussed in the Contao call, we want to keep this solution - but only change it for Contao 4.13 and leave it as a known limitation in 4.9.

core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -741,6 +743,23 @@ public function validate()
...
+			$blnFinalize = func_get_arg(0);
+		}
+
+		if ($blnFinalize)
+		{
+			@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+
+			if ($this instanceof FinalizableWidgetInterface)
+			{
+				$this->finalize();
```
ausi: Should we only deprecate `validate()` (without `false`) for finalizable widgets?

```suggestion
			if ($this instanceof FinalizableWidgetInterface)
			{
				@trigger_error('Finalizing a widget during Widget::validate() has been deprecated and will no longer work in Contao 5.0. Use Widget::validate(false) and Widget::finalize() instead.', E_USER_DEPRECATED);

				$this->finalize();
```

core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php
```diff
@@ -91,6 +91,11 @@ protected function compile()
 			// Close account
 			if (!$objWidget->hasErrors())
 			{
+				if ($objWidget instanceof FinalizableWidgetInterface)
+				{
+					$objWidget->finalize();
+				}
+
```
ausi, line 98: If we use `finalize()` here, don’t we need to use `$objWidget->validate(false);` on line 80 then too?

Same problem with other `finalize()` calls in this pull request (except `Form`). Doesn’t the widget now get finalized twice?

ausi (changes requested):

core-bundle/src/Resources/contao/modules/ModuleCloseAccount.php
fritzmg: True, I missed that.

fritzmg: Should be fixed now.

ausi (approved):

core-bundle/src/Resources/contao/library/Contao/Widget.php
```diff
@@ -741,6 +743,23 @@ public function validate()
...
+		$blnFinalize = true;
+
+		if (\func_num_args() > 0)
+		{
+			$blnFinalize = func_get_arg(0);
+		}
+
+		if ($blnFinalize)
+		{
+			if ($this instanceof FinalizableWidgetInterface)
```
aschempp, line 756: this can be merged into one if condition

core-bundle/src/Resources/contao/modules/ModuleChangePassword.php
```diff
@@ -169,6 +170,19 @@ protected function compile()
 			}
 
 			$strFields .= $objWidget->parse();
+			$arrWidgets[] = $objWidget;
```
aschempp, line 173: how about we only collect the "finalizable" widgets to not need to loop over them again (applies to the form code as well)?

aschempp (changes requested):

fritzmg:
Closing in favor of #3734

PR #3601, opened 2021-10-25
Inline fragments should not compile the front end template

### LINKED ISSUE #3263: Make sure Template#getResponse() is used correctly

aschempp:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3263

This is a draft attempt to not handle inline fragments like page content. However, ESI fragments **should** actually behave like the main page controller, it **should** minify markup etc.

An alternative idea by @fritzmg and me would be to register a response listener that minifies the main response content etc., but we're not sure that's a good idea 🙃

### DIFF
--- a/core-bundle/src/Controller/AbstractFragmentController.php
+++ b/core-bundle/src/Controller/AbstractFragmentController.php
@@ -85,11 +85,11 @@ abstract class AbstractFragmentController extends AbstractController implements 
      */
     protected function createTemplate(Model $model, string $templateName): Template
     {
-        $request = $this->get('request_stack')->getCurrentRequest();
-
         if (isset($this->options['template'])) {
             $templateName = $this->options['template'];
         }
+
+        $request = $this->get('request_stack')->getCurrentRequest();
 
         if ($model->customTpl) {
             // Use the custom template unless it is a back end request
--- a/core-bundle/src/Controller/AbstractFragmentController.php
+++ b/core-bundle/src/Controller/AbstractFragmentController.php
@@ -86,8 +86,6 @@ abstract class AbstractFragmentController extends AbstractController implements 
     protected function createTemplate(Model $model, string $templateName): Template
     {
         $request = $this->get('request_stack')->getCurrentRequest();
-        $mainRequest = $this->get('request_stack')->getMasterRequest();
-        $templateClass = FragmentTemplate::class;
 
         if (isset($this->options['template'])) {
             $templateName = $this->options['template'];
@@ -100,9 +98,11 @@ abstract class AbstractFragmentController extends AbstractController implements 
             }
         }
 
+        $templateClass = FragmentTemplate::class;
+
         // Current request is the main request (e.g. ESI fragment), so we have to replace
-        // insert tags etc. on the template output.
-        if ($request === $mainRequest) {
+        // insert tags etc. on the template output
+        if ($request === $this->get('request_stack')->getMasterRequest()) {
             $templateClass = FrontendTemplate::class;
         }
 
--- a/core-bundle/src/Controller/AbstractFragmentController.php
+++ b/core-bundle/src/Controller/AbstractFragmentController.php
@@ -14,6 +14,7 @@ namespace Contao\CoreBundle\Controller;
 
 use Contao\CoreBundle\Fragment\FragmentOptionsAwareInterface;
 use Contao\CoreBundle\Routing\ScopeMatcher;
+use Contao\FragmentTemplate;
 use Contao\FrontendTemplate;
 use Contao\Model;
 use Contao\PageModel;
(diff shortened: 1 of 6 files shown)


fritzmg (dismissed):

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -82,20 +83,27 @@ protected function getPageModel(): ?PageModel
...
-            $request = $this->get('request_stack')->getCurrentRequest();
-
             // Use the custom template unless it is a back end request
             if (null === $request || !$this->get('contao.routing.scope_matcher')->isBackendRequest($request)) {
                 $templateName = $model->customTpl;
             }
         }
 
-        $template = $this->get('contao.framework')->createInstance(FrontendTemplate::class, [$templateName]);
+        // Current request is the main request, which means this must be an ESI fragment
```
Toflar: Why ESI? Could also be hinclude or any other custom fragment strategy?

aschempp: yes it could be hinclude, but ESI is just the most use(d/ful) one 🤷

Toflar:
```suggestion
        // Current request is the main request, which means we must have a full template render, not just a fragment template (e.g. for ESI requests)
```

Toflar (dismissed):

core-bundle/src/Resources/contao/classes/FragmentTemplate.php
```diff
@@ -0,0 +1,40 @@
...
+		return $response;
+	}
+
+	protected function compile()
+	{
+		$this->strBuffer = $this->parse();
+	}
+}
+
+class_alias(FragmentTemplate::class, 'FragmentTemplate');
```
leofeyer: This is not required for new classes.

leofeyer (commented):
The initial comment says "This is a draft attempt" but the status of the PR is "ready for review". Which one is correct?

core-bundle/src/Resources/contao/classes/FragmentTemplate.php
aschempp: Removed in 1ec4e2d29ca8f41210d878e1216cd8a16b3b4115

aschempp:
> The initial comment says "This is a draft attempt" but the status of the PR is "ready for review". Which one is correct?

Sorry about that. The draft was meant before we discussed this in the plenum at the dev meeting. Since we all agreed/couldn't think of a better way this is no longer considered a draft.

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
```diff
@@ -243,9 +275,13 @@ private function mockContainerWithFrameworkTemplate(string $templateName): Conta
...
-        $container->set('contao.framework', $framework);
+        $this->container->set('contao.framework', $framework);
+
+        return $this->container;
+    }
 
-        return $container;
+    private function getContentModel(): ContentModel
+    {
+        return (new \ReflectionClass(ContentModel::class))->newInstanceWithoutConstructor();
```
leofeyer, line 285: Why is this necessary?

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
```diff
@@ -228,11 +235,36 @@ public function testDoesNotSetTheSharedMaxAgeIfTheElementHasNeitherAStartNorASto
...
+        $currentRequest = new Request([], [], ['_scope' => 'frontend']);
+        $requestStack = $this->container->get('request_stack');
+
+        $requestStack->push(new Request()); // Main request
+        $requestStack->push($currentRequest); // Sub request
+
+        $controller = new TestController();
+        $controller->setContainer($this->container);
+
+        $controller(new Request(), $this->getContentModel(), 'main');
```
leofeyer: Should this not use `$currentRequest`?

```suggestion
        $controller($currentRequest, $this->getContentModel(), 'main');
```

core-bundle/tests/Controller/FrontendModule/FrontendModuleControllerTest.php
```diff
@@ -158,10 +190,14 @@ private function mockContainerWithFrameworkTemplate(string $templateName): Conta
...
+        $this->container->set('contao.framework', $framework);
+        $this->container->set('contao.routing.scope_matcher', $this->mockScopeMatcher());
+
+        return $this->container;
+    }
 
-        return $container;
+    private function getModuleModel(): ModuleModel
+    {
+        return (new \ReflectionClass(ModuleModel::class))->newInstanceWithoutConstructor();
```
leofeyer: Why can we no longer use `new ModuleModel()`?

aschempp: Creating an instance of `Model` will extract DCA information and fail miserably without the framework booted.

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
aschempp: Hah, yes it was meant to. But the test shows it works anyway, because the given one is not the main request 😅

leofeyer (approved):

m-vo:
Returning a legacy class from a stub like so…

```php
        $framework = $this->mockContaoFramework();
        $framework
            ->expects($this->once())
            ->method('createInstance')
            ->with(FragmentTemplate::class, ['ce_test'])
            ->willReturn(new FragmentTemplate('ce_test'))
        ;
```

broke my local tests. The class was not autoloaded: `Error : Class "Contao\FragmentTemplate" not found`.

I think this is somewhat expected as our class loader isn't running, isn't it? I could fix it by requiring the actual file.


Edit: Hmm, I switched branches today and now the error isn't occuring anymore. :thinking: I do not have an explanation, though…

PR #3603, opened 2021-10-25
Add a setting for allowed insert tags

ausi:


m-vo (dismissed):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -132,6 +138,12 @@ protected function doReplace($strBuffer, $blnCache)
 				continue;
 			}
 
+			if (preg_match($allowedTagsRegex, $elements[0]) !== 1)
+			{
+				$strBuffer .= '{{' . $strTag . '}}';
```
aschempp: doesn't this cause endless recursion?

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -565,6 +566,22 @@ static function (array $attributes): array {
         ;
     }
 
+    private function addInsertTagsNode(): NodeDefinition
+    {
+        return (new TreeBuilder('insert_tags'))
+            ->getRootNode()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('allowed_tags')
```
aschempp, line 584: I really don't like to create a config group for this only option. Maybe it could be added to `contao.security` or find a way to merge it with the `sanitizer` config? 😬

ausi: `security` and `two_factor` also have only one option, so this should not be a problem I think.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: No, see tests ☺️

core-bundle/src/DependencyInjection/Configuration.php
SeverinGloeckle: As long as only this one option is planned for the foreseeable future, I'd also prefer `contao.security.allowed_insert_tags` over `contao.insert_tags.allowed_tags` since limiting the available insert tags is most likely done for security reasons.

I could imagine options to configure the caching or maybe the delimiters for insert tags though. Then, the more general configuration group `contao.insert_tags` would be useful.

ausi: I’m pretty sure that we will get more settings for the `contao.insert_tags.` category in the future.

Toflar (dismissed):

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -565,6 +566,22 @@ static function (array $attributes): array {
...
+                    ->info('A list of allowed insert tags.')
+                    ->example(['*_url', 'request_token'])
```
fritzmg, line 586: How do I disable the `insert_content` insert tag specifically for example? I'll have to write `['insert_content::*']`, correct? If yes, I think we should add that to the example.

fritzmg (changes requested):

core-bundle/src/DependencyInjection/Configuration.php
ausi: No, only the name is checked. So `'insert_content'` would be the correct setting to **allow** it.

fritzmg: But it is checked against `$strTag`, and `$strTag` doesn't contain just the insert tag's name, but the whole string without any modifiers (e.g. "`insert_content::129`"). But according to the test it works. I am a little confused about it.

ausi: It is checked against `$elements[0]` in line 141.

fritzmg: Right, my bad :)

fritzmg (dismissed):

core-bundle/src/DependencyInjection/Configuration.php
aschempp: in your own words, that would be pretty easy to deprecate and move then, right? 😝

ausi: Sure, I’m happy with both `contao.insert_tags.allowed_tags` as well as `contao.allowed_insert_tags` but this would create a new key under `contao.` too so I don’t see any benefit. Especially if we already know that we want to add more insert tags related things in the future.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -104,9 +106,13 @@ protected function doReplace($strBuffer, $blnCache)
 		}
 
 		$strBuffer = '';
-		$container = System::getContainer();
 		$blnFeUserLoggedIn = $container->get('contao.security.token_checker')->hasFrontendUser();
 
+		$allowedTagsRegex = '(' . implode('|', array_map(static function ($allowedTag)
```
aschempp: I wonder whether this is something we could actually do when storing the configuration in the container, so it only needs to be done once for cache warmup instead of every time tags are replaced (for every template!)?

aschempp (changes requested):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: Fixed in 5b3d2c68bc903460b71e921787e1582bda42d264

m-vo (dismissed):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
aschempp: why build and reset something on every request instead of once if the container can never change? 🤷

core-bundle/src/DependencyInjection/Configuration.php
aschempp: I liked @SeverinGloeckle's idea of `contao.security.allowed_insert_tags` 😊

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: So that compiler passes can change the parameter without having to deal with the regex.

core-bundle/src/DependencyInjection/Configuration.php
ausi: For me it does not belong to `contao.security`

leofeyer:
@ausi There is a conflict now that needs to be resolved. 🙈

ausi:
> There is a conflict now that needs to be resolved. 🙈

Done

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3604, opened 2021-10-25
Use the back end access voter instead of hasAccess() and isAllowed()

aschempp:
This PR rewrites all back end access permissions to the security voter. I also moved the `BackendUser::isAllowed` and `BackendUser::canEditFieldsOf` to the voter itself, since they are pretty simple to re-implement. This means these methods are now unit tests 🎉 and I have (hopefully) improved the flow for inherited page permissions.

I did not yet rewrite `BackendUser::hasAccess()` since I'm unsure how `pagemounts` and `filemounts` access really works. And I think we can do that in a follow-up PR to keep things simpler.


core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -1229,6 +1229,11 @@ public static function disableModule()
 	{
 		trigger_deprecation('contao/core-bundle', '4.0', 'Using "Contao\System::disableModule()" has been deprecated and will no longer work in Contao 5.0. Use Composer to add or remove modules.');
 	}
+
+	public static function isGranted($attributes, $subject = null)
```
Toflar: `System::isGranted()` sounds wrong because it's missing the "security" info. What about `isUserGranted()`?

Toflar (dismissed):
Awesome work!! I've double checked every instance and could not spot any issue!
The only thing I'm questioning is the name of the shortcut for `System::isGranted()` because it's not the system that's granted anything but that's a nitpick so I'm approving anyway.

core-bundle/src/Resources/contao/library/Contao/System.php
aschempp: It is always `isGranted()` in Symfony, e.g. in the abstract controller (https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L223). Also, it is not related to the backend user, you can check access/grants on anything in the Symfony security system.

aschempp:
PR finalized, please see updated PR description.

Toflar (dismissed):
More unit tests for our permissions 🎉 Great work!

leofeyer (changes requested):
I don't like `System::isGranted()`, because we should not add new functionality to legacy classes IMHO. I understand that `System::getContainer()->get('security.helper')->isGranted()` is longer but I would still prefer it.

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -1206,8 +1207,10 @@ public function showJsLibraryHint($dc)
 			return;
 		}
 
+		$security = System::getContainer()->get('security.helper');
+
 		// Return if the user cannot access the layout module (see #6190)
-		if (!$this->User->hasAccess('themes', 'modules') || !$this->User->hasAccess('layout', 'themes'))
+		if (!$security->isGranted(ContaoCorePermissions::USER_CAN_ACCESS_MODULE, 'themes') || !$security->isGranted(ContaoCorePermissions::USER_CAN_ACCESS_LAYOUTS))
```
leofeyer, line 1214: Shouldn't this be `USER_CAN_ACCESS_THEME`?

core-bundle/src/Resources/contao/dca/tl_form_field.php
```diff
@@ -731,7 +732,7 @@ public function getFields()
 
 		foreach ($GLOBALS['TL_FFL'] as $k=>$v)
 		{
-			if ($this->User->hasAccess($k, 'fields'))
+			if (System::getContainer()->get('security.helper')->isGranted(ContaoCorePermissions::USER_CAN_ACCESS_FIELD_TYPE, $k))
```
leofeyer: Retrieving the security helper should be done outside the loop.

core-bundle/src/Security/ContaoCorePermissions.php
```diff
@@ -111,11 +111,41 @@ final class ContaoCorePermissions
     public const USER_CAN_ACCESS_FIELD_TYPE = 'contao_user.fields';
 
     /**
-     * Access is granted if the current user can access the theme.
-     * Subject must be an ID of table tl_theme.
+     * Access is granted if the current user can access the theme module.
+     * Subject must be theme module name.
```
leofeyer, line 115:
```suggestion
     * Access is granted if the current user can access the theme module.
```

core-bundle/src/Security/Voter/BackendAccessVoter.php
```diff
@@ -94,28 +116,98 @@ private function hasAccess($subject, string $field, BackendUser $user): bool
...
 
         if (!\is_array($subject)) {
-            return false;
+            $page = $this->framework->getAdapter(PageModel::class)->findByPk($subject);
+
+            if (!$page instanceof PageModel) {
+                return false;
+            }
+
+            $subject = $page->row();
```
leofeyer, line 128: This looks strange to me. The same code is already on line 118. Do we have to remove lines 117–119?

core-bundle/src/Resources/contao/dca/tl_content.php
aschempp: No 😅

core-bundle/src/Security/Voter/BackendAccessVoter.php
aschempp: The first case handles a PageModel being passed. The second one supports loading the pagemodel if an ID is given.

core-bundle/src/Resources/contao/dca/tl_form_field.php
aschempp: fixed in 45800a6d689de1e18415bcdb67a1a62af3fa8d0f

leofeyer (approved):

PR #3605, opened 2021-10-25
Do not generate routes for error pages

### LINKED ISSUE #3415: Improve the error page handling in the sitemap controller

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3415
| Docs PR or issue | -

This PR implements the discussed solution for not generating any routes for the special `error_*` page types. The `PageRegistry` will throw a `RouteNotFoundException` when the route for any of the `error_*` pages is retrieved. The "non-routable" page types are hard coded in the `PageRegistry`, similar to the legacy page types which do not have content composition (`PageRegistry::DISABLE_CONTENT_COMPOSITION` & `PageRegistry::DISABLE_ROUTING`).

Additionally the `SitemapController` checks the `_format` of the processed page and only adds pages with `html` content to the sitemap.


aschempp:
As discussed with @fritzmg, we want to investigate the options to have routing disallowed through the page controller configuration. We could replace the `compiler_class` option to prevent a page route from being compiled but still be able to fetch it from the registry for its configuration.

core-bundle/tests/Fixtures/Functional/Routing/news.yml
```diff
@@ -79,4 +79,14 @@ tl_news_archive:
...
+
+tl_page:
+    - id: 333
+      pid: 1
+      tstamp: 1636915068
+      sorting: 512
+      title: News reader
+      alias: news-reader
+      type: regular
+      published: '1'
```
fritzmg, line 92: @aschempp I am not sure why the functional tests did not fail before here. The news archive of the functional tests reference a page ID of 3 for the news reader page. However, no such page will actually be created during the functional tests - and thus URL generation will fail for the news items.

_Note:_ I used a completely new page ID, because there would be duplicate primary key errors during the tests if I used `3`.

fritzmg:
You can now disable routing for a page via

```php
/**
 * @Page(path=false)
 */
```

or

```yaml
tags:
    - { name: contao.page, path: false }
```

I have also updated all navigation modules and templates. If the navigation items contain a non-routable entry, then it will now output as a `<span>`. And in the Quick navigation and Quick link module those pages will be a disabled option.

fritzmg:
Discussion point: since a page represented by a `PageModel` can now be a non-routable page, I wonder if we should allow `PageModel::getAbsoluteUrl` and `PageModel::getFrontendUrl` to return `null`, i.e. implement the try catch block in there as opposed to in each navigation module.

aschempp (changes requested):
General approach looks very good to me, the route compiler ensures the route can never be generated as discussed. I would however change the routable check. It should be in the page registry (similar to `supportContentComposition`) and it could then probably check for the path in config instead of the route compiler.

Also this needs some adjustments after https://github.com/contao/contao/pull/3714 is merged, but I'm sure you're aware of that.

core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php
```diff
@@ -95,11 +96,17 @@ protected function getRouteConfig(Reference $reference, Definition $definition,
         $path = $attributes['path'] ?? null;
         $pathRegex = null;
 
-        if (null !== $path && 0 === strncmp($path, '/', 1)) {
+        if (\is_string($path) && 0 === strncmp($path, '/', 1)) {
             $compiledRoute = (new Route($path))->compile();
             $pathRegex = $compiledRoute->getRegex();
         }
 
+        if (false === $path) {
```
aschempp: the `$path` of `RouteConfig` cannot be boolean at the moment. This should probably be changed (including `getPath()`) and we should check the path value instead of the compiler pass in `PageRegistry::isRoutable()`

aschempp (changes requested):
I think we should not `null`-handle routes that are not routable. Any route that throws an exception (e.g. because of missing parameters) is not routable either, and we already handle these by logging and not showing e.g. in the navigation. I think the same should apply to non-routable pages of this type.

The PR points out a few issues with generating routes (e.g. `ModuleCustomnav` and `ModuleQuicklinks`) that are not handled at the moment. I think these are actually bugs that should be fixed in 4.12 (or 4.9?).

core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php
fritzmg: The `$path` in `RouteConfig` can be null, so we could just set it to `null` here. The important thing is that the route can be disabled by setting `path` to `false` in your service tag (which is already possible). I do wonder why this didn't cause an issue though in my testing yet 🤔

fritzmg:
> The PR points out a few issues with generating routes (e.g. ModuleCustomnav and ModuleQuicklinks) that are not handled at the moment. I think these are actually bugs that should be fixed in 4.12 (or 4.9?).

Only in 4.13 we will have page types that are non-routable by default. The only issues that can occur in 4.9 and 4.12 are with pages that have `requireItem` enabled. These can be fixed yes - there is already a separate or at least related PR for that though: #3540

core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php
aschempp: `null` is not the same as `false` for the path in a route config. `null` means "only use the alias", so I would update the method and allow boolean false as well.

aschempp:
What I meant was `ModuleCustomnav` etc. should do that same as we do in other places:

```php
try {
    $href = $objSubpage->getFrontendUrl();
} catch (ExceptionInterface $exception) {
    System::log('Unable to generate URL for page ID '.$objSubpage->id.': '.$exception->getMessage(), __METHOD__, ContaoContext::ERROR);
    continue 2;
}
```

#3540 does not seem to do that. Shall I create a separate PR for this (targeted at 4.12)?

fritzmg:
Setting this up for discussion, as @aschempp are in disagreement over whether non-routable pages should really never show up in any navigation 😉. (If they don't, you wouldn't be able to implement folder pages for example as a Page Controller with `@Page(path=false)`.)

fritzmg:
As discussed with @aschempp I will make the following changes to this PR:

* The `RouteNotFoundException` will not be handled separately in the navigation modules. This will either be implemented in #3540 for `4.12` or in a separate PR by @aschempp. This affects `ModuleCustomnav`,  `ModuleQuicklink` and `ModuleQuicknav`. They will not work with non routable pages after this PR is merged - however they also currently would not work with any `requireItem` page. So this needs to be handled separately for `4.12`.
* The template changes to the navigation modules will be reverted as well.

core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php
fritzmg: Done. I had to revert some of your changes of https://github.com/contao/contao/pull/3714 though.

fritzmg:
Failing tests seem unrelated.

SeverinGloeckle:
I'm out of the loop on this topic.

#3540 fixed the handling of pages with `requireItem` within the frontend preview and `ModuleBreadcrumb`:
- The frontend preview silently swallows the `InvalidParameterException` and starts the preview on the index page.
- The breadcrumbs module silently ignores pages which require an item.

For `ModuleCustomnav` and `ModuleQuicknav`, linking to a page which requires an item throws an exception – which probably is the right thing at this point and the editor should not create a link to such a page.
I'm not convinced that silently ignoring pages with `requireItem` is the right thing to do here, since an editor may wonder why these pages do not show in the modules even though they are selected.

It's a similar case with `ModuleQuicklink`. A page with `requireItem` should also be flagged as `hidden` to not end up in the navigation. One could argue that it would be nice if this module silently ignores pages which require an item.

If pages with `requireItem` should be handled explicitly in any of these modules, I can adjust #3540 so the fixes are all grouped together. In this case, please let me know how these pages should be dealt with specifically in the three modules.

aschempp (changes requested):
see https://github.com/fritzmg/contao/pull/3 for my suggestions and fixes

aschempp:
Btw I noticed these changes will result in that the error page types (and other non-routable) pages can no longer be previewed in the front end. The only way to "see" the `401 Access Denied` page is now by accessing a protected page without being logged in. This might be cumbersome to style the page and content, maybe we should allow routing these pages in preview or dev mode?

fritzmg:
Good point. Imho it's still normal that we cannot access them directly. May be we can throw appropriate exceptions when trying to access these routes in the front end, so that the content is still visible, without having to allow routing on these pages in general?

aschempp:
That would only work for error pages though.

fritzmg:
Which is fine, since they are the only ones with content. However, do you think we should generally allow non-routable pages in the preview, if they have content composition enabled?

fritzmg:
See https://github.com/contao/contao/pull/3605/commits/ccb85c3a66bcb5ae87d31ca6342770aa1c35260a - this would allow non-routable pages to still work via the `preview.php` entry point. So for example http://example.com/not-authenticated will result in 404 as the route will not exist, while http://example.com/preview.php/not-authenticated will still render the 401 page. However, those pages will also show up in http://example.com/preview.php/sitemap.xml then …

aschempp:
I think this is a problem we should solve separately (see https://github.com/contao/contao/issues/3775), can you please revert https://github.com/contao/contao/commit/ccb85c3a66bcb5ae87d31ca6342770aa1c35260a so this PR can be merged? 😊

fritzmg:
Sure thing :). Btw. I will have to revert your changes to the `UnroutablePageRouteCompiler` class, because we do need to check whether the given `Route` object is an instance of `PageRoute` - since only the `PageRoute` will have a `getPageModel` method.

aschempp (dismissed):
I would always throw a `RouteNotFoundException` in the compiler class regardless of the route type. Not sure the exception message is relevant, it could simply be `This route is not routable` since we usually catch and log it with more details anyway. Adding a hint about the page type won't harm, but it could as well show the page ID if we really need debug info.

core-bundle/src/DependencyInjection/Attribute/AsPage.php
```diff
@@ -20,7 +20,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class AsPage
 {
-    public function __construct(public ?string $type = null, public ?string $path = null, public array $requirements = [], public array $options = [], public array $defaults = [], public array $methods = [], string $locale = null, string $format = null, public bool $contentComposition = true, public ?string $urlSuffix = null)
+    /**
+     * @param string|bool|null $path
+     */
+    public function __construct(public ?string $type = null, public $path = null, public array $requirements = [], public array $options = [], public array $defaults = [], public array $methods = [], string $locale = null, string $format = null, public bool $contentComposition = true, public ?string $urlSuffix = null)
```
leofeyer, line 26: This change seems wrong. The path should be either a string or `null`. Why did you allow `bool` as well?

core-bundle/src/Routing/AbstractPageRouteProvider.php
```diff
@@ -73,7 +76,9 @@ protected function findCandidatePages(Request $request): array
         }
 
         /** @var array<PageModel> */
```
leofeyer:
```suggestion
        /** @var array<PageModel> $models */
```

core-bundle/src/DependencyInjection/Attribute/AsPage.php
fritzmg: Because we can now do

```
/**
 * @Page(path=false)
 */
```

in order to disable routing for a page.

leofeyer: What does `path=null` do then?

fritzmg: Oh wait, @aschempp changed it for the wrong property.

fritzmg: If no `path` is defined for a Page Controller, then the behavior will be the default one, i.e. it will only use the alias for the path of the route. That's the same as before.

leofeyer: It should be like this IMHO:

```php
if (null === $path) {
    if (\in_array($type, self::DISABLE_ROUTING, true)) {
        $path = '';
        $options['compiler_class'] = UnroutablePageRouteCompiler::class;
    } else {
        $path = '/'.($pageModel->alias ?: $pageModel->id).'{!parameters}';
        $defaults['parameters'] = '';
        $requirements['parameters'] = $pageModel->requireItem ? '/.+' : '(/.+?)?';
    }
}
```

fritzmg: But that would not cover `@Page(path=false)`.

leofeyer: Yes, I don't want to use `path=false` at all.

fritzmg: How else do you want to define pages that are not routable?

leofeyer: With `path=null` and the code outlined above?

fritzmg: That would not allow you to disable routing for your page controller. If you set `@Page()` or `@Page(path=null)` (which is equivalent) then the default behavior of only using the `alias` for the path applies - and routing will still be enabled.

aschempp: because the whole point of this PR is to allow `false` as a meaning of disabling routing 😂

aschempp: @leofeyer it looks like you don't yet know how page routes work. A path can either be `null` to just be the alias, it can be relative to append the alias, or it can be absolute to override the page alias. Using `false` now adds the "disable routing" feature. We cannot use `null` for that.

aschempp (dismissed):

leofeyer:
> I would always throw a `RouteNotFoundException` in the compiler class regardless of the route type.

@fritzmg What about this? Sounds reasonable to me.

aschempp (dismissed):

leofeyer (changes requested):
IMHO there is still a bug in the `Page::getPath()` method:

```php
public function getPath(): ?string
{
    return $this->path;
}
```

It should not have a return type but a `@return string|bool|null` annotation, shouldn't it?

fritzmg:
Fixed. Shouldn't PHPstan have noticed this? 🤔

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

core-bundle/src/DependencyInjection/Attribute/AsPage.php
aschempp: FYI I also just figured out why we have a difference between `null` and `''` (empty string). If the path is `null`, we generate the default behaviour of allowing any parameter after the alias. If the path is an empty string, the path will be exactly the alias, and no optional parameters.

PR #3606, opened 2021-10-25
Keep insert tags as chunked text and handle them in the HTML escaper

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

We need to keep semantics when replacing insert tags. This way we can escape text around an insert tag with the HTML strategy and the insert tag itself without (raw).

### Examples:

Assuming the context variable `text` would contain `<br> {{br}}`, these would be the results with the different filters:

| Description | Template Content | Output |
|---|---|---|
| no insert tag replacement | `{{ text }}` | `&lt;br&gt; {{br}}` |
| insert tag replacement with escaping | `{{ text\|insert_tag }}` | `&lt;br&gt; &lt;br&gt;` |
| raw insert tag, escaped outer text | `{{ text\|insert_tag_raw }}` | `&lt;br&gt; <br>` |
| all raw with insert_tag | `{{ text\|insert_tag\|raw }}` | `<br> <br>` |
| all raw with insert_tag_raw (redundant) | `{{ text\|insert_tag_raw\|raw }}` | `<br> <br>` |

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -962,9 +962,7 @@ services:
         arguments:
             - '@Contao\CoreBundle\Image\Studio\FigureRenderer'
 
-    Contao\CoreBundle\Twig\Runtime\InsertTagRuntime:
-        arguments:
-            - '@contao.framework'
+    Contao\CoreBundle\Twig\Runtime\InsertTagRuntime: ~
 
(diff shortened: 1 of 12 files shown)


ausi (approved):

leofeyer:
Thank you @m-vo.

m-vo:
Thank you for the big contribution @ausi!

ausi:
Thank you @m-vo very much for the great twig implementation! 🥳

PR #3610, opened 2021-10-26
Fix possibly undefined variables

aschempp:


Toflar (dismissed):

fritzmg (dismissed):

bytehead (dismissed):

SeverinGloeckle:
I'd suggest additionally applying the same fix for the other bundles, since `$GLOBALS['TL_CSS_UNITS']` is used in these DCAs as well:
- tl_calendar_events
- tl_faw
- tl_news

leofeyer:
How can `$GLOBALS['TL_CSS_UNITS']` ever be undefined? It is loaded when the Contao framework is loaded.

https://github.com/contao/contao/blob/b9cd28d141b243e31d6e3d93fcad2f8c6536f9f9/core-bundle/src/Resources/contao/config/config.php#L406-L407

aschempp:
it was not loaded in my unit tests I guess 🙃

leofeyer (changes requested):
I would prefer if you fixed your unit test instead. 🙈

SeverinGloeckle:
> I would prefer if you fixed your unit test instead. 🙈

You are right that this is usually only necessary in unit tests, but particularly there it is a very common requirement to load a DCA when testing legacy models or backend functions. Initializing the framework in each test is possible, but slows down tests considerably.

The change can also be implemented easily and without side effects.

I would very much welcome it to be able to load (most of the regular) DCAs stand-alone, without causing undefined variables with `E_WARNING`.

It's pretty much the same reasoning as for `$GLOBALS['TL_LANG']` in DCAs (from #2568)::

https://github.com/contao/contao/blob/f562af12895bbfaa9adb51107af6d4f18ba6af86/core-bundle/src/Resources/contao/dca/tl_content.php#L100

fritzmg (approved):

m-vo:
Can we use `?? []` instead of `?? array()` as we do for `$GLOBALS['TL_LANG']['MSC']['deleteConfirm']`?

fritzmg:
> Can we use `?? []` instead of `?? array()` as we do for `$GLOBALS['TL_LANG']['MSC']['deleteConfirm']`?

That's `$GLOBALS['TL_LANG']['MSC']['deleteConfirm'] ?? null`. The CS fixer would revert the `[]` to `array()` (for legacy code).

m-vo:
Ah you're right. My bad. :facepalm:

m-vo (approved):

leofeyer (changes requested):
I am against merging this, because

1. the issue only occurs in the unit tests and should therefore be fixed there;
2. the PR only fixes the "problem" for `TL_CSS_UNITS` and not for the other global variables.

m-vo:
> the issue only occurs in the unit tests and should therefore be fixed there

 We could also implement something like outlined here: https://github.com/contao/contao/issues/3663 (though not for 4.9)

aschempp:
Isn't that very similar to every other place where we did a patch on variables that might not be there in tests etc? This PR fixes one possible minor problem. It certainly does not fix everything but it doesn't harm either 🤷

PR #3611, opened 2021-10-26
Decorate the access decision manager

Toflar:
Contao only works with the priority strategy and in the back end, we have to allow if all voters abstain whereas in the front end, it's the opposite.

This PR implements this logic by decorating the core access decision manager and delegating it to the Symfony one (default security settings) in case it's not a Contao request to provide full compatibility with apps that add Contao as a bundle.


core-bundle/src/Security/Authentication/AccessDecisionManager.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        // Back end is allowed by default, front end is the opposite
+        $allowIfAllAbstain = 'BE' === $cacheKey;
+
+        return $this->cache[$cacheKey] = new SymfonyAccessDecisionManager(
+            $this->voters,
+            SymfonyAccessDecisionManager::STRATEGY_PRIORITY,
+            $allowIfAllAbstain
+        );
```
aschempp: I think these should be separate services registered through the container and injected into this decoration service.

aschempp (changes requested):

core-bundle/src/Security/Authentication/AccessDecisionManager.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $this->inner = $inner;
+        $this->backendDecisionManager = $backendDecisionManager;
+        $this->frontendDecisionManager = $frontendDecisionManager;
+        $this->scopeMatcher = $scopeMatcher;
+        $this->requestStack = $requestStack;
+    }
+
+    public function decide(TokenInterface $token, array $attributes, $object = null)
+    {
```
aschempp, line 36:
```suggestion
    {
        if (\method_exists($token, 'getFirewallName')) {
            switch($token->getFirewallName()) {
                case 'contao_backend':
                    return $this->backendDecisionManager->decide($token, $attributes, $object);
                    
                case 'contao_frontend':
                    return $this->frontendDecisionManager->decide($token, $attributes, $object);
                    
                default:
                    return $this->inner->decide($token, $attributes, $object);
            }
        }
```

aschempp:
I had a long thought about this again, and I think we're making this more complicated than it has to be. Its hard because its about security and we're trying to limit security risks. However, I think that is never entirely possible and ultimately is always up to the application admin. Let me explain:

1. first of all, having a separate access decision for front end and back end like in this PR no longer makes sense, since we agreed both should behave the same (use `priority` strategy and deny if there is no voter for an attribute).
2. in 95% of all cases, there is only one voter for a subject, which means the strategy is irrelevant.
3. a voter can decide on a topic (attribute), but it cannot know whether it has the final vote. Our priority strategy makes this more predictable, but it does not make it absolute. 

We are trying to make security voting predictable for extension developers, in the rare case that there is more than one voter for an attribute. However, this is ultimately not possible, because extensions can override each other and in the end someone could even override our access decision manager all over again.

I think the sane default solution should simply be to set the strategy in our manager-bundle skeleton and expect non-managed Symfony users to adopt that. Same as we do with the firewall configuration. One can override the firewall configuration with the possible outcome that the backend becomes accessible without a login. We can't prevent these mistakes 🤷

bytehead:
I'm all on the side of @Toflar:

> This PR implements this logic by decorating the core access decision manager and delegating it to the Symfony one (default security settings) in case it's not a Contao request to provide full compatibility with apps that add Contao as a bundle.

bytehead:
> I think the sane default solution should simply be to set the strategy in our manager-bundle skeleton and expect non-managed Symfony users to adopt that.

Contao is simply not the center of the universe in a Symfony application context. You just make the regular Symfony users life hard. The Contao strategy (priority) should only apply to Contao scoped requests if somehow possible.

Toflar:
I think decorating is the easiest solution to achieve compatibility with both. But probably we don't need a different ADM for both, FE and BE. They are the same anyway.
And maybe we should simply decide on `$token->getUser() instanceof Contao\User`.
That sounds like the easiest version to begin with.

aschempp:
> Contao is simply not the center of the universe in a Symfony application context.

I agree with that. But if your application configuration is not compatible with Contao, stuff simply might not work as designed. That is not only applicable to security but to a lot of other settings as well. If you disable ESI/don't have a cache proxy, requests that are supposed to affect your HTTP cache might not do so. If you change the Symfony framework's CSRF protection, Contao forms might miss out on the request token check. If you disable UTF8 support for routing, users can enter chinese characters as page alias and their pages will never be routable.

> And maybe we should simply decide on `$token->getUser() instanceof Contao\User`.
> That sounds like the easiest version to begin with.

A firewall is not related to a specific user object at all. Yes, the **voter** _might_ not be able to vote on anything but a `Contao\User`. But the **access decision manager** should still ask **all voters** for their vote. A voter does not know whether its decision is the final one. A voter also does not know whether the call was made on a Contao request or in what firewall, and as far as I understand it's not supposed to know.

---

Ultimately, I think any approach will work 99% of the time and it just does not matter. It boils down to the question whether we want to add a complex service decoration for the very unlikely case that someone really needs to have a different strategy. If they do, they can still decorate the access decision manager themselves and do exactly what their application needs. For us, its whether we want to add two lines of configuration or a complex service decorator that still does not know whether what it does is ultimately right for the current application.

I'm in favour of having just the simple configuration and let app developer handle edge cases, because they need to anyways.

leofeyer:
As discussed in the Contao call, we want to keep our separate access decision manager. Also, we do not want two decision managers for the back end and the front end. One should suffice.

Toflar:
Updated as discussed yesterday. Super simple now 💪

core-bundle/tests/Security/Authentication/AccessDecisionManagerTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $accessDecisionManager = new AccessDecisionManager(
+            $inner,
+            $contao,
+            $this->mockScopeMatcher(),
+            $requestStack
+        );
+
+        $accessDecisionManager->decide($this->createMock(TokenInterface::class), []);
+    }
+
```
bytehead, line 82: Isn't here a test missing with 
```
$inner = $this->createAccessDecisionManager(false);
$contao = $this->createAccessDecisionManager(true);
```
?

bytehead (changes requested):

core-bundle/tests/Security/Authentication/AccessDecisionManagerTest.php
Toflar: Lol, that got lost, sorry. It's back now.

bytehead: Thanks! :)

bytehead (approved):

m-vo (approved):

core-bundle/src/Security/Authentication/AccessDecisionManager.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->inner = $inner;
+        $this->contaoAccessDecisionManager = $contaoAccessDecisionManager;
+        $this->scopeMatcher = $scopeMatcher;
+        $this->requestStack = $requestStack;
+    }
+
+    public function decide(TokenInterface $token, array $attributes, $object = null)
+    {
+        $request = $this->requestStack->getCurrentRequest();
```
aschempp: The firewall only operates on the main request (https://github.com/symfony/symfony/blob/a5aaf2c9dc9cc04e472af32d3427a98c5ecebcb8/src/Symfony/Component/Security/Http/Firewall.php#L52) so I think we should check the main request here.

aschempp (changes requested):

core-bundle/src/Security/Authentication/AccessDecisionManager.php
Toflar: Updated accordingly :)

aschempp (approved):

core-bundle/src/Resources/config/services.yml
```diff
@@ -629,6 +629,21 @@ services:
...
+    contao.security.access_decision_manager:
+        decorates: security.access.decision_manager
+        class: Contao\CoreBundle\Security\Authentication\AccessDecisionManager
+        arguments:
+            - '@.inner'
+            - '@contao.security.access_decision_manager.contao'
+            - '@contao.routing.scope_matcher'
+            - '@request_stack'
+
+    contao.security.access_decision_manager.contao:
```
```suggestion
    contao.security.access_decision_manager.priority:
```

WDYT?

Toflar: I don't mind :)

m-vo (approved):

leofeyer:
Thank you @Toflar.

PR #3613, opened 2021-10-26
Add a root page dependent module selector

bytehead:
To-Do
-----

- [x] Naming
- [x] Tests

The idea is to get rid of multiple page layouts (or inserttags with module IDs in combination with `{{iflng::*}}` tags) for multiple root pages (and languages).

Usage
-----

You can now easy configure you modules for every root page available:

Bildschirmfoto 2021-12-15 um 22 36 21

and include this module in your layout: 

Bildschirmfoto 2021-12-15 um 22 39 41

This module will then render the configured module for each configured root page. 

Developers
----------

The module provides an additional input type to use in your own code:

**Use the default configuration** (loads all available modules):
```php
<?php
// config/dca/tl_modules.php

$GLOBALS['TL_DCA']['tl_module']['fields'] += [
    'myRootPageDependentModule' => [
        'inputType' => 'rootPageDependentSelect',
        'eval' => [
            'tl_class' => 'w50',
        ],
        'sql' => 'blob NULL',
    ],
];
```

There is even more and you can customize this widget to your needs:

**use `options`** (for custom list of modules):
```php
<?php
// config/dca/tl_modules.php

$GLOBALS['TL_DCA']['tl_module']['fields'] += [
    'myRootPageDependentModule' => [
        'inputType' => 'rootPageDependentSelect',
        'options' => [
            0 => 'My module 0',
            1 => 'My module 1',
        ],
        'eval' => [
            'tl_class' => 'w50',
        ],
        'sql' => 'blob NULL',
    ],
];
```

**use `options_callback`** (for custom list of modules):
```php
<?php
// config/dca/tl_modules.php

$GLOBALS['TL_DCA']['tl_module']['fields'] += [
    'myRootPageDependentModule' => [
        'inputType' => 'rootPageDependentSelect',
        'options_callback' => ['my.service_id', 'methodName'],
        'eval' => [
            'tl_class' => 'w50',
        ],
        'sql' => 'blob NULL',
    ],
];

// you can also use this with service tagging, see https://docs.contao.org/dev/framework/dca/#registering-callbacks
```

**use `eval['modules']`** (to filter for custom types):
```php
<?php
// config/dca/tl_modules.php

$GLOBALS['TL_DCA']['tl_module']['fields'] += [
    'myRootPageDependentModule' => [
        'inputType' => 'rootPageDependentSelect',
        'eval' => [
            'tl_class' => 'w50',
            'modules' => [
                'navigation',
                'customnav',
                'search',
                'html',
                'myCustomModule',
            ],
        ],
        'sql' => 'blob NULL',
    ],
];
```

**Change the label for the blank option**:
```php
<?php
// config/dca/tl_modules.php

$GLOBALS['TL_DCA']['tl_module']['fields'] += [
    'myRootPageDependentModule' => [
        'inputType' => 'rootPageDependentSelect',
        'eval' => [
            'tl_class' => 'w50',
            'blankOptionLabel' => 'My Label'
        ],
        'sql' => 'blob NULL',
    ],
];
```

_Note:_ If you just want to change the label for the blank option of the default field, the key is: `tl_module.rootPageDependentModulesBlankOptionLabel`.


bezin:
I really love this new feature! I used your [bundle](https://github.com/1up-lab/contao-language-dependent-modules-bundle) as an inspiration when I ran into a somewhat similiar situation: 

I have multiple root pages that share almost everything, esp. the layout, but some modules differ because of different module settings. They are all in German (think of one company, but different page trees for different branches). I developed a root page dependent module like the language dependent module which then chooses the module to render depending on the current root page. 

I would find this also very handy in the core. What do you think?

Toflar:
We actually discussed that this should be root page dependent, not language... ;)

asaage:
I've been using https://github.com/terminal42/contao-rootcontent for that kind of usecase and i can't see a benefit here at all yet.

bytehead:
It is actually root page dependent. Useful naming is a big task here :sweat_smile:

bezin:
Is it? You fetch the whole row but store the language code: 

https://github.com/contao/contao/commit/b81717c4e677105eb614f1be9c611e237fd5655a#diff-6a50f8ed0c115f3ecf9d5ae43de21850af61cc76364c727e6050ed086c33e6dcR198-R202

This would not work in my scenario, as all root pages have the same language code.

core-bundle/src/EventListener/DataContainer/LanguageDependentModuleListener.php
```diff
@@ -0,0 +1,238 @@
...
+    public function onEditModule(DataContainer $dc): string
+    {
+        $wizards = [];
+        $values = StringUtil::deserialize($dc->value, true);
+
+        if (empty($values)) {
+            return '';
+        }
+
+        System::loadLanguageFile('tl_content');
```
aschempp: no need to load the language file if you use the translator.

core-bundle/src/EventListener/DataContainer/LanguageDependentModuleListener.php
```diff
@@ -0,0 +1,238 @@
...
+                sprintf('LNG.%s', $rootPage['language']),
+                [],
+                'contao_languages'
+            );
+        }
+
+        return $languages;
+    }
+
+    private function getFrontendModules(array $types = []): array
```
aschempp: this should only show the modules of the current theme (same where you create the selector module).

aschempp (changes requested):
as said previously, this needs to take account of actual root pages not languages. In a multi-domain environment, the same language might be on multiple root pages.

core-bundle/src/EventListener/DataContainer/LanguageDependentModuleListener.php
bytehead: Removed in 9e5e12d03.

core-bundle/src/EventListener/DataContainer/LanguageDependentModuleListener.php
bytehead: Added in 3b8c80c46.

bytehead:
I've rebased against 4.x and put some fresh screenshots in the description. Before writing extensive tests - RFC 🙃

core-bundle/src/Resources/contao/config/config.php
```diff
@@ -332,34 +333,35 @@
...
+	'listWizard'              => ListWizard::class,
+	'optionWizard'            => OptionWizard::class,
+	'moduleWizard'            => ModuleWizard::class,
+	'keyValueWizard'          => KeyValueWizard::class,
+	'imageSize'               => ImageSize::class,
+	'timePeriod'              => TimePeriod::class,
+	'metaWizard'              => MetaWizard::class,
+	'sectionWizard'           => SectionWizard::class,
+	'serpPreview'             => SerpPreview::class,
+	'rootPageDependentModule' => RootPageDependentModule::class
```
Toflar: Why would you register a new fragment the old way?

core-bundle/src/Resources/config/listener.yml
```diff
@@ -116,6 +116,19 @@ services:
...
+        public: true
+        arguments:
+            - '@database_connection'
+            - '@translator'
+            - '@contao.routing.scope_matcher'
+            - '@request_stack'
+            - '@Contao\CoreBundle\Csrf\ContaoCsrfTokenManager'
+            - '%contao.csrf_token_name%'
+        tags:
+            - { name: contao.hook, hook: loadDataContainer }
```
Toflar: Why not just the `@Hook()` annotation?

core-bundle/src/Resources/contao/templates/modules/mod_root_page_dependent_module.html5
```diff
@@ -0,0 +1 @@
+<?= $this->module;
```
Toflar: Do you see any value in having an extra template here? I would just render it directly. We do the same in `ContentModule` for example. There's no point in having just one rendered HTML variable in a template.

core-bundle/src/Resources/contao/widgets/RootPageDependentModule.php
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao;
+
+/**
+ * @property array $rootPages
+ */
+class RootPageDependentModule extends SelectMenu
```
Toflar: Can you explain why we need a special widget?

Toflar (changes requested):

core-bundle/src/Resources/contao/widgets/RootPageDependentModule.php
bytehead: How would you render this thing with multiple selects?

core-bundle/src/Resources/contao/config/config.php
bytehead: Possibly kind of "operational blindness" 😬

core-bundle/src/Resources/config/listener.yml
bytehead: Will update. 👍

core-bundle/src/Resources/contao/templates/modules/mod_root_page_dependent_module.html5
bytehead: Fine with rendering directly - didn't think about it. 👍

bezin: I actually don't mind the template. What if I want to add a specific wrapper to this particular module, e.g. a CSS class that depends on the root page?

core-bundle/src/Resources/contao/widgets/RootPageDependentModule.php
Toflar: Let me rephrase: I can see why you need a special widget here but why is this extending `SelectMenu` then? Or in other words: You add a special widget which is only here for `rootPages`. That makes no sense to me. Why not allow a `MultiSelectMenu`. You already load lots of stuff with callbacks anyway so maybe it can be generalized? Right now it's a mix.

bytehead: I see, I like the idea. 👍

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -125,6 +125,7 @@
 		'template'                    => '{title_legend},name,headline,type;{template_legend},data,customTpl;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
 		'rssReader'                   => '{title_legend},name,headline,type;{config_legend},rss_feed,numberOfItems,perPage,skipFirst,rss_cache;{template_legend:hide},rss_template;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
 		'two_factor'                  => '{title_legend},name,headline,type;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},cssID',
+		'root_page_dependent_module'  => '{title_legend},name,type;{config_legend},rootPageDependentModules;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},cssID'
```
fritzmg:
```suggestion
		'root_page_dependent_module'  => '{title_legend},name,type;{config_legend},rootPageDependentModules;{template_legend:hide},customTpl;{protected_legend:hide},protected'
```

The CSS classes and CSS ID are never output since the template does not extend from `block_searchable` or `block_unsearchable` - or will be removed altogether (https://github.com/contao/contao/pull/3613/files#r777346193).

fritzmg (changes requested):

bezin:
What about naming it only **Root page dependent module** ~~configuration~~? That is still specific enough and would short it a bit.

core-bundle/src/Resources/contao/dca/tl_module.php
bytehead: We could extend from `block_unsearchable` though (see https://github.com/contao/contao/pull/3613#discussion_r777360303).

core-bundle/src/Resources/contao/config/config.php
bytehead: Removed in f6b8d50ce.

core-bundle/src/Resources/config/listener.yml
bytehead: Adjusted in b638dd2cc.

core-bundle/src/Resources/contao/templates/modules/mod_root_page_dependent_module.html5
bytehead: Fixed in 86cbde127.

core-bundle/src/Resources/contao/dca/tl_module.php
bytehead: Removed in e97704c2f, as the rendered module itself brings its own classes (see https://github.com/contao/contao/commit/86cbde12746d431da23cabc1b8dff82d41bd0440 as well).

core-bundle/src/Resources/contao/config/config.php
bytehead: This is actually not the fragment, it's the backend form field type. Will readd.

core-bundle/src/Resources/config/listener.yml
bytehead: Should only have removed the tag 😂 I'll add it with the change to a more generic solution.

core-bundle/src/Controller/FrontendModule/RootPageDependentModuleController.php
```diff
@@ -0,0 +1,45 @@
...
+        $modules = StringUtil::deserialize($model->rootPageDependentModules);
+        $content = '';
+
+        if (\is_array($modules) && \array_key_exists($pageModel->rootId, $modules)) {
+            $content = $controller->getFrontendModule($modules[$pageModel->rootId]);
+        }
+
+        $response = new Response($content);
+
+        $this->markResponseForInternalCaching($response);
```
aschempp: this makes no sense since the controller does not include caching information anyway.
Also the rendered module might come from a fragment response that already affected caching (not sure if that works, but that would be a separate issue).

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -634,7 +635,23 @@
 			'inputType'               => 'text',
 			'eval'                    => array('multiple'=>true, 'size'=>2, 'tl_class'=>'w50'),
 			'sql'                     => "varchar(255) NOT NULL default ''"
-		)
+		),
+		'rootPageDependentModules' => array
+		(
+			'exclude'                 => true,
+			'inputType'               => 'rootPageDependentModule',
+			'options_callback'        => array('contao.listener.data_container.root_page_dependent_module', 'onOptionsCallback'),
```
aschempp: this should be done through annotation or service tag

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -634,7 +635,23 @@
...
+			'save_callback'           => array
+			(
+				array('contao.listener.data_container.root_page_dependent_module', 'onSaveCallback')
```
aschempp: this should be done through annotation or service tag

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -634,7 +635,23 @@
...
+			),
+			'wizard'                  => array
+			(
+				array('contao.listener.data_container.root_page_dependent_module', 'onEditModule')
```
aschempp: this should be done through annotation or service tag

aschempp (dismissed):
I guess the `MultiSelectMenu` is missing as well?

core-bundle/src/EventListener/DataContainer/RootPageDependentModuleListener.php
```diff
@@ -0,0 +1,206 @@
...
+            $GLOBALS['TL_DCA'][$table]['fields'][$key] = $fieldConfig;
+        }
+    }
+
+    public function onOptionsCallback(DataContainer $dc): array
+    {
+        $options = [];
+        $types = $GLOBALS['TL_DCA'][$dc->table]['fields'][$dc->field]['eval']['modules'] ?? [];
+
+        $statement = $this->connection->executeQuery(
```
m-vo: `$statement` will likely get overwritten a few lines down, might be better suited in an else clause. :upside_down_face:

core-bundle/src/Controller/FrontendModule/RootPageDependentModuleController.php
bytehead: cc @Toflar

Toflar: It does make sense but the tags are missing. When you edit this FE module, the pages where this module is placed should be invalidated. So the problem here is not the `markResponseForInternalCaching()`, that's correct. The problem is the missing `tagResponse()` call.

bytehead: Add tag in 971a70a7f.

core-bundle/src/Resources/contao/dca/tl_module.php
bytehead: Done in 6c85b7b8a.

core-bundle/src/Resources/contao/dca/tl_module.php
bytehead: Done in 6c85b7b8a.

core-bundle/src/Resources/contao/dca/tl_module.php
bytehead: Done in 6c85b7b8a.

core-bundle/src/Resources/contao/widgets/RootPageDependentModule.php
bytehead: After our disscusion on Slack I went for `RootPageDependentSelect` as a generalized solution would be more in the direction of a MCW / group widget. Would you prefer `RootPageSelect` instead?

core-bundle/src/EventListener/DataContainer/RootPageDependentModuleListener.php
bytehead: It's actually ment to be overwritten if there is some types available to filter for. Possibly more a personal thing, as I don't like else clauses 🙈

core-bundle/src/Resources/contao/widgets/RootPageDependentModule.php
leofeyer: `RootPageDependentSelect` 👍

core-bundle/src/EventListener/DataContainer/RootPageDependentModuleListener.php
bytehead: Refactored in 5c9f0a50b.

core-bundle/src/Controller/FrontendModule/RootPageDependentModuleController.php
aschempp: `markResponseForInternalCaching` tells our fragment handler to merge `Cache-Control` headers, if the response has any. Calling it is useful if you do add cache headers, or if one does not know if the parent method does (that's why the twig render method calls it). It is not useful if you do not add any caching info to the response (as this module does).

Toflar: Hmm true, you would actually kill caching now because the default is `private, no-store` 🙈  So that has to go again. But I think I'd leave the changes in abstract anyway because there surely are other use cases where a controller would want to use that method.

bytehead: So remove the tag again?

Toflar: Nooo, the tag is correct, the cache call is not. I'll suggest.

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -0,0 +1,46 @@
...
+        $content = '';
+
+        if (\is_array($modules) && \array_key_exists($pageModel->rootId, $modules)) {
+            $content = $controller->getFrontendModule($modules[$pageModel->rootId]);
+        }
+
+        $response = new Response($content);
+
+        $this->tagResponse($model);
+        $this->markResponseForInternalCaching($response);
```
Toflar:
```suggestion
        // Do not call $this->markResponseForInternalCaching($response); as this would turn all responses into
        // private, no-store as our response never has caching information.
```

core-bundle/src/EventListener/DataContainer/RootPageDependentSelectListener.php
```diff
@@ -0,0 +1,207 @@
...
+        $this->scopeMatcher = $scopeMatcher;
+        $this->requestStack = $requestStack;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->csrfTokenName = $csrfTokenName;
+    }
+
+    /**
+     * @Hook("loadDataContainer")
+     */
+    public function onLoadDataContainer(string $table): void
```
Toflar: Why is the logic not in the widget itself? It's a very specific widget that loads stuff per root page. Why setting them from the outside if the widget itself already knows about them?

Toflar (changes requested):
General remark: I think you could use `iterateAssociative()` instead of `fetchAllAssociative()` to optimize memory usage.

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
bytehead: Fixed in 4d1878353.

core-bundle/src/EventListener/DataContainer/RootPageDependentSelectListener.php
bytehead: Valid point. This has possibly historical reasons how this thing evolved 😄

core-bundle/src/Controller/FrontendModule/RootPageDependentModuleController.php
bytehead: Fixed in 4d18783.

core-bundle/src/EventListener/DataContainer/RootPageDependentSelectListener.php
bytehead: I've restructured it in c742a6b6e.

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -0,0 +1,47 @@
...
+        $content = '';
+
+        if (\is_array($modules) && \array_key_exists($pageModel->rootId, $modules)) {
+            $content = $controller->getFrontendModule($modules[$pageModel->rootId]);
+        }
+
+        $response = new Response($content);
+
+        $this->tagResponse($model);
+        // Do not call $this->markResponseForInternalCaching($response); as this would turn all responses into
```
Toflar: The explanation is actually not true, just remove the comment. Sorry 🙈

bytehead: Removed in 2b801f251.

fritzmg (dismissed):

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
```diff
@@ -0,0 +1,102 @@
...
+{
+	public function generate(): string
+	{
+		$fields = array();
+		$originalLabel = $this->strLabel;
+		$cssClasses = 'tl_select tl_chosen';
+		$rootPages = PageModel::findByPid(0, array('order' => 'sorting'));
+		$wizard = StringUtil::deserialize($this->wizard);
+
+		$this->blankOptionLabel = $GLOBALS['TL_LANG']['tl_module'][sprintf('%sBlankOptionLabel', $this->name)];
```
Toflar: What's this? Why not the regular one?

Toflar (dismissed):
Question regarding the blank label and I don't understand why the widget is untested and not using adapters but other than that I think it's "good enough" :)

core-bundle/src/EventListener/Widget/RootPageDependentSelectListener.php
```diff
@@ -0,0 +1,136 @@
...
+    private TranslatorInterface $translator;
+    private CsrfTokenManagerInterface $csrfTokenManager;
+    private string $csrfTokenName;
+
+    public function __construct(Connection $connection, TranslatorInterface $translator, CsrfTokenManagerInterface $csrfTokenManager, string $csrfTokenName)
+    {
+        $this->connection = $connection;
+        $this->translator = $translator;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->csrfTokenName = $csrfTokenName;
```
m-vo: You could inject the `ContaoCsrfTokenManager` and use our new `getDefaultTokenValue()` method instead of injecting the manager and the name separately.

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
```diff
@@ -0,0 +1,102 @@
...
+		$this->strLabel = $originalLabel;
+
+		return implode('', $fields);
+	}
+
+	protected function isSelected($option): string
+	{
+		if (empty($this->varValue) && empty($_POST) && ($option['default'] ?? null))
+		{
+			return static::optionSelected((string) 1, 1);
```
m-vo: You need this to generate the string ` selected`, right? This should IMHO get rid of the implicit casts, then:

```suggestion
			return static::optionSelected('1', '1'); // selected
```

Or just use the value:

```suggestion
			return ' selected';
```

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
```diff
@@ -0,0 +1,102 @@
...
+namespace Contao;
+
+class RootPageDependentSelect extends SelectMenu
+{
+	public function generate(): string
+	{
+		$fields = array();
+		$originalLabel = $this->strLabel;
+		$cssClasses = 'tl_select tl_chosen';
+		$rootPages = PageModel::findByPid(0, array('order' => 'sorting'));
```
m-vo: In the `RouteProvider` we find the pages by `type = 'root'`. Should we do this here as well?

m-vo (dismissed):
:heart: People will love this!

I left some remarks but nothing to stop the PR. :slightly_smiling_face:

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: The option to set a custom blank label instead of using the core default one (`-`):

https://github.com/contao/contao/blob/cf2ca71c614232865301443b71c66b6a31152a38/core-bundle/src/Resources/contao/library/Contao/Widget.php#L1360-L1364

It gives you an insight, what option you should select for: 

Bildschirmfoto 2022-01-16 um 12 05 19

And think about having multiple `rootPageDependentSelect` widgets - you probably don't have the same blank label depending on your use case. This way you can fully customize it. Or is there another, more prefered way to achieve this? :-)

core-bundle/src/EventListener/Widget/RootPageDependentSelectListener.php
bytehead: Fully agree! :)

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: Yes definitely.

core-bundle/src/EventListener/Widget/RootPageDependentSelectListener.php
bytehead: Adjusted in de2b7dc15.

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: Fixed in 9f56fda61.

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: Adjusted in a4762f93b.

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: as @Toflar is in holidays - WDYT @contao/developers?

bytehead:
> Question regarding the blank label and I don't understand why the widget is untested and not using adapters but other than that I think it's "good enough" :)

I've added a test for the widget, it uses the adapter for the `PageModel` and gets the translator from the container now (not using globals anymore).

m-vo (dismissed):

leofeyer (dismissed):

m-vo (dismissed):

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -0,0 +1,39 @@
...
+        $modules = StringUtil::deserialize($model->rootPageDependentModules);
+        $content = '';
+
+        if (\is_array($modules) && \array_key_exists($pageModel->rootId, $modules)) {
+            $content = $controller->getFrontendModule($modules[$pageModel->rootId]);
+        }
+
+        $this->tagResponse($model);
+
+        return new Response($content);
```
leofeyer: IMHO, this method should look like this:

```php
        if (!$pageModel = $this->getPageModel()) {
            return new Response(''); // This can happen in the back end preview if no module has been selected
        }

        $modules = StringUtil::deserialize($model->rootPageDependentModules);

        if (empty($modules) || !\is_array($modules) || !\array_key_exists($pageModel->rootId, $modules)) {
            return new Response(''); // No need for further processing if there are no modules
        }

        $controller = $this->container->get('contao.framework')->getAdapter(Controller::class);
        $content = $controller->getFrontendModule($modules[$pageModel->rootId]);

        $this->tagResponse($model);

        return new Response($content);
```

@bytehead WDYT?

bytehead: 👍 adjusted in f2984b71b.

m-vo (dismissed):

leofeyer (dismissed):

leofeyer (dismissed):

core-bundle/tests/Controller/FrontendModule/RootPageDependentModulesControllerTest.php
```diff
@@ -0,0 +1,185 @@
...
+            ->method('trans')
+            ->willReturn('Root Page Dependent Modules')
+        ;
+
+        $container = $this->mockContainer();
+        $container->set('contao.routing.scope_matcher', $scopeMatcher);
+        $container->set('router', $router);
+        $container->set('translator', $translator);
+
+        TemplateLoader::addFile('be_wildcard', __DIR__.'/../../../src/Resources/contao/templates/backend');
```
m-vo: The path should probably be wrapped in `Path:: canonicalize()`.

leofeyer: Changed in 41f654ba9f716bf69d92288977a57f50c63ecd3c. 👍

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #3614, opened 2021-10-26
Add a nonce to all string placeholders

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

This adds a nonce to all `[[TL_*]]` placeholders we are using to prevent users entering them in text fields.


Toflar (dismissed):

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -114,6 +117,8 @@ public function initialize(bool $isFrontend = false): void
         if (!$this->legacyRouting) {
             $this->throwOnLegacyRoutingHooks();
         }
+
+        self::$nonce = random_bytes(6);
```
ausi:
```suggestion
        self::$nonce = bin2hex(random_bytes(16));
```

As some code might rely on valid UTF-8 code in templates we should only use “safe” ACII characters I think.

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -835,19 +836,21 @@ public static function replaceDynamicScriptTags($strBuffer)
...
 
-		$arrReplace['[[TL_JQUERY]]'] = $strScripts;
+		$nonce = ContaoFramework::getNonce();
+
+		$arrReplace["[[TL_JQUERY_$nonce]]"] = $strScripts;
 		$strScripts = '';
 
 		// Add the internal MooTools scripts
-		if (!empty($GLOBALS['TL_MOOTOOLS']) && \is_array($GLOBALS['TL_MOOTOOLS']))
+		if (!empty($GLOBALS["TL_MOOTOOLS_$nonce]]"]) && \is_array($GLOBALS['TL_MOOTOOLS']))
```
ausi:
```suggestion
		if (!empty($GLOBALS['TL_MOOTOOLS']) && \is_array($GLOBALS['TL_MOOTOOLS']))
```

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -835,19 +836,21 @@ public static function replaceDynamicScriptTags($strBuffer)
...
 		{
-			foreach (array_unique($GLOBALS['TL_MOOTOOLS']) as $script)
+			foreach (array_unique($GLOBALS["TL_MOOTOOLS_$nonce]]"]) as $script)
```
ausi:
```suggestion
			foreach (array_unique($GLOBALS['TL_MOOTOOLS']) as $script)
```

core-bundle/src/Twig/Interop/PhpTemplateProxyNodeVisitor.php
```diff
@@ -67,7 +68,9 @@ private function configurePhpTemplateProxy(ModuleNode $node): void
                 continue;
             }
 
-            $blockNodes[$name] = new BlockNode($name, new TextNode('[[TL_PARENT]]', 0), 0);
+            $nonce = ContaoFramework::getNonce();
+
+            $blockNodes[$name] = new BlockNode($name, new TextNode("[[TL_PARENT_$nonce]]", 0), 0);
```
ausi: Can there be cases where this block node get cached by twig?

ausi (changes requested):

core-bundle/src/Framework/ContaoFramework.php
m-vo: Agreed. :+1:

core-bundle/src/Resources/contao/library/Contao/Controller.php
m-vo: This place is accessing the global variable not replacing a `[[…]]` string, no?

ausi: Correct. That’s why we should not use the nonce here I think.

core-bundle/src/Twig/Interop/PhpTemplateProxyNodeVisitor.php
m-vo: Very good point! First I thought all of this is dynamic anyways, but we're actually writing block nodes here. 

Fixed with a custom node in 4c0273ab36c3eea9aa38f894a05306716b669b1e.

core-bundle/src/Resources/contao/library/Contao/Controller.php
m-vo: Oops, I read the diff the wrong way. :smile:

ausi (approved):

aschempp (changes requested):
as discussed at the developer meeting, we should find a better place than the `ContaoFramework` to store the nonce. The `ContaoFramework` is about the PHP framework to me, not the front end layout/templating system.

ausi:
> we should find a better place than the `ContaoFramework` to store the nonce.

I would suggest `TemplateInheritance::$nonce` and `TemplateInheritance::getNonce()` then.

Toflar:
We chose ContaoFramwork because it's something that will get removed once we're Twig-only.
But so will TemplateInheritance so that's fine too.

aschempp:
Looking into the code, `TemplateInheritance` seems good to me.

I'm not fully sure how (recursive) template inheritance works, but it looks like the variable does not need to be static but could be for each trait? `TemplateInheritance` could then use its own nonce for `TL_PARENT`. For the case of `PageRegular`, the nonce can be retrieved from `$this->Template->getNonce()` and forwarded to `Controller::replaceDynamicScriptTags()`. This might also have the advantage of never possibly overriding each other.

But if that does not work with recursive templates (I'm sure @ausi knowns 😊) then a static variable certainly works as well.

m-vo:
> but it looks like the variable does not need to be static but could be for each trait

We need to access the nonce from three different places: `PageRegular`, the `TemplateInheritance` trait and from within rendered Twig templates. None of these are services, so a static reference makes sense to me. It's basically a global but with the benefit of a typed interface and an `@internal` annotation, so that we can remove it at any point.

There are two different things (head tags that are replaced at a late stage / internals of the PHP template system to make inheritance work) but IMHO it's overkill to use different values/scopes here. 

> The ContaoFramework is about the PHP framework to me, not the front end layout/templating system.

Yes, true. I'd still go with it because we only need it when working with the legacy framework. And the `ContaoFramework` can generate the nonce when being initialized. 

We cannot use the `TemplateInheritance` trait. Calling static traits or accessing static properties in traits is deprecated in PHP8.1: https://3v4l.org/lJUTV

aschempp:
> We cannot use the `TemplateInheritance` trait. Calling static traits or accessing static properties in traits is deprecated in PHP8.1: https://3v4l.org/lJUTV

Works for me as soon as the trait is implemted by an actual class 🤷 
https://3v4l.org/eiIiE
But my point was to not use a static anyway, since only the template should know about it.

m-vo:
> But my point was to not use a static anyway, since only the template should know about it.

We do not have a class for our legacy template logic, though. `FrontendTemplate`, `BackendTemplate`, `Widget`, … all are using the `TemplateInheritance` trait.

aschempp:
Here's a PR that hopefully explains what I mean: https://github.com/m-vo/contao/pull/7/

The nonce in dynamic script tags are only relevant to the current template and the front end page output. Therefore it does not need to by static in the framework.

m-vo:
> Here's a PR that hopefully explains what I mean: m-vo#7

This does not work from within Twig templates, see `PhpTemplateParentReference`.

m-vo:
todo: 
- [x] rebase + resolve conflicts
- [x] rename `PhpTemplateParentReference` &rarr; `PhpTemplateParentReferenceNode`

leofeyer:
As discussed in the Contao call, we want to keep the method in the `ContaoFramework` class.

ausi (approved):

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -114,6 +117,8 @@ public function initialize(bool $isFrontend = false): void
         if (!$this->legacyRouting) {
             $this->throwOnLegacyRoutingHooks();
         }
+
+        self::$nonce = bin2hex(random_bytes(16));
```
leofeyer: We could move this into the `getNonce()` method so we only generate the nonce if someone uses it.

```php
public static function getNonce(): string
{
    if ('' === self::$nonce) {
        self::$nonce = bin2hex(random_bytes(16));
    }

    return self::$nonce;
}
```

m-vo: Done in 015450554.

ausi (approved):

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -82,6 +83,8 @@ public function reset(): void
         Input::resetUnusedGet();
         InsertTags::reset();
         Registry::getInstance()->reset();
+
+        self::$nonce = '';
```
leofeyer: I think this line needs to go before the if condition on line 77, otherwise the nonce will not be reset if the framework has not been initialized.

Toflar (approved):

leofeyer (changes requested):

core-bundle/src/Framework/ContaoFramework.php
m-vo: Right, now that we're setting the nonce in the getter. Changed in 66e358f70.

ausi (approved):
❤️

leofeyer:
Thank you @m-vo.

PR #3615, opened 2021-10-26
Add the maker bundle

sheeep:
| Q                | A
| -----------------| ---
| Fixed issues     |
| Docs PR or issue | tbd.

This PR adds a new bundle to the Contao ecosystem in `contao/contao`. The `contao/maker-bundle` provides commands to easily generate empty Content Elements, Frontend Modules, Hooks, DCA-Callbacks and Event-Listener for Contao specific events by guiding you through a simple CLI-wizard.

This bundle is heavily inspired by and relies on the `symfony/maker-bundle`.

Upon installation, five different commands will be available for usage:

```
  make:contao:content-element             Creates an empty content element
  make:contao:dca-callback                Creates a dca callback
  make:contao:event-listener              Creates an event listener for a Contao event
  make:contao:frontend-module             Creates an empty frontend module
  make:contao:hook                        Creates a hook
```

## Principles & Process

The `make:contao:dca-callback` command (for example) allows you to generate a DCA-Callback by following the wizard presented to you. You'll be asked to provide a `name`, a `table` and a `target`. For both `table` and `target` there is a suggestion list available, which implements the [auto completer](https://symfony.com/doc/current/components/console/helpers/questionhelper.html#autocompletion). If your target happens to be a "dynamic" target (for example `fields.{field}.load`), you'll additionally be asked to provide the value for the dynamic part (in this case `field`).

The `tables` suggestion list will be compiled from files in `dca/` folders found by `Contao\CoreBundle\Config\ResourceFinder`. The `targets` are hard-coded and compiled from [the documentation](https://docs.contao.org/dev/reference/dca/callbacks/). For example, the definition for the DCA callback `config.onrestore_version` is defined as such:

```php
[
    // ...

    'config.onrestore_version' => new CallbackDefinition(new MethodDefinition('void', [
        'table' => 'string',
        'pid' => 'int',
        'versionNumber' => 'int',
        'recordData' => 'array',
    ])),

    // ...
]
```

This allows us to create a specific `__invoke` method with all the parameters and the return type already set which is probably the thing that saves the most time when using this bundle. 

The generated `__invoke` method for above config looks like this:

```php
public function __invoke(string $table, int $pid, int $versionNumber, array $recordData): void
{
    // Do something …
}
```

A more complete example (suggestion lists omitted):

```
⇢ php bin/console make:contao:dca-callback

 Choose a class name for your callback:
 > MyDcaCallback

 Choose a table for this callback:
 > tl_page

 Choose a target for this callback:
 > fields.{field}.options

 Please choose a value for "field":
 > meep

 created: src/EventListener/MyDcaCallback.php
```

File generated:

```php
<?php

declare(strict_types=1);

namespace App\EventListener;

use Contao\DataContainer;

/**
 * @Callback(table="tl_page", target="fields.meep.options")
 */
class MyDcaCallback
{
    public function __invoke(DataContainer $dataContainer): array
    {
        // Do something …
    }
}
```

⚠️ Since the suggestion lists (and definitions) for hooks, events & callbacks are hard coded in the `maker-bundle`, we'll also need to maintain these lists, if we want to be able to auto-generate methods of said listeners.

`make:contao:dca-callback`, `make:contao:event-listener` and `make:contao:hook` all create a class file in `EventListener/` with the corresponding service annotation (`@Hook`...).

But both the commands`make:contao:content-element` and `make:contao:frontend-module` furthermore allow you to create dca-, translation- and template files.

An example for the creation of a content element:

```
⇢ php bin/console make:contao:content-element

 Choose a class name for your content element [GrumpyPopsicle]:
 > GrumpyPopsicle

 Do you want to add an empty DCA palette? (yes/no) [yes]:
 > yes

 Suggested Categories:
 * texts
 * accordion
 * slider
 * links
 * media
 * files
 * includes
 * custom

 Choose a category:
 > media

 Do you want to add translations? (yes/no) [yes]:
 > yes

 Choose an english name for this element:
 > My fancy element

 Choose an english description for this element:
 > A description for my fancy element                

 Do you want to add another translation? (yes/no) [no]:
 > yes

 What language do you want the element name translate to? (e.g. de):
 > de

 Choose a translated name for this element:
 > Mein schickes Inhaltselement

 Choose a translated description for this element:
 > Eine Beschreibung für mein schickes Inhaltselement

 Do you want to add another translation? (yes/no) [no]:
 > 

 updated: contao/dca/tl_content.php
 updated: contao/languages/en/default.xlf
 updated: contao/languages/de/default.xlf
 created: src/Controller/ContentElement/GrumpyPopsicle.php
 created: contao/templates/ce_grumpy_popsicle.html5
```

Since I already had a `contao/dca/tl_content.php`, a default palette was merged into the existing file:

```php
<?php

declare(strict_types=1);

// ... existing declarations

$GLOBALS['TL_DCA']['tl_content']['palettes']['grumpy_popsicle'] = '
    {type_legend},type,headline;
    {template_legend:hide},customTpl;
    {protected_legend:hide},protected;
    {expert_legend:hide},guests,cssID;
    {invisible_legend:hide},invisible,start,stop;
';
```

Same goes for the translation file (skipped the `de` one, but same principle applies):

```xml
<?xml version="1.0"?>
<xliff version="1.1">
  <file>
    <body>
      <!-- ... existing declarations -->
      <trans-unit id="CTE.grumpy_popsicle.0">
        <source>My fancy element</source>
      </trans-unit>
      <trans-unit id="CTE.grumpy_popsicle.1">
        <source>A description for my fancy element</source>
      </trans-unit>
    </body>
  </file>
</xliff>
```

New dca palettes are appended at the end of the file. XLF-Files will be merged together. It is important to note, that no existing keys will be overwritten. If `CTE.grumpy_popsicle.0` already existed, the new one will be skipped!

Furthermore, a controller class and template was created (see output log above):

```php
<?php

declare(strict_types=1);

use Contao\ContentModel;
use Contao\Template;

/**
 * @ContentElement("grumpy_popsicle", category="media")
 */
class GrumpyPopsicle extends AbstractContentElementController
{
    {
    }
}
```

```html
<?php $this->extend('block_searchable'); ?>

<?php $this->endblock(); ?>
```

Same goes for the creation process of a frontend module.

## Things to consider (⇢ RFC)

1. There is no real "error handling" when it comes to valid names for class names. It uses [`Str::asClassName`](https://github.com/symfony/maker-bundle/blob/f76abb1e4facd712a5cdbf35aff998e87c416801/src/Str.php#L61) internally, but you are very well allowed to provide the name `DataContainer` for the creation of a dca callback, which will result in the following code (stripped to the relevant part):

```php
<?php

use Contao\DataContainer;

class DataContainer
{
    public function __invoke(DataContainer $dataContainer): array
    {
    }
}
```

2. The XLF-Merger will dump the merged files with `tabsize=2` and I have not found a proper way to detect/configure/change this setting
3. In Contao, there are still some events that are dispatched with a name unrelated to the class name of the `Event` object (see all except the last one: https://docs.contao.org/dev/reference/events/). As soon as we refactor those events, we should auto generate listeners the proper way, since now the `event`-key is still provided in the annotation.
4. Since we're using the command registration infrastructure from `symfony/maker-bundle` (basically the tag `maker.command`) we also need to activate this bundle, otherwise the commands `make:contao:...` will not be registered. It could be useful to refactor this, so we at least don't need to active a second bundle.
5. We should find a better namespace name for the two services that currently resides in `Contao\MakerBundle\Code`
6. There is probably a lot more to add to the basic repository configuration files (monorepo/CI/...)

Sorry for the delay ❤️


maker-bundle/src/Maker/AbstractFragmentMaker.php
```diff
@@ -0,0 +1,174 @@
...
+        // Ask whether language files should be generated
+        $command->addArgument('addTranslation', InputArgument::OPTIONAL);
+        $question = new ConfirmationQuestion('Do you want to add translations?', true);
+        $input->setArgument('addTranslation', $io->askQuestion($question));
+
+        if ($input->getArgument('addTranslation')) {
+            $command
+                ->addArgument('translatedName', InputArgument::OPTIONAL, 'Choose an english name for this element')
+                ->addArgument('translatedDescription', InputArgument::OPTIONAL, 'Choose an english description for this element')
+                // FIXME: ->addArgument('languages', InputArgument::OPTIONAL, 'What languages do you want the element name translate to? (e.g. <fg=yellow>en, fr</>)')
```
leofeyer: This has been commented, so I have added a `FIXME` mark. Is this the reason why the PR is still a draft?

bytehead: May because of `Things to consider`?

monorepo.yml
```diff
@@ -60,6 +60,9 @@ repositories:
             81df735ec6aa9333baef09459f610bd3847ab946: bb713b10ef98dcaf52395e2cef8f4b7f682faec5
             a06d4de66922982fcadc9ee4a2b6d71a21171753: 51c69f1089e0f39fb6bfd7fb71cf2430ef5bde01
 
+    maker-bundle:
+        url: https://github.com/contao/manager-bundle.git
```
ausi:
```suggestion
        url: https://github.com/contao/maker-bundle.git
```

leofeyer:
> 1. There is no real "error handling" when it comes to valid names for class names. It uses [`Str::asClassName`](https://github.com/symfony/maker-bundle/blob/f76abb1e4facd712a5cdbf35aff998e87c416801/src/Str.php#L61) internally, but you are very well allowed to provide the name `DataContainer` for the creation of a dca callback

I don't think we have to handle this case. At least not now.

>  1. The XLF-Merger will dump the merged files with `tabsize=2` and I have not found a proper way to detect/configure/change this setting

Actually, `tabsize=2` is what we want to use. 😄 

> 3. Since we're using the command registration infrastructure from `symfony/maker-bundle` (basically the tag `maker.command`) we also need to activate this bundle, otherwise the commands `make:contao:...` will not be registered. It could be useful to refactor this, so we at least don't need to active a second bundle.

I don't think it hurts to activate the bundle. What are your concerns?

> 4. We should find a better namespace name for the two services that currently resides in `Contao\MakerBundle\Code`

I have renamed the namespace to `Contao\MakerBundle\Reflection` now.

leofeyer:
This is all nice and shiny now and ready for review. 🎉

Toflar (commented):
In terms of the coding style and architecture, I think there are no obvious issues so that's fine :) In terms of functionality, I cannot really say anything until I start using it.
But that shouldn't hold us back from merging.

The only thing that came to my mind is whether we should merge https://github.com/contao/contao/pull/3619 first and then either generate annotations or attributes depending on the PHP version used to run the `make` commands. I guess Symfony does the same here.

maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Contao\ContentModel;
+use Contao\CoreBundle\Controller\ContentElement\AbstractContentElementController;
+use Contao\CoreBundle\ServiceAnnotation\ContentElement;
+use Contao\Template;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+<?php if (PHP_VERSION_ID >= 80000): ?>
+#[AsContentElement("<?= $elementName ?>", "<?= $category ?>")]
```
leofeyer: @aschempp Is this OK or would you prefer `#[AsContentElement("<?= $elementName ?>", category: "<?= $category ?>")]`?

aschempp: I would prefer with named argument just to make it same as the annotation

leofeyer: Changed in cf1253ff9506af0e77d0e861296f86236eff5926.

leofeyer:
PHP 8 attributes support has been added in 9f0a8e87c94a93f483dc53e8dacf3c1d66019c7e. 🎉

maker-bundle/README.md
```diff
@@ -0,0 +1,46 @@
...
+composer require contao/maker-bundle --dev
+```
+
+## Usage
+
+This bundle provides several commands under the `make:` namespace. You can list
+them all with the following command:
+
+```
+php bin/console list make:contao
```
ausi:
```suggestion
php vendor/bin/contao-console list make:contao
```
I think we are using `vendor/bin/contao-console` in the Contao docs.

maker-bundle/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Available commands for the "make:contao" namespace:
+  make:contao:content-element  Creates an empty content element
+  make:contao:dca-callback     Creates a dca callback
+  make:contao:event-listener   Creates an event listener for a Contao event
+  make:contao:frontend-module  Creates an empty front end module
+  make:contao:hook             Creates a hook
```
ausi:
```suggestion
Available commands for the "make:contao" namespace:
  make:contao:content-element  Creates a new content element
  make:contao:dca-callback     Creates a new DCA callback listener
  make:contao:event-listener   Creates a new event listener
  make:contao:frontend-module  Creates a new front end module
  make:contao:hook             Creates a new hook listener
```
The command description seem to have changed ☺️

maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php
```diff
@@ -0,0 +1,31 @@
+<?= "<?php\n" ?>
+
+declare(strict_types=1);
+
+namespace App\Controller\ContentElement;
+
+use Contao\ContentModel;
+use Contao\CoreBundle\Controller\ContentElement\AbstractContentElementController;
+<?php if (PHP_VERSION_ID >= 80000): ?>
```
ausi: Checking the current PHP version can cause unwanted effects if the project itself allows a lower verion. Can we instead read the php requirement from the root *composer.json* of the project?

maker-bundle/src/Resources/config/hooks.yaml
```diff
@@ -0,0 +1,566 @@
...
+            file: array
+
+    getContentElement:
+        return_type: string
+        arguments:
+            contentModel: Contao\ContentModel
+            buffer: string
+            contentElement: Contao\ContentElement
+
+    getCountries:
```
ausi, line 204: Some of these hooks are already deprecated. Might be helpful if we can show that in the maker too IMO.
(But maybe in a seperate pull request)

maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php
leofeyer: There is a helper class in the maker bundle that reads the `composer.lock` file.

https://github.com/symfony/maker-bundle/blob/716eee9c8b10b33e682df1b7d80b9061887e9691/src/Util/PhpCompatUtil.php#L32-L37

However, the helper only checks `platform-overrides`. Would this suffice in our case, too?

ausi: I think it would be sufficient for this edge case, because if someone has this problem in a project, they can then fix it by adding `platform-overrides` to their composer.json.

leofeyer: Implemented in 39c3af3735efacaedeca0f35df7c599345d23a04.

ausi (approved):

PR #3616, opened 2021-10-26
Support canonical URLs in the front end

Toflar:
This includes most of the features discussed in https://github.com/contao/contao/issues/3097#issuecomment-952045251


Toflar:
This is ready for review :)

core-bundle/src/Routing/ResponseContext/HtmlHeadBag/HtmlHeadBag.php
```diff
@@ -53,4 +57,55 @@ public function setMetaRobots(string $metaRobots): self
...
+    {
+        if ($this->canonicalUri) {
+            return $this->canonicalUri;
+        }
+
+        $params = [];
+
+        foreach ($request->query->all() as $originalParam => $value) {
+            foreach ($this->getKeepParamsForCanonical() as $param) {
+                $regex = sprintf('/^%s$/', str_replace('*', '.*', $param));
```
m-vo:
```suggestion
                $regex = sprintf('/^%s$/', str_replace('\*', '.*', preg_quote($param, '/')));
```

core-bundle/src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -26,6 +26,24 @@
...
+        <source>Manual rel="canonical" selection</source>
+      </trans-unit>
+      <trans-unit id="tl_page.canonicalLink.1">
+        <source>You can manually select the desired target page for rel="canonical" support.</source>
+      </trans-unit>
+      <trans-unit id="tl_page.canonicalKeepParams.0">
+        <source>Query parameters to keep for rel="canonical"</source>
+      </trans-unit>
+      <trans-unit id="tl_page.canonicalKeepParams.1">
+        <source>By default, Contao points to itself for rel="canonical". You may add a comma-separated list of query parameters which should be kept. Use "*" as wildcard.</source>
```
fritzmg:
```suggestion
        <source>By default, Contao points to itself for rel="canonical". You may add a comma-separated list of query parameters which should be kept. Use "*" as a wildcard.</source>
```

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
```diff
@@ -13,6 +13,12 @@
       <meta name="generator" content="Contao Open Source CMS">
     <?php $this->endblock(); ?>
 
+    <?php $this->block('canonical'); ?>
+    <?php if ($this->canonical): ?>
+    <link rel="canonical" href="<?= $this->canonical; ?>">
+    <?php endif; ?>
+    <?php $this->endblock(); ?>
```
fritzmg:
```suggestion
    <?php $this->block('canonical'); ?>
      <?php if ($this->canonical): ?>
        <link rel="canonical" href="<?= $this->canonical; ?>">
      <?php endif; ?>
    <?php $this->endblock(); ?>
```

Btw. what if we used `$GLOBALS['TL_HEAD'][]` instead for anything that is added via the `HtmlHeaderBag`? This would then not require a `fe_page` adjustment.

fritzmg (changes requested):

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
Toflar: Yeah, that's planned for the future ;)

core-bundle/src/Routing/ResponseContext/HtmlHeadBag/HtmlHeadBag.php
```diff
@@ -53,4 +57,55 @@ public function setMetaRobots(string $metaRobots): self
...
+    {
+        $this->canonicalUri = $canonicalUri;
+
+        return $this;
+    }
+
+    public function getCanonicalUri(Request $request): string
+    {
+        if ($this->canonicalUri) {
+            return $this->canonicalUri;
```
ausi: Should we let Symfony normalize this URL too?

ausi (dismissed):

core-bundle/src/Routing/ResponseContext/HtmlHeadBag/HtmlHeadBag.php
Toflar: Good idea, done in 627ab4868dcc4cbbb1af27e932fae9ae7fd39710.

ausi (dismissed):

fritzmg (dismissed):

leofeyer:
@Toflar Can you resolve the conflicts please?

Toflar:
Done.

Toflar:
Failing tests unrelated.

ausi (dismissed):

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
```diff
@@ -13,6 +13,12 @@
       <meta name="generator" content="Contao Open Source CMS">
     <?php $this->endblock(); ?>
 
+    <?php $this->block('canonical'); ?>
```
leofeyer: Why does it have to be a block? It seems very unlikely to me that someone needs to replace it with a custom implementation.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -85,6 +90,26 @@ public function createContaoWebpageResponseContext(PageModel $pageModel): Respon
             $htmlHeadBag->setMetaRobots($pageModel->robots);
         }
 
+        if ($pageModel->enableCanonical && $pageModel->canonicalLink) {
+            /** @var Controller $controller */
+            $controller = $this->contaoFramework->getAdapter(Controller::class);
+            $url = $controller->replaceInsertTags($pageModel->canonicalLink, false);
```
leofeyer, line 99: `Controller::replaceInsertTags` is deprecated. Can we use `(new InsertTags())->replace()` here instead?

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
Toflar: The more blocks, the better. The `fe_page` is our fundamental template so it should be absolutely stuffed with blocks imho.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
```diff
@@ -85,6 +90,26 @@ public function createContaoWebpageResponseContext(PageModel $pageModel): Respon
...
+
+            // Ensure absolute links
+            // FIXME: Remove once we remove support for relative urls
+            if (!preg_match('#^https?://#', $url)) {
+                /** @var Environment $environment */
+                $environment = $this->contaoFramework->getAdapter(Environment::class);
+                $url = $environment->get('base').$url;
```
leofeyer: This should be possible without the `Environment` class. I can try to make it work if you want.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: No we can't because we cannot mock that. We will be able to replace this with @ausi's new service he's building in https://github.com/contao/contao/pull/3638

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
Toflar: Feel free.

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
fritzmg: Imho it is likely that this may need to be replaced. If, for example, you have another solution to add a canonical tag via `$GLOBALS['TL_HEAD']`, you can simply add

```php
<?php $this->block('canonical') ?>
<?php $this->endblock(); ?>
```

to your `fe_page` template, in order to disable the built in solution. Of course you could also overwrite the `$this->canonical` template variable via a `parseTemplate` hook instead. Or use `$this->canonical = null` somewhere in the beginning of the template.

core-bundle/src/Routing/ResponseContext/CoreResponseContextFactory.php
leofeyer: See https://github.com/Toflar/contao/pull/7

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
leofeyer: If someone implements a custom canonical logic, it is very likely that they will only overwrite `$this->canonical` and still use the `<link rel="canonical" href="<?= $this->canonical; ?>">` markup – because there is no other way to set the canonical URL in an HTML document!

Therefore, a block is not needed here.

fritzmg (dismissed):

ausi (dismissed):

aschempp (commented):
Looks very good! One particular problem I see is that it is not possible to multi-edit the canonical information (because the fields won't be added because `$dc->id` isn't there).

One idea to work around this would be to always show the fields in the page configuration. And then either hide them through the current onload-callback. Or maybe we could even say if a canonicalLink was entered manually, we always show that regardless of root settings?

leofeyer:
Maybe we should just disable the fields like we do in the `DisableAppConfiguredSettingsListener`? Then they will always be visible, even in "override all" mode.

PR is here: https://github.com/Toflar/contao/pull/9

core-bundle/src/Resources/contao/templates/frontend/fe_page.html5
leofeyer: https://github.com/Toflar/contao/pull/10

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3617, opened 2021-10-26
Do not index documents if the canonical URL does not match

Toflar:
https://github.com/contao/contao/issues/3097#issuecomment-952045251


fritzmg (approved):

fritzmg:
Should we may be backport this to `4.9` as well? I would consider it a bug that non-matching canonical URLs can be added to the search index currently.

Toflar:
I asked and we decided not to :)

fritzmg:
Ah, I missed that then ;)

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #3618, opened 2021-10-26
Introduce FragmentResponseStack service

fritzmg:
This introduces a `FragmentResponseStack` service as discussed on the developer meeting. The `SubrequestCacheSubscriber` now doesn't implement the `CacheResponseStrategy` stack itself, but instead uses the `FragmentResponseStack` service and calls `init()` for each main request, `add()` for each sub response and `finalize()` for each main response.

The latter then merges all the `Cache-Control` headers of all responses of the current stack.

This service can then be used in other places to add additional `Cache-Control` response headers to be merged into the main request (e.g. from insert tags or from legacy content elements, modules and articles).


Toflar:
I like this concept 👍

core-bundle/src/Fragment/FragmentResponseStack.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @var array<FragmentResponseCollection>
+     */
+    private array $collectionStack = [];
+
+    /**
+     * Initializes and pushes new fragment response collection for a new main request.
+     */
+    public function init(): void
```
aschempp, line 33: `init` sounds like something that should only be done once. Maybe use `start()`?

core-bundle/src/Fragment/FragmentResponseCollection.php
```diff
@@ -0,0 +1,36 @@
...
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Fragment;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class FragmentResponseCollection
```
aschempp, line 17: not sure why we need a class that does nothing else than a simple array would? also, why not just use the `ResponseCacheStrategy`?

core-bundle/src/Fragment/FragmentResponseStack.php
```diff
@@ -0,0 +1,90 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Fragment;
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\HttpCache\ResponseCacheStrategy;
+use Symfony\Contracts\Service\ResetInterface;
+
+final class FragmentResponseStack implements ResetInterface
```
aschempp, line 19: since we want to use this for other stuff than fragments, maybe something like `ResponseCacheStack` or `ResponseCacheCollection` in relation to `ResponseCacheStrategy`?

core-bundle/src/Fragment/FragmentResponseStack.php
```diff
@@ -0,0 +1,90 @@
...
+            return;
+        }
+
+        $this->currentCollection->add($response);
+    }
+
+    /**
+     * Finalizes the current fragment response collection on the given response.
+     */
+    public function finalize(Response $response): void
```
aschempp, line 57: maybe change to `finish` so the workflow is `->start() -> add() -> finish()` 🙃 
if it's meant to be semantically meaningful, it should probably be `mergeCacheHeaders` (and `collectCacheHeaders()`)

core-bundle/src/Fragment/FragmentResponseStack.php
aschempp, line 19: I don't think we use `final` for service classes, only for POPOs atm.

aschempp (changes requested):
concept looks good to me, but I think it could be simplified a bit 🙃

core-bundle/src/Fragment/FragmentResponseCollection.php
fritzmg: > also, why not just use the `ResponseCacheStrategy`?

We want to hold arbitrary responses first and then finalize them - e.g. apply a `ResponseCacheStrategy` according to the collected responses. But I am fine with removing the collection class and just using arrays of arrays (holding Responses) for the stack instead.

core-bundle/src/Fragment/FragmentResponseStack.php
fritzmg: Sounds good to me 👍

core-bundle/src/Fragment/FragmentResponseCollection.php
aschempp: what I meant was that that's exactly what the `ResponseCacheStrategy` does internally (and currently does directly in the listener) 🙃

core-bundle/src/Fragment/FragmentResponseStack.php
m-vo: The `ContaoMailer` is final as well (but also implements the `MailerInterface`). :upside_down_face:

core-bundle/src/Fragment/FragmentResponseStack.php
m-vo: Would it make sense to throw exceptions if you'd for instance try to call `add` on a finialized or `finalize` on a not yet started stack? So that configuration errors won't slip through.

fritzmg: @m-vo the first one would not work, because the stack is popped during `finalize`. Any subsequent `add` would then add the response to that stack, if available. There is no way of knowing whether that response does not belong to that stack, is there?

Toflar (commented):
Are you finishing this @fritzmg ? :)

PR #3619, opened 2021-10-26
Add PHP8 attributes for our existing service annotations

aschempp:
Replaces https://github.com/terminal42/service-annotation-bundle/pull/6

The new naming follows the [Symfony convention](https://symfony.com/blog/new-in-symfony-5-3-service-autoconfiguration-and-attributes#autoconfigurable-attributes)

> These attribute names follow the pattern #[As...]. Symfony 5.3 provides #[AsCommand] to define a PHP class as a Symfony console command, #[AsEventListener] to define a PHP class as a Symfony event listener, etc.


core-bundle/src/DependencyInjection/Attribute/AsPage.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace Contao\CoreBundle\DependencyInjection\Attribute;
+
+/**
+ * An attribute class for page controllers.
+ *
+ * @see \Symfony\Component\Routing\Annotation\Route
+ */
+class AsPage
```
leofeyer, line 21: No `#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]` here?

leofeyer:
Can you give an example of how to use the new feature?

aschempp:
```
// register hook
#[AsHook("outputFrontendPage")]

// registering two hooks for different methods
#[AsHook("outputFrontendPage", method: "foo")]
#[AsHook("parseTemplate", method: "bar")]

// register content element
#[AsContentElement(category: "texts")]
#[AsContentElement("my_type", category: "texts")]
```

Using named arguments you can skip the optional arguments 😊  
Unfortunately, Symfonys service attributes are not supported on methods, but the same can be done through the method argument.

core-bundle/src/DependencyInjection/Attribute/AsPage.php
aschempp: fixed in 44c5d82c75a9f8fdeaaef8760e97c51f702f34f5

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -119,6 +129,55 @@ public function load(array $configs, ContainerBuilder $container): void
             ->registerForAutoconfiguration(ContentCompositionInterface::class)
             ->addTag('contao.page')
         ;
+
+        $container->registerAttributeForAutoconfiguration(
+            AsContentElement::class,
+            static function (ChildDefinition $definition, AsContentElement $attribute): void {
+                $definition->addTag(ContentElementReference::TAG_NAME, $attribute->attributes);
```
leofeyer, line 136: I haven't seen `$attribute->attributes` here before. Isn't it usually `get_object_vars($attribute)`?

leofeyer (commented):
Maybe I am missing something but the constructor arguments do not match the annotations in the `ServiceAnnotation` namespace, do they? E. g.

```
// AsContentElement
string $type = null,
string $category,
string $template = null,
string $method = null,
string $renderer = null,
array ...$attributes
```

```
// ContentElement
@Attributes({
    @Attribute("value", type = "string"),
    @Attribute("category", required = true, type = "string"),
    @Attribute("template", type = "string"),
    @Attribute("renderer", type = "string"),
    @Attribute("attributes", type = "array"),
})
```

Is this correct?

---

**Edit:** As discussed with @aschempp, this is correct. The `value` attribute corresponds to the `$type` property and the `$method` property is used to assign the attribute to a method, because

> Unfortunately, Symfonys service attributes are not supported on methods, but the same can be done through the method argument.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
aschempp: it can be done however we want. I found Symfony to use `get_object_vars` because they define public properties for the possible tag attributes. In the case of content elements and frontend modules, we allow all unknown properties and forward them to the fragment/renderer. In annotations, this was only possible through an array of attributes as a last-resort. With PHP though we can use unlimited named arguments for these attributes 😎

core-bundle/src/DependencyInjection/Attribute/AsContentElement.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Service tag to autoconfigure content elements.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+class AsContentElement
+{
+    public array $attributes;
+
+    public function __construct(string $type = null, string $category, string $template = null, string $method = null, string $renderer = null, array ...$attributes)
```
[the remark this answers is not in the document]
aschempp: unfortunately, this is correct. The type is automatically inferred from the class name, but the category must be set. If one enters all arguments in the list, the type should be first, as it was with the annotation. But one can use names arguments to only set the category, exactly like now in annotations.

core-bundle/src/DependencyInjection/Attribute/AsPage.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * An attribute class for page controllers.
+ *
+ * @see \Symfony\Component\Routing\Annotation\Route
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+class AsPage
+{
+    public function __construct(public ?string $type = null, public bool $contentComposition = true, public ?string $path = null, public ?string $urlSuffix = null, public array $requirements = [], public array $options = [], public array $defaults = [], public array $methods = [])
+    {
```
leofeyer, line 24: The `Page` service annotation also has logic in its constructor:

https://github.com/contao/contao/blob/cf615b4640f0a00c1e2a19f5604cb1b86a6d8c4f/core-bundle/src/ServiceAnnotation/Page.php#L39-L57

Is this not required here, too?

core-bundle/src/DependencyInjection/Attribute/AsCronJob.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Contao\CoreBundle\DependencyInjection\Attribute;
+
+/**
+ * An attribute to register a Contao cron job.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+class AsCronJob
+{
+    public function __construct(public string $interval, public ?string $method = null)
+    {
```
leofeyer, line 22: The CronJob service annotation also has the following logic:

https://github.com/contao/contao/blob/cf615b4640f0a00c1e2a19f5604cb1b86a6d8c4f/core-bundle/src/ServiceAnnotation/CronJob.php#L40-L41

aschempp: I don't think we need to escape values in attributes anymore?

fritzmg: Yeah, this was only necessary for annotations since they are in PHP comments. You cannot write

```php
/**
 * @CronJob(*/5 * * *)
 */
```

for example, since `*/` closes the comment. You need to escape the `*` or `/`:

```php
/**
 * @CronJob(*\/5 * * *)
 */
```

With PHP attributes this is (presumably) not necessary.

leofeyer (approved):

PR #3620, opened 2021-10-26
Deprecate the request_token insert tag

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | see https://github.com/contao/docs/pull/992

On the route to a better insert tag handling: This deprecates the `{{request_token}}` insert tag. The value should be provided by the controller instead.


Toflar (dismissed):

bytehead (dismissed):

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -44,16 +45,18 @@ class ModuleEventReader extends Events
...
-		if ($request && System::getContainer()->get('contao.routing.scope_matcher')->isBackendRequest($request))
+		if ($request && $container->get('contao.routing.scope_matcher')->isBackendRequest($request))
 		{
 			$objTemplate = new BackendTemplate('be_wildcard');
 			$objTemplate->wildcard = '### ' . $GLOBALS['TL_LANG']['FMD']['eventreader'][0] . ' ###';
 			$objTemplate->title = $this->headline;
 			$objTemplate->id = $this->id;
 			$objTemplate->link = $this->name;
 			$objTemplate->href = 'contao/main.php?do=themes&amp;table=tl_module&amp;act=edit&amp;id=' . $this->id;
+			$objTemplate->requestToken = $container->get(ContaoCsrfTokenManager::class)->getToken($container->getParameter('contao.csrf_token_name'))->getValue();
```
fritzmg: This only sets the `requestToken` variable for the back end wildcard template, not for the actual template used in the front end.

fritzmg (changes requested):

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
m-vo: Thanks for checking. Super silly mistake, changed in 370f3a5b2c2d8d6ec7be15f9acb20dad99681ff0. :facepalm:

ausi (dismissed):

aschempp (changes requested):
I totally understand the reason behind this, but its very bad DX to me. Can we maybe add a `ContaoCsrfTokenManager->getFrontendTokenValue()` as a shortcut?

aschempp (dismissed):
Even more so, since this is always used in (frontend) templates, we might want a `FrontendTemplate::getRequestToken()` method instead?

m-vo:
> Can we maybe add a ContaoCsrfTokenManager->getFrontendTokenValue() as a shortcut?

That's a very good point. :+1: You would probably always use the `contao.csrf_token_name` parameter with the `ContaoCsrfTokenManager` in both new and old code. I'm going to refactor this if noone is screaming stop. :smile: 

> we might want a FrontendTemplate::getRequestToken() method instead?

I'm not sure about this one. I'd first make the changes in this PR and then - if a need arises - implement a Twig function and a Template function using the service. Maybe I'm wrong but I feel injecting the token should be the responsibility of the controller.

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
```diff
@@ -60,4 +62,13 @@ public function canSkipTokenValidation(Request $request, string $tokenCookieName
             && !($request->hasSession() && $request->getSession()->isStarted())
         ;
     }
+
+    public function getFrontendTokenValue(): string
```
m-vo, line 66: We could additionally overwrite `getToken(string $tokenName)` with `getToken(string $tokenName = null)` and use the configured frontend token name when the `$tokenName` argument wasn't set.

fritzmg:
> Can we maybe add a `ContaoCsrfTokenManager->getFrontendTokenValue()` as a shortcut?

Yes please 👍 I always found the current way very cumbersome (always having to inject both the parameter and the manager in every service where I need a request token - which also always needs an additional `bind` configuration for autowiring in each new project).

m-vo:
> Yes please +1

see https://github.com/contao/contao/commit/409b33cdd625dd8ac2d5fa7c1751a594fe503d69 :slightly_smiling_face:

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
fritzmg: I would prefer this 👍

aschempp: I think having a separate method makes this more explicit. Not passing a value basically means _give me what you want_, which is not the same. But it would probably be acceptable for the developer 😂

aschempp:
> Maybe I'm wrong but I feel injecting the token should be the responsibility of the controller.

I don't think thats the case. The controller does not (need to) know about the request token, only the form in the template needs to have it imho.

Toflar:
I agree it's the responsibility of the controller. Because the check **should** not be global on every POST request but instead, every form should check individually, like e.g. Symfony Forms does.

leofeyer:
@m-vo Does your implementation still work with cached pages? I think that was our primary motivation to use the insert tag, so it is correctly replaced even if the page is loaded from cache.

m-vo:
> Does your implementation still work with cached pages? 

IMHO nothing changes, because our `CsrfTokenCookieSubscriber` just strips the tokens (the real representation, not the insert tag) from the rendered response (see L141):

https://github.com/contao/contao/blob/109441dbb4db438832ecd27db399c20d06476fe2/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php#L127-L153

leofeyer:
If a page with a request token is cached for one year, I doubt that the token value will still be in `$this->tokenStorage->getUsedTokens()` after 6 months or longer, will it?

ausi:
> If a page with a request token is cached for one year, I doubt that the token value will still be in `$this->tokenStorage->getUsedTokens()` after 6 months or longer, will it?

The request token gets removed before it is stored in the cache. `$this->tokenStorage->getUsedTokens()` only lives in memory, so only for the current request.

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
m-vo: I agree with both. Should we vote?

:+1: = just the `getFrontendTokenValue()` method like it is currently implemented

:tada: = adjust `getToken(string $tokenName = null)` and return the front end token if `null` was given

:rocket: = same + `getFrontendTokenValue()` method that basically returns `$this->getToken()->getValue()`

ausi (approved):

leofeyer:
@m-vo There are some conflicting files now. 🙈 Can you please resolve these conflicts?

m-vo:
@contao/developers  Before I resolve conflicts, any opinion on this https://github.com/contao/contao/pull/3620#discussion_r739650183?

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
Toflar: Always be explicit. We just had trouble with people doing `Model::findByPk(null)` - let's not repeat the story.

m-vo: So then you intended to vote for :+1: instead of :rocket:?

Toflar: 🤦 yeah of course :D

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -714,6 +714,8 @@ static function ($allowedTag)
 
 				// Request token
 				case 'request_token':
+					trigger_deprecation('contao/core-bundle', '4.13', 'Using the request_token insert tag has been deprecated and will no longer work in Contao 5.0.');
+
 					$arrCache[$strTag] = REQUEST_TOKEN;
```
Toflar, line 719: Should probably also just take the service so we can get rid of all the usages of the constants as well?

Toflar (commented):
One final nitpick before I approve :)

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
m-vo: The `REQUEST_TOKEN` constant is currently used 100+ times in the code base. Let's refactor this in another PR, wdyt?

Toflar (approved):

fritzmg (approved):

leofeyer:
Thank you @m-vo.

PR #3622, opened 2021-10-26
Make replacing insert tags more granular

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | -
| Docs PR or issue | -

Make replacing insert tag more granular. And potentially destroy hopes.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
@@ -127,8 +127,6 @@ class FrontendTemplate extends Template
 			}
 		}
 
-		// Replace insert tags
-		$this->strBuffer = $this->replaceInsertTags($this->strBuffer);
 		$this->strBuffer = $this->replaceDynamicScriptTags($this->strBuffer); // see #4203
 
 		// HOOK: allow to modify the compiled markup (see #4291)
--- a/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
+++ b/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
@@ -159,6 +159,9 @@ trait TemplateInheritance
 			}
 		}
 
+		// Replace insert tags
+		$strBuffer = Controller::replaceInsertTags($strBuffer);
+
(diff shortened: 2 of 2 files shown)


m-vo:
> Make replacing insert tag more granular.

Some background info: Insert tags are currently replaced at a very late stage. At this point all semantics are gone and we're left with a string. Replacing anything on strings is a dangerous task and doing it recursively (like we're doing it with insert tags) can yield all sorts of problems. We do not even know the context (html? js? html attributes? user content? trusted?  …) anymore.

The correct solution is to move replacing insert tags closer to the actual content they reside in. This is why rendering Twig templates won't replace insert tags on the rendered output at all (it can be explicitly enabled *per variable*, see #3606). Using the old template engine, this PR will make sure they're only replaced *per template*, i.e. after the template was parsed. 

> And potentially destroy hopes.

This will effectively disallow insert tags that span across multiple content elements* like so:

```
// HTML element 1
{{iflng::en}}

// Text element
foo bar

// HTML element 2
{{iflng}}
```

Spoiler alert: For this use case there's a better variant around the door: #3613 

*) the effect of it not working anymore will only become noticeable as soon as the templates rendering the parent templates (e.g. `mod_article`, `fe_page`) are rendered with Twig and **not** replacing insert tags anymore.

aschempp (changes requested):
Looks good to me!

Just requesting changes to not get this merged until we get https://github.com/contao/contao/pull/3601 sorted. Currently fragment templates already do replace insert tags, which might collide with some of the changes here.

ausi (approved):
I don’t see conflicts with #3601 here

leofeyer:
Thank you @m-vo.

fritzmg:
The changes in this PR essentially reverted a change from #3601. The change in #3601 would have made it so that insert tags would not be replaced when calling `FragmentTemplate::getResponse`. However, in this PR the insert tag replacement was moved into `inherit` which will _always_ be called when any template is parsed.

This in general causes insert tags to be always replaced in the back end now (as well as when calling `FragmentTemplate::getResponse`) - which causes various issues (e.g. the picker will not return an insert tag anymore, but the already replaced URL for a page).

PR #3625, opened 2021-10-27
Add a template element and module

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/core-bundle/pull/1315#issuecomment-358484679
| Docs PR or issue | contao/docs#866


m-vo (dismissed):
Not sure about `category: texts` but I do not have a better idea either. :slightly_smiling_face:

bytehead:
maybe `misc` 🙃

bytehead (dismissed):

ausi:
> Not sure about `category: texts`

I followed the convention of the HTML element which is also in the `texts` category.

bytehead (dismissed):

m-vo (dismissed):

aschempp:
I don't understand what this element is supposed to be useful for, maybe you can explain?

ausi:
> I don't understand what this element is supposed to be useful for, maybe you can explain?

https://github.com/contao/core-bundle/pull/1315#issuecomment-358484679

fritzmg:
Personally I would put it into the `includes` category, as you "include" a template.

Will there also be a template module?

ausi:
> Will there also be a template module?

Would make sense IMO. 
@contao/developers please vote 👍 👎

aschempp:
> > I don't understand what this element is supposed to be useful for, maybe you can explain?
> 
> [contao/core-bundle#1315 (comment)](https://github.com/contao/core-bundle/pull/1315#issuecomment-358484679)

well that pull request was merged. The only thing "missing" was your mention of _a key value wizard_ that noone took up to in that PR 😅 .
I use HTML modules a lot since then, but what I meant was what would the key value wizard be useful for? If I want the user to enter data in the back end, I would create an content element for that (or use your awesome custom-elements extension). But having a key value wizard **without** knowing what keys/values are available seems counterintuitive to me.

ausi:
> If I want the user to enter data in the back end, I would create an content element for that

The same could be said about the usage of the HTML element with custom templates. You could create a custom element for that instead too.

> what would the key value wizard be useful for?

Arbitrary data that the template can use however it wants.

aschempp:
> > If I want the user to enter data in the back end, I would create an content element for that
> 
> The same could be said about the usage of the HTML element with custom templates. You could create a custom element for that instead too.

yes, thats exactly my point. The HTML element and this template element are functionally equal. Contrary to elements that have settings. The only difference I spot is the "arbitrary key/value wizard". Maybe you can give a real life example what this would be useful for? I honestly don't want people to start using a key value wizard and tell their customers to "fill this and that in" instead of using custom elements 🤷

ausi:
> Will there also be a template module?

Done in 1fec68a5ace0ad42c0cdaeed9037b9236bccdea9

m-vo (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

fkaminski:
Based on the 4.13.0 RC1, the content element is called "**Eigenes Template**" in German.
As a module, this can then be found under "**Individuelles Template**".

Maybe the name should be consistent?

m-vo:
Agreed, but that's not related to the code but the translations managed on Transifex.

fritzmg:
> Agreed, but that's not related to the code but the translations managed on Transifex.

Translations issues should still be reported here, as there is no other way to do it. @fkaminski may be open a separate issue about this.

fkaminski:
see #3950

PR #3626, opened 2021-10-27
Support arbitrary classes as Twig context

### LINKED ISSUE #3565: Support Twig for widgets
**Affected version(s)**

Contao 4.12

**Description**

Currently, `TemplateInheritance#renderTwigSurrogateIfExists()` only renders a surrogate if the object is a `Template` instance. We're doing this so we can reliably build the right context for the Twig version (injecting `$arrData` making the methods available and so on…).

https://github.com/contao/contao/blob/82286f4c90e3bf2bc7857ddca6db7823fe271738/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php#L365-L368

We're also having widgets, though that are using the `TemplateInheritance`.

```php
abstract class Widget extends Controller
{
	use TemplateInheritance;

	// …
}
```

This for instance means it's currently not possible to overwrite `form_textfield` with `form_textfield.html.twig`. 

We can of course allow other classes as well in the check but we must then also make sure we're building the right context e.g. with a new `ContextHelper::fromContaoWidget()` method.

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3565 
| Docs PR or issue | Todo

This PR enables creating Twig contexts from arbitrary classes. Examples being our `Widget` class for forms and other things. This task isn't trivial as templates in the old system are *directly* accessing anything (including private functions/members/… and dynamic assignments) from within the object itself.

One bummer: I did not find a way to support `__get` access, because the virtual names (`label` instead of `strlabel`) cannot be discovered programmatically.

This PR also refactors the `ContextHelper` into a service.


m-vo:
> One bummer: I did not find a way to support __get access, because the virtual names (label instead of strlabel) cannot be discovered programmatically.

The object now gets stored under `data` as well.
So you could access magic methods for instance by writing `{{ data.label }}`.

core-bundle/src/Twig/Interop/ContextFactory.php
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Find all members including those that were dynamically set ($this->foo = 'bar').
+     */
+    private function getAllMembers(object $object): \Generator
+    {
+        // See https://externals.io/message/105697#105697
+        $mangledObjectVars = \function_exists('get_mangled_object_vars') ?
+            get_mangled_object_vars($object) : (array) $object;
```
ausi:
```suggestion
        // See https://externals.io/message/105697#105697
        // Backwards compatibility with PHP < 7.4
        $mangledObjectVars = \function_exists('get_mangled_object_vars') ?
            get_mangled_object_vars($object) : (array) $object;
```
Comment so that we don’t forget to remove this in the future?

core-bundle/src/Twig/Interop/ContextFactory.php
```diff
@@ -0,0 +1,159 @@
...
+
+        foreach ($mangledObjectVars as $key => $value) {
+            if (preg_match('/^\x0(\S+)\x0(\S+)$/', $key, $matches)) {
+                // Protected or private member
+                if ('*' === $matches[1] || \get_class($object) === $matches[1]) {
+                    yield $matches[2] => $value;
+                    continue;
+                }
+
+                throw new \InvalidArgumentException('Could not identify prepended key.');
```
ausi: Wouldn’t this now fail for every private member of a parent class?

core-bundle/src/Twig/Interop/ContextFactory.php
m-vo: Sounds good!

core-bundle/src/Twig/Interop/ContextFactory.php
m-vo: After a discussion with @ausi on slack, we came to the conclusion that private members, functions and constants should be ignored. I refactored the code in cf2d5b6c5.

core-bundle/src/Twig/Interop/ContextFactory.php
```diff
@@ -0,0 +1,175 @@
...
+            if (0 === strpos($name, '__')) {
+                continue;
+            }
+
+            $context[$name] = $this->getCallableWrapper($method->getClosure($object), $name);
+        }
+
+        if (!isset($context['data'])) {
+            $context['data'] = $object;
+        }
```
ausi: This could easily result in naming conflicts I would guess. Can we use a more uncommon (for template variables) name. How about using `this`?

core-bundle/src/Twig/Interop/ContextFactory.php
```diff
@@ -0,0 +1,175 @@
...
+                // Protected member
+                if ('*' === $matches[1]) {
+                    yield $matches[2] => $value;
+                }
+
+                continue;
+            }
+
+            // Public or dynamic member
+            yield $key => $value;
```
ausi:
```suggestion
            if (strncmp($key, "\0*\0", 3) === 0) {
                // Protected member
                $key = substr($key, 3);
            }

            if ("\0" !== $key[0]) {
                yield $key => $value;
            }
```
Nitpick ☺️

ausi (approved):
❤️

core-bundle/src/Twig/Interop/ContextFactory.php
m-vo: Right, after the private members are gone, this looks simpler now :smile:

core-bundle/src/Twig/Interop/ContextFactory.php
m-vo: Agreed. Changed in 98419a429.

ausi (approved):

core-bundle/tests/Twig/TwigIntegrationTest.php
```diff
@@ -0,0 +1,84 @@
...
+        (new Filesystem())->remove(Path::join($this->getTempDir(), 'templates'));
+
+        TemplateLoader::reset();
+
+        unset($GLOBALS['TL_LANG'], $GLOBALS['TL_FFL']);
+    }
+
+    public function testRendersWidgets(): void
+    {
+        $content = "{{ strClass }}\n{{ strLabel }} {{ data.label }}\n {{ getErrorAsString }}";
```
ausi:
```suggestion
        $content = "{{ strClass }}\n{{ strLabel }} {{ this.label }}\n {{ getErrorAsString }}";
```

m-vo: oops :see_no_evil:

ausi (dismissed):

core-bundle/src/Resources/config/services.yml
```diff
@@ -921,6 +921,8 @@ services:
             - '@Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader'
         public: true
 
+    Contao\CoreBundle\Twig\Interop\ContextFactory: ~
```
leofeyer: With regards to #3686 and upstream mergeability, this should be `contao.twig.context_factory`.

leofeyer (changes requested):

core-bundle/src/Resources/config/services.yml
m-vo: Changed in 815b891.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3627, opened 2021-10-28
Add the missing dot-env features

aschempp:
This enables the Contao Manager to know what configuration can be done in the `.env` file.

PS: this needs to be updated after merging into 4.12

### DIFF
--- a/manager-bundle/src/ContaoManager/Plugin.php
+++ b/manager-bundle/src/ContaoManager/Plugin.php
@@ -197,6 +197,13 @@ class Plugin implements BundlePluginInterface, ConfigPluginInterface, RoutingPlu
     {
         return [
             'dot-env' => [
+                'APP_SECRET',
+                'APP_ENV',
+                'COOKIE_WHITELIST',
+                'DATABASE_URL',
+                'DISABLE_HTTP_CACHE',
+                'MAILER_URL',
+                'TRACE_LEVEL',
                 'TRUSTED_PROXIES',
                 'TRUSTED_HOSTS',
             ],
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

Toflar (approved):

bytehead (approved):

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -197,6 +197,13 @@ public function getApiFeatures(): array
     {
         return [
             'dot-env' => [
+                'APP_SECRET',
+                'APP_ENV',
+                'COOKIE_WHITELIST',
+                'DATABASE_URL',
+                'DISABLE_HTTP_CACHE',
+                'MAILER_URL',
```
fritzmg, line 205: You can also use `MAILER_DSN` (as this is the Symfony Standard). Though not sure if we need to support _every_ possibility anyway.

aschempp: I don't think we could have used that in Contao 4.4 right? The Manager must know what's supported, and this config is the only way to tell 🙃

fritzmg: No, not in 4.4, not even 4.9 even I think. Only since 4.10.

aschempp: Thats why we need to announce it in the plugin. It seems to be there in Contao 4.9 though: 
https://github.com/contao/contao/blob/4.9/manager-bundle/src/Resources/skeleton/config/config.yml#L80
https://github.com/contao/contao/blob/4.9/manager-bundle/src/ContaoManager/Plugin.php#L250

fritzmg: > It seems to be there in Contao 4.9 though:

Yes, `MAILER_URL`, but not `MAILER_DSN`.

aschempp: ah, got you! But this PR is for 4.9, so we need to update this for 4.12 as soon as this is merged 🙃

PR #3628, opened 2021-10-28
Make image width and height overwritable in the upload widget

doishub:
In addition to the upload size and file extensions, image widths and heights can also be overwritten within the upload field.


fritzmg:
The _Docs PR or issue_ is meant to be a link to a github.com/contao/docs Pull Request explaining the new features - or at least an issue that new features need to be explained :)

core-bundle/src/Resources/contao/forms/FormFileUpload.php
```diff
@@ -179,23 +181,23 @@ public function validate()
 
 		if ($arrImageSize = @getimagesize($file['tmp_name']))
 		{
-			$intImageWidth = Config::get('imageWidth');
+			$intImageWidth = $this->getMaximumImageWidth();
```
fritzmg:
```suggestion
			$intImageWidth = $this->maxImageWidth ?: Config::get('imageWidth');
```

imho we do not need a separate method for that

fritzmg (changes requested):

fritzmg (dismissed):

core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
```diff
@@ -152,6 +152,18 @@
       <trans-unit id="tl_form_field.maxlength.1">
         <source>Limit the field length to a certain number of characters (text) or bytes (file uploads).</source>
       </trans-unit>
+      <trans-unit id="tl_form_field.maxImageWidth.0">
+        <source>Maximum image width</source>
+      </trans-unit>
+      <trans-unit id="tl_form_field.maxImageWidth.1">
+        <source>Here you can enter the maximum width for image uploads in pixels.</source>
```
m-vo: Should there be an explanation that `0` means "use the value from the config"?

fritzmg: The translation is exactly the same as in `tl_settings`: https://github.com/contao/contao/blob/f562af12895bbfaa9adb51107af6d4f18ba6af86/core-bundle/src/Resources/contao/languages/en/tl_settings.xlf#L110-L121

If we want to change this, it should be changed in both places.

doishub: Yes makes sense, like so "Here you can enter the maximum width for image uploads in pixels. Set to 0 to use the default settings."?

doishub: > If we want to change this, it should be changed in both places.

Only within the upload field the default settings are used as fallback. I think it would have to be changed only within the upload field.

m-vo: Setting `0` in the config also has a special meaning, i.e.: no limit at all.

doishub: Ah now I see it. We have three levels to make these settings:
1. config.yml
2. settings
3. upload field

When the values are set in config, the fields inside settings are disabled. Therefore I would almost say, no further explanation would have to follow here. However, in the field we should include this.

aschempp: `config.yml` and settings are the same. If a config is set in the `config.yml`, you cannot override it in the back end, that's why the field is disabled. I don't think the form field should be disabled in that case though, and the global config still applies.

Regarding the labels, I would suggest to keep them the same in both places (= copy to the form field language file). This way [Transifex](https://www.transifex.com) should re-use all translations automatically for the form fields.

fritzmg: Unless we want to be able to disable the max image width and height completely for a form field, while still having a max image width and height in the system setting. This is currently not possible, as the field will always be `0` and in that case it will fall back to the system setting.

core-bundle/src/Resources/contao/forms/FormFileUpload.php
```diff
@@ -179,23 +181,23 @@ public function validate()
 
 		if ($arrImageSize = @getimagesize($file['tmp_name']))
 		{
-			$intImageWidth = Config::get('imageWidth');
+			$intImageWidth = $this->maxImageWidth ?: Config::get('imageWidth');
```
Toflar, line 186:
```suggestion
			$intImageWidth = $this->maxImageWidth ? min($this->maxImageWidth, Config::get('imageWidth')) : Config::get('imageWidth');
```

I don't think as a user you should be able to increase the maximum of the system settings, only decrease. As a system admin, I want to be able to control maximum values.

doishub: Basically I agree with you, but this is exactly the case I want to achieve with this PR. My customer wants to keep the limit (configuration settings) in all forms, only in this form it should be possible to upload bigger files (images). 🤔

fritzmg: ~~In that case you can remove/not add that field in the user rights.~~

fritzmg: May be add a save callback, so that only admins can set it higher than the system limit? 🤔

core-bundle/src/Resources/contao/languages/en/tl_form_field.xlf
aschempp: good question. I don't think that would be sensible though. If I configure a system wide image limitation, shouldn't that apply to forms as well? Alternatively, if the system configuration is (mostly) meant for the back end uploads, we might consider migrating the configuration to all form widgets and not inherit system settings at all. Just maybe use the system setting as default for new form fields.

core-bundle/src/Resources/contao/forms/FormFileUpload.php
aschempp: if that were useful the admin could simply hide the field for the back end user.

eki89: > I don't think as a user you should be able to increase the maximum of the system settings, only decrease. As a system admin, I want to be able to control maximum values.

I agree with you. However, it is a special case when the file size is set higher for an upload field. If you need to set the system-wide maximum value higher for this special case, it also follows that all default upload fields would have to be limited manually.

Splitting imageWidth into a default value and an absolute maximum value could help. In this case a checkbox to activate the limitation in upload fields could be a solution to handle all cases.

Using a special value for the backend could additionally be more flexible.

ausi (dismissed):

core-bundle/src/Resources/contao/forms/FormFileUpload.php
leofeyer: As discussed in the Contao call, we explicitly want to allow exceeding the maximum dimensions.

eki89:
As discussed in todays Contao Call, we should add the information that 0 means "use the value from the config".

doishub:
Unfortunately, I could not join the conversation this time. Thank you for discussing the issue. 

I have adjusted the description once again. 😉

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @doishub.

Issue #1842, opened 2020-06-17
Add image size to news and calendar feed for teaser images


bezin:
Hello there,

while working with the Contao core news feed I stumbled upon the issue, that the link to the teaser images will point to the source image in the files folder. However, these images might be a couple of MB in size. I would prefer to define a specific image size for this, so that the link points to an optimized thumbnail in the feed and even takes the important area into account.

This should be quite easily doable and I would be happy to provide a PR.

Another issue on the other hand is, that images from content elements would need to be resized to. 

I ran into this issue when we worked  on a newsletter template for CleverReach; they provide the feature, that one can dynamically include content in the newsletter via an RSS feed. However, the current implementation would include the 4000px×2800px source file. We implemented our own RSS feed to work around this issue in the application.

Any thoughts welcome :-)

asaage:
> the link to the teaser images will point to the source image in the files folder

Shouldn't the teaser-images be linked to the news-details-view?
At least for news-lists that is the default case - and if you specify the full-size/lightbox-view
tl_layout->lightboxSize would be applied.
I would prefer the feed to behave likewise.

Thoughts on the other issue:
I use the core-newsletter solution and added templates to the be_tinyNews.html5.
https://www.tiny.cloud/docs/plugins/template/
The url of those templates point to a stripped-down hidden page on my site wich just displays a newslist wich can be configured independently.

bezin:
> Shouldn't the teaser-images be linked to the news-details-view?

What do you mean? The URL of the feed item links to the detail page, yes. But I am talking about the image file itself. This will be the source file from /files (i.e. the path stored in tl_files). 

Your solution for the newsletter sounds interesting however we need a more powerful solution :-)

leofeyer:
@ausi /cc

ausi:
The feature would be an image size field in the `tl_news_feed` DCA?
But the image sizes of the content elements in the news article would stay as configured in the content elements?

leofeyer:
Since the only purpose of this is to prevent the large source image to be used in the RSS feed, we might as well either

a) use the fallback image size from the page layout or
b) hardcode a reasonable default format for images in RSS feeds.

I doubt that the image size will be different per feed.

bezin:
Hello, I totally missed out on this discussion due to my holidays 🙈 

> The feature would be an image size field in the tl_news_feed DCA?
But the image sizes of the content elements in the news article would stay as configured in the content elements?

[...]

PR #3630, opened 2021-10-29
Support image sizes in news and calendar feeds

bezin:
Implements #1842


calendar-bundle/src/Resources/contao/dca/tl_calendar_feed.php
```diff
@@ -211,7 +211,19 @@
 			'inputType'               => 'textarea',
 			'eval'                    => array('style'=>'height:60px', 'tl_class'=>'clr'),
 			'sql'                     => "text NULL"
-		)
+		),
+		'imageSize' => array
```
fritzmg: In `tl_module`, the field is called `imgSize` and everywhere else just `size`. I think we should use either of these (may be `imgSize`) in order to have more consistency.

fritzmg (changes requested):

fritzmg (dismissed):
LGTM 👍

calendar-bundle/src/Resources/contao/languages/en/tl_calendar_feed.xlf
```diff
@@ -56,6 +56,12 @@
       <trans-unit id="tl_calendar_feed.description.1">
         <source>Here you can enter a short description of the calendar feed.</source>
       </trans-unit>
+      <trans-unit id="tl_calendar_feed.imageSize.0">
+        <source>Image size</source>
+      </trans-unit>
+      <trans-unit id="tl_calendar_feed.imageSize.1">
```
aschempp:
```suggestion
      <trans-unit id="tl_calendar_feed.imgSize.0">
        <source>Image size</source>
      </trans-unit>
      <trans-unit id="tl_calendar_feed.imgSize.1">
```

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -113,12 +117,29 @@ public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
...
+		{
+			$image = System::getContainer()->get('contao.image.image_factory')->create($rootDir . '/' . $strFile, $imageSize);
+
+			if ($image instanceof DeferredImageInterface)
+			{
+				System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($image);
+			}
+
+			$fileUrl = $strUrl . System::urlEncode($image->getUrl($rootDir));
+			$relativeFilePath = str_replace($rootDir . '/', '', $image->getPath());
```
aschempp: this should probably use `PathUtils::makeRelative()` (not sure if that's the exact name 🙈 )

news-bundle/src/Resources/contao/languages/en/tl_news_feed.xlf
```diff
@@ -56,6 +56,12 @@
       <trans-unit id="tl_news_feed.description.1">
         <source>Here you can enter a short description of the news feed.</source>
       </trans-unit>
+      <trans-unit id="tl_news_feed.imageSize.0">
+        <source>Image size</source>
+      </trans-unit>
+      <trans-unit id="tl_news_feed.imageSize.1">
```
aschempp:
```suggestion
      <trans-unit id="tl_news_feed.imgSize.0">
        <source>Image size</source>
      </trans-unit>
      <trans-unit id="tl_news_feed.imgSize.1">
```

aschempp (changes requested):

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
m-vo: Yup, it's called `Path`: :slightly_smiling_face: 

```php
$relativeFilePath = Path::makeRelative($image->getPath(), $rootDir);
```

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -113,12 +118,29 @@ public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
 			$strUrl = Environment::get('base');
 		}
 
+		$fileUrl = $strUrl . System::urlEncode($strFile);
 		$objFile = new File($strFile);
 
+		$imageSize = StringUtil::deserialize($imageSize, true);
```
leofeyer: This line is overwriting the method argument, which is not a good thing to do. Maybe you can use a different name for the unserialized content?

calendar-bundle/src/Resources/contao/dca/tl_calendar_feed.php
```diff
@@ -211,7 +211,19 @@
...
+		'imgSize' => array
+		(
+			'exclude'                 => true,
+			'inputType'               => 'imageSize',
+			'reference'               => &$GLOBALS['TL_LANG']['MSC'],
+			'eval'                    => array('rgxp'=>'natural', 'includeBlankOption'=>true, 'nospace'=>true, 'helpwizard'=>true, 'tl_class'=>'w50'),
+			'options_callback' => static function ()
+			{
+				return System::getContainer()->get('contao.image.image_sizes')->getOptionsForUser(BackendUser::getInstance());
+			},
```
m-vo: What about putting this in a class, tagging it with…

```
@Callback(table="tl_calendar_feed ", target="fields.imgSize.options")
@Callback(table="tl_news_feed ", target="fields.imgSize.options")
```
… and using proper DI for both instances?

bezin: I really dislike the tagging pattern for select fields: It completely removes any reference to the options from the field definition: You look at the configuration and have no clue a) which options it has or b) even where these are compiled. You have to find the implementation manually somewhere. I would prefer defining a callabe over tagging if we want to remove the anonymous function. 

However, if this it where Contao wants to go I will bow to that.

SeverinGloeckle: You are definitely right that the loss of reference between an option field and its options is inconvenient.

The notation of a callable / service call can at least provide a hint on where to look for the options source (e.g. `contao.image.image_sizes`), but that still requires a manual lookup of the service ID and a public service - so it's not perfect either.

For the specific use-case of image sizes, there probably should be an easier way to get the options for the current backend user. Either the `ImageSizes` service should know about the current request and optionally infer the current backend user from it or a more general `BackendUserSettings` service should be added which can be queried for options and settings of the current backend user.

Alternatively, since the `ImageSize` widget is already querying the available options from the image sizes service in its validator: Maybe this logic can also be used as a default so you don't have to define the options for "regular" image size options at all?

However, as the current one-line callback to get the image sizes is used in multiple places already, I think it should be used here, too (for now, and maybe get refactored later together with similar callbacks).

SeverinGloeckle: As a side note: In my opinion, the specification of a closure as a callback should be completely discouraged in the long term, as these can only be defined within PHP code. A conversion / dumping of a DCA into another format - for example YAML - as a result would be impossible.

So that's definitely something to think about in the future.

bezin: >  but that still requires a manual lookup of the service ID and a public service - so it's not perfect either.

Yes, true – not perfect, but at least I have a service ID to search for in my IDE.

> However, as the current one-line callback to get the image sizes is used in multiple places already, I think it should be used here, too (for now, and maybe get refactored later together with similar callbacks).

That is what I think, too.

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
leofeyer: aa374353b51684cecff5afe6fff22836b5e07b5f did not fix this issue, I'm afraid, because now both the argument and the variable are called `$size`. One should be called `$imageSize` and the other one `$size`.

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -101,9 +104,11 @@ public function __isset($strKey)
 	 * @param string $strUrl   The base URL
 	 * @param string $strMedia The media type
 	 */
-	public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
+	public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure', $size = null)
```
aschempp: `$size` is missing in the docbloc

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -113,12 +118,29 @@ public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
...
 
+		$size = StringUtil::deserialize($size, true);
+
+		if ($size && $objFile->isImage)
+		{
+			$image = System::getContainer()->get('contao.image.image_factory')->create($rootDir . '/' . $strFile, $size);
+
+			if ($image instanceof DeferredImageInterface)
+			{
+				System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($image);
```
aschempp: why do we need to force-resize deferred images? They are not embedded in the feed so they would just be created on request?

aschempp (changes requested):

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
bezin: True, this is probably specific to my original use case. We used the RSS feed to import news articles into and send them with CleverReach. Their client did not wait long enough for the images to be created and just did not load them at all. 

We could make DeferredImages the default, but I would like to add a checkbox then, that allows me to create images on feed creation. WDYT?

leofeyer: @ausi /cc

aschempp: if you have that specific problem, you should configure that in your application – disable deferred images or run the image-generation command in a cron job.

ausi: > but I would like to add a checkbox then, that allows me to create images on feed creation. WDYT?

I would not add this checkbox as this requirement is a quite specific use-case IMO.

But we should make sure that your use-case is possible at least. Is there an event or hook that you can use to intercept the feed contents? If so, it might be possible to resize the images like the newsletter module does with `$objFile->exists() || $objFile->createIfDeferred()`: https://github.com/contao/contao/blob/4299a5116b73cdbcd024c2d86b0533ae15ec3137/core-bundle/src/Resources/contao/library/Contao/Email.php#L476

bezin: Agreed, I will look into that. As far as I see, there is no Hook/Event yet, so we would have to add one. I will figure out, where would be the best place.

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
bezin: Fixed

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -113,12 +118,23 @@ public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
...
 		$objFile = new File($strFile);
 
+		$size = StringUtil::deserialize($imageSize, true);
+
+		if ($size && $objFile->isImage)
+		{
+			$image = System::getContainer()->get('contao.image.image_factory')->create($rootDir . '/' . $strFile, $size);
+			$fileUrl = $strUrl . System::urlEncode($image->getUrl($rootDir));
+			$relativeFilePath = Path::makeRelative($image->getPath(), $rootDir);
+			$objFile = new File($relativeFilePath);
```
aschempp: Are you sure to use `File` class for a non-resized image?

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
m-vo: Alternatively: If we migrate the feed composition to templates, this would be easily adjustable for users as well.

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
ausi: I don’t understand the question. It overwrites the `$objFile` variable from above which seems to be correct.

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
aschempp: > Alternatively: If we migrate the feed composition to templates, this would be easily adjustable for users as well.

not related to this PR, and something we might look at when rewriting to a controller 😉

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
bezin: As far as I see the file is not required to be present, when instantiating `File`? However, I added a check if the file exists, before I get `$objFile->size` some lines below to avoid a warning

calendar-bundle/src/Resources/contao/dca/tl_calendar_feed.php
leofeyer: I have created a separate issue for discussing this: #3907

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
leofeyer: Resolved through 1883a64eb8fcdaa29eef11f9c6eb295a6c096fd8

calendar-bundle/src/Resources/contao/dca/tl_calendar_feed.php
bezin: I am currently refactoring this: I implement a general purpose listener that is used in all other places in the core, too (`tl_layout`, `tl_module`, `tl_content`). I then use this in `tl_news` and `tl_news_feed` as well as `tl_calendar_event`  and `tl_calendar_feed`.

leofeyer (dismissed):

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -113,15 +118,29 @@ public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
 			$strUrl = Environment::get('base');
 		}
 
+		$fileUrl = $strUrl . System::urlEncode($strFile);
 		$objFile = new File($strFile);
+		$size = StringUtil::deserialize($imageSize, true);
+
+		if ($size && $objFile->isImage)
+		{
+			$image = System::getContainer()->get('contao.image.image_factory')->create($rootDir . '/' . $strFile, $size);
```
m-vo:
```suggestion
			$image = System::getContainer()->get('contao.image.image_factory')->create(Path::join($rootDir, $strFile), $size);
```

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
```diff
@@ -97,13 +99,16 @@ public function __isset($strKey)
...
+	 * @param string $strMedia  The media type
+	 * @param mixed  $imageSize The image size
 	 */
-	public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure')
+	public function addEnclosure($strFile, $strUrl=null, $strMedia='enclosure', $imageSize = null)
 	{
-		if (!$strFile || !file_exists(System::getContainer()->getParameter('kernel.project_dir') . '/' . $strFile))
+		$rootDir = System::getContainer()->getParameter('kernel.project_dir');
+
+		if (!$strFile || !file_exists($rootDir . '/' . $strFile))
```
m-vo:
```suggestion
		if (!$strFile || !file_exists(Path::join($rootDir, $strFile)))
```

core-bundle/src/Resources/contao/library/Contao/FeedItem.php
bezin: https://github.com/contao/contao/pull/3630/commits/18bac58c6522ab73dcea26867d8e851f1cdb1bda

m-vo (dismissed):

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -265,10 +265,6 @@
 			'inputType'               => 'imageSize',
 			'reference'               => &$GLOBALS['TL_LANG']['MSC'],
 			'eval'                    => array('rgxp'=>'natural', 'includeBlankOption'=>true, 'nospace'=>true, 'helpwizard'=>true, 'tl_class'=>'w50'),
-			'options_callback' => static function ()
```
leofeyer, line 268: Why did you remove the options callback here (and in tl_layout and tl_module)?

bezin: Because I added a [general purpose listener for this and use it via annotations](https://github.com/contao/contao/pull/3630/files#diff-d71a0a24c80180705e7fd49548488f67bc086cb5cc0e547dea38794fb68428f5R21-R23)

aschempp (dismissed):

leofeyer (approved):

leofeyer:
Thank you @bezin.

PR #3631, opened 2021-11-01
Backup management on CLI

Toflar:
Follow up on https://github.com/contao/contao/pull/3594.

This PR provides 3 new commands:

- `contao:backup:create` - to create a new backup
- `contao:backup:restore` - to restore a given backup
- `contao:backup:list` - to list all available backups

Configuration options:

- The maximum number of backups Contao should keep (defaults to `5`).
- The target directory where it should manage the backups (defaults to `var/backups`)
- Tables to ignore for backups (defaults to things like the search index or our crawler queue)

Additional features:

- Our `contao:migrate` command now automatically creates a backup before migrating which makes things a lot safer :-) You can optionally disable it using `--no-backup`.
- `--format=json` support for the `create` and `list` commands for a possible future integration in e.g. the Contao Manager.

Impressions:

Bildschirmfoto 2021-11-01 um 14 04 31
Bildschirmfoto 2021-11-01 um 14 04 50
Bildschirmfoto 2021-11-01 um 14 05 03

ToDo:

- [x] Tests


core-bundle/src/Command/Backup/AbstractBackupCommand.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->backupManager = $backupManager;
+
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('file', InputArgument::OPTIONAL, 'The path to the SQL dump to process.')
+            ->addOption('ignore-tables', 'i', InputOption::VALUE_OPTIONAL, 'A comma-separated list of database tables to ignore. Defaults to the backup configuration (contao.backup.ignore_tables).')
```
aschempp: are you sure the option value is optional? Also, we might want to use the array format, such command usually work like `--ignore-table=tl_foo --ignore-table=tl_bar`

core-bundle/src/Command/Backup/BackupRestoreCommand.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    protected static $defaultName = 'contao:backup:restore';
+
+    protected function configure(): void
+    {
+        parent::configure();
+
+        $this
+            ->addOption('force', null, InputOption::VALUE_NONE, 'By default, this command checks whether the backup has been generated with Contao too. Use --force to disable this check.')
+            ->setDescription('Restores a backup.')
```
aschempp, line 34: any reason this does not support JSON as well?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,311 @@
...
+
+    public function __construct(Connection $connection, string $backupDir, array $tablesToIgnore, int $keepMax)
+    {
+        $this->connection = $connection;
+        $this->backupDir = $backupDir;
+        $this->tablesToIgnore = $tablesToIgnore;
+        $this->keepMax = $keepMax;
+    }
+
+    public function createCreateConfig(): CreateConfig
```
aschempp, line 43:
```suggestion
    public function createBackupConfig(): CreateConfig
```

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -84,6 +92,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
     }
 
+    private function backup(): void
+    {
+        $config = $this->backupManager->createCreateConfig();
+        $this->io->info(sprintf(
+            'Creating a database dump to "%s" with the default options. Use --no-backup to disable this feature.',
+            $config->getBackup()->getFilepath()
+        ));
+        $this->backupManager->create($config);
```
aschempp: what if the backup fails here?

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -565,6 +566,29 @@ static function (array $attributes): array {
         ;
     }
 
+    private function addBackupNode(): NodeDefinition
+    {
+        return (new TreeBuilder('backup'))
+            ->getRootNode()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->scalarNode('backup_dir')
```
aschempp:
```suggestion
                ->scalarNode('target_dir')
```

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,311 @@
...
+     *
+     * @return array<Backup>
+     */
+    public function listBackups(): array
+    {
+        (new Filesystem())->mkdir($this->backupDir); // Ensures, the directory exists
+
+        $backups = [];
+        $files = Finder::create()
+            ->files() // TODO: depth == 0?
```
aschempp: since this is supposed to work with our backups only, I would say `yes`. We might even hardcode the (allowe(d)) file extensions?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,311 @@
...
+            ->in($this->backupDir)
+            ->ignoreDotFiles(true)
+            ->ignoreVCS(true)
+        ;
+
+        foreach ($files as $file) {
+            $backups[] = new Backup($file->getPathname());
+        }
+
+        uasort($backups, static fn (Backup $a, Backup $b) => $b->getCreatedAt() <=> $a->getCreatedAt());
```
aschempp: `Finder` allows for a sorting callback, maybe we should use that?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * @throws BackupManagerException
+     */
+    private function doCreate(CreateConfig $config): void
+    {
+        $backup = $config->getBackup();
+
+        // Ensure the target file exists and is empty
+        (new Filesystem())->dumpFile($backup->getFilepath(), '');
```
aschempp, line 144: I think we shouldn't overwrite if the file already exists 🤔
also, `fopen` already does that and truncates the file?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * @throws BackupManagerException
+     */
+    private function executeWrappedQuery(string $query): void
+    {
+        try {
+            $this->connection->executeQuery($query);
+        } catch (Exception $e) {
+            throw new BackupManagerException($e->getMessage());
```
aschempp: should probably pass `$previous` (applies to other places as well)

core-bundle/src/Doctrine/Backup/Config/CreateConfig.php
```diff
@@ -0,0 +1,31 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Doctrine\Backup\Config;
+
+class CreateConfig extends AbstractConfig
```
aschempp, line 15:
```suggestion
class BackupConfig extends AbstractConfig
```
?

core-bundle/src/Resources/config/services.yml
```diff
@@ -268,6 +268,13 @@ services:
         tags:
             - { name: data_collector, template: '@ContaoCore/Collector/contao.html.twig', id: contao }
 
+    Contao\CoreBundle\Doctrine\Backup\BackupManager:
+        arguments:
+            - '@database_connection'
+            - '' # Set by extension
+            - [] # Set by extension
+            - 0 # Set by extension
```
aschempp: I would apply our sane default config here as well just to be explicit

aschempp (commented):
Looks very simple and promising! 🎉 

Three side questions coming up with this PR
 1. should we at some point stop grouping _every_ command from Contao into the `contao:` prefix? Also because this would fit into the `database` namespace, because it does not do Contao backups but database backups 🤔 
 2. I wonder if this belongs to `core-bundle` or rather should be in the `manager-bundle`. For 99% of our users it wouldn't matter, but if you run a Symfony app with Contao you might now expect to have Contao-specific DB backup commands and probably use other tools like `slimdump` directly? Maybe @bytehead @sheeep have some input on this.

bytehead:
> 2. I wonder if this belongs to core-bundle or rather should be in the manager-bundle. For 99% of our users it wouldn't matter, but if you run a Symfony app with Contao you might now expect to have Contao-specific DB backup commands and probably use other tools like slimdump directly? Maybe @bytehead @sheeep have some input on this.

I would like to see it in the core as it is a CLI command. I don't see the point here, where it should belong to the manager (bundle). In some applications we even separate the DB backups between Contao and the Symfony application. IMHO this would fit to the core bundle.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Yes we should, that's exactly why I'm doing this here. And I'm using `Filesystem` because it's safer :)

core-bundle/src/Doctrine/Backup/Config/CreateConfig.php
Toflar: Everything is Backup config but this one is the create config. So that's correct.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: No we cannot because we don't sort in the finder.

core-bundle/src/Command/MigrateCommand.php
Toflar: Then it will throw an exception which the command shows accordingly. This is on purpose.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Same here, that's correct.

core-bundle/src/Command/Backup/BackupRestoreCommand.php
Toflar: Because at the moment it does not provide any more info other than "worked" or "failed" which the status code already provides.

core-bundle/src/Command/Backup/AbstractBackupCommand.php
Toflar: Yes it is optional. If you don't provide anything, the config is taken into account. And I prefer the comma-separated notation because I don't want to get sore fingers from typing.

Toflar:
>     2\. I wonder if this belongs to `core-bundle` or rather should be in the `manager-bundle`. For 99% of our users it wouldn't matter, but if you run a Symfony app with Contao you might now expect to have Contao-specific DB backup commands and probably use other tools like `slimdump` directly? Maybe @bytehead @sheeep have some input on this.

This was already decided at the core devs meeting. That's why it's in core-bundle.

core-bundle/src/DependencyInjection/Configuration.php
Toflar: I don't like that so I won't use that suggestion.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Done in f2a2fcbaca9ddc1b1ccda92e57319e728667927f.

core-bundle/src/Resources/config/services.yml
Toflar: Done in 72abbf77b357a98f9b483f9b1edda4113ab9bf0f.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Of course yes, this is just a todo for me for the tests, so I'm going to resolve this discussion.
I don't mind about the used extensions but maybe I'll just limit them to `sql` and `sql.gz`.

Toflar:
@contao/developers this is now completely unit tested and ready for review. @fritzmg would you mind checking the failing tests on windows? They're all related to new lines and directory separators but I'm going to go nuts if I cannot fix those locally 😟

fritzmg:
> @fritzmg would you mind checking the failing tests on windows? They're all related to new lines and directory separators but I'm going to go nuts if I cannot fix those locally 😟

I'll check them, yes :)

fritzmg:
Done: https://github.com/Toflar/contao/pull/8

Toflar:
I want to ease replacing the core dumper a little bit by moving the actual write/stream responsibility to the backup manager.

Toflar:
Alright, this is ready for review now :) The `DumperInterface` can now easily stream single lines and the `BackupManager` will write and transparently compress the file. It's super easy to replace the core `Dumper` now. Generators FTW :) 
Ah, and while I was on it, I added native support for views 😎

core-bundle/src/Command/Backup/AbstractBackupCommand.php
m-vo: What about adding a `getIgnoredTables(): array` method to the `BackupManager` and then listing the effectively ignored ones here?

core-bundle/src/Command/Backup/AbstractBackupCommand.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Contao\CoreBundle\Command\Backup;
+
+use Contao\CoreBundle\Doctrine\Backup\BackupManager;
+use Contao\CoreBundle\Doctrine\Backup\Config\AbstractConfig;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+
```
m-vo, line 21:
```suggestion

/**
 * @internal
 */
```

core-bundle/src/Command/Backup/AbstractBackupCommand.php
```diff
@@ -0,0 +1,65 @@
...
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('file', InputArgument::OPTIONAL, 'The path to the SQL dump to process.')
+            ->addOption('ignore-tables', 'i', InputOption::VALUE_OPTIONAL, 'A comma-separated list of database tables to ignore. Defaults to the backup configuration (contao.backup.ignore_tables).')
+            ->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format (txt, json)', 'txt')
+        ;
+    }
+
```
m-vo, line 44: I think with this we might remove the `@var` annotations in the child classes.
```suggestion

/**
 * @template T
 * @param T $config
 * @psalm-return T
 */
```

core-bundle/src/Command/Backup/BackupCreateCommand.php
```diff
@@ -0,0 +1,70 @@
...
+            ->setDescription('Creates a new backup.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $config = $this->backupManager->createCreateConfig();
+        /** @var CreateConfig $config */
```
m-vo:
```suggestion
```

core-bundle/src/Command/Backup/BackupRestoreCommand.php
```diff
@@ -0,0 +1,76 @@
...
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $config = $this->backupManager->createRestoreConfig();
+        /** @var RestoreConfig $config */
```
m-vo:
```suggestion
```

core-bundle/src/Command/Backup/BackupRestoreCommand.php
```diff
@@ -0,0 +1,76 @@
...
+class BackupRestoreCommand extends AbstractBackupCommand
+{
+    protected static $defaultName = 'contao:backup:restore';
+
+    protected function configure(): void
+    {
+        parent::configure();
+
+        $this
+            ->addOption('force', null, InputOption::VALUE_NONE, 'By default, this command checks whether the backup has been generated with Contao too. Use --force to disable this check.')
```
m-vo: Maybe we should explicitly name the command, so that people do not confuse it with extensions?
```suggestion
            ->addOption('force', null, InputOption::VALUE_NONE, 'By default, this command makes sure the backup has been generated with "contao:backup:create". Use --force to disable this check.')
```

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @throws BackupManagerException
+     */
+    private function executeTransactional(\Closure $func): void
+    {
+        $isAutoCommit = $this->connection->isAutoCommit();
+
+        if ($isAutoCommit) {
+            $this->connection->setAutoCommit(false);
+        }
```
m-vo, line 121: Why is this needed? Isn't it enough to start the transaction beforehand and commit it afterwards / use `Connection#transactional()` + closure?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @throws BackupManagerException
+     */
+    public function create(CreateConfig $config): void
+    {
+        $this->executeTransactional(
+            function () use ($config): void {
+                $this->doCreate($config);
+            }
+        );
```
m-vo:
```suggestion
        $this->executeTransactional(fn() => doCreate($config));
```

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @throws BackupManagerException
+     */
+    public function restore(RestoreConfig $config): void
+    {
+        $this->executeTransactional(
+            function () use ($config): void {
+                $this->doRestore($config);
+            }
+        );
```
m-vo:
```suggestion
        $this->executeTransactional(fn() => doRestore($config));
```

core-bundle/src/Doctrine/Backup/DumperInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+interface DumperInterface
+{
+    /**
+     * The data yielded is streamed into the backup file. Every yielded string
+     * will be a new line.
+     *
+     * @throws BackupManagerException in case anything went wrong
+     */
+    public function dump(Connection $connection, CreateConfig $config): \Generator;
```
m-vo, line 24: Maybe worth annotating that the strings must be returned `@return \Generator<string>`? Is a generator mandatory or do we just need something that is iterable?

core-bundle/src/Resources/config/services.yml
```diff
@@ -268,6 +268,17 @@ services:
...
+    Contao\CoreBundle\Doctrine\Backup\DumperInterface:
+        class: Contao\CoreBundle\Doctrine\Backup\Dumper
+
+    Contao\CoreBundle\Doctrine\Backup\BackupManager:
+        arguments:
+            - '@database_connection'
+            - '@Contao\CoreBundle\Doctrine\Backup\DumperInterface'
+            - '%kernel.project_dir%/var/backups' # Set by extension
+            - [tl_crawl_queue, tl_log, tl_search, tl_search_index, tl_search_term] # Set by extension
+            - 5 # Set by extension
```
m-vo: When are these values used if the extension sets them anyway?

m-vo (changes requested):
Awesome feature! :heart:  I'm looking forward to this!

One or two general things/questions:

* Is `Doctrine` a good namespace for this? To me this is somewhat orthogonal to the things from this vendor.
* A typical use case for ignored tables might be to skip additional ones when importing. Like for instance `tl_user` to still be able to log in :smile: . I thought we could maybe allow an `&` in the list of ignored tables that would stand for the configured ones. So like `--ignore-tables=&,tl_user` or is this overkill?

core-bundle/src/Command/Backup/AbstractBackupCommand.php
Toflar: Everything in the `Command` namespace is not covered anyway: https://docs.contao.org/dev/internals/bc-promise/

m-vo: But you added it to the other commands :slightly_smiling_face:.

core-bundle/src/Command/Backup/AbstractBackupCommand.php
Toflar: Done in 0d669948283e7401fa7e317d9e8e415332cb8707.

core-bundle/src/Command/Backup/BackupRestoreCommand.php
Toflar: Done in 0d669948283e7401fa7e317d9e8e415332cb8707.

core-bundle/src/Command/Backup/BackupCreateCommand.php
Toflar: Done in 0d669948283e7401fa7e317d9e8e415332cb8707.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: This was the only way I could get a proper rollback to work while trying to actually backup a DB.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: That's the CS fixer here :)

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Same here, CS fixer.

core-bundle/src/Doctrine/Backup/DumperInterface.php
Toflar: I wanted to make it mandatory so people make sure, they actually stream stuff. Otherwise one builds a huge array in memory again ;) But I agree with the docs which I have improved in 08d98c8bc21c55a75dc891e465bc77f9be37d37b.

core-bundle/src/Resources/config/services.yml
Toflar: Because @aschempp wanted them to be here :)

Toflar:
>     * A typical use case for ignored tables might be to skip additional ones when importing. Like for instance `tl_user` to still be able to log in 😄 . I thought we could maybe allow an `&` in the list of ignored tables that would stand for the configured ones. So like `--ignore-tables=&,tl_user` or is this overkill?

Hmm...didn't we have some kind of similar use case lately where @ausi implemented adding/removing? Where was that? Did we use +/-?

core-bundle/src/Doctrine/Backup/BackupManager.php
m-vo: The backup process is read only, so commiting a transaction (either explicitly or by setting auto commit) would just release the locks, no? You cannot rollback imports anyways because `MySQL` does not support transactions for DDL queries.

core-bundle/src/Doctrine/Backup/BackupManager.php
m-vo: Ah, I see :see_no_evil:. 

/cc @leofeyer Can we adjust this? I think this issue will occur more often in the future.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: I lack the knowledge here so I'd love to have you checkout the branch and do some tests yourself. It worked for me 🙈

core-bundle/src/Doctrine/Backup/DumperInterface.php
m-vo: Good point with streaming. (Now people are going to write `yield from ArrayIterator($array);` :smile: )

ausi:
> Where was that? Did we use +/-?

https://github.com/contao/contao/blob/f2e179360dd7e1c5a10d657ca92f0746e18a0508/core-bundle/src/DependencyInjection/Configuration.php#L366

core-bundle/src/Doctrine/Backup/DumperInterface.php
Toflar: Haha, yeah well...it requires enough knowledge about generators to actually **not** do that ;)

Toflar:
@m-vo would you be okay if I implemented `+tl_user` and `-tl_user`?

m-vo:
> @m-vo would you be okay if I implemented +tl_user and -tl_user?

This would be even better than my proposal. :tada:

Toflar:
Added in 06083e8e7e9a950959e7c6bc33e83efa8f89df84.

core-bundle/src/Command/Backup/AbstractBackupCommand.php
Toflar: I see, fixed in 5d909578efc061bc74c3c1954068ef1698b0af14.

core-bundle/src/Command/Backup/AbstractBackupCommand.php
Toflar: In theory there might be different ones for creating and restoring so I don't know if we really want that.

frontendschlampe:
Just a question: What about exporting views? Can I also take them out of backup? Because many webhoster don't allow to import views - you have to recreate them after importing database without views.

m-vo:
> Because many webhoster don't allow to import views

You mean when importing with `mysql`? This command does not use the CLI but runs queries instead. So this should work out of the box if you *can* create views in general.

frontendschlampe:
Actually I export my databases without the views and after importing them, I recreate the views. So I mean if it's possible to ignore views if I make a backup with `contao:backup:create`

Toflar:
>  Because many webhoster don't allow to import views - you have to recreate them after importing database without views.

Either you are allowed to execute a `CREATE VIEW` query or not. There is no difference between the `backup:restore` command and you executing that query in a tool manually.

fkaminski:
I would just like to ask carefully whether an option would be possible that could provide the export for the install tool. 

Analogous to the BackupDB extension:
https://github.com/do-while/contao-BackupDB/wiki/BackupDB-for-Contao-4.4-(EN)

fkaminski:
@Toflar Too bad - Understand: Thanks for comment

fritzmg:
Unfortunately I am not able to import the created backups via another tool like SQLyog:

```
Query:
INSERT INTO `tl_news` (`id`, `pid`, `tstamp`, `headline`, `alias`, `author`, `date`, `time`, `subheadline`, `teaser`, `addImage`, `overwriteMeta`, `singleSRC`, `alt`, `imageTitle`, `size`, `imagemargin`, `imageUrl`, `fullsize`, `caption`, `floating`, `addEnclosure`, `enclosure`, `source`, `jumpTo`, `articleId`, `url`, `target`, `cssClass`, `noComments`, `featured`, `published`, `start`, `stop`, `description`, `pageTitle`, `robots`, `fb_schedule_time`, `fb_schedule_post`, `fromFb`, `fbPostId`, `fbData`, `fbMessage`, `fbPostAsPhoto`, `postToFb`, `twitter_schedule_time`, `twitter_schedule_post`, `tweetImage`, `fromTwitter`, `tweetId`, `twitterData`, `twitterMessage`, `postToTwitter`, `languageMain`) VALUES ('90', '5', '1619452842', 'Tweet from @fritz_mg', 'tweet-from-fritz_mg-6132697a1bf7e', '0', '1619452842', '1619452842', '', '…', '1', '', UNHEX('D82C13F00CE411EC9433D8BBC1145C99'), '', '', '', '', '', '1', '', 'above', '', NULL, 'external', '0', '0', 'https://twitter.com/fritz_mg/status/1386711846464405505/photo/1', '1', '', '', '', '', '', '', NULL, '', '', NULL, '', '', '', NULL, NULL, '', '', NULL, '', '', '1', '1386711846464405505', '…', '', '', '0')

Error occured at:2021-11-22 10:48:30
Line no.:440
Error Code: 1366 - Incorrect string value: '\xF0\x9F\x98\x81.<...' for column `c412dev`.`tl_news`.`teaser` at row 1
```

Though interestingly it still works with `mysql c412dev < backup__20211122104707.sql` - but the `Incorrect string value` still does occur, if I execute the generated `INSERT INTO` statement manually. Restoring via `contao:backup:restore` also works 🤔 

// edit: the issue is an Emoji contained in that news teaser. This is likely an issue with SQLyog itself, since everything else (charset and collation) is otherwise correct (`utf8mb4`).

core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private function formatValueForDump($value, Column $column, Connection $connection): string
+    {
+        if (null === $value) {
+            return 'NULL';
+        }
+
+        if ($column->getType() instanceof BinaryType) {
+            return sprintf('UNHEX(%s)', $connection->quote($value));
```
fritzmg, line 115: `mysqldump` converts binary fields to `0xd82c13f00ce411ec9433d8bbc1145c99` for example with the `--hex-blob` option instead of `UNHEX('D82C13F00CE411EC9433D8BBC1145C99')`. May be we should do the same here to, in order to avoid any issues with binary data (e.g. https://github.com/contao/contao/pull/3631#issuecomment-975400899).

// edit: nvm. the original issue was not with the `singleSRC` field anyway.

core-bundle/src/Doctrine/Backup/Backup.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public function getSize(): int
+    {
+        return (int) filesize($this->getFilepath());
+    }
+
+    public function getHumanReadableSize(): string
+    {
+        $base = log($this->getSize()) / log(1024);
+        $suffix = ['B', 'KB', 'MB', 'GB', 'TB'][floor($base)];
```
fritzmg:
```suggestion
        $suffix = ['B', 'KiB', 'MiB', 'GiB', 'TiB'][floor($base)];
```

In Contao we use kibibyte etc. 🙃

fritzmg (approved):
LGTM 👍 
My local tests showed no (obvious) issues. My other mentioned issues are solely related to the SQL tool I used.

core-bundle/src/Doctrine/Backup/Backup.php
Toflar: Yeah I actually wanted to reuse `System::getReadableSize()` here but that's still a static method, uses superglobals  etc. We might do that once that was migrated to a proper service as well (cannot be static due to the dep on the translator service).

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -0,0 +1,265 @@
...
+            ->name(Backup::VALID_BACKUP_NAME_REGEX)
+        ;
+
+        foreach ($files as $file) {
+            $backups[] = new Backup($file->getPathname());
+        }
+
+        uasort($backups, static fn (Backup $a, Backup $b) => $b->getCreatedAt() <=> $a->getCreatedAt());
+
+        return array_values($backups);
```
ausi:
```suggestion
        usort($backups, static fn (Backup $a, Backup $b) => $b->getCreatedAt() <=> $a->getCreatedAt());

        return $backups;
```
Nitpick ☺️

ausi (approved):
❤️

ausi (approved):

ausi (approved):

leofeyer:
CS fixes are here: https://github.com/Toflar/contao/pull/11

core-bundle/src/Doctrine/Backup/Backup.php
```diff
@@ -0,0 +1,95 @@
...
+
+        (new Filesystem())->dumpFile($filepath, '');
+
+        return new self($filepath);
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'createdAt' => $this->getCreatedAt()->format(\DateTimeInterface::ISO8601),
```
leofeyer: Can this be ignored or should we change it?

Toflar: We should probably use ATOM, yeah. You can change that, no problem :)

core-bundle/tests/Doctrine/Backup/DumperTest.php
```diff
@@ -0,0 +1,218 @@
...
+        $connection
+            ->expects($this->once())
+            ->method('createSchemaManager')
+            ->willThrowException(new \Exception('Error!'))
+        ;
+
+        $dumper = new Dumper();
+        $config = (new CreateConfig($backup))->withGzCompression(false);
+
+        iterator_to_array($dumper->dump($connection, $config), false);
```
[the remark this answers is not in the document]
Toflar: That's because of your CS changes. You changed `@return static` to `self`. `self` is the same class (= `AbstractConfig`). But `static` would be correct. Which is natively only available as of PHP 8, see https://wiki.php.net/rfc/static_return_type

core-bundle/src/Resources/config/services.yml
leofeyer: Does not make any sense to me. 🤷‍♂️

Toflar: Imho it should just be like in other places so if we don't do it elsewhere, we shouldn't do it here either. So that's a CS fix as well imho.

m-vo: I'd vote for removing. Writing `Set by extension` and - at the same line - still setting data is kind of incoherent communication.

core-bundle/src/Doctrine/Backup/BackupManager.php
m-vo: Let's leave this as is. I added it to my list + can still make another PR if we want to change something later on.

m-vo (approved):

core-bundle/src/Doctrine/Backup/Backup.php
leofeyer: I have added the changes to [Toflar#11](https://github.com/Toflar/contao/pull/11).

core-bundle/src/Resources/config/services.yml
leofeyer: Especially since PHPStorm already shows the argument type:

core-bundle/src/Doctrine/Backup/BackupManager.php
leofeyer: The CS fixer does support fn() functions (see a11da204ece9bc72a2a2a350366957c0f3c0b9d2).

leofeyer:
Thank you @Toflar.

PR #3633, opened 2021-11-02
Restore parent records first, if it does not exist

### LINKED ISSUE #3612: Avoid data-loss if parent and child were deleted independently
While working on #3498 I try to tackle more issues with the general undo procedure.

It is possible to lose data for good, when it comes to parent-child data.

**To reproduce:**

1. Delete a news item from the back-end
2. Delete the parent news archive of this item from the back-end

Go to the restore module: You now have two entries, one for the item and one for the archive

3. Restore the news item (The data will be written back into `tl_news`, inspect with any database GUI)
4. Go to the news module again

Now the controllers `reviseTable` kicks in and deletes the news item again, as it references a non-existing news archive.
The item data is now lost. 

**Ideas to solve this:**

* Disable the restore button, if the parent does not exist and show a hint, which parent has to be restored before or have a dialog box that asks the editor if you want to restore the related parent item first. This would require a new interface element (proper dialog box) or could be done with a confirm box like the delete action. 
* Restore the parent before the child automatically and show a global message afterwards

bezin:
Fixes #3612 

This PR adds a parent check before we restore a record: 
* We check if the parent and the parents parents exis (recursively). 
* If no, we restore them from the undo table as well, if possible.
* If the direct parent does not exist and cannot be restored from the undo table, we disable the restore button, so the data does not get lost


bezin:
This PR adds some checks before we restore a record: 
* We check if the parent and the parents parents exist. 
* If no, we restore them if we can restore them from the undo table as well

bezin:
Try to restore a record where the parent is missing and not restorable:


Restore a record where the parent is missing but can be restored prior:


I am really bad at error messages. Comments appreciated ;-)

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6238,6 +6215,100 @@ public function initPicker(PickerInterface $picker)
...
+
+				// Trigger the undo_callback
+				if (\is_array($GLOBALS['TL_DCA'][$table]['config']['onundo_callback']))
+				{
+					foreach ($GLOBALS['TL_DCA'][$table]['config']['onundo_callback'] as $callback)
+					{
+						if (\is_array($callback))
+						{
+							$this->import($callback[0]);
+							$this->{$callback[0]}->{$callback[1]}($table, $row, $this);
```
bezin, line 6301: This could be a problem: `$this` is always the DataContainer of the original undo record, that the user intended to restore. But `$table` or `$row` might be a parent record or a parents parent record. Not sure if I should create a new instance of `DataContainer` here?

core-bundle/src/EventListener/UndoOperationsListener.php
```diff
@@ -0,0 +1,149 @@
...
+
+        /** @var Image $image */
+        $image = $this->framework->getAdapter(Image::class);
+        $this->image = $image;
+    }
+
+    /**
+     * @Callback(table="tl_undo", target="list.operations.undo.button")
+     */
+    public function onRenderUndoButton(array $row, ?string $href, string $label, string $title, ?string $icon, string $attributes = ''): string
```
aschempp, line 78: since this listener has only one public method (as it should!), the method should be `__invoke` and you can place the `@Callback` annotation directly on the class. Also please mark the class as `@internal` please.

aschempp (commented):
Looks good! I wonder how the performance is if we need to do all these operations for every undo-row?

Also, can you please explain the `originalId`? The only place it seems used is where `data` is unserialised anyway, so the same data would be available?

core-bundle/src/EventListener/UndoOperationsListener.php
bezin: I agree! Since this PR is related to #3498 I would then also add a separate listener for the `jumpToParent` there and move everything to a `EventListener/Undo`  sub-namespace?

bezin: I implemented this in https://github.com/contao/contao/pull/3498/commits/2f8a4cdaba576237bbcd73fca5f9933fba282b93. If nobody objects to this I would update this PR here accordingly.

bezin:
> Looks good! I wonder how the performance is if we need to do all these operations for every undo-row?
> 
> Also, can you please explain the `originalId`? The only place it seems used is where `data` is unserialised anyway, so the same data would be available?

Hi @aschempp missed out on your comment: If you refer to https://github.com/contao/contao/pull/3633/files#diff-8b238ee3f300eb74ed8ebce5333c8d5514d0f3f9f199108e445ac85839cda8b4R72 you are right this variable is redundant.

leofeyer:
@bezin What about this pull request? It was closed automatically when the `4.9` milestone was removed, but it does not seem resolved.

bezin:
I have no idea what state the implementation is in. Will have to check.

PR #3623, opened 2021-10-27
Do not set the server_version

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1696

For the record as described in https://github.com/contao/contao/pull/3623#issuecomment-952708988 the `CommandSchedulerListener` must only load the `Cron` service when it needs it. Otherwise the install tool would fail when terminating because the `Cron` service needs ORM metadata which itself needs to know the `server_version`.

PR #3635, opened 2021-11-03
Use service container and add explanation

aschempp:
see #3623


ausi (changes requested):
I don’t unserstand this change, using a service locator for the framework and the connection has no benefits. Also the tests seem to be failing now.

PR #3637, opened 2021-11-03
Fix the decorator pattern in the FailTolerantProxyCacheWarmer

aschempp:
related to #3623


ausi (commented):
I don’t think extending the class of the inner service is bad practise.
But I’m OK either way, so if the majority wants to merge this I’m 👍

_(but tests and coding style needs to be fixed)_

aschempp:
coding style fixed. `FailTolerantProxyCacheWarmer` does not seem to have any tests at all 😮

leofeyer (approved):

PR #3638, opened 2021-11-03
Add an insert tags service

ausi:
Currently a draft...


Toflar:
A cool new feature here would be to add some max depth/recursion detection to prevent endless loops which are currently possible. I think we could throw an exception when entering level...say...50.

ausi:
> A cool new feature here would be to add some max depth/recursion detection to prevent endless loops which are currently possible.

Done in ff74edb40a3eced52787cce15b75844400e60ad6

core-bundle/src/InsertTag/ChunkedText.php
```diff
@@ -10,7 +10,7 @@
  * @license LGPL-3.0-or-later
  */
 
-namespace Contao\CoreBundle\Twig\Interop;
+namespace Contao\CoreBundle\InsertTag;
 
 /**
  * @experimental
```
m-vo: Should we remove the `experimental` flag for this class?

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -62,8 +69,43 @@ public static function reset()
 	 * @param boolean $blnAsChunks If true a ChunkedText object is returned instead of the string
 	 *
 	 * @return string|ChunkedText The text with the replaced tags
+	 *
+	 * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0.
+	 *             Use the InsertTagParser service instead.
 	 */
 	public function replace($strBuffer, $blnCache=true, $blnAsChunks=false)
```
m-vo: If this will get merged into 4.13 we could probably remove `$blnAsChunks` again (it's not released yet) here and just keep it in the internal method.

core-bundle/src/Twig/Runtime/InsertTagRuntime.php
```diff
@@ -12,34 +12,34 @@
...
+        $this->insertTagParser = $insertTagParser;
     }
 
     public function renderInsertTag(string $insertTag): string
     {
-        return $this->replaceInsertTags('{{'.$insertTag.'}}');
+        return $this->insertTagParser->render($insertTag);
     }
 
     public function replaceInsertTags(string $text): string
```
m-vo, line 36: Should we rename the methods here as well to match those of the `InsertTagParser`?

m-vo (commented):
I like this a lot. :heart:

core-bundle/src/Twig/Runtime/InsertTagRuntime.php
ausi: I think the current names are OK, but It’s your call. I like that `raw` is included in the `replaceInsertTagsChunkedRaw()` method name.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: Great idea, I removed the paramter completely in 77dcae0f97225927cd647407c3ecc87583b98538

core-bundle/src/InsertTag/ChunkedText.php
ausi: Done in 0b95a171f81e7aff76dc00b4e42a92d766c0ba87

m-vo (dismissed):
With `@experimental` now gone, we should probably mark the constructor of `ChunkedText` `@internal`, because the alternating raw/text chunks are probably an implementation detail of the legacy `InsertTag` class.

ausi:
> we should probably mark the constructor of `ChunkedText` `@internal`

Done in 2401f966883e3d4d07e8df8acb8afc165d6ef4f3

m-vo (approved):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -33,6 +33,13 @@
  */
 class InsertTags extends Controller
 {
+	private const MAX_NESTING_LEVEL = 100;
+
+	/**
+	 * @var int
+	 */
+	private static $intRecursionCount = 0;
```
leofeyer, line 41:
```suggestion
	private static int $intRecursionCount = 0;
```

core-bundle/src/InsertTag/InsertTagParser.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return $chunked[0][1];
+    }
+
+    public function reset(): void
+    {
+        InsertTags::reset();
+    }
+
+    private function callLegacyClass(string $input, bool $allowEsiTags): ChunkedText
```
leofeyer, line 66: It is rather `$preserveUncachedTags` than `$allowEsiTags`, but since it is a private method, it should not matter.

leofeyer (approved):

core-bundle/src/InsertTag/InsertTagParser.php
ausi: AFAIK, all insert tags get always replaced. With `$blnCache = true` some of the tags get replaced with ESI tags, with `$blnCache = false` all tags get replaced directly. Because I always have to check the code to know what `$blnCache` does I named this parameter `$allowEsiTags` because I think this describes it best.
It doesn’t really “preserve uncached tags” I think.

leofeyer:
Thank you @ausi.

PR #3643, opened 2021-11-04
Fix the DB query in the Versions class

### LINKED ISSUE #3642: Prepared statement fails in Contao\Versions
```An exception occurred while executing a query: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens```

After pulling in 3cadec7dfd444eefced20da70d5ca21408b3ec82 I get the error message mentioned above after logging into the backend. It originates in the Versions class, when counting versions:

https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/classes/Versions.php#L655-L656

I logged in as Admin, hence the ` AND userid=?` part was not added to the prepared statement. The user id however is always passed to `execute()`.

### PR DESCRIPTION
**Pull request by leofeyer**
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3642
| Docs PR or issue | -

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -651,9 +651,16 @@ class Versions extends Controller
 		$objUser = BackendUser::getInstance();
 		$objDatabase = Database::getInstance();
 
+		$params = array();
+
+		if (!$objUser->isAdmin)
+		{
+			$params[] = $objUser->id;
+		}
+
 		// Get the total number of versions
 		$objTotal = $objDatabase->prepare("SELECT COUNT(*) AS count FROM tl_version WHERE editUrl IS NOT NULL" . (!$objUser->isAdmin ? " AND userid=?" : ""))
-								->execute($objUser->id);
+								->execute(...$params);
 
 		$intLast   = ceil($objTotal->count / 30);
 		$intPage   = Input::get('vp') ?? 1;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by Toflar — DISMISSED** [review]
Not sure why the CI stuff is in this PR? But looks good anyway :)

File: core-bundle/src/Resources/contao/classes/Versions.php [hunk 1]
```diff
@@ -651,9 +651,16 @@ public static function addToTemplate(BackendTemplate $objTemplate)
 		$objUser = BackendUser::getInstance();
 		$objDatabase = Database::getInstance();
 
+		$params = array();
+
+		if (!$objUser->isAdmin)
+		{
+			$params[] = $objUser->id;
+		}
+
```
m-vo: Less negation and shorter?
```suggestion
		$params = $objUser->isAdmin ? array() : array($objUser->id);
```

m-vo (dismissed):

bytehead (dismissed):

ausi (dismissed):

core-bundle/src/Resources/contao/classes/Versions.php
leofeyer: Changed in 1f77c2334dddb2a1ca2341f2abb1ac6f49392c48.

leofeyer:
> Not sure why the CI stuff is in this PR? But looks good anyway :)

Removed in 1f77c2334dddb2a1ca2341f2abb1ac6f49392c48.

ausi (approved):

PR #3644, opened 2021-11-04
Allow MODE_PARENT without child_record_callback

fritzmg:
Currently you _always_ have to define a `child_record_callback`, if you use `MODE_PARENT` for your DCA. This makes sense for `tl_content`, as you want to render the content elements there, rather than any specific labels. However, for other DCAs that simply want to show a list of data records for a specific parent, it's a bit cumbersome that you always have to define a `child_record_callback`, even if it's just a simple one (like [here](https://docs.contao.org/dev/guides/dca/#list-1)). And it can be a little [confusing](https://github.com/contao/docs/pull/867) as well.

This PR might be a bit bold, but let's see 😉. It moves the general label creation into its own method (thus also reducing duplicated code) and uses this as the fallback label, when no `child_record_callback` was defined.

_Note:_ code review might be a bit confusing, since it removes a big `if { … }` block and thus everything moves to the left. I wish Github would be able to detect such whitespace only changes, just like VSCode or presumably PHPStorm can.


fritzmg:
_Note:_ for easier review, I have re-added the indentations of the removed condition block in `DC_Table::parentView` for now.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6330,6 +6202,104 @@ public function initPicker(PickerInterface $picker)
 
 		return $attributes;
 	}
+
+	/**
+	 * @param array  $row   The data record
+	 * @param string $table The table of the data record
+	 */
+	private function getLabel($row, $table = null): string
```
Toflar: Can we rename this method? So it answers the question to "Get label of what?"

Toflar (commented):
That's been on my list forever but I've never come round to create a PR.
I'm very 👍 on this :)

core-bundle/src/Resources/contao/drivers/DC_Table.php
fritzmg: May be `getRecordLabel`? And then add _Generates the label for a given record as defined in its DCA_ to the docblock.

news-bundle/src/Resources/contao/dca/tl_news.php
```diff
@@ -81,8 +80,11 @@
 			'fields'                  => array('date'),
 			'headerFields'            => array('title', 'jumpTo', 'tstamp', 'protected', 'allowComments'),
 			'panelLayout'             => 'filter;sort,search,limit',
-			'child_record_callback'   => array('tl_news', 'listNewsArticles'),
-			'child_record_class'      => 'no_padding'
```
fritzmg, line 85: _Note:_ this seems to be a remnant of old Contao versions, as such a CSS class does not exist anymore.

core-bundle/src/Resources/contao/drivers/DC_Table.php
bezin: I did not check the function line by line, but I just wanted to note this message overlaps with the universal picker and the improved undo module (#3498), which both need to display the label, too. Is there any way to actually share this logic?

fritzmg: @bezin good point. May be a DCA utility service of some sort, that can handle such tasks?

fritzmg:
I have renamed the method to `generateRecordLabel` and moved it into the `DataContainer` parent class, as the function is pretty agnostic towards the precise data container implementation, with some exceptions (like showing labels from related entries). It is also a public static method now, so that @bezin could use it like this:

```php
DataContainer::generateRecordLabel($row, $table);
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1541,6 +1541,106 @@ public static function getDriverForTable(string $table): string
...
 		return $dataContainer;
 	}
+
+	/**
+	 * Generates the label for a given data record according to the DCA configuration.
+	 *
+	 * @param array  $row   The data record
+	 * @param string $table The name of the data container
+	 */
+	public static function generateRecordLabel($row, $table): string
```
Toflar: I don't think this is a good idea, is it? It's a static method that has dependencies to other services.
But then again, this happens all over the place in the "old" DCA stuff.

fritzmg: Indeed. Since the future of the DCA is so unclear (see also the other PRs/discussions) I wasn't sure whether I should really introduce a dedicated helper service just for this.

Toflar: Yeah, let's leave it like that then. Every DC is instantiated using `new ...` directly anyway and does not get the DB injected. We can only fix that by completely rewriting lots of parts. Possibly based on the new config stuff then which @SeverinGloeckle seems to be working on 😉

Toflar (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
m-vo: Can we maybe use an instance method instead? Then we can at least refactor the service usages to a locator or shared properties at some point. 

Personally, I'd go with a helper service because I think this functionality is quite orthogonal to the rest of the DC concept (maybe the `row` parameter needs to be named `fields`). Even after the data containers are gone this could IMHO live on for BC without hurting anyone. But up to you.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1541,6 +1541,106 @@ public static function getDriverForTable(string $table): string
...
+	{
+		$showFields = $GLOBALS['TL_DCA'][$table]['list']['label']['fields'] ?? array('id');
```
aschempp: I don't think we should fall back to the ID. How does the current implementation behave if the fields are not defined? Maybe throw an exception that either fields or label_callback must be defined?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4426,151 +4383,152 @@ protected function parentView()
...
+					$this->import($strClass);
+					$return .= '</div>' . $this->$strClass->$strMethod($row[$i]) . '</div>';
+				}
+				elseif (\is_callable($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['child_record_callback'] ?? null))
+				{
+					$return .= '</div>' . $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['child_record_callback']($row[$i]) . '</div>';
+				}
+				else
+				{
+					$return .= '</div><div class="tl_content_left">' . parent::generateRecordLabel($row[$i], $this->strTable) . '</div></div>';
```
aschempp: Why does this have a `tl_content_left` class even though the other `if-else` do not?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3940,48 +3938,7 @@ protected function generateTree($table, $id, $arrPrevNext, $blnHasSorting, $intM
 			$return .= '<a href="' . $this->addToUrl('ptg=' . $id) . '" title="' . StringUtil::specialchars($alt) . '" onclick="Backend.getScrollOffset();return AjaxRequest.toggleStructure(this,\'' . $node . '_' . $id . '\',' . $level . ',' . ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? '') . ')">' . Image::getHtml($img, '', 'style="margin-right:2px"') . '</a>';
 		}
 
-		foreach ($showFields as $k=>$v)
-		{
-			// Decrypt the value
-			if ($GLOBALS['TL_DCA'][$table]['fields'][$v]['eval']['encrypt'] ?? null)
-			{
-				$objRow->$v = Encryption::decrypt(StringUtil::deserialize($objRow->$v));
```
aschempp, line 4067: this previously decrypted the database value, which is later on forwarded to the `label_callback`. With these changes, this is no longer the case, but do we still support `encrypt` anyway?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1541,6 +1541,106 @@ public static function getDriverForTable(string $table): string
...
+				else
+				{
+					$args[$k] = $row[$v] ? Date::parse(Config::get('datimFormat'), $row[$v]) : '-';
+				}
+			}
+			elseif (($GLOBALS['TL_DCA'][$table]['fields'][$v]['eval']['isBoolean'] ?? null) || (($GLOBALS['TL_DCA'][$table]['fields'][$v]['inputType'] ?? null) == 'checkbox' && !($GLOBALS['TL_DCA'][$table]['fields'][$v]['eval']['multiple'] ?? null)))
+			{
+				$args[$k] = $row[$v] ? $GLOBALS['TL_LANG']['MSC']['yes'] : $GLOBALS['TL_LANG']['MSC']['no'];
+			}
+			elseif (isset($row[$v]))
```
aschempp, line 1641: this whole condition seems not to be present previously, where does that come from?

aschempp (changes requested):

core-bundle/src/Resources/contao/drivers/DC_Table.php
fritzmg: Hm, true. It is deprecated, but still supported otherwise.

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: > Can we maybe use an instance method instead?

The idea was that @bezin for example could use this method for his PR as well. Although he might have a DataContainer instance there as well? 🤔

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: It is present previously: https://github.com/contao/contao/blob/3ce5a595d30a4cb26361f03f368f9bf38faa560e/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4935-L4962

core-bundle/src/Resources/contao/drivers/DC_Table.php
fritzmg: On second thought: `Encryption::decrypt` will just result in a `BadMethodCallException` anyway. So I think we can just dispense this.

core-bundle/src/Resources/contao/drivers/DC_Table.php
fritzmg: It works without it too, but all the core's `child_record_callback`s always return the label wrapped in `<div class="tl_content_left">`. May be that was necessary in previous Contao versions?

fritzmg: There is actually a tiny visual difference between wrapping it with `tl_content_left` and without. The line-height is different. Without the wrapper the line-height is 14px and with the wrapper it is 16px.

/cc @leofeyer

aschempp (commented):
I find it impossible to tell if this is all right, since just noone knows whats really going on in DC anyway. But I compared it to the Picker method, and it's pretty much identical (but even a bit better).

It seems to not handle the `label_callback` though, and I'm wondering if it should. If so, we might need to pass an (optional?) DC object into the method because that would be forwarded to the callback.

fritzmg:
> I find it impossible to tell if this is all right, since just noone knows whats really going on in DC anyway. But I compared it to the Picker method, and it's pretty much identical (but even a bit better).

I know, that's why I wrote "this may be a bit bold" 😁.

> It seems to not handle the label_callback though, and I'm wondering if it should. If so, we might need to pass an (optional?) DC object into the method because that would be forwarded to the callback.

True, I'll have a look at integrating the label callback for the new automatic child record labels.

core-bundle/src/Resources/contao/drivers/DC_Table.php
fritzmg: On third thought: may be we need to keep it. Previously you would get an exception as soon as the value would be decrypted. If I remove this then no exception will be thrown.

fritzmg:
The Picker now also uses `DataContainer::generateRecordLabel` and the `label_callback` is now also executed for `DataContainer::parentView`. The last commit also fixes an oversight for the `listView` - we need the columns separated there. Thus I adjusted `DataContainer::generateRecordLabel` so that you can pass the `$args` array by reference. Not exactly pretty, but could not come up with something better so far.

aschempp (commented):
Shouldn't the callback be executed _inside_ the method to make sure it is always done?

fritzmg:
> Shouldn't the callback be executed inside the method to make sure it is always done?

Easier said than done. The `label_callback` in its current form is very specific to the actual list mode. Depending on the mode, the callback arguments and return values differ. Thus I decided to keep this out of the method. Currently we have 3 different method signatures and two different return types. See the implementation in the `Picker` - this gives you a good overview on the differences.

Also depending on the mode, the label might be adjusted depending on the presence of the callback (see `DC_Table::generateTree` for example).

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4547,151 +4504,168 @@ protected function parentView()
...
+					$this->import($strClass);
+					$return .= '</div>' . $this->$strClass->$strMethod($row[$i]) . '</div>';
+				}
+				elseif (\is_callable($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['child_record_callback'] ?? null))
+				{
+					$return .= '</div>' . $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['child_record_callback']($row[$i]) . '</div>';
+				}
+				else
+				{
+					$label = parent::generateRecordLabel($row[$i], $this->strTable);
```
leofeyer: Why do you generate the label here if it is potentially overwritten by a label_callback? Maybe we should generate it in an `else` condition instead.

fritzmg: The `label_callback` requires the original label to be present.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1583,6 +1583,108 @@ public static function getDriverForTable(string $table): string
...
+					$args[$k] = $row[$v] ? Date::parse(Config::get('timeFormat'), $row[$v]) : '-';
+				}
+				else
+				{
+					$args[$k] = $row[$v] ? Date::parse(Config::get('datimFormat'), $row[$v]) : '-';
+				}
+			}
+			elseif (($GLOBALS['TL_DCA'][$table]['fields'][$v]['eval']['isBoolean'] ?? null) || (($GLOBALS['TL_DCA'][$table]['fields'][$v]['inputType'] ?? null) == 'checkbox' && !($GLOBALS['TL_DCA'][$table]['fields'][$v]['eval']['multiple'] ?? null)))
+			{
+				$args[$k] = $row[$v] ? $GLOBALS['TL_LANG']['MSC']['yes'] : $GLOBALS['TL_LANG']['MSC']['no'];
```
leofeyer, line 1639: This has been handled differently in tree view before:

https://github.com/contao/contao/blob/6e3156a0703917cd137d26ecbe3765130aed8276/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4085-L4088

Did you intentionally change it?

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1583,6 +1583,108 @@ public static function getDriverForTable(string $table): string
...
+	/**
+	 * Generates the label for a given data record according to the DCA configuration.
+	 *
+	 * @param array  $row   The data record
+	 * @param string $table The name of the data container
+	 * @param array  $args  The generated labels for each column
+	 */
+	public static function generateRecordLabel(array $row, string $table, array &$args = array()): string
+	{
+		$mode = $GLOBALS['TL_DCA'][$table]['list']['sorting']['mode'] ?? self::MODE_SORTED;
```
leofeyer: The `$mode` variable is not used anywhere. Can it be removed?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1583,6 +1583,108 @@ public static function getDriverForTable(string $table): string
...
+		$showFields = $GLOBALS['TL_DCA'][$table]['list']['label']['fields'];
```
```suggestion
		$showFields = $GLOBALS['TL_DCA'][$table]['list']['label']['fields'] ?? array();
```

Can we be sure that `$GLOBALS['TL_DCA'][$table]['list']['label']['fields']` is always set or should we add a fallback for PHP 8?

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: @bezin Do you have a DataContainer instance or do you need the static method?

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: Not intentionally, no. But which version is more correct? Probably the one from the list view? (i.e. the one currently implemented in the method)

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: Currently the DCA implementation assumes that it must be set. See also https://github.com/contao/contao/pull/3644#discussion_r759940288

core-bundle/src/Resources/contao/classes/DataContainer.php
fritzmg: Actually I think we can make it non static. A `DataContainer` instance is available here: https://github.com/contao/contao/pull/3498/files#diff-f2b278b76e6ee91923f9b1ebfdedf63d82cf70c4ef365467acb11b011b7642adR91

bezin: Yes, I have one as I have to create one for the `label_callback` anyways.

fritzmg: Then I'll change it accordingly.

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: IDK. Can you post a screenshot of how the change affects the tree view labels?

fritzmg: The core does not use tree views with boolean fields in their label. But you if you add

```php
// contao/dca/tl_page.php
$GLOBALS['TL_DCA']['tl_page']['list']['label']['fields'] = ['title', 'published'];
$GLOBALS['TL_DCA']['tl_page']['list']['label']['format'] = '%s <span style="opacity:0.5;font-size:0.8em;">published: %s</span>';
```

then this is what it looks like now:

image

and this is what it looks like with the current state of the PR:

image

I can of course check the mode and make sure that in the tree view it will show the field's label when true and not yes/no. Though I think this should be consistent across _all_ views.

leofeyer: > Though I think this should be consistent across all views.

I agree. 👍

fritzmg:
For the record: I can still try to integrate the label callbacks into the method itself, if desired. This would save us more code duplication (at least in the Picker and the new Undo view).

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: Adding `tl_content_left` is correct. 👍

leofeyer (dismissed):
The PR is ready to be merged now. You still want to change the following, don't you?

fritzmg:
> The PR is ready to be merged now. You still want to change the following, don't you?

Yes, I think it's better if we do it now (we can't really change it afterwards, unless we introduce a new service or method). Will do it later today, probably.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4059,70 +4057,10 @@ protected function generateTree($table, $id, $arrPrevNext, $blnHasSorting, $intM
...
-		elseif (\is_callable($GLOBALS['TL_DCA'][$table]['list']['label']['label_callback'] ?? null))
-		{
-			$return .= $GLOBALS['TL_DCA'][$table]['list']['label']['label_callback']($objRow->row(), $label, $this, '', false, $blnProtected, $isVisibleRootTrailPage);
-		}
-		else
-		{
-			$return .= Image::getHtml('iconPLAIN.svg') . ' ' . $label;
-		}
-
+		$return .= $this->generateRecordLabel($objRow->row(), $table, $blnProtected, $isVisibleRootTrailPage);
```
leofeyer, line 4063: It seems that you have dropped the `$return .= Image::getHtml('iconPLAIN.svg') . ' ' . $label;` case, haven't you?

fritzmg: Added in 933ad2c00

leofeyer (changes requested):
Unfortunately, there is still something odd.

### System log

**Before**



**After**



### Member groups

**Before**



**After**



### Comments

**Before**



**After**

fritzmg:
Fixed in 2b6e17fccf0d80a7dbbe1862aebbc6503c717304

leofeyer (approved):

leofeyer (changes requested):
It seems we are still missing something:

**Before** 



**After**

leofeyer:
It seems to be related to the protection status.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1583,6 +1583,159 @@ public static function getDriverForTable(string $table): string
...
+		$mode = $GLOBALS['TL_DCA'][$table]['list']['sorting']['mode'] ?? self::MODE_SORTED;
+
+		// Execute label_callback
+		if (\is_array($labelConfig['label_callback'] ?? null) || \is_callable($labelConfig['label_callback'] ?? null))
+		{
+			if (\in_array($mode, array(self::MODE_TREE, self::MODE_TREE_EXTENDED)))
+			{
+				if (\is_array($labelConfig['label_callback'] ?? null))
+				{
+					$label = System::importStatic($labelConfig['label_callback'][0])->{$labelConfig['label_callback'][1]}($row, $label, $this, '', $protected, $isVisibleRootTrailPage);
```
leofeyer: This is why the labels of the protected pages are missing: the original arguments were `$objRow->row(), $label, $this, '', false, $blnProtected, $isVisibleRootTrailPage`. It seems you have dropped the fifth argument?

fritzmg: Hm, no, `$blnProtected` is passed [here](https://github.com/contao/contao/pull/3644/files#diff-e4bf1535e70e3fbc2b9692ba8add4c45237305ff815df14c7ba9c9ce53dc9204R4063) and then passed to the callback as the 5th argument in the line you referenced, but I'll analyse.

leofeyer: Adding the fifth argument again fixes the issue for me:



Tested in the regular view and in the picker view. No idea if I am overlooking something though.

fritzmg: Aaah, yes... the argument that is always `false` (and the one before is always an empty string) 😁 I forgot about that one.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3645, opened 2021-11-04
Restore backwards compatiblilty for DB Statement

### LINKED ISSUE #3642: Prepared statement fails in Contao\Versions

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3642

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
@@ -284,6 +284,10 @@ class Statement
 		{
 			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
 		}
+		// Executing the query with the wrong number of params causes a
+		// DriverException in PDO and an ArgumentCountError in mysqli. For
+		// backwards compatibility we catch them and fix the number of arguments
+		// if possible.
 		catch (DriverException|\ArgumentCountError $exception)
 		{
 			if (!$arrParams || \count($arrParams) <= ($intTokenCount = $this->countQueryTokens()))
@@ -296,7 +300,14 @@ class Statement
 			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
 
 			// Only trigger the deprecation if the parameter count was the reason for the exception and the previous call did not throw
-			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s::execute(null)" or passing more parameters than "?" tokens has been deprecated and will no longer work in Contao 5.0. Use the correct number of parameters instead.', __CLASS__);
+			if ($this->arrLastUsedParams === array(null))
+			{
+				trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s::execute(null)" has been deprecated and will no longer work in Contao 5.0. Omit the NULL parameters instead.', __CLASS__);
+			}
+			else
+			{
+				trigger_deprecation('contao/core-bundle', '4.13', 'Passing more parameters than "?" tokens has been deprecated and will no longer work in Contao 5.0. Use the correct number of parameters instead.', __CLASS__);
+			}
 		}
 
 		// No result set available
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
@@ -284,18 +284,19 @@ class Statement
 		{
 			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
 		}
-		catch (DriverException $exception)
+		catch (DriverException|\ArgumentCountError $exception)
 		{
-			if ($arrParams !== array(null))
+			if (!$arrParams || \count($arrParams) <= ($intTokenCount = $this->countQueryTokens()))
 			{
 				throw $exception;
 			}
 
-			// Backwards compatibility for calling Statement::execute(null)
-			$this->statement = $this->resConnection->executeQuery($this->strQuery);
+			$arrParams = \array_slice($arrParams, 0, $intTokenCount);
+
+			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
 
 			// Only trigger the deprecation if the parameter count was the reason for the exception and the previous call did not throw
-			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s::execute(null)" has been deprecated and will no longer work in Contao 5.0. Use "%s::execute() instead."', __CLASS__, __CLASS__);
+			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s::execute(null)" or passing more parameters than "?" tokens has been deprecated and will no longer work in Contao 5.0. Use the correct number of parameters instead.', __CLASS__);
 		}
 
 		// No result set available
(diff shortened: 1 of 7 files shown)


core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -284,18 +284,19 @@ static function ($varParam)
 		{
 			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
 		}
-		catch (DriverException $exception)
+		catch (DriverException|\ArgumentCountError $exception)
 		{
-			if ($arrParams !== array(null))
+			if (!$arrParams || \count($arrParams) <= ($intTokenCount = $this->countQueryTokens()))
```
leofeyer: How about moving this logic into the `countQueryTokens()` method as well and renaming it to `isTokenCountMismatch()` or something else? Then we could simplify this condition:

```suggestion
			if ($this->isTokenCountMismatch($arrParams))
```

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -750,6 +750,13 @@ public function preventSaving($blnKeepClone=true)
 	 */
 	public static function findByPk($varValue, array $arrOptions=array())
 	{
+		if ($varValue === null)
+		{
+			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s::findByPk(null)" has been deprecated and will no longer work in Contao 5.0.', __CLASS__);
```
leofeyer: Not sure if this is good. What if I want to pass `null` as argument to e.g. get `SELECT * FROM table WHERE field=null`?

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
ausi: Not possible because `$intTokenCount` is reused in the code below.

core-bundle/src/Resources/contao/library/Contao/Model.php
ausi: That’s why I only added it for `findByPk`, AFAIK we don’t support nullable primary key columns.

m-vo: IMHO nullable primary keys never make sense as it's semantically the same as deleting the row. :smile:

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -308,6 +309,15 @@ static function ($varParam)
 		return new Result($this->statement, $this->strQuery);
 	}
 
+	/**
+	 * @return int
+	 */
+	private function countQueryTokens()
+	{
+		// Backwards compatibility
```
leofeyer: How is this a BC layer?

core-bundle/src/Resources/contao/library/Contao/Model.php
leofeyer: But we also have `findById()` and `findByIdOrAlias()`. 🙈

core-bundle/src/Routing/RouteProvider.php
```diff
@@ -301,7 +301,7 @@ private function findRootPages(string $httpHost): array
             $rootPages = $pages->getModels();
         }
 
-        $pages = $pageModel->findBy(["tl_page.alias='index' OR tl_page.alias='/'"], null);
+        $pages = $pageModel->findAll(['column' => ["tl_page.alias='index' OR tl_page.alias='/'"]]);
```
leofeyer: I like the original syntax more. Can't we preserve this kind of usage?

core-bundle/src/Resources/contao/library/Contao/Model.php
ausi: Yes, but `findByIdOrAlias()` could potentially allow `null` so we cannot disallow that I think.

Calling `findByPk(null)` is most certainly wrong, but for all other columns we cannot assume that.

core-bundle/src/Routing/RouteProvider.php
ausi: I don’t think so because `null` can be an actual value that gets used by a `?` in the query.

How about the following?
```suggestion
        $pages = $pageModel->findBy(["tl_page.alias=? OR tl_page.alias=?"], ['index', '/']);
```

ausi:
```suggestion
        $pages = $pageModel->findBy(["tl_page.alias='index' OR tl_page.alias='/'"], []);
```
This could also work maybe (need to test :)

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
ausi: It counts the tokens the way the old implementation did which is error-prone and only needed for deprecated calls with too many arguments.

core-bundle/src/Resources/contao/library/Contao/Model.php
leofeyer: Not even for `findById()`? `findByPk()` is actually just an alias for `findById()`.

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
leofeyer: How will we count the tokens in the new implementation?

core-bundle/src/Routing/RouteProvider.php
leofeyer: The first one is a good compromise. 👍

core-bundle/src/Resources/contao/library/Contao/Model.php
ausi: Added deprecation for `findById()` in 325ad84cfa783bbe8f95c006f785da1db02f41b3

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
ausi: We will not. Once the deprecation layer is removed, for a mismatch between the number of tokens and arguments the database will throw an exception.

core-bundle/src/Routing/RouteProvider.php
ausi: Fixed in 8d303adc813138d747225458a9633c7678d819eb

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
leofeyer: What do you think about https://github.com/ausi/contao-contao/pull/1? I find it more clear this way.

ausi: Both exceptions handle the same case. One is for pdo, the other for mysqli. So we cannot split them up.

ausi: But I should probably add a comment that clarifies this.

ausi: Added a comment in e7d12adb8ebdabbc55563001c8e5278bb63ad98f

leofeyer: WDYT about https://github.com/ausi/contao-contao/pull/2?

m-vo (dismissed):

leofeyer (approved):

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #3647, opened 2021-11-05
Set language of redirect page in global context

rabauss:
Currently the language of a redirect page is not set in global context. This is especially a problem if you use language dependent insert tags in the redirect.

Affected Version: 4.9+

I'm not sure if we need to add more lines from the `prepare` method of `PageRegular`. The supposed lines are sufficient for my problem with the missing language context.

https://github.com/contao/contao/blob/2a19b186afd59495e9fbd5174d0deb40543c5079/core-bundle/src/Resources/contao/pages/PageRegular.php#L63-L69

### DIFF
--- a/core-bundle/src/Resources/contao/pages/PageRedirect.php
+++ b/core-bundle/src/Resources/contao/pages/PageRedirect.php
@@ -27,6 +27,8 @@ class PageRedirect extends Frontend
 	 */
 	public function generate($objPage)
 	{
+		$this->prepare($objPage);
+
 		$this->redirect($this->replaceInsertTags($objPage->url, false), $this->getRedirectStatusCode($objPage));
 	}
 
@@ -39,6 +41,8 @@ class PageRedirect extends Frontend
 	 */
 	public function getResponse($objPage)
 	{
+		$this->prepare($objPage);
+
 		return new RedirectResponse($this->replaceInsertTags($objPage->url, false), $this->getRedirectStatusCode($objPage));
 	}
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

leofeyer:
> This is especially a problem if you use language dependent insert tags in the redirect.

What exactly does that mean?

ausi:
> > This is especially a problem if you use language dependent insert tags in the redirect.
> 
> What exactly does that mean?

I would guess something like this:  
`{{iflng::de}}{{link_url::1}}{{iflng}}{{iflng::en}}{{link_url::2}}{{iflng}}`

rabauss:
Well, the example from @ausi would be a matching one.
We have a custom entity with DCMultilingual which can be picked and we generate a inserttag with ID of the fallback language which is then replaced by the language model if there is the current language in the global context.

Of course I could also use the language from global `$objPage` but I thought with this PR we get a reusable solution ;-)

core-bundle/src/Resources/contao/pages/PageRedirect.php
```diff
@@ -53,6 +57,21 @@ protected function getRedirectStatusCode($objPage)
 	{
 		return ($objPage->redirect == 'temporary') ? 303 : 301;
 	}
+
+	/**
+	 * @param PageModel $objPage
+	 */
+	private function prepare($objPage)
+	{
+		$GLOBALS['TL_KEYWORDS'] = '';
```
leofeyer, line 66:
```suggestion
```

core-bundle/src/Resources/contao/pages/PageRedirect.php
```diff
@@ -53,6 +57,21 @@ protected function getRedirectStatusCode($objPage)
...
+		$GLOBALS['TL_LANGUAGE'] = $objPage->language;
```
leofeyer, line 67:
```suggestion
		$GLOBALS['TL_LANGUAGE'] = $objPage->language;
```

leofeyer:
```suggestion
		$GLOBALS['TL_LANGUAGE'] = $objPage->language;
```

leofeyer (changes requested):

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @rabauss.

PR #3654, opened 2021-11-09
Ensure defined admin email is always used as sender

### LINKED ISSUE #3573: Custom sender mail address in OptInToken
Currently, it is not possible to use a custom sender address when sending an email with the OptIn-Service.
In the `Contao\CoreBundle\OptIn\OptInToken` a new instance of the `Contao\Email`-class is created. Since there is no sender-address, the Email-class uses the administration-email from the settings.

However, sometimes it would be useful to use another email address.
When you are in a multi-domain-system there could be different page-administrators per root page. Sending an email with the OptIn-Service therefore should use the administrator email of the current root page (if there is a root page (frontend)).

There is already the possibility to extend the OptIn-Service in a way that you can add a custom sender address. But you have to copy a lot of code from the original OptInToken-class because the service uses exactly this class.

```php
return new OptInToken($optIn, $this->framework);
```

https://github.com/contao/contao/blob/4.x/core-bundle/src/OptIn/OptIn.php#L61

Would it be possible to add a method to the `OptInToken`-class to set a custom sender address?
Something like:

```php
public function setSender(string $address, ?string $name = null): self
{
    $this->senderAddress = $address;
    if (null !== $name) {
        $this->senderName = $name;
    }

    return $this;
}
```

So it would be possible to set the sender address for the email instance:

```php
/** @var Email $email */
$email = $this->framework->createInstance(Email::class);
$email->subject = $this->model->emailSubject;
$email->text = $this->model->emailText;

if ('' !== $this->senderAddress) {
    $email->from = $this->senderAddress;
}

if ('' !== $this->senderName) {
    $email->fromName = $this->senderName;
}

$email->sendTo($this->model->email);
```

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3573
| Docs PR or issue | -

This PR fixes #3573 by setting the `$GLOBALS['TL_ADMIN_EMAIL']` value as the sender within `OptInToken`, if available. This PR also ensures that `Contao\Email` _always_ uses `$GLOBALS['TL_ADMIN_EMAIL']` if available.

### DIFF
--- a/core-bundle/src/OptIn/OptInToken.php
+++ b/core-bundle/src/OptIn/OptInToken.php
@@ -135,6 +135,8 @@ class OptInToken implements OptInTokenInterface
         $email = $this->framework->createInstance(Email::class);
         $email->subject = $this->model->emailSubject;
         $email->text = $this->model->emailText;
+        $email->from = $GLOBALS['TL_ADMIN_EMAIL'] ?? null;
+        $email->fromName = $GLOBALS['TL_ADMIN_NAME'] ?? null;
         $email->sendTo($this->model->email);
     }
 
--- a/core-bundle/src/Resources/contao/library/Contao/Email.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Email.php
@@ -433,7 +433,15 @@ class Email
 		// Add the administrator e-mail as default sender
 		if (!$this->strSender)
 		{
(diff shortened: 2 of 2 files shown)


Toflar (approved):

m-vo (approved):

leofeyer:
Thank you @fritzmg.

PR #3655, opened 2021-11-09
Unset page model again after parsing in backend

Toflar:
Otherwise our profiler data collector tries to load the page layout in the back end.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -1013,12 +1013,10 @@ class tl_page extends Contao\Backend
 			return '';
 		}
 
-		global $objPage;
+		// Set the global page object, so we can replace the insert tags
+		$GLOBALS['objPage'] = $model;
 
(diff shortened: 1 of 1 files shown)


aschempp (approved):

fritzmg (approved):

bytehead (approved):

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1028,6 +1026,10 @@ static function ($strVal)
 				explode('{{page::pageTitle}}', $layout->titleTag ?: '{{page::pageTitle}} - {{page::rootPageTitle}}', 2)
 			)
 		);
+
+		unset($GLOBALS['objPage']);
```
m-vo, line 1030: Should we only set and unset if `$GLOBALS['objPage']` wasn't set before?

leofeyer:
It seems @Toflar and I have both created a PR (see #3656). 🙈

PR #3659, opened 2021-11-09
Add a cache tag service for entity/model classes

### LINKED ISSUE #3648: Shortcut for cache tagging with entities
Assume there is an entity `Foo` with table `tl_foo` of which all elements should be output in a `FooListFragment`, a single one in a `FooDetailsFragment` and maybe some more selected ones in a `SelectedFoosDetailsFragment`. 

You would then need to tag you response like so:
```php
// FooListFragment
$this->tagResponse(['contao.db.tl_foo']);

// FooDetailsFragment
$this->tagResponse(['contao.db.tl_foo.' . $element->getId()]);

// SelectedFoosDetailsFragment
$this->tagResponse(
    array_map(static fn(Foo $element): string => 'contao.db.tl_foo.' . $element->getId(), $elements);
);


```

Now, you're losing your abstraction - the `tl_foo` table name gets baked into the code and will probably be missed out when refactoring. And, though small, it's also repetitive code showing up at many places. 

I wonder if we could create a shortcut that works like this:

```php
// FooListFragment
$this->tagResponse(Foo::class);

// FooDetailsFragment
$this->tagResponse([$element]);

// SelectedFoosDetailsFragment
$this->tagResponse($elements);
```

I.e. when passing an entity class name, we substitute it with `['contao.db.<table-name>']` and if a collection of objects is passed we substitute each with `contao.db.<table-name>.<pk>`. All mappings (table name, pk) are available via doctrine's class metadata. 

The same would work with our models I guess.

Wdyt?

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Closes #3648 
| Docs PR or issue | see https://github.com/contao/docs/pull/990

This PR implements an `EntityCacheTags` service that can generate, tag a response with and invalidate cache tags in the form `contao.db.<name>`/`contao.db.<name>.<pk>` from entity classes/instances. 

There is also a method that auto detects the input that is now used inside `AbstractController#tagResponse()`. This allows a usage like so:

```php
// Output a blog post  with author and comments
public function action(BlogPost $post): Response {
    $this->tagResponse([$post, $post->getAuthor(), $post->getComments()];

    return $this->render('blog_post.html.twig', ['post' => $post]);
}
```

The service as well as the autodetection supports both doctrine entites, Contao models as well as their collection counterparts.


core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * @param iterable|string|object|null $target
+     *
+     * @return array<int, string>
+     */
+    public function getTags($target): array
+    {
+        if (empty($target)) {
```
ausi:
```suggestion
        if (!$target) {
```
Nitpick ☺️

core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,107 @@
...
+            return [];
+        }
+
+        if (\is_string($target)) {
+            if (class_exists($target, false)) {
```
ausi: As using classes with `FooBar::class` does not autoload the class I think this check could lead to weird bugs. How about instead checking if it is a valid PHP FQCN?

```suggestion
            if (preg_match('/^(?:[a-z_\x80-\xff][a-z0-9_\x80-\xff]*\\\\?)+(?<!\\\\)$/i', $target)) {
```

core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,107 @@
...
+                    return [$this->getTagForEntityClass($target)];
+                } catch (\InvalidArgumentException $e) {
+                    // ignore
+                }
+            }
+
+            return [$target];
+        }
+
+        if (is_iterable($target)) {
```
ausi: Could this lead to problems if an entity class implements `Traversable`?

core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        $getMetadata = function (string $className) {
+            try {
+                return $this->entityManager->getClassMetadata($className);
+            } catch (MappingException $e) {
+                return null;
+            }
+        };
+
+        return $this->classMetadata[$className] ?? ($this->classMetadata[$className] = $getMetadata($className));
```
ausi:
```suggestion
        $this->classMetadata[$className] ??= $getMetadata($className);

        return $this->classMetadata[$className];
```
Nitpick ☺️

core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,107 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Cache;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\Persistence\Mapping\MappingException;
+
+class EntityTagger
```
Toflar: Why is this called `Tagger` but it does not tag? It returns tags but it does not tag. Why does it not tag right away? ;)

ausi: Returning the tags instead of tagging right away could have the benefit that you could use its methods for clearing the cache based on entities too maybe?

m-vo: Correct, that's the idea. 

But I agree that the naming could be improved. Any suggestions? :slightly_smiling_face:

ausi: `EntityCacheTags`?

Toflar: Sure, I think I would inject the response tagger here and have two methods. One for fetching the tags and one for tagging right away (forwarding the tags to the response tagger). That way the name is correct and both options are available. Also, you then don't have to inject both, the response tagger plus this new service if you want to tag right away :)

m-vo: Just thinking: do we also want this concept to work with models? :thinking: Then we should maybe get rid of the `Entity` in the name.

fritzmg: > Just thinking: do we also want this concept to work with models? 🤔 Then we should maybe get rid of the `Entity` in the name.

Imho yes. But I think using "Entity" in the name is fine. I see "Entity" as a general term and not exclusive to just doctrine entities.

core-bundle/src/Cache/EntityTagger.php
m-vo: Agreed, this is probably better suited!

core-bundle/src/Cache/EntityTagger.php
m-vo: Hmm, maybe. But it's very unlikely to implement this on an entity. We could alternatively check for doctrine's `Collection` interface. Wdyt?

 The only important thing is to *not* use arrays, because then iterating and getting the id would not be lazy anymore. :slightly_smiling_face:

ausi: > We could alternatively check for doctrine's `Collection` interface. Wdyt?

👍

core-bundle/src/Cache/EntityTagger.php
```diff
@@ -0,0 +1,109 @@
...
+            try {
+                return $this->entityManager->getClassMetadata($className);
+            } catch (MappingException $e) {
+                return null;
+            }
+        };
+
+        $this->classMetadata[$className] ??= $getMetadata($className);
+
+        return $this->classMetadata[$className];
```
ausi: Looking at the code again, I think we can simplify it even more ☺️

```suggestion
        return $this->classMetadata[$className] ??= $getMetadata($className);
```

m-vo: Ha! I already did this, but didn't push it yet. :smile:

core-bundle/src/Cache/EntityTagger.php
m-vo: done so in  c30a621b3a8463b22f2fd5e3ddb6b2a4d0431668

m-vo:
I reworked the concept. Now the service is called `EntityCacheTags` and supports deriving tags, tagging and invalidating. There is also support for models built in.

m-vo:
I rebased the PR, resolved conflicts and updated the initial description. This should be ready to review.

core-bundle/src/Cache/EntityCacheTags.php
```diff
@@ -0,0 +1,319 @@
...
+    public function __construct(EntityManagerInterface $entityManager, ResponseTagger $responseTagger = null, CacheInvalidator $cacheInvalidator = null)
+    {
+        $this->entityManager = $entityManager;
+        $this->responseTagger = $responseTagger;
+        $this->cacheInvalidator = $cacheInvalidator;
+    }
+
+    /**
+     * Derives a cache tag from an entity class and returns it.
+     */
```
ausi, line 52:
```suggestion
     *
     * @param class-string $className
     */
```
Should we annotate the type `class-string` here so that the IDE and our analysis tools know this should not be some other string?

core-bundle/src/Cache/EntityCacheTags.php
```diff
@@ -0,0 +1,319 @@
...
+            ->getUnitOfWork()
+            ->getSingleIdentifierValue($instance)
+        ;
+
+        return sprintf('contao.db.%s.%s', $metadata->getTableName(), $identifier);
+    }
+
+    /**
+     * Derives a cache tag from a model class and returns it.
+     */
```
ausi, line 87:
```suggestion
     *
     * @param class-string<Model> $className
     */
```
Same here.

ausi (dismissed):

core-bundle/src/Cache/EntityCacheTags.php
m-vo: Good point! I like!

m-vo:
Rebased + conflicts resolved. Again…

ausi (approved):

Toflar (approved):
Nice work! ❤️

phpstan.neon.dist
```diff
@@ -38,6 +38,11 @@ parameters:
         # Ignore the wrong return type hint of the UrlGeneratorInterface::generate() method
         - '#Method Contao\\CoreBundle\\Picker\\AbstractPickerProvider::generateUrl\(\) never returns null so it can be removed from the return type\.#'
 
+        # Ignore wrong class-string when testing the guard that enforces it at runtime
+        -
+            message: '#Parameter \#1 \$className of method Contao\\CoreBundle\\Cache\\EntityCacheTags::getTagForModelClass\(\) expects class-string<Contao\\Model>, string given\.#'
```
leofeyer: This looks fishy. PHPStan handles generics very well, so this should rather be fixed than ignored, shouldn't it?

m-vo: The problem is: PHPStan follows the annotations in a stricter way than programmers do, in fact in our setting they are treated as truth even if PHP doesn't enforce them at runtime. In this case the guard should IMHO be there for good DX, so that you can a clear exception if the passed class string isn't a model class.

core-bundle/src/Cache/EntityCacheTags.php
```diff
@@ -0,0 +1,323 @@
...
+
+        return sprintf('contao.db.%s', \call_user_func([$className, 'getTable']));
+    }
+
+    /**
+     * Derives a cache tag from a model instance and returns it.
+     */
+    public function getTagForModelInstance(Model $instance): string
+    {
+        return sprintf('contao.db.%s.%s', $instance::getTable(), $instance->id);
```
leofeyer, line 102:
```suggestion
        return sprintf('contao.db.%s.%s', $instance->getTable(), $instance->id);
```

If we called the `getTable()` method non-statically, we could use `mockClassWithProperties()` in the unit tests instead of `(new \ReflectionClass())->newInstanceWithoutConstructor()`. WDYT?

m-vo: But it's a static method and calling it with `->` will hopefully be illegal at some point in the future, no? :smile: Static analyzers do not like this as well.

I'd prefer it this way round, but I leave this up to you.

leofeyer: > But it's a static method and calling it with `->` will hopefully be illegal at some point in the future, no? 😄 Static analyzers do not like this as well.

I hope not, otherwise our `Adapter` class will no longer work.

> I'd prefer it this way round, but I leave this up to you.

It does not seem to work anyway. The `Static method cannot be invoked on mock object" warning keeps popping up. 🤷‍♂️

phpstan.neon.dist
ausi: Can we make sure to only ignore the warning in the unit test and not anywhere else?

m-vo: It's already scoped to the exact file.

ausi: Right, sorry 🙈 _Didn’t see the line after the comment ☺️_

leofeyer: I think that is what the `@phpstan-ignore-next-line` annotation is for. I have already changed that in 72927064aa24f0483c1401be8eef38faa1a417c0.

leofeyer (approved):

phpstan.neon.dist
m-vo: :+1: Totally forgot about that one.

leofeyer:
Thank you @m-vo.

PR #3660, opened 2021-11-10
Add Chosen to select menus in the backend DCA filters

qzminski:
This pull request adds the Chosen plugin to select menus (to both, filter dropdowns and the search dropdown) in the backend DCA filters. Screenshot:

Zrzut ekranu 2021-11-10 o 11 03 43

One thing to consider is that it is possible to hit the "Enter" key and submit the form with the current regular select menus. This is not yet the case with the "Chosen" plugin, but we might want to add that.

Feature proposal per @Toflar request.


fritzmg (dismissed):

Toflar (dismissed):

bytehead (dismissed):

m-vo (dismissed):

aschempp:
can we make this configurable somehow? I don't think chosen makes sense for **all** filters, but it should be selectively enabled for some?

leofeyer:
I think we did not add Chosen to these drop-downs on purpose, because it would affect the page loading time. Similar issues had been reported e. g. in https://github.com/contao/core/issues/8696.

Maybe we can replace Chosen with a modern and more performant solution?

qzminski:
Chosen has performance issues with hundreds of options, but it seems to work fast with my test case of tl_page. @Toflar initially suggested that Chosen should be added only to the dropdowns with more than N items. Perhaps that would be an intermediate solution?

Toflar:
I think replacing chosen and performance are two different topics because even the most performant script will be slow when generating an in browser autocomplete thingy with 5000 entries :) 
So if we want to optimize this we should probably go for something like

- Select all `.tl_chosen`
- Check if there is at least one option, if not return
- If there are options, count them
- If  <= 5 or >= <put in max performance number>, return
- Otherwise do your magic

I'd propose 100 for the max performance number. I think that's a reasonable max to start with. We can still increase if we find out 200 are no problem either. And as for the minimum: I just don't think it's worth to have a chosen field if there are only 5 options.

m-vo:
>  And as for the minimum: I just don't think it's worth to have a chosen field if there are only 5 options.

I don't think making the behavior dependent on the number of items is good UX. Consistency/coherence is what makes a product feel good because you do not have to think about things anymore.

Toflar:
I think we should merge this as is then. The mentioned issue was having 23.844 elements in our old include element. I don't think we can argue that it was chosen's fault in that case. It's just entirely wrong to use a dropdown in the first place if you expect that many entries :) We have solved this in the meantime by using the universal picker.

I don't think we'll ever get to a filter that has that many values. And again, if we do, then it would be the same problem with a regular dropdown and we need to find a different widget type there too.

aschempp (dismissed):

leofeyer:
Ok, let's change it again and see what happens. 👍 However, right now the PR is pretty inconsistent:



And in the file mananger (`DC_Folder`), there is no `tl_chosen` at all. Is this the desired behavior?

qzminski:
I have pushed the latest changes to make it all consistent.

leofeyer (approved):

leofeyer:
Thank you @qzminski.

PR #3664, opened 2021-11-10
Update PHPStan to v1.0

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes -
| Docs PR or issue | -

Failing checks after update + migration are here: [e96dd069a35f372512409f9a3491520fb095d249](https://github.com/contao/contao/pull/3664/commits/e96dd069a35f372512409f9a3491520fb095d249)

### DIFF
--- a/core-bundle/src/Command/ResizeImagesCommand.php
+++ b/core-bundle/src/Command/ResizeImagesCommand.php
@@ -211,7 +211,7 @@ class ResizeImagesCommand extends Command
                 if ($timeLimit && microtime(true) - $startTime + $sleep > $timeLimit) {
                     $this->io->writeln('Time limit of '.$timeLimit.' seconds reached.');
 
-                    return 0;
+                    break;
                 }
 
                 usleep((int) ($sleep * 1000000));
--- a/core-bundle/src/Monolog/ContaoTableProcessor.php
+++ b/core-bundle/src/Monolog/ContaoTableProcessor.php
@@ -46,7 +46,7 @@ class ContaoTableProcessor implements ProcessorInterface
 
         $context = $record['context']['contao'];
         $request = $this->requestStack->getCurrentRequest();
-        $level = $record['level'] ?? 0;
+        $level = $record['level'];
 
         $this->updateAction($context, $level);
         $this->updateBrowser($context, $request);
--- a/core-bundle/src/Routing/PageUrlGenerator.php
+++ b/core-bundle/src/Routing/PageUrlGenerator.php
@@ -38,6 +38,9 @@ class PageUrlGenerator extends SymfonyUrlGenerator
         $this->pageRegistry = $pageRegistry;
     }
 
+    /**
+     * @throws RouteParametersException
+     */
     public function generate(string $name, array $parameters = [], int $referenceType = self::ABSOLUTE_PATH): string
     {
         if (
--- a/core-bundle/src/Routing/PageUrlGenerator.php
+++ b/core-bundle/src/Routing/PageUrlGenerator.php
@@ -38,9 +38,6 @@ class PageUrlGenerator extends SymfonyUrlGenerator
         $this->pageRegistry = $pageRegistry;
     }
 
-    /**
-     * @throws RouteParametersException
-     */
     public function generate(string $name, array $parameters = [], int $referenceType = self::ABSOLUTE_PATH): string
     {
         if (
--- a/core-bundle/tests/Picker/TablePickerProviderTest.php
+++ b/core-bundle/tests/Picker/TablePickerProviderTest.php
@@ -31,12 +31,16 @@ class TablePickerProviderTest extends ContaoTestCase
 {
     protected function setUp(): void
     {
+        parent::setUp();
+
         $GLOBALS['TL_DCA'] = [];
         $GLOBALS['BE_MOD'] = [];
     }
 
     protected function tearDown(): void
     {
+        parent::tearDown();
+
         unset($GLOBALS['TL_DCA'], $GLOBALS['BE_MOD']);
     }
 
--- a/core-bundle/tests/Picker/TablePickerProviderTest.php
+++ b/core-bundle/tests/Picker/TablePickerProviderTest.php
@@ -501,7 +501,7 @@ class TablePickerProviderTest extends ContaoTestCase
             ->expects($this->exactly(\count($expectedItems)))
             ->method('createItem')
             ->withConsecutive(...$expectedItems)
-            ->willReturn($menu ?? $this->createMock(ItemInterface::class))
+            ->willReturn($menu)
         ;
 
         return new TablePickerProvider(
--- a/core-bundle/src/Exception/RouteParametersException.php
+++ b/core-bundle/src/Exception/RouteParametersException.php
@@ -13,9 +13,10 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Exception;
 
 use Symfony\Component\Routing\Exception\ExceptionInterface;
+use Symfony\Component\Routing\Exception\InvalidParameterException;
 use Symfony\Component\Routing\Route;
 
-class RouteParametersException extends \InvalidArgumentException implements ExceptionInterface
+class RouteParametersException extends InvalidParameterException
 {
     private Route $route;
     private array $parameters;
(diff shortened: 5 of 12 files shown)


m-vo:
I'm not 100% sure about 3b8c844862ad3dfb2f0d821b6edb16f6f312c664 but PHPStan's output where it did know all the values was convincing to me :smile:. 

> Offset 'level' on array{message: string, context: array, level: 100|200|250|300|400|500|550|600, level_name: 'ALERT'|'CRITICAL'|'DEBUG'|'EMERGENCY'|'ERROR'|'INFO'|'NOTICE'|'WARNING', channel: string, datetime: DateTimeImmutable, extra: array} on left side of ?? always exists and is not nullable.

core-bundle/src/Command/ResizeImagesCommand.php
```diff
@@ -211,7 +211,7 @@ private function executeInCurrentProcess(float $timeLimit, float $concurrent, bo
                 if ($timeLimit && microtime(true) - $startTime + $sleep > $timeLimit) {
                     $this->io->writeln('Time limit of '.$timeLimit.' seconds reached.');
 
-                    return 0;
+                    break;
```
m-vo, line 214: @ausi Can you double check this is correct? When breaking out of the foreach loop we should now IMHO hit the `return` on L244 if L237 does not evaluate to `true`.

ausi: As the only code after the foreach is the `return 0;` line I’m pretty sure this change has the exact behavior as before. The `finally` block gets executed with both versions.

Why does phpstan want to change this to `break`?

m-vo: > Why does phpstan want to change this to break?

Because the `return 1` from the `finally` block would overwrite this return which is supposedly bad practice (My PHPStorm complains about this as well.). It's at least a bit hard to understand I guess. :upside_down_face: 

The exact message from PHPStan was: `This return is overwritten by a different one in the finally block below.`.

core-bundle/src/Monolog/ContaoTableProcessor.php
```diff
@@ -46,7 +46,7 @@ public function __invoke(array $record): array
 
         $context = $record['context']['contao'];
         $request = $this->requestStack->getCurrentRequest();
-        $level = $record['level'] ?? 0;
+        $level = $record['level'];
```
leofeyer, line 49: Why?

core-bundle/src/Routing/PageUrlGenerator.php
```diff
@@ -38,6 +38,9 @@ public function __construct(RouteProviderInterface $provider, PageRegistry $page
         $this->pageRegistry = $pageRegistry;
     }
 
+    /**
+     * @throws RouteParametersException
```
leofeyer: We are not using `@throws` annotations anywhere. Why does PHPStan complain about this one?

core-bundle/src/Monolog/ContaoTableProcessor.php
bytehead: See here: https://github.com/contao/contao/pull/3664#issuecomment-965826213

core-bundle/tests/Picker/TablePickerProviderTest.php
```diff
@@ -497,7 +501,7 @@ private function createMenuTableProvider(array $modules, string $current, ItemIn
             ->expects($this->exactly(\count($expectedItems)))
             ->method('createItem')
             ->withConsecutive(...$expectedItems)
-            ->willReturn($menu ?? $this->createMock(ItemInterface::class))
+            ->willReturn($menu)
```
leofeyer, line 501: Why? `$menu` can be `null`, so we cannot change this.

core-bundle/src/Command/ResizeImagesCommand.php
ausi: > This return is overwritten by a different one in the finally block below.

That’s the beauty of the `finally` block to me ☺️
But I agree, a `break` here is better for understanding the code correctly.

core-bundle/tests/Picker/TablePickerProviderTest.php
m-vo: It is set a few lines up in the `else` block.

core-bundle/src/Routing/PageUrlGenerator.php
m-vo: I'm not sure why PHPStan cannot figure this out, but IMHO it's correct to annotate exceptions that are thrown as apart of the "business logic", i.e. no guards but custom exceptions you would catch as part of the normal control flow.

leofeyer: What is the exact PHPStan error message?

core-bundle/src/Monolog/ContaoTableProcessor.php
leofeyer: > I'm not 100% sure about [3b8c844](https://github.com/contao/contao/commit/3b8c844862ad3dfb2f0d821b6edb16f6f312c664) but PHPStan's output where it did know all the values was convincing to me 😄.
> 
> > Offset 'level' on array{message: string, context: array, level: 100|200|250|300|400|500|550|600, level_name: 'ALERT'|'CRITICAL'|'DEBUG'|'EMERGENCY'|'ERROR'|'INFO'|'NOTICE'|'WARNING', channel: string, datetime: DateTimeImmutable, extra: array} on left side of ?? always exists and is not nullable.

How can we be sure? This is our own class and there is no guarantee that `$record` contains a `level` key. After all, the `?? 0` is probably there for a reason?

core-bundle/src/Routing/PageUrlGenerator.php
m-vo: ~~See https://github.com/contao/contao/pull/3664#issuecomment-965826213~~ or click the link in the initial description.

My bad, confused or with the other one.

m-vo: > Dead catch - Contao\CoreBundle\Exception\RouteParametersException is never thrown in the try block.

m-vo: Seems that rethrowing as another type isn't detected.

leofeyer: Then it is obviously not a Contao issue. We should probably report it and ignore it in the PHPStan config file until it has been fixed.

m-vo: I spent some time debugging it. It is indeed no bug in PHPStan. The reason the error is reported is because the `Symfony\Component\Routing\Generator\UrlGeneratorInterface` annotates the following exceptions that may be thrown (regular control flow):

https://github.com/symfony/symfony/blob/b560ec4a1d11f7b7e78a4bb9ed75b84397222a26/src/Symfony/Component/Routing/Generator/UrlGeneratorInterface.php#L80-L82

Our own `RouteParametersException` is of course none of them. IMHO this is a bug because now we're not following the interface anymore. People implementing it are only supposed to handle the annotated exceptions.

I made the `RouteParametersException` extend from `InvalidParameterException` in 6158ad1115f9eb85af47cce69671e1ce7a772a20. This way our `PrettyErrorScreenListener` still can access the the additional route information but we're compliant again. 
@aschempp Is this the right exception to extend?

core-bundle/src/Monolog/ContaoTableProcessor.php
m-vo: > How can we be sure? 

The array structure comes from monolog and they are annotating it; see https://github.com/Seldaek/monolog/blob/f19a2ae873d42f4494f3fb29de9dadb91a3f7601/src/Monolog/Logger.php#L30-L32.

----

Basically the `ProcessorInterface` says it is a `Record`.

The `Logger` defines a `Record` looks like this:
```
array{message: string, context: mixed[], level: Level, level_name: LevelName, channel: string, datetime: \DateTimeImmutable, extra: mixed[]}
```

And also that a `Level` looks like this:
```
Logger::DEBUG|Logger::INFO|Logger::NOTICE|Logger::WARNING|Logger::ERROR|Logger::CRITICAL|Logger::ALERT|Logger::EMERGENCY
```

These constants evaluate to `100|200|250|300|400|500`.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3665, opened 2021-11-11
Handle doctrine binary types in version compare view

SeverinGloeckle:
The `Versions::compare` method tries to detect binary DCA fields to transform their values to a readable UUID. However, this detection does not work if you declare your DCA field using the built-in types from `Doctrine\DBAL\Types\Types`:

Not detected as binary field:

````php
'sql' => ['type' => Doctrine\DBAL\Types\Types::BINARY, 'length' => 16, 'fixed' => true, 'notnull' => false]
````

Detected as binary field:

````php
'sql' => 'binary(16) NULL'
````

This pull request adds Doctrine's two built-in binary types to the detection.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -13,6 +13,7 @@ namespace Contao;
 use Contao\CoreBundle\Exception\ResponseException;
 use Doctrine\DBAL\Types\BinaryType;
 use Doctrine\DBAL\Types\BlobType;
+use Doctrine\DBAL\Types\Types;
 
 /**
  * Provide methods to handle versioning.
@@ -458,7 +459,7 @@ class Versions extends Controller
 
 						if (\is_array($arrFields[$k]))
 						{
-							$blnIsBinary = $arrFields[$k]['type'] === BinaryType::class || $arrFields[$k]['type'] === BlobType::class;
+							$blnIsBinary = \in_array($arrFields[$k]['type'] ?? null, array(BinaryType::class, BlobType::class, Types::BINARY, Types::BLOB), true);
 						}
 						else
 						{
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Contao\CoreBundle\Doctrine\DBAL\Types\BinaryStringType;
 use Contao\CoreBundle\Exception\ResponseException;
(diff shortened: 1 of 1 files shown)


SeverinGloeckle:
The existing comparison in `Versions::compare` is based on Doctrine's type classes (e.g. `\Doctrine\DBAL\Types\BinaryType`). 

The following declaration is currently handled correctly in `Versions::compare`, however it's an unkown column type when running a migration for the field:

````php
'sql' => ['type' => \Doctrine\DBAL\Types\BinaryType::class, 'length' => 16, 'fixed' => true, 'notnull' => false]
````

I'm not sure what's the use case with the type classes in `Versions::compare`, so I decided to extend the existing checks instead of replacing them.

aschempp:
Did you see we have a special Doctrine type for something similar to this? Maybe this solves your problem.
https://github.com/contao/contao/blob/4.x/core-bundle/src/Doctrine/DBAL/Types/BinaryStringType.php

SeverinGloeckle:
> Did you see we have a special Doctrine type for something similar to this? Maybe this solves your problem. https://github.com/contao/contao/blob/4.x/core-bundle/src/Doctrine/DBAL/Types/BinaryStringType.php

Thanks for the hint. 

This is helpful and works for defining the binary type for the migration – however, the version compare view doesn't detect `BinaryStringType` as a binary field either.

I'd argue its name should be added to the check as well:

````php
\in_array(..., array(BinaryType::class, BlobType::class, Types::BINARY, Types::BLOB, BinaryStringType::NAME), true);
````

Maybe you can also share some insights on the existing `BinaryType::class` and `BlobType::class` checks?

aschempp:
I don't really know about the existing checks. The custom type was added because Doctrine handles binary fields as streams by default, which didn't work with our serialized binary strings.

SeverinGloeckle:
I have added `BinaryStringType::NAME` to the check as well.

All of these are now handled as binary fields in the version compare view:

````php
'sql' => 'binary(16) NULL'
'sql' => 'blob NULL'
'sql' => ['type' => Doctrine\DBAL\Types\Types::BINARY, ...]
'sql' => ['type' => Doctrine\DBAL\Types\Types::BLOB, ...]
'sql' => ['type' => Contao\CoreBundle\Doctrine\DBAL\Types\BinaryStringType::NAME, ...]
````

If anyone has information about the two checks for `BinaryType::class` and `BlobType::class`, I would be happy to add some comments to the checks. 

Otherwise, this may be merged from my side.

ausi (approved):

fritzmg (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3668, opened 2021-11-11
Do not automatically focus input fields with autocomplete

### LINKED ISSUE #3651: Unsightly effect of autoFocusFirstInputField
Contao 4.9

Every time I open a user for editing I get this annoying popup as can be seen in the screenshot.

Screen-2021-11-06_16-46-37-2

I think it's the result of `autocomplete = "username"` and at the same time the JS function `autoFocusFirstInputField()`.

IMO we should exclude fields with the attribute `autocomplete` from the function `autoFocusFirstInputField()` in this line:
https://github.com/contao/contao/blob/8468881d30b54c448310bc2456920d4ec328e088/core-bundle/src/Resources/public/core.js#L2597.

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3651

The autofocus filter now skips inputs with autocomplete (besides `autocomplete="off"`). This prevents issues when autofocusing inputs with autocomplete plugins as described in #3651.

It is possible that the focused input is no longer the first visible input field (e.g. when editing a user account, the second input "Name" receives the focus), which might confuse some editors. However, I think the advantage of not having to deal with autocomplete plugins outweighs this.


aschempp:
Interesting issue. I can imagine there are two problems
 1. if you **add a new record**, focussing the first input field _with_ autocomplete seems totally fine to me, since I need to enter a username
 2. if I'm **editing a record**, I'm not sure the autofocus is useful at all. It probably is to "tab" through the form, but I'm unsure that's the reason we actually have that feature?

fritzmg:
As mentioned in #3651 this should be fixed in the browser, in my opinion.

SeverinGloeckle:
> 1. if you **add a new record**, focussing the first input field _with_ autocomplete seems totally fine to me, since I need to enter a username
> 2. if I'm **editing a record**, I'm not sure the autofocus is useful at all. It probably is to "tab" through the form, but I'm unsure that's the reason we actually have that feature?

The autofocus is useful both when creating new records and when editing existing data since it speeds up your keyboard workflow.

However, you can argue that autofocusing an autocomplete input (and thus possibly triggering any autocomplete plugin) is fine when creating a new record.

We could skip autocomplete inputs only if they already contain a value, then:

````javascript
... (!item.get('autocomplete') || item.get('autocomplete') === 'off' || !item.get('value'))
````

This is a useful improvement and I have added the additional check.

> As mentioned in #3651 this should be fixed in the browser, in my opinion.

Valid point. We could just blame the browser and/or the autocomplete plugin for the behaviour. I'm not sure whether showing autocomplete options is wrong/unintended when autofocusing an input, though.

As far as I can see, autocomplete is only used in `tl_member` and `tl_user` and only in `tl_user` it's the first input that uses autocompletion.

So both ignoring the issue or adding the simple fix to move the autofocus to the next widget seem valid to me.

ausi:
> So both ignoring the issue or adding the simple fix to move the autofocus to the next widget seem valid to me.

Agree, I’m fine either way too.

> 2\. if I'm **editing a record**, I'm not sure the autofocus is useful at all.

If we can detect this easily, I think this would be the best solution to the problem. 
Only use autofocus for creating new records 👍

aschempp (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3672, opened 2021-11-12
Remove a lot of inline phpDoc comments now that all tools support generics

leofeyer:
Isn't the code much nicer this way? 😊

### DIFF
--- a/core-bundle/tests/Crawl/Escargot/Subscriber/SearchIndexSubscriberTest.php
+++ b/core-bundle/tests/Crawl/Escargot/Subscriber/SearchIndexSubscriberTest.php
@@ -17,6 +17,7 @@ use Contao\CoreBundle\Crawl\Escargot\Subscriber\SubscriberResult;
 use Contao\CoreBundle\Search\Indexer\IndexerException;
 use Contao\CoreBundle\Search\Indexer\IndexerInterface;
 use Nyholm\Psr7\Uri;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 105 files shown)


bytehead (dismissed):

core-bundle/tests/Crawl/Escargot/Subscriber/SearchIndexSubscriberTest.php
```diff
@@ -498,10 +497,7 @@ static function (string $key) use ($statusCode) {
         return $response;
     }
 
-    /**
-     * @return TranslatorInterface&MockObject
-     */
-    private function getTranslator(): TranslatorInterface
+    private function getTranslator()
```
ausi: Is it necessary to remove the `TranslatorInterface` return type so that the static analysis tools can infer `TranslatorInterface&MockObject` as the correct type?

m-vo: > Is it necessary to remove the TranslatorInterface return type so that the static analysis tools can infer TranslatorInterface&MockObject as the correct type?

If so I'd vote to keep the annotation instead.

Maybe this change would also make it harder to switch to real PHP8.1 intersection types at some point in a semi-automated way.

leofeyer: I have added the return types to the mock methods again in c1307c6ab1b38e929e45b95fffa1bcf594ba86b0. I have also fixed the methods which did not follow the naming rule "methods that return a mock object must begin with `mock`".

ausi (dismissed):

PR #3674, opened 2021-11-14
Add a DCA option to collapse inactive checkbox groups

### LINKED ISSUE #3629: Benutzergruppen Erlaubte Felder
Kann man in der Legend Erlaubten Feldern als Standart alle zu klappen, außer diese die konfiguriert sind?
Damit hätte man eine bessere Übersicht.

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3629
| Docs PR or issue | https://github.com/contao/docs/pull/888

If a checkbox widget contains multiple groups of options, it would be useful if the widget could be initialized with most of the groups collapsed (see #3629).

To selectively add this feature to some checkbox widgets, this draft adds a new DCA option `fields.eval.checkboxGroup` which itself can take up to three boolean options:
- `collapseInactive`: Collapse groups with no selected values
- `fallbackToFirst`: Open the first group of the widget if there are no selected values
- `overwriteSession`: Allow the other two options to change a previous manual user interaction with the widget's groups

The options can be combined in any way.

For the extensive access rights checkbox widgets in `tl_user` and `tl_user_group`, I'd suggest enabling all three options.

This way, only the first group of options will be visible when opening the edit view for the first time, which reduces the list of options while still presenting the editor an idea of the options available (instead of closing all groups without `fallbackToFirst: true`).

Whether the session should be overwritten is debatable. To me, it seems appropriate for the extensive option lists in `tl_user` and `tl_user_group` and ensures a consistent view for the editor (always showing groups with selected options, if there are any).

Initial view:

> Bildschirmfoto 2021-11-14 um 20 56 13

View with selected options:

> Bildschirmfoto 2021-11-14 um 20 56 30

### Questions & todos:
- Is a nested option really necessary or is a simple flag enough (e.g. `collapseInactiveGroups`, grouping all three options together)?
- Are there any more checkbox widgets that should get initialized this way?
- Extend the DCA eval documentation.


leofeyer:
> * Is a nested option really necessary or is a simple flag enough (e.g. `collapseInactiveGroups`, grouping all three options together)?

IMHO, a simple flag would suffice.

aschempp:
what is an "inactive group"?

SeverinGloeckle:
It's a group of checkbox options where no option is selected (see screenshots in the first post). 

We should probably make the context more clear by naming it `collapseInactiveCheckboxGroups` if it's a simple flag.

fritzmg:
Or may be `collapseUncheckedGroups`?

SeverinGloeckle:
> Or may be `collapseUncheckedGroups`?

I like `unchecked` over `inactive`!

I'd suggest `collapseUncheckedOptionGroups` to add context to the groups part again.

Toflar:
Just a random thought: Wouldn't that be a good default behaviour? I mean, if we can save space we should by default 🤷‍♂️ Maybe there's not even the need for a setting?

fritzmg:
> Just a random thought: Wouldn't that be a good default behaviour?

Not for small groups or a small amount of groups as it would require an additional click every time.

SeverinGloeckle:
I have merged the three options into a single option `eval.collapseUncheckedOptionGroups`.

Toflar (dismissed):
I like that! It's a simple and useful addition 👍

ausi (dismissed):

aschempp (dismissed):
Simple and good! 

I'd rename `collapseUncheckedOptionGroups` to `collapseUncheckedGroups` since there is nothing else but "options" anyway. Following our DCA style it could even be just `collapse` and people will "get to know" what it does 😂

ausi:
+1 for the naming `collapseUncheckedGroups`

SeverinGloeckle:
I have renamed the option from `collapseUncheckedOptionGroups` to `collapseUncheckedGroups`.

Toflar (approved):

ausi (approved):

leofeyer:
Thank you @SeverinGloeckle.

PR #3676, opened 2021-11-15
Replace FragmentRendererPass with tagged locator

aschempp:
The `FragmentRendererPass` arguments have become deprecated in https://github.com/symfony/symfony/pull/40468. Luckily, we don't actually need to use the compiler pass to get a list of services 😅

### DIFF
--- a/core-bundle/src/ContaoCoreBundle.php
+++ b/core-bundle/src/ContaoCoreBundle.php
@@ -107,7 +107,6 @@ class ContaoCoreBundle extends Bundle
             )
         );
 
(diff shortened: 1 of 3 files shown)


fritzmg (approved):

bytehead (approved):

m-vo (approved):

leofeyer:
We should also delete the `FragmentRendererPass` class and test class (if exists), shouldn't we?

m-vo:
> We should also delete the FragmentRendererPass class

It's not ours :smile:

leofeyer:
True. 😄

PR #3680, opened 2021-11-16
Fix a wrong method usage

leofeyer:
Follow-up on #3659.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -257,6 +257,11 @@ abstract class Module extends Frontend
 	 */
 	protected function getResponseCacheTags(): array
 	{
+		if (!$this->objModel instanceof Model)
+		{
+			return array();
+		}
+
 		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelInstance($this->objModel));
 	}
 
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -257,7 +257,7 @@ abstract class Module extends Frontend
 	 */
 	protected function getResponseCacheTags(): array
 	{
-		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelClass($this->objModel));
+		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelInstance($this->objModel));
 	}
 
 	/**
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -99,7 +99,7 @@ abstract class Module extends Frontend
 
 	/**
 	 * Model
-	 * @var ModuleModel
+	 * @var ?ModuleModel
 	 */
 	protected $objModel;
 
--- a/core-bundle/src/Resources/contao/modules/Module.php
+++ b/core-bundle/src/Resources/contao/modules/Module.php
@@ -99,7 +99,7 @@ abstract class Module extends Frontend
 
 	/**
 	 * Model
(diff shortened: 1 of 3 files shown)


core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -257,7 +257,7 @@ abstract protected function compile();
 	 */
 	protected function getResponseCacheTags(): array
 	{
-		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelClass($this->objModel));
+		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelInstance($this->objModel));
```
ausi, line 265: Should we also check that `$this->objModel` is not null?

m-vo (approved):

core-bundle/src/Resources/contao/modules/Module.php
leofeyer: We probably should. 👍

leofeyer: Changed in b7ca2d1d0bb00eb434a5321153cdec05eec9f120.

m-vo: How can this be null? The property is annotated with `@var ModuleModel` as well. On the other hand: the constructor type hints are wrong as well.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -257,7 +257,12 @@ abstract protected function compile();
 	 */
 	protected function getResponseCacheTags(): array
 	{
-		return array(System::getContainer()->get(EntityCacheTags::class)->getTagForModelClass($this->objModel));
+		if (!$this->objModel instanceof Model)
```
m-vo: Why don't we check for `null` here? 
We should probably also adjust the typehint in L104 and return type of `getModel` to `?ModuleModel`, no?

leofeyer: Because the `getTagForModelInstance()` method expects a `Model` object.

m-vo: Yes sure, but how can `$this->objModel` be anything else than a `Model` or `null` (if not defined). It gets set in the constructor where the type is checked.

leofeyer: Changed in 60f279d3d835d53335127c8ac0e06be80ffc4d8a.

m-vo:
I just realized, that `ModuleArticle` will probably have to get this change as well…

leofeyer:
That's out of scope of this PR, though.

m-vo:
> That's out of scope of this PR, though.

IMO it's related as it was also introduced in #3659. See https://github.com/leofeyer/contao/pull/1

ausi (approved):

leofeyer:
True. Added in 1cbd271b0396e2c76197cf4a9d6ae5be58277cbe.

leofeyer:
This actually applies to even more classes such as e.g. the `Hybrid` class. And there are quite a few classes that use `$this->objModel` without checking `null` such as e.g. the `ContentElement` class. Therefore, I guess we can safely assume that `Module::$objModel` is never supposed to be `null`.

So IMHO we should add an exception in line 139 instead:

```php
if (!$objModule instanceof Model)
{
    throw new \LogicException('No model given');
}
```

fritzmg (approved):

m-vo (approved):

aschempp:
This change will prevent a content element etc. from being created from a database result, which was the behaviour we had before models. That's why we check for a model in the first if-statement. Is that intentional, maybe it should only be done in Contao 5?

leofeyer:
You are absolutely right. Fixed in #3735.

PR #3682, opened 2021-11-16
Fix type error when trying to copy website root

fritzmg:
Currently you cannot copy a website root in Contao `4.12.3` and `4.12.x-dev` due to the following error:

```
TypeError:

Return value of Contao\CoreBundle\Util\LocaleUtil::formatAsLocale() must be of the type string, null returned
```

This error can also happen if an extension, like `terminal42/contao-changelanguage`, executes `PageModel::findWithDetails(…)` with the ID of the newly created record - since the record will have no language set yet at that point.

This PR will fix both instances.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/LegacyRoutingListener.php
+++ b/core-bundle/src/EventListener/DataContainer/LegacyRoutingListener.php
@@ -84,7 +84,7 @@ class LegacyRoutingListener
      */
     public function overrideUrlPrefix($value, DataContainer $dc): ?string
     {
-        return $this->prependLocale && $dc->activeRecord->language ? LocaleUtil::formatAsLanguageTag($dc->activeRecord->language) : '';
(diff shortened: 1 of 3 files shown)


fritzmg:
@ausi or should `LocaleUtil::formatAsLocale` be able to handle empty (or unknown) language tags instead?

e.g.

```php
return array_pop($locales) ?? $languageTag;
```

or

```php
return array_pop($locales) ?? '';
```

ausi:
I’m +1 for `array_pop($locales) ?? ''` but the `LocaleUtil` was created by @aschempp I think.

fritzmg:
Right, sorry 🙈 I have changed the fix to `LocaleUtil` now, as I would prefer that as well, for ease of use within Contao.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #763, opened 2019-09-16
Use classnames instead of service IDs

leofeyer:
This PR replaces the `contao.` service IDs with class names as recommended in Symfony 4.

### TODO

- [ ] Also convert the private services
- [ ] Add aliases for the private services
- [ ] Check for `->getDefinition()->getClass()`, which now returns `null`

PR #3683, opened 2021-11-16
Use classnames instead of service IDs

leofeyer:
This is a new take on #763. Before I do all the work, I would like to know if you agree with the concept.

Basically, I am planning on having a nice, forward-compatible `services.yml` file that imports a `deprecated.yml` file with all the old service IDs. In Contao 5, all we have to do is to remove the `deprecated.yml` files.

@contao/developers WDYT?


ausi:
LGTM!

fritzmg:
👍 from me. Though do we really want to remove these service IDs in Contao 5.0 already? Personally I am still in favour, just wondering.

m-vo:
Just to be sure, we're aware of this? https://symfony.com/doc/current/bundles/best_practices.html#services

image

Because we're half bundle half application framework? And we don't care about inconsistencies e.g. when there is a need for a class to be used twice as two different services?

leofeyer:
I was aware that we are not supposed to use autowiring in a bundle. But that we cannot use FQCNs is new information.

leofeyer:
Closed in favor of #3686.

PR #2788, opened 2021-02-22
Fix accessing Model\Collection instead of Model in ModuleFaqPage

m-vo:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2786 
| Docs PR or issue | -

`$figureBuilder->setMetadata($objFaq->getOverwriteMetadata())` was failing because `$objFaq` is a `Model\Collection` instead of a `Model`. To prevent this rather hard to spot error in the future, I adjusted how we're iterating over the results instead.

PR #3684, opened 2021-11-16
Use the metadata for the player caption

### LINKED ISSUE #2873: Get Subtititle (Untertitel) for Video/Auto Element from the File Metadata in the Filemanager
The Video/Audio content element has its subtitle in the element's settings. It would be an improvement if the subtitle is displayed from the file metadata. Like the metadata of an image is processed.
It would be more coherent in the users workflow to edit all subtitles for media elements in the file manager.

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2873
| Docs PR or issue | -

With this PR the Video/Audio player will automatically use the caption from the file manager, if present. If multiple sources are used, the first non-empty caption will be used. The caption within the player element will overwrite the caption from the file manager. This PR also fixes an issue where the `title` in the front end could be empty, instead of using the file name as the title.


Toflar (dismissed):

core-bundle/src/Resources/contao/elements/ContentMedia.php
```diff
@@ -130,22 +130,27 @@ protected function compile()
...
 		$strLanguage = LocaleUtil::formatAsLocale($objPage->language);
 
+		// Determine caption
+		$strCaption = $this->playerCaption;
+
 		// Pass File objects to the template
 		while ($objFiles->next())
 		{
-			$arrMeta = StringUtil::deserialize($objFiles->meta);
+			$objMeta = $objFiles->current()->getMetadata($strLanguage);
```
m-vo: I'd prefer if we would rewrite L137 to a `foreach($objFiles as $fileModel)` instead of using `current()`. This is IMHO less likely to introduce subtle bugs.

m-vo (dismissed):
LGTM, only a small nitpick.

core-bundle/src/Resources/contao/elements/ContentMedia.php
fritzmg: Agreed, but I wanted to keep the PR minimal. This should be done in a separate PR and replace the `while` with `foreach` in all other places as well.

m-vo: I'm not sure it's worth refactoring everywhere in the legacy code. But it's probably easy enough at the places where new changes are introduced? But up to you.

For reference in https://github.com/contao/contao/pull/2788/files we fixed the bug I introduced earlier on that could have been avoided if I had done the minimal refactoring in the first place. :smile:

core-bundle/src/Resources/contao/elements/ContentMedia.php
```diff
@@ -130,22 +130,27 @@ protected function compile()
...
-			{
-				$strTitle = $arrMeta[$strLanguage]['title'];
-			}
-			else
+			if (null !== $objMeta)
 			{
-				$strTitle = $objFiles->name;
+				$strTitle = $objMeta->getTitle();
+
+				if (empty($strCaption))
```
leofeyer, line 143: This check should be first in the loop, so we don't call `getMetadata()` etc. if we already know that `$strCaption` will not be overwritten anyway.

leofeyer (changes requested):

core-bundle/src/Resources/contao/elements/ContentMedia.php
leofeyer: I'm with @m-vo on this.

core-bundle/src/Resources/contao/elements/ContentMedia.php
fritzmg: That won't work, because we always need the title.

core-bundle/src/Resources/contao/elements/ContentMedia.php
fritzmg: Done :)

core-bundle/src/Resources/contao/elements/ContentMedia.php
leofeyer: I see, however, I still don't understand the logic.

> If multiple sources are used, the first non-empty caption will be used.

How can this be correct? It seems pretty random to me.

> The caption within the player element will overwrite the caption from the file manager.

IIRC, the caption of the file manager will overwrite the caption set in the player element, won't it?

fritzmg: > How can this be correct? It seems pretty random to me.

It is essentially random as you cannot sort the sources in the back end. But since there can only be one caption we can only extract the caption from one file - which might as well just be whichever source happens to be first.

> IIRC, the caption of the file manager will overwrite the caption set in the player element, won't it?

No, not in the current implementation. If a caption is set in the player element, `if (empty($strCaption))` will be `false` and thus no caption is used from any source.

leofeyer: Exactly. 😄

fritzmg: ? The caption of the player element will overwrite the caption of the file manager, not the other way around is what I am saying ;)

leofeyer (commented):
I find it strange to randomly use the first non-empty title from the files collection. @contao/developers WDYT?

Toflar:
I guess you'd want an explicit "Fallback" checkbox?

fritzmg:
> I find it strange to randomly use the first non-empty title from the files collection. @contao/developers WDYT?

You mean caption. The title is used for _every_ source (since that goes into the `title` attribute of the `<source>`).

leofeyer:
> I guess you'd want an explicit "Fallback" checkbox?

Yes, something like that. Or maybe the random caption is OK because we can assume that all media files represent the same media only in different resolutions?

fritzmg:
> Or maybe the random caption is OK because we can assume that all media files represent the same media only in different resolutions?

That's how I see it. It doesn't matter in which of the sources you write the caption. As long as it is defined in one of them, it will automatically be displayed in the player.

m-vo (dismissed):

core-bundle/src/Resources/contao/elements/ContentMedia.php
```diff
@@ -125,27 +125,31 @@ protected function compile()
...
-
 		// Convert the language to a locale (see #5678)
 		$strLanguage = LocaleUtil::formatAsLocale($objPage->language);
 
+		// Determine caption
+		$strCaption = $this->playerCaption;
+
 		// Pass File objects to the template
-		while ($objFiles->next())
+		foreach ($objFiles as $objFile)
```
ausi: Using `$objFile` for both the model and the file object on line 151 can lead to confusion I think.
How about naming it `$objFileModel`?
```suggestion
		foreach ($objFiles as $objFileModel)
```

fritzmg: Done in db6d12026

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3686, opened 2021-11-17
Do not use FQCN service IDs for non-autowiring services

leofeyer:
Follow-up on #3683.

As we have learned from the [Symfony docs](https://symfony.com/doc/current/bundles/best_practices.html#services), we must not use FQCNs as service IDs in our bundles (except maybe for the manager-bundle, because it defines the application). So this is how it would look in the future.


bytehead:
I don't get the move from `.yaml` to `.php`. What's the idea behind?

m-vo:
This
```yaml
contao.cron:
    class: Contao\CoreBundle\Cron\Cron

# Backwards compatibility
Contao\CoreBundle\Cron\Cron: '@contao.cron'
```
would be equivalent to
```yaml
Contao\CoreBundle\Cron\Cron:
    alias: 'contao.cron'
```
?

Or can we use `deprecated` in the first case to just deprecate the usage via the FQCN?

leofeyer:
Yes, the two are equivalent AFAIK.

leofeyer:
You can e.g. see this in the `services.yml` file of the manager bundle:

```yml
services:
    # Backwards compatibility
    Contao\ManagerBundle\Controller\DebugController:
        alias: contao_manager.controller.debug
        public: true
        deprecated:
            package: contao/manager-bundle
            version: 4.13
            message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "contao_manager.controller.debug" instead.
```

m-vo:
> You can e.g. see this in the services.yml file of the manager bundle:

But that's the other way round. Deprecating the prefixed name instead of the FQCN.

leofeyer:
This one is ready for review now. 🎉 

@contao/developers Which services need to be available for autowiring? We currently have these:

```yml
# Autowiring aliases
Contao\CoreBundle\Cache\EntityCacheTags: '@contao.cache.entity_cache_tags'
Contao\CoreBundle\Config\ResourceFinderInterface: '@contao.resource_finder'
Contao\CoreBundle\Framework\ContaoFramework: '@contao.framework'
Contao\CoreBundle\Image\ImageFactoryInterface: '@contao.image.image_factory'
Contao\CoreBundle\Image\PictureFactoryInterface: '@contao.image.picture_factory'
Contao\CoreBundle\Image\Studio\Studio: '@contao.image.studio'
Contao\CoreBundle\InsertTag\InsertTagParser: '@contao.insert_tag_parser'
Contao\CoreBundle\Picker\PickerBuilderInterface: '@contao.picker.builder'
Contao\CoreBundle\Routing\ScopeMatcher: '@contao.routing.scope_matcher'
Contao\CoreBundle\Security\Authentication\Token\TokenChecker: '@contao.security.token_checker'
Contao\CoreBundle\Security\TwoFactor\Authenticator: '@contao.security.two_factor.authenticator'
Contao\CoreBundle\Security\TwoFactor\TrustedDeviceManager: '@contao.security.two_factor.trusted_device_manager'
Contao\CoreBundle\Slug\Slug: '@contao.slug'
Contao\CoreBundle\String\SimpleTokenParser: '@contao.string.simple_token_parser'
```

Other candidates would be:

- `contao.intl.countries`
- `contao.intl.locales`
- `contao.response_context.accessor`
- `contao.string.html_decoder`

WDYT?

leofeyer:
Oh, and do we really need autowiring aliases for `contao.security.two_factor.authenticator` and `contao.security.two_factor.trusted_device_manager`? @bytehead /cc

bytehead:
> Oh, and do we really need autowiring aliases for contao.security.two_factor.authenticator and contao.security.two_factor.trusted_device_manager?

I don't need any 🤷‍♂️  I'll alias them in my `service.yaml`, if I need them aliased in my application. 🙃

ausi:
`contao.intl.countries`, `contao.intl.locales` and `contao.string.html_decoder` should get an alias too IMO.

fritzmg:
> @contao/developers Which services need to be available for autowiring? We currently have these:

`contao.opt-in` should get one.

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -527,7 +527,7 @@ private function doBuild(): Figure
         $settings = clone $this;
 
         $imageResult = $this->locator
-            ->get(Studio::class)
+            ->get('contao.image.studio')
```
aschempp, line 530: `Studio::class` should get an autowiring alias and this change can then be reverted.

aschempp (commented):
Any particular reason you removed all the `listeners.yml` and gave them bundle service names? But we keep the controllers as FQCN (which I like!)? As far as I remember, we defined event listeners and controllers (and maybe commands) to be "our application" and e.g. use autoconfig (and maybe FQCN)?

leofeyer:
> Any particular reason you removed all the `listeners.yml` and gave them bundle service names? But we keep the controllers as FQCN (which I like!)?

Controllers only work with FQCNs (see https://www.py4u.net/discuss/1851134), therefore I could not get rid of the FQCNs in this case.

> As far as I remember, we defined event listeners and controllers (and maybe commands) to be "our application" and e.g. use autoconfig (and maybe FQCN)?

Autoconfig yes, FQCNs no, autowiring no.

leofeyer:
I have added the requested autowiring aliases in df41964d689a82e7716402f4583251ddafe59757.

aschempp:
https://symfony.com/doc/current/bundles/best_practices.html#services 

> If the bundle defines services, they must be prefixed with the bundle alias instead of using fully qualified class names like you do in your project services. For example, AcmeBlogBundle services must be prefixed with acme_blog. The reason is that bundles shouldn't rely on features such as service autowiring **or autoconfiguration** to not impose an overhead when compiling application services.


According to the docs, we shouldn't use autoconfiguration either (which I disagree with as well 🤷 )

leofeyer:
<del>We all disagreed with this, that's why we are allowing autoconfiguration. But I can try to remove it as well if you want.</del>

**Edit:** As discussed in the Contao call, we want to keep using autoconfiguration.

bytehead:
Somehow the functional tests are failing after removing the `TwoFactor` autowiring aliases.

leofeyer:
Let's revert then. 😄

core-bundle/src/Image/Studio/Studio.php
```diff
@@ -67,13 +68,13 @@ public function createLightboxImage($filePathOrImage, string $url = null, $sizeC
...
         return [
-            self::class,
+            'contao.image.studio' => self::class,
             'contao.image.picture_factory' => PictureFactoryInterface::class,
             'contao.image.image_factory' => ImageFactoryInterface::class,
             'contao.image.resizer' => ResizerInterface::class,
             'contao.assets.files_context' => ContaoContext::class,
             'contao.framework' => ContaoFramework::class,
-            'event_dispatcher' => 'event_dispatcher',
+            'event_dispatcher' => EventDispatcherInterface::class,
```
ausi: Using the class names would again make use of autowiring, right?
So we would have to use only the service names I think:
```suggestion
            'contao.image.studio',
            'contao.image.picture_factory',
            'contao.image.image_factory',
            'contao.image.resizer',
            'contao.assets.files_context',
            'contao.framework',
            'event_dispatcher',
```

ausi:
> Somehow the functional tests are failing after removing the `TwoFactor` autowiring aliases.

I think this is because we are using FQCNs in the `getSubscribedServices()` methods, and therefore we are using autowiring with these services. See https://github.com/contao/contao/pull/3686#discussion_r752555567

core-bundle/src/Image/Studio/Studio.php
m-vo: > Using the class names would again make use of autowiring, right?

I'd say so, too. The key is only used for the internal reference - you could for instance set `'foo' => 'event_dispatcher'` and then use it like `$container->get('foo')`. Writing `<numeric-key> => <value>` is equivalent to `<value> => <value>`.

leofeyer: If I change this, however, I get the following error:

```
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: "Contao\CoreBundle\Image\Studio\Studio::getSubscribedServices()" must return valid PHP types for service "contao.image.studio" key "0", "contao.image.studio" returned.
```

leofeyer: It would work like this, however, I am not sure if this is better:

```php
return [
    'contao.image.studio' => 'contao.image.studio',
    'contao.image.picture_factory' => 'contao.image.picture_factory',
    'contao.image.image_factory' => 'contao.image.image_factory',
    'contao.image.resizer' => 'contao.image.resizer',
    'contao.assets.files_context' => 'contao.assets.files_context',
    'contao.framework' => 'contao.framework',
    'event_dispatcher' => 'event_dispatcher',
];
```

m-vo: Ah maybe the shortcut is only working with class names. It sure is better if we're not using the FQCN version, so that you're able to decorate the service.

ausi: Fixed in 58a480e8dac91ed3a7cb93bea6eeaa4206d7b8f3

leofeyer: Changed in 58a480e8dac91ed3a7cb93bea6eeaa4206d7b8f3.

ausi (approved):
❤️

leofeyer:
Thank you @ausi for the `!service_locator` hint. I wouldn't have thought of that. 🦊

PR #3692, opened 2021-11-17
Remove the script name before signing a redirect URI

### LINKED ISSUE #3689: Backend login does not redirect back on preview links
**Affected version(s)**

Contao 4.9

**Description**

Open a preview link without a valid backend session, you'll get redirected to the backend login but not redirected after the successful login:

https://demo.contao.org/preview.php/en/content-elements.html

bytehead:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3689
| Docs PR or issue | n/a

Remove any script name (normally `/preview.php`) from the generated URI before signing, otherwise the signed request is invalid after the redirect.

### DIFF
--- a/core-bundle/src/EventListener/PreviewAuthenticationListener.php
+++ b/core-bundle/src/EventListener/PreviewAuthenticationListener.php
@@ -65,6 +65,8 @@ class PreviewAuthenticationListener
         }
 
         $context = $this->router->getContext();
+        $baseUrl = $context->getBaseUrl();
+
         $context->setBaseUrl('');
 
         $url = $this->router->generate(
@@ -72,6 +74,8 @@ class PreviewAuthenticationListener
             ['redirect' => $request->getUri()],
             UrlGeneratorInterface::ABSOLUTE_URL
         );
+
+        $context->setBaseUrl($baseUrl);
 
         $event->setResponse(new RedirectResponse($this->uriSigner->sign($url)));
     }
--- a/core-bundle/src/EventListener/PreviewAuthenticationListener.php
+++ b/core-bundle/src/EventListener/PreviewAuthenticationListener.php
@@ -70,6 +70,8 @@ class PreviewAuthenticationListener
             UrlGeneratorInterface::ABSOLUTE_URL
         );
 
+        $url = str_replace($request->getScriptName(), '', $url);
+
         $event->setResponse(new RedirectResponse($this->uriSigner->sign($url)));
     }
 }
--- a/core-bundle/src/EventListener/PreviewAuthenticationListener.php
+++ b/core-bundle/src/EventListener/PreviewAuthenticationListener.php
@@ -64,13 +64,14 @@ class PreviewAuthenticationListener
             return;
         }
 
+        $context = $this->router->getContext();
+        $context->setBaseUrl('');
+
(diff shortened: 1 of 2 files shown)


fritzmg (approved):

core-bundle/src/EventListener/PreviewAuthenticationListener.php
```diff
@@ -70,6 +70,8 @@ public function __invoke(RequestEvent $event): void
             UrlGeneratorInterface::ABSOLUTE_URL
         );
 
+        $url = str_replace($request->getScriptName(), '', $url);
```
ausi: Is a string replace a good idea here? Shouldn’t we instead change the router context by calling `$this->router->getContext()` and `$this->router->setContext()`?

bytehead: I thought about the same, but was not sure if it's a good idea? 🤔 I would actually prefer it.

bytehead: Adjusted in b23ccafda.

core-bundle/src/EventListener/PreviewAuthenticationListener.php
```diff
@@ -64,6 +64,9 @@ public function __invoke(RequestEvent $event): void
             return;
         }
 
+        $context = $this->router->getContext();
+        $context->setBaseUrl('');
+
```
ausi, line 71: Should we maybe reset the base URL after using the router? Or maybe clone the context and reset to the old context afterwards?

bytehead: Good idea.

bytehead: Fixed in 7e3aa7615.

ausi (approved):

aschempp:
do I understand it correctly that we need to remove the base because otherwise the system would automatically redirect to the backend login without preview.php? Maybe it wouldn't be better to not force the redirect on the login route?

bytehead:
The request would be signed for 

```
https://demo.contao.org/preview.php/contao/login.html?redirect=...
``` 

but you will be redirected to 

```
https://demo.contao.org/contao/login?....
```

where the sign isn't valid obviously.

bytehead:
why would you redirect to the backend login with `/preview.php`?

leofeyer (approved):

aschempp:
I'm just thinking about handling other edge cases. With your changes, our current problem will be solved for sure, but we would need to rewrite the context in every other place where we force a back end login and then redirect.

Shouldn't this also be solved by setting `_allow_preview` on the `conto_backend_login` route?

bytehead:
Where else then in the `PreviewAuthenticationListener` should this happen?

leofeyer:
> Shouldn't this also be solved by setting `_allow_preview` on the `conto_backend_login` route?

Wouldn't this prevent the redirect (see #3464) for the `/preview.php/contao/login` route?

leofeyer:
As discussed in the Contao call, we want to merge the PR as is:

leofeyer:
Thank you @bytehead.

PR #3693, opened 2021-11-17
Do not replace insert tags in the back end

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes https://github.com/contao/contao/pull/3622#issuecomment-971806487
| Docs PR or issue | -

This PR fixes the problems mentioned in https://github.com/contao/contao/pull/3622#issuecomment-971806487 by checking the request scope within `TemplateInheritance::inherit`. That's the only solution I can think of, while also keeping the granularity that https://github.com/contao/contao/pull/3622 intended to introduce when it comes to replacing insert tags.


m-vo (dismissed):
:heart: Thanks for fixing! I like the solution, it's simple and it does not affect the new implementation with Twig where insert tags are never replaced without explicitly denoting.

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -160,8 +161,13 @@ public function inherit()
 			}
 		}
 
-		// Replace insert tags
-		$strBuffer = System::getContainer()->get(InsertTagParser::class)->replace($strBuffer);
+		// Replace insert tags, but not for the back end
+		$request = $container->get('request_stack')->getCurrentRequest();
+
+		if (null === $request || !$container->get('contao.routing.scope_matcher')->isBackendRequest($request))
```
leofeyer: Shouldn't we use `isFrontendRequest()` instead? Because we only want insert tags to be replaced in the front end and not in any other scope (even though there is none yet).

fritzmg: I am not sure. For example, if a template is parsed on the console and the insert tags are not automatically replaced, they need to be replaced manually, if you need them replaced. However, at that point you lose the granularity of replacing insert tags per template again, rather than replacing them on the finished string.

imho with the new concept (replacing in `TemplateInheritance::inherit`) the insert tags should always be replaced, except for the back end.

But then again, the new concept might actually be a BC break since some applications might not want insert tags to be replaced automatically when using `$template->parse()` and thus may be we need to replace insert tags _only_ for the Contao frontend scope indeed. But then may be insert tags should be replaced somewhere else entirely anyway?

@contao/developers wdyt?

bytehead: > Because we only want insert tags to be replaced in the front end and not in any other scope (even though there is none yet).

You could also argue, that you only don't want them to be replaced in the backend.

May there is at some point a scope `API` or the like where it also should be replaced I think.

ausi: IMO the check for the backend scope is correct. Can we additionally add a parameter to `$template->parse()` that controls if insert tags should be replaced?

- `$template->parse()` => replace insert-tags if not in backend scope
- `$template->parse(true)` => replace insert-tags always
- `$template->parse(false)` => do not replace insert-tags

leofeyer: Do you want to add the parameter in this PR or in a separate one?

ausi: I would say however @fritzmg prefers ☺️

fritzmg: I'll add it to this PR :)

m-vo: > Can we additionally add a parameter to $template->parse() that controls if insert tags should be replaced?

I'm not sure about this one. This would suggest that insert tags are replaced when passing `true` even though they wouldn't if I use a Twig template. If we add a parameter I'd go for somthing like `$neverReplace = false` that can be set to `true`, but I'd just omit it tbh.

leofeyer (dismissed):

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
fritzmg: Done. Not sure if I like though 😬

leofeyer: @fritzmg Please revert the last commit. @m-vo has explained why the parameter is not a good idea.

fritzmg: Right, haven't seen that comment. But still, since insert tags are now _always_ replaced now (except for the back end), which was previously not the case, shouldn't we allow to disable replacing insert tags _somehow_? May be via the Insert Tag service?

```php
$this->insertTags->pushDisable();
$template->parse();
$this->insertTags->popDisable();
```

m-vo: They were always replaced ever since, just at higher granularity level. And they are never replaced in Twig. 

The only thing I could imagine is someone using templates that aren't output in the frontend and which includes patterns used that look like insert tags `{{foo}}` that now get removed.

fritzmg: > They were always replaced ever since, just at higher granularity level. And they are never replaced in Twig.

No, that's not true. They were never replaced with `$template->parse()`, only with `$template->output()` or `$template->getResponse()` (previously).

m-vo: Yes, I meant 'if output in the front end' which is where templates are probably used 99% of the time. :smile:

fritzmg: Just to reiterate: this is where insert tags are replaced for the front end in Contao 4.9.22:

https://github.com/contao/contao/blob/0406c1b94e4780b021c44bf646dac7392ff5ce36/core-bundle/src/Resources/contao/classes/FrontendTemplate.php#L127

Executing

```php
$template = new FrontendTemplate('foobar');
$buffer = $template->parse();
```

will result in insert tags being present and unreplaced in `$buffer`. `FrontendTemplate` will replace insert tags _only_ in `FrontendTemplate::compile` and `FrontendTemplate::compile` will _only_ be executed when using `FrontendTemplate::output` or `FrontendTemplate::getResponse`.

fritzmg: And @andyschempp 's PR https://github.com/contao/contao/pull/3601 introduced a new `FragmentTemplate` class, which overrides the `compile` method, so that it will only execute `$this->parse()` without replacing any insert tags (or doing any other tasks that are actually only relevant to the `fe_page` template.

ausi: We should go with something like `parse($skipReplaceInsertTags = false)` then I think. Twig templates however should ignore this parameter completely I think.

fritzmg: So may be we actually need to revert https://github.com/contao/contao/pull/3622/ in order to keep backwards compatibility. I do think it is a problem that `$template->parse()` now replaces insert tags without any control over it.

m-vo: Should we discuss this in our call later this day? Might be easier :relaxed:.

Edit: I see @leo already added the label. :slightly_smiling_face:

leofeyer: As discussed in the Contao call, we want to reverse the logic and only replace insert tags in the front end. We also want to add an `$preserveInsertTagsInLegacyTemplates` argument to the `parse()` method.

fritzmg:
Btw. the change from #3622 might be considered a BC break in any case. Before, you were able to sort of adjust the output of the hardcoded insert tags of the core by replacing them to another insert tag in an `outputFrontendTemplate` hook and then implement a `replaceInsertTag` hook for that new insert tag.

Since the insert tags are now not replaced between the `outputFrontendTemplate` and `modifyFrontendPage` hooks anymore, this is now not possible anymore. However, I think we should introduce a better mechanism to change the output of _any_ insert tag anyway (see also https://github.com/contao/contao/pull/3716#issuecomment-976378960).

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
fritzmg: > As discussed in the Contao call, we want to reverse the logic and only replace insert tags in the front end. We also want to add an `$preserveInsertTagsInLegacyTemplates` argument to the `parse()` method.

Done in f0dd1ca

m-vo:
Should the modifications to `parse()` also be done for the `Widget` class?

fritzmg:
> Should the modifications to `parse()` also be done for the `Widget` class?

Yes, I missed that. Will do.

aschempp (commented):
any particular reason we use the new "argument as comment" syntax?

fritzmg:
> any particular reason we use the new "argument as comment" syntax?

We can't define an actual argument because of backwards compatibility.

fritzmg:
Ooph, I just realised that most front end Widgets implement their own `parse` method.

I am actually now in favor of being able to enable/disable insert tag replacing globally (e.g. via `$insertTagParser->disable()`). Or revert https://github.com/contao/contao/pull/3622 (as it constitutes a BC break) - otherwise I find this too messy.

fritzmg:
Closing in favor of #4450

PR #3694, opened 2021-11-17
Add the URI and page ID to log entries

### LINKED ISSUE #2389: Show the affected pages in the system log
When you have a look onto the error messages there's no detailed information which page is affected in detail. 

So when I have a bunch of errors like this, I absolutely don't know where to look.
Bildschirmfoto 2020-10-01 um 09 37 56

The **page ID** would be very helpful, maybe even the **article or content ID**, as well as a **quickjump to the page/content element**.

Maybe even a link to the **frontend page** and the page that was visited before (if possible).

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2389

This pull requests adds a URI and page ID field to `tl_log` and extends the `ContaoContext` for Monolog with the URI of the current request and the ID of the current page (if available). Both fields are searchable and sortable in the log view.

In #2389 it was suggested to also add references to the article and/or content element to a log entry. This is rather tricky and since many log entries are related to neither an article nor a content element, I decided to skip these.

The full request URI already provides a lot of context to find the source content of a log entry.

## Operations for tl_log

The tl_log view in the backend is also extended by two operations:
- `edit articles` to jump to the articles of the page related to a log entry 
- `jump to URI` to open the URI of a log entry in a new tab

Bildschirmfoto 2021-11-17 um 19 44 14

I'm not sure I like the two new operations (that's why they are still missing their tests). While both actions can be very convenient when debugging the source of a log entry, they clutter the log view and are useless for many entries.

I'd rather link the request URI and maybe add a link to edit a page and its articles in the show popup. Sadly, this is currently not possible since all values are encoded via `StringUtil::specialchars()` in the popup. 

Maybe this can be adjusted in another pull request so one can add custom HTML code to the table cells in the show popup. I'll happily drop the two new operations in this case.

## Additional notes:
- The URI is urldecoded, but encoded via `StringUtil::specialchars()`. Encoding is probably not necessary but I remembered CVE-2021-35210 and decided it's better to be safe than sorry.
- The URI in the log context may differ slightly from the real URI of a request since Symfony's `$request->getUri()` rebuilds the query string (transforming `foo.html?foo=bar&baz` to `foo.html?baz=&foo=bar`. This should be irrelevant for the logging.
- Can there be any reason to not store the URI under certain conditions (e.g. with sensitive data)?


aschempp:
I generally like the idea of adding additional context to a log entry. I wonder if we should allow for arbitrary data (e.g. a JSON/serialized array on each log entry). I also like the buttons, but I would simply not render the link button for entries that do not have a URL. That would clean up the UI a lot I think.

SeverinGloeckle:
As discussed in the Contao call, the operations should be removed from the view.

In the future, the URI in the show popup should become a clickable link. This requires adjustments in `DC_Table::show()`, which I will add in a follow up pull request.

SeverinGloeckle:
Follow-up adjustments to `DC_Table::show` should be delayed until a decision has been made about the future of the system log (see https://github.com/contao/contao-manager/issues/84 and https://github.com/contao/contao/pull/3702#issuecomment-976323815).

Since both `URI` and `page ID` from this pull request are added to the `ContaoContext`, they may also be used in other log views in the future. So from my point of view, there are no changes required for this request regardless of the future of the system log.

SeverinGloeckle:
The implementation of the system logger service in #3702 does not influence this pull request.

I update the `tl_log.uri` field to be in line with https://github.com/contao/contao/pull/3812.

An option to jump to the page, article or URI of a log entry may be added as follow up. For now, the feature is as follows:
- Automatically add the URI and the page ID to each log message.
- Show the URI and page ID in the `show` popup of the system log.
- Add sort and search options for URI and page ID in the system log.

If there's no problems with this:

> ## Additional notes:
> * The URI is urldecoded, but encoded via `StringUtil::specialchars()`. Encoding is probably not necessary but I remembered [CVE-2021-35210](https://github.com/advisories/GHSA-h58v-c6rf-g9f7) and decided it's better to be safe than sorry.
> * The URI in the log context may differ slightly from the real URI of a request since Symfony's `$request->getUri()` rebuilds the query string (transforming `foo.html?foo=bar&baz` to `foo.html?baz=&foo=bar`. This should be irrelevant for the logging.
> * Can there be any reason to not store the URI under certain conditions (e.g. with sensitive data)?

then this request is ready for review.

core-bundle/tests/Monolog/ContaoTableProcessorTest.php
```diff
@@ -306,6 +307,88 @@ public function sourceProvider(): \Generator
...
+
+        /** @var ContaoContext $context */
+        $context = $record['extra']['contao'];
+
+        $this->assertSame($uri, $context->getUri());
+    }
+
+    public function requestProvider(): \Generator
+    {
+        yield 'regular URL' => [Request::create('https://www.contao.org/foo?bar=baz'), 'https://www.contao.org/foo?bar=baz'];
```
ausi:
```suggestion
        yield 'regular URL' => [Request::create('https://www.contao.org/foo?bar=baz'), 'https://www.contao.org/foo?bar=baz'];
        yield 'encoded URL' => [Request::create('https://www.contao.org/foo?bar=baz&foo=b%20r'), 'https://www.contao.org/foo?bar=baz&amp;foo=b%20r'];
```

core-bundle/src/Monolog/ContaoTableHandler.php
```diff
@@ -65,22 +63,22 @@ public function handle(array $record): bool
...
-        $this->statement->executeStatement([
+        $this->getConnection()->insert('tl_log', [
             'tstamp' => $date->format('U'),
             'text' => StringUtil::specialchars((string) $record['formatted']),
             'source' => (string) $context->getSource(),
             'action' => (string) $context->getAction(),
             'username' => (string) $context->getUsername(),
             'func' => $context->getFunc(),
             'browser' => StringUtil::specialchars((string) $context->getBrowser()),
+            'uri' => StringUtil::specialchars(urldecode($context->getUri() ?? '')),
```
ausi:
```suggestion
            'uri' => StringUtil::specialchars($context->getUri() ?? ''),
```

I don’t think it should be URL-decoded here (and this would potentially be dangerous)

ausi (commented):
> Can there be any reason to not store the URI under certain conditions (e.g. with sensitive data)?

IMO no. Sensitive data should never appear in the query string, but only in the POST body.

leofeyer (commented):
The PR is working fine, thank you @SeverinGloeckle. I was wondering if we should limit adding URIs to the front end? Because this does not really help anyone, does it?



@Toflar @ausi /cc

fritzmg:
imho it wouldn't hurt. However I would add `overflow:hidden; text-overflow: ellipsis;` to the field (and link the URL).

leofeyer:
> imho it wouldn't hurt.

IMHO it would. We are storing hundreds of longish URLs in the table although we already know that they will not be helpful. 🤷‍♂️

Toflar:
I also think it doesn't hurt. Logs are cleaned up. And every webserver logs all the request URIs in its access log too so 🤷‍♂️ 
But if for some reason you think this is a problem, we can also limit the additional info to the FE to begin with. Although, I'd prefer to have it in the BE as well, see how it goes and remove it in case there's an issue.

ausi (approved):

ausi:
> I was wondering if we should limit adding URIs to the front end?

I think even in the backend the URL can be very helpful for debugging purposes.

leofeyer:
Thank you @SeverinGloeckle.

PR #3695, opened 2021-11-17
Switch to Symfony's version of the Path helper

m-vo:
The `webmozart/path-util` package got abandoned but the `Path` helper class was added to the `Filesystem` component of Symfony 5.4. 

Besides the `getFilename()` method removed (which is basically `basename`) and the use of a variadic instead of an array parameter in `getLongestCommonBasePath()` everything stayed the same.

The CI jobs must be rerun once Symfony 5.4 is released.


ausi (dismissed):

Toflar (dismissed):

bytehead (dismissed):

core-bundle/tests/Twig/Loader/ContaoFilesystemLoaderTest.php
```diff
@@ -664,7 +664,7 @@ private function mockFilemtime(array $pathToMtime): void
                     namespace %s;
 
                     function filemtime(string $filename) {
-                        if (null !== ($mtime = unserialize('%s')[\Webmozart\PathUtil\Path::canonicalize($filename)] ?? null)) {
+                        if (null !== ($mtime = unserialize('%s')[\Symfony\Component\Filesystem\Path::canonicalize($filename)] ?? null)) {
```
leofeyer, line 667: Why are we using a FQCN here?

m-vo: It's part of a string. We could also write an import, but this looks simpler to me.

leofeyer: > It's part of a string.

Really? It seems like an array key to me. 🤔

ausi: @leofeyer the whole code block is part of a string (using the nowdoc syntax), see line 663: `<<<'EOPHP'`

aschempp:
@m-vo 4.x now requires Symfony 5.4, can you please resolve the merge conflicts?

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3696, opened 2021-11-17
Lower the maximum insert tag recursion level

m-vo:
With the insert tag recursion limit of 100 you'll hit the default xdebug recursion limit when running tests:

```
1) Contao\CoreBundle\Tests\Contao\InsertTagsTest::testInfiniteRecursionWithCatchInsertTag
Error: Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames
```

For me it started working with 78  and lower. This PR sets it to 64. We could alternatively increase the level by calling sth. like `ini_set('xdebug.max_nesting_level', 1024);` in the tests but IMHO 64 is more than enough.

Credits @ausi for debugging.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -33,7 +33,7 @@ use Webmozart\PathUtil\Path;
  */
 class InsertTags extends Controller
 {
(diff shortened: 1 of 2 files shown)


ausi (approved):

fritzmg (approved):

leofeyer:
Thank you @m-vo.

PR #3698, opened 2021-11-18
Support stateless routes

aschempp:
While finishing #3619 I noticed a flag in Symfony to define a route as stateless: https://symfony.com/doc/current/routing.html#stateless-routes
This appears to be similar to our "always load from cache flag". This PR enables `stateless` if the cache flag is set.

I'm unsure this is actually our case. I'm pretty sure people will use the `alwaysLoadFromCache` option even if they have extensions that use the cache. I know I did, but simply because the extension in use (Isotope 😬) does not support ESI etc. However, as far as the documentation goes, it will only produce a warning in production, so this should only help to debug in dev and not affect live systems.

### DIFF
--- a/core-bundle/src/Routing/Page/PageRoute.php
+++ b/core-bundle/src/Routing/Page/PageRoute.php
@@ -43,6 +43,7 @@ class PageRoute extends Route implements RouteObjectInterface
                 '_scope' => ContaoCoreBundle::SCOPE_FRONTEND,
                 '_locale' => LocaleUtil::formatAsLocale($pageModel->rootLanguage),
                 '_format' => 'html',
+                '_stateless' => $pageModel->includeCache && $pageModel->alwaysLoadFromCache,
                 '_canonical_route' => 'tl_page.'.$pageModel->id,
             ],
             $defaults
--- a/core-bundle/tests/Routing/Page/PageRouteTest.php
+++ b/core-bundle/tests/Routing/Page/PageRouteTest.php
@@ -108,6 +108,21 @@ class PageRouteTest extends TestCase
         $this->assertSame(['https'], $route->getSchemes());
     }
 
+    public function testRouteIsStatelessWhenPageIsAlwaysLoadedFromCache(): void
+    {
+        $route = new PageRoute($this->mockPageModel(['includeCache' => '', 'alwaysLoadFromCache' => '']));
+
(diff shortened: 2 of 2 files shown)


Toflar (changes requested):
I don't think we should do this. The attribute is used to determine whether a session should have been used on this route at all or not.
The place where this attribute is used is in the `AbstractSessionListener` (https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpKernel/EventListener/AbstractSessionListener.php#L238-L250).

It will report if you accessed the session even though you have a route which should not access the session.
Our use case is, however, different. You are allowed to access the session if that checkbox is set. We ensure the response will be private in that case anyway (`MakeResponsePrivateListener` which by the way also disables half of the logic of the `AbstractSessionListener`) so nothing will happen. However, if there is no session, it will generate a cache entry and also use it for other logged in users (aka no `Vary: Cookie` header). 
So imho this change is pointless.

aschempp:
> You are allowed to access the session if that checkbox is set. We ensure the response will be private in that case anyway so nothing will happen. However, if there is no session, it will generate a cache entry and also use it for other logged in users.

You're absolutely right. But this essentially means as soon as a cache entry is created, noone (including the logged-in user) will access the page with a session. Isn't the no-cache-if-session only a precaution handling that should never happen?

Toflar:
Contao ensures the cache entry is only created when allowed to create. No matter whether our checkbox is set or not. The checkbox only affects **the delivery** of a potentially already existing entry. This is done by **not** setting `Vary: Cookie` when the entry is created.
But your changes would now cause exceptions for people accessing the session on those pages. But in our case, that's totally valid. It will just cause the request to be uncacheable.

aschempp:
> But your changes would now cause exceptions for people accessing the session on those pages.

Yes, it will cause an exception **in debug mode**, not in production. 

> But in our case, that's totally valid. It will just cause the request to be uncacheable.

Is it though? It _will_ cause the request to be uncacheable as a precaution, but imho it should never happen in the first place?

Toflar:
It is totally valid to access the session. It is even desired! This PR is definitely wrong :)

leofeyer:
<del>As discussed in the Contao call, we do not want to prevent accessing the session even if the checkbox is enabled.</del>

PR #3702, opened 2021-11-18
Add a system logger service

### LINKED ISSUE #3701: Create a SystemLogger helper service
https://github.com/contao/contao/pull/3458#discussion_r751436071 @fritzmg 

> On that note: since `System::log` is also deprecated - you would proabably need to change any `$this->log` or `System::log` occurrence to:
> 
> ```php
> System::getContainer()->get('logger')->log('…', ['contao' => new ContaoContext(__METHOD__, ContaoContext::GENERAL)]);
> ```
> 
> However, I personally always find this rather cumbersome and still often use just `System::log`, since it's more convenient. May be we could introduce a `SystemLogger` helper service to improve DX here? Follow up PR of course, since out of scope.

SeverinGloeckle:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3701 

This request adds a new service `Contao\CoreBundle\Monolog\SystemLogger` to make it easier to create log messages within a Contao application.

With regular `LoggerInterface`:
````php
$this->logger->log(LogLevel::INFO, 'Info message', ['contao' => new ContaoContext(__METHOD__)]);
$this->logger->log(LogLevel::ERROR, 'Error message', ['contao' => new ContaoContext(__METHOD__, ContaoContext::ERROR)]);
$this->logger->log(LogLevel::INFO, 'Message with custom action', ['contao' => new ContaoContext(__METHOD__, ContaoContext::CRON)]);
````

With `SystemLogger`:
````php
$this->logger->info('Info message');
$this->logger->error('Error message');
$this->logger->log(ContaoContext::CRON, 'Message with custom action');
````

## Getting caller from debug trace
The calling method is resolved within the service using `debug_backtrace`. The function is optimized for performance with `DEBUG_BACKTRACE_IGNORE_ARGS` and a sensible limit.

The method can be set explicitly when triggering the log message:

````php
$this->logger->info('Info message', 'Class::method');
````

If we decide against using `debug_backtrace`, this `$method` parameter would become a required parameter in all three log methods.

## Not a `LoggerInterface`
The `SystemLogger` is designed specifically to create log messages with a Contao context. It does not implement `Psr\Log\LoggerInterface`. 

For messages with other log levels or other contexts, you would still have to use the regular logger and set a potential `ContaoContext` yourself.

## Private service
The `SystemLogger` service is not public. For legacy use, the existing, deprecated `System::log()` may be used.

### CODE REVIEW

**Review by fritzmg — COMMENTED** [review]
LGTM 👍 Though since this service is there for convenience anyway, I would vote for implementing shorthand functions for _all_ system log categories that the core uses, not just `info` and `error`, i.e. also `->cron()`, `->configuration()` etc.

**Comment by fritzmg** [comment]
> The `SystemLogger` service is not public. For legacy use, the existing, deprecated `System::log()` may be used.

Why though? `System::log` is deprecated and the core is not really supposed to use deprecated functionality (especially from its own framework). All `System::log(…)` entries should be replaced with `$container->get(SystemLogger::class)->…`. Some instances were already replaced with `$container->get('@logger')->log(…)`.

**Comment by SeverinGloeckle** [comment]
> I would vote for implementing shorthand functions for _all_ system log categories that the core uses, not just `info` and `error`, i.e. also `->cron()`, `->configuration()` etc.

Good idea. I have added the other actions from `ContaoContext` in https://github.com/contao/contao/pull/3702/commits/f2a9fea245db240910bd0df70395af8701985f80. 

We now have:
- access
- configuration
- cron
- email
- error
- files
- forms
- general (equivalent to info)
- info
- log (provide a custom action)

Two actions from `ContaoContext` are not added:
- newsletter, since it is not in the core bundle
- repository, since I guess that's deprecated

I decided to keep `SystemLogger::info()` as it is more intuitive than `SystemLogger::general()`.


>> The SystemLogger service is not public. For legacy use, the existing, deprecated System::log() may be used.

> Why though? System::log is deprecated and the core is not really supposed to use deprecated functionality (especially from its own framework). All System::log(…) entries should be replaced with $container->get(SystemLogger::class)->…. Some instances were already replaced with $container->get('@logger')->log(…).

To enforce best practices by keeping services private whenever possible.

From my point of view, anything without dependency injection must use `System::log` while anything else can inject the logger service via proper dependency injection.

Maybe we can add a public alias for the logger service if it absolutely has to be used in legacy code, but I'd rather use the regular logger in this case.

**Comment by m-vo** [comment]
> From my point of view, anything without dependency injection must use System::log while anything else can inject the logger service via proper dependency injection.

Please don't. We decided that we must not use deprecated functions ourselves and this would stop `System::log()` from becoming deprecated. Introducing a new service and using it *also* in the legacy code is IMHO the right thing to do. Making it `public` doesn't hurt as long as we're booting the legacy framework anyways.

File: core-bundle/src/Resources/config/services.yml [hunk 1]
```diff
@@ -1023,3 +1023,7 @@ services:
             version: 4.13
             message: Using the "%service_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "Contao\CoreBundle\String\SimpleTokenParser" instead.
         public: true
+
+    Contao\CoreBundle\Monolog\SystemLogger:
+        arguments:
+            - '@logger'
```
m-vo: The logger is an optional dependency everywhere else. Should/must we follow this here as well?

core-bundle/src/Resources/config/services.yml
```diff
@@ -1023,3 +1023,7 @@ services:
             version: 4.13
             message: Using the "%service_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "Contao\CoreBundle\String\SimpleTokenParser" instead.
         public: true
+
+    Contao\CoreBundle\Monolog\SystemLogger:
```
m-vo: According to #3686 this should be a service id + an additional alias.

core-bundle/src/Resources/config/services.yml
SeverinGloeckle: Logging may be optional for some services, but as this service is used for logging only, it relies on the logger in every method. 

You would probably make the new system logger an optional dependeny in your services, though.

core-bundle/src/Resources/config/services.yml
SeverinGloeckle: Yes, good point. I guess the ID would be `contao.monolog.system_logger`, then.

I'll wait until #3686 is merged and rebase this pull request on top of it.

SeverinGloeckle:
@m-vo  and @fritzmg are probably right. 

I'd still argue that a defensive approach with a private service may be more adequate (`System::log` would be replaced by getting the regular logger and manually setting the `ContaoContext' ), but that's nothing to argue about. I'll just make the new service public.

While at it, I'll replace the remaining calls to `System::log` with the new service and check whether the new service should also be used in places that previously used the regular logger with a Contao context.

m-vo:
> While at it, I'll replace the remaining calls to System::log with the new service and check whether the new service should also be used in places that previously used the regular logger with a Contao context.

Sounds good 👍

aschempp:
I totally understand simplifying this. However, on the last Contao Call we discussed about the different logging options (file vs. backend etc) and it wasn't really clear what to use when. There are also plans to add a log viewer to the Contao Manager (https://github.com/contao/contao-manager/issues/84). I would prefer to discuss this topic first, because we might consider to remove or rewrite the back end log in an upcoming version and therefore shouldn't add such a service.

SeverinGloeckle:
Good call.

Changing the system log or adding a new log viewer would also influence #3694 and especially its potential follow-ups.

I was re-thinking this utility logger over the past days and already changed quite a bit. It is more about providing convenient methods to add a `ContaoContext` to a log action now – which probably will be useful regardless of the type of view for the log (assuming the `ContaoContext` will still be used).

I'll update this pull request soon so there's a better basis for a discussion on the topic.

aschempp:
Just another quick though. Instead of adding a service, we could add (static) convenience methods to the `ContaoContext` class that simply expects the logger service as additional argument.

SeverinGloeckle:
> Instead of adding a service, we could add (static) convenience methods to the `ContaoContext` class that simply expects the logger service as additional argument.

That sounds tempting at first. However, in this case, I still need both the logger and the `ContaoContext`.

In my opinion, for optimal comfort it should be possible to create a typical Contao log message (e.g. `CRON`) without the `ContaoContext` class.

I had briefly experimented with a decorator for the Monolog logger. Sadly, this doesn't work out due to the necessity to add new methods to the logger.

Therefore, I would suggest a fluid interface:

````php
// Current
$logger->log('Purged the search tables', __METHOD__, ContaoContext::CRON);

// Super convenient (state of this pull request, logger only knows the convenience methods)
$logger->cron('Purged the search tables');

// New proposal (logger implements LoggerInterface, provides pre-defined contexts)
$logger->asContaoCron()->info('Purged the search tables');
````

In this example, the `asContaoCron()` defines the `ContaoContext` with `ContaoContext::CRON` for the following log action.

There are also additional utility methods, which may be combined as needed:

````php
$logger
  ->withContaoAction('CUSTOM')
  ->withContaoFunc('MyClass::myMethod')
  ->warning('Log message');

$logger
  ->asContaoAccess()
  ->withContaoUsername('admin')
  ->warning('Switched user');
````

This way, the typical Contao contexts can be added comfortably, while the developer still has all the freedom to choose the log level, add more contexts, etc.

**Comment by m-vo** [comment]
Why can't we add convenience methods to a decorated/additional logger? If you want those you'd inject this specific service. 

IMHO a usage like `$logger->cron('Purged the search tables');` should be the way to go, everything else seems unnecessarily complex for a helper. If you want something specific you can still use the current explicit way with the logger service and a `ContaoContext` that you create yourself.

**Comment by aschempp** [comment]
Again, we need to discuss this anyway. Thinking about it again, we must not introduce another logger service, because it would break PSR and make everything harder to test. The `ContaoContext` is just additional information for the logger, but monolog could be reconfigured to not write that into the Contao backend (but e.g. to a centralized logging system). 

Since `ContaoContext` is necessary anyway, my idea was something like that:

```php
class ContaoContext
{
    public static function cron(LoggerInterface $logger, string $message /* ... */)
    {
        $logger->log($message, ['contao' => new self(/* ... */)]);
    }
}
```

SeverinGloeckle:
> Why can't we add convenience methods to a decorated/additional logger?

Sorry, I was unclear. I meant decorate as in _make it a decorator of the logger_ to enhance the regular logger with the utility methods. This didn't feel right since the enhanced logger and the original logger can't be replaced by each other which contradicts the decorater pattern in my view.

At first I came up with `$logger->cron()` as well. But this leaves no possibility to define the log level so we couldn't use the logger to replace something like this:

https://github.com/contao/contao/blob/24b1731cb798ba9e6dc81135f2b1771d0d9fccd9/core-bundle/src/Image/ImageFactory.php#L293-L305

Additionally, if the Contao actions are directly mapped to methods, the utility logger cannot implement the regular `LoggerInterface` since there would be a hard conflict for `error()` and a conceptual mismatch between the utility methods, which add a context, and the regular logger methods, which do not.

In some cases, you could end up with having to inject both the regular logger and the utility logger if you want to log something conveniently with a Contao context and want to log something with a non-info log level.

So if it's just about replacing the `System::log()` calls, there's nothing wrong with `$logger->cron()` and similar. 

But in my opinion, a developer friendly logger enhancement should provide more. 

> Thinking about it again, we must not introduce another logger service, because it would break PSR and make everything harder to test.

That's how I ended up with a fluid interface to add the `ContaoContext` before calling any of the regular PSR logger methods.

Testing would be pretty similar to the regular logger, but instead of one assertion for the log method and its context argument, you'd have two assertions for method calls - plus you don't have to setup the expected `ContaoContext` in your tests.

````php
// Before
$context = [
    'contao' => new ContaoContext(
        'Contao\CoreBundle\EventListener\Security\SwitchUserListener::__invoke',
        ContaoContext::ACCESS
    ),
];

$logger
    ->expects($this->once())
    ->method('info')
    ->with($message, $context)
;

// After
$logger
    ->expects($this->once())
    ->method('asContaoAccess')
    ->willReturn($logger)
;

$logger
    ->expects($this->once())
    ->method('info')
    ->with($message)
;
````

The only downside in testing is the fact that if you don't care about the contexts, you cannot inject a `NullLogger` since it lacks the utility methods.

----

I updated the request with the current state of my code. Note that not all logger calls with a Contao context are replaced with the new logger, yet.

Maybe we can use this to see whether there's a preference for a utility-only service or a logger enhancement – or if the utility is better provided by the `ContaoContext`.

**Comment by fritzmg** [comment]
> At first I came up with `$logger->cron()` as well. But this leaves no possibility to define the log level so we couldn't use the logger to replace something like this:

@SeverinGloeckle  Do you mean because that code uses `$logger->warning`, even though it uses `ContaoContext::ERROR`? This is something we can ignore I think. It was probably done in error and the log level does not matter for the system log anyway (currently).

> Thinking about it again, we must not introduce another logger service, because it would break PSR and make everything harder to test.

@aschempp Why would it break PSR? It's just a helper service for the logger, not a logger in itself.

> The ContaoContext is just additional information for the logger, but monolog could be reconfigured to not write that into the Contao backend (but e.g. to a centralized logging system).

@aschempp But that wouldn't matter, would it? If monolog was reconfigured to do something else with these contao messages, then this adjustment would still be in effect when using the helper service.

**Comment by SeverinGloeckle** [comment]
> ... and the log level does not matter for the system log anyway (currently).

It doesn't matter for the system log, but it could matter for any custom listeners or for a Contao log in the future. That's why adding a Contao context to the regular logger sounds like the most useful utility feature to me.

As @aschempp suggested, this topic is probably `up for discussion`.

**Comment by aschempp** [comment]
As discussed in todays Contao Call, the idea is to use monolog channels to target the back end log. So as a developer, you would define this service:

```yaml
my_bundle.listener.foobar:
    class: MyBundle\Listener\FooClass
    tags:
        - { name: monolog.logger, channel: contao.cron }
```

This should wire you a service that logs to the Contao `tl_log` table and uses the `cron` action. It should also automatically wire the calling method through `debug_backtrace()`.

On the Contao side, we should automatically handle all `contao.xxx` channels. This means
 1. we need a compiler pass that runs after `LoggerChannelPass`
 2. we either need to find all services through their name (its pretty static, see https://github.com/symfony/monolog-bundle/blob/fde12fc628162787a4e53877abadc30047fd868b/DependencyInjection/Compiler/LoggerChannelPass.php#L50) or through their parent definition if possible (https://github.com/symfony/monolog-bundle/blob/fde12fc628162787a4e53877abadc30047fd868b/DependencyInjection/Compiler/LoggerChannelPass.php#L147)
 3. check if the argument 0 matches our schema (channel = `contao.xxx`)
 4. decorate the service with our `BackendLogger` (not the ideal name) that gets the inner service and the action parsed from channel argument

This `BackendLogger` should look something like this (assuming it decorates the service)

```php
class BackendLogger implements \Psr\Log\LoggerInterface
{
    private LoggerInterface $inner;
    private string $action;

    public function __construct(LoggerInterface $inner, string $action)
    {
        $this->inner = $inner;
        $this->action = $action;
    }

    // and any other method in the interface
    public function emergency($message, array $context = array())
    {
        $this->inner->emergency($message, $this->addContaoContext($context));
    }

    private function addContaoContext(array $context)
    {
        if (!($context['contao'] ?? null) instanceof ContaoContext) {
            $context['contao'] = new ContaoContext(
                $this->action,
                // some magic to get calling method from backtrace
            );
        }

        return $context;
    }
}
```

SeverinGloeckle:
As discussed in the Contao call, I have changed the implementation as described by @aschempp.

### Tag with logger channel

If your service wants to log to a specific Contao channel, you can simply tag your service accordingly:

```yaml
App\EventListener\MyListener:
  tags:
    - { name: monolog.logger, channel: contao.cron }
```

```php
public function __construct(LoggerInterface $logger)
{
    // Preconfigured logger that logs to the contao.cron channel and adds a ContaoContext with action CRON
    $this->logger = $logger;
}
```

### Loggers for default actions

To get rid of the `System::log()` deprecations, the `LoggerChannelPass` also creates logger services for the regular Contao actions:
* `monolog.logger.contao.access`
* `monolog.logger.contao.configuration`
* `monolog.logger.contao.cron`
* `monolog.logger.contao.error`
* `monolog.logger.contao.files`
* `monolog.logger.contao.forms`
* `monolog.logger.contao.general`

These channels are configurable by the configuration option `contao.monolog.default_channels` so users/extensions could add additional channels if they want to get a predefined, public logger service for a custom channel.

All calls to `System::log()` are thus replaced as follows:

```php
// Previously
$this->log('Generated news feed "' . $objFeed->feedName . '.xml"', __METHOD__, ContaoContext::CRON);

// New
System::getContainer()->get('monolog.logger.contao.cron')->info('Generated news feed "' . $objFeed->feedName . '.xml"');
```

### Loggers by parameter name

The services for the default actions may also be injected in a service by type hinting `LoggerInterface` and the [appropriate logger channel name](https://symfony.com/doc/current/logging/channels_handlers.html#monolog-autowire-channels):

```php
public function __construct(LoggerInterface $contaoAccessLogger)
{
    // Logger with contao.access channel
    $this->logger = $contaoAccessLogger;
}
```

This way, you may even use multiple predefined loggers within one service.

### Log levels & log channels / Contao actions

To anticipate a possible discussion: The service ID relates to the log channel - within the channel itself, all log levels are available. The following syntax can make perfect sense:

```php
// Channel contao.error and log level error
System::getContainer()->get('monolog.logger.contao.error')->error('Some error');
```

Since now you may define the log level within the error channel:

```php
// Channel contao.error and log level warning
System::getContainer()->get('monolog.logger.contao.error')->warning('Some warning');

// Channel contao.error and log level emergency
System::getContainer()->get('monolog.logger.contao.error')->emergency('Some warning');
```

This distinction did not exist before. But I think this is a good extension for the future.

core-bundle/src/EventListener/Security/SwitchUserListener.php
```diff
@@ -47,9 +46,6 @@ public function __invoke(SwitchUserEvent $event): void
         $sourceUser = $token->getUserIdentifier();
         $targetUser = $event->getTargetUser()->getUserIdentifier();
 
-        $this->logger->info(
-            sprintf('User "%s" has switched to user "%s"', $sourceUser, $targetUser),
-            ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS, $sourceUser)]
-        );
+        $this->logger->info(sprintf('User "%s" has switched to user "%s"', $sourceUser, $targetUser));
```
aschempp, line 47: Monolog/PSR supports placeholders for the message, should we use that? https://github.com/Seldaek/monolog/blob/main/doc/message-structure.md#log-message-structure
we would need to make sure the placeholders are replaced for the backend log message, but I think that's already the case.

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -263,10 +262,10 @@ public static function handleRunOnce()
 
 			if (!unlink($file))
 			{
-				throw new \Exception("The file $strRelpath cannot be deleted. Please remove the file manually and correct the file permission settings on your server.");
+				throw new \Exception('The file ' . $strRelpath . ' cannot be deleted. Please remove the file manually and correct the file permission settings on your server.');
```
aschempp, line 268: this looks like an unrelated change

core-bundle/src/Resources/contao/library/Contao/Email.php
```diff
@@ -603,7 +603,7 @@ public function sendTo()
 			}
 		}
 
-		System::log($strMessage, __METHOD__, $this->strLogFile);
+		System::getContainer()->get('contao.monolog.logger')->log($this->strLogFile, $strMessage);
```
aschempp: this looks like a wrong service name. Also, what is `$this->strLogFile`?

core-bundle/src/DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -0,0 +1,98 @@
...
+                $this->loggers[] = $id;
+            }
+        }
+
+        foreach ($container->getDefinitions() as $id => $definition) {
+            if (!\in_array($id, $this->loggers, true) && $this->isContaoChannelLoggerDefinition($definition)) {
+                $container->setDefinition(
+                    $id,
+                    new Definition(SystemLogger::class, [$definition, $this->getContaoActionFromChannel($definition->getArgument(0))])
+                );
```
aschempp: This should create a decorated service instead of overriding the monolog service. Something like that:

```php
$container
    ->register("contao._logger.$id", SystemLogger::class)
    ->setDecoratedService($id)
    ->setArguments([new Reference("contao._logger.$id.inner"), $this->getContaoActionFromChannel($definition->getArgument(0))]);
```

We could even have a template service (similar to `monolog.logger_prototype`) but that might be over the tops.

aschempp (changes requested):
Looks very good! And apparently the concept seems to work? 🎉 

I would not add a channel/action configuration to the Contao bundle though. First of all, anyone using a custom channels in this new way really should use dependency injection and tag the service. If that is not possible for whatever reason, there is still `monolog.additional_channels` that can be configured on the app or through a `ManagerPlugin` that offers the exact same functionality.

m-vo:
Wohoo! :tada: :heart: 

Just to understand the concept: for each configured action a logger service definition `monolog.logger.contao.[action]` gets created, right? Why can't we create them on our own in the `ContaoCoreExtension`?

```php
// Contao\CoreBundle\DependencyInjection\ContaoCoreExtension (pseudo code)

foreach($actions as $action) {
    $definition  = new Definition(LoggerInterface::class);
    // $definition->setArgument(…)

    $name = "contao.logger.$action";
    $container->setDefinition($name, $definition);
    $container->registerAliasForArgument($name, LoggerInterface::class, "{$name}Logger");
}
```

My question basically is: What's the benefit to let monolog create them if we need to change the definitions anyways?

aschempp:
> My question basically is: What's the benefit to let monolog create them if we need to change the definitions anyways?

see first code example in https://github.com/contao/contao/pull/3702#issuecomment-995926557 😉

m-vo:
>> My question basically is: What's the benefit to let monolog create them if we need to change the definitions anyways?
>
> see first code example in #3702 (comment) wink

The first code example is config that wouldn't be needed in my example, no?

(Plus the added benefit, that you in fact *can* inject multiple instances, even with explicitly configuring. But maybe I'm misunderstanding something.)

aschempp:
it wouldn't be needed if you follow the argument name convention. But I would consider the monolog service tag to be more explicit and follow the best practice.

core-bundle/src/EventListener/Security/SwitchUserListener.php
SeverinGloeckle: This would be nice. However, this is out of the scope of this pull request in my opinion and may be added at a later point in time.

core-bundle/src/Resources/contao/classes/Backend.php
SeverinGloeckle: I changed the quotes for the log message in the line below and couldn't resist the urge to "correct" the quotes in this line as well. It looked weird having pretty much the same construct within 2 lines but with different syntax.

core-bundle/src/Resources/contao/library/Contao/Email.php
SeverinGloeckle: Good catch. Its leftover from the former logger implementation.

In fact, there is no simple change for this line since `$this->strLogFile` – and thus the `action` of the Contao context – is set by the newsletter bundle to a dynamic value `ContaoContext::NEWSLETTER . '_' . $objNewsletter->id`.

Therefor the logging has to be wired up manually at this point. I have changed this in https://github.com/contao/contao/pull/3702/commits/24d4871cad77e59068682a98a94e873067b03dfd

core-bundle/src/DependencyInjection/Compiler/LoggerChannelPass.php
SeverinGloeckle: You are right. I have changed this in https://github.com/contao/contao/pull/3702/commits/8f6649e0b6510d78b42086fc072642a624080dd9 – simpler and more transparent.

SeverinGloeckle:
> The first code example is config that wouldn't be needed in my example, no?
> 
> (Plus the added benefit, that you in fact _can_ inject multiple instances, even with explicitly configuring. But maybe I'm misunderstanding something.)

You are correct: While there already was logic in place to create a dedicated, public logger service for each standard Contao action, the alias registering was missing. I have fixed this in https://github.com/contao/contao/pull/3702/commits/4c2c3d8a5c90f70c7500fc880bd6e5a7af787d29 – so wiring by parameter name (and injecting multiple logger channels) works correctly now.

SeverinGloeckle:
> I would not add a channel/action configuration to the Contao bundle though. First of all, anyone using a custom channels in this new way really should use dependency injection and tag the service. If that is not possible for whatever reason, there is still `monolog.additional_channels` that can be configured on the app or through a `ManagerPlugin` that offers the exact same functionality.


The configuration option targets two audiences:
- Developers of legacy modules who want to replace `System::log()`. If they used custom actions, they would have to generate a ContaoContext for each message – with a config option, they can add their log action and replace the log call with a one-liner `System::getContainer()->get('monolog.logger.contao.my_action').
- Developers that prefer to inject the logger channel by parameter name and want to use custom Contao actions/channels.

However, I'd be fine with dropping the configuration option (as long as we keep the 8 loggers for the default actions).

aschempp:
> (https://github.com/contao/contao/commit/4c2c3d8a5c90f70c7500fc880bd6e5a7af787d29) – so wiring by parameter name (and injecting multiple logger channels) works correctly now.

Are you sure the parameter name should be the logger name? Wouldn't that need a parameter `__construct(LoggerInterface $contao.cron.logger)`? @m-vo 


> * Developers of legacy modules who want to replace `System::log()`. If they used custom actions, they would have to generate a ContaoContext for each message – with a config option, they can add their log action and replace the log call with a one-liner `System::getContainer()->get('monolog.logger.contao.my_action').
> * Developers that prefer to inject the logger channel by parameter name and want to use custom Contao actions/channels.

What I meant is both of this is still possible through the standard monolog configuration: https://symfony.com/doc/current/logging/channels_handlers.html#configure-additional-channels-without-tagged-services

SeverinGloeckle:
> Are you sure the parameter name should be the logger name? Wouldn't that need a parameter `__construct(LoggerInterface $contao.cron.logger)`? @m-vo

Yes, the parameter names of the loggers are correct. Each channel gets converted to camel case by Symfony so `contao.cron.logger` becomes `LoggerInterface $contaoCronLogger`:

https://github.com/symfony/symfony/blob/aec12ab33d585b2dd18c6dfe868e4b9d3381d892/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1319-L1336


> What I meant is both of this is still possible through the standard monolog configuration:

Great point, the additional services are then directly created and configured by Monolog! I have adjusted the setup of the channels in https://github.com/contao/contao/pull/3702/commits/3073954e76348f9e0d5b19a2af33709816763ea8.

The naming of the parameters remains as described above (camel case of the channel, e.g. `LoggerInterface $contaoCronLogger`).

aschempp (dismissed):
Code looks good now 🎉 
I'm unsure whether the Contao monolog channels should be done through prepend in the core bundle or through the Manager plugin. Doing it in the plugin means that Symfony app users **must** manually add the same configuration or the system might fail. That's true for the framework configuration as well though 🤷

SeverinGloeckle:
> I'm unsure whether the Contao monolog channels should be done through prepend in the core bundle or through the Manager plugin. 

For me, a configuration in the bundle extension has a somewhat higher significance than in the manager plugin. Since the logger channels are critical functionality, I would leave the setup in `ContaoCoreExtension`.

You are right that the corresponding configuration should be documented for use in a standard Symfony application (similar to [here](https://docs.contao.org/dev/getting-started/initial-setup/symfony-application/contao-4.9/#configure-your-contao-installation), but for Contao 4.13).

Generally, there are two things to be documented:
1. working with the logger service 
	- via service tagging
	- via autowiring
	- in legacy code
	- adding additional channels
	- mocking in unit tests
2. configuring the logger channels in a standard Symfony application
    - add the Monolog logger channels for Contao

I can add documentation for point 1. The second topic may be added by someone with an overview of all necessary configuration for Contao 4.13.

Can you add the `needs documentation` badge please?

SeverinGloeckle:
By the way, I can't reproduce the failing ECS test locally. It looks like an error in the parallel execution of the fixer, but maybe someone can review that just to be on the safe side?

m-vo:
> By the way, I can't reproduce the failing ECS test locally. It looks like an error in the parallel execution of the fixer, but maybe someone can review that just to be on the safe side?

When I first ran it locally, the error occured as well, but then disappeared. I restarted the jobs - let's see if it's a race condition glitch or something alike…

aschempp:
> > I'm unsure whether the Contao monolog channels should be done through prepend in the core bundle or through the Manager plugin.
> 
> For me, a configuration in the bundle extension has a somewhat higher significance than in the manager plugin. Since the logger channels are critical functionality, I would leave the setup in `ContaoCoreExtension`.

Well it's not that it has higher significance. A prepended configuration is _always_ there, also for a Symfony application. A plugin configuration is only set in the Managed Edition. But a prepend configuration might be correct here, as far as I remember we added the fake `monolog.logger` channel in https://github.com/contao/contao/blob/69129f9a519689fbf6611e550ca9dc41d6003f23/core-bundle/src/Resources/config/services.yml#L398 just to have monolog create a `contao` channel 😅 

Regarding the documentation, I think this is something we should promote for new bundles. I didn't find any documentation on the `ContaoContext`, so we should probably just add a description of "new way" or "old way" when injecting a logger.

core-bundle/src/Resources/contao/classes/Backend.php
leofeyer: Since PHP 8, using double quotes and interpolation is more performant than single quotes and concatenation, so I guess the change should be reverted. I did not find the link though, maybe @m-vo can help?

leofeyer (commented):
I like this PR a lot. There is only one little thing that bothers me:

```php
$container->->get('monolog.logger.contao.error')->error('Website search failed');
```

Does this imply that I can log something other than an error in the `contao.error` channel? Or that I can log an error in the `contao.access` channel?

SeverinGloeckle:
> Does this imply that I can log something other than an error in the `contao.error` channel? Or that I can log an error in the `contao.access` channel?

Yes. As explained above:

> ### Log levels & log channels / Contao actions
> To anticipate a possible discussion: The service ID relates to the log channel - within the channel itself, all log levels are available. The following syntax can make perfect sense:
> 
> ```
> // Channel contao.error and log level error
> System::getContainer()->get('monolog.logger.contao.error')->error('Some error');
> ```
> 
> Since now you may define the log level within the error channel:
> 
> ```
> // Channel contao.error and log level warning
> System::getContainer()->get('monolog.logger.contao.error')->warning('Some warning');
> 
> // Channel contao.error and log level emergency
> System::getContainer()->get('monolog.logger.contao.error')->emergency('Some warning');
> ```
> 
> This distinction did not exist before. But I think this is a good extension for the future.

Since the system log currently doesn't do anything with the log levels, there's no reason to chose a special level _for now_, but there might be in the future.

For example, there could be an interactive log viewer in the Contao manager which let's you filter by channel (e.g. `access` to see everything access related) or log level (e.g. `error` and above to see all errors, criticals, alerts and emergencies) – and a combination of both, to see errors in the access channel only.

core-bundle/src/Resources/contao/classes/Backend.php
SeverinGloeckle: For me, it is more about consistency than performance, as the other log actions (even in the same file) use string concatenation or `sprintf`. 

If interpolation is more performant, we should probably check the whole system for such optimisations in a follow up pull request.

I can of course revert the change if that is preferred.

m-vo: Currently, interpolation seems to be the fastest (https://3v4l.org/atThi#v8.1.1) but it doesn't matter much, i.e. it's a micro optimization that probably doesn't matter anywhere. And this test is super-artificial.

However, string interpolation beats sprintf/concatenation in terms of readability. That's a personal opinion of course, but it's backed by literally all modern languages introducing it. :upside_down_face:

m-vo: (Dang it, three times edit for 4 lines of message - now it's right hopefully. :sweat_smile:)

SeverinGloeckle: Thanks for the interesting check. I'd still vote for a consistent style for now – if there should be changes, a separate issue to discuss and agree on a style for string concatenation is probably the proper way to go.

leofeyer:
The PR looks very good now, however, there are 7 places where the `ContaoContext` is still used. Did you omit them on purpose?

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Controller/Page/RootPageController.php#L45-L48

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Image/ImageFactory.php#L294-L303

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Security/Authentication/AuthenticationFailureHandler.php#L57-L60

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Security/Authentication/AuthenticationSuccessHandler.php#L104-L107

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Security/Authentication/FrontendPreviewAuthenticator.php#L103-L106

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Security/Logout/LogoutHandler.php#L56-L59

https://github.com/contao/contao/blob/ed7abca0709e08b7c9beee0ba55246cbdfa04369/core-bundle/src/Security/User/ContaoUserProvider.php#L133-L136

core-bundle/src/Resources/contao/classes/Backend.php
leofeyer: We could not find the link to the document anymore. It might have been deleted or even never existed 😄, so we will stick to Symfony's existing "prefer string concatenation over interpolation" rule.

core-bundle/src/Resources/config/services.yml
```diff
@@ -241,6 +241,8 @@ services:
             - '%contao.image.valid_extensions%'
             - '%kernel.project_dir%/%contao.upload_path%'
             - '@?logger'
+        tags:
+            - { name: monolog.logger, channel: contao.error }
```
aschempp, line 245: 🥳  easy peasy now!

SeverinGloeckle:
> ... there are 7 places where the `ContaoContext` is still used. Did you omit them on purpose?

`RootPageController`:
As a service subscriber, the controller only has access to the default logger. 

There are various ways to handle this:
1. A controller has to fall back to creating the `ContaoContext` manually.
2. A controller may request custom loggers by providing an extended `getSubscribedServices`.
3. A controller has to inject the required logger(s) via autowiring and constructor injection (e.g. `__construct(LoggerInterface $contaoErrorLogger)`).
4. The controller's service gets tagged with the desired log channel (e.g. `- { name: monolog.logger, channel: contao.error }` _- this only works if the whole controller logs to a single channel._
5. Contao's `AbstractController` provides service for all default logger channels (this would probably require some autowire magic for the channels).

I think variant 1 is appropriate for `RootPageController`.

For content or module controllers, I'd probably go with variant 3 or 4.

`ImageFactory`:
I have injected the correct logger channel and remove the `ContaoContext` in https://github.com/contao/contao/pull/3702/commits/2a1d80c59bb005045867969d25720a6dd909b6dc.

`Security/*`:
These handlers explicitly define a username for the context and thus have to create a custom `ContaoContext` manually.

fritzmg:
I'd go for 2 or 3 for the `RootPageController`.

leofeyer:
> I think variant 1 is appropriate for `RootPageController`.
> 
> For content or module controllers, I'd probably go with variant 3 or 4.

@aschempp WDYT?

aschempp:
A controllers should use constructor injection. ServiceSubscriber should only be used for parent methods that might not be called. Rewriting is not related to this PR though, so I wouldn't fix it here.

SeverinGloeckle:
As discussed in the Contao call, the error logger is now injected into `RootPageController`.

This request is now complete.

leofeyer (approved):

leofeyer:
Thank you @SeverinGloeckle for your patience and hard work. 😉

PR #3706, opened 2021-11-19
Add a service ID linter and adjust the service IDs

leofeyer:
This PR adds a linter to make sure the service IDs are consistent. <del>I have already made all the changes, however, the PR is still a draft, because I am not yet sure how to make the linter part of our CI chain. Maybe add it to the CS fixer?</del>

**Edit:** The linter is now in `tools/service-linter`.


core-bundle/src/Image/Studio/ImageResult.php
```diff
@@ -204,10 +204,10 @@ public function createIfDeferred(): void
             return;
         }
 
-        $resizer = $this->locator->get('contao.image.resizer');
+        $resizer = $this->locator->get('contao.image.legacy_resizer');
```
leofeyer, line 207: @ausi: I am not sure about this change. The class is called `LegacyResizer` but there seems to be no non-legacy resizer?

bytehead:
🥳 CS fixer sounds reasonable to me.

m-vo:
> Maybe add it to the CS fixer?

Sounds like a good place for me. As this doesn't need the application to run, this could also live as a simple script + composer.json (or single-command sf app) under tools/service-linter.

core-bundle/src/Command/LintServiceIdsCommand.php
```diff
@@ -0,0 +1,218 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $files = Finder::create()
+            ->files()
+            ->name('*.yml')
+            ->path('-bundle/src/Resources/config')
+            ->in(Path::join($this->projectDir, 'vendor/contao/contao'))
+        ;
+
+        $tc = 0;
```
m-vo: What does `tc` and `fc` stand for? :see_no_evil:

bytehead: I guess `totalCount` and `fileCount` 🙃

core-bundle/src/Image/Studio/ImageResult.php
ausi: Not sure. This class will probably go away in Contao 5.0 and there will be no class at all anymore or it would be called resizer. So keeping the service called `contao.image.resizer` would be better here I think.

leofeyer:
This one is ready for review now. 🎉

core-bundle/src/Resources/config/migrations.yml
```diff
@@ -2,117 +2,117 @@ services:
...
 
     # Backwards compatibility
     Contao\CoreBundle\Migration\MigrationCollection:
-        alias: contao.migration.migration_collection
+        alias: contao.migration.collection
         deprecated:
             package: contao/core-bundle
-            version: 4.13
-            message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "contao.migration.migration_collection" instead.
+            version_: 4.13
```
ausi:
```suggestion
            version: 4.13
```
mistake?

ausi (approved):
Spotted a small mistake (I think)

Instead of a seperate package `contao/service-id-linter` I would probably have put the files directly in the `tools/service-linter` folder (so that PRs have the option to change the command if a new service ID has problems with it).

Otherwise ❤️🎉🚀

ausi (approved):

richardhj:
I want to note that this PR introduces incompatibilities when it comes to service decoration.

Decorating a service by its old ID does not work, as for example here:
https://github.com/richardhj/contao-theme-framework/blob/c4100b108ac0d18532b36146868ed2b820aca11d/config/services.yml#L56-L57

Is this something that is worth addressing?

leofeyer:
The Twig integration is still experimental, therefore we are allowed to change the service IDs without providing a BC layer.

richardhj:
> The Twig integration is still experimental, therefore we are allowed to change the service IDs without providing a BC layer.

This is fine if it is the only affected service, no? (Haven't checked other services.)

PR #3711, opened 2021-11-21
Fix being unable to re-register after token expired

### LINKED ISSUE #3709: Follow-up registration is not possible under certain conditions
**Affected version(s)**

4.12.3

**Description**

There’s a period of time, when the follow-up registration is impossible, because the token has expired, but the member still exists.

Suppose, token is 23-hours old by the time Cron runs (daily by default). Then, an hour later it expires, and the user essentially stuck until the next cronjob runs (so, by default –  for the next 23 hours). He can’t log in (‘account has been disabled’), he can’t activate (‘invalid token’), and he can’t re-register (‘account exists’).

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3709
| Docs PR or issue | TBA

As explained in #3709, there can be a period where a user can neither log-in (since the account has not been activated), nor activate the account (since the token expired) nor register again (since the registration has not yet been purged). This is mainly because the registrations are only purged via the `purgeRegistrations` cronjob and this cronjob only runs every 24 hours. Thus this period where a member cannot login, activate their account or re-register can last up to almost 24 hours.

This PR fixes that by ensuring via `tl_member.config.onload` that the registrations are always purged immediately and so members can re-register immediately after their nonactivated registration expired. This also ensures that the back end never shows any registrations that should have actually been purged already.

This PR also changes the `purgeRegistrations` and `purgeOptInTokens` cronjobs to _hourly_ cronjobs, instead of daily cronjobs. Otherwise expired tokens and registrations can remain in the database almost up to 24 hours longer than intended. Technically we should probably use _minutely_, but since our other data purging cronjob, `PurgeExpiredDataCron`, is also only running _hourly_, I set these to hourly as well.


m-vo (dismissed):

core-bundle/src/EventListener/DataContainer/PurgeMemberRegistrationsListener.php
```diff
@@ -0,0 +1,56 @@
...
+    public function __invoke(): void
+    {
+        $request = $this->requestStack->getCurrentRequest();
+
+        // Do not purge during any back end edit actions
+        if (null !== $request && $this->scopeMatcher->isBackendRequest($request) && null !== $request->query->get('act')) {
+            return;
+        }
+
+        (new Automator())->purgeRegistrations();
```
leofeyer: We should not use the legacy `Automator` class here but copy the few lines of logic instead.

core-bundle/src/Resources/contao/config/config.php
```diff
@@ -380,10 +380,12 @@
 		'purgeTempFolder' => array('Contao\Automator', 'purgeTempFolder'),
 		'purgeSearchCache' => array('Contao\Automator', 'purgeSearchCache'),
 		'generateSitemap' => array('Contao\Automator', 'generateSitemap'),
+	),
+	'hourly' => array
+	(
 		'purgeRegistrations' => array('Contao\Automator', 'purgeRegistrations'),
-		'purgeOptInTokens' => array('Contao\Automator', 'purgeOptInTokens')
+		'purgeOptInTokens' => array('Contao\Automator', 'purgeOptInTokens'),
```
leofeyer: This could be a (future-proof) listener as well actually.

leofeyer (changes requested):

core-bundle/src/Resources/contao/config/config.php
fritzmg: Should we put all purge cron tasks into the existing `PurgeExpiredDataCron` may be?

core-bundle/src/Cron/PurgeMemberRegistrationsCron.php
```diff
@@ -0,0 +1,57 @@
...
+
+        /** @var MemberModel $memberModel */
+        $memberModel = $this->framework->getAdapter(MemberModel::class);
+
+        foreach ($memberModel->findExpiredRegistrations() ?? [] as $member) {
+            $member->delete();
+        }
+
+        if (null !== $this->logger) {
+            $this->logger->info('Purged the unactivated member registrations', ['contao' => new ContaoContext(__METHOD__, ContaoContext::CRON)]);
```
aschempp: this should only be logged if something was actually done. It previously didn't log if the member list was empty. And maybe it could include the count?

core-bundle/src/Cron/PurgeOptInTokensCron.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->optIn = $optIn;
+        $this->logger = $logger;
+    }
+
+    public function __invoke(): void
+    {
+        $this->optIn->purgeTokens();
+
+        if (null !== $this->logger) {
+            $this->logger->info('Purged the expired double opt-in tokens', ['contao' => new ContaoContext(__METHOD__, ContaoContext::CRON)]);
```
aschempp: this should only be logged if something was purged, otherwise it pollutes the back end log every hour for no reason (e.g. even if a system does not have registration at all)

core-bundle/src/EventListener/DataContainer/PurgeMemberRegistrationsListener.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Purges the member registrations in the front and back end, whenever tl_member is loaded (#3711).
+ *
+ * @internal
+ *
+ * @Callback(table="tl_member", target="config.onload")
+ */
+class PurgeMemberRegistrationsListener
```
aschempp: We should use the `reviseTable` hook instead of the `onload_callback`. `reviseTable` is only executed in exactly the case where this should be done, and it supports reloading the UI if something was deleted (dunno if that's actually necessary 🤷 )

aschempp (dismissed):

core-bundle/src/Cron/PurgeOptInTokensCron.php
fritzmg: `OptIn::purgeTokens` currently only returns `void`. Can we change the return type of a public function in a bugfix release? Although I could also query the database before and after in order to figure out if and how many opt in tokens were purged.

core-bundle/src/EventListener/DataContainer/PurgeMemberRegistrationsListener.php
fritzmg: Should we use `onload` only for the front end then, so that it is executed for any module related to `tl_member`?

core-bundle/src/Cron/PurgeOptInTokensCron.php
aschempp: I think we can, since we're only adding stuff. Anyone using the current implementation would ignore the returned value.

core-bundle/src/EventListener/DataContainer/PurgeMemberRegistrationsListener.php
aschempp: why do we need to purge in the front end? To make sure a user can register again? maybe we could explicitly check that on registration?

fritzmg: > why do we need to purge in the front end? To make sure a user can register again?

Yes, otherwise this wouldn't fix the problem.

> maybe we could explicitly check that on registration?

Yes, I guess it's sufficient to explicitly do it in `ModuleRegistration` only.

core-bundle/src/Cron/PurgeOptInTokensCron.php
fritzmg: That won't be possible since it's part of the interface. So I am counting the entries before and after now.

core-bundle/src/EventListener/PurgeExpiredMemberRegistrationsListener.php
```diff
@@ -0,0 +1,53 @@
...
+     * @var ContaoFramework
+     */
+    private $framework;
+
+    public function __construct(ContaoFramework $framework)
+    {
+        $this->framework = $framework;
+    }
+
+    public function __invoke(string $table): ?bool
```
leofeyer: IIRC, the method always returns `null`, so the return type does not match the method body. Which one is wrong?

core-bundle/src/Resources/config/services.yml
```diff
@@ -207,6 +207,19 @@ services:
             - '@contao.framework'
             - '@database_connection'
 
+    contao.cron.purge_member_registrations:
+        class: Contao\CoreBundle\Cron\PurgeMemberRegistrationsCron
+        arguments:
+            - '@contao.framework'
+            - '@?logger'
+
+    contao.cron.purge_optin_tokens:
```
```suggestion
    contao.cron.purge_opt_in_tokens:
```

core-bundle/src/Resources/contao/library/Contao/Automator.php
```diff
@@ -238,6 +238,8 @@ public function purgeTempFolder()
 	 */
 	public function purgeRegistrations()
 	{
+		@trigger_error(sprintf('Using "%s" has been deprecated and will be removed in Contao 5.0. Use MemberModel::findExpiredRegistrations() to find and delete expired registrations instead.', __METHOD__), E_USER_DEPRECATED);
```
leofeyer: Please also add an `@deprecated` annotation in the phpDoc comment.

leofeyer (changes requested):
Please revert the changes to the `ContaoCoreExtensionTest` class. It will make it more difficult to merge this PR upstream.

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
```diff
@@ -89,6 +89,11 @@ protected function compile()
 			}
 		}
 
+		// Purge expired opt-in tokens (#3709)
+		/** @var OptIn $optIn */
+		$optIn = System::getContainer()->get('contao.opt-in');
+		$optIn->purgeTokens();
```
leofeyer: I don't like that this code is now always run in the registration module, because it will affect the performance. Purging tokens should be done in the cron job only.

core-bundle/src/Cron/PurgeOptInTokensCron.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->database = $database;
+    }
+
+    public function __invoke(): void
+    {
+        $count = (int) $this->database->fetchOne('SELECT COUNT(*) FROM tl_opt_in');
+
+        $this->optIn->purgeTokens();
+
+        $count -= (int) $this->database->fetchOne('SELECT COUNT(*) FROM tl_opt_in');
```
leofeyer: I don't like this. Is it really necessary to know how many tokens have been purged? If so, the `purgeTokens()` method should probably return the value.

leofeyer (changes requested):

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
fritzmg: > Purging tokens should be done in the cron job only.

We can't rely on the cron job for purging tokens. No matter how often the cron job runs, there will always be a time period where it is possible that a member can neither login, nor activate, nor re-register.

core-bundle/src/Cron/PurgeOptInTokensCron.php
fritzmg: > I don't like this. Is it really necessary to know how many tokens have been purged?

No, it's just nice to have. We did not have this before, so I am fine with removing it.

> If so, the purgeTokens() method should probably return the value.

As already mentioned in https://github.com/contao/contao/pull/3711#discussion_r75826067 we cannot change the return type as it is part of the interface.

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
leofeyer: The period has been reduced from up to 24 hours to up to 60 minutes. That should suffice for most scenarios. And if someone really needs to further reduce the period, they can adjust the cron job to run minutely, can't they?

core-bundle/src/EventListener/PurgeExpiredMemberRegistrationsListener.php
fritzmg: The code that executes the hook checks the return value of the hook for `=== true`. So any hook implementation can either return `true` or something arbitrary, that's why I chose this return type. We can also define it as `void` as it would not matter in this case (just feels wrong to me).

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
fritzmg: > That should suffice for most scenarios.

It makes it less likely for it to occur. But I wouldn't consider a bug fixed just because it is less likely to occur.

core-bundle/src/Resources/config/services.yml
fritzmg: Done

core-bundle/src/Resources/contao/library/Contao/Automator.php
fritzmg: Done

fritzmg:
> Please revert the changes to the `ContaoCoreExtensionTest` class. It will make it more difficult to merge this PR upstream.

Done. However, do we not test for whether our services are properly registered anymore? I missed that we have removed that at some point.

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
fritzmg: @leofeyer what about adding 

```php
if (Input::post('FORM_SUBMIT') == $strFormId)
```

then the purging will only be done when the form is actually submitted, ensuring that you will always be able to re-register (if applicable).

fritzmg: Or we additionally introduce a `purgeTokensForEmail` function, so that we can target the affected email address specifically. This would then only be one additional database query per regular registration.

```php
if (Input::post('FORM_SUBMIT') == $strFormId && $email == Input::post('email'))
{
    $optIn->purgeTokensForEmail($email);
}
```

leofeyer: @contao/developers WDYT?

core-bundle/src/Cron/PurgeOptInTokensCron.php
leofeyer: Then I am for dropping the `$count` variable entirely.

fritzmg: Done.

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
fritzmg: Actually, the opt-in tokens do not matter. We need to make sure that the expired registrations are deleted, when the form is submitted.

leofeyer:
I am adding this to the discussion list, so we can discuss https://github.com/contao/contao/pull/3711#pullrequestreview-818056715 in the next Contao call.

fritzmg:
> I am adding this to the discussion list, so we can discuss [#3711 (review)](https://github.com/contao/contao/pull/3711#pullrequestreview-818056715) in the next Contao call.

Not sure there is much to discuss. Have you looked at the new version?

The registration module definitely needs to make sure that registrations are purged before any new registration attempt. The current implementation will purge the registrations _only_ for the provided email address and _only_ during the POST request when registering. So performance wise this should be totally fine.

leofeyer:
As discussed in the Contao call, we do not want to deprecate the Automator methods, because they are still used for the `contao:automator` command.

fritzmg:
Done, only the necessary changes in the registration module to fix this issue remains.

ausi (dismissed):

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
```diff
@@ -89,6 +89,17 @@ protected function compile()
 			}
 		}
 
+		$strFormId = 'tl_registration_' . $this->id;
+
+		// Purge expired registrations (#3709)
+		if (Input::post('FORM_SUBMIT') == $strFormId && $email = Input::post('email'))
+		{
+			foreach (MemberModel::findExpiredRegistrations(array('column' => array('email = ?'), 'value' => array($email))) ?? array() as $member)
```
leofeyer: How can this work? IIRC, `$arrOptions` is ignored in the model:

https://github.com/contao/contao/blob/61cf04641590af805a371cbc7d900f0b6eefe9fa/core-bundle/src/Resources/contao/models/MemberModel.php#L241-L255

leofeyer (changes requested):

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
fritzmg: True, I did not realise. Then I guess I need to implement a custom query in order to find expired registrations only for a particular email address.

leofeyer: How about adding a `findExpiredRegistrationsByEmail()` method to the `MemberModel` class? Or maybe we can adjust the existing query if `$arrOptions['email']` is set?

fritzmg: I'd vote for the former. The `findExpiredRegistrations` method having an `$arrOptions` argument is likely unintended.

fritzmg:
I have implemented a `MemberModel::findExpiredRegistrationByEmail` method as discussed. The new method is basically the opposite to `MemberModel::findUnactivatedByEmail`. I have also added deprecations for the unused `$arrOptions` arguments, which we can then remove entirely in Contao 5.

core-bundle/src/Resources/contao/models/MemberModel.php
```diff
@@ -240,6 +245,11 @@ public static function findUnactivatedByEmail($strEmail, array $arrOptions=array
 	 */
 	public static function findExpiredRegistrations(array $arrOptions=array())
 	{
+		if (func_num_args() > 0)
+		{
+			trigger_deprecation('contao/core-bundle', '4.9', 'Calling MemberModel::findExpiredRegistrations() with an argument has been deprecated and will no longer work in Contao 5.0. Do not pass $arrOptions as an argument anymore');
```
leofeyer: Since the argument was never used, you can just remove it.

Toflar (dismissed):

aschempp (dismissed):
Not sure why we're not simply removing all expired registrations, but this should work as well 🙃

fritzmg:
> Not sure why we're not simply removing all expired registrations, but this should work as well 🙃

See this discussion: https://github.com/contao/contao/pull/3711#discussion_r758549907

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3714, opened 2021-11-23
Do not unnecessarily fetch the PageRoute twice

aschempp:
Preparing feedback for https://github.com/contao/contao/pull/3605 I noticed these changes would prevent the route object from being loaded from the page registry multiple times.


core-bundle/src/Routing/RouteProvider.php
```diff
@@ -206,7 +203,7 @@ private function addRoutesForRootPage(PageModel $page, array &$routes): void
             $route->getMethods()
         );
 
-        if (!$urlPrefix || (!$this->legacyRouting && $page->disableLanguageRedirect)) {
+        if (!$urlPrefix || (!$this->legacyRouting && $page->loadDetails() && $page->disableLanguageRedirect)) {
```
fritzmg: `PageModel::loadDetails()` does not return a boolean, not sure if we should put that in there?

aschempp: it returns the page itself, which "equals" to true. It's only meant to inline-load the details before `disableLanguageRedirect` is queried

ausi:
```suggestion
        if (!$urlPrefix || (!$this->legacyRouting && $page->loadDetails()->disableLanguageRedirect)) {
```
Maybe something like this makes that more clear?

aschempp: good idea, changed in 018aa0e3ee56f2b18139f6eaa876805b744f679c. That also solves the phpstan issue 😆

fritzmg (approved):

Toflar (approved):

leofeyer (approved):

PR #3715, opened 2021-11-23
Delete obsolete language files

ausi:
It looks like the update-language script does not remove obsolete translation files.


ausi:
For better backwards compatibility we should probably merge #3716 first.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3716, opened 2021-11-23
Fix backwards compatibility for label::LNG and CNT

### LINKED ISSUE #3712: Missing LNG labels when using an insert tag
**Affected version(s)**

Contao 4.12

**Description**

The insert tag `{{label::LNG:de}}` outputs an empty string, because the `System::loadLanguageFile('languages')` no longer loads the correct files. I think it happened in this commit, where those files were removed:

https://github.com/contao/contao/commit/8f975da820a9dca0415d9da9b669953a41a502c2#diff-96f48168e3462d4b0437106abbc1dc5d8ec39a9aec76ad1da4a334e4a5ae9990

Could it be a BC break? Or is this new behavior intended?

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3712

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -12,6 +12,8 @@ namespace Contao;
 
 use Contao\CoreBundle\Controller\InsertTagsController;
 use Contao\CoreBundle\Image\Studio\FigureRenderer;
+use Contao\CoreBundle\Intl\Countries;
+use Contao\CoreBundle\Intl\Locales;
 use Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag\HtmlHeadBag;
 use Contao\CoreBundle\Routing\ResponseContext\ResponseContextAccessor;
 use Contao\CoreBundle\Util\LocaleUtil;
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -232,6 +234,36 @@ protected function doReplace($strBuffer, $blnCache)
 						break;
 					}
 
+					if ($keys[0] == 'LNG' && \count($keys) == 2)
+					{
+						trigger_deprecation('contao/core-bundle', '4.12', 'Using the label::LNG insert tag has been deprecated and will no longer work in Contao 5.0.');
```
Toflar: The alternative is missing :)

fritzmg:
Why do we want to deprecate the usage of `label::LNG` and `label::CNT`?

ausi:
> Why do we want to deprecate the usage of `label::LNG` and `label::CNT`?

Because they are no longer from the translation files.

> The alternative is missing :)

Should we maybe add seperate insert tags for the language and country services?

Toflar:
Well, we can only deprecate if there's an alternative suggestion. Otherwise it cannot be deprecated.

fritzmg:
> Because they are no longer from the translation files.

But `Contries::getCountries()` still loads them from the translator as far as I can see? And so does `Locales::getDisplayNames()`.

ausi:
> But `Contries::getCountries()` still loads them from the translator as far as I can see? And so does `Locales::getDisplayNames()`.

In order to be able to overwrite a country or language label the translation files can be used. But in the core they are no longer contained in `$GLOBALS['TL_LANG']` so I think explicit insert tags for country and languages would make the difference more clear.

fritzmg:
imho it is not necessary to introduce a new insert tag. While the default names are no longer stored in the translations, from the user's point of view nothing really changed as they can still access the names via `{{label::CNT:*}}`/`{{label::LNG:*}}` and they can also adjust them via Contao translations as usual. So from the user's point of view it's still a "label".

ausi:
> So from the user's point of view it's still a "label".

Removed deprecation in 0961714d9b9ef9324653581597d7d62782d4aa1a

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: Removed deprecation in 0961714d9b9ef9324653581597d7d62782d4aa1a

fritzmg (approved):

fritzmg:
There is one argument in favor of a separate insert tag though: we could use the `replaceInsertTags` hook and use a service with proper DI. However, I would much rather see the possibility within Contao to be able to override its hardcoded insert tags. The `replaceInsertTags` hook will not be executed for any "internal" insert tag and thus it is not possible to change any of them.

With a new hook or event in place to be able to adjust the replacement of _any_ insert tag we could introduce services just for `label::CNT` and `label::LNG` for example.

Toflar (approved):

leofeyer:
Thank you @ausi.

aschempp (approved):

PR #3717, opened 2021-11-23
Move the ECS configuration into the tools folder

leofeyer:
So we can remove the `contao/contao` specific configuration from `contao/easy-coding-standard`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
                   composer bin ecs install --no-interaction --no-progress
 
             - name: Run ECS
-              run: tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests --config tools/ecs/vendor/contao/easy-coding-standard/config/default.php --no-progress-bar --ansi
+              run: tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests --config tools/ecs/config/default.php --no-progress-bar --ansi
 
     ecs-legacy:
         name: ECS Legacy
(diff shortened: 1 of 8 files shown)


aschempp:
I use the default config in the bundle a lot in my projects 😬

bytehead:
I do as well, at least the templates.

```
"php-cs-fixer": [
    "@php vendor/bin/php-cs-fixer fix --dry-run -vvv --ansi",
    "@php vendor/bin/ecs check templates --config vendor/contao/easy-coding-standard/config/template.php --ansi"
],
```

bytehead:
I don't get the idea to remove the configs from the package? @leofeyer can you share your ideas behind?

leofeyer:
You should only use the `config/contao.php` configuration set and define your custom rules in your `ecs.php` file.

The `default.php` file has a lot of specific configuration such as references to the `core-bundle` and `manager-bundle`, which both do not exist in your installation, or a header comment fixer that says _This file is part of Contao_, which is also not true in your application. Why would you want to use this configuration in your own project?

bytehead:
I basically use the template configuration only. The other ones make not really sense, I'm there with you.

leofeyer:
Maybe we should provide a `config/contao-templates.php` configuration? @ausi WDYT?

ausi:
I think that would be great.

Using the following code will still work in the future, correct? https://github.com/ausi/slug-generator/blob/4ed3f8cec67d8ca136ca9473934f151cc860544d/ecs.php#L14

bytehead:
> Using the following code will still work in the future, correct? ausi/slug-generator@4ed3f8c/ecs.php#L14

Yes, this is still in the library available.

aschempp:
I'm actually using the `self.php`, not sure if that's still here: https://github.com/terminal42/contao-pageimage/blob/main/ecs.php

leofeyer:
That is as wrong as it can be. 😄 The `self.php` is the ECS file of the library itself, which is now the [`ecs.php`](https://github.com/contao/easy-coding-standard/blob/main/ecs.php) file in the project root.

PR #3719, opened 2021-11-23
Forward error handling to routing controller

aschempp:
This is a first step to get rid of our `PageErrorXXX` classes, since they are currently "hardcoded" to generate some error pages. We already have the `PrettyErrorScreenListener` that generates the page content. Instead of calling `PageError404()->getResponse()` this will not essentially call `FrontendIndex::renderPage()` with the given page type, which in turn might use the old page class. As soon as there is a real controller for an error page (possibly my next PR), this will make the page classes obsolete.


core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -531,6 +531,34 @@ public static function findFirstPublishedRegularByPid($intPid, array $arrOptions
 		return static::findOneBy($arrColumns, $intPid, $arrOptions);
 	}
 
+	/**
+	 * Find the first published regular page by its parent ID
```
Toflar: Regular?

Toflar (commented):
Looks very good to me! Just a question before I approve: Don't we need the old page types for BC? We can throw them away soon ;)

core-bundle/src/Resources/contao/models/PageModel.php
aschempp: fixed in dad9b92cf9e4ecdf0f44aef8ff65624953855320

aschempp:
> Don't we need the old page types for BC?

Not exactly sure what you mean? The controller for all legacy page types is `FrontendIndex::renderPage()`, which still calls the `TL_PTY` objects: https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/controllers/FrontendIndex.php#L295

Toflar (approved):

core-bundle/src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -132,32 +138,33 @@ private function renderErrorScreenByType(int $type, ExceptionEvent $event): void
 
         $processing = true;
 
-        if (null !== ($response = $this->getResponseFromPageHandler($type))) {
-            $event->setResponse($response);
-        }
-
-        $processing = false;
-    }
+        try {
```
leofeyer, line 141: The try block seems to be longer than it should be. Is it not just required around the `$this->httpKernel->handle()` method?

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -531,6 +531,34 @@ public static function findFirstPublishedRegularByPid($intPid, array $arrOptions
...
+	/**
+	 * Find the first published page by its type and parent ID
+	 *
+	 * @param string  $strType    The page type
+	 * @param integer $intPid     The parent page's ID
+	 * @param array   $arrOptions An optional options array
+	 *
+	 * @return PageModel|null The model or null if there is no published regular page
+	 */
+	public static function findFirstPublishedTypeByPid($strType, $intPid, array $arrOptions=array())
```
```suggestion
	public static function findFirstPublishedByTypeAndPid($strType, $intPid, array $arrOptions=array())
```

core-bundle/src/EventListener/PrettyErrorScreenListener.php
aschempp: changed in 2752e29fc707bce414fbb0e41a3f2b0980a30c75

leofeyer (approved):

PR #3720, opened 2021-11-24
Fix security permissions for custom backend paths

aschempp:
This is a follow-up to https://github.com/contao/contao/pull/3472 and should fix the issues mentioned in https://github.com/contao/contao/pull/3472#discussion_r708917236

Additionally found that the install tool was not using the custom backend route, and the lexik maintenance mode did not work when the back end path was configured.

- [x] Should we also update the `README` to reflect the changes to the system configuration?


Toflar (approved):

bytehead (approved):
Looks correct to me.  Apart from the changes in `*/**/Plugin.php`. Why that?

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -77,12 +77,12 @@ public function getBundles(ParserInterface $parser): array
     {
         $configs = [
             BundleConfig::create(FrameworkBundle::class),
-            BundleConfig::create(SecurityBundle::class)->setLoadAfter([FrameworkBundle::class]),
+            BundleConfig::create(SecurityBundle::class)->setLoadAfter([FrameworkBundle::class, ContaoCoreBundle::class]),
```
richardhj: I guess this fixes the issue that I encountered: "the service parameter '%contao.backend.route_prefix%' is not defined" 🙈

aschempp:
> Looks correct to me. Apart from the changes in `*/**/Plugin.php`. Why that?

You mean the order of loading / load after? This fixes https://github.com/contao/contao/pull/3472#discussion_r747242565 which is incorrect, the parameter was not available because of the loading order. As far as I can tell, it shouldn't matter for us.

richardhj (approved):

manager-bundle/src/ContaoManager/Plugin.php
aschempp: exactly 🙃

fritzmg (approved):

core-bundle/src/ContaoManager/Plugin.php
```diff
@@ -52,13 +50,11 @@ public function getBundles(ParserInterface $parser): array
                 ->setLoadAfter(
                     [
                         FrameworkBundle::class,
-                        SecurityBundle::class,
```
leofeyer: I am not sure about this change. The core bundle loads the SchebTwoFactorBundle, which decorates the authentication provider and checks for `security.authenticator.manager` when adding its listeners. How does this work when the security bundle is loaded after the core bundle?

leofeyer (changes requested):

core-bundle/src/ContaoManager/Plugin.php
aschempp: idk, but wouldn't that mean the `SchebTwoFactorBundle` needs `loadAfter` SecurityBundle and not the current statement?

aschempp:
I did rethink the problem, and the `PrependExtensionInterface` is exactly here for such problems, so thats how we can easily fix it without affecting the plugin order 😎

manager-bundle/src/Resources/skeleton/config/security.yml
```diff
@@ -51,7 +51,7 @@ security:
 
     # Backwards compatibility
```
richardhj, line 52: Remove this comment then?
```suggestion
```

aschempp: dunno what that is for actually 😂 @leofeyer ?

bytehead: Can be removed IMO, introduced in here: https://github.com/contao/contao/commit/024e0938ef7e94159cc03ba67a71f261cc0e03bc

richardhj:
I did add this line in the previous PR because I thought we don't need these rules anymore.

manager-bundle/src/Resources/skeleton/config/security.yml
richardhj: I did add this line in the previous PR because I thought we don't need these rules anymore.

leofeyer (approved):

PR #3722, opened 2021-11-25
Improve concept of replacing insert tags after parsing the template

ausi:
Alternative to #3622 because of https://github.com/contao/contao/pull/3693#issuecomment-977760667

@fritzmg @m-vo WDYT?

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
@@ -60,7 +60,7 @@ class ModuleCalendar extends Events
 			$objTemplate->link = $this->name;
 			$objTemplate->href = 'contao/main.php?do=themes&amp;table=tl_module&amp;act=edit&amp;id=' . $this->id;
 
-			return $objTemplate->parseWithInsertTags();
+			return $objTemplate->parseTemplate();
 		}
 
 		$this->cal_calendar = $this->sortOutProtected(StringUtil::deserialize($this->cal_calendar, true));
@@ -184,7 +184,7 @@ class ModuleCalendar extends Events
 		$objTemplate->weeks = $this->compileWeeks();
 		$objTemplate->substr = $GLOBALS['TL_LANG']['MSC']['dayShortLength'];
 
-		$this->Template->calendar = $objTemplate->parseWithInsertTags();
+		$this->Template->calendar = $objTemplate->parseTemplate();
 	}
 
 	/**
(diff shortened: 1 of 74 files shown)


leofeyer:
LGTM. 👍 I would probably call one method `parse()` and the other one `parseWithInsertTags()` or `parseWithoutInsertTags()` depending on what the default behavior shall be.

m-vo:
Maybe `parseWithInsertTags` could be named `parseAndReplaceInsertTags()`? So that *with* cannot be interpreted as "leaving them in". :thinking:

ausi:
I discussed this with @fritzmg on Slack and we proably should fix some things before we can replace in insert-tags in the templates directly:

1. Fix BC for `outputFrontendTemplate` (insert tags need to be non-replaced there) and deprecate it (I think)
2. Create an alternative way to overwrite an insert tag from the core (this is currently possible by hacking `outputFrontendTemplate`)
3. Fix BC for not replacing insert tags in the backend

m-vo:
> Create an alternative way to overwrite an insert tag from the core (this is currently possible by hacking outputFrontendTemplate)

I'm all :+1: for allowing insert tags to be overwritten (e.g. with insert tags as a service + priority tagging) but I would not consider removing the ability for this sort of hack a BC break. :smile:

fritzmg:
imho for now we should revert #3622. This will fix #3693 and not introduce any BC breaks. In Contao 5.0 we can then move insert tag replacement into `parse`/`inherit` again - assuming we introduce the changes mentioned by @ausi to `4.13`.

ausi:
> imho for now we should revert #3622. This will fix #3693 and not introduce any BC breaks.

I’m not sure if that is neccesary. The current version of this PR tries to keep BC and thus should solve point 1 and 3.

aschempp:
This PR looks reealy strange to me. It deprecated a method every developer uses since forever just because we change(d) the default behaviour of our PHP template system without benefit (not speaking of the Twig benefits!) even though that template system is supposed to be replaced anyway?

fritzmg:
> I’m not sure if that is neccesary. The current version of this PR tries to keep BC and thus should solve point 1 and 3.

There is still another BC issue. Previously insert tags were replaced after everything was already processed. Some insert tags might rely on that, i.e. an insert tag might rely on the global page model to have the updated information that a news reader for example sets (`fritzmg/contao-sharebuttons` relies on that).

ausi:
> imho for now we should revert #3622.

I guess you are right ☺️

So we would change this behavior in Contao 5.0, but is there a way to make the transition more convenient for developers?

fritzmg:
May be this is something we can only document in the `UPGRADE.md`? And may be also announce it via different channels.

leofeyer:
So the plan is to revert the changes for now and to discuss how to implement https://github.com/contao/contao/pull/3722#issuecomment-979301369 before we apply the changes again?

ausi:
> So the plan is to revert the changes for now and to discuss how to implement [#3722 (comment)](https://github.com/contao/contao/pull/3722#issuecomment-979301369) before we apply the changes again?

I think so, yes, we need to deprecate `outputFrontendTemplate` and:
> Create an alternative way to overwrite an insert tag from the core (this is currently possible by hacking `outputFrontendTemplate`)

And apply #3622 only in Contao 5.0.

@m-vo WDYT?

ausi:
This can be closed because of #4450
right?

PR #3723, opened 2021-11-25
Stop using the LegacyEventDispatcherProxy class

### LINKED ISSUE #2862: Deprecated classes in Symfony 5.1

leofeyer:
Partly fixes #2862.

### DIFF
--- a/manager-bundle/src/HttpKernel/ContaoCache.php
+++ b/manager-bundle/src/HttpKernel/ContaoCache.php
@@ -52,11 +52,7 @@ class ContaoCache extends HttpCache implements CacheInvalidation
 
     public function getEventDispatcher(): EventDispatcher
     {
-        if (!$this->eventDispatcher) {
-            $this->eventDispatcher = new EventDispatcher();
-        }
-
-        return $this->eventDispatcher;
+        return $this->eventDispatcher ??= new EventDispatcher();
     }
 
     public function fetch(Request $request, $catch = false): Response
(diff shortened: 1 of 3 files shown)


ausi (approved):

manager-bundle/src/HttpKernel/ContaoCache.php
```diff
@@ -49,6 +50,15 @@ public function __construct(ContaoKernel $kernel, string $cacheDir = null)
         $this->addSubscriber(new CleanupCacheTagsListener());
     }
 
+    public function getEventDispatcher(): EventDispatcher
+    {
+        if (!$this->eventDispatcher) {
+            $this->eventDispatcher = new EventDispatcher();
+        }
+
+        return $this->eventDispatcher;
```
m-vo: With PHP>=7.4 this could be shortened to:

```suggestion
        return $this->eventDispatcher ??= new EventDispatcher();
```

:upside_down_face:

m-vo (approved):

PR #3724, opened 2021-11-25
Add the missing @throws tags

leofeyer:
This is a follow-up on https://github.com/Toflar/contao/pull/11#discussion_r755230151 that aims to clarify how the changes would look like, so we can agree on a concept. This is my PHPStorm config:

Note that I have added `\ReflectionException` as unchecked exception, which might be wrong. 🤔 See https://phpstan.org/blog/bring-your-exceptions-under-control#checked-vs.-unchecked-exceptions

### DIFF
--- a/core-bundle/src/Controller/BackendCsvImportController.php
+++ b/core-bundle/src/Controller/BackendCsvImportController.php
@@ -20,6 +20,7 @@ use Contao\DataContainer;
 use Contao\FileUpload;
 use Contao\Message;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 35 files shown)


m-vo:
I appreciate the effort but I think most of the exceptions from this PR should not be annotated with the `@throws` tag. 

Take the `FigureBuilder` class for instance. Noone should ever catch the `ContainerException` or `NotFoundExceptionInterface`. If these exceptions get thrown they indicate a programmer error, so they should bubble up and make the code fail. The `build()` method on the other side will throw an `InvalidResourceException` if a resource isn't present - something a back end user could have configured or due to a file that was deleted in the meantime. This exception should IMHO be annotated. And in fact - it already is :stuck_out_tongue:: https://github.com/contao/contao/blob/4.x/core-bundle/src/Image/Studio/FigureBuilder.php#L493

leofeyer:
> Noone should ever catch the `ContainerException` or `NotFoundExceptionInterface`.

I agree. That's why I have created the PR, so we can discuss which exceptions to add to the list of unchecked exceptions.

leofeyer:
I am closing this in favor of #3816.

PR #3725, opened 2021-11-26
PHP 8.1

Toflar:
PHP 8.1.0 has been released so it's time to run our tests on 8.1 as well and switch to 8.2 for nightly build tests.

There's quite a bit of work also in dependencies to be done but this PR is here to track the progress :)


m-vo (dismissed):
Yei! 

I don't understand why the 8.1 job failed resolving dependencies while it worked on 8.2, though.

bytehead (dismissed):

Toflar:
> I don't understand why the 8.1 job failed resolving dependencies while it worked on 8.2, though.

Because we run the nightly ones with `--ignore-platform-req=php`.

There are quite a few dependencies that need to be updated before we can get compatible with 8.1 ;)

Toflar:
Dependencies have been updated. We're ready :)

core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -22,6 +22,11 @@
  */
 class PageRegular extends Frontend
 {
+	/**
+	 * @var FrontendTemplate
+	 */
+	protected $Template;
```
leofeyer: This property is already defined here:

https://github.com/contao/contao/blob/f842b78a8f5a78d0dd16aacc3528e84134e18f4e/core-bundle/src/Resources/contao/library/Contao/System.php#L54

Why did you add it?

fritzmg (dismissed):

core-bundle/src/Resources/contao/pages/PageRegular.php
Toflar: Not in 4.9...

fritzmg: It's not actually defined. The docblock is wrong, as not everyone inheriting from `Contao\System` will automatically have `$this->Template`. `PageRegular` is the odd one out here, because in all other classes where the legacy framework uses `$this->Template` it is covered by `__get` and `__set` there and stored in `$this->arrData`. But that's not the case here, thus it needs to be defined, otherwise PHP 8.2 will cause an error.

leofeyer: > Not in 4.9...

The code snippet above shows the `4.9` branch.

Toflar: Well, I don't know. For PHP 8.2 it's not there.

leofeyer: > `PageRegular` is the odd one out here, because in all other classes where the legacy framework uses `$this->Template` it is covered by `__get` and `__set` and stored in `$this->arrData`. But that's not the case here

@fritzmg I don't understand. PageRegular does not define its own getters or setters, therefore it should inherit them from the parent class.

fritzmg: It's not present in any Contao version. It just does not raise a deprecation prior to PHP 8.2.

Toflar: It's just a comment. The property needs to exist.

leofeyer: What is the exact error message?

fritzmg: > @fritzmg I don't understand. PageRegular does not define its own getters or setters, therefore it should inherit them from the parent class.

The parent class does not provide them either.

> What is the exact error message?

https://github.com/contao/contao/runs/4521842534

```
Creation of dynamic property Contao\PageRegular::$Template is deprecated
```

Toflar: https://wiki.php.net/rfc/deprecate_dynamic_properties

leofeyer: Yes, but that is not what we are doing. We are using magic getters and setters to add objects to `$this->arrObjects`.

https://github.com/contao/contao/blob/f842b78a8f5a78d0dd16aacc3528e84134e18f4e/core-bundle/src/Resources/contao/library/Contao/System.php#L133-L149

And this should continue to work in PHP 8.2 as well. So maybe we are looking at a PHP 8.2 bug here?

fritzmg: > We are using magic getters and setters to add objects to $this->arrObjects.

As already said: _not_ in `PageRegular`. There is no `__set` to handle `$this->Template = …` in `PageRegular`. That's why this issue occurs.

Toflar: Yeah but we create a new one, so that would be `__set()`, not `__get()`.

Toflar: With `__set()` it works properly. See https://3v4l.org/akWau/rfc#vgit.master.

leofeyer: Yes, we should fix this in the `System` class. If I search for `$this->Template =` in PhpStorm, there are more than 20 matches that will all trigger the same warning in PHP 8.2 otherwise.

Toflar: I don't think that's a good idea. We want them to be local to the class and not stored away in some random inheritance object.

fritzmg: >  I search for $this->Template = in PhpStorm, there are more than 20 matches that will all trigger the same warning in PHP 8.2 otherwise.

That's not the case. All other classes that use `$this->Template = …` already have `__set`. It's just `PageRegular` that does not.

leofeyer: PR is here: #3822

leofeyer (approved):

bytehead (approved):

leofeyer:
Thank you @Toflar.

PR #3726, opened 2021-11-26
Unlock PHPStan level 6

leofeyer:
@m-vo @ausi I am not sure how to fix the remaining 10 issues:

Any idea?


ausi:
> Any idea?

08fcf1947b6d5c121664252ec9d37fa7dc94b1c2 ☺️

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -351,6 +351,8 @@ public function setLocale(?string $locale): self
      *
      * Set the value to null to remove it. If you want to explicitly remove an
      * auto-generated value from the results, set the $forceRemove flag to true.
+     *
+     * @param string|int|bool $forceRemove
```
ausi:
```suggestion
     * @param bool $forceRemove
```
why not `bool` only?

leofeyer: Changed in 2e46f7a8ef9084c1ecd417917cf7717745d76c99.

core-bundle/tests/Twig/Interop/ContaoEscaperTest.php
```diff
@@ -71,6 +73,8 @@ public function provideHtmlInput(): \Generator
     }
 
     /**
+     * @param string|int $input
+     *
      * @dataProvider provideHtmlAttributeInput
      */
     public function testEscapesHtmlAttributes($input, string $expectedOutput): void
```
ausi: Many of these can be changed I think.

```suggestion
     * @dataProvider provideHtmlAttributeInput
     */
    public function testEscapesHtmlAttributes(string $input, string $expectedOutput): void
```

leofeyer: Changed in 4856530db904f2246a3fe79850e1cdf76217efa9. However, the others in the class cannot be changed, because `provideHtmlInput()` returns `123` in line 60.

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -117,7 +117,7 @@ private function authenticatePreview(Request $request): void
         $frontendUsername = $this->tokenChecker->getFrontendUsername();
 
         if ($this->security->isGranted('ROLE_ALLOWED_TO_SWITCH_MEMBER')) {
-            $frontendUsername = $request->request->get('user');
+            $frontendUsername = (string) $request->request->get('user');
```
ausi: Should we change the check from 5 lines below from `if(null !== $frontendUsername)` to `if($frontendUsername)` ?

core-bundle/src/Picker/AbstractTablePickerProvider.php
```diff
@@ -160,7 +160,7 @@ public function getDcaAttributes(PickerConfig $config): array
         return $attributes;
     }
 
-    public function convertDcaValue(PickerConfig $config, $value)
+    public function convertDcaValue(PickerConfig $config, $value): int
```
ausi: Isn’t this a BC break because every class that extends this abstract class has to add the return type now too?

ausi (commented):
I created a PR with some changes I noticed while reviewing: https://github.com/leofeyer/contao/pull/3

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: Fixed in 5cef696c5265efc21944489b2ae81d9a481f6d48.

core-bundle/src/Picker/AbstractTablePickerProvider.php
leofeyer: Good catch! Reverted in 5cef696c5265efc21944489b2ae81d9a481f6d48.

ausi (approved):

PR #3727, opened 2021-11-26
Link parent elements in the back end breadcrumb trail

Toflar:
Here we go, yet another user feature :-)

Explanation is pretty simple.
Before:

Bildschirmfoto 2021-11-26 um 14 39 51

After:

Bildschirmfoto 2021-11-26 um 14 39 59


bytehead (dismissed):
I like a lot!

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -537,19 +539,17 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 							$trail[] = ' <span>' . $GLOBALS['TL_LANG']['MOD'][$table] . '</span>';
 						}
 
+						$strUrl = $container->get('router')->generate('contao_backend', array
+						(
+							'do' => $container->get('request_stack')->getCurrentRequest()->query->get('do'),
+							'table' => $table,
+							'id' => $objRow->id,
+							'ref' => $container->get('request_stack')->getCurrentRequest()->attributes->get('_contao_referer_id'),
```
aschempp: isnt the referrer ID supposed to change on every request?

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -537,19 +539,17 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 							$trail[] = ' <span>' . $GLOBALS['TL_LANG']['MOD'][$table] . '</span>';
 						}
 
+						$strUrl = $container->get('router')->generate('contao_backend', array
+						(
+							'do' => $container->get('request_stack')->getCurrentRequest()->query->get('do'),
+							'table' => $table,
```
aschempp: how come you're using `$table` here but the loop uses `$ptable`? looks like this won't work with more than one level?

core-bundle/src/Resources/contao/classes/Backend.php
Toflar: Yes, the request attribut contains the new one :)

core-bundle/src/Resources/contao/classes/Backend.php
Toflar: Hmm yeah depends on the mode I guess. When editing a content element, `tl_content` is actually correct 🤔

Toflar: For the article case, it is correct. I don't even know if we have anywhere where we go deeper than that?

Toflar: `$table` is actually correct :) If you edit the content element of an event, it will look like this now:

Bildschirmfoto 2021-11-26 um 16 13 23

* `Contao North-Day` links to `contao?do=calendar&table=tl_content&id=6.`
* `Contao Public Example Events` links to `contao?do=calendar&table=tl_calendar_events&id=3`

fritzmg (dismissed):
I tested and I like 👍

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -537,19 +539,17 @@ protected function getBackendModule($module, PickerInterface $picker = null)
...
-						}
-						elseif ($objRow->name)
-						{
-							$trail[] = ' <span>' . $objRow->name . '</span>';
-						}
-						elseif ($objRow->headline)
-						{
-							$trail[] = ' <span>' . $objRow->headline . '</span>';
-						}
+						$trail[] = sprintf(' <span><a href="%s">%s</a></span>', $strUrl, ($objRow->title ?? $objRow->name ?? $objRow->headline));
```
m-vo: Will this always work or should we use `?:` instead of `??`?

m-vo (approved):
Noice.

core-bundle/src/Resources/contao/themes/flexible/main.css
```diff
@@ -355,6 +355,9 @@ body:not(.fullscreen) #container {
 	text-overflow:ellipsis;
 	line-height:22px;
 }
+#main_headline span a {
```
richardhj:
```suggestion
#main_headline span a:hover {
```

I find the underline too obtrusive.

Toflar: I would like to leave that to @leofeyer because there surely was an idea behind how links are formatted in the back end and I personally don't care anyway :)

leofeyer: <del>I agree with the suggested change.</del>

Turns out I do not agree eventually. 🙈

core-bundle/src/Resources/contao/classes/Backend.php
Toflar: Ah right, we probably should also ignore empty values. Fixed in 1fa1fc05f8bdbe136fd99c9a4f8a132e9c94782f.

core-bundle/src/Resources/contao/themes/flexible/main.css
Toflar: Looks nicer indeed. PR adjusted.

m-vo (dismissed):

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -537,19 +539,17 @@ protected function getBackendModule($module, PickerInterface $picker = null)
...
-						}
-						elseif ($objRow->name)
-						{
-							$trail[] = ' <span>' . $objRow->name . '</span>';
-						}
-						elseif ($objRow->headline)
-						{
-							$trail[] = ' <span>' . $objRow->headline . '</span>';
-						}
+						$trail[] = sprintf(' <span><a href="%s">%s</a></span>', $strUrl, ($objRow->title ?: $objRow->name ?: $objRow->headline));
```
aschempp: This previously did not add a breadcrumb if the record neither had a title, name or headline. I don't think the current behaviour is correct, maybe we should fall back to the ID or not change the behaviour?

aschempp (changes requested):

core-bundle/src/Resources/contao/classes/Backend.php
Toflar: I think not showing anything is an issue. Actually the field should be configurable and just fall back to one of those three but I wouldn't know where to put that config in the DCA. And then fall back to the ID maybe, yes?

Toflar: But it could still be merged as is because ignoring it like we did for years cannot be the desired behaviour? :D

leofeyer: I am also not sure about this. Even though it might be a bug, I would prefer if we did not change it now. 🙈

Toflar: Adjusted in https://github.com/contao/contao/pull/3727/commits/1799660d1fbdc79c7b9bf88ed092916bf33e653c then.

aschempp (dismissed):

leofeyer (dismissed):

core-bundle/src/Resources/contao/classes/Backend.php
leofeyer: Excellent solution. 👍

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3729, opened 2021-11-26
Maintenance mode per root page

### LINKED ISSUE #1705: Maintenance mode with Multi Domain
I wish the Maintenance mode could be configured for a multi domain environment -
so per domain. ..

aschempp:
With this feature, a user can put a root page into maintenance mode. It either generates a 503 page on that root page (meaning one can style the maintenance message from the back end) or falls back to our existing maintenance template.

**About the _maintenance login module_**
The _maintenance login module_ allows to bypass the maintenance mode. This can be useful e.g. to allow a client to preview the page under maintenance. It uses our JWT cookie (which we also use to enable debug mode), therefore the module is part of the `contao/manager-bundle`.
The module intentionally does not output anything if the maintenance mode is not enabled. It does not only allow to login if maintenance is enabled, it also allows to logout of the "bypass maintenance". This means it can be placed anywhere in the layout at all times and will only show the login&logout option once the maintenance mode is enabled.

**Questions/Ideas:**
 - [ ] Not sure the current checkbox in `tl_page` is at its ideal place?

This PR intentionally does not remove `lexik/maintenance-bundle` since 1. it works on the command line and 2. puts the whole system in maintenance. Both might be used e.g. for deployments.

**TODOs:**
 - [x] bypass maintenance in frontend preview mode
 - [x] change the listener priority to be after the firewall, since we need permission checks (also for frontend preview)
 - [x] Set `path=false` on `@Page` once https://github.com/contao/contao/pull/3605 is merged
 - [x] Add a _bypass maintenance login module_ that can be placed on the 503 page.


leofeyer:
> This PR intentionally does not remove `lexik/maintenance-bundle`

I wish it would, though. 😞

Toflar:
The complementary PR will follow next week. It's already done on my local machine ;)

Toflar:
> * should maintenance be bypassed in frontend preview mode?

I think that's an important requirement yes. That's exactly what it's meant for. You should be able to put the site into maintenance mode, work on it and then publish, like we do now with our lexik implementation. However, what's new is that we can do this root page based 👍

fritzmg (commented):
I am not sure about the changes in `PageModel`, i.e. hard coding the error page type, even though it is implemented as a Page Controller. If your App or an extension implements its own non-routeable Page Controller, it will not be included in that list.

aschempp:
thats not really related to this PR though

fritzmg (dismissed):

Toflar (changes requested):
Only the route flag is missing here imho :)

aschempp:
FYI I have moved the `MSC.maintenanceMode` label into the manager-bundle in 238ccd06fd5f632047970b7a44e5f50bed34f806 since it will be removed in core-bundle in #3737

core-bundle/src/EventListener/ServiceUnavailableListener.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        $pageModel->loadDetails();
+
+        if ($pageModel->maintenanceMode) {
+            throw new ServiceUnavailableException(sprintf('Domain %s is in maintenance mode', $pageModel->dns));
+        }
+    }
+
+    private function isDisabledByJwt(Request $request): bool
```
Toflar: My problem with this (yes, this also applies to other usages of the `JwtManager` is that if you do not use the `manager-bundle` you cannot use this whole listener. And if @bytehead wanted to provide some custom way to circumvent the maintenance mode, they would have to duplicate the whole listener.
The alternative would be to adjust this listener to check for a certain request attribute and a separate listener with a higher priority in the `manager-bundle` that does set this attribute. Wdyt?

aschempp: Well the same applies to debug mode, does it not? I mean the whole "bypass maintenance through login" will not work anyway because there's no way to set a JWT?

Toflar: Exactly, that's what I'm saying. This should be changed as well. If @bytehead wants to provide this logic, they have absolutely no way of achieving this. Because the `JwtManager` is a concrete class coming from the `manager-bundle` which they cannot use because it does all the ME stuff they don't need.

Toflar: We could also simply dispatch an event 🤔 Then we could use that in the manager-bundle based on the JWT and custom implementation can be easily added. Like...dunno...disable it when logged in using Shibboleth. But then again, we can also implement that once someone needs it.

bytehead: > disable it when logged in 

Right now you won't see the 503 when logged in (backend).

I don't get it, why you check by the JWT token instead if there is a logged in backend user to skip the maintenance mode?

aschempp: The idea is that a client can login in the front end on the maintenance page to bypass it. I don't think this needs to be available without a Managed Edition.

bytehead: 🤷‍♂️ But login in to the backend means still bypass the maintenance page?

aschempp: correct, it is disabled in preview mode (`preview.php`). Not in the regular front end entry point even if logged in to the back end. Not sure if your Symfony installations have a preview endpoint?

Toflar (dismissed):

core-bundle/src/EventListener/ServiceUnavailableListener.php
bytehead: > Not in the regular front end entry point even if logged in to the back end

But this is the case right now in 4.9 AFAIK.

> Not sure if your Symfony installations have a preview endpoint?

We do have a preview endpoint 👍

core-bundle/src/Controller/Page/Error503PageController.php
```diff
@@ -0,0 +1,43 @@
...
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\ServiceAnnotation\Page;
+use Contao\FrontendIndex;
+use Contao\PageModel;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @Page("error_503", path=false)
+ */
+class Error503PageController extends AbstractController
```
leofeyer: The class has essentially the same code as the `Error4xxPageController` class, except the `supportsContentComposition()` method which is missing here IMHO. Do we still need a separate class?

core-bundle/src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -107,7 +107,9 @@ private function handleException(ExceptionEvent $event): void
                 break;
 
             case $exception instanceof ServiceUnavailableHttpException:
-                $this->renderTemplate('service_unavailable', 503, $event);
+                if (!$this->renderErrorScreenByType(503, $event)) {
+                    $this->renderTemplate('service_unavailable', 503, $event);
```
leofeyer, line 113: We should add a deprecation here, so we can change this to `$this->renderErrorScreenByType(503, $event);` in Contao 5. WDYT?

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -593,6 +593,12 @@
       <trans-unit id="PTY.error_404.1">
         <source>If a visitor requests a non-existent page, a 404 error page will be loaded instead.</source>
       </trans-unit>
+      <trans-unit id="PTY.error_503.0">
+        <source>503 Service unavailable</source>
+      </trans-unit>
+      <trans-unit id="PTY.error_503.1">
+        <source>If maintenance mode is enabled in the root page, a 503 error page will be show to all requests.</source>
```
```suggestion
        <source>If the maintenance mode is enabled in the root page, a 503 error page will be shown for all requests.</source>
```

core-bundle/src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -38,6 +38,12 @@
       <trans-unit id="tl_page.robots.1">
         <source>Here you can define how search engines handle the page.</source>
       </trans-unit>
+      <trans-unit id="tl_page.maintenanceMode.0">
+        <source>Enable maintenance mode</source>
+      </trans-unit>
+      <trans-unit id="tl_page.maintenanceMode.1">
+        <source>Shows a page to all visitors that this website is currently under maintenance.</source>
```
leofeyer:
```suggestion
        <source>Show visitors a page that this website is currently being maintained.</source>
```

leofeyer:
Why do we need the username/password authentication? I figured we could bypass the maintenance mode in the back end preview.

core-bundle/src/Controller/Page/Error503PageController.php
aschempp: Yes, because this one always supports content composition. We could check for the type though.

core-bundle/src/EventListener/PrettyErrorScreenListener.php
aschempp: I don't understand what you mean? We do want to show the template if there is no page, or you mean we should just show the generic Symfony 503 in that case?

aschempp:
The username/password authentication is a new feature that does NOT require a back end login to bypass the maintenance mode. Which is useful for a client to view a page. It basically replaces the need for https://github.com/terminal42/contao-root-protection

core-bundle/src/EventListener/PrettyErrorScreenListener.php
leofeyer: Exactly. Just like we do for the 4xx pages, too. You either have a page in the site structure or the generic Symfony 503.

aschempp: Hmm, I don't think thats best. We'll need the maintenance template for #3737 anyway, and as long as I do not have multiple root pages I could just reuse that template in both cases.

Toflar: I would also leave that as is :)

core-bundle/src/Controller/Page/Error503PageController.php
aschempp: Changed in dd61a4f67876ce024a61eeb319215623e17117e2

leofeyer:
I'm not sure I like this. A client who does not have back end access? That sounds like a rare case that would be better solved in an extension IMHO.

aschempp:
It's the case I have with every major client. Their people (regular users/management that needs to review a new page) never have backend access, and explaining them how to use the frontend preview would be cumbersome. Anyway, the module is not mandatory on a 503 page, its just a feature you can use if you need to, but I agree most people will not and that won't harm 😊

Toflar:
Agreed. I have thise use case for almost every bigger project. It's a "decision maker" thing. They don't want to fiddle around with preview and user accounts and back end access. They just want to give their final approval before anything goes live. They'll never edit anything themselves. This is a killer feature you are free to use or not :)

bytehead:
I see the case for this as well. But is this somehow possible to provide in an non-managed edition?

core-bundle/src/EventListener/ServiceUnavailableListener.php
aschempp: I noticed this is actually available in any Contao version through the `_bypass_maintenance` flag.

aschempp:
> But is this somehow possible to provide in an non-managed edition?

Well you cannot use the login module since it requires the `JwtManager`, which only works in the `ContaoKernel` of the Managed Edition. You _can_ use a listener to set `_bypass_maintenance`, e.g. depending on a client's IP address.

We _might_ be able to refactor the JwtManager to work in the core bundle (except for the debug mode), but I'm not sure that's worth the effort and might be a BC break 🤷

fritzmg:
We have the same use case - but also for the preview, not just the maintenance mode, i.e. not just for final approval before the whole site goes live. New content in an existing, running page also needs to be approved and our clients typically just create published pages for that which are only hidden from the menu - instead of using the front end preview to show hidden content, because not all the decision makers have a back end login.

aschempp:
> We have the same use case - but also for the preview, not just the maintenance mode, i.e. not just for final approval before the whole site goes live. New content in an existing, running page also needs to be approved and our clients typically just create published pages for that which are only hidden from the menu - instead of using the front end preview to show hidden content, because not all the decision makers have a back end login.

not sure how that's related here?

fritzmg:
> not sure how that's related here?

If we have a feature in place to allow access to otherwise unavailable content (e.g. maintenance in this case) with out a back end login - then that  could may be also be used for the same use case in a different context. Hence: yes it's related and an argument in favor of the system proposed here.

leofeyer:
> They just want to give their final approval before anything goes live. 

> Their people (regular users/management that needs to review a new page) never have backend access, and explaining them how to use the frontend preview would be cumbersome.

So how do they approve unpublished pages without the front end preview?

aschempp:
> So how do they approve unpublished pages without the front end preview?

We currently don't… I guess we could implement something similar, but I'm not sure that related?

leofeyer:
Both seems unrelated to me. Can you please move the contao-root-protection related changes into a separate PR so we can discuss them and extend them with a solution for the FE preview if applicable?

aschempp:
I have removed the maintenance login feature since it's controversial and I already have other ideas for it 😊

core-bundle/src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -107,7 +107,9 @@ private function handleException(ExceptionEvent $event): void
...
+                }
```
leofeyer:
```suggestion
                $this->renderErrorScreenByType(503, $event);

                if (!$event->hasResponse()) {
                    $this->renderTemplate('service_unavailable', 503, $event);
                }
```

This way we do not have to add a return value to the `renderErrorScreenByType()` method.

leofeyer (changes requested):
This PR is almost ready to be merged, however, I have noticed an odd "error" option in the page type drop-down menu that has not been there before:



@aschempp Any idea where that is coming from?

aschempp:
>  I have noticed an odd "error" option in the page type drop-down menu that has not been there before:
> 
> 
> 
> @aschempp Any idea where that is coming from?

Jup, that's coming from here: https://github.com/contao/contao/pull/3846#issuecomment-999546178

leofeyer (approved):

PR #3732, opened 2021-11-27
Disable form autocomplete for preview bar

fritzmg:
The browser's login autofill functionality currently may prevent the preview bar from working properly, if the browser happens to fill in the `name="user"` input field with a username which is not actually available (anymore). This was introduced in Contao 4.9.14 (https://github.com/contao/contao/commit/2e762a3eee71ce6076d51cb3fd64cfa0b6ce6758).

**Reproduction**

1. Create a new member.
2. Login with that member in the front end and save the login data in your browser (e.g. Firefox).
3. Delete the member.
4. Log into the back end and open the front end preview. 
  » The browser should autofill the username field in the preview bar with the previously used username.
5. Select _show_ for _Unpublished_ and click apply.

Applying the _Unpublished_ change will fail silently since a username of a front end member was submitted that does not actually exist. To work around this you always have to manually delete the autofilled member first currently.

This PR fixes that by setting `autocomplete="off"` on the preview bar form. Only removing the `autocomplete` attribute from the `input` field would also work, though I think it's better to specifically disable it.

### DIFF
--- a/core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
+++ b/core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
@@ -3,14 +3,14 @@
     <a href="#" title="{{ 'MSC.openToolbar'|trans }}">{{ include('@ContaoCore/Collector/contao.svg') }}</a>
 </div>
 <div class="cto-toolbar__inside">
-    <form action="{{ action }}" method="post">
+    <form action="{{ action }}" method="post" autocomplete="off">
         <div class="formbody">
             <input type="hidden" name="FORM_SUBMIT" value="tl_switch">
             <input type="hidden" name="REQUEST_TOKEN" value="{{ request_token }}">
             {% if canSwitchUser %}
                 <div>
                     <label for="ctrl_user">{{ 'MSC.feUser'|trans }}:</label>
-                    <input type="text" name="user" id="ctrl_user" list="userlist" class="tl_text user" placeholder="{{ 'MSC.username'|trans }}" value="{{ user }}" autocomplete="username">
+                    <input type="text" name="user" id="ctrl_user" list="userlist" class="tl_text user" placeholder="{{ 'MSC.username'|trans }}" value="{{ user }}">
                     <datalist id="userlist"></datalist>
                 </div>
             {% endif %}


ausi (approved):

aschempp (commented):
shouldn't we also handle the invalid username or at least not fail to toggle unpublished changes?

core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
```diff
@@ -3,14 +3,14 @@
     <a href="#" title="{{ 'MSC.openToolbar'|trans }}">{{ include('@ContaoCore/Collector/contao.svg') }}</a>
 </div>
 <div class="cto-toolbar__inside">
-    <form action="{{ action }}" method="post">
+    <form action="{{ action }}" method="post" autocomplete="off">
```
leofeyer, line 6: Did you test that the datalist autosuggest still works this way?

fritzmg: The auto suggest currently does not work in Firefox at all (#3733). But if I both add `autocomplete="off"` to the `<form>` _and_ remove `placeholder` from the input, it works, yes.

fritzmg:
> shouldn't we also handle the invalid username or at least not fail to toggle unpublished changes?

Indeed. I made a few simple changes so that it will now show a JS alert box with an error message, if the preview settings could not be applied:

image

leofeyer (changes requested):
I would prefer to only fix the bug in this PR and to add the new features in the `4.x` branch. Especially since there are new strings to be translated.

fritzmg:
I'd consider the silent failure the actual bug. But if you insist I can make a new PR for `4.x` of course. I would however still fix the tests in this PR. The changes revealed that some of them were wrong and didn't actually test what they were supposed to.

leofeyer:
But the fixes in the unit tests are not related to this PR, are they?

fritzmg:
> But the fixes in the unit tests are not related to this PR, are they?

They are. The changes to `BackendPreviewSwitchController::authenticatePreview` make them fail - and revealed that some things weren't actually tested properly.

leofeyer:
I don't understand. To me, the PR was complete after the fist commit (81bcc641e1af3a96f62afa775e9807f30c7b5c5d). Anything after that should be a separate PR – or even two separate PRs, one bugfix for Contao 4.9 and a new feature for Contao 4.13.

fritzmg:
> To me, the PR was complete after the fist commit ([81bcc64](https://github.com/contao/contao/commit/81bcc641e1af3a96f62afa775e9807f30c7b5c5d)).

Changing only that would still leave the problem that applying the preview settings fails silently, if you type in a member username that does not exist. imho this should also be fixed in `4.9`, but if we cannot add new translation strings in a bugfix release, then I guess it can only go into `4.13` (or we leave it English in 4.9, which would still be better than nothing).

leofeyer:
> Changing only that would still leave the problem that applying the preview settings fails silently, if you type in a member username that does not exist

I know, that is why I have said _Anything after that should be a separate PR – or even two separate PRs, one bugfix for Contao 4.9 and a new feature for Contao 4.13_.

Fixing one problem per PR will make it easier for everyone to review and merge changes.

fritzmg:
👍 done

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3734, opened 2021-11-27
Delete uploaded files if a form does not validate

### LINKED ISSUE #1185: Residual file via frontend upload when server-side form validation fails because of other field

fritzmg:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #1185
| Docs PR or issue | -

This is an alternative, less invasive approach to #3600 in order to fix #1185. Instead of introducing a new mechanic for the legacy framework, this PR instead fixes the specific, original problem by making sure the `Form` removes any uploaded files, if the form itself did not validate. `Form` already purges `$_SESSION['FILES']` when the form did not validate here:

https://github.com/contao/contao/blob/ba5f1ba905f54918641267f2e90965d3d1153df0/core-bundle/src/Resources/contao/forms/Form.php#L262-L271

so the `Form` might as well also properly delete the uploads.

### DIFF
--- a/core-bundle/src/Resources/contao/forms/Form.php
+++ b/core-bundle/src/Resources/contao/forms/Form.php
@@ -259,6 +259,30 @@ class Form extends Hybrid
 			$this->processFormData($arrSubmitted, $arrLabels, $arrFields);
 		}
 
+		// Remove any uploads, if form did not validate (#1185)
+		if ($doNotSubmit && $hasUpload && !empty($_SESSION['FILES']))
+		{
+			foreach ($_SESSION['FILES'] as $field => $upload)
+			{
+				if (empty($arrFields[$field]))
+				{
+					continue;
+				}
+
+				if (!empty($upload['uuid']) && null !== ($file = FilesModel::findById($upload['uuid'])))
+				{
+					$file->delete();
+				}
+
+				if (is_file($upload['tmp_name']))
+				{
+					unlink($upload['tmp_name']);
+				}
+
+				unset($_SESSION['FILES'][$field]);
+			}
+		}
+
 		// Add a warning to the page title
 		if ($doNotSubmit && !Environment::get('isAjaxRequest'))
 		{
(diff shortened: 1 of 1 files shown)


aschempp (approved):
I like that a lot more than the other PR. It will automatically work for any upload widget without requiring adjustments (e.g. our fineuploader I would assume).

leofeyer:
> I like that a lot more than the other PR.

So do I.

core-bundle/src/Resources/contao/forms/Form.php
```diff
@@ -259,6 +259,30 @@ protected function compile()
 			$this->processFormData($arrSubmitted, $arrLabels, $arrFields);
 		}
 
+		// Remove any uploads, if form did not validate (#1185)
+		if ($doNotSubmit && $hasUpload && !empty($_SESSION['FILES']))
```
leofeyer, line 263: Do we not need the `&& !Environment::get('isAjaxRequest')` check here as well?

fritzmg: I don't think so. It looks like a bug to me that `$_SESSION['FILES']` was not purged in the past when the form was sent via AJAX. Or conversely: it should not matter whether the form was sent via AJAX or not. If the form does not validate, any uploaded file should be deleted, regardless of the type of request, should it not?

leofeyer:
Thank you @fritzmg.

PR #3737, opened 2021-11-29
Improve the system maintenance mode

Toflar:
This is the complementary PR for https://github.com/contao/contao/pull/3729.

Once https://github.com/contao/contao/pull/3729 is merged, we'll have a better maintenance mode which can be enabled per root page and the content can be managed and styled individually.
This covers the "content maintenance mode" (aka system is running but parts of the website are just not ready yet).

What's missing in order to get rid of the lexik maintenance bundle completely is a new "system maintenance mode" (aka not even the system is running). This is a use case we have for example when updating a system, running migrations/new deployments etc.
But to make this a real improvement over the lexik maintenance `kernel.request` listener, we have to take this even a step further because it might be that the application isn't running at all because for example the whole `vendor` directory isn't even installed (yet). The maintenance message should still be shown then. 

That's why I have designed this new mode with a simple "if a given file exists, show and abort" approach directly in our `index.php` before even loading any vendors or autoloading files. That means you can also enable the "system maintenance mode" without having any CLI running. 

The implementation is based on a Twig template which defaults to `@ContaoCore/Error/service_unavailable.html.twig` so it's also backwards compatible ;-)

The new `contao:maintenance-mode` command takes `enable` or `disable` as arguments (can also use `on` or `off`).
When enabling, you can pass a custom Twig template name as well as custom template variables for maximum flexibility.


Toflar:
Ready for review but cannot be merged without https://github.com/contao/contao/pull/3729.

fritzmg (commented):
Should we also add

```
RewriteCond %{DOCUMENT_ROOT}/maintenance.html -f
RewriteCond %{SCRIPT_FILENAME} !maintenance.html
RewriteRule ^ /maintenance.html [R=503,L]
ErrorDocument 503 /maintenance.html
```

to the `.htaccess` of the skeleton? This would bypass PHP completely for any servers running Apache with `AllowOverride`.

Toflar:
I also want to try to use templates here :)

fritzmg:
Btw. this will break some of our websites after deployment, because we already have a `maintenance.html` in our document root 😁. We are using a `maintenance.enable` flag there (instead of renaming the file):

```
RewriteCond %{DOCUMENT_ROOT}/maintenance.html -f
RewriteCond %{DOCUMENT_ROOT}/maintenance.enable -f
RewriteCond %{SCRIPT_FILENAME} !maintenance.html
RewriteRule ^ /maintenance.html [R=503,L]
ErrorDocument 503 /maintenance.html
```

Not a huge issue. I can provide a follow up PR with a migration though.

Toflar:
I would like to ignore the `.htaccess` changes for this PR as I'm not really sure we want to have those. 

The implementation is updated and we're now using Twig :) This means, if you have already adjusted the `service_unavailable` template, it will work for you out of the box 🎉

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    private function enable(string $templateName, string $templateVars): void
+    {
+        // Render the template and write it to maintenance.html
+        $this->filesystem->dumpFile(
+            Path::join($this->webDir, 'maintenance.html'),
+            $this->twig->render($templateName, array_merge([
+                'statusCode' => 503,
+                'language' => 'en',
+                'template' => $templateName,
```
m-vo: Just an idea: We could in theory even generate a file per root page language (`maintenance.html`, `maintenance.de.html`, `maintenance.fr.html` …) and match accordingly in the `index.php`. 

But probably also something for another PR (and while watching https://github.com/symfony/symfony/issues/35925).

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('contao:maintenance-mode')
+            ->addArgument('state', InputArgument::REQUIRED, 'Use "enable" to enable and "disable" to disable the maintenance mode. If the state is already the desired one, nothing happens. You can also use "on" and "off".')
+            ->addOption('template', 't', InputOption::VALUE_REQUIRED, 'Allows to take a different Twig template name when enabling the maintenance mode.', '@ContaoCore/Error/service_unavailable.html.twig')
+            ->addOption('templateVars', null, InputOption::VALUE_OPTIONAL, 'Add custom template variables to the Twig template when enabling the maintenance mode (provide as JSON).', '{}')
+            ->setDescription('Changes the state of the system maintenance mode. Without any argument, it toggles between the current state.')
```
m-vo: I think I'd remove toggling. People will use this in scripts and promoting only the explicit way will likely remove errors.

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,96 @@
...
+class MaintenanceModeCommand extends Command
+{
+    private string $webDir;
+    private Environment $twig;
+    private Filesystem $filesystem;
+
+    public function __construct(string $webDir, Environment $twig, Filesystem $filesystem = null)
+    {
+        $this->webDir = $webDir;
+        $this->twig = $twig;
```
m-vo, line 36: Should we call the property for the `\Twig\Environment` `$twig` or `$environment`? We're currently mixing this up but I think we should stick to one.

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,96 @@
...
+            ->addArgument('state', InputArgument::REQUIRED, 'Use "enable" to enable and "disable" to disable the maintenance mode. If the state is already the desired one, nothing happens. You can also use "on" and "off".')
+            ->addOption('template', 't', InputOption::VALUE_REQUIRED, 'Allows to take a different Twig template name when enabling the maintenance mode.', '@ContaoCore/Error/service_unavailable.html.twig')
+            ->addOption('templateVars', null, InputOption::VALUE_OPTIONAL, 'Add custom template variables to the Twig template when enabling the maintenance mode (provide as JSON).', '{}')
+            ->setDescription('Changes the state of the system maintenance mode. Without any argument, it toggles between the current state.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $shouldEnable = \in_array($input->getArgument('state'), ['enable', 'on'], true);
```
m-vo, line 55: What about using `filter_var`? This handles probably all "boolean config" cases.

```suggestion
        $shouldEnable = filter_var($input->getArgument('state'), FILTER_VALIDATE_BOOLEAN);
```

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,96 @@
...
+
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('contao:maintenance-mode')
+            ->addArgument('state', InputArgument::REQUIRED, 'Use "enable" to enable and "disable" to disable the maintenance mode. If the state is already the desired one, nothing happens. You can also use "on" and "off".')
+            ->addOption('template', 't', InputOption::VALUE_REQUIRED, 'Allows to take a different Twig template name when enabling the maintenance mode.', '@ContaoCore/Error/service_unavailable.html.twig')
```
m-vo, line 47: I wonder if we should add a new `maintenance.html.twig` template to the Contao template folder and use `@Contao/maintenance.html.twig` here. Slowly moving away from bundle templates.

Could be done in a later PR as well.

m-vo (changes requested):
I really like this one! A much better solution than before + we can get rid of another deprecated bundle! :tada: 

Code looks good to me, I just left some minor remarks.

And: What's about BC? Can we really just remove classes and functions like `Maintenance` or  `Messages#maintenanceCheck()`?

manager-bundle/src/Command/MaintenanceModeCommand.php
Toflar: It's not toggling at all, that's a wrong description, I'll fix it.

manager-bundle/src/Command/MaintenanceModeCommand.php
Toflar: To me, it's always been `$twig`. In all my projects because `environment` can be anything.

manager-bundle/src/Command/MaintenanceModeCommand.php
Toflar: I would prefer keeping the existing code because it has to match the command docs. Otherwise we have to document that `1` and `yes` work as well. I still hope that Symfony Console gets support for enums at some point so the options would be predefined anyway.

manager-bundle/src/Command/MaintenanceModeCommand.php
Toflar: I don't know the exact implications but it's your area of expertise :) I chose the existing one because people might have already adjusted it. That means they have the major benefit of not having to change a thing at all :)
So if we want to change anything here, we should probably do that later, yes.

fritzmg (dismissed):
LGTM 👍 I like the simplistic approach - and being largely independent from a running application once active, especially if you have additional rules for your web server. It's something we use in many of our projects in any case, as already mentioned.

Toflar:
> And: What's about BC? Can we really just remove classes and functions like Maintenance or Messages#maintenanceCheck()?

@m-vo I don't see any reason why anybody would've ever used this. If you have called the `Maintenance` class yourself you were probably drunk and the message doesn't need to be there because the whole system is now in maintenance mode. So nothing is running :)

manager-bundle/src/Command/MaintenanceModeCommand.php
m-vo: As this is the only thing you're typically interfacing with I think `$twig` is a good idea. :+1:  Let's adjust this in a another PR.

manager-bundle/src/Command/MaintenanceModeCommand.php
m-vo: On my list.

m-vo (dismissed):

core-bundle/src/Resources/contao/classes/Messages.php
```diff
@@ -31,35 +31,6 @@ public function versionCheck()
 		return '';
 	}
 
-	/**
-	 * Check for maintenance mode
-	 *
-	 * @return string
-	 */
-	public function maintenanceCheck()
```
leofeyer, line 41: We should probably keep the method, deprecate it and return an empty string.

manager-bundle/src/Resources/skeleton/public/index.php
```diff
@@ -23,6 +23,18 @@
     stream_wrapper_unregister('phar');
 }
 
+// System maintenance mode comes first as it has to work even if the vendor directory does not exist
+if (file_exists(__DIR__.'/maintenance.html')) {
```
leofeyer: Are you sure that this works? If there is a file named `maintenance.html` in the document root, the web server will not rewrite the request to the `index.php` file, will it?

core-bundle/src/Exception/ServiceUnavailableException.php
```diff
@@ -12,12 +12,8 @@
 
 namespace Contao\CoreBundle\Exception;
 
-trigger_deprecation('contao/core-bundle', '4.1', 'Using the "Contao\CoreBundle\Exception\ServiceUnavailableException" class has been deprecated and will no longer work in Contao 5.0. Use the "Lexik\Bundle\MaintenanceBundle\Exception\ServiceUnavailableException" class instead.');
+use Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException;
```
leofeyer, line 15: IMHO we should keep the deprecation and update it to say `use the Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException class instead.`

core-bundle/src/Exception/ServiceUnavailableException.php
```diff
@@ -12,12 +12,8 @@
...
 
-/**
- * @deprecated Deprecated since Contao 4.1, to be removed in Contao 5.0; use the
- *             Lexik\Bundle\MaintenanceBundle\Exception\ServiceUnavailableException instead
- */
-class ServiceUnavailableException extends \RuntimeException
+class ServiceUnavailableException extends ServiceUnavailableHttpException
```
leofeyer: Are you using this class anywhere?

core-bundle/src/Resources/contao/classes/Messages.php
Toflar: As I already said to @m-vo I don't see why anyone would've ever called this. It was a `getSystemMessages` hook listener so there's absolutely no valid use case to have called this method yourself.

manager-bundle/src/Resources/skeleton/public/index.php
Toflar: If the maintenance mode is enabled, anybody who calls `foobar.com/maintenance.html` will get the same as if they called `foobar.com/foobar`, yes. But who cares 🤷‍♂️

core-bundle/src/Exception/ServiceUnavailableException.php
Toflar: I need to bring it back because you should use this one instead of the lexik one. We're using it in the PrettyErrorScreenListener. But once https://github.com/contao/contao/pull/3729 is merged (which has to happen first), that will already be the case. So any remarks regarding this exception should be done there :)

aschempp: FYI undeprecating this is also done by https://github.com/contao/contao/pull/3729, but I don't think we should change the parent method. We don't extend `NotFoundHttpException` in our `PageNotFoundException` either etc. Our `ExceptionConverterListener` will correctly handle and convert the exception.

leofeyer: #3729 has been merged meanwhile, so this actually needs to be reverted, right?

**Edit:** Yes, it does. The unit tests fail with this change, so I have reverted it in f6a0279f322e7ea37cd74df8ebffa9c7e92b8a41.

core-bundle/src/Resources/contao/classes/Messages.php
leofeyer: Ok, let's see if this assumption is correct. 😄

manager-bundle/src/Resources/skeleton/public/index.php
leofeyer: No, they will not get the same, because `foobar.com/maintenance.html` will return status `200 OK` whereas `foobar.com/foobar` will return status `503 Service unavailable`. Can we not write the file to `var/maintenance` instead? Then every request will be answered with `503 Service unavailable`.

leofeyer (changes requested):

Toflar:
The `maintenance.html` is now witten to `var` :) https://github.com/contao/contao/pull/3737/commits/2714b27a0b9e131c4691653186e23275faf816f0

leofeyer (approved):

leofeyer:
Thank you @Toflar.

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $shouldEnable = \in_array($input->getArgument('state'), ['enable', 'on'], true);
+        $io = new SymfonyStyle($input, $output);
+
+        if ($shouldEnable) {
+            $this->enable($input->getOption('template'), $input->getOption('templateVars'));
+            $io->success('Maintenance mode enabled.');
+        } else {
+            $this->disable();
+            $io->success('Maintenance mode disabled.');
```
aschempp, line 63: This means I can write `contao:maintenance foobar` and it will disable the mode?

manager-bundle/src/Resources/skeleton/public/index.php
```diff
@@ -23,6 +23,18 @@
     stream_wrapper_unregister('phar');
 }
 
+// System maintenance mode comes first as it has to work even if the vendor directory does not exist
+if (file_exists(__DIR__.'/../var/maintenance.html')) {
+    $contents = file_get_contents(__DIR__.'/../var/maintenance.html');
+
+    header('HTTP/1.1 503 Service Unavailable', true, 503);
```
aschempp, line 30: According to the PHP docs, we should probably use `http_​response_​code()` instead of this.

manager-bundle/src/Command/MaintenanceModeCommand.php
aschempp: fixed in https://github.com/contao/contao/pull/3883

manager-bundle/src/Resources/skeleton/public/index.php
leofeyer: Are you going to fix this in #3883 as well?

aschempp: Not related to the command imho. I'm not entirely sure what the correct headers would be, but it's still on my radar. But feel free to create a PR if you find time, otherwise I'll probably do that next year.

PR #3739, opened 2021-11-29
Handle the routing exception in more navigation modules

aschempp:
As discussed in #3605, we already handle this exception in the other navigation modules but it was probably forgotten here.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleCustomnav.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleCustomnav.php
@@ -151,7 +151,7 @@ class ModuleCustomnav extends Module
 				$trail = \in_array($objModel->id, $objPage->trail);
 
 				// Use the path without query string to check for active pages (see #480)
-				[$path] = explode('?', Environment::get('request'), 2);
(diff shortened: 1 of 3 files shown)


fritzmg:
Doesn't this occur in `4.9` as well?

aschempp:
we don't handle these in 4.9, I think it was specifically introduced with the new "legacy" routing

fritzmg (approved):

Toflar (approved):

leofeyer:
> we already handle this exception in the other navigation modules but it was probably forgotten here.

Where exactly do we handle this exception?

fritzmg:
> Where exactly do we handle this exception?

For `ModuleNavigation` and `ModuleSitemap` here: 

https://github.com/contao/contao/blob/b1419cc981cf59fd451342863494e61b0ed04c07/core-bundle/src/Resources/contao/modules/Module.php#L363-L386

For `ModuleBreadcrumb` here:

https://github.com/contao/contao/blob/b1419cc981cf59fd451342863494e61b0ed04c07/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php#L250-L262

PR #3741, opened 2021-11-29
Make sure GUIDs in RSS feeds are unique

### LINKED ISSUE #3572: veranstaltungen.xml does not validate
The RSS feed of the events does not validate:

See https://validator.w3.org/feed/check.cgi?url=https://contao.org/share/veranstaltungen.xml

grafik

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3572
| Docs PR or issue | -

Use custom GUIDs (with a `#Y-m-d` suffix) for repeated events, so the GUIDs remain unique in the RSS file.

### DIFF
--- a/calendar-bundle/src/Resources/contao/classes/Calendar.php
+++ b/calendar-bundle/src/Resources/contao/classes/Calendar.php
@@ -195,7 +195,7 @@ class Calendar extends Frontend
 
 						if ($intStartTime >= $time)
 						{
-							$this->addEvent($objArticle, $intStartTime, $intEndTime, $strUrl);
+							$this->addEvent($objArticle, $intStartTime, $intEndTime, $strUrl, '', true);
 						}
 					}
 				}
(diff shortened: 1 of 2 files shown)


calendar-bundle/src/Resources/contao/classes/Calendar.php
```diff
@@ -231,13 +231,17 @@ protected function generateFiles($arrFeed)
...
 					$objItem->title = $event['title'];
 					$objItem->link = $event['link'];
 					$objItem->published = $event['tstamp'];
 					$objItem->begin = $event['startTime'];
 					$objItem->end = $event['endTime'];
 
+					if ($event['isRepeated'] ?? null)
+					{
+						$objItem->guid = $event['link'] . '#' . date('Y-m-d', $event['startTime']);
+					}
```
fritzmg, line 243: imho it would be fine to always append `#Y-m-d`. Then you would not need a new parameter.

leofeyer: No, the first event should have the real URL IMHO.

core-bundle/src/Resources/contao/library/Contao/Feed.php
```diff
@@ -142,15 +142,7 @@ public function generateRss()
...
-				// Add the isPermaLink attribute if the guid is not a link (see #4930)
-				if (strncmp($objItem->guid, 'http://', 7) !== 0 && strncmp($objItem->guid, 'https://', 8) !== 0)
-				{
-					$xml .= '<guid isPermaLink="false">' . $objItem->guid . '</guid>';
-				}
-				else
-				{
-					$xml .= '<guid>' . $objItem->guid . '</guid>';
-				}
+				$xml .= '<guid isPermaLink="false">' . $objItem->guid . '</guid>';
```
leofeyer, line 145: I have simplified this, because if `$objItem->guid` is set, we can assume that the GUID is not a link. Otherwise it would have been set under `$objItem->link`.

calendar-bundle/src/Resources/contao/classes/Calendar.php
ausi: ~~I also think we sould only append the hash for the duplicates, not for the first one. This way we would only get `isPermaLink="false"` for the duplicate URLs but not for the first one.~~ This is alredy the case I see now 🙈

ausi (approved):

PR #3742, opened 2021-11-29
Show error if preview settings could not be applied

fritzmg:
This is a follow up to #3732 as discussed there. While #3732 prevents the browser from auto filling the member username field in the preview bar, part of the problem is that applying the settings fails silently, if you specify a member username that does not exist.

This PR fixes that by responding with an appropriate error message.

image

The necessary changes also revealed a problem in the tests (that's why there are also so many changes in the tests).


aschempp (commented):
Very simple solution to the problem. I would prefer to optimise this a bit though. The controller could return a JSON result so the preview bar actually knows what happened. 

A `400 Bad Request` seems incorrect to me. If the user does not have permissions (not `ROLE_USER`) a `401`/`403` would be more appropriate. And apparently this permission check is incorrectly performed in the controller itself (`$user instanceof BackendUser` should be `denyAccessUnlessGranted('ROLE_USER')`. Then we don't need error-checking the `authenticateFrontendGuest` anyway.

fritzmg:
> The controller could return a JSON result so the preview bar actually knows what happened.

I thought about that too, but I find it superfluous. All wee need is a status and an error message - which the HTTP response provides already.

> A `400 Bad Request` seems incorrect to me. If the user does not have permissions (not `ROLE_USER`) a `401`/`403` would be more appropriate. And apparently this permission check is incorrectly performed in the controller itself (`$user instanceof BackendUser` should be `denyAccessUnlessGranted('ROLE_USER')`. Then we don't need error-checking the `authenticateFrontendGuest` anyway.

We would still need error-checking, since the form is sent via AJAX and thus we need to handle the response and show the error message.

fritzmg:
I have made a few changes:

* The `BackendPreviewSwitchController` now checks for `ROLE_USER` and responds with either `401` or `403` accordingly.
* The JavaScript now shows the error message when it was a `400` response, otherwise it will reload the page.
* In this case the only error message we need will be the one when there is an invalid username given. In all other cases a `401` or `403` response would have been generated before hand (or `400` if it was not an AJAX request in the first place).

This causes the following improvement:

* If you are in the `preview.php` entry point and you lose your authentication (e.g. your session/cookie expires or gets deleted) and if you then click _Apply_ in the preview bar, the controller will now respond with `401` or `403` which in turn will cause the JavaScript to reload the page.
* This in turn will cause a `302` response to the back end login, where you will be able to log in again.
* After logging in, you will redirected back to the front end preview where you were left off.

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -90,16 +95,19 @@ public function __construct(FrontendPreviewAuthenticator $previewAuthenticator,
...
     }
 
     /**
      * @Route("/contao/preview_switch", name="contao_backend_switch")
      */
     public function __invoke(Request $request): Response
     {
-        $user = $this->security->getUser();
+        if (!$this->security->isGranted('ROLE_USER')) {
+            return new Response('Access Denied', $this->security->isGranted('IS_AUTHENTICATED_FULLY') ? Response::HTTP_FORBIDDEN : Response::HTTP_UNAUTHORIZED);
```
aschempp: This should throw an exception to let security and prettyerrorscreen handle the response. If the controller would extend `AbstractController` this would ve as simple as `$this->denyAccessUnlessGranted('ROLE_USER')`

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -238,8 +238,8 @@
                 const formData = new FormData(form);
 
                 request('POST', form.action, formData, function () {
-                    if (204 === this.status) {
-                        location.reload();
+                    if (400 === this.status) {
+                        alert(this.responseText ? this.responseText : 'An unknown error occurred.');
```
```suggestion
                        alert(this.responseText || 'An unknown error occurred.');
```

In JavaScript `||` behaves like `?:` in PHP.

ausi (dismissed):

ausi (dismissed):

aschempp (commented):
I still don't think `400 Bad Request` is the correct status code if a username was not found. You wouldn't return a `400` if a form was submitted and an input field did not validate?

Anyway, merging as-is works and sure is an improvement, we can always improve later again 🙃

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: `$this->denyAccessUnlessGranted('ROLE_USER')` only generates a `403` status. Also I think it's overkill to extend from `AbstractController` just for this. Furthermore I don't think the `PrettyErrorScreenHandler` should be invoked for AJAX requests.

fritzmg:
> I still don't think 400 Bad Request is the correct status code if a username was not found. You wouldn't return a 400 if a form was submitted and an input field did not validate?

400 is the most common status code to indicate invalid parameters or invalid parameter values.

core-bundle/src/Controller/BackendPreviewSwitchController.php
aschempp: yes, it only generates a `403` because apparently Symfony always does that, so I would be fine with not differentiating between `401` and `403`. The PrettyErrorScreenHandler only renders if the request expects an HTML response.

leofeyer (changes requested):
Instead of an alert box that interrupts the workflow, I'd prefer if we, for example, highlighted the field in red or showed a flash message that does not need to be confirmed.

fritzmg:
Do we have any existing code/styles for that?

fritzmg:
And if we don't I don't think it is worth investing effort into it, since this will only be seen in edge cases anyway. Either you leave the field empty, or you select an auto-completed username. Also imho it should not be just a flash message that does not have to be confirmed. Otherwise you might again not realise that the settings did not actually apply.

ausi:
`input.setCustomValidity()` could be an alternative 😎 https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity

fritzmg:
Done, looks like this now (in Google Chrome for instance):

image

ausi (dismissed):

aschempp:
Nice! I would personally remove the `Preview settings could not be applied` and just add the message related to the input field (`Invalid username XY`), but choose what you prefer 😊

fritzmg:
I would keep it as is - in order to communicate that the "apply" action failed.

aschempp:
> I would keep it as is - in order to communicate that the "apply" action failed.

If any input field in the form has an error, I would always assume the whole form failed. Isn't that the point of (client side) errors in forms?

fritzmg:
But it's not client side. The request is actually sent and the error response regards the whole request - i.e. the apply action of the whole form.

aschempp:
yeah its server side validation? If you submit a form in the form generator, you wouldn't expect the email to be sent if your name input comes back with an error message? Or if you register an account and your password is too short, you wouldn't expect to get an account without a password (= having the "other" fields be applied)?

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: Imho if we should leave it as it is.

fritzmg:
@aschempp I still disagree. I think it is fine as it is now.

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -155,13 +161,19 @@ private function authenticatePreview(Request $request): void
         $showUnpublished = 'hide' !== $request->request->get('unpublished');
 
         if (null !== $frontendUsername) {
-            $this->previewAuthenticator->authenticateFrontendUser($frontendUsername, $showUnpublished);
+            if (!$this->previewAuthenticator->authenticateFrontendUser($frontendUsername, $showUnpublished)) {
+                $message = $this->translator->trans('ERR.previewSwitchInvalidUsername', [$frontendUsername], 'contao_default');
+
+                return new Response($message, Response::HTTP_BAD_REQUEST);
```
leofeyer: This is wrong. Bad request is to be used if a request was syntactically incorrect and not understood by the server, which is not the case here. If a user is attempting to authenticate, but provides invalid credentials, the response should have a status of 401.

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: @fritzmg Please don‘t resolve other people‘s conversations unless you actually have resolved their issues (e.g. by merging their suggestion or adjusting the code to their recommendations).

leofeyer: I also think that we should use `$this->denyAccessUnlessGranted('ROLE_USER')` here or at least behave the same (which means not returning a `401` status code).

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -239,6 +239,9 @@
       <trans-unit id="ERR.moduleWithoutColumn">
         <source>Please assign all modules to a column!</source>
       </trans-unit>
+      <trans-unit id="ERR.previewSwitchInvalidUsername">
+        <source>Preview settings could not be applied: invalid username "%s".</source>
```
```suggestion
        <source>Preview settings could not be applied: Unknown username "%s".</source>
```

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -233,25 +233,36 @@
...
+            const userField = form.querySelector('input[name="user"]');
+
             form.addEventListener('submit', function (e) {
                 e.preventDefault();
                 const formData = new FormData(form);
 
                 request('POST', form.action, formData, function () {
-                    if (204 === this.status) {
-                        location.reload();
+                    if (400 === this.status) {
```
leofeyer: This should be 401, too.

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: I don't think that would be correct. This already is an authenticated request and thus 401 would indicate that you are not authenticated (which you are). Also we need to be able to distinguish a non-authenticated request.

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
fritzmg: See above. The response will be `401` if you are not authenticated (in which case we want to reload the preview page so that you can log in again) and `400` if the provided username does not exist, so that we can show the error.

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: The method name `authenticateFrontendUser()` already says that this is in fact an authentication request. 🤷‍♂️

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: > Please don‘t resolve other people‘s conversations unless you actually have resolved their issues (e.g. by merging their suggestion or adjusting the code to their recommendations).

I think we need a better process here then - because then PRs potentially never can get merged.

> I also think that we should use $this->denyAccessUnlessGranted('ROLE_USER') here or at least behave the same (which means not returning a 401 status code).

Please elaborate why, as I still disagree. The response should be 401 if you are not authenticated and 403 if you are authenticated but not allowed to use the preview.

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: But the request is already authenticated and 401 would indicate to you that you are not authenticated - but you are authenticated. 

Also we need to distinguish between not authenticated and invalid username for the JavaScript to work. Currently we only rely on status codes.

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: Because that is what Symfony‘s `denyAccessUnlessGranted()` method does.

fritzmg: But that method is supposed to be only used when you are authenticated, as it will only generate a 403 response. For the same reason we cannot use this method in Contao otherwise, as we want to distinguish between 401 (Not Authenticated) and 403 (Access Denied).

Furthermore, as already explained above, we do not want to throw an exception here as we do not want the prettyerrorscreen listener to show a pretty error screen as this is only used for an AJAX request.

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: As discussed in the Contao call, we want to use status `422` here.

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -239,6 +239,9 @@
       <trans-unit id="ERR.moduleWithoutColumn">
         <source>Please assign all modules to a column!</source>
       </trans-unit>
+      <trans-unit id="ERR.previewSwitchInvalidUsername">
+        <source>Preview settings could not be applied: Unknown username "%s".</source>
```
```suggestion
        <source>Unknown username: %s</source>
```

As discussed in the Contao call, we don‘t want to show the _Preview settings could not be applied_ part.

core-bundle/src/Controller/BackendPreviewSwitchController.php
leofeyer: As discussed in the Contao call, we want to throw an `AccessDeniedException` here. (Just for the record: `IS_AUTHENTICATED_FULLY` would have had to be replaced with `!IS_AUTHENTICATED`.

fritzmg: For the record: I still disagree with this as the exception will now render the full pretty error screen, which I find unnecessary for this AJAX request. But implemented in 77a7ad6fa anyway.

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: Implemented in 77a7ad6fa

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
fritzmg: 422 is now used in 77a7ad6fa

fritzmg:
> input.setCustomValidity() could be an alternative 😎 https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity

Unfortunately in FireFox this will only be visible if you set `signon.showAutoCompleteFooter` to `false` in your `about:config` - as FireFox still identifies this as a login username field despite #3732 😑. The only way to make Firefox _not_ identify this as a username field is to remove/replace `placeholder="Username"` as far as I tested 🙄

core-bundle/src/Controller/BackendPreviewSwitchController.php
fritzmg: Actually, since we are not using `400` for the response to indicate that the username is invalid anymore, I have reverted this to the original state.

aschempp (dismissed):

Toflar (dismissed):
❤️

fritzmg:
@leofeyer anything missing to get this merged?

ausi (dismissed):

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -233,25 +233,36 @@
...
-                    if (204 === this.status) {
-                        location.reload();
+                    if (422 === this.status) {
+                        if (userField) {
+                            userField.setCustomValidity(this.responseText || 'An unknown error occurred.');
+                            userField.reportValidity();
+                        }
                     } else {
                         location.reload();
                     }
```
m-vo: Maybe simpler if the first condition is inverted? (opinionated)
```suggestion
                    if (422 !== this.status) {
                        location.reload();
                    }

                    if (userField) {
                            userField.setCustomValidity(this.responseText || 'An unknown error occurred.');
                            userField.reportValidity();
                    }
```

m-vo (dismissed):

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
m-vo: (btw. I hate the diff view on this :smile:)

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3743, opened 2021-11-29
Fix the PasswordHasherFactory usage

bytehead:
Missing fixes for #3257.

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
@@ -32,6 +32,7 @@ class MakeServicesPublicPass implements CompilerPassInterface
         'security.authentication.trust_resolver',
         'security.firewall.map',
         'security.logout_url_generator',
+        'security.password_hasher_factory',
         'security.helper',
         'uri_signer',
     ];
@@ -39,7 +40,6 @@ class MakeServicesPublicPass implements CompilerPassInterface
     private const ALIASES = [
         'database_connection',
         'mailer',
-        'security.password_hasher_factory',
     ];
 
     public function process(ContainerBuilder $container): void
--- a/core-bundle/src/Resources/contao/controllers/BackendPassword.php
+++ b/core-bundle/src/Resources/contao/controllers/BackendPassword.php
@@ -12,6 +12,7 @@ namespace Contao;
 
 use Contao\CoreBundle\Exception\AccessDeniedException;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\PasswordHasher\PasswordHasherInterface;
 
 /**
  * Back end help wizard.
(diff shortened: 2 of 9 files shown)


aschempp (dismissed):

bytehead:
Tests fail possibly unrelated:

```
There was 1 failure:

1) Contao\ManagerBundle\Tests\Api\ApplicationTest::testGetsCommandsFromPlugins
Failed asserting that actual size 6 matches expected size 4.
```

bytehead:
And what happened to PHPStan? 

```
 ------ ----------------------------------------------------------------------- 
  Line   src/Security/Authentication/ContaoLoginAuthenticationListener.php      
 ------ ----------------------------------------------------------------------- 
  79     Parameter #1 $user of class                                            
         Symfony\Component\Security\Core\Authentication\Token\UsernamePassword  
         Token constructor expects                                              
         Symfony\Component\Security\Core\User\UserInterface, string given.      
 ------ -----------------------------------------------------------------------
```

```php
class UsernamePasswordToken extends AbstractToken
{
    private $credentials;
    private $firewallName;

    /**
     * @param string|\Stringable|UserInterface $user        The username (like a nickname, email address, etc.) or a UserInterface instance
     * @param mixed                            $credentials
     * @param string[]                         $roles
     *
     * @throws \InvalidArgumentException
     */
    public function __construct($user, $credentials, string $firewallName, array $roles = [])
```

fritzmg:
> Tests fail possibly unrelated:

See https://github.com/contao/contao/pull/3745

bytehead:
PHPStan fails because of https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Core/Authentication/Token/UsernamePasswordToken.php#L27.

bytehead:
PHPStan fixes in #3746

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
```diff
@@ -32,14 +32,14 @@ class MakeServicesPublicPass implements CompilerPassInterface
         'security.authentication.trust_resolver',
         'security.firewall.map',
         'security.logout_url_generator',
+        'security.password_hasher_factory',
```
Toflar, line 35: This reveals an issue in the logic of this compiler pass. It shouldn't matter wheter it's a service or an alias. We should use `findDefinition()` which checks both, alias and service ID.

Toflar: And catch the exception to do nothing to keep the existing `->has()` logic.

bytehead: You want to fix that issue in this PR?

Toflar: I can create a separate PR :)

bytehead: I'm fine with fixing it - I just don't know if the scope here is correct?

aschempp: I don't think it is 😬

bytehead: I'll provide a separate PR 👍

bytehead: @Toflar see https://github.com/contao/contao/pull/3759

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #3744, opened 2021-11-29
Show bubbled exceptions in the pretty error screen listener

aschempp:
If debug mode is enabled, the pretty error screen listener will silently swallow any exceptions in the sub request. With this change, the subrequest exception will be shown in the inspector instead of the one that triggered the error screen. I noticed this while working on #3729, an error on a 503 page module never showed up in the inspector.


Toflar:
Is that really 4.13 only?
Looks good but needs a test to prevent regressions :)

aschempp:
> Is that really 4.13 only?

see https://github.com/contao/contao/pull/3719

aschempp:
Failing tests related to #3748/#3745 and https://github.com/contao/contao/pull/3746

Toflar (approved):

PR #3745, opened 2021-11-29
Check for specific commands in ApplicationTest

fritzmg:
Symfony 5.4 added two new commands in https://github.com/symfony/symfony/pull/42251 - `_complete` and `completion`. This makes the test fail, since it checks for exactly 4 commands. To support _prefer lowest_ this PR fixes this by checking for an intersection of the commands instead.

### DIFF
--- a/manager-bundle/tests/Api/ApplicationTest.php
+++ b/manager-bundle/tests/Api/ApplicationTest.php
@@ -106,11 +106,10 @@ class ApplicationTest extends ContaoTestCase
         $application = $this->getApplication();
         $application->setPluginLoader($pluginLoader);
 
-        /** @var array $commands */
-        $commands = $application->all();
+        $expectedCommands = ['help', 'list', 'version', 'config:get'];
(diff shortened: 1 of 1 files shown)


bytehead:
PHPStan fixes are in #3747.

bytehead (approved):
I wonder why the unit tests still failing, changes looking correct to me.

leofeyer (changes requested):
Since the purpose of the test is to check if the `config:get` command has been added, we can simply drop the `$this->assertCount(4, $commands)` assertion.

leofeyer:
> I wonder why the unit tests still failing, changes looking correct to me.

It is due to the remaining self deprecations I think.

aschempp:
alternatively see https://github.com/contao/contao/pull/3748, especially regarding raising the minimum Symfony version to 5.4

fritzmg:
Closing as the changes from https://github.com/contao/contao/pull/3749 will be backported by @leofeyer.

PR #3746, opened 2021-11-29
Upgrade symfony/security-bundle to 5.4 and fix TokenInterface usage

bytehead:
Fixes the tests for example in #3743 or #3745 and all upcoming PRs.


bytehead:
Still failing unit tests will bi fixed by #3745.

core-bundle/src/Security/Authentication/Token/FrontendPreviewToken.php
```diff
@@ -23,7 +24,7 @@ public function __construct(?FrontendUser $user, bool $showUnpublished)
     {
         if (null === $user) {
             parent::__construct();
-            $this->setUser('anon.');
+            $this->setUser(new InMemoryUser('anon.', null));
```
leofeyer: According to Symfony, _anonymous is now represented by the absence of a token_.

https://github.com/symfony/symfony/blob/e0fdec0f739f7324e1338940a5b42c787a81167b/src/Symfony/Component/Security/Core/Authentication/Token/AnonymousToken.php#L21

Shouldn't the code be something like this then?

```php
if (null === $user) {
    parent::__construct();
} else {
    parent::__construct($user->getRoles());

    $this->setUser($user);
    $this->setAuthenticated(true);
}

$this->showUnpublished = $showUnpublished;
```

@aschempp /cc

core-bundle/tests/Security/Authentication/Token/FrontendPreviewTokenTest.php
```diff
@@ -56,9 +56,9 @@ public function testSerializesItself(): void
         $serialized = $token->__serialize();
 
         if (method_exists($token, 'getRoles')) {
```
leofeyer: This should always be true in Symfony 5.4, so you might remove the if condition. Not strictly related to this PR though.

richardhj:
The current 4.x still uses the AnonymousToken in the frontend to represent an unauthorized access.

@bytehead Do you know how to fix this?

Screenshot 2021-12-01 at 15 40 53

core-bundle/src/Security/Authentication/Token/FrontendPreviewToken.php
bytehead: Correct 👍

bytehead:
@richardhj 

> Do you know how to fix this?

Possibly in with the new security system for Contao 5. For 4.x the anonymous token will stay.

core-bundle/src/Security/Authentication/Token/FrontendPreviewToken.php
bytehead: Fixed in d485127b699c3a4d1e83647fa33f59475f3fb149.

core-bundle/tests/Security/Authentication/Token/FrontendPreviewTokenTest.php
bytehead: Fixed in 1f3b2532f.

core-bundle/src/Security/Authentication/Token/FrontendPreviewToken.php
bytehead: Had to undo d485127b699c3a4d1e83647fa33f59475f3fb149 in 530a62c6f. If we change to `anonymous is now represented by the absence of a token`, we have to adjust more or less all security related stuff.

bytehead:
Had to add some PHPStan ignores as we cannot get rid of some stuff unless we exchange the whole security to the new authentication system.

bytehead:
@leofeyer I'll fix the new deprecations in a next PR or after #3750 has been merged 👍

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #3749, opened 2021-11-30
Fix the CI chain

leofeyer:
This PR fixes the CI chain. Don't mind the failing PHPStan job, it will be fixed by #3746.


fritzmg:
Shouldn't this be targeted towards `4.12`? The same jobs are failing there as well (since 4.12 can be used with Symfony 5.4).

leofeyer:
Yes, I will backport the changes.

PR #3750, opened 2021-11-30
Make Symfony 5.4 the minimum requirement

leofeyer:
This PR makes Symfony 5.4 the minimum requirement and fixes all deprecations except the security related ones, which @bytehead is going to fix in #3746.


core-bundle/README.md
```diff
@@ -121,10 +121,10 @@ security:
                     - contao.security.logout_handler
 
     access_control:
-        - { path: ^%contao.backend.route_prefix%/login$, roles: IS_AUTHENTICATED_ANONYMOUSLY }
-        - { path: ^%contao.backend.route_prefix%/logout$, roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: ^%contao.backend.route_prefix%/login$, roles: PUBLIC_ACCESS }
```
leofeyer, line 124: @bytehead @aschempp Is this correct?

core-bundle/src/Controller/Page/RootPageController.php
```diff
@@ -39,8 +39,8 @@ private function getNextPage(int $rootPageId): PageModel
             return $nextPage;
         }
 
-        if ($this->has('logger')) {
-            $this->get('logger')->error(
+        if ($this->container->has('logger')) {
+            $this->container->get('logger')->error(
```
leofeyer, line 43: I really think this should have been a constructor injection.

Toflar: Everything should be constructor injection if it does not need to be conditionally loaded. E.g. every controller that uses "service A" all the time, should use constructor injection and never subscribe to it. You should only subscribe if you provide e.g. abstract functions other children *can* use. Just like `AbstractController` itself does. There are other ideas like `AbstractJobType` etc. But if the code needs the service, it should always use constructor injection.
Probably not in this PR though ;)

aschempp (approved):

core-bundle/README.md
aschempp: Looks correct to me

Toflar (approved):

core-bundle/README.md
bytehead: Yep, seems to be correct.

bytehead (approved):

PR #3758, opened 2021-12-01
Add transitive dependency on symfony/twig-bridge

richardhj:
| Q                | A
| -----------------| ---

We use functions like `{% trans_default_domain %}` which are part of `symfony/twig-bridge`. As a consequence, we have a transitive dependency on this package. This PR adds the missing dependency.

I noticed this issue when (while being on 4.13@dev) I was updated on symfony/twig-bridge v6.0 which is currently incompatible with Contao (because of the usage of AnonymousToken, for example).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -104,6 +104,7 @@
         "symfony/stopwatch": "4.4.*",
         "symfony/swiftmailer-bundle": "^2.6.2 || ^3.1.5",
         "symfony/translation": "4.4.*",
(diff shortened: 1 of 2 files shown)


m-vo (approved):

Toflar (approved):

bytehead (approved):

leofeyer:
Thank you @richardhj.

PR #3759, opened 2021-12-01
Unify the MakeServicesPublicPass

bytehead:
See https://github.com/contao/contao/pull/3743#discussion_r759472200

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
@@ -44,9 +44,7 @@ class MakeServicesPublicPass implements CompilerPassInterface
             if ($container->hasAlias($id)) {
                 $alias = $container->getAlias($id);
                 $alias->setPublic(true);
-            }
-
-            if ($container->hasDefinition($id)) {
+            } elseif ($container->hasDefinition($id)) {
                 $definition = $container->getDefinition($id);
                 $definition->setPublic(true);
             }
--- a/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
@@ -24,18 +24,18 @@ class MakeServicesPublicPass implements CompilerPassInterface
 {
     private const IDS = [
         'assets.packages',
+        'database_connection',
         'debug.stopwatch',
         'fragment.handler',
         'lexik_maintenance.driver.factory',
         'monolog.logger.contao',
         'security.authentication.trust_resolver',
-        'security.firewall.map',
-        'security.logout_url_generator',
-        'security.helper',
-        'uri_signer',
-        'database_connection',
-        'swiftmailer.mailer',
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):
I think this is the correct way to go. Aliases are exactly here to make sure that old service references still work. So for our compiler pass it shouldn't matter which one is a service and which one is an alias. They just have to end up being public.

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
```diff
@@ -22,43 +22,34 @@
...
             }
 
-            $definition = $container->getDefinition($service);
-            $definition->setPublic(true);
-        }
-
-        foreach (self::ALIASES as $alias) {
-            if (!$container->hasAlias($alias)) {
-                continue;
+            if ($container->hasDefinition($id)) {
```
leofeyer: Should this be an `elseif` condition, so we do not look for definitions if the ID is an alias?

bytehead: I don't like `elseif`, if the `$id` is an alias then `$container->hasDefinition($id)` will always evaluate to false.

leofeyer: Yes, but if we already know that the ID is an alias, the `hasDefinition()` call is unnecessary, isn't it?

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
```diff
@@ -22,43 +22,34 @@
...
 
     public function process(ContainerBuilder $container): void
     {
-        foreach (self::SERVICES as $service) {
-            if (!$container->hasDefinition($service)) {
-                continue;
+        foreach (self::IDS as $id) {
+            if ($container->hasAlias($id)) {
+                $alias = $container->getAlias($id);
+                $alias->setPublic(true);
```
bytehead, line 46:
```suggestion
                $alias->setPublic(true);
                
                continue;
```

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
bytehead: Maybe like https://github.com/contao/contao/pull/3759#discussion_r760963494?

bytehead: It's IMO a personal thing. My eyes always hurt a bit if I see `elseif` when not really needed 😅

core-bundle/src/DependencyInjection/Compiler/MakeServicesPublicPass.php
leofeyer: That's even worse IMHO. No hard feelings. 😄

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #3765, opened 2021-12-02
Do not deprecate the autowiring aliases

leofeyer:
Follow-up on #3764

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -963,6 +963,9 @@ services:
         alias: contao.opt_in
         public: true # backwards compatibility
     Contao\CoreBundle\Picker\PickerBuilderInterface: '@contao.picker.builder'
+    Contao\CoreBundle\Routing\Page\PageRegistry:
+        alias: contao.routing.page_registry
+        public: true # backwards compatibility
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/config/services.yml
```diff
@@ -1039,14 +1046,6 @@ services:
 
     Contao\CoreBundle\Framework\ContaoFrameworkInterface: '@contao.framework'
```
aschempp, line 1040: This should be deprecated, we don't use that interface. Not sure it was ever available for autowiring?

aschempp (commented):
Can you please also add autowiring for the PageRegistry?

core-bundle/src/Resources/config/services.yml
leofeyer: Yes, it was.

https://github.com/contao/contao/blob/904fd1846abbb051783b8b6d043abaffab3553bb/core-bundle/src/Resources/config/services.yml#L29-L30

I have just moved it down into the "backwards compatibility" section now.

leofeyer:
> Can you please also add autowiring for the PageRegistry?

Added in 1852f1490272a88cd9df93f9de96a208ba233de6.

aschempp (approved):

PR #3766, opened 2021-12-03
Remove page from index if "Do not search" is checked

### LINKED ISSUE #3740: Remove page from index, if `Do not search` is checked
**Affected version(s)**

Contao 4.9.22

**Description**

Remove page from the search index, if it is already indexed by Contao, when in afterwards `Do not search` is checked from the pages setting.

tsarma:
As pointed out here, I am proposing this pull request
https://github.com/contao/contao/issues/3740#issuecomment-984652753

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -38,7 +38,8 @@ $GLOBALS['TL_DCA']['tl_page'] = array
 		'onsubmit_callback' => array
 		(
 			array('tl_page', 'scheduleUpdate'),
-			array('tl_page', 'generateArticle')
+			array('tl_page', 'generateArticle'),
+			array('tl_page', 'removeIndexEntry')
 		),
 		'sql' => array
 		(
@@ -1119,6 +1120,23 @@ class tl_page extends Contao\Backend
 		$objSession->set('sitemap_updater', array_unique($session));
 	}
 
+	/**
+	 * Remove already indexed page, if Do not search is checked
+	 */
+	public function removeIndexEntry(Contao\DataContainer $dc)
+	{
+		// Return if there is no ID
+		if (!$dc->activeRecord || !$dc->activeRecord->id) {
+			return;
+		}
+
+		if ($dc->activeRecord->noSearch || $dc->activeRecord->type != 'regular') {
+			$objPage = Contao\PageModel::findById($dc->activeRecord->id);
+			$pageUrl = $objPage->getAbsoluteUrl();
+			Contao\Search::removeEntry($pageUrl);
+		}
+	}
+	
 	/**
 	 * Auto-generate a page alias if it has not been set yet
 	 *


core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -1119,6 +1120,23 @@ public function scheduleUpdate(Contao\DataContainer $dc)
...
+	{
+		// Return if there is no ID
+		if (!$dc->activeRecord || !$dc->activeRecord->id) {
+			return;
+		}
+
+		if ($dc->activeRecord->noSearch || $dc->activeRecord->type != 'regular') {
+			$objPage = Contao\PageModel::findById($dc->activeRecord->id);
+			$pageUrl = $objPage->getAbsoluteUrl();
+			Contao\Search::removeEntry($pageUrl);
```
ausi, line 1136: I think we should use the `delete()` method of ~~the `contao.search.delegating_indexer` service instead here.~~

Toflar (changes requested):
Thanks for the contribution :) This definitely cannot be merged as is, though :) We need

* A separate DCA callback listener. No `tl_page.php` adjustments needed. You can find an example in core e.g. here: https://github.com/contao/contao/blob/4.x/core-bundle/src/EventListener/DataContainer/PageTypeOptionsListener.php#L24-L26
* This new listener needs to get the `contao.search.indexer` injected.
* It needs unit tests.

core-bundle/src/Resources/contao/dca/tl_page.php
Toflar: No, the `contao.search.indexer`. :)

SeverinGloeckle:
While the adjustment itself sounds very good, I unfortunately see some conceptual problems:

- Isn't it likely that an editor would assume that a page that is removed from the index when `Do not search` is activated is automatically added back to the index when the option is deactivated?
- If the now-to-be-ignored page has items (e.g. a news reader page), not only the URI of the actual page should be removed from the index, but all its items as well. In fact, the page itself may be unroutable and thus not even have a URI at all.

An alternative approach would be to filter pages with `Do not search` when retrieving the actual search results (similar to protected pages). However, this would be rather slow for larger websites since the actual page record has to be checked for each search result.

fritzmg:
> Isn't it likely that an editor would assume that a page that is removed from the index when `Do not search` is activated is automatically added back to the index when the option is deactivated?

Since pages aren't immediately added to the search index in any case (only when they are visited in the front end), I would assume no.

> If the now-to-be-ignored page has items (e.g. a news reader page), not only the URI of the actual page should be removed from the index, but all its items as well. In fact, the page itself may be unroutable and thus not even have a URI at all.

That's not a problem. Instead of removing by URL, the entries should just be deleted based on the `pageId`, which is already stored with each `tl_search` entry.

core-bundle/src/Resources/contao/dca/tl_page.php
fritzmg: The pages need to be deleted according to the page ID, which is currently not possible via the `contao.search.indexer`.

Toflar: That's incorrect though. For the search indexer we are only allowed to work with external URLs, not page IDs. It's 100% abstracted and must not know anything about any additional meta data.

aschempp:
We already have a listener that purges the search index if a page is deleted (or the alias is changed), and it intentionally does only work on the database because we purge by page ID (like @fritzmg) mentioned. The `robots` handling should simply be added to https://github.com/contao/contao/blob/4.x/core-bundle/src/EventListener/DataContainer/PageUrlListener.php

Toflar:
That listener is wrong too then.

aschempp:
Can we fix the issue at hand in the already provided implementation and you can rewrite it to a perfect score in a later PR? 😉

core-bundle/src/Resources/contao/dca/tl_page.php
fritzmg: But then it would be impossible to accurately and quickly delete entries from the search index. If you have a page with a newsreader on it, then all the news URLs will need to be purged as well if you enable `noSearch` for the page. How would you gather the URLs to delete then? And the page's original URL will not even be present if `requireItem` is active.

Toflar: Yes, it's impossible currently. And that's correct.

Toflar:
I'll take care and provide a PR at some point because it seems I'm the only one who knows what to do in terms of keeping our search index abstraction intact :)

Toflar:
So I think we can split the fix here in two TODOs:

1. The original issue by @tsarma to also purge the search index when `noSearch` is changing
2. My problem, that this listener deals with the search index entirely wrong.

I'll fix 2 myelf at some point but @tsarma, if you want to, you can still fix your problem :-)
What you need to do in my opinion is the following:

1. Register another callback on the `noSearch` field, just like we do for the alias change: https://github.com/contao/contao/blob/4.x/core-bundle/src/EventListener/DataContainer/PageUrlListener.php#L81-L93 - check whether the value has changed and if so, call `$this->purgeSearchIndex((int) $dc->id)` :) 
2. Write a unit test to cover your new callback, essentially it's also more or less copying https://github.com/contao/contao/blob/4.x/core-bundle/tests/EventListener/DataContainer/PageUrlListenerTest.php#L753-L788

core-bundle/src/Resources/contao/dca/tl_page.php
ausi: Can’t we just update the delete method of the indexer? Because the page ID is known from the JSON-LD so removing all pages that are related to a page ID should be no problem: https://github.com/contao/contao/blob/927a912b96d98499be6b589c32623434d7a30b17/core-bundle/src/Search/Indexer/DefaultIndexer.php#L110-L116

Toflar: Yes, that's what I'll do :)

tsarma:
Hi there, 
I am trying this callback on field `noSearch`, Contao 4.9.23 and it's not calling for the reason I don't know! Any idea?

```
<?php

declare(strict_types=1);

/*
 * This file is part of Contao.
 *
 * (c) Leo Feyer
 *
 * @license LGPL-3.0-or-later
 */

namespace Contao\CoreBundle\EventListener\DataContainer;

use Contao\DataContainer;
use Contao\Search;
use Doctrine\DBAL\Connection;
use Contao\CoreBundle\Framework\ContaoFramework;

/**
 * @Callback(table="tl_page", target="fields.noSearch.save")
 */
class PageSearchSettingListenter
{
    private ContaoFramework $framework;

    public function __construct(ContaoFramework $framework, Connection $connection)
    {
        $this->framework = $framework;
        $this->connection = $connection;
    }


    public function __invoke($value, DataContainer $dc)
    {
        if ($dc->activeRecord->noSearch || $dc->activeRecord->type != 'regular') {
            $this->purgeSearchIndex((int) $dc->activeRecord->pid);
        }

        return $value;
    }

    public function purgeSearchIndex(int $pageId): void
    {
        $urls = $this->connection->fetchFirstColumn(
            'SELECT url FROM tl_search WHERE pid=:pageId',
            ['pageId' => $pageId]
        );

        $search = $this->framework->getAdapter(Search::class);

        foreach ($urls as $url) {
            $search->removeEntry($url);
        }
    }
}

```

ausi:
@tsarma did you clear the container cache? (/var/cache/*)

SeverinGloeckle:
You have to explicitly define the listener in `listener.yml`, similar to this:

https://github.com/contao/contao/blob/904fd1846abbb051783b8b6d043abaffab3553bb/core-bundle/src/Resources/config/listener.yml#L107-L115

and then clear the application cache.

By the way, there's a typo in your class name, PageSearchSettingListen**~~t~~**er.

tsarma:
Thank you @SeverinGloeckle, @ausi  , Now it is calling my class. I've also corrected the typo too.

aschempp:
Closed in favour of https://github.com/contao/contao/pull/3839

PR #3767, opened 2021-12-03
Run the code inspector

leofeyer:


news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
```diff
@@ -16,13 +16,15 @@
...
- * Front end module "news reader".
+ * Front end module "newsreader".
  *
  * @property Comments $Comments
  * @property string   $com_template
  * @property array    $news_archives
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ *
+ * @todo Rename to ModuleNewsreader in Contao 5.0
```
fritzmg, line 27: Why? Then we also need to rename `ModuleNewsList` to `ModuleNewslist`, don't we?

ausi (approved):

news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
leofeyer: No, because "newslist" is not a word but "newsreader" is. 🤷‍♂️

leofeyer: But that also seems debatable.

leofeyer: It seems that "newsreader" means "Nachrichtensprecher", so we should actually stick with "news reader". 🤦‍♂️ I am going to adjust the code inspector accordingly.

leofeyer: Reverted in f6a1068b74b819a2874aa9f74f25b4d2169b92ed.

fritzmg: I see - yes that should be changed.

PR #3769, opened 2021-12-03
Run the code inspector (again)

leofeyer:


calendar-bundle/src/EventListener/PreviewUrlConvertListener.php
```diff
@@ -41,7 +41,7 @@ public function __invoke(PreviewUrlConvertEvent $event): void
 
         $request = $event->getRequest();
 
-        if (null === $request || null === ($eventModel = $this->getEventModel($request))) {
```
leofeyer, line 44: Because `$event->getRequest()` no longer returns `null`.

core-bundle/src/EventListener/Security/SwitchUserListener.php
```diff
@@ -46,11 +45,7 @@ public function __invoke(SwitchUserEvent $event): void
         }
 
         $sourceUser = $token->getUserIdentifier();
-        $targetUser = $event->getTargetUser();
-
-        if ($targetUser instanceof UserInterface) {
-            $targetUser = $targetUser->getUserIdentifier();
-        }
+        $targetUser = $event->getTargetUser()->getUserIdentifier();
```
leofeyer, line 48: Because `getTargetUser()` always returns an instance of `UserInterface` now.

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -391,10 +391,6 @@ public function buildInheritanceChains(): void
      */
     private function getThemeTemplateName(string $name): ?string
     {
-        if (!\is_string($name)) {
```
leofeyer, line 394: I forgot to do this (see https://github.com/contao/contao/pull/3343/files#diff-e36795bda9eb0543af8a7afe0e4f71cd1e57f8d12fcf3bb49c8a2cdc8f2df976R402).

installation-bundle/src/InstallTool.php
```diff
@@ -101,7 +101,7 @@ public function canConnectToDatabase(?string $name): bool
         } catch (\Exception $e) {
         }
 
-        if (null === $name || null === $this->connection) {
```
leofeyer, line 104: Because `$this->connection` can never be `null`.

calendar-bundle/src/EventListener/PreviewUrlConvertListener.php
```diff
@@ -41,7 +41,7 @@ public function __invoke(PreviewUrlConvertEvent $event): void
...
+        if (null === ($eventModel = $this->getEventModel($request))) {
```
ausi: https://github.com/contao/contao/pull/3769/files#r761986625
```suggestion
        if (null === ($eventModel = $this->getEventModel($event->getRequest()))) {
```
We could directly inline it then I think?

core-bundle/tests/Doctrine/Backup/BackupManagerTest.php
```diff
@@ -428,8 +428,13 @@ private function getBackupDir(): string
...
     {
-        $connection = $connection ?? $this->createMock(Connection::class);
-        $dumper = $dumper ?? $this->createMock(DumperInterface::class);
+        if (null === $connection) {
+            $connection = $this->createMock(Connection::class);
+        }
+
+        if (null === $dumper) {
+            $dumper = $this->createMock(DumperInterface::class);
+        }
```
ausi:
```suggestion
        $connection ??= $this->createMock(Connection::class);
        $dumper ??= $this->createMock(DumperInterface::class);
```

ausi: This applies to all code blocks of this kind in this PR I think.

fritzmg: Why is this change necessary in the first place though? I mean what's the problem with the original version?

leofeyer: This is the original problem:



But @ausi's suggestion to use `??=` is an even nicer solution.

leofeyer: Changed in f2c019aec55bade213d1f2dc37f079a4a1a031e6.

calendar-bundle/src/EventListener/PreviewUrlConvertListener.php
leofeyer: Changed in f2c019aec55bade213d1f2dc37f079a4a1a031e6.

ausi (dismissed):

core-bundle/tests/Cache/EntityCacheTagsTest.php
```diff
@@ -258,10 +258,9 @@ public function testDelegatesToCacheInvalidator(): void
...
     {
-        return new EntityCacheTags(
-            $this->getTestEntityManager(),
-            $responseTagger ?? $this->createMock(ResponseTagger::class),
-            $cacheInvalidator ?? $this->createMock(CacheInvalidator::class)
-        );
+        $responseTagger ??= $this->createMock(ResponseTagger::class);
+        $cacheInvalidator ??= $this->createMock(CacheInvalidator::class);
+
+        return new EntityCacheTags($this->getTestEntityManager(), $responseTagger, $cacheInvalidator);
```
m-vo: IMHO the old version was easier to read. :see_no_evil: 


I think I prefer this…

```php 
doFoo(
    $firstArgument ?? new Fallback(),
    $secondArgument ?? $alternative,
);
```

over this

```php 
$firstArgument ??= new Fallback();
$secondArgument ??= $alternative;

doFoo($firstArgument, $secondArgument);
```

in general.

It's basically the same amount of code but with the first option, the place where the "fallback chain" is put, is exactly the place it is used. The code basically reads "use this as the argument, fallback to this, fallback to that, …".

In the second one, you do not know the usage (yet) when reading the assignments. Another point: Why would you overwrite an argument if not needed? WIth `$a ??= $b` the original value of `$a` is gone.

ausi: I think it could also be read as default values for the parameters, which is where the `$this->createMock(…)` would go if PHP would support expressions as default values. From that perspective using `??=` at the beginning of the method makes sense IMO.

But I would intuitively also write it in the style the old version was.
But I’m happy either way ☺️

core-bundle/tests/EventListener/Security/TwoFactorFrontendListenerTest.php
```diff
@@ -552,12 +547,7 @@ private function mockScopeMatcher(bool $hasFrontendUser, RequestEvent $event): S
     private function getRequestEvent(Request $request = null, Response $response = null): RequestEvent
     {
         $kernel = $this->createMock(Kernel::class);
-
-        if (null === $request) {
-            $request = new Request();
-        }
-
-        $event = new RequestEvent($kernel, $request, HttpKernelInterface::MAIN_REQUEST);
+        $event = new RequestEvent($kernel, $request ?? new Request(), HttpKernelInterface::MAIN_REQUEST);
```
bytehead, line 550: To apply same codestyle as (e.g.) in https://github.com/contao/contao/pull/3769/files#diff-5ced46fdf62feba7ee121f368149866f881e3047bdf98517b2cbad4ccc1e5487R277:

```suggestion
        $request ??= new Request();

        $event = new RequestEvent($kernel, $request, HttpKernelInterface::MAIN_REQUEST);
```

core-bundle/tests/OptIn/OptInTokenTest.php
```diff
@@ -306,10 +306,6 @@ public function testDoesNotRequireSubjectAndTextToResendToken(): void
 
     private function getToken(OptInModel $model, ContaoFramework $framework = null): OptInTokenInterface
     {
-        if (null === $framework) {
-            $framework = $this->mockContaoFramework();
-        }
-
-        return new OptInToken($model, $framework);
+        return new OptInToken($model, $framework ?? $this->mockContaoFramework());
```
bytehead, line 309: Same codestyle here:

```suggestion
        $framework ??= $this->mockContaoFramework();

        return new OptInToken($model, $framework);
```

core-bundle/tests/String/SimpleTokenParserTest.php
```diff
@@ -644,10 +644,6 @@ public function getFunctions(): array
 
     private function getParser(SimpleTokenExpressionLanguage $expressionLanguage = null): SimpleTokenParser
     {
-        if (null === $expressionLanguage) {
-            $expressionLanguage = new SimpleTokenExpressionLanguage();
-        }
-
-        return new SimpleTokenParser($expressionLanguage);
+        return new SimpleTokenParser($expressionLanguage ?? new SimpleTokenExpressionLanguage());
```
bytehead, line 647: Same codestyle here:

```suggestion
        $expressionLanguage ??= new SimpleTokenExpressionLanguage();

        return new SimpleTokenParser($expressionLanguage);
```

manager-bundle/tests/Api/ApplicationTest.php
```diff
@@ -140,10 +140,6 @@ public function testThrowsExceptionIfPluginReturnsInvalidCommand(): void
 
     private function getApplication(string $path = null): Application
     {
-        if (null === $path) {
-            $path = $this->getTempDir();
-        }
-
-        return new Application($path);
+        return new Application($path ?? $this->getTempDir());
```
bytehead, line 143: Same here:

```suggestion
        $path ??= $this->getTempDir();

        return new Application($path);
```

bytehead (changes requested):
Looks good to me apart the inconsistent code style issues mentioned.

core-bundle/tests/Cache/EntityCacheTagsTest.php
leofeyer: > Another point: Why would you overwrite an argument if not needed? WIth `$a ??= $b` the original value of `$a` is gone.

That's not a point, because the same happens with the old `if (null === $a) { $a = $b; }` code.

leofeyer: I have reverted some of the changes in 99c07707065b4e90a76685af287ca03db8156e04 to explicitly allow the `$var ?? $this->createMock()` construct.

bytehead (approved):

ausi (approved):

PR #3774, opened 2021-12-06
Add a DBAFS service and integrate Flysystem

m-vo:
| Fixed issues     | see https://github.com/contao/docs/pull/1001    |
| Docs PR or issue | todo |

Currently code deals with files and file metadata by directly using PHP filesystem functions, querying/updating `tl_files`, using `FilesModel`s and the `Dbafs` class. With this PR, there will be the concept of a virtual filesystem (VFS) which aims at getting your only interfaces for files and file metadata in the future. 

### Architecture
There are 3 main components:
 
1) The `MountManager` manages access to multiple [Flysystem](https://flysystem.thephpleague.com/v2/docs/) adapters each bound to a virtual path - the mount path. The `MountManager` looks very similar to a Flysystem `FilesystemAdapter` but it a) completely abstracts the Flysystem so that we can exchange it or upgrade to a newer version in the future and b) it selects the right adapter based on the given path. If adapter *A* was mounted to `files` and adapter *B* was mounted to `files/media`, a call to `files/foo.jpg` would resolve to `foo.jpg` in adapter *A* while `files/media/videos/cat.mov` would resolve to `videos/cat.mov` in adapter *B*.

2) The `DbafsManager` allows DBAFS services to be registered (also with a prefix path). This also looks similar to a filesystem interface but also supports UUIDs along paths to locate resources and has methods to get and set extra metadata. Calls will resolve to the right DBAFS by prefix (in case of paths) or by searching across multiple in case of UUIDs (which are unique system-wide).

3) Finally, the `VirtualFilesystem` is the user interface to both worlds: filesystems and DBAFS. We typically have a single `MountManager` and a single `DbafsManager` that is injected in each `VirtualFilesystem` instance. In each? Yes, you can have as many VFS as you like: Every instance can be constrained to a path prefix and can optionally be made readonly. So if you're dealing with, say, image assets you might want to have a `$imageAssetsStorage` where each call will automatically be prefixed with `assets/images` and where you cannot escape.

### Configuration
There is no bundle configuration, instead you can use the `FilesystemConfiguration` helper object that sets the appropriate things in the container. The easiest way to do so is by implementing the `ConfigureFilesystemInterface` in your extension class:

```php
class FooExtension extends Extension implements ConfigureFilesystemInterface
{
    public function configureFilesystem(FilesystemConfiguration $config): void
    {
        $config
            // Mount a local adapter to 'files/foo'
            ->mountLocalAdapter('files/foo', '/here/is/the/real/path')

            // Add a new readonly VFS that can be injected with 'VirtualFilesystemInterface $fooStorage'
            ->addVirtualFilesystem('foo', 'files/foo', $readonly = true)
        ;

        // for more things to configure, have a look at the 'FilesystemConfiguration' class
    }

    // …
}
```

With this, it should be very easy for extension developers to add and mount own adapters, DBAFS services and virtual filesystems.

### Usage
Here is a short example on how you would interact with a VFS, in this case the `filesStorage` which is prefixed to `files` and made available by the core. 

```php
use Contao\CoreBundle\Filesystem\VirtualFilesystemInterface;
use Symfony\Component\Uid\Uuid;

class Foo
{   
    public function __construct(private VirtualFilesystemInterface $filesStorage) {}

    /** Using a special storage **/
    public function demo1(Uuid $uuid): void {
        // Check if files/test.txt exists - either by path or UUID
        $this->filesStorage->fileExists('test.txt');
        $this->filesStorage->fileExists($uuid);

        // Read/move/write
        $this->filesStorage->read('foo.dat');
        $this->filesStorage->move(new Uuid('d6c8c82a-e588-11eb-9478-e7674e53b388'), 'foo/test.txt');
        $this->filesStorage->write('test.txt', 'some content');

        // Use metadata
        $data = $this->filesStorage->getExtraMetadata('images/cat.jpg');
        $data['metadata']['en']; // Metadata class
        $data['importantPart']; // ImportantPart class

        // List directories
        foreach($this->filesStorage->listContents('foo') as $entry) {
            $entry->getPath(); // string            
            $entry->isFile(); // bool
            $entry->getExtraMetadata(); // like above
            // …
        }
}
```

### DBAFS
The DBAFS functionality got a major makeover. Here are some facts about the default implementation, that is now already used in the `filesync` command:

* You can specify one or multiple files/directories as a scope. This allows super fast partial synchronization that still keeps directory hashes intact. If you would execute `contao:filesync foo/* bar/baz/* images/cat.jpg` everything under the directory `foo` the direct children of `bar/baz` as well as the single file `images/cat.jpg` would get synchronized (possibly including parent directories). In fact this is used by the `VirtualFilesystem`, which automatically does partial syncs when creating/deleting/copying/moving files.

* The hash generation is abstracted into a service. This for instance allows to create individual hash functions. The default implementation already supports different hash functions, so you could for instance configure to use the much faster `xxh128` when using PHP8.1. 

* The computed changes (when synchronizing) are put in a `ChangeSet` (container class), that is both used to update the database as well as generate reports. The command now uses this to print a nice summary: 

    image

* Our `Dbafs` class now supports last modified timestamps (can be enabled/disabled) - this allows skipping rehashing if the file mtime didn't change since the last time.

The way the new service works (considering the database) is fully BC with the current implementation but **a lot** faster.

#### A word on models/entities
It's IMHO wrong to use models or entities for anything related to file metadata. This does not allow any filesystem abstraction and it's the wrong ownership (Contao model system / doctrine). This is why we should declare `tl_files` (or however it will be called in the future) as an implementation detail and drop the `FilesModel`. You should only use the DBAFS to access and alter metadata and files themselves.

#### Extra Metadata
There are two new events `contao.filesystem.retrieve_dbafs_metadata` and `contao.filesystem.store_dbafs_metadata` that are dispatched after/before metadata is loaded from/stored into the database by our default `Dbafs` service. When retrieving, the data can be enhanced (e.g. `ImportantPart` and `File\Metadata` classes instead of arrays), while when storing, it must be normalized into values that can be stored in the database again. This, together with the ability to set the DBAFS table allows for a simple way to build your own DBAFS storage/table with your own attributes.


core-bundle/src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,416 @@
...
+use League\Flysystem\InvalidStreamProvided;
+use League\Flysystem\PathNormalizer;
+use League\Flysystem\StorageAttributes;
+use League\Flysystem\WhitespacePathNormalizer;
+use ProxyManager\Factory\LazyLoadingGhostFactory;
+use ProxyManager\Proxy\GhostObjectInterface;
+
+/**
+ * This Flysystem operator wraps a DBAFS layer around the underlying filesystem
+ * adaptor: Listing content will only access the database, operations that can
```
Toflar:
```suggestion
 * adapter: Listing content will only access the database, operations that can
```

core-bundle/src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,416 @@
...
+ * Additionally, there is support to directly retrieve the 'extra metadata'
+ * from a file's / directory's StorageAttributes via a dedicated method.
+ *
+ * We widen the FilesystemOperator interface - instead of calling methods with
+ * resource paths (string), you can also pass in an ID (int) or UUID of a
+ * synchronized DBAFS resource.
+ *
+ * @phpstan-import-type ExtraMetadata from Dbafs
+ */
+final class Filesystem implements FilesystemOperator
```
Toflar: The biggest problem I have with this is the name. Our extension developers will associate `Filesystem` with the local drive for sure. Especially because it's already taken by Symfony's `Filesystem` class which indeed provides methods to operate on the local drive. What about we call it what it really is `VirtualFilesystem`? It would also make a lot of things clearer.

core-bundle/src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,416 @@
...
+        $this->config = $config ?? new Config();
+        $this->pathNormalizer = $pathNormalizer ?? new WhitespacePathNormalizer();
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param string|int $location
+     */
+    public function fileExists($location): bool
```
Toflar: All the methods should allow to force ignoring the Dbafs if they want to. It should not be the default though.

core-bundle/src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,416 @@
...
+    private FilesystemAdapter $adapter;
+    private Config $config;
+    private PathNormalizer $pathNormalizer;
+    private Dbafs $dbafs;
+    private LazyLoadingGhostFactory $proxyFactory;
+
+    public function __construct(Dbafs $dbafs, FilesystemAdapter $adapter, LazyLoadingGhostFactory $proxyFactory = null, Config $config = null, PathNormalizer $pathNormalizer = null)
+    {
+        $this->dbafs = $dbafs;
+        $this->adapter = $adapter;
```
Toflar: In order to really benefit from `Flysystem` I think we need to have the `Mount Manager` as first class citizen here. We don't have to implement it in core now but I think it's a fairly safe bet that we want to have this in the future. So that I can mount `/files/dropbox` to some dropbox account and `/files/s3` to some s3 bucket.
I'm not sure if the access details etc. have to go to the `Dbafs` or if it should reside in files like `.public` to ensure access is availalbe also without `Dbafs`. However, I'm rather sure, we cannot use the protocol specification `Flysystem` uses by default. Hence, I'd love to already address this now because now is the time to think about this conceptually.

core-bundle/src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,416 @@
...
+    {
+        return empty($options) ? $this->config : $this->config->extend($options);
+    }
+
+    /**
+     * Returns a @see StorageAttributes ghost object that only has the path and
+     * extra metadata set. Accessing any other property will trigger fully
+     * initializing it and thus asking the underlying adapter for attributes.
+     */
+    private function getLazyStorageAttributes(array $record): StorageAttributes
```
Toflar: Fancy stuff! <3

Toflar (commented):
I would also question whether we need a bundle at all. Given the fact that we just had this discussion again with getting rid of e.g. the Lexik Bundle due to it not fitting our exact needs. If we have our own way to wire other adapters we don't really need the stuff provided by bundles, do we?

core-bundle/src/Filesystem/Filesystem.php
m-vo: :+1: already added to the todos :slightly_smiling_face:

core-bundle/src/Filesystem/Filesystem.php
m-vo: Thought about this as well and just went with `Filesystem` because Flysystem did. 

I like `VirtualFilesystem`.

Alternatives would maybe be:
* `Flysystem`
* `FilesystemOperator`
* `FlysystemOperator`

We might also want to reword `Dbafs` in this go because it's less of a **f**ile**s**ystem but rather a "metadata storage indexed by paths".

Toflar: I don't care about `Dbafs` that much because it's an internal implementation detail now :) Which is great by the way!
I prefer `VirtualFilesystem` over the others. I'm trying to imagine the users of that service and for them, I think `VirtualFilesystem` tells a lot more than `Flysystem` (does it fly?) or `FlysystemOperator` (so it not only flies but it also has an operator that flies it?).
Of course we'll have devs that know exactly what `Flysystem` is. But they'll also understand `VirtualFilesystem` ;)

core-bundle/src/Filesystem/Filesystem.php
m-vo: In the flysystem bundle there is a `\League\Flysystem\Filesystem` operator instance for each configured adapter + some magic in the extension, that makes the DI component inject the right one according to how you name your variable (so `default_storage` for `Filesystem $defaultStorage`). 

It's already possible if you manually configure it, but we could do the same magic for our operator for sure. You should then get another DBAFS service instance with its own database as well.

The mount manager is another layer on top of filesystem operators, so this should IMHO work out of the box (minus extra functionality built into ours like a flag to bypass the DBAFS).

Toflar: Yeah that's exactly what I'm saying. I don't think this makes sense. In our case, the only service any developer ever needs is the `VirtualFilesystem` and the different sources should be managed inside it based on the path. Hence we don't need any bundle imho.

core-bundle/src/Filesystem/Filesystem.php
m-vo: I named it `DbafsFilesystem` for now. Lew's discuss this again after the mount manager / virtual filesystem concept is established.

core-bundle/src/Filesystem/Filesystem.php
m-vo: I started outlining a mount manager concept. I will update the description once there is more to see and then we can discuss based on this new approach.

m-vo:
I updated the initial description to reflect the new implementations.

core-bundle/src/Event/DbafsMetadataEvent.php
```diff
@@ -0,0 +1,80 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Event;
+
+class DbafsMetadataEvent
```
m-vo: Note to myself: there should probably be two events: one for normalization, one for denormalization. (Maybe use serializer component?)

core-bundle/src/Filesystem/VirtualFilesystem.php
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Example:
+ *   If operator A is mounted to 'files' and operator B to 'files/media',
+ *   calling read('files/media/foo') will issue a call to B->read('foo')
+ *   while calling fileExists('files/bar/baz') will issue a call to
+ *   A->fileExists('bar/baz') instead.
+ *
+ * todo: handle and rethrow exceptions
+ */
+class VirtualFilesystem implements FilesystemOperator
```
m-vo: Note to myself: both the `DbafsFilesystem` and the `VirtualFilesystem` should probably implement an interface that includes metadata support. Maybe `DbafsFilesystemOperator extends FilesystemOperator`?

core-bundle/src/Filesystem/DbafsFilesystem.php
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    /**
+     * Resolve DBAFS IDs and UUIDs to paths. If a path was given, normalize it.
+     *
+     * @param string|int $identifier
+     */
+    private function resolvePath($identifier): string
+    {
+        if (is_numeric($identifier)) {
```
ausi: This could be problematic I think. A folder with the name `123` would be inaccessible, same if someone names a file or folder afer a UUID.

How about using a value object that represents a uuid or an ID? `DbafsPath::fromUuid(…)`, `DbafsPath::fromId(…)`?

m-vo: The problem with stringable objects is that you cannot pass them though abstractions that annotate `string` instead of `stringable`. Think of any regular `FilesystemOperator` or our own `VirtualFilesystem` with prefix paths.

We could maybe require a certain format that is more unlikely to match a folder name: `foo/bar/#123` or `[123]` for ID 123. :thinking: I'm not sure if UUIDs are a problem. Why would you name a folder like this? (matching an actual entry in this exact DBAFS table)

ausi: How about a protocol prefix? `dbafs:123` and `dbafs:d6c8c82a-e588-11eb-9478-e7674e53b388`? Or `dbafs://123` and `dbafs://d6c8c82a-e588-11eb-9478-e7674e53b388`?

m-vo: Works for me, too. I'm not sure the protocol in the middle of a string looks "correct" to me, but this could only need some getting used to it.

```php
$fooStorage->extraMetadata("dbafs://$uuid");
$virtualFilesystem->read("foo/bar/dbafs://$uuid");
```

ausi: If it can appear inside of other paths I would prefer a folder prefix I think like `~/123` and `~/d6c8c82a-e588-11eb-9478-e7674e53b388`. This way I can also list all IDs and files by listing the contents in the `~` folder.

ausi: Or maybe two specific prefixes: `~id/123` and `~uuid/d6c8c82a-e588-11eb-9478-e7674e53b388`

m-vo: As briefly discussed with @ausi on slack we could/should probably make UUIDs first class citizens of an interface that both the `DbafsFilesystem` and the `VirtualFilesystem` implements. They should be unique across all dbafs instances anyway and it's a necessity that we can reference them from content elements and other places. This would also mean you do not need a prefix for them in the `VirtualFilesystem`.

So we're just left with IDs. :thinking:

core-bundle/src/Filesystem/Filesystem.php
m-vo: implemented in e21aa8064f352740870ee50415f3d3d37121d29a

core-bundle/src/Event/DbafsMetadataEvent.php
m-vo: implemented in e21aa8064f352740870ee50415f3d3d37121d29a

core-bundle/src/Filesystem/VirtualFilesystem.php
m-vo: implemented in e21aa8064f352740870ee50415f3d3d37121d29a

core-bundle/src/Filesystem/DbafsFilesystem.php
m-vo: I reworked the concept in e21aa8064f352740870ee50415f3d3d37121d29a. Now there are only UUIDs and they are "first class citizens".

m-vo:
Initial description is updated again to reflect the new changes/implementations:

 - make UUIDs first class citizens + use symfony/uid
 - introduce interfaces
 - add modes for SYNCED_ONLY/BYPASS_DBAFS
 - implement setExtraMetadata + add event to normalize data
 - rethrow exceptions with adjusted locations in the VirtualFilesystem

core-bundle/src/Filesystem/Dbafs.php
```diff
@@ -275,7 +273,9 @@ private function doComputeChangeSet(array $dbPaths, array $allDbHashesByPath, \G
 
         $isPartialSync = \count($dbPaths) !== \count($allDbHashesByPath);
 
-        foreach ($filesystemIterator as $path => $hash) {
```
m-vo: Memo to myself: revert this change again (was only intended for debugging))

core-bundle/src/Crawl/Escargot/Factory.php
```diff
@@ -93,7 +93,7 @@ public function createLazyQueue(): LazyQueue
     {
         return new LazyQueue(
             new InMemoryQueue(),
-            new DoctrineQueue($this->connection, static fn (): string => Uuid::uuid4()->toString(), 'tl_crawl_queue')
+            new DoctrineQueue($this->connection, static fn (): string => Uuid::v4()->toRfc4122(), 'tl_crawl_queue')
```
Toflar: Could be a separate little PR and merged real quick so it does not pollute the diff and gets its own changelog entry :)

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -62,6 +69,9 @@ public function prepend(ContainerBuilder $container): void
 
         // Prepend the backend route prefix to make it available for third-party bundle configuration
         $container->setParameter('contao.backend.route_prefix', $config['backend']['route_prefix']);
+
+        // todo: how do I do this correctly?
+        $container->setParameter('contao.upload_path', $config['upload_path']);
```
Toflar: Imho that is correct if you want it to be available for users to be used in their config.yml.

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -0,0 +1,103 @@
...
+use Contao\CoreBundle\Event\StoreDbafsMetadataEvent;
+use Contao\CoreBundle\File\Metadata;
+use Contao\Image\ImportantPart;
+use Contao\StringUtil;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * @internal
+ */
+class DbafsMetadataSubscriber implements EventSubscriberInterface
```
Toflar, line 26: I would consider making this one event because imho the 99% case is you need both, the serialization and unserialization part. Could be even handled with the symfony serializer which is already extensible but maybe that's a bit too much complexity and the single event is easier.

core-bundle/src/Filesystem/Dbafs.php
```diff
@@ -0,0 +1,863 @@
...
+     * @var array<string, string|null>
+     */
+    private array $pathByUuid = [];
+
+    /**
+     * @var array<int, string|null>
+     */
+    private array $pathById = [];
+
+    public function __construct(Connection $connection, EventDispatcherInterface $eventDispatcher, string $table, string $hashAlgorithm)
```
Toflar: Proposal: Let's make the hash algorithm its own service, it would provide multiple advantages:

```php
interface FileContentHasherInterface
{
    public function getHash(string $fileContents): string
}
```

The default service could work like this:

```php
class NativeFileContentHasher implements FileHasherInterface
{
    private ?string $nativeFunctionName = null;

    public function __construct(string $nativeFunctionName)
    {
        $this->nativeFunctionName = $nativeFunctionName;
    }

    public function getHash(string $fileContents): string
    {
        return hash($this->nativeFunctionName, $fileContents);
    }

    public static function createWithBestSuitedFromPlatform(): self
    {
        $algo = 'md5';

        if (PHP_VERSION_ID > 80100) {
            $algo = 'xxh128';
        }

        return new self($algo);
    }
}
```

Advantages:

- Easy to reuse across all `dbafs` configurations because they'll most likely use the same algorithm so they can all reuse the same service.
- Automatically uses the best native hash function available by default without us having to do anything (just use the factory method as service factory)
- Easy to test custom alternatives and see how they perform
- Better testability

Things to consider:

- What happens if the algorithm changes? No hash will match anymore, does that mean, we need a migration or can we do something like the password API (`password_*`) does internally. E.g. storing `$md5$<hash>` or `$xxh128$<hash>` and some `needsRehash()`? Basically really the same as the password API just not with cryptographically secure hashes?
- Just a wild thought: Might also be interesting to pass on the operator to that interface? I'm just thinking that maybe it's not needed to request a 4 GB file from an S3 storage in order to calculate it's hash if the operator already knows this and it's just metadata on the file. But then again, it's probably mixing different hash algorithms because S3 uses md5 and someone else something different. But there's always only one operator per file path so...could work ;)

core-bundle/src/Filesystem/UnableToResolveUuid.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Filesystem;
+
+use League\Flysystem\FilesystemException;
+use Symfony\Component\Uid\Uuid;
+
+class UnableToResolveUuid extends \RuntimeException implements FilesystemException
```
Toflar: Afaik we use the `Exception` suffix everywhere.

core-bundle/src/Filesystem/UnableToSetExtraMetadata.php
```diff
@@ -0,0 +1,23 @@
...
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Filesystem;
+
+use League\Flysystem\FilesystemException;
+
+class UnableToSetExtraMetadata extends \RuntimeException implements FilesystemException
```
Toflar: Same here.

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -615,6 +618,132 @@ private function addBackupNode(): NodeDefinition
...
+                            ->ifString()
+                            ->then(
+                                static fn (string $value): int => (int) (new NumberComparator($value))->getTarget()
+                            )
+                        ->end()
+                    ->end()
+                    ->integerNode('bulk_insert_size')
+                        ->info('We are using bulk inserts to improve the database performance when writing lots of entries. You can fine tune the size here.')
+                        ->min(1)
+                        ->defaultValue(100)
```
Toflar: Random thought: Would it make sense to have something like PHP's OPCache `validate_tstamps` setting here to check for the last modified date of the file when comparing the hashes.
Just because I want to omit downloading a 4 GB file from S3 to calculate the hash if in fact we knew that it's still going to be the same if the last modified date did not change. Should be enabled by default in order to increase performance and decrease traffic usage but one should be able to disable it if they know that a given file system does not update the last modified date for example. Wdyt about this idea?

core-bundle/src/Filesystem/UnableToResolveUuid.php
m-vo: I know, I was torn between our naming convention and placing the class in the `Exception` namespace versus following the flysystem convention https://github.com/thephpleague/flysystem/tree/2.x/src. Still not sure :stuck_out_tongue:.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
m-vo: Ah, I see, I'm using it in the manage-bundle, that is why it's needed.

By this way: Should we rather hardcode/prepend the config that the core will use? Or shall we document it + put it in the manager bundle skeleton config (like implemented now)? /cc @bytehead

core-bundle/src/Filesystem/Dbafs.php
m-vo: reverted in b71037bb544f8b0e83660d2b7e3f275306664f48

core-bundle/src/Crawl/Escargot/Factory.php
m-vo: Agreed. See https://github.com/contao/contao/pull/3813.

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
m-vo: I thought about the serializer as well. But IMHO this is not about serialization but "only" the normalization/denormalization part because it's always tied to a database. The serializer is probably overkill for this.

> I would consider making this one event because imho the 99% case is you need both, the serialization and unserialization part.

Yes, you would probably need both. But how would you differentiate? One time you should be able to set the metadata, one time the row data.

Toflar: Hmm. True^^ Didn't think about it properly, sorry.

core-bundle/src/DependencyInjection/Configuration.php
m-vo: Interesting, I like this idea!

I think we could implement it in a future PR without breaking anything and keep this one as big as it already is: :smile: 
* introduce a new fileld `lastModified` to `tl_files`
* add the function to the `Dbafs` class (disabled by default, also makes sense, because the service should IMO require the minimal amount of columns by default)
* enable it in the config

core-bundle/src/Filesystem/Dbafs.php
m-vo: > What happens if the algorithm changes? No hash will match anymore, does that mean, we need a migration 

Changing the algorithm and directly triggering a sync afterwards (i.e. not altering the filesystem) should rebuild the table just fine - it should be analogous to changing the contents of all files at once but still keeping UUIDs/paths the same.

It should IMHO be a user decision for this reason, though. Making it dependent of the `PHP_VERSION_ID` will probably cause confusion when people switch environments/deploy stuff. I think I would change the default for Contao 5 where we have PHP8 as a minimum requirement anyways.


> Let's make the hash algorithm its own service,

Currently you can simply change a value in the config (this is how I tested switching algorithms) to test out and set different algorithms. I mean, I'm all for abstracting stuff but PHP's `hash(string $algo, string $data)` already kind of does that, no? :slightly_smiling_face:

core-bundle/src/DependencyInjection/Configuration.php
Toflar: True, the new abstraction would allow for something like that. We should definitely not forget about that. But creating an issue for a PR that's not even merged yet sounds a bit wrong :D

core-bundle/src/Filesystem/Dbafs.php
Toflar: Yeah but the hash algo is part of every `dbafs` config so you have to make sure you update all of them. But yeah, it might be true that it's unnecessary. Just wondering if that whole concept is gonna work out with remote FS. I mean...calculating a hash for a 4GB remote file is going to take a while. That why I'm thinking about maybe somehow outsourcing that to the adapter. Because, well...is `xxh128` faster than `md5`? Hell yes. But is it also faster with S3 if in fact the `md5` hash was part of the meta data and nothing needs to be downloaded? Hell no.

m-vo: Yes, I think the algorithm choice is more something for the (fast) local filesystem which will be the default for 99% of our users. And it's probably more about the default and the ability to configure it _should_ you want to tweak it, rather than a setting everyone has to fiddle with.

Your idea from https://github.com/contao/contao/pull/3774#discussion_r767493300 will probably make the larger difference when using a DBAFS with slow filesystems. If you want to benefit from filesystems that can provide file hashes via their metadata, I've got the feeling you will end up writing your own specialized `DbafsInterface` anyways.

Toflar: I thought about this again and I think it's really a major design flaw to always execute the file content hashing locally. It **really should** happen in the adapter because things can be easily precalculated on any remote host instead of transferring thousands of megabytes only for a hash. However, I did some research on Flysystem and why this is not a first class citizen like e.g. `lastModified()`. And it seems like it's mostly because it's just not standardized. Not even in S3. But can I convince you to at least have something like `$hash = $event->getHash() ?: hash(....);` or so?

core-bundle/src/Filesystem/Dbafs.php
```diff
@@ -0,0 +1,861 @@
...
+     * specific path first).
+     *
+     * @param array<int, string> $searchPaths       non-empty list of search paths
+     * @param array<int, string> $parentDirectories parent directories to consider
+     *
+     * @return \Generator<string, (string|false|null)>
+     */
+    private function getFilesystemPaths(FilesystemAdapter $filesystem, array $searchPaths, array $parentDirectories): \Generator
+    {
+        $getHash = fn (string $path): string => hash($this->hashAlgorithm, $filesystem->read($path));
```
Toflar:
```suggestion
        $getHash = function (string $path) use ($filesystem, $path): string {
            $ctx = hash_init($this->hashAlgorithm);
            hash_update_stream($ctx, $filesystem->readStream($path));
            return hash_final($ctx);
        };
```

Not sure if it works but that would be a lot more memory efficient ;)

m-vo: Works! 

(and is already implemented in 7b97f8cf41d13bfcd8f14290b91a2f6bde97d3e5)

core-bundle/src/Filesystem/Dbafs.php
m-vo: An abstraction for the hashing is now in place with 7b97f8cf41d13bfcd8f14290b91a2f6bde97d3e5 and using "last modified" to skip generating hashes is catered for in bf7c0b07aa61983aa05d5b1fb4f6afa837467e3f (not implemented in the `Dbafs`, yet, but part of the interface).

core-bundle/src/DependencyInjection/Configuration.php
m-vo: see bf7c0b07aa61983aa05d5b1fb4f6afa837467e3f - it must be opt in, I guess (even when implemented into the `Dbafs`) because the only sensible thing then is to include the file time in the hashes as well, which would force people to recompute them.

m-vo: Ok it's implemented now. :see_no_evil:

core-bundle/src/Filesystem/UnableToResolveUuid.php
Toflar: Imho it should be consistent across our bundle, not the PHP universe :)

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -684,6 +684,10 @@ private function addVirtualFilesystemNode(): NodeDefinition
                         ->min(1)
                         ->defaultValue(100)
                     ->end()
+                    ->booleanNode('use_last_modified')
+                        ->info('When enabled, rehashing a file might be skipped if it has not changed since the last time. Note, that the DBAFS service requires a "lastModified" column in the database in this case. All hashes will initially be rewritten to include the file time.')
+                        ->defaultFalse()
```
Toflar: I would argue to switch to `defaultTrue()`. I think we can be fairly sure that the 99% case will be that `filemtime` is actually correct. We're even sort of relying on it for other purposes in Contao e.g. the image thumbnail hashes. So for those who use a filesystem that does not update the last modified date, they'll have a few other places to address and not just this config.

core-bundle/src/Filesystem/Dbafs.php
```diff
@@ -52,6 +52,7 @@ class Dbafs implements ResetInterface, DbafsInterface
     private string $dbPathPrefix = '';
     private int $maxFileSize = 2147483648; // 2 GiB
     private int $bulkInsertSize = 100;
+    private bool $useLastModified = false;
```
Toflar: Again, I'd propose to make that `true` by default.

core-bundle/src/Filesystem/HashGeneratorInterface.php
```diff
@@ -22,8 +22,12 @@ interface HashGeneratorInterface
      * If a value for $lastModified is provided, it is allowed to return null
      * instead of a hash. Doing this indicates that the hash has not changed in
      * the meantime and does not need to be recomputed.
+     *
+     * If you know the last modified timestamp (e.g. because you read it from
+     * the filesystem), assign the value to &$fileLastModified. This will prevent
+     * the calling code from reading it (again).
      */
-    public function hashFileContent(FilesystemAdapter $filesystem, string $path, int $lastModified = null): ?string;
+    public function hashFileContent(FilesystemAdapter $filesystem, string $path, int $lastModified = null, int &$fileLastModified = null): ?string;
```
Toflar: My gut feeling tells me, that this method might get another property or return value for some reason we do not know now. I'd thus propose to use a data object as argument and return value. It would allow us to easily extend the object with another property and we wouldn't have any reference handling here.

core-bundle/src/DependencyInjection/Configuration.php
m-vo: There are different places where default values are in play:

* the bundle config (this)
* how it is configured for the core (skeleton config)
* the class itself

My Idea was: with the bundle config defaulting to `false`, you do not have to create a `lastModified` column in your own DBAFS table (minimum is: `id`, `uid`, `pid`, `path`, `hash`, `type` + your own columns). But if we want this to be `true`, I'm fine with it as well. Same reasoning for the class. So: bare minimum + opt-in. These things should probably match for a better DX when creating definitions from scratch.

If we set the core's config (basically `tl_files`) to `true`, we're introducing a change to the database. I.e. the need for recomputing all hashes (migration?) and possibly breaking other peoples code that rely on it. Unlikely but still. :neutral_face:

core-bundle/src/Filesystem/HashGeneratorInterface.php
m-vo: Had this same feeling when writing the code and tried to avoid another object. :see_no_evil:  But I'm already inclined to adding one.

core-bundle/src/DependencyInjection/Configuration.php
Toflar: Why not having a general `meta` JSON column for stuff like this? Nobody queries for `last_modified` anyway, do they?

m-vo: > Nobody queries for last_modified anyway, do they?

Hmm, it's then very easy to show the last files that were changed for instance by adding an index + ordering. But it's more that we do not need a JSON field/serialized stuff anymore (I think):

If you want a new storage, say for your company videos with speaker information, you would create a storage definition, enable DBAFS and add a new table `tl_company_videos`, e.g. via a DCA definition. You would probably want to add a column `speaker`,  maybe `language` and so on to the default fields. (If you want/need  last modified, add it as well and enable it.) So every dbafs storage can have its specialized fields without interfering.

Toflar: Then I think it's totally fine to just require this field to be present in the table. The hash update shouldn't be a problem either. No migration needed imho.

m-vo: Discussed this with @ausi. Aaaand the solution is to *not* include the file time in the hash (because it kills the moving detection). I could work around this in another way. And now that there aren't any changes to the hashing needed, I'm absolutely in favor of enabling it by default. :smile:

core-bundle/src/Filesystem/HashGeneratorInterface.php
m-vo: I implemented a result object (context) in fe99dfa520c5a43463cc4089be5afdb15a1a4239.

core-bundle/src/Filesystem/Dbafs.php
m-vo: see https://github.com/contao/contao/pull/3774/commits/e508cfca00293922378cab8189c12d8e3469e141

core-bundle/src/DependencyInjection/Configuration.php
m-vo: see https://github.com/contao/contao/pull/3774/commits/f8129c5c71399250008a6bf29e8dad8c83b22d01 +  https://github.com/contao/contao/pull/3774/commits/e508cfca00293922378cab8189c12d8e3469e141

core-bundle/src/Filesystem/UnableToResolveUuid.php
m-vo: see 0b70134f9d7456b3c8862ac203ce120dadce8405

core-bundle/src/DependencyInjection/Filesystem/FilesystemConfig.php
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * The $mountPath must be a path relative to and inside the project root
+     * (e.g. 'files/foo' or 'assets/images').
+     *
+     * If you do not set a name, the id/alias for the adapter service will be
+     * derived from the mount path.
+     */
+    public function mountAdapter(string $adapter, array $options, string $mountPath, string $name = null): self
+    {
+        $name ??= (new SlugGenerator((new SlugOptions())->setDelimiter('_')))->generate($mountPath);
```
Toflar: Even though this is @ausi's library, I don't think we should have a dependency here. `Container` already has helper methods like `Container::underscore()` and `Container::camelize()`. Not sure if either one is fitting but maybe we can have a non-dependency variant using one of those.

core-bundle/src/Filesystem/Dbafs/DbafsInterface.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Filesystem\Dbafs;
+
+use Contao\CoreBundle\Filesystem\FilesystemItem;
+use Symfony\Component\Uid\Uuid;
+
+interface DbafsInterface
```
Toflar, line 21: Would love to have some better description of what `string $path` is. Is it relative? Does it have to be relative? How - as a developer - am I supposed to use it? :)

core-bundle/src/Filesystem/Dbafs/DbafsInterface.php
```diff
@@ -0,0 +1,77 @@
...
+     *   'foo/*' = foo and only direct child resources of foo
+     *
+     * @param string ...$paths relative paths inside the filesystem root
+     */
+    public function sync(string ...$paths): ChangeSet;
+
+    /**
+     * Returns true if this DBAFS sets the key 'lastModified' in the returned records.
+     */
+    public function supportsLastModified(): bool;
```
Toflar: I would generalize this as `supportsFeature(DbafsInterface::FEATURE_LAST_MODIFIED)`.

core-bundle/src/Filesystem/Dbafs/DbafsManager.php
```diff
@@ -0,0 +1,305 @@
...
+    /**
+     * Make sure that all DBAFS with a more specific prefix are also supporting
+     * everything each less specific one does. For example a DBAFS with prefix
+     * 'files/media' must also support 'fileSize' if the DBAFS under 'files'
+     * does. It could, however, support additional properties like 'mimeType',
+     * even if the 'files' DBAFS does not.
+     */
+    private function validateTransitiveProperties(): void
+    {
+        $supports = [
```
Toflar: If we had a generalized `$dbafs->supportsFeature(DbafsInterface::FEATURE_LAST_MODIFIED)` and the like, we could easily add another feature in the future. Adding a constant to the interface wouldn't break it. Adding a method would.

core-bundle/src/Filesystem/VirtualFilesystem.php
```diff
@@ -0,0 +1,278 @@
...
+     */
+    private function resolve($location): string
+    {
+        if ($location instanceof Uuid) {
+            return $this->dbafsManager->resolveUuid($location, $this->prefix);
+        }
+
+        $path = Path::canonicalize($location);
+
+        if (str_starts_with($path, '..')) {
```
Toflar, line 295: Imho that's not secure. I can do `files/../../..` to circumvent this. That's why we have `Validator::isInsecurePath()`. Maybe the Symfony `Path` library can do this for us? /cc @ausi

core-bundle/src/Resources/config/services.yml
```diff
@@ -178,6 +178,26 @@ services:
...
+
+    contao.filesystem.dbafs_manager:
+        class: Contao\CoreBundle\Filesystem\Dbafs\DbafsManager
+
+    contao.filesystem.mount_manager:
+        class: Contao\CoreBundle\Filesystem\MountManager
+        arguments:
+            - ~
+
+    contao.filesystem.virtual_factory: # really, service linter? ;-)
```
Toflar: Maybe create a separate issue with an example if you'd expect something different here. Then this can be fixed prior to this PR (or later and we rename but the comment can be removed).

aschempp (commented):
Judging by the sheer size and complexity of this PR, I feel like this should be considered experimental atm. and go into Contao 5 🤷.  It will be impossible to really review this in the next 2 weeks for 4.13 …

Sidenote: I didn't see any changes to the existing `Dbafs` class, is the new implementation fully BC on the table data etc. so we can just keep the class as-is?

bytehead:
If we consider this experimental we could still ship it with 4.13 (as the twig feature), no?

m-vo:
> Judging by the sheer size and complexity of this PR, I feel like this should be considered experimental atm

I had a call with @Toflar, @ausi and @leo where this was the conclusion IIRC, i.e. the idea to make it an experimental feature in 4.13.

> Sidenote: I didn't see any changes to the existing Dbafs class, is the new implementation fully BC on the table data etc. so we can just keep the class as-is?

Yes, it's fully BC. Rewiring the existing stuff, especially usages in the back end, would be things for follow-up PRs.

core-bundle/src/Filesystem/VirtualFilesystem.php
m-vo: The path is canonicalized at this point, so I think this is fine:

> During normalization, all slashes are replaced by forward slashes ("/").
Furthermore, all "." and ".." segments are removed as far as possible.
".." segments at the beginning of relative paths are not removed.

*(from the Path:canonicalize() doc block)*

Toflar: If we have a testcase for that, I'm fine with it :)

m-vo: Coming up… :smile:

core-bundle/src/DependencyInjection/Filesystem/FilesystemConfig.php
m-vo: adjusted in 78b38ad37d0437b1aee0ab788f629ca2d14ab715

core-bundle/src/Filesystem/Dbafs/DbafsManager.php
m-vo: Good idea! :heart:  Implemented in e252d95b7e30d1cfe7918be08e1078709ccd608e.

I went with `getSupportedFeatures()` isntead of `supportsFeature()` because this way round they can be validated without knowing the complete list.

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -176,6 +181,21 @@ static function (ChildDefinition $definition, AsPickerProvider $attribute): void
         );
     }
 
+    public function configureFilesystem(FilesystemConfig $config): void
+    {
+        $filesStorageName = 'files';
+
+        $config
+            ->mountLocalAdapter('%kernel.project_dir%/files', 'files')
```
Toflar: Don't we need `contao.upload_path` here somewhere?

m-vo: Yes, already changed locally. :smile:

m-vo: see 4dedecd301365d938bea37d068722fdd6edea19f

core-bundle/src/Filesystem/Dbafs/DbafsInterface.php
m-vo: It's all relative to the DBAFS root. So typically, `foo/bar` in a DBAFS registered to the prefix `files` will effectively target `files/foo/bar`. Prefixes are a concept of the `DbafsManager`, though, which I deliberatley not included in the `DbafsInterface`. 

Improved the descriptions in see 2c4b2fdc1212666d3f2ee7dc6dfd49ff5d038c5d

core-bundle/src/Filesystem/VirtualFilesystem.php
m-vo: Done in 27cbbbe1a449528f7cb0c9e36002a209dcf06e19

core-bundle/src/Resources/config/services.yml
m-vo: Removed in 27cbbbe1a449528f7cb0c9e36002a209dcf06e19

m-vo:
@contao/reviewers I updated the initial description and added a short paragraph about the architecture, that might make it easier to review this beast.

aschempp:
Without checking the code, the update description raises the following questions/feedback for me:
1. if everything is abstracted from Flysystem, do we need to require Flysystem at all? Or could our `MountManager`/`FilesystemAdapter` just "have an adapter that supports Flysystem?
2. The concept of `ConfigureFilesystemInterface` is something I haven't seen before. The compiler pass means the configuration is not part of the basic bundle configuration, doesn't have a config tree and would not enable a bundle to allow configuration for it. It feels like this should be tagged services instead, each handling their filesystem and possible something like a `Delegating...` class tag receives all tagged services and iterates over them.

3. > This, together with the ability to set the DBAFS table allows for a simple way to build your own DBAFS storage/table with your own attributes.

    Does that mean I can use a separate DBAFS table to add metadata to my Dropbox files? Otherwise I don't see a point in being able to rename `tl_files` etc.

m-vo:
> 1. if everything is abstracted from Flysystem, do we need to require Flysystem at all? Or could our `MountManager`/`FilesystemAdapter` just "have an adapter that supports Flysystem?

Situation is: the adapters do the heavy lifting to support the various (remote) file systems and are well tested. The other (small) bit we are using is orchestration to create adapter services in Symfony. If we need something else, the current architecture allows us doing so (e.g. by making the  `MountManager` accept it as well) without breaking anyone's code.

Abstracting the `League\…\FilesystemAdapter` interface as well doesn't make much sense to me, as noone will be interfacing with it in their code. People will be using the VFS, however, and there it's good that we do not throw `League\*` exceptions or return `League\*` filesystem items and so on.


> 2. The concept of `ConfigureFilesystemInterface` is something I haven't seen before. The compiler pass means the configuration is not part of the basic bundle configuration, doesn't have a config tree and would not enable a bundle to allow configuration for it. 

This was how it was implemented before: a (rather complex) config, that was evaluated by our extension to create and configure the respective services. If you wanted to change something in your extension you needed to prepend configuration (rather cumbersome) or alter the container in a compiler pass (rather unstable and also very cumbersome). 

Now you get a nice, easy to use, config object with shortcuts for the most common use cases (e.g. `mountAdapter()` and `mountLocalAdapter()`).

What's the downside to this approach in your opinion?


> It feels like this should be tagged services instead, each handling their filesystem and possible something like a `Delegating...` class tag receives all tagged services and iterates over them.

Tags do not work here, I think. It's often about configuring or changing something without introducing code. 

Think of mounting a folder somewhere: It would only make sense if you would create the adapter service yourself (which you typically don't). Whether `mountAdapter()`, in the current implementation, creates a adapter service and registers a method call to the `MountManager` or creates a service and tags it (to be injected in the `MountManager`) is a nuance, isn't it? 


> Does that mean I can use a separate DBAFS table to add metadata to my Dropbox files? Otherwise I don't see a point in being able to rename `tl_files` etc.

Yes, you can have multiple DBAFS each with their own table containting the "extra metadata columns" you need (plus: your own hash generator, individual settings, …).

m-vo:
Note to myself: move the events to the `Filesystem` namespace.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3776, opened 2021-12-07
Check page no search setting

tsarma:
As discussed, and suggested by @Toflar. This creates a callback on the `noSearch` field to remove indexes when it's checked. (https://github.com/contao/contao/pull/3766)
But for a unit test, I don't have any idea, how to write one!

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PageSearchSettingListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PageSearchSettingListener.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\EventListener\DataContainer;
+
+use Contao\CoreBundle\ServiceAnnotation\Callback;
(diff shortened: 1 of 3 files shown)


aschempp (changes requested):
Instead of adding a new class, I would suggest to add the additional callback to the existing class, just below https://github.com/contao/contao/blob/4.x/core-bundle/src/EventListener/DataContainer/PageUrlListener.php#L81-L93 or thelike. For the unit test, maybe you can also copy the existing one and adjust?

Regarding the code in this implementation, you should check `$value` and not `$dc->activeRecord` and I don't see why you would purge anything that is not type `regular`?

tsarma:
> Instead of adding a new class, I would suggest to add the additional callback to the existing class.

Hi @aschempp, there is no such class `PageUrlListener` in the branch `4.9`.
Regarding the code, I'll check against `$value`instead of `$dc->activeRecord` and not `regular` page check was there just in case, if someone changes some regular page, which might already be in the index, into some other type, say extrenal url.

aschempp:
Hmm, I guess we can/should only "fix" that for Contao 4.12 then. I assume purging the search if the alias changes is also not implemented in 4.9?

aschempp:
Closed in favour of https://github.com/contao/contao/pull/3839

PR #3778, opened 2021-12-08
Ensure type-safety when replacing legacy insert tags

aschempp:
I just had a case where something would pass `null` instead of a string to the legacy method.

`Argument 1 passed to Contao\CoreBundle\InsertTag\InsertTagParser::replace() must be of the type string, null given, called in vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php on line 810`

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -807,10 +807,10 @@ abstract class Controller extends System
 
 		if ($blnCache)
 		{
-			return $parser->replace($strBuffer);
+			return $parser->replace((string) $strBuffer);
 		}
 
-		return $parser->replaceInline($strBuffer);
+		return $parser->replaceInline((string) $strBuffer);
 	}
 
 	/**


fritzmg (approved):

bytehead (approved):

m-vo:
Shouldn't this be fixed in the calling method? `Controller::replaceInsertTags()` has a `@param string  $strBuffer` typehint.

ausi:
> Shouldn't this be fixed in the calling method?

IMO both.

ausi (approved):

aschempp:
> Shouldn't this be fixed in the calling method? `Controller::replaceInsertTags()` has a `@param string $strBuffer` typehint.

technically both, but it's a third-party extension I can't control. And this currently breaks BC because it previously works.

fritzmg:
May be add a deprecation, so that we can drop support for mixed values? Though we did not introduce it for https://github.com/contao/contao/issues/3001 either.

m-vo:
> And this currently breaks BC because it previously works.

It did work before by accident because noone enforced the type hint. In general I don't like these silent casts tbh. It swallows errors that should occur because the usage is wrong. 

In this case I would still be fine merging it for the sake of simplicity and because the legacy code is full of wrong type hints. :neutral_face:

ausi:
> May be add a deprecation

The whole method is already deprecated.

PR #3780, opened 2021-12-08
Allow to ignore SSL errors when updating search index

aschempp:
I had to locally update the search index of a site where my machine was unable to verify the peer certificate for. This simple option allowed to do that. Not sure if `--ignore-certificate` is the best term, maybe something about `-ssl` or `https`? Maybe someone knows a bundle that has a similar flag?

### DIFF
--- a/core-bundle/src/Command/CrawlCommand.php
+++ b/core-bundle/src/Command/CrawlCommand.php
@@ -75,6 +75,7 @@ class CrawlCommand extends Command
             ->addOption('no-progress', null, InputOption::VALUE_NONE, 'Disables the progress bar output')
             ->addOption('enable-debug-csv', null, InputOption::VALUE_NONE, 'Writes the crawl debug log into a separate CSV file')
             ->addOption('debug-csv-path', null, InputOption::VALUE_REQUIRED, 'The path of the debug log CSV file', Path::join(getcwd(), 'crawl_debug_log.csv'))
+            ->addOption('ignore-certificate', null, InputOption::VALUE_NONE, 'Ignore the TLS/SSL certificate of indexed hosts')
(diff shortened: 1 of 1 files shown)


core-bundle/src/Command/CrawlCommand.php
```diff
@@ -75,6 +75,7 @@ protected function configure(): void
             ->addOption('no-progress', null, InputOption::VALUE_NONE, 'Disables the progress bar output')
             ->addOption('enable-debug-csv', null, InputOption::VALUE_NONE, 'Writes the crawl debug log into a separate CSV file')
             ->addOption('debug-csv-path', null, InputOption::VALUE_REQUIRED, 'The path of the debug log CSV file', Path::join(getcwd(), 'crawl_debug_log.csv'))
+            ->addOption('ignore-certificate', null, InputOption::VALUE_NONE, 'Ignore the TLS/SSL certificate of indexed hosts')
```
bytehead, line 78: Naming examples from other tools:

- `curl`: `--insecure`
- Chrome:  `--ignore-certificate-errors`

But `ignore-certificate` seems valid to me as you set `verify_peer` and `verify_host` to `false` and therefore a certificate error cannot occur anymore.

bytehead (approved):

fritzmg (approved):
You could also set this via 

```yaml
# config/config.yaml
contao:
    crawl:
        default_http_client_options:
            verify_peer: false
            verify_host: false
```

but having the command option is definitely more convenient, so that you do not have to have an environment dependent configuration just for this.

Toflar (changes requested):
Against this. Never makes sense in production and the config options for local development exist ever since.

PR #3785, opened 2021-12-09
Show member groups for content elements when protected

### LINKED ISSUE #3784: Show chosen member groups on protected content elements
If a content element is protected, I'd prefer to see the member group in the brackets. At the moment you only see that the item is protected, but if there are multiple identical elements following it's not clear whom each item is protected for.  

A simple listing of used member groups would help, e.g. (protected: group1) or (protected: group1, group2, group3)

Bildschirmfoto 2021-12-09 um 11 35 14

fritzmg:
| Fixed issues     | fixes #3784      |
| Docs PR or issue | - |

This will show the chosen member groups (including "Guests") when an element is protected:

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -19,6 +19,7 @@ use Contao\CoreBundle\Security\ContaoCorePermissions;
 use Contao\DataContainer;
 use Contao\Image;
 use Contao\Input;
+use Contao\MemberGroupModel;
 use Contao\Message;
 use Contao\PageModel;
 use Contao\StringUtil;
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -1291,7 +1292,23 @@ public function addCteType($arrRow)
 		// Add the protection status
 		if ($arrRow['protected'])
 		{
-			$type .= ' (' . $GLOBALS['TL_LANG']['MSC']['protected'] . ')';
+			$groupIds = StringUtil::deserialize($arrRow['groups'], true);
+			$groupNames = array();
+
+			if (!empty($groupIds))
+			{
+				if (in_array(-1, array_map('intval', $groupIds)))
```
leofeyer: The int cast is redundant if you do not also set the third argument to `true` (which will make `in_array()` type-safe). Since this is legacy code, just omit the `array_map()` entirely.

leofeyer (changes requested):

core-bundle/src/Resources/contao/dca/tl_content.php
fritzmg: Yeah, I relied on the CS fixer to fix that 🙈. I'd rather have it type safe and change it as such.

m-vo:
Should we maybe change `protected: group1, group2` to something like `only show for: group1, group2` to make it clear it's not protected for those groups and accessible for anyone else? :slightly_smiling_face:

fritzmg:
We probably need a more general concept for easily displaying publication status at a glance for any record (not necessarily just content elements). However - looking at the discourse in https://github.com/contao/core-bundle/pull/940 it feels like an impossible task 😄

aschempp:
> We probably need a more general concept for easily displaying publication status …

Would that mean this PR is not ready to merge / should possible be removed from 4.13?

fritzmg:
Personally I am fine with merging as is. Any sort of improvement to the wording or a rework in a general concept can still be done in the future.

ausi (approved):

PR #3781, opened 2021-12-08
Fix unknown "ansi_to_html" filter error

fritzmg:
| Q                | A               |
|------------------|-----------------|
| Fixed issues     | Fixes #3772      |
| Docs PR or issue | - |

Restores the erroneously removed `ansi_to_html` twig filter.

PR #3786, opened 2021-12-09
Use lower case for "fixes" in PR template

fritzmg:
I noticed that GitHub is never automatically linking the original issues when creating a PR with the current template, even though it says `Fixes #…`. It appears that it only works when it's lower case (which I find peculiar).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 | Q                | A               |
 |------------------|-----------------|
-| Fixed issues     | Fixes #...      |
+| Fixed issues     | fixes #...      |
 | Docs PR or issue | contao/docs#... |
 
 <!--
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,4 @@
-| Q                | A               |
-|------------------|-----------------|
-| Fixed issues     | fixes #...      |
-| Docs PR or issue | contao/docs#... |
+Fixes #...
 
 <!--
 Bugfixes should be based on the 4.9 or 4.12 branch and features on the 4.x
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-Fixes #...
+Fixes #
 
 <!--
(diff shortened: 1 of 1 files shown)


bytehead (approved):

bytehead (dismissed):
May rebase to 4.9 or is it branch independently?

fritzmg:
> May rebase to 4.9 or is it branch independently?

I think GitHub pulls this information from the default branch, which is `4.x`, doesn't it @leofeyer ?

leofeyer:
Why do you want to change this? IMHO the column is a new sentence and thus the first word should be capitalized.

fritzmg:
> Why do you want to change this? IMHO the column is a new sentence and thus the first word should be capitalized.

When using the capitalized version GitHub does not automatically link the issue. You will need to manually link the issue after the PR is created. With the lower case version the issue will be linked automatically (and thus the linked issue will be closed automatically, when the PR is merged). For example:

* Here the referenced issue was not automatically linked, I had to link it manually after the fact: https://github.com/contao/contao/pull/3781
* Here the referenced issue was automatically linked, without me having to do so manually: https://github.com/contao/contao/pull/3785

leofeyer:
I see. 🤦‍♂️ 

While we are on this, do we also want to talk about simplifying the PR template? Maybe shorten it to

```
Fixes #

<!--
Bugfixes should be based on the 4.9 or 4.12 branch and features on the 4.x
branch. Select the correct branch in the "base:" drop-down menu above.

Replace this notice with a short README for your feature/bugfix. This will help
people to understand your PR and can be used as a start for the documentation.
-->
```

Because I find myself erasing the `contao/docs#...` part most of the time.

fritzmg:
Fine with me. I think we should handle documentation differently anyway. I would propose to introduce a new label like `needs documentation` that anyone can add (or I can add after the fact), if they did not already provide a PR for it - which is the case almost always anyway, because often times it's too much work to provide a PR _and_ also document the new features right away, especially when there is no guarantee yet, that the feature will be merged at all. This way I can keep track of what's merged and still not documented and then I can create issues or PRs in `contao/docs` accordingly.

leofeyer:
OK, then you adjust the PR and I will create the label.

fritzmg:
Changed now. I kept the uppercase this time because I think it works if it's outside the table at the beginning. We can always change it again anyway.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3787, opened 2021-12-09
Do not sort the referrers

### LINKED ISSUE #3779: Redirecting to article list after content element insert
**Affected version(s)**

Contao 4.9.23 

**Description**

Backend Issue:
After the Update to 4.9.23 user is being redirected to the article overview after copying and inserting a content element within an article.
One should stay within the article that the CE is inserted in.

leofeyer:
| Fixed issues     | Fixes #3779      |
| Docs PR or issue | - |

### DIFF
--- a/core-bundle/src/EventListener/StoreRefererListener.php
+++ b/core-bundle/src/EventListener/StoreRefererListener.php
@@ -89,10 +89,6 @@ class StoreRefererListener
         // Set new current referer
         $referers[$refererId]['current'] = $this->getRelativeRequestUri($request);
 
-        // Makes testing easier
-        ksort($referers[$refererId]);
-        ksort($referers);
-
         $session->set($key, $referers);
     }
 
--- a/core-bundle/tests/EventListener/StoreRefererListenerTest.php
+++ b/core-bundle/tests/EventListener/StoreRefererListenerTest.php
@@ -64,8 +64,8 @@ class StoreRefererListenerTest extends TestCase
             null,
             [
                 'newRefererId' => [
-                    'current' => 'path/of/contao?having&query&string=1',
                     'last' => '',
+                    'current' => 'path/of/contao?having&query&string=1',
                 ],
             ],
         ];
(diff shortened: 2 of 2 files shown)


Toflar:
Can anybody explain what the problem with sorting is so we can cover it in the tests. As you can see from the tests now, sorting does absolutely no harm to the existing tests?

leofeyer:
I assume that the `end()` call in the following line is causing the issue:

https://github.com/contao/contao/blob/5e3eee9408650977d3218b48c33f3c3b5b40e5dd/core-bundle/src/Resources/contao/library/Contao/System.php#L379

That seems to be the place where the sorting of the referrers matters.

Toflar:
Shouldn't it take the correct key there then? Otherwise, why do we give them keys like `current` and `last` :D 
Seems like this should be fixed in the `getReferer()` method.
Relying on a certain order within a named (!) array seems like it's gonna break again at some point.

leofeyer:
There is not only `last` and `current`, the array will also contain table name keys such as `tl_article`. I think we should not mess with the current implementation, especially since we are planning to replace it at some point anyway.

ausi (approved):

fritzmg (approved):
@Toflar the changes are correct since Contao should take the referrer that was inserted into the session _last_ when there is **no** (valid) `ref` available. However, there seems to be still a bug somewhere. All the back end actions pretty much always have the `ref` parameter, yet it often happens that `$session[$ref]` is empty. See this for example:

1. Open the site structure for example.
2. Open the _Profile_ link of the top right menu in a new tab.
3. Navigate to the _Articles_ in the **first** tab.
4. Click _Save & Close_ in the **second** tab.

In theory the second tab should take you back to the site structure - but the code when opening the URL of _2.)_ will end up in the `else` block of the condition (even though the `ref` parameter is present and should be valid) and pull the last inserted session reference - and thus the second tab ends up in the _Articles_ view.

But that should be handled in a different PR.

PR #3788, opened 2021-12-09
Clarify internal style sheets deprecation

fritzmg:
Even though we are specifically removing the internal stylesheets in Contao 5.0 - the message in the back end still says:

> The internal CSS editor has been deprecated and will be removed **in one of the next Contao versions**! Please consider exporting your existing style sheets and re-adding them to the page layout as external style sheets.

The question when the internal stylesheets are removed pops up in the community forum and slack every once in a while - and since we have a clear answer now, the message should be updated to reflect that as well.

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/default.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/default.xlf
@@ -1971,7 +1971,7 @@
         <source>Disable</source>
       </trans-unit>
       <trans-unit id="MSC.internalCssEditor">
-        <source>The internal CSS editor has been deprecated and will be removed in one of the next Contao versions! Please consider exporting your existing style sheets and re-adding them to the page layout as external style sheets.</source>
+        <source>The internal CSS editor has been deprecated and will be removed in Contao 5.0! Please consider exporting your existing style sheets and re-adding them to the page layout as external style sheets.</source>
       </trans-unit>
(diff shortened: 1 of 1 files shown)


ausi (approved):

aschempp:
Since we also know the specific date, should we also add that? Also consider the fact this message might appear in a few years on a Contao 4 installation, so maybe text something that works in all cases?

bytehead (approved):

leofeyer:
@aschempp What do you suggest?

m-vo (approved):

m-vo:
> Also consider the fact this message might appear in a few years on a Contao 4 installation

After the release of Contao 5 we could also just change all messages including "will be removed in Contao 5" to "was removed in Contao 5" in the 4.* branches.

aschempp (approved):
> After the release of Contao 5 we could also just change all messages including "will be removed in Contao 5" to "was removed in Contao 5" in the 4.* branches.

I don't think that would be worth the effort. Apart from the fact that not everything marked as `will be removed in Contao 5` WILL actually be removed in Contao 5, it means "if you upgrade from current 4 to next 5" which still is kinda "future" if Contao 5 is released.

> What do you suggest?

after re-reading and thinking what I just said above, I think the PR can stay as is 😆

leofeyer:
Thank you @fritzmg.

PR #3790, opened 2021-12-09
Add a debounce wrapper to the username autosuggester

ameotoko:
This adds a simple debounce wrapper to the members autosuggester in Preview Bar. It prevents autosuggester from sending POST request on each key stroke; instead, it waits while you type and makes a request 1 second after you've done.

### DIFF
--- a/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
+++ b/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
@@ -248,6 +248,7 @@
 
             if (document.createElement('datalist') && window.HTMLDataListElement) {
                 const userField = form.querySelector('input[name="user"]');
+                let timer;
 
                 if (userField) {
                     userField.addEventListener('input', function () {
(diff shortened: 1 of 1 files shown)


aschempp (dismissed):
Code looks good to me. Not sure 1000ms is too much, most debounce implementations I know use something around 300 (since you usually type more than 3 keys a second anyway I guess)

fritzmg (commented):
I would also lower the value to 300 or at least 500ms.

Btw. I would consider this a bugfix and thus base it against 4.9.

ameotoko:
> Btw. I would consider this a bugfix and thus base it against 4.9.

How do I rebase it properly?

aschempp:
> How do I rebase it properly?

Depends on your tool. You can (interactively) rebase in a lot of GIT tool (e.g. PHPStorm). Or if you wanna do a simple approach, create a new branch from `4.9` and cherry-pick the changes. Ideally you then rename the branch to the same as this and push the changes to the same remote, so this pull request will be updated.

ameotoko:
I would also add `LIMIT 20` to the db query, but it might be too opinionated, what do you think?

The first request (with two initial letters of the username) already returns 3000 results for me; does it make sense for browser to download, `JSON.parse` and insert 3000 elements into the DOM at this point? I find `select`s unusable anyway, when options list is higher then screen height. Or would you rather consider it an edge-case?

ameotoko:
> Not sure 1000ms is too much

Changed to 300 while rebasing, sorry😅

aschempp:
> The first request (with two initial letters of the username) already returns 3000 results for me; 

I've seen other tools only show results after the third letter. Not sure that's useful for just a few DB entries though...

ausi:
> I would also add `LIMIT 20` to the db query, but it might be too opinionated, what do you think?

I  would use something like `100` but IMO yes, this is a good idea to increase performance for projects with a huge number of members.

aschempp:
maybe this is something that should also be added in https://github.com/contao/contao/pull/3890 @bennyborn ?

bennyborn:
Sure. I could incorporate the debounce into my PR.

fritzmg (approved):

fritzmg:
What for? Let's just merge this one first.

aschempp (approved):

leofeyer:
Thank you @ameotoko.

PR #3791, opened 2021-12-10
Add annotations that generating page URLs can throw an exception

aschempp:
Added this to 4.12 since we already have the new routing there with exceptions.


Toflar (approved):

bytehead (approved):

leofeyer (changes requested):
I won't merge this until we have discussed #3724.

aschempp:
I understand, but this `@throws` exactly is relevant because we DO want people to catch it – contrary to most other exceptions as mentioned by @m-vo

leofeyer:
See #3816 for a new approach then.

leofeyer (approved):

PR #3792, opened 2021-12-10
Fix more PHP8 warnings

aschempp:
Stuff that came up during my PHP8 trials. Not sure if we should further optimize the `nav_default` template. I think in my case it is used by ChangeLanguage which does not set all array properties.


core-bundle/src/Resources/contao/templates/navigation/nav_default.html5
```diff
@@ -4,7 +4,7 @@
...
+      <li<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?>><a href="<?= $item['href'] ?: './' ?>" title="<?= $item['pageTitle'] ?: $item['title'] ?>"<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?><?php if ('' !== $item['accesskey']): ?> accesskey="<?= $item['accesskey'] ?>"<?php endif; ?><?php if ($item['tabindex']): ?> tabindex="<?= $item['tabindex'] ?>"<?php endif; ?><?= $item['target'] ?><?= $item['rel'] ?? '' ?><?php if (!empty($item['subitems'])): ?> aria-haspopup="true"<?php endif; ?> itemprop="url"><span itemprop="name"><?= $item['link'] ?></span></a><?= $item['subitems'] ?></li>
     <?php endif; ?>
   <?php endforeach; ?>
-</ul>
+</ul>
```
bytehead: CS?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -889,8 +889,13 @@ protected function generateGlobalButtons()
 			}
 
 			$v = \is_array($v) ? $v : array($v);
-			$label = \is_array($v['label']) ? $v['label'][0] : $v['label'];
-			$title = \is_array($v['label']) ? $v['label'][1] : $v['label'];
+			$title = $label = $k;
```
leofeyer, line 892: This is a behavior change. Label and title used to be `null` if `$v['label']` was not set.

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp: https://github.com/contao/contao/pull/3792/files/8346b5c7e7479fc04c4d3edcc8b601a81046c40a#diff-03666a52b2ba705242aa20dd08b23a30db45719f48607261eabdb45d2cec3b36R915-R923 😉

aschempp: https://github.com/contao/contao/blob/8346b5c7e7479fc04c4d3edcc8b601a81046c40a/core-bundle/src/Resources/contao/classes/DataContainer.php#L915-L923 😉

core-bundle/src/Resources/contao/templates/navigation/nav_default.html5
aschempp: pretty sure the .editorconfig caused that. @leofeyer is that wrong in the PR or was wrong in the template?

leofeyer (approved):

core-bundle/src/Resources/contao/templates/navigation/nav_default.html5
leofeyer: It was wrong in the PR but I have already reverted the change.

PR #3793, opened 2021-12-10
Allow creating nested folders in the file manager

leofeyer:
Implements contao/core#2907

@ausi Any security concerns?


aschempp:
what if you enter `foo/../bar` in your demo setup?

leofeyer:
It will not get past `Validator::isInsecurePath()`.

core-bundle/src/Resources/contao/dca/tl_files.php
```diff
@@ -607,8 +607,9 @@ public function addFileLocation($value, DataContainer $dc)
 	public function checkFilename($varValue, DataContainer $dc)
 	{
 		$varValue = str_replace('"', '', $varValue);
+		$varValue = trim($varValue, '/');
```
ausi: I think we should probably split the value here on `/` and validate every single part with the code below (not allowing `.` at the and and checking maxlength)

leofeyer: Changed in 2c037b1e716c692bc469ac15b48eaac617b766c9.

ausi: The maxlength check is still on the whole path, but it should be checked on every single chunk I think?

leofeyer: Changed in 37e60da26f54a1f9c88dc333005e3eef0041be95.

leofeyer:
I have just noticed another problem: The `checkFilename()` changes now also allow slashes when **renaming** files or folders. 🤔

ausi:
> I have just noticed another problem: The `checkFilename()` changes now also allow slashes when **renaming** files or folders. 🤔

This can be seen as a feature I think. I often use this in IDEs to move files into a new subfolder.

ausi (dismissed):

aschempp:
> I have just noticed another problem: The `checkFilename()` changes now also allow slashes when **renaming** files or folders. 🤔

what happens if you rename a file and include slashes? creating subdirectories seems unintiutive to me?

ausi:
> what happens if you rename a file and include slashes? creating subdirectories seems unintiutive to me?

Does it? I always use this in code editors to move files to non-yet-existing subdirectories...

leofeyer:
> This can be seen as a feature I think. I often use this in IDEs to move files into a new subfolder.

I agree, it just does not work yet. 😄

Aybee:
> what happens if you rename a file and include slashes? creating subdirectories seems unintiutive to me?

I also use this technique in filezilla.

leofeyer:
Turns out we cannot modify the folder structure when renaming existing resources.

```
files/
    foo/
        bar/
            baz/
```

When trying to rename the `files/foo/bar` folder to `bar/bar`, the following exception is thrown:



Therefore I have adjusted the code to only allow using nested folders when creating new folders (as initially requested).

Toflar (approved):

PR #3794, opened 2021-12-10
Directly jump to news archives if the user can only access one

leofeyer:
Implements contao/core#4603

If a non-admin user has access to only one news archive, a listener modifies the URL of the back end menu item "news" to point directly to this new archive instead of to the news archives overview page.

@contao/developers Is this something we want to add to news, events, FAQs and forms? And if so, do you agree with the implementation?

I am aware that unit tests and the other listeners are still missing.

### DIFF
--- a/news-bundle/src/EventListener/BackendMenuListener.php
+++ b/news-bundle/src/EventListener/BackendMenuListener.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\NewsBundle\EventListener;
+
+use Contao\CoreBundle\Event\MenuEvent;
+use Contao\NewsBundle\Security\ContaoNewsPermissions;
+use Contao\StringUtil;
+use Symfony\Component\Security\Core\Security;
+
+/**
+ * Adjusts the link to the news module, so that non-admin users who can only
+ * access one single news archive are directly taken there instead of to the
+ * news archives overview page.
+ *
+ * @internal
+ */
+class BackendMenuListener
+{
+    private Security $security;
+
+    public function __construct(Security $security)
+    {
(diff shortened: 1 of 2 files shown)


Toflar:
I would absolutely love this feature, yes! However, implementation-wise I would try to unify it maybe? At least in an abstract listener that can be reused by others? Dunno if a DCA configuration would work but I can see that other developers would want the same behaviour in their extensions.

fritzmg:
imho this should be implemented directly in the data container. The data container would then redirect to the first child (under these conditions). Could then be configurable in the DCA.

aschempp:
I like the feature a lot. I don't think it should be implemented this way though, it should rather be an `onload_callback`. It should also check on `config.closed` and not just the permissions, since that could be overridden by another extension. Maybe that's what @fritzmg meant? Have a config flag in the DCA to forward to the first item in the list, if there is only one item and the DCA is closed?

leofeyer:
> Have a config flag in the DCA to forward to the first item in the list, if there is only one item and the DCA is closed?

You have to check the user permissions, because admins and users who are allowed to create new archives need to see the archive overview page to add archives.

asaage:
> users who are allowed to create new archives need to see the archive overview page to add archives.

That was my concern too.

You could also argue if it's useful to jump from Site-structure directly into Contentelements if there is just one Article
image

If in doubt i'd say dont touch it.

leofeyer:
As discussed in the Contao call, <del>we want to try to implement this in the `DataContainer` class and base the check on wether the table is closed and not deletable (which effectively checks `ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES` and `ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES`).</del>

<del>We also should add a `config.forwardToFirst` flag to enable the feature, so people can disable it in case they want to.</del>

However, this will break global actions (e.g. "edit RSS feeds" or "show info") and probably other actions, therefore we do not want to implement this. If someone really needs the feature, they can create an extension based on the code of this PR.

PR #3795, opened 2021-12-10
Fix issues with non-admin users

leofeyer:
These are issues I have noticed while working on #3794.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3983,7 +3983,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 		}
 
 		$label = preg_replace('/\(\) ?|\[] ?|{} ?|<> ?/', '', $label);
-		$isVisibleRootTrailPage = \in_array($id, $this->visibleRootTrails);
+		$isVisibleRootTrailPage = $needsCheck ? \in_array($id, $this->visibleRootTrails) : \in_array($objRow->pid, $this->visibleRootTrails);
 
 		// Call the label_callback ($row, $label, $this)
 		if (\is_array($GLOBALS['TL_DCA'][$table]['list']['label']['label_callback'] ?? null))
@@ -6283,7 +6283,10 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 		}
 
 		// Fetch all children of the root
-		$this->rootChildren = $this->Database->getChildRecords($this->root, $table, $this->Database->fieldExists('sorting', $table));
+		if ($this->Database->fieldExists('sorting', $table))
+		{
+			$this->rootChildren = $this->Database->getChildRecords($this->root, $table, true);
+		}
 	}
 
 	/**
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3838,11 +3838,12 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 	 */
 	protected function generateTree($table, $id, $arrPrevNext, $blnHasSorting, $intMargin=0, $arrClipboard=null, $blnCircularReference=false, $protectedPage=false, $blnNoRecursion=false, $arrFound=array())
 	{
-		// Must be either visible in the root trail or allowed by permissions (or their children)
-		// In the extended tree view (mode 6) we have to check on the parent table only
-		$needsCheck = ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) == self::MODE_TREE || $table !== $this->strTable;
+		// Check if the ID is visible in the root trail or allowed by permissions (or their children)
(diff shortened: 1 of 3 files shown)


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3983,7 +3983,7 @@ protected function generateTree($table, $id, $arrPrevNext, $blnHasSorting, $intM
 		}
 
 		$label = preg_replace('/\(\) ?|\[] ?|{} ?|<> ?/', '', $label);
-		$isVisibleRootTrailPage = \in_array($id, $this->visibleRootTrails);
+		$isVisibleRootTrailPage = $needsCheck ? \in_array($id, $this->visibleRootTrails) : \in_array($objRow->pid, $this->visibleRootTrails);
```
leofeyer: Without this check, Contao will display buttons for articles on trail pages that end in a 403 error. @Toflar Not sure if that is the correct fix?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6283,7 +6283,10 @@ protected function initRoots()
 		}
 
 		// Fetch all children of the root
-		$this->rootChildren = $this->Database->getChildRecords($this->root, $table, $this->Database->fieldExists('sorting', $table));
```
leofeyer, line 6286: This caused a DB error for tables without a `pid` column:



@Toflar Since all tables with a `sorting` column also have a `pid` column in Contao, I have solved it this way. We could also check for a `pid` column in the if condition instead. WDYT?

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I don't understand the change at all. Why do you want to change it?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6283,7 +6283,10 @@ protected function initRoots()
...
+		if ($this->Database->fieldExists('sorting', $table))
```
Toflar, line 6286: Can you explain that change?

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: These two are root trail pages. They are not mounted, so when the user clicks the edit button, it will trigger a 403 exception. After the change, these pages will not have buttons anymore.

Toflar: One is `tl_page` and one is `tl_article`. The permission is configured on `tl_page`. So you need to make sure you take the parent ID there, I see.
Hmm, I see the problem but I don't think this is the right solution. I mean, especially because the naming is super confusing now. `$needsCheck` should be renamed then, wdyt?

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Same here, I see the problem but I don't think the solution is the right one :D It should be the same logic as everywhere else? Probably checking on the mode?

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: I am only reusing the `$needsCheck` variable because it is already there. 😄 We might as well duplicate the if condition or come up with something more sophisticated.

Toflar: Yeah I just want to make sure we still understand the code in the future :)

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: Yes, we could check for `$this->treeView` as in line 6241. Or we could move line 6286 into the if condition above:

```php
if ($GLOBALS['TL_DCA'][$table]['list']['sorting']['showRootTrails'] ?? null)
{
	foreach ($this->root as $id)
	{
		$this->visibleRootTrails = array_unique(array_merge($this->visibleRootTrails, $this->Database->getParentRecords($id, $table, true)));
	}

	$this->rootChildren = $this->Database->getChildRecords($this->root, $table, $this->Database->fieldExists('sorting', $table));
}
```

Or is `$this->rootChildren` independent from the `showRootTrails` setting?

Toflar: I think it's independent. It just contains the children of `root` which is a permission thing. `showRootTrails` is not related to permissions directly. Imho it's dependent on the tree mode. If we're not in the tree mode, we don't have any `root` children, right?

leofeyer: Fixed in 0a30a0026a9881aa5d16e2625ad2e846542ecfba.

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: Improved in 0a30a0026a9881aa5d16e2625ad2e846542ecfba.

Toflar (approved):

PR #2362, opened 2020-09-25
Correctly handle IDNA hostnames in the root page

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2335, #2239
| Docs PR or issue | -

PR #3800, opened 2021-12-10
Fix routing with local domain and port

### LINKED ISSUE #3783: No root page found if server runs on specific port
**Affected version(s)**

Contao 4.9

**Description**

I have a Contao installation that runs on port 8080 in my local dev environment. I used to configure the DNS with port number. 
Since https://github.com/contao/contao/pull/2362 the port number is stripped from the dns config, which breaks routing alltogether. I just experienced this now as it is the first time since a long time I saved the root page, actually running the new checkDns method. So all time until then, the DNS setting was still stored in the database including the port number. 

In short: Running your website under `example.local:8080` with DNS setting `example.local` does not work in Contao 4.9. 

I manually applied the changes made in https://github.com/contao/contao/pull/2673 to `Environment` and `UrlMatcher` and it fixes these issues, too.

bezin:
Fixes #3783


bezin:
I removed the port from the dns field in the test database, because the port is no longer stored there.

aschempp (approved):
I can only _assume_ this is correct from reading the linked issues and pull requests. No idea why we removed the port in `tl_page.dns` but only fixed the routing in 4.10 in the first place 🤷

bezin:
The bug was introduced with #2362: Fixing issues with umlauts in domains was the primary goal. The decoding of the domains also strips the port numbers from the string. I assume that this was an unintended side-effect of that PR.

leofeyer:
Thank you @bezin.

PR #3804, opened 2021-12-10
Always show debug log and fetch crawl status earlier

### LINKED ISSUE #1371: Crawlvorgang startet nach Neuladen der Seite erneut
Wenn in der Systemwartung ein Crawlvorgang durchgeführt wurde und die Ergebnisseite dann neu geladen wird, dann wird der Crawlvorgang mit derselben Jobid neu gestartet anstatt direkt die vorhandenen Ergebnisse darzustellen.

Toflar:
This allows to always download the crawl debug log file even while the crawler is running and it also fixes https://github.com/contao/contao/issues/1371 because it actually always continued on page reload, you just could only see the current status after the first crawl process which might take 10 - 20 seconds depending on the settings.
This is solved now by simply executing the first ajax request without actually initiating the crawl process.


leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3810, opened 2021-12-13
Use mode constants in Picker widget

bezin:
Uses mode constants in two more places in the Picker widget. I took the opportunity and re-scanned the code base, but I did not find any more places. Maybe these are the last :-)


aschempp (approved):

leofeyer:
Thank you @bezin.

PR #3811, opened 2021-12-13
Prevent installing the Symfony framework bundle in version 6

leofeyer:
We must not install `symfony/framework-bundle` in version 6, otherwise there will be a fatal error:

```
PHP Fatal error:  Declaration of Contao\CoreBundle\Controller\AbstractController::getSubscribedServices() must be compatible with Symfony\Bundle\FrameworkBundle\Controller\AbstractController::getSubscribedServices(): array in vendor/contao/contao/core-bundle/src/Controller/AbstractController.php on line 28
```

### DIFF
--- a/tools/ecs/composer.lock
+++ b/tools/ecs/composer.lock
@@ -172,16 +172,16 @@
         },
         {
             "name": "slevomat/coding-standard",
-            "version": "7.0.16",
+            "version": "7.0.18",
             "source": {
                 "type": "git",
                 "url": "https://github.com/slevomat/coding-standard.git",
-                "reference": "14c324b2f2f0072933036c2f3abaeda16a56dcd3"
+                "reference": "b81ac84f41a4797dc25c8ede1b0718e2a74be0fc"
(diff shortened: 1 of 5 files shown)


bytehead (approved):

bytehead:
But in the end (4.13 release), things should be compatible with 5.4 or 6 at the same time, no?

Toflar:
It's just CI, doesn't matter

leofeyer:
> But in the end (4.13 release), things should be compatible with 5.4 or 6 at the same time, no?

We would have to ship two different versions of the class (and probably other classes as well) to be compatible with both Symfony 5 and 6. I doubt that we want that.

Contao 4.13 LTS is bound to Symfony 5.4 LTS. It will not work with Symfony 6.

leofeyer:
Oh, and we cannot even ship two different versions of the class, because if we added the return type to our `AbstractController`, every child class in every extension would break. So we can only change this in Contao 5.

PR #3812, opened 2021-12-13
Increase the length of URL fields

fritzmg:
Currently all URL fields use the same maximum length as most other typical text fields, i.e. 255. However, in some cases this will not be enough. URLs can often be way longer than that and then you will either not be able to enter that URL in the back end, or if the URL is stored via an import script for example in the database directly, you will get an error (when using the strict mode).

As discussed with @ausi and @Toflar briefly in Slack, using `2048` should be a sane default for URL fields. Any URLs with a length above that might cause problems in some legacy systems (I think IE does not support URL lengths above 2048 for example, not that that's very relevant these days).

Regarding the maximum row size in the database: this should not be a problem. As it turns out, the solution to the "row size too large error" can also be to _increase_ the length of varchar fields, depending on the used collation. For `utf8mb4`, any varchar field with a length of 64 and above is already stored in "overflow pages". For `utf8` the minimum length is `86`. See https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/#fitting-more-columns-on-overflow-pages

The only problem related to the length that might still occur is, when you use said field in an index. But since these fields are not used in any index (by default) it should be no problem either.

### DIFF
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -401,8 +401,8 @@ $GLOBALS['TL_DCA']['tl_calendar_events'] = array
 			'exclude'                 => true,
 			'search'                  => true,
 			'inputType'               => 'text',
-			'eval'                    => array('rgxp'=>'url', 'decodeEntities'=>true, 'maxlength'=>255, 'dcaPicker'=>true, 'tl_class'=>'w50 wizard'),
-			'sql'                     => "varchar(255) NOT NULL default ''"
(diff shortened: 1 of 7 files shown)


leofeyer:
I guess the few people who ran into this issue were able to solve it by changing the column definition. The rest (probably 99,9%) never had this issue and never will. So why change it for everyone now?

ausi (approved):

ausi:
> So why change it for everyone now?

IMO to follow the de facto standard of URL lengths.

bytehead (approved):

m-vo (approved):

aschempp (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #703, opened 2019-09-01
Add support for lazy loading images

ausi:
See #433

I only added support for `loading="lazy"` for now because `eager` and `auto` don’t change the default browser behavior currently.

PR #3815, opened 2021-12-13
Enable image lazy loading in file manager

fritzmg:
Opening the file manager while a folder is open that contains a large amount of images can cause a lot of load on the web server since the browser will try to load all images immediately and the server will then either serve the images directly - or still has to generate the images, if not already done (if the image cache was cleared previously for instance).

Contao 4.9 supports deferred resizing and thus we can take full advantage of `loading="lazy"` here. I consider it a bug that we did not add this alongside https://github.com/contao/contao/pull/703 back then.


Toflar (approved):

aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #3816, opened 2021-12-13
Rework the @throws annotations

leofeyer:
With regard to #3791, this is a new attempt to create rules for `@throws` annotations.

As already discussed, we do not want to add `@throws` exceptions to every method that calls another method that calls another method that might throw an exception. Instead, we want to add `@throws` to indicate that an exception should be handled in a try/catch block in contrast to letting it bubble up to the general exception handler.

### Example 1

```php
/**
 * @throws \InvalidArgumentException If the file does not exist
 */
public function loadFile(string $file): SplFileInfo
{
    // ...
}
```

We add the `@throws` annotation here to indicate that the method should not be used without a try/catch block.

### Example 2

```php
public function __construct(string $func, string $action = null, string $username = null)
{
    if ('' === $func) {
        throw new \InvalidArgumentException('The function name in the Contao context must not be empty');
    }

    // ...
}
```

We do not add a `@throws` annotation here because we want the exception to bubble up to the general exception listener. The developer is not supposed to add a try/catch block here but to provide a non empty function name.

### Conclusions

Following these rules, `@throws` annotations never make sense for

- private methods (because only we use them);
- internal classes (e.g. event listeners);
- classes that nobody is supposed to work with (e.g. controllers);
- any exception that is supposed to bubble up to the general exception handler.

They primarily make sense for

- interfaces and abstract classes;
- methods that should only be used inside a try/catch block.

### Ambiguous cases

How about invalid argument exceptions? Are these meant to bubble up or is a developer supposed to wrap them in a try/catch block?

### DIFF
--- a/core-bundle/src/Controller/BackendController.php
+++ b/core-bundle/src/Controller/BackendController.php
@@ -183,8 +183,6 @@ class BackendController extends AbstractController
      * It will determine the current provider URL based on the value, which is usually
      * read dynamically via JavaScript.
      *
-     * @throws BadRequestHttpException
-     *
      * @Route("/picker", name="contao_backend_picker")
      */
     public function pickerAction(Request $request): RedirectResponse
--- a/core-bundle/src/Controller/FrontendController.php
+++ b/core-bundle/src/Controller/FrontendController.php
@@ -68,6 +68,8 @@ class FrontendController extends AbstractController
      * Symfony will un-authenticate the user automatically by calling this route.
      *
      * @Route("/_contao/logout", name="contao_frontend_logout")
+     *
+     * @return never
      */
     public function logoutAction(): void
     {
--- a/core-bundle/src/Controller/FrontendController.php
+++ b/core-bundle/src/Controller/FrontendController.php
@@ -67,8 +67,6 @@ class FrontendController extends AbstractController
     /**
      * Symfony will un-authenticate the user automatically by calling this route.
      *
-     * @throws LogoutException
-     *
      * @Route("/_contao/logout", name="contao_frontend_logout")
      */
     public function logoutAction(): void
--- a/core-bundle/src/Crawl/Escargot/Factory.php
+++ b/core-bundle/src/Crawl/Escargot/Factory.php
@@ -196,9 +196,6 @@ class Factory
         }
     }
 
-    /**
-     * @throws \InvalidArgumentException
-     */
     private function validateSubscribers(array $selectedSubscribers): array
     {
         $msg = sprintf(
--- a/core-bundle/src/Crawl/Escargot/Factory.php
+++ b/core-bundle/src/Crawl/Escargot/Factory.php
@@ -138,9 +138,6 @@ class Factory
         return $collection;
     }
 
-    /**
-     * @throws \InvalidArgumentException
-     */
     public function create(BaseUriCollection $baseUris, QueueInterface $queue, array $selectedSubscribers, array $clientOptions = []): Escargot
     {
         $escargot = Escargot::create($baseUris, $queue)->withHttpClient($this->createHttpClient($clientOptions));
@@ -152,7 +149,6 @@ class Factory
     }
 
     /**
-     * @throws \InvalidArgumentException
      * @throws InvalidJobIdException
      */
     public function createFromJobId(string $jobId, QueueInterface $queue, array $selectedSubscribers, array $clientOptions = []): Escargot
--- a/core-bundle/src/DataContainer/PaletteManipulator.php
+++ b/core-bundle/src/DataContainer/PaletteManipulator.php
@@ -35,6 +35,8 @@ class PaletteManipulator
      *
      * @param string|array $parent
      * @param bool         $hide
+     *
+     * @throws PalettePositionException
      */
     public function addLegend(string $name, $parent, string $position = self::POSITION_AFTER, $hide = false): self
     {
@@ -146,6 +148,9 @@ class PaletteManipulator
         return $this->implode($config);
     }
 
+    /**
+     * @throws PalettePositionException
+     */
     private function validatePosition(string $position): void
     {
         static $positions = [
--- a/core-bundle/src/DataContainer/PaletteManipulator.php
+++ b/core-bundle/src/DataContainer/PaletteManipulator.php
@@ -146,9 +146,6 @@ class PaletteManipulator
         return $this->implode($config);
     }
 
-    /**
-     * @throws PalettePositionException
-     */
     private function validatePosition(string $position): void
     {
         static $positions = [
--- a/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
@@ -211,9 +211,6 @@ class PageUrlListener implements ResetInterface
         }
     }
 
-    /**
-     * @throws DuplicateAliasException
-     */
     private function aliasExists(string $currentAlias, int $currentId, PageModel $currentPage, bool $throw = false): bool
     {
         $currentPage->loadDetails();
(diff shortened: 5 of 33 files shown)


ausi:
> How about invalid argument exceptions? Are these meant to bubble up or is a developer supposed to wrap them in a try/catch block?

IMO almost all of them should bubble up. There might be some exceptions (☺️) where it is not possible for the developer to validate the argument beforehand and in such cases it might make sense.

> private methods (because only we use them);

I think it is still valid if (and only if) the public method of the same class adds the same `@throws` like:

```php
/**
 * @throws SomethingTheDeveloperShouldCatchException
 */
public function foo ()
{
	$this->bar();
}

/**
 * @throws SomethingTheDeveloperShouldCatchException
 */
private function bar ()
{
	throw new SomethingTheDeveloperShouldCatchException();
}
```

m-vo:
> IMO almost all of them should bubble up. 

I agree. IMHO all exceptions of the type `SomethingTheDeveloperShouldCatchException` should be custom exceptions, i.e. not from the global namespace. So maybe we can rule these out? At least for code we own.

aschempp:
sounds good to me

> private methods (because only we use them);

I think we can still have them to let ourselves know that we should handle the exception when using the private method.

m-vo:
>> private methods (because only we use them);
> 
>I think we can still have them to let ourselves know that we should handle the exception when using the private method.

And in this case they must IMO also be annotated if not handled by the consuming method (that itself could be public API).

core-bundle/src/Security/User/UserChecker.php
```diff
@@ -50,9 +50,6 @@ public function checkPostAuth(UserInterface $user): void
     {
     }
 
-    /**
-     * @throws LockedException
-     */
```
m-vo, line 55: This should stay, I guess. The function has bascially a `never` return type for a reason.

Following the logic of my last comment, the exception would need to be added to `checkPreAuth()` as well. I do know too less of the context of this code, but I assume a consumer calling the public method will need to handle `LockedExceptions`?

m-vo: (same for some other private methods above, but I did not want to spam the changed files view)

leofeyer: No, it should be removed.

Keeping the annotation would imply "please handle this exception by wrapping the method call in a try/catch block". That is exactly what we don't want. We want the exception to bubble up, so it is converted into an error message by the exception handler.

core-bundle/src/Controller/BackendController.php
```diff
@@ -183,8 +183,6 @@ public function alertsAction(): Response
      * It will determine the current provider URL based on the value, which is usually
      * read dynamically via JavaScript.
      *
-     * @throws BadRequestHttpException
-     *
```
ausi, line 187: This should stay IMO. In the case where someone calls this method manually, they should handle the `BadRequestHttpException`.

core-bundle/src/Controller/FrontendController.php
```diff
@@ -67,8 +67,6 @@ public function shareAction(): RedirectResponse
     /**
      * Symfony will un-authenticate the user automatically by calling this route.
      *
-     * @throws LogoutException
-     *
```
ausi, line 71: This should stay too IMO. And maybe combined with `@return never`.

core-bundle/src/Crawl/Escargot/Factory.php
```diff
@@ -196,9 +196,6 @@ private function registerSubscribers(Escargot $escargot, array $selectedSubscrib
         }
     }
 
-    /**
-     * @throws \InvalidArgumentException
```
ausi, line 200: `\InvalidArgumentException` needs to be removed from the other methods in this class too IMO.

core-bundle/src/DataContainer/PaletteManipulator.php
```diff
@@ -146,9 +146,6 @@ public function applyToString(string $palette, bool $skipLegends = false): strin
         return $this->implode($config);
     }
 
-    /**
-     * @throws PalettePositionException
-     */
```
ausi: IMO this should not be removed (as I would not exempt private methods from the rule for `@throws`).

And `addLegend()` needs to get an `@throws PalettePositionException` too.

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -211,9 +211,6 @@ private function recursiveValidatePages(int $pid, PageModel $rootPage): void
         }
     }
 
-    /**
-     * @throws DuplicateAliasException
-     */
```
ausi: This is a good example IMO that shows why we should not exclude private methods from the `@throws` rule. In all other methods of this class we need to try/catch this exception (as we already do) and the `@throws` annotation would help us not to forget that.

core-bundle/src/Search/Indexer/DefaultIndexer.php
```diff
@@ -122,9 +122,6 @@ public function clear(): void
         $this->connection->executeStatement('TRUNCATE TABLE tl_search_term');
     }
 
-    /**
-     * @throws IndexerException
-     */
```
ausi: Maybe use `@return never` here?
_(Just a nitpick, can be ignored ☺️)_

ausi (changes requested):
Overall 👍 on _“we want to add @throws to indicate that an exception should be handled in a try/catch block in contrast to letting it bubble up to the general exception handler.”_

I spotted some places where I think it might not be correct yet.

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
leofeyer: Agreed. 👍

core-bundle/src/Controller/BackendController.php
ausi: Changed my mind I think. Controller and Listener don’t need `@throws`...

leofeyer: I don't think anyone will ever call the method manually, but I'm OK with keeping the annotation here. 👍

core-bundle/src/Controller/FrontendController.php
leofeyer: Added in 4b170582f71bd8906902e2b5fd5ef2205966ae50.

core-bundle/src/Crawl/Escargot/Factory.php
leofeyer: Changed in 4b170582f71bd8906902e2b5fd5ef2205966ae50.

core-bundle/src/DataContainer/PaletteManipulator.php
leofeyer: Added in 4b170582f71bd8906902e2b5fd5ef2205966ae50.

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
leofeyer: Added in 4b170582f71bd8906902e2b5fd5ef2205966ae50 again.

core-bundle/src/Search/Indexer/DefaultIndexer.php
leofeyer: Added in 4b170582f71bd8906902e2b5fd5ef2205966ae50.

ausi (approved):

m-vo (approved):
> Changed my mind I think. Controller and Listener don’t need @throws...

I'd also add everything related to the build process to this. Plus everywhere an interface is implemented, because there the *interface* should define what can be thrown and the implementation must adhere to this.

Not sure if we can put these "rules" into an automatic fixer?

PR #3819, opened 2021-12-14
Add comments to our interfaces and abstract classes

leofeyer:
This PR adds comments to our interfaces and abstract classes so people know how these will change in Contao 5. It is pretty much the same thing that Symfony did.

### DIFF
--- a/core-bundle/src/Command/AbstractLockedCommand.php
+++ b/core-bundle/src/Command/AbstractLockedCommand.php
@@ -70,7 +70,7 @@ abstract class AbstractLockedCommand extends Command implements ContainerAwareIn
     /**
      * @return int
      */
-    abstract protected function executeLocked(InputInterface $input, OutputInterface $output)/*: int*/;
+    abstract protected function executeLocked(InputInterface $input, OutputInterface $output);
 
     /**
      * Creates an installation specific folder in the temporary directory and returns its path.
--- a/core-bundle/src/Command/AbstractLockedCommand.php
+++ b/core-bundle/src/Command/AbstractLockedCommand.php
@@ -70,7 +70,7 @@ abstract class AbstractLockedCommand extends Command implements ContainerAwareIn
     /**
      * @return int
      */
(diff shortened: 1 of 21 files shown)


core-bundle/src/Command/AbstractLockedCommand.php
```diff
@@ -70,7 +70,7 @@ final protected function execute(InputInterface $input, OutputInterface $output)
     /**
      * @return int
      */
-    abstract protected function executeLocked(InputInterface $input, OutputInterface $output);
+    abstract protected function executeLocked(InputInterface $input, OutputInterface $output)/*: int*/;
```
fritzmg: This class will be removed in Contao 5.0 anyway btw.

leofeyer: Reverted in 8a11b062f49367e50172992bca97f89b629824ad.

installation-bundle/src/Database/AbstractVersionUpdate.php
```diff
@@ -37,15 +37,15 @@ public function __construct(Connection $connection)
     /**
      * @return bool
      */
-    public function hasMessage()
+    public function hasMessage()/*: bool*/
```
fritzmg: This class is also deprecated 🙃

aschempp:
what's the benefit of making this effort? I mean no developer will ever see this except by looking at the code, and the docblock already gives the same information?

fritzmg:
The docblock does not specify the enforced type though. Within Contao at least we allow parameter and return types to be different from the docblock.

leofeyer:
> what's the benefit of making this effort?

Why did Symfony do it? And why did we do it in the `PictureFactoryInterface`?

https://github.com/contao/contao/blob/bfff2382777eb19e9a16697f3b8db9b4d37b60e1/core-bundle/src/Image/PictureFactoryInterface.php#L40

ausi (approved):

aschempp:
should we also set `#[\ReturnTypeWillChange]` ?

PR #3822, opened 2021-12-14
Add a Template property to the Controller class

leofeyer:
See https://github.com/contao/contao/pull/3725/files#r768774313

This PR ensures that all controller-like classes have a `$Template` property, so the template object is not added to `$this->arrData` through a magic setter. It also provides forward compatibility with PHP 8.2.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/BackendModule.php
+++ b/core-bundle/src/Resources/contao/classes/BackendModule.php
@@ -32,6 +32,11 @@ abstract class BackendModule extends Backend
 	protected $objDc;
 
 	/**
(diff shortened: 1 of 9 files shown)


aschempp:
Shouldn't we add this to `\Contao\Controller` instead?

leofeyer:
> Shouldn't we add this to `\Contao\Controller` instead?

Yes. Changed in 57db9cb31b752cb195105d053d2adac528786a62.

Toflar (approved):

bytehead (approved):

PR #3835, opened 2021-12-16
Remove the alias field from unroutable pages

aschempp:
Since error pages are no longer routable (https://github.com/contao/contao/pull/3605), it does not make sense to have an alias and check it when comparing with other pages.


core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -203,7 +215,7 @@ private function recursiveValidatePages(int $pid, PageModel $rootPage): void
 
         /** @var PageModel $page */
         foreach ($pages as $page) {
-            if ($page->alias) {
+            if ($page->alias && $this->pageRegistry->isRoutable($page)) {
```
leofeyer, line 183: This condition is also checked at the beginning of the `aliasExists()` method. Are both checks necessary?

aschempp: yes, because this checks `$page` but we forward `$rootPage` to the `checkAlias` method

aschempp: But I have improved that in ed539ca88ea7797492e8077dca7f0f6666a05e6c since we don't actually need to check `$rootPage`

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -287,6 +290,29 @@ private function aliasExists(string $currentAlias, int $currentId, PageModel $cu
...
+            }
+
+            if (null !== ($urlPrefix = $input->post('urlPrefix'))) {
+                $pageModel->urlPrefix = $urlPrefix;
+            }
+
+            if (null !== ($urlSuffix = $input->post('urlSuffix'))) {
+                $pageModel->urlSuffix = $urlSuffix;
+            }
+        }
```
fritzmg, line 275: Should we add `$pageModel->preventSaving()` here just to be safe?

aschempp: no, this would create another copy of the page even though they already are through `loadDetails`

aschempp: I don't think so. Its a private method, we're the only one calling it and just need to make sure. But I actually fixed/improved that in b70f907495de097e6b124bb294752e01deaa1bcb

fritzmg: `preventSaving` is public, but with `findWithDetails` `preventSaving` is already called now, yes 👍

fritzmg (dismissed):

leofeyer (dismissed):

leofeyer (approved):

PR #3836, opened 2021-12-16
Report HTTP exceptions as warnings for the crawler instead of errors

Toflar:
I propose to change failing HTTP requests in our `SearchIndexSubscriber` to be reported as warnings instead of errors. 
Errors cause a hard failure which e.g. cause `contao:crawl` to return a non `0` exit code. This indicates that something went wrong. However, everything was just fine, there are just URLs that end up in e.g. a 401 or whatever else which can be happily ignored by the search index subscriber.

What cannot be ignored is if there's a real failure during indexing a document (e.g. a DB issue when writing to our search tables) which is already the case now.


leofeyer (approved):
I agree. Let's see what the other @contao/developers think.

fritzmg (approved):

bytehead (approved):

leofeyer:
Thank you @Toflar.

PR #3838, opened 2021-12-17
Use KnpMenu for navigation modules

### LINKED ISSUE #3162: make navigation extendable
Hi Leo

One thing that I really find annoying in Contao is that the navigation module is limited to the tl_page table and there's no way to hook upon.

Some scenarios:
- The catalog provides the possibility to generate a navigation out of a catalog but I can't tell the catalog to add its generated navigation *_into_ Home->Products so it in fact generates a submenu of "Products"
- Isotope could have the same use case as the catalog
- I wrote a customer specific module where the editors can describe sports. Every sport has to have its own navigation point and the two subpages "Description" and "Rules". The whole thing should be generated automatically. No problem at all except for adding the sports navigation to Home->Association->Sports.
- I could imagine that this could be very interesting when writing a front end admin panel for any software. Let's say you want to integrate any API from a third party application and provide the functionalities of this application within the default Contao navigation module. Same problem here...

I don't think that a hook in the current navigation module is the way to go because that would end up in too many useless hook calls (and you'd need to run through every registered hook for every page).
I rather think of a new library called "Navigation" that controls a navigation array (that can be extended - that's the point) and all the ModuleNavigation has to do would be

```
$this->Template->items = Navigation::generate($trail[$level]);
```

The fact that there exist several extensions (found "xNavigation" and "backboneit_navigation", maybe there are others) that try to solve this issue (but with hooks...) should also be evidence of the need to extend the navigation of Contao.

What do you think? ;)

--- _Originally created on June 14th, 2011, at 10:06am (ID 3162)_

richardhj:
## About

This PR introduces KnpMenu for the frontend navigation modules.

The benefits of KnpMenu will be ~better Twig implementation~ (postponed) and [using events to allow a menu to be extended](https://symfony.com/bundles/KnpMenuBundle/current/events.html).

Fixes contao/core#3162.

## Changes

We introduce the [`FrontendMenuBuilder`](https://github.com/contao/contao/pull/3838/files#diff-fbaac5f5ae13f5efd13bd4be3057a998ae9fdd7f762f39b8839911bd5eb11f8f) that contains all the logic that was used to be found in the `Module` class.

In the `Module` class, however, we [restore compatibility](https://github.com/contao/contao/pull/3838/files#diff-072e13a8027ac3afc897fcf9634710245dd6b6a962ce2c759e3b48cc5d6b439cR422-R435) of the new KnpMenu instances with the `nav_default` templates.

### Extend the menu using an event

See https://symfony.com/bundles/KnpMenuBundle/current/events.html.

```php
// src/EventListener/BackendMenuListener.php
namespace App\EventListener;

use App\Controller\BackendController;
use Contao\CoreBundle\Event\FrontendMenuEvent;
use Terminal42\ServiceAnnotationBundle\Annotation\ServiceTag;

/**
 * @ServiceTag("kernel.event_listener", event="contao.frontend_menu_build")
 */
class FrontendMenuListener
{
    public function __invoke(FrontendMenuEvent $event): void
    {
        $factory = $event->getFactory();
        $tree = $event->getTree();

        if ($event->getLevel() > 1) {
            return;
        }

        $node = $factory
            ->createItem('foobar')
        ;

        $tree->addChild($node);
    }
}
```


core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -185,6 +185,12 @@
 			'eval'                    => array('helpwizard'=>true, 'chosen'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50'),
 			'sql'                     => "varchar(64) NOT NULL default 'navigation'"
 		),
+		'menuAlias' => array(
+			'exclude'   => true,
+			'inputType' => 'text',
+			'eval'      => array('maxlength' => 255, 'tl_class' => 'w50 clr', 'unique' => true),
```
richardhj: Thinking about the `menuAlias`, I would suggest making the alias non-unique and filtering for the current theme in the menu provider, i.e., `$moduleAdapter->findBy(['menuAlias' => $name, 'pid' => $themeId])`.

This allows easy switching of themes in the front-end. But at the same time, this would disallow using navigation modules not defined in the current theme...

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -275,126 +274,47 @@ protected function getResponseCacheTags(): array
 	 *
 	 * @return string
 	 */
-	protected function renderNavigation($pid, $level=1, $host=null, $language=null)
+	protected function renderNavigation($pid, $level=1, $host=null/*, $language=null*/)
```
richardhj: There is no usage of `$language` in the current version of Module.php anyways.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,314 @@
...
+
+                    break;
+            }
+
+            $this->populateMenuItem($item, $requestPage, $page, $href, $options);
+            $root->addChild($item);
+        }
+
+        // TODO this is the same event as for the backend. must introduce new event
+        $menuEvent = new MenuEvent($this->factory, $root);
```
richardhj: Speaking from my tests, you cannot have the same event class dispatched at different spots because this means that the "[simpler event dispatching](https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching)" does not work then. 
Is this correct?

fritzmg: Within the event listener, you will check for `$tree->getName()`. The back end main menu is called `mainMenu` for example. Currently you call the front end menu `root`, but I guess it should be something like `frontEndMenu`.

However, I think within the event you should have access to things like the `ModuleModel` instance for example. So may be we need a new event after all. I think it would be better to have a `BackEndMenuEvent` and a `FrontEndMenuEvent` or something.

core-bundle/src/Resources/contao/dca/tl_module.php
m-vo: What about finding any with the given alias and preferring the one from the current theme if there are multiples and one matches the `$themeId`?

On that note: should we autogenerate aliases when saving like everywhere else?

core-bundle/src/Menu/FrontendMenuBuilder.php
m-vo: > I think it would be better to have a BackEndMenuEvent and a FrontEndMenuEvent or something.

:+1: for separating these two concerns.

core-bundle/src/Resources/contao/dca/tl_module.php
richardhj: 👎 for autogenerating the alias. Because if I name the module "Navigation - Main navigation" I don't want the alias to be sth like "navigation___main_navigation". Also, the alias really is not required, if you don't want to use the Knp Menu Provider / Twig integration.

m-vo: > " I don't want the alias to be sth like "navigation___main_navigation"

The alias could just be generated from the module type + id: `custom_navigation_123`.

richardhj:
I updated the ModuleCustomnav to use KnpMenu as well in cad7b0f198831a561e443808a9d7dbc0aa1da58a.

core-bundle/src/Resources/contao/dca/tl_module.php
richardhj: This format, however, looks good. But still, whenever an alias is generated in Contao, it is, because it's necessary. In this case, however, the feature is totally optional.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,333 @@
...
+    public function getMenu(int $pid, int $level = 1, string $host = null, array $options = []): ItemInterface
+    {
+        $root = $this->factory->createItem('root');
+
+        if (null === ($pages = $this->getPages($pid, $options))) {
+            return $root;
+        }
+
+        $request = $this->requestStack->getCurrentRequest();
+        $requestPage = $request->attributes->get('pageModel');
```
richardhj: I wonder if `$requestPage` should be part of the method signature…

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Event added in f6e69a855c1f9ba91f09a063d148d242db55ad79.

core-bundle/src/Resources/contao/dca/tl_module.php
richardhj: Alias generation added in 305158e77b525a7fb4fbf8372e1d2c5b222f726e.

richardhj: Ambiguous menu alias and automatic theme switching added in e0ef24cc5ce79c80d004c7a01c7525df9f557160.

core-bundle/src/Menu/FrontendMenuBuilder.php
Toflar: I don't think that's needed. It can be achieved by push/pop on the request stack if really needed. Otherwise we could also argue why the logged-in status is not an argument :)

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,342 @@
...
+                'page' => PageModel::findByPk($row['id']),
+                'hasSubpages' => (bool) $row['hasSubpages'],
+            ],
+            $pages
+        );
+    }
+
+    private function generateUri(PageModel $pageModel, ItemInterface $menuItem): ?string
+    {
+        if ('redirect' === $pageModel->type) {
```
Toflar, line 206: I find this code in multiple places. Also in my own apps. Any reason the router doesn't do this directly? /cc @aschempp

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,342 @@
...
+
+    private function populateMenuItem(MenuItem $item, PageModel $requestPage, PageModel $page, ?string $href, array $options = []): void
+    {
+        $extra = $page->row();
+        $trail = \in_array($page->id, $requestPage->trail, false);
+
+        $item->setUri($href);
+
+        // Use the path without query string to check for active pages (see #480)
+        $path = current(explode('?', Environment::get('request'), 2));
```
Toflar: No more legacy access please :)

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,342 @@
...
+            $cssClass = implode('', [
+                $hasSubmenu ? 'submenu' : '',
+                $page->protected ? ' protected' : '',
+                $trail ? ' trail' : '',
+                $page->cssClass ? ' '.$page->cssClass : '',
+            ]);
+
+            // Mark pages on the same level (see #2419)
+            if ($page->pid === $requestPage->pid) {
+                $cssClass .= ' sibling';
```
Toflar: Looks like an array would be easier?

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -185,6 +185,12 @@
 			'eval'                    => array('helpwizard'=>true, 'chosen'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50'),
 			'sql'                     => "varchar(64) NOT NULL default 'navigation'"
 		),
+		'menuAlias' => array(
```
Toflar: I don't see why this would be a specific `menuAlias`? Either having an alias for other purposes like insert tags would be useful too or it doesn't make sense here either. In other words: If we want aliases then imho we can introduce that as separate PR and imho they should be unique too (maybe per theme). Then it could also be optimized in the model registry :)

core-bundle/src/Menu/NavigationModuleProvider.php
```diff
@@ -0,0 +1,96 @@
...
+    public function get(string $name, array $options = []): ItemInterface
+    {
+        $request = $this->requestStack->getCurrentRequest();
+        $moduleAdapter = $this->framework->getAdapter(ModuleModel::class);
+        $layoutAdapter = $this->framework->getAdapter(LayoutModel::class);
+
+        $t = ModuleModel::getTable();
+        $currentPage = null !== $request ? $request->attributes->get('pageModel') : null;
+        $layout = null !== $currentPage ? $layoutAdapter->findByPk($currentPage->layout) : null;
+        $module = $moduleAdapter->findBy('menuAlias', $name, [
```
Toflar: If aliases were a general concept, we could make them unique. See other comments.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -420,9 +340,13 @@ protected function renderNavigation($pid, $level=1, $host=null, $language=null)
 	 * @param string    $href
 	 *
 	 * @return array
+	 *
+	 * @deprecated Will be removed in Contao 5.0
```
Toflar: Use what then? ;)

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -490,6 +414,26 @@ protected function compileNavigationRow(PageModel $objPage, PageModel $objSubpag
 		return $row;
 	}
 
+	/**
+	 * Prepare menu item for usage in nav_* templates.
+	 *
+	 * @return array
+	 */
+	protected function compileMenuItem(ItemInterface $menuItem, string $subitems = ''): array
```
Toflar, line 444: `private`? I can use the event to midify now, no?

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -498,9 +442,13 @@ protected function compileNavigationRow(PageModel $objPage, PageModel $objSubpag
 	 * @param boolean $blnIsSitemap  If true, the sitemap settings apply
 	 *
 	 * @return array<array{page:PageModel,hasSubpages:bool}>|null
+	 *
+	 * @deprecated Will be removed in Contao 5.0
```
Toflar: Explain what to use now if you were using this.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,342 @@
...
+
+        $menuEvent = new FrontendMenuEvent($this->factory, $root, $pid, $level, $options);
+        $this->dispatcher->dispatch($menuEvent, ContaoCoreEvents::FRONTEND_MENU_BUILD);
+
+        return $root;
+    }
+
+    private function getPages(int $pid, array $options): ?array
+    {
+        if (0 === $pid && $options['pages']) {
```
Toflar, line 149:
```suggestion
        // Custom page choice like e.g. for the custom navigation module
        if (0 === $pid && $options['pages']) {
```

Toflar (commented):
Looks great, thank you! The only debatable thing here is the module alias imho and I don't see how it is required for this PR to work. I would thus extract it as it would otherwise hold up this PR completely for no reason.
If that was removed and tests are submitted before Dec 31st, I think we could even get it in for 4.13 as it's otherwise an easy one to approve for me :)

core-bundle/src/Resources/contao/dca/tl_module.php
richardhj: If we remove the alias for now, we would also need to remove the Menu Provider, i.e. Twig integration. It would be a pity to not be able to fetch navigation modules in Twig templates, but I understand your concern for a more general alias concept in modules.

core-bundle/src/Resources/contao/modules/Module.php
richardhj: The ModuleCustomnav is using this method as we do not use `renderNavigation()` for custom navigations.

core-bundle/src/Resources/contao/dca/tl_module.php
Toflar: It wasn't possible until now either. This PR consists of 2 features

- extending the navigation using events
- extended Twig support

Both have added values in their own right, sure. But in order to ease the review process we should keep the chunks as small as possible.

core-bundle/src/Menu/FrontendMenuBuilder.php
aschempp: Unfortunately no, the router can only generate a URL based on the route, and that can never be the actual target URL outside the page structure. It would generate a link to that page (alias) which then forwards, which is certainly not what we want. I tried to fix this in https://github.com/contao/contao/pull/2031/files but failed so far, I noticed recently that cmf-routing actually has something similar but haven't come up with an implementation yet.

aschempp:
I agree with @Toflar on the menu alias stuff. I understand why you like that, but it goes against the general concept of Contao and its front end modules in my opinion. I also fail to see why that would be necessary for navigation modules but not e.g. for breadcrumb or anything else related to layouts.

core-bundle/src/Resources/contao/dca/tl_module.php
richardhj: Alias thingy removed in 885a3eb and also see #3875

core-bundle/src/Event/ContaoCoreEvents.php
```diff
@@ -21,6 +21,13 @@ final class ContaoCoreEvents
      */
     public const BACKEND_MENU_BUILD = 'contao.backend_menu_build';
 
+    /**
+     * The contao.frontend_menu_build event is triggered when a frontend navigation module menu is built.
+     *
+     * @see FrontendMenuEvent
+     */
+    public const FRONTEND_MENU_BUILD = 'contao.frontend_menu_build';
```
fritzmg: Best practise is to not use string constants anymore for events. We did not introduce a string constant for the `FilterPageTypeEvent` either for example.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,359 @@
...
+            if (null === $href = $this->generateUri($page, $item)) {
+                continue;
+            }
+
+            $this->populateMenuItem($item, $request, $page, $href, $options);
+            $root->addChild($item);
+        }
+
+        $menuEvent = new FrontendMenuEvent($this->factory, $root, $pid, $level, $options);
+        $this->dispatcher->dispatch($menuEvent, ContaoCoreEvents::FRONTEND_MENU_BUILD);
```
fritzmg:
```suggestion
        $this->dispatcher->dispatch($menuEvent);
```

fritzmg (changes requested):

core-bundle/src/Event/ContaoCoreEvents.php
richardhj: yes, I was wondering. Changed in 427dc02cce71dad9735afb01b927b5542e82f95f.

core-bundle/src/Resources/config/services.yml
```diff
@@ -379,6 +379,21 @@ services:
...
 
+    contao.menu.frontend_builder:
+        class: Contao\CoreBundle\Menu\FrontendMenuBuilder
+        arguments:
+            - '@Knp\Menu\FactoryInterface'
+            - '@request_stack'
+            - '@event_dispatcher'
+            - '@database_connection'
+            - '@contao.routing.page_registry'
+            - '@=service("contao.framework").getAdapter("Contao\\PageModel")'
```
Toflar, line 382: We never did this so far but we sure can :) Should we? /cc @contao/developers

leofeyer: Definitely yes. 👍

aschempp: Not sure this is bad for our tests?

m-vo:
> I agree with @Toflar on the menu alias stuff. I understand why you like that, but it goes against the general concept of Contao and its front end modules in my opinion.

But being able to use the Twig function is extremely helpful when building custom layouts. I typically create my own fe_page and directly build the needed menus in there (or in an included sub file). Especially for complex headers with integrated menus that is so much easier.

core-bundle/src/Resources/config/services.yml
leofeyer: @richardhj Does this work reliably without calling `$this->framework->initialize()` in advance? Because so far we have been initializing the framework before every call to `$this->framework->getAdapter()`.

richardhj:
> But being able to use the Twig function is extremely helpful when building custom layouts. I typically create my own fe_page and directly build the needed menus in there (or in an included sub file). Especially for complex headers with integrated menus that is so much easier.

Same for me. I will start two new PRs (one about the aliases, see #3875, one about the Menu Provider) and maybe we can squeeze them in 4.13 yet.

core-bundle/src/Event/FrontendMenuEvent.php
```diff
@@ -0,0 +1,70 @@
...
+
+class FrontendMenuEvent extends Event
+{
+    private FactoryInterface $factory;
+    private ItemInterface $tree;
+    private int $pid;
+    private int $level;
+    private array $options;
+
+    public function __construct(FactoryInterface $factory, ItemInterface $tree, int $pid, int $level, array $options)
```
fritzmg: I am not sure about the `$options` array. For example - if I want to change the menu _only_ for the `navigation` module, but not for the `customnav` module, I could check for `'navigation' === $event->getOptions()['type']` - however, since the options can contain data from basically anything, I can never be sure that this was actually a Contao module of type `navigation`.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,366 @@
...
+                ) {
+                    $item->setDisplayChildren(false);
+                }
+            }
+
+            if (null === $href = $this->generateUri($page, $item)) {
+                continue;
+            }
+
+            $this->populateMenuItem($item, $request, $page, $href, $options);
```
fritzmg: This should be moved up so that it is excuted before the children are generated. Otherwise `$event->getTree()->getExtras()` will be empty when the `FrontendMenuEvent` is dispatched (since it will be executed depth first) and thus you cannot access the information of the page for the current root node.

Example: depending on a page's setting I want to add new children to the page. Ideally I would implement the following:

```php
namespace App\EventListener;

use Contao\CoreBundle\Event\FrontendMenuEvent;
use Terminal42\ServiceAnnotationBundle\Annotation\ServiceTag;

/**
 * @ServiceTag("kernel.event_listener")
 */
class FrontendMenuListener
{
    public function __invoke(FrontendMenuEvent $event): void
    {
        $tree = $event->getTree();

        // From tl_page.foobar
        if (!$tree->getExtras()['foobar']) {
            return;
        }

        $factory = $event->getFactory();
        $node = $factory->createItem('foobar');
        $tree->addChild($node);
    }
}
```

But in the current state of this PR this would not be possible. Instead I would have to traverse the children each time, which is inefficient.

fritzmg (changes requested):

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Check out 239a7c024239f274b8e61bea01cd053a57973b58

fritzmg (commented):
I also notice that the `FrontendMenuEvent` is not fired for any pages that do not have any children. I again would have to traverse the children of every root in the event listener in order to dynamically add children for items that do not have any children yet.

If we keep the current implementation of firing the event for basically every item - then I think it should be fired for every page, not just for pages that do not have any children in the site structure.

richardhj:
> I also notice that the `FrontendMenuEvent` is not fired for any pages that do not have any children. I again would have to traverse the children of every root in the event listener in order to dynamically add children for items that do not have any children yet.
> 
> If we keep the current implementation of firing the event for basically every item - then I think it should be fired for every page, not just for pages that do not have any children in the site structure.

To fix this, we would need to dispatch the event here as well: https://github.com/contao/contao/pull/3838/files#diff-fbaac5f5ae13f5efd13bd4be3057a998ae9fdd7f762f39b8839911bd5eb11f8fR80
Agree?

fritzmg:
> To fix this, we would need to dispatch the event here as well: https://github.com/contao/contao/pull/3838/files#diff-fbaac5f5ae13f5efd13bd4be3057a998ae9fdd7f762f39b8839911bd5eb11f8fR80
> Agree?

You could return an empty array [here](https://github.com/contao/contao/pull/3838/files#diff-fbaac5f5ae13f5efd13bd4be3057a998ae9fdd7f762f39b8839911bd5eb11f8fR171) instead and remove your referenced early out.

richardhj:
Event dispatching fixed in e91af7634ed9be8ff646f420a7d5b0cd8b2abd8b.

core-bundle/src/Resources/config/services.yml
richardhj: Works like a charm.

The only issue I recognized is that [`$this->mockAdapter(...$methods)`](https://github.com/contao/test-case/blob/ceea79483c6040ab5d90c2b9cefd6a2ec5edf696/src/ContaoTestCase.php#L132) creates an Adapter instance without actual class. [`$this->class` is null](https://github.com/contao/contao/blob/6a815a472a46c44c4dbf3885f44c229a26fb8510/core-bundle/src/Framework/Adapter.php#L46) so that I decided to mock every method I use from that Adapter.

Another fix would be to update [`$this->mockAdapter(...$methods)`](https://github.com/contao/test-case/blob/ceea79483c6040ab5d90c2b9cefd6a2ec5edf696/src/ContaoTestCase.php#L132) to allow setting a class.

core-bundle/src/Event/FrontendMenuEvent.php
richardhj: We can make some options mandatory, but I don't think any need for action in that concern.
What I currently don't like on the $options array is that we have an option called 'isSitemap' that is not something coming from the module row.

fritzmg:
> Event dispatching fixed in [e91af76](https://github.com/contao/contao/commit/e91af7634ed9be8ff646f420a7d5b0cd8b2abd8b).

Hm, it still does not fire for every page. I'll debug further.

fritzmg:
The

```php
            if (!$hasSubpages) {
                continue;
            }
```
check would need to be removed as well, if we want to fire the event for every page in the navigation.

fritzmg:
Even when removing this check I am still unable to add new children via the event unfortunately (at least to leaves without any chilrden).

richardhj:
> Hm, it still does not fire for every page. I'll debug further.

We don't want to trigger the event for every menu item, but for every root node per level.
So if we have a menu with two submenus, we want the event to be dispatched three times (1x root, 2x submenu root nodes).

Please check out 427380afe6f0aada38327714b8826a1f5e9159a9. I wrote a test for this issue.

fritzmg:
Yes, but every page is a potential root node, so in the end it would be triggered for every page, would it not? 

Another issue I found is the implementation in `Module::renderNavigation`. `FrontendMenuBuilder::getMenu` already recursively builds the whole tree - but then in `Module::renderNavigation` it will be built for every level again.

richardhj:
> Yes, but every page is a potential root node, so in the end it would be triggered for every page, would it not?

You're right 😆 

> Another issue I found is the implementation in `Module::renderNavigation`. `FrontendMenuBuilder::getMenu` already recursively builds the whole tree - but then in `Module::renderNavigation` it will be built for every level again.

Good point. We either have to adjust renderNavigation() to traverse through the existing tree, or configure the MenuBuilder with a maxDepth=1 or so.

fritzmg:
Yeah I would go for something like this:

<details>
  <summary>Click to expand!</summary>
  
```php
protected function renderNavigation($pid, $level=1, $host=null/*, $language=null*/)
{
    /** @var FrontendMenuBuilder $menuBuilder */
    $menuBuilder = System::getContainer()->get('contao.menu.frontend_builder');
    $root = System::getContainer()->get('knp_menu.factory')->createItem('root');

    $options = $this->arrData;
    $options += array('isSitemap' => $this instanceof ModuleSitemap);

    $menu = $menuBuilder->getMenu($root, $pid, $level, $host, $options);

    if (!$menu->count())
    {
        return '';
    }

    return $this->renderNavigationItems($menu);
}

protected function renderNavigationItems(ItemInterface $rootItem): string
{
    $objTemplate = new FrontendTemplate($this->navigationTpl ?: 'nav_default');
    $objTemplate->pid = $rootItem->getExtras()['pid'] ?? null;
    $objTemplate->type = static::class;
    $objTemplate->cssID = $this->cssID; // see #4897
    $objTemplate->level = 'level_' . ($rootItem->getLevel() + 1);
    $objTemplate->module = $this; // see #155

    $objTemplate->items = $this->compileMenuItems($rootItem);

    return !empty($objTemplate->items) ? $objTemplate->parse() : '';
}

protected function compileMenuItems(ItemInterface $rootItem): array
{
    $items = array();

    foreach ($rootItem as $menuItem)
    {
        if ($menuItem->hasChildren() && $menuItem->getDisplayChildren()) {
            $subitems = $this->renderNavigationItems($menuItem);
        }

        $items[] = $this->compileMenuItem($menuItem, $subitems ?? '');
    }

    // Add first and last classes
    if (!empty($items))
    {
        $last = \count($items) - 1;

        $items[0]['class'] = trim($items[0]['class'] . ' first');
        $items[$last]['class'] = trim($items[$last]['class'] . ' last');
    }

    return $items;
}
```
</details>

This works as far as I tested and now my added menu items also appear. Still need to test in more detail though. 

This also removes the necessity of having to have a `pageModel` "extra" for each menu item, which is not very intuitive if you want to add arbitrary menu items to Contao's main navigation.

fritzmg:
I can provide a PR for your branch with this plus a few other suggestions, if you'd like. For example, you don't really need to pass the `$level` to `FrontendMenuBuilder::getMenu` - you can just use `$root->getLevel() + 1` instead (KnpMenu starts its level count at zero) - because the menu is built depth first and as soon as a child is added to its parent it will have the correct level when `getMenu` is called recursively on it.

richardhj:
This looks fine! Please create a PR and also include your suggestions :-)

core-bundle/src/Resources/contao/modules/Module.php
fritzmg: Yes, this seems to be a compatibility leftover from Contao 2. It's not even used in Contao 3.5.40 at least. But, for BC you will need to keep this, otherwise extending from any `Contao\Module` and overriding this protected method will break. I'll incorporate a deprecation in my PR.

fritzmg: Actually this is a BC break introduced in Contao `3.5.7`. Prior to that version you were able to override the language - but Contao `3.5.7` introduced the usage of `PageModel::getFrontendUrl`, without carrying this parameter over. Technically we could fix this in 4.9 and thus we would need to continue supporting this, but that's probably not worth it … So I would just deprecate this parameter now and remove it in 5.0 for good.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -499,9 +444,13 @@ protected function compileNavigationRow(PageModel $objPage, PageModel $objSubpag
 	 * @param boolean $blnIsSitemap  If true, the sitemap settings apply
 	 *
 	 * @return array<array{page:PageModel,hasSubpages:bool}>|null
+	 *
+	 * @deprecated Will be removed in Contao 5.0, no replacement is given. This method was moved to FrontendMenuBuilder#findPagesByPid().
```
fritzmg, line 471: Then `FrontendMenuBuilder::findPagesByPid` is the replacement, is it not?

core-bundle/src/Event/FrontendMenuEvent.php
fritzmg: I rather meant that the event should may be store the module it originated from so that you could check the module type. Though I am not sure that's a good idea either.

core-bundle/src/Resources/contao/modules/Module.php
richardhj: I don't know why we would want to make this method public and static.

richardhj: Also see contao/contao#2935.

core-bundle/src/Event/FrontendMenuEvent.php
richardhj: The MenuBuilder should not have a dependency on Modules IMHO.

Toflar:
I was just analyzing one of my sites because it was a bit slow and I noticed a pretty stupid thing: There's a main navigation module on top of the page and a secondary in the footer. 
No data here is cached whatsoever so all the work of finding the pages is done twice.
Could this PR improve the situation @richardhj?

leofeyer:
> No data here is cached whatsoever so all the work of finding the pages is done twice.

The models should be cached in the registry.

Toflar:
The registry is completely irrelevant. Not a single query can be saved here. See `getPublishedSubpagesByPid()` which executes queries that circumvent the registry completely.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * @return array<array{page:PageModel,hasSubpages:bool}>
+     */
+    private function findPagesByPid(int $pid, bool $showHidden = false, bool $isSitemap = false): array
+    {
+        $time = Date::floorToMinute();
+        $blnBeUserLoggedIn = $this->tokenChecker->hasBackendUser() && $this->tokenChecker->isPreviewMode();
+        $unroutableTypes = $this->pageRegistry->getUnroutableTypes();
+
+        $pages = $this->connection
```
Toflar, line 186: This PR would finally allow us to improve this logic a bit. We execute this query for every single page and it can never be cached. However, if we were to restructure the logic a bit, we could maybe improve this.
Given the following structure:

- 1
    - 2
        - 21
        - 22
        - 23
    - 3
        - 31
        - 32
        - 33

Currently, if we render the navigation for ID 1, it will execute this query for every single page. However, e.g. for ID 2 and 3 we already know the respective 3 subpages. So instead of requesting that query for every single one of them, maybe they could be loaded in a single one?

core-bundle/src/Resources/contao/modules/Module.php
fritzmg: It does not need to be static. If you are referring to my usage of `::` - I am not sure using `#` to denote object methods is common? Even PHP uses `::` for everything.

richardhj: If we want others to use this method, we would need to extract this method into some kind of repository. Here, it should stay an internal method IMHO.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Mhhh, I am unsure about this.

First, the query actually does not return the IDs of the next level.
Second, recursive queries are only possible from MySQL 8+; it looks like: https://stackoverflow.com/a/33737203

All I can offer is that we cache the information whether the page has subpages or not. This will save some resources for re-querying empty pages. See https://github.com/richardhj/contao/commit/6587efc470c02bc2cfa4969619378ddd8cb7e111 for this purpose.
However, this comes with problems, e.g., different navigation modules on the same page differ in the $showHidden option.

But maybe this is all about my inability to write complex MySQL queries 😄

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,362 @@
...
+            'showLevel' => 0,
+            'hardLimit' => false,
+            'isSitemap' => false,
+        ], $options);
+
+        $pages = $this->getPages($pid, $options);
+
+        $request = $this->requestStack->getCurrentRequest();
+        /** @var PageModel|null $currentPage */
+        $currentPage = $request->attributes->get('pageModel');
```
fritzmg: In an ESI request this attribute will only contain the page ID. Until we have a general service to retrieve the current page model, I think we still need to use `$GLOBALS['objPage']` here. /cc @aschempp

core-bundle/src/Menu/FrontendMenuBuilder.php
Toflar: Maybe @ausi has an idea? 😊

core-bundle/contao/modules/ModuleNavigation.php
```diff
@@ -69,22 +69,10 @@ protected function compile()
...
-
-		// Overwrite the domain and language if the reference page belongs to a different root page (see #3765)
-		if ($this->defineRoot && $this->rootPage > 0)
-		{
-			$objRootPage = PageModel::findWithDetails($this->rootPage);
-
-			$lang = $objRootPage->rootLanguage;
-			$host = $objRootPage->domain;
-		}
-
```
fritzmg: _Note:_ this code was outdated since a while now, since `PageModel::getFrontendUrl` will already generate an URL with the correct domain and language. That's why I removed it and also deprecated the corresponding method parameters of `Module::renderNavigation`.

fritzmg (changes requested):
There is still a problem when using a **Start level** in the navigation module for example. `FrontendMenuBuilder::populateMenuItem` will not get executed for the first levels and thus in the `FrontendMenuEvent` the page's properties are missing, which makes it difficult to add any submenu items via the event.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: or use the PageModelAdapter to fetch the page

richardhj: awaiting feedback from Andy concerning 02f81fd5b7657ee522b428516f5e0b7a5b16ec98.

richardhj:
> There is still a problem when using a **Start level** in the navigation module for example. `FrontendMenuBuilder::populateMenuItem` will not get executed for the first levels and thus in the `FrontendMenuEvent` the page's properties are missing, which makes it difficult to add any submenu items via the event.

Have you debugged this? Do we need to call `populateMenuItem()` twice? or split up the function?

richardhj:
BTW we haven't discussed yet if we want to trigger the event for every page at all. An alternative would be to trigger the event only for root-level items (i.e., `$item->getParent() is NULL`). I guess it is also a matter of performance. Just saying 🤷

fritzmg:
> BTW we haven't discussed yet if we want to trigger the event for every page at all. An alternative would be to trigger the event only for root-level items (i.e., `$item->getParent() is NULL`). I guess it is also a matter of performance. Just saying 🤷

For performance it's probably better to fire the event for every page (and it's more convenient), otherwise every event subscriber who wants to alter a menu item depending on a page's settinga has to traverse the whole tree again every time. 

I've made a small demo extension for testing: https://github.com/inspiredminds/contao-news-navigation/. Currently, all I have to do is [this](https://github.com/inspiredminds/contao-news-navigation/blob/187e7f5d5e81692f7e7f0b66be8916bae3798a63/src/EventListener/FrontendMenuListener.php#L39) in order to check whether the current menu item should receive additional sub items.

fritzmg:
> Have you debugged this? 

Yes, the problem is that currently the root item never gets populated. This is fine for the custom navigation, as the root item is just "virtual", but for the regular navigation the root item is always a real page. Either the website root - or a regular page, in case the **Start level** is greater than zero.

> Do we need to call `populateMenuItem()` twice? or split up the function?

I have an idea on how to restructure `getMenu` so that we don't have to call `populateMenuItem` twice. I'll create a PR again 🙃

core-bundle/src/Menu/FrontendMenuBuilder.php
fritzmg: May be we should put this in a common service. Currently we have `ContaoContext::getPageModel` and `AbstractFragmentController::getPageModel` which do (almost) exactly the same thing, i.e. we have duplicated code.

richardhj: 🤔 yes, someone decide pls

core-bundle/src/Menu/FrontendMenuBuilder.php
m-vo: In my experience the "with" or "connect by prior" functionality of DBMS are most of the time more suited for getting a sorted list (in "tree form") than to save performance. And as @richardhj already stated, it's not supported in older MySQL versions.

If it's just about saving requests, it would probably be faster and much simpler to just get a projection of all records - in this case a `SELECT id, pid FROM tl_page WHERE <published, not hidden, …>` (once) to build a lookup table.

core-bundle/src/Menu/FrontendMenuBuilder.php
m-vo: I had the same issue in the image studio code as well and ended up using `$GLOBALS['objPage']` because there was no better solution. Big :+1: for a service from my side.

core-bundle/src/Resources/contao/modules/Module.php
m-vo: @fritzmg Here is an example from the Doctrine docs where `#` is used to denote object context: https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/query-builder.html#sql-query-builder I kind of like this :slightly_smiling_face:

fritzmg: But `__METHOD__` always outputs `::` and you use `parent::` in object methods as well ;)

m-vo: Yes, and you can (still) use `->` to call static methods. These things … ehm … "haven't developed yet". :smile:

fritzmg: > If we want others to use this method, we would need to extract this method into some kind of repository. Here, it should stay an internal method IMHO.

Yes, just saying I am not sure whether we should really deprecate a public method without providing an alternative.

richardhj:
I decided to trigger the MenuEvent only once in 98abdf960ac4fc230252237f5aadd978a172a27e, meaning, for the entire menu the event is only triggered once instead of for every single item. And this is why:
- We end up very easily firing the event a dozen times for a single navigation menu, even if there is no listener making use of the event. I don't think there is a necessity for it.
- When the event is triggered for a non-root item, traversing up to the root level (`$item->getParent()`) did not work. So if you, for example, want to know the root $pid of the menu item in question, listening on a child did not work out either.

fritzmg (changes requested):
The current approach still does not populate the extras for the root item. The problem is that you are checking and populating everything only in the `foreach` loop of the children of the current element, which will always leave out the root node. Personally I prefer [my approach](https://github.com/richardhj/contao/pull/3) which ensures that every node is processed equally. I was also planning on adjusting the implementation, so that you do not have to take care of the CSS classes and `isActive` extra etc. yourself when implementing a listener and only use `$item->setCurrent(true);` for example.

So in essence I would implement:

```php
if (null === $root->getParent()) {
    // For the root level, trigger the event to allow modifying the menu
    $menuEvent = new FrontendMenuEvent($this->factory, $root, $pid, $options);
    $this->dispatcher->dispatch($menuEvent);

    // Traverse the menu to set navigation module specific extras
    $this->setNavigationExtras($root);
}
```

richardhj:
> The current approach still does not populate the extras for the root item. The problem is that you are checking and populating everything only in the `foreach` loop of the children of the current element, which will always leave out the root node. Personally I prefer [my approach](https://github.com/richardhj/contao/pull/3) which ensures that every node is processed equally. I was also planning on adjusting the implementation, so that you do not have to take care of the CSS classes and `isActive` extra etc. yourself when implementing a listener and only use `$item->setCurrent(true);` for example.
> 
> So in essence I would implement:
> 
> ```
> if (null === $root->getParent()) {
>     // For the root level, trigger the event to allow modifying the menu
>     $menuEvent = new FrontendMenuEvent($this->factory, $root, $pid, $options);
>     $this->dispatcher->dispatch($menuEvent);
> 
>     // Traverse the menu to set navigation module specific extras
>     $this->setNavigationExtras($root);
> }
> ```

I considered this as well and sadly, you have to traverse all items again in `setNavigationExtras()`.
I think we are totally fine if we move the logic into here: https://github.com/contao/contao/pull/3838/files#diff-072e13a8027ac3afc897fcf9634710245dd6b6a962ce2c759e3b48cc5d6b439cR444 because this is all about legacy compatibility anyways.

fritzmg:
> because this is all about legacy compatibility anyways.

Well Contao's navigation module isn't exactly legacy. There is no alternative navigation module available. But yes, if we want to avoid to traverse the menu tree again, then any event subscriber needs to set the appropriate extras for the navigation module itself.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            if (0 === strncasecmp($href, 'mailto:', 7)) {
+                return StringUtil::encodeEmail($href);
+            }
+
+            return $href;
+        }
+
+        if ('root' === $pageModel->type) {
+            // Overwrite the alias to link to the empty URL or language URL (see #1641)
+            $pageModel->alias = 'index';
```
aschempp: This looks strange, did that exist already somewhere? Since the first subpages does not necessarily have the `alias` index, this could create incorrect links.

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            } else {
+                $jumpTo = $this->pageModelAdapter->findFirstPublishedRegularByPid($pageModel->id);
+            }
+
+            // Hide the link if the target page is invisible
+            if (
+                !$jumpTo instanceof PageModel
+                || (!$jumpTo->loadDetails()->isPublic && !$this->tokenChecker->isPreviewMode())
+            ) {
+                $menuItem->setDisplay(false);
```
aschempp: how does this work? Only fetching published pages, but then add the page as invisible if it's not published? Also, we're still adding the page if it is not published? Or did you maybe want to check for `tl_page.hide` if it should be hidden in the navigation?

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+    private function populateMenuItem(ItemInterface $item, Request $request, PageModel $page, ?string $href, bool $hasSubmenu, array $options = []): void
+    {
+        /** @var PageModel|int|null $currentPage */
+        $currentPage = $request->attributes->get('pageModel');
+
+        // Support ESI requests
+        if (null !== $currentPage && !$currentPage instanceof PageModel) {
+            /** @var PageModel|null $currentPage */
+            $currentPage = $this->pageModelAdapter->findByPk($currentPage);
+        }
```
aschempp: This will not work with inline fragments. They also have a numeric `pageModel` but should use the global object (see https://github.com/contao/contao/pull/2506). Also, this logic would be re-run for every single menu item?

core-bundle/src/Menu/FrontendMenuBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            $item->setLinkAttribute('title', $title);
+        }
+
+        if ($page->accesskey) {
+            $item->setLinkAttribute('accesskey', $page->accesskey);
+        }
+
+        if ($page->tabindex) {
+            $item->setLinkAttribute('tabindex', $page->tabindex);
+        }
```
aschempp, line 306: `tabindex` support seems to be deprecated (https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/modules/Module.php), are we intentionally supporting it here?

core-bundle/contao/modules/Module.php
```diff
@@ -276,130 +275,71 @@ protected function getResponseCacheTags(): array
 	 *
 	 * @return string
 	 */
-	protected function renderNavigation($pid, $level=1, $host=null, $language=null)
+	protected function renderNavigation($pid, $level=null, $host=null, $language=null)
 	{
-		// Get all active subpages
```
aschempp: shouldn't we deprecated this whole method?

core-bundle/src/Event/FrontendMenuEvent.php
```diff
@@ -0,0 +1,60 @@
...
+class FrontendMenuEvent extends Event
+{
+    private FactoryInterface $factory;
+    private ItemInterface $menu;
+    private int $pid;
+    private array $options;
+
+    public function __construct(FactoryInterface $factory, ItemInterface $menu, int $pid, array $options)
+    {
+        $this->factory = $factory;
```
aschempp, line 28: why is the factory passed to the event? It seems to be a container service so one should use dependency injection in the event listener instead of getting it from the event?

core-bundle/src/Menu/FrontendMenuBuilder.php
fritzmg: It will be removed in Contao 5.0.

core-bundle/src/Event/FrontendMenuEvent.php
fritzmg: The `Contao\CoreBundle\Event\MenuEvent` does the same, but I am also not sure why.

richardhj: Official docs do so as well: https://symfony.com/bundles/KnpMenuBundle/current/events.html#create-the-event-object
It may be because the concrete Factory can differ from the one in the container, no?.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: This is the reference: https://github.com/contao/contao/blob/846437e41504ce7b0c864fe73436d68f6f5aaaf6/core-bundle/src/Resources/contao/modules/ModuleCustomnav.php#L113-L117
but maybe this is some legacy stuff as well.

fritzmg: Yeah, it's to make the root URL pretty, in case the page is rendered as part of a navigation. Otherwise the URL will be `https://example.com/alias-of-the-root-page` instead of just `https://example.com`.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: I did not change the semantics here either:
https://github.com/contao/contao/blob/4166f4eeccafb9cdcbd15e5e89728bbb4f1f3a81/core-bundle/src/Resources/contao/modules/Module.php#L367-L370

But I think returning NULL here makes more sense so it doesn't get added to the menu in the first place.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: The deprecation notice is still in place:

https://github.com/contao/contao/pull/3838/files/9ebce42ab5ed85b481c129bd7c959f3a8b3a3755#diff-fbaac5f5ae13f5efd13bd4be3057a998ae9fdd7f762f39b8839911bd5eb11f8fR107

However, why is it deprecated for the case being > 0?

core-bundle/contao/modules/Module.php
richardhj: Hmm, what should be the exact concept? Right now this method is used. Like it is called in ModuleNavigation.
But maybe in Contao 5, we can move away from the old navigation modules and templates.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: See a8cb646f3976153edbe1148bb55a8849ac6bcc8d

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: See f5105224bb51236ad27408f64a21239f32cd11dd

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Checked the related issue. Only tabindex > 0 is deprecated. Otherwise, the attribute stays intact.
See #3141 and #2949.

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Replicated the code in f5105224bb51236ad27408f64a21239f32cd11dd again. But I am not copying the tests for this method 🤣 We need a general service the other day.

richardhj:
@fritzmg can you update your extension in consideration of 4141108ce75cfb4a054375187bbff50fc59044fd?

core-bundle/src/Event/FrontendMenuEvent.php
aschempp: Or because KnpMenu also works without a DI container. But I guess its fine this way as well 😊

core-bundle/contao/modules/Module.php
aschempp: Ah, I assumed the navigation module uses KnpMenu itself. Don't we do that for BC or another reason?

core-bundle/src/Menu/FrontendMenuBuilder.php
aschempp: which means you can only have `tabindex = 0` which means none?

core-bundle/src/Menu/FrontendMenuBuilder.php
aschempp: but that (also existing) code does not make any sense to me, does it to you? Or does `isPublic` also check whether the root page is published? I think we have `isRootPublished` for that, which _might_ make sense …? @contao/developers

core-bundle/src/Menu/FrontendMenuBuilder.php
aschempp: I understand what it does, but its still wrong. We would need to fetch the first published subpage and render that instead. Apart from the fact that re-implementing this all the time is a PITA and should be done by routing, but unfortunately that does not work (yet) 🙄 (see https://github.com/contao/contao/pull/2031)

core-bundle/src/Menu/FrontendMenuBuilder.php
aschempp: see my suggestion in https://github.com/contao/contao/pull/3930

core-bundle/src/Menu/FrontendMenuBuilder.php
richardhj: Tabindex of 0 and -1 are the most common values. I wonder if there are logical values besides these two, but the allowed values are more than just 0.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

core-bundle/contao/modules/Module.php
richardhj: The idea is to have best backwards compatibility and this is why I kept the renderNavigation method that is returning a string. This method takes the KnpMenu and transforms the values to be compatible with the nav_default template.
Both the mod_navigation and nav_default templates stay compatible and current.

Have a look at the Module#compileNavigationItem in where the KnpMenuItem is transformed to an array being compatible with nav_default.

We can, however, think about an upgrade path. This ideally would look like to only use the Twig integration. This, however, is something that we should target in Contao 6.

core-bundle/src/Menu/FrontendMenuBuilder.php
fritzmg: So we keep it the way it is and always was for now, no? Until we have something better.

fritzmg:
> @fritzmg can you update your extension in consideration of 4141108?

Yes, but the following still needs to be dealt with imho:

> The current approach still does not populate the extras for the root item. The problem is that you are checking and populating everything only in the foreach loop of the children of the current element, which will always leave out the root node.

You did not provide any further feedback for [my PR](https://github.com/richardhj/contao/pull/3) yet. I would still prefer my approach as this would ensure that all pages are treated equally.

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -492,6 +434,31 @@ protected function compileNavigationRow(PageModel $objPage, PageModel $objSubpag
...
+		$item['subitems'] = $subitems;
+		$item['isActive'] = $menuItem->isCurrent();
+		$item['link'] = $item['title'];
+		$item['title'] = StringUtil::specialchars($menuItem->getLinkAttribute('title'), true);
+		$item['pageTitle'] = StringUtil::specialchars($item['pageTitle'], true);
+		$item['description'] = str_replace(array("\n", "\r"), array(' ', ''), $item['description']);
+		$item['href'] = $menuItem->getUri();
+		$item['nofollow'] = (strncmp($menuItem->getLinkAttribute('rel', ''), 'noindex,nofollow', 16) === 0); // backwards compatibility
+		$item['target'] = ($v = $menuItem->getLinkAttribute('target')) ? " target=\"$v\"" : '';
+		$item['rel'] = ($v = $menuItem->getLinkAttribute('rel')) ? " rel=\"$v\"" : '';
```
fritzmg, line 457: Btw. I think we can put even more things into `Module::compileMenuItem` that are mostly only related to the current navigation module of Contao. E.g. CSS class generation: you already provide the `pageModel` in the extras and thus you could also let `Module::compileMenuItem` set the appropriate CSS classes based on the `pageModel` instead of the `FrontendMenuBuilder`. Would make the `FrontendMenuBuilder` cleaner yet again :). Just a thought.

fritzmg:
@richardhj will you continue to work on this? :) I still think my changes (see PR) should be incorporated as the navigation would then be built fully recursively and thus always also include all necessary information in the root item.

richardhj:
> @richardhj will you continue to work on this? :) I still think my changes (see PR) should be incorporated as the navigation would then be built fully recursively and thus always also include all necessary information in the root item.

See richardhj/contao#3.

PR #3839, opened 2021-12-17
Remove page from index if "Do not search" is checked

aschempp:
fixes #3766
fixes #3776

As pointed out by @tsarma, we should purge the search index if the `noSearch` checkbox is enabled in the page settings. Since I'm working on the same class in https://github.com/contao/contao/pull/3835 (and follow-up PRs) I took the liberty of implementing this and also checking `noindex` in the `robots` field. @Toflar is it correct this does not index the page as well?

As discussed at the Contao Call yesterday, purging the search on alias change was implemented in Contao 4.12, so we see this as a bug fix for 4.12 and not Contao 4.9 (see https://github.com/contao/contao/pull/3776#issuecomment-988063486).
Thinking of this, we could very easily add the new `PageSearchListener` class to Contao 4.9 though and just remove the methods in `PageUrlListener` from 4.12, what do you think @contao/developers ?

@Toflar pointed out in https://github.com/contao/contao/pull/3766 that this is only an intermedia solution, because we should ultimately purge through the search indexer and not directly through the Contao search engine. However this will be much more complex and out of scope for now.


Toflar (dismissed):

fritzmg (dismissed):
> Thinking of this, we could very easily add the new `PageSearchListener` class to Contao 4.9 though and just remove the methods in `PageUrlListener` from 4.12, what do you think @contao/developers ?

Agreed. Though I am fine either way.

aschempp:
As discussed with @leofeyer I actually rebased this on 4.13 only, since it would be quite the pita to merge upstream and could be seen as a feature anyway 🙃

Toflar (dismissed):

leofeyer (approved):

PR #3841, opened 2021-12-17
Remove an unused setup method

aschempp:
Figured out this setup method is never used because every test method calls `getContainerWithFrameworkTemplate` which sets `System::setContainer` anyway.


bytehead (approved):
If tests pass 👍

PR #3843, opened 2021-12-17
Add route priority and allow the same page alias with different parameters

aschempp:
This PR *finally* adds one of the most requested features to Contao: having two separate pages for a `news.html` page with the news list, and `news/foobar.html` for the details of a news with alias `foobar`. Two separate pages means two separate page layouts etc.

However, with great power comes great responsibility! It is now possible to configure two pages that listen to the same URL. To work around this, there are multiple precautions:

**1. Handling of URL parameters**

For `regular` pages in Contao, it is still not possible to create two pages with the same alias and the same configuration. It is however possible to create one page with `requireItem` enabled and one with `requireItem` disabled. That's actually how one can realise the behaviour described above in a plain Contao 4.13.

Contao will automatically prioritise a page with `requireItem` enabled so it will only render the non-`requireItem` page if there is nothing but the alias in the URL. 

If a custom page type follows the standard Contao behaviour (`path=null`, alias and `requireItem`), it will just behave exactly the same.

**2. Handling of custom page types with parameters** 

Our new page controllers can have additional route parameters thanks to Symfony routing. In my current client project, I have the following page controller annotation:

```
/**
 * @Page(path="{country}/{type}/{id}-{title}", requirements={"id"="\d+"})
 */
```

This means the respective page type will match a URL that starts with the page alias and ends with country, type and id/title. For example this would be `my-alias/switzerland/foobar/17-this-is-a-title.html`. I can - at the same time - have a regular page for `my-alias.html`, `my-alias/switzerland.html` etc.
In extreme cases, we cannot "know" what page to prioritise, therefore the user has to do that. I have added a route priority configuration to the page settings. Contao will also tell the user if there are pages that potentially conflict with the current one, thanks to Symfony route matching.

Bildschirmfoto 2021-12-17 um 16 27 33


core-bundle/src/EventListener/DataContainer/PageRoutingListener.php
```diff
@@ -0,0 +1,111 @@
...
+
+        return $this->twig->render(
+            '@ContaoCore/Backend/be_route_conflicts.html.twig',
+            [
+                'conflicts' => $conflicts,
+            ]
+        );
+    }
+
+    private function buildUrl(string $alias, string $urlPrefix, string $urlSuffix): string
```
Toflar, line 111: Can you explain why we don't use the router here? This is confusing for me. If there's a technical reason for that, I think it's worth a comment in the code for future reference :)

Toflar (commented):
Looks good to me :) Should we have a routing functional test for this, like for all the other routing related features?

core-bundle/src/EventListener/DataContainer/PageRoutingListener.php
aschempp: added in 24ad8b84429e4b58c92b4c20f01dbf062a6decc3

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -191,65 +199,85 @@ private function recursiveValidatePages(int $pid, PageModel $rootPage): void
...
+        if (null === $currentUrl || 0 === $routeCollection->count()) {
+            return false;
         }
 
-        return false;
+        $request = Request::create($currentUrl);
+
+        try {
+            $this->routeMatcher->finalMatch($routeCollection, $request);
+            // TODO return value is the which page conflicts, should we display that?
```
leofeyer: What about this comment?

leofeyer (dismissed):

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
aschempp: I suggest we merge it as-is and add a new issue to discuss this (maybe in the next call).

leofeyer (dismissed):

leofeyer (approved):

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
leofeyer: See #3903

PR #3846, opened 2021-12-19
Fixed debug:pages command and show dynamic content composition

aschempp:
While implementing https://github.com/contao/contao/pull/3845 I wanted to show if a page has dynamic content composition in the `debug:pages` command. However, the command didn't work anymore since we renamed the service name!

I also ended up removing the autoconfiguration for our page controller interfaces, since they are only causing issues. Even though I registered a page controller with custom name (`error_404` etc.), the interface caused a page controller with `error4xx` to be registered.

### DIFF
--- a/core-bundle/src/Command/DebugPagesCommand.php
+++ b/core-bundle/src/Command/DebugPagesCommand.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Command;
 
 use Contao\CoreBundle\Framework\ContaoFramework;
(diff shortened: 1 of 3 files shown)


Toflar (approved):

leofeyer:
> I also ended up removing the autoconfiguration for our page controller interfaces, since they are only causing issues. Even though I registered a page controller with custom name (`error_404` etc.), the interface caused a page controller with `error4xx` to be registered.

Shouldn't we try to fix the problem instead? @contao/developers Are you using this in your apps or bundles?

PR #3848, opened 2021-12-20
Add file previews for downloads

ausi:
~~Currently a draft ☺️~~

This pull request adds a preview factory that can be used to get a preview image for any file:

Adding support for new file formats should be as easy as creating a service that implements `PreviewProviderInterface` and is tagged with `contao.preview_provider`.

ToDo:

- [x] Agree on the concept
- ~~Implement deferred previews~~ Not easy it seems 😭
- [x] Discuss if we need to support multiple pages/frames like multipage PDFs, videos, animated GIF and so on.
- [x] Use the previews in the file manager
- [x] Use the previews in the download element
- [x] Use the previews in the downloads element
- [x] Check if we need to integrate somethings from filesystem PR #3774
- [x] Check if there are more file formats that we want to support out of the box (currently everything Imagine (Imagick, Gmagick or GD) can handle is supported) => _no (easy to implement) file format came up so far so I think it is fine for now. Also adding support for more formats can perfectly be done later or by an extension._


core-bundle/src/Image/Preview/PreviewProviderInterface.php
```diff
@@ -0,0 +1,44 @@
...
+interface PreviewProviderInterface
+{
+    public function supports(string $path, string $fileHeader = ''): bool;
+
+    /**
+     * @param int $size Minimal size the preview image should have in each
+     *                  dimension. Can be larger for implementation reasons or
+     *                  smaller if there is not enough data.
+     */
+    public function generatePreview(string $sourcePath, int $size, string $targetPath): void;
```
Toflar: What if generating fails?

core-bundle/src/Image/Preview/PreviewProviderInterface.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Calculate the exact dimensions the preview image will have when generated.
+     *
+     * @param int    $size       See self::generatePreview()
+     * @param string $fileHeader First X bytes of the source file
```
Toflar: What's `X bytes`, who defines how many bytes are required for the file header?

Toflar: Maybe loop over all providers and ask how many bytes they need at least to determine the file header and then take `max()`?

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -0,0 +1,197 @@
...
+        // Supported image formats do not need an extra preview image
+        if (\in_array(strtolower(pathinfo($path, PATHINFO_EXTENSION)), $this->validImageExtensions, true)) {
+            return $path;
+        }
+
+        // Round up to the next highest power of two
+        $size = (int) (2 ** log($size, 2));
+
+        // Minimum size for previews is 512
+        $size = max(512, $size);
```
Toflar: Any technical reason for this? Or should this be configurable? If not, then it should probably be a public constant as I can imagine other services might want to check for this. (or have a `isSizeBigEnoughWhatever()`).

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -0,0 +1,197 @@
...
+        $cachePath = $this->createCachePath($path, $size);
+        $globPattern = preg_replace('/[*?[{\\\\]/', '\\\\$0', $this->cacheDir.'/'.$cachePath).'.*';
+
+        foreach (glob($globPattern) as $cacheFile) {
+            if (\in_array(pathinfo($cacheFile, PATHINFO_EXTENSION), $this->validImageExtensions, true)) {
+                return $cacheFile;
+            }
+        }
+
+        $first1024Bytes = file_get_contents($path, false, null, 0, 1024);
```
Toflar: Ah, here are the bytes. Again, maybe there's a provider that needs 2 KB to determine support. We should fetch that from them.

Toflar (commented):
> Discuss if we need to support multiple pages/frames like multipage PDFs, videos, animated GIF and so on.

I don't think we need that for a start. But we could already support an `array $options` on our interface to pass on provider-specific information. If e.g. we wanted to have a configuration on PDF files in our DBAFS as to which page to use for previews, then this info could be passed on via `$options`.

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -0,0 +1,197 @@
...
+
+        $provider = null;
+
+        foreach ($this->previewProviders as $provider) {
+            if ($provider->supports($path, $first1024Bytes)) {
+                try {
+                    $format = $provider->getImageFormat($path, $size, $first1024Bytes);
+                    $targetPath = $this->cacheDir.'/'.$cachePath.'.'.$format;
```
m-vo:
```suggestion
                    $targetPath = Path::join($this->cacheDir, "$cachePath.$format");
```
:relaxed:

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -0,0 +1,197 @@
...
+        $hashData = [
+            Path::makeRelative($path, $this->cacheDir),
+            (string) $size,
+            (string) filemtime($path),
+        ];
+
+        $hash = substr(md5(implode('|', $hashData)), 0, 9);
+        $pathinfo = pathinfo($path);
+
+        return $hash[0].'/'.$pathinfo['filename'].'-'.substr($hash, 1);
```
m-vo: path util eye triggered :eye: 

```suggestion
        return Path::join($hash[0], $pathinfo['filename'].'-'.substr($hash, 1));
```

core-bundle/src/Image/Preview/PreviewFactory.php
ausi: I want to reduce the number of preview images that have to be generated, as this is a non-performant task. A constant is the way to go IMO. 👍

core-bundle/src/Image/Preview/PreviewProviderInterface.php
ausi: Will create a custom exception for that and add an `@throws` here.

core-bundle/src/Image/Preview/PreviewProviderInterface.php
ausi: > Maybe loop over all providers and ask how many bytes they need at least to determine the file header and then take `max()`?

👍 This way we can also not read the file at all if no provider needs it 💪

core-bundle/src/Image/Preview/PreviewFactory.php
ausi: 60ac9b4e27389b262981f671daaf19a6b577897c

core-bundle/src/Image/Preview/PreviewFactory.php
ausi: 1cd029830d63475ec9d74eca2c19945e0852ffd2

core-bundle/src/Image/Preview/PreviewProviderInterface.php
ausi: 1cd029830d63475ec9d74eca2c19945e0852ffd2

core-bundle/src/Image/Preview/PreviewProviderInterface.php
ausi: 80b270b1b0084668baa8c02c2d33eac648208a28

ausi:
> But we could already support an `array $options` on our interface to pass on provider-specific information.

Done in 4771c7f3d6cce92ab9bf1d32dd130fd7f5c6b179

ausi:
I’m pretty happy with the state of this PR now. And will add unit tests and some fine tuning shortly.

The public interface of the preview factory now looks like this:

```php
$previewFactory->createPreview('/var/www/files/foo.pdf');
// Returns an image object with the configured default size pointing to
// "/var/www/assets/images/previews/9/foo-fbf35a7e-2.png"

$previewFactory->createPreview('/var/www/files/foo.pdf', 1024, 2, [/* options */]);
// Returns an image object of the second page of the PDF with a minimum size
// (in each dimension) of 1024 pixels and passing the options to the providers

$previewFactory->createPreviewImage('/var/www/files/foo.pdf', [100, 100, 'crop']);
// Returns an image object that was resized based on the preview image that fits
// the required dimensions

$previewFactory->createPreviewPicture('/var/www/files/foo.pdf', new PictureConfiguration(…)); // picture object
$previewFactory->createPreviewFigure('/var/www/files/foo.pdf', '_predefined_size'); // figure object
$previewFactory->createPreviewFigureBuilder('/var/www/files/foo.pdf', '_predefined_size'); // figure builder object

$previewFactory->createPreviews('/var/www/files/foo.pdf', 1024, 5, 1, [/* options */]);
// Returns an array of image objects of the first five pages of the PDF

$previewFactory->createPreviewImages('/var/www/files/foo.pdf', [100, 100, 'crop']);
// Returns an array of image objects for all pages in the PDF that got resized
// based on the preview images that fits the required dimensions

$previewFactory->createPreviewPictures('/var/www/files/foo.pdf', new PictureConfiguration(…)); // picture object array
$previewFactory->createPreviewFigures('/var/www/files/foo.pdf', '_predefined_size'); // figure object array
```

And this is the interface the preview providers have to implement:

```php
interface PreviewProviderInterface
{
    public function getFileHeaderSize(): int;

    public function supports(string $path, string $fileHeader = ''): bool;

    /**
     * @param int      $size               Minimal size the preview image should
     *                                     have in each dimension. Can be larger
     *                                     for implementation reasons or smaller
     *                                     if there is not enough data.
     * @param \Closure $targetPathCallback Returns the target path without file
     *                                     extension for the specified page
     *                                     number
     *
     * @phpstan-param \Closure(int): string $targetPathCallback
     *
     * @throws UnableToGeneratePreviewException
     *
     * @return iterable<string> Target paths including the file extension
     */
    public function generatePreviews(string $sourcePath, int $size, \Closure $targetPathCallback, int $lastPage = PHP_INT_MAX, int $firstPage = 1, array $options = []): iterable;
}
```

The core has two providers by default, an `ImaginePreviewProvider` that uses the available `Imagine` adapter and a `FallbackPreviewProvider` (with a very low priority) that creates SVG images with file icons that show what format it is, like:  
xls-icon

Deferred previews are postponed for now as it is very difficult to base a deferred image on a deferred preview image. But I got some ideas how we can still get them with a different approach.

core-bundle/src/Resources/contao/elements/ContentDownload.php
```diff
@@ -140,6 +142,63 @@ protected function compile()
...
+			foreach ($previews as $index => $preview)
+			{
+				$builder->fromImage($preview);
+
+				if (!empty($lightboxPreviews[$index]))
+				{
+					$builder->setLightboxResourceOrUrl($lightboxPreviews[$index]);
+				}
+
+				$previews[$index] = $builder->build()->getLegacyTemplateData();
```
ausi:
```suggestion
				$previews[$index] = $builder->build();
```

No need to use legacy data here I think.

ausi:
As discussed with @Toflar on Slack:

- [x] Make the hash longer and include the app secret
- Do the same for the images (seperate PR) https://github.com/contao/image/issues/87
- ~~Don’t make the preview images accessible via http~~ no problem in being accessible if the file names cannot be guessed.
- ~~Do not generate previews in the file manager popup for private files `This file does not have a preview because it’s in a protected folder and preview files are publicly accessible`~~ see above.

core-bundle/src/Resources/contao/elements/ContentDownload.php
ausi: Done in 08140cb31f376e1d26dc395405ac2ef81f564409

Toflar (dismissed):

ausi:
The only missing todo is now to see if there is something we can or should prepare for the PR #3774

We could make the interface to also accept streams (resource handles) instead of only file paths, but in order to make extension developers lifes easier this should go into a seperate interface (so that not every provider has to support streams). And in that case it would be better to create this interface after the filesystem topic has settled IMO.
@m-vo agree? 🥺

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -326,6 +326,38 @@ static function (array $value): array {
                     ->prototype('scalar')->end()
                     ->defaultValue(['jpg', 'jpeg', 'gif', 'png', 'tif', 'tiff', 'bmp', 'svg', 'svgz', 'webp'])
                 ->end()
+                ->arrayNode('preview')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('target_dir')
+                            ->info('The target directory for the cached previews.')
+                            ->example('%kernel.project_dir%/assets/previews')
```
leofeyer, line 334: Why can't we reuse the `assets/images` directory?

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -326,6 +326,38 @@ static function (array $value): array {
...
+                            ->min(1)
+                            ->max(65535)
+                            ->defaultValue(512)
+                        ->end()
+                        ->integerNode('max_size')
+                            ->min(1)
+                            ->max(65535)
+                            ->defaultValue(1024)
+                        ->end()
+                        ->booleanNode('fallbacks')
```
leofeyer: Should we name this `fallback_images` or `show_fallback_image`?

core-bundle/src/Image/Preview/ImaginePreviewProvider.php
```diff
@@ -0,0 +1,206 @@
...
+    public function getFileHeaderSize(): int
+    {
+        return 0;
+    }
+
+    public function supports(string $path, string $fileHeader = '', array $options = []): bool
+    {
+        $format = strtolower(pathinfo($path, PATHINFO_EXTENSION));
+
+        return $this->imagineSupportsFormat($format);
```
leofeyer, line 46: This method throws an "Unsupported Imagine implementation" exception. Should we handle it or let it bubble up?

core-bundle/src/Resources/contao/elements/ContentDownload.php
```diff
@@ -140,6 +142,63 @@ protected function compile()
 		$this->Template->meta = $arrMeta;
 		$this->Template->extension = $objFile->extension;
 		$this->Template->path = $objFile->dirname;
+		$this->Template->previews = $this->getPreviews($objFile->path);
+	}
+
+	protected function getPreviews(string $path): array
```
leofeyer, line 148:
```suggestion
	private function getPreviews(string $path): array
```

core-bundle/src/DependencyInjection/Configuration.php
aschempp: Without checking into the code, shouldn't this be a developer decision for the element that creates the preview? I mean a download element might _know_ it always wants fallbacks, but other elements might not? Or something the user decides through a checkbox. It doesn't sound right to enable or disable this globally in the application.

ausi: We could do both, an option for the FallbackProvider as well as the current config. But I think disabling these images for the whole project is still a valid use case.

ausi: How about naming it `enable_fallback_images` or `enable_fallbacks`?

Toflar: Could be as "simple" as having a dropdown rather than a checkbox for preview. `disable`, `enable_with_fallback` and `enable_without_fallback`?

core-bundle/src/DependencyInjection/Configuration.php
ausi: We could, but letting two different services handle the same directory to store image files is not a good idea IMO. It might also make sense to mount assets/previews somewhere else for performance or storage reasons. I think a clear distinction here is better than reusing the assets/images folder.

core-bundle/src/Image/Preview/ImaginePreviewProvider.php
ausi: I think we should let it bubble up because it would be some sort of a misconfiguration.

core-bundle/src/DependencyInjection/Configuration.php
leofeyer: Not sure about that.

First of all, I would prefer to have all generated images in the same folder. Storing thumbnails in `assets/images` and previews in `assets/previews` will confuse people, because they have to know the difference in order to find a file.

Second of all, `assets/images` is optimized in terms of storage and can handle a large number of generated files. It might even be mounted somewhere else already. So why build everything again?

core-bundle/src/DependencyInjection/Configuration.php
leofeyer: I like `enable_fallback_images`. I also like @Toflar's suggestion if you decide to go this way.

Toflar: The global fallback is required anyway. E.g. to decide for the file manager.

aschempp: > E.g. to decide for the file manager.

Which could/should be a user config, similar to "show preview images"? 🤷

core-bundle/src/DependencyInjection/Configuration.php
ausi: How about keeping it how it is now but change the default for `preview.target_dir` to `assets/images/previews` ?
```suggestion
                            ->example('%kernel.project_dir%/assets/images/previews')
```

leofeyer: Do you mean storing all the files in one single folder or do you mean `assets/images/previews/0`, `assets/images/previews/1` and so on?

ausi: yes, `assets/images/previews/0`, `assets/images/previews/1` and so on.

ausi: ~~I will change the default to `assets/images/previews`~~

leofeyer: As briefly discussed, `assets/images/deferred` should be moved to `var/deferred-images` in Contao 5, therefore it makes sense to store the previews in `assets/previews` so we can have a "clean" `assets/images` directory in Contao 5 again. See https://github.com/contao/contao/issues/3898.

aschempp:
looking at https://github.com/contao/contao/issues/3898 this PR also means preview images of possibly protected files are now publicly available right? 🤔

Toflar:
Yes, which is why the hash is longer and includes the secret.

leofeyer:
> this PR also means preview images of possibly protected files are now publicly available right? 🤔

This is a problem indeed. A minified version of an image is one thing, but a minified version of a confidential text file or PDF is another thing. Isn't this a potential information disclosure issue?

Toflar:
@ausi and I have already discussed this. That's why the hash is longer and the secret is part of the hash so it's unguessable.

ausi:
The file name includes a keyed hash with over 2^65 possibilities. It should be impossible to brute force that using HTTP requests.

leofeyer:
We could also store the preview images in a non-accessible location and inline them in the HTML document. Since previews are small, this should not generate too much overhead, right?

ausi:
This would not work as resized versions of the previews would show up in assets/images anyways.

> Since previews are small, this should not generate too much overhead, right?

Previews are as large as any other images, this depends completely on the use case.

leofeyer:
As briefly discussed in Slack, we want to merge the PR as is. The obfuscated URLs will look like `assets/previews/w/06-ticket-a6B6UBSExR.png` and should be unguessable via HTTP brute-force.

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -0,0 +1,532 @@
...
+        $hashData = [
+            Path::makeRelative($path, $this->cacheDir),
+            (string) $size,
+            (string) filemtime($path),
+            ...array_keys($previewOptions),
+            ...array_values($previewOptions),
+        ];
+
+        $hash = hash_hmac('sha256', md5(implode('|', $hashData)), $this->secret, true);
+        $hash = substr(strtr(base64_encode($hash), '+/', '-_'), 0, 11);
```
Toflar: @ausi @leofeyer  what do you think about making `11` dynamic to increase security even more? It has to be deterministic of course but if it's based on the secret and the $hash it would increase the number of possibilities even more.

```suggestion
        $length = 11 + hexdec(md5($hash . $this->secret)[0]);
        $hash = substr(strtr(base64_encode($hash), '+/', '-_'), 0, $length);
```

Maybe you have a better idea but this variant would generate lengths between 11 and 26.

leo: Wrong Leo!

ausi: I’m not sure if this is how to calculate the brute force possibilities in this case, but I’ve got a 1/16th chance that the length is 11, this can (maybe?) just be multiplied with the probability of 1/(2^65) which would be less then increasing the length from 11 to 12.

So just increasing the length would be better I think (and less magic).

Toflar: Aaah combinatorics :) Wasn't my strength but imho this is permutation (order matters) with repetition (every hex char can appear multiple times). So that would mean

Fixed to 11: `16^11 = 1.75 * 10^13`
Fixed to 26: `16^26 = 2.02 * 10^31`

And if we have a range between 11 and 26, that should be the summary of all of those.
So `16^11 + 16^12 + .... + 16^26`. Which would end up being something like `2.16 * 10^31`.

So yeah, we can just increase the length of the hash, making it dynamic does not add a lot more possibilities.

Toflar: I mean it does add a lot more (`0.14 * 10^31` is still a huge number) but compared to the already huge `2.02 ^ 10^31` it doesn't make any difference :D

Toflar: I'd be more comfortable with more than `11`, though :)

m-vo: It's easier to think of 'bits of information' in this case. A shorter input is basically just padded with zeroes - only length matters.

ausi: I try another go on this: 

Considering case-insensitive file systems, we have basically 38^11 possible values for the hash. Brute forcing 1000 URLs per second would take around 7.5 million years to try out every hash or an average of 3.8 million years to crack one single hash.


| Chars | Million years max | Million years average |
|-------|------------------:|----------------------:|
| 11    |                 8 |                     4 |
| 12    |               287 |                   144 |
| 13    |             10924 |                  5462 |
| 14    |            415111 |                207555 |
| 15    |          15774200 |               7887100 |
| 16    |         599419599 |             299709800 |

Increasing it to 12 should be enough IMO.

Toflar: What we're missing here is that the possibilities alone are not sufficient for this decision. We also have to account for the probability of shared hash information. See https://en.wikipedia.org/wiki/Birthday_problem. 
I don't really know but instead of arguing I would simply pick a number high enough (e.g. 20) so it doesn't matter at all. Saving 8 letters makes no sense imho.

m-vo: > We also have to account for the probability of shared hash information.

The probability of a hash collision only depends on the number of hashed elements. With around 1M images in the system and a length of 64bit (38 values per char = ~6bit * 11 = ~66bit) [the probability](https://stackoverflow.com/a/62667633/3146819) [is approx.](https://www.wolframalpha.com/input/?i=1+-+e%5E%28-k%5E2%2F2%5E%28n%2B1%29%29) 2,7 * 10^-8 or in other words: negligible.

ausi: > I don't really know but instead of arguing I would simply pick a number high enough (e.g. 20) so it doesn't matter at all.

That’s what I tried to do by increasing the 36 bits of contao/image to ~64 bits.

@contao/developers should we vote on this one?
- 👎 for 11 characters, e.g.: assets/previews/w/foobar-a6B6UBSExR.png
- 👍 for 16 characters, e.g.: assets/previews/w/foobar-a6B6UBSExRhbpJP.png
- 🚀 for 20 characters, e.g.: assets/previews/w/foobar-a6B6UBSExRhbpJPMTY5.png

leofeyer: This is currently blocking the PR from being merged. @Toflar You have not voted yet, I think.

Toflar: Just did. I simply don't see why we would want a short hash. It makes absolutely no sense to me.

ausi: Increased to 16 in 130d645107c72b519dca519cfb184d04a67b1d03

Toflar (dismissed):

core-bundle/src/DependencyInjection/Configuration.php
ausi: Renamed to `enable_fallback_images` in af2ae0f9022b5b555b648e547a5b9d46cb00774f

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -1080,6 +1088,10 @@ public function adjustDcaByType($dc)
 			case 'image':
 				$GLOBALS['TL_DCA']['tl_content']['fields']['imagemargin']['eval']['tl_class'] .= ' clr';
 				break;
+
+			case 'downloads':
```
leofeyer, line 1093: What about `case 'download'`? It the field not required there?

leofeyer:
There is a little UX flaw that we might want to fix:

core-bundle/src/Resources/contao/dca/tl_content.php
ausi: Fixed in a22492738d2564b14414a141c86dfaff6fed5cec

ausi:
> There is a little UX flaw that we might want to fix

Done in e6a1cc52218372f4e4f7be442d59a7abedda8988

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -326,6 +326,38 @@ static function (array $value): array {
                     ->prototype('scalar')->end()
                     ->defaultValue(['jpg', 'jpeg', 'gif', 'png', 'tif', 'tiff', 'bmp', 'svg', 'svgz', 'webp'])
                 ->end()
+                ->arrayNode('preview')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('target_dir')
```
m-vo, line 332: Something we might want to discuss related to #3774: 

With VFSs in place, we could/should probably deprecate all configurable paths and suggest to use mounting instead. This would include the image assets path, the upload dir, this new config and possibly more.

The current default (whatever it is `assets/images` or `var/images` …) would then become just an identifier. If we wanted to switch locations later on, we could use mounting ourselves. (The implementation would get a `VirtualFilesystem` prefixed with the identifier and would not care about the location at all.)

leofeyer:
> `StringUtil`? Looks pretty useful for other use cases.

Why are we not using https://github.com/ChristianRiesen/base32?

bytehead:
We already have https://github.com/paragonie/constant_time_encoding as a dependency for security related stuff that also encodes with base32: https://github.com/contao/contao/blob/4.x/core-bundle/src/Controller/FrontendModule/TwoFactorController.php#L166

So no need for @ausi to provide our own method - right? 🙃

ausi:
> So no need for @ausi to provide our own method - right? 🙃

I’m not sure, it doesn’t seem to be very performant as it tries to protect against timing attacks.

leofeyer (approved):

leofeyer:
Thank you @ausi.

bytehead:
@ausi what about a notice in `StringUtil::encodeBase32` to use the time constant one if used in security context?

PR #3850, opened 2021-12-20
Add a backup retention policy

Toflar:
This brings even more power to our new backup management CLI introduced in https://github.com/contao/contao/pull/3631.

Besides `keep_max` we can now also configure `keep_intervals` which by default is configured to `['1D', '7D', '14D', '1M']`.
`keep_max` is a very simple configuration. If you configure it to `10`, it will just keep 10 backups and on the 11th, the oldest one of those will get removed.
However, this is likely not the behavior you'd want to as if you create 10 backups today, chances are pretty high that you are not interested in any of the other 9 of today. However, you'd probably be interested in the backup of yesterday, a week ago or even a month ago.
Here's where the new `keep_intervals` configuration comes in. As soon as you configured it, it will always keep the one you just created, plus the **oldest** one of every interval.

So with its default configuration, Contao will now only every keep `5` backups as a maximum (remains unchanged) but even if you run `contao:backup:create` 10 times today, it will not keep the latest 5 but instead

* The latest one
* The oldest created from `latest one - 1 day`
* The oldest created from `latest one - 7 days`
* The oldest created from `latest one - 14 days`
* The oldest created from `latest one - 1 month`


aschempp (commented):
Nice addition! Since we have a `RetentionPolicy` class, wouldn't it make sense for the class to decide if the backup should be kept? That way we (or possibly anyone) could add new features to that class and the backup process would automatically "respect" it?

Toflar:
I thought about this but then the policy needs to be an interface as well. And we need to be able to configure a different service. All completely doable, just not sure if that's the current scope or if it's worth the effort. But I can certainly look into it.

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -608,9 +608,14 @@ private function addBackupNode(): NodeDefinition
                     ->scalarPrototype()->end()
                 ->end()
                 ->integerNode('keep_max')
-                    ->info('The maximum number of backups to keep.')
+                    ->info('The maximum number of backups to keep. Use 0 to keep all the backups forever.')
                     ->defaultValue(5)
                 ->end()
+                ->arrayNode('keep_periods')
+                    ->info('The latest backup plus the oldest of every configured period will be kept.')
+                    ->defaultValue([1, 7, 14, 30])
```
m-vo: Can we maybe support strings like `7d`, `2w`, `1m` here? I think this would make for a more meaningful config.

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -232,14 +233,44 @@ private function doRestore(RestoreConfig $config): void
...
+                }
             }
+        }
+
+        // Then cleanup according to maximum amount of backups to keep
+        if ($keepMax > 0) {
+            $i = 0;
 
-            ++$i;
+            foreach ($this->listBackups() as $backup) {
```
m-vo: What about `foreach (array_values($this->listBackups()) as $index => $backup) {` here and dropping the counter?

core-bundle/src/Doctrine/Backup/Config/RetentionPolicy.php
```diff
@@ -0,0 +1,48 @@
...
+    public function getKeepPeriods(): array
+    {
+        return $this->keepPeriods;
+    }
+
+    private function validateAndSortKeepPeriods(array $keepPeriods): array
+    {
+        foreach ($keepPeriods as $numberOfDays) {
+            if (!\is_int($numberOfDays)) {
+                throw new \InvalidArgumentException('$keepPeriods must be an array of integers.');
```
m-vo: We should IMHO annotate `$keepPeriods` with `array<int>` in the constructor. :slightly_smiling_face:

tools/ecs/config/default.php
```diff
@@ -37,5 +37,5 @@
     ;
 
     $parameters->set(Option::LINE_ENDING, "\n");
-    $parameters->set(Option::CACHE_DIRECTORY, sys_get_temp_dir().'/ecs_default_cache');
+    //$parameters->set(Option::CACHE_DIRECTORY, sys_get_temp_dir().'/ecs_default_cache');
```
m-vo: Accidental commit?

m-vo (commented):
Nice and helpful addition! :tada: 

Since `RetentionPolicy` is only an object holding config, we might want to close it, i.e. make it final with a private constructor and a static `createFromFoo(): self` method. This way we could add/change things (or open it) later on.

core-bundle/src/DependencyInjection/Configuration.php
Toflar: That was real PITA because `DateInterval` instances are not comparable by default but I've found a smart workaround :P Implemented in https://github.com/contao/contao/pull/3850/commits/d0c63da6771500eb134222ea95a259459c81f52f

Toflar:
We now have an even more flexible way of specifying the intervals and there's a `RetentionPolicyInterface` now which decides which backups should be kept so you can completely re-invent and implement your custom solutions 😎

core-bundle/src/DependencyInjection/Configuration.php
m-vo: Nice!

core-bundle/src/Doctrine/Backup/RetentionPolicy.php
```diff
@@ -0,0 +1,107 @@
...
+        // Then cleanup according to maximum amount of backups to keep
+        if ($keepMax > 0) {
+            $toKeep = \array_slice($toKeep, 0, $keepMax);
+        }
+
+        return $toKeep;
+    }
+
+    /**
+     * @throws \Exception
```
leofeyer: This type hint needs to be removed (see #3816).

leofeyer (changes requested):

core-bundle/src/Doctrine/Backup/RetentionPolicy.php
Toflar: I've just pushed https://github.com/contao/contao/pull/3850/commits/63928ec6b1615d6bef024d23bbecdf6915035965 to already validate things in the configuration. I left the comment because it's now `public static` and you are expected to handle it. Still wrong?

leofeyer: Yes, sorry. General exceptions such as `\Exception`, `\RuntimeException` or `\LogicException` are by definition always supposed to bubble up. If you want to add an exception that other developers are supposed to handle in a try/catch block, you should add (or use) a specific exception.

ausi (dismissed):

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -365,10 +365,18 @@ private function handleBackup(array $config, ContainerBuilder $container): void
...
+        $retentionPolicy->setArgument(0, $config['backup']['keep_max']);
+        $retentionPolicy->setArgument(1, $config['backup']['keep_intervals']);
+
         $dbDumper = $container->getDefinition('contao.doctrine.backup_manager');
-        $dbDumper->replaceArgument(2, $config['backup']['directory']);
-        $dbDumper->replaceArgument(3, $config['backup']['ignore_tables']);
-        $dbDumper->replaceArgument(4, $config['backup']['keep_max']);
+        $dbDumper->setArgument(2, $config['backup']['directory']);
+        $dbDumper->setArgument(3, $config['backup']['ignore_tables']);
+        $dbDumper->setArgument(4, $retentionPolicy);
```
leofeyer, line 379: The fourth argument is the retention policy already. Can this line be omitted?

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #3852, opened 2021-12-21
Dispatch an event before rendering Twig templates

m-vo:
This replaces the `Twig\Environment` with our own that dispatches an event before rendering. The intention is to provide an alternative to the `parseTemplate`, `parseFrontendTemplate`, `parseBackendTemplate`, `parseWidget`, `outputFrontendTemplate` and `outputBackendTemplate` hooks. 

I went for a single event that runs *before* rendering and that allows adjusting the context. If you need to adjust the output (which is way less sane), you can now add a `PostRenderTransformer` in the event. This way we can at least encapsulate dangerous things (like adjusting HTML) and there will be a better DX for that:

Example:
```php
public function __invoke(TwigRenderEvent $event): void
{
    // adjust the context
    $event->setContext(array_merge($event->getContext(), ['hello' => 'world']));

    // add `data-foo="bar" to every wrapper
    $event->addPostRenderTransformer(new HtmlWrapperTransformer(['data-foo' => 'bar']));
}
```

We could ship transformers for the most typical scenarios out of the box.

Feedback welcome before I go ahead, polish and add tests. :relaxed:


m-vo:
I have to rework this concept, I guess. If events should also be fired whenever a file is included/embedded, we'll need a more sophisticated solution that injects an event dispatching node in the AST when rendering (or something along this line).

PR #3857, opened 2021-12-22
Fix fragments in RSS feeds

### LINKED ISSUE #3856: Cannot use fragments for RSS feeds with "Full stories"
### Affected version(s)

4.9, 4.12, 4.x

### Description

If one of your news articles uses fragment content elements, the following error will occur when executing `contao:cron` and the RSS feed is set to "Full stories":

```
In FragmentHandler.php line 82:

  [LogicException]
  Rendering a fragment can only be done when handling a Request.
```

### Reproduction

1. Create a custom content element as fragment [1].
2. Create a news archive.
3. Create a news article within the archive.
4. Add the fragment content element to the detail content of the news article.
5. Create an RSS feed for the news archive.
6. Set **Export settings** to _Full stories_.
7. Execute `vendor/bin/contao-console contao:cron`.

[1]

```php
// src/Controller/ContentElement/HelloWorldElementController.php

use Contao\ContentModel;
use Contao\Template;

/**
 * @ContentElement(category="texts")
 */
class HelloWorldElementController extends AbstractContentElementController
{
    {
    }
}
```

### Stack trace

```
In FragmentHandler.php line 82:

  [LogicException]
  Rendering a fragment can only be done when handling a Request.  


Exception trace:
  at vendor\symfony\http-kernel\Fragment\FragmentHandler.php:82
 Symfony\Component\HttpKernel\Fragment\FragmentHandler->render() at vendor\contao\contao\core-bundle\src\Fragment\FragmentHandler.php:84
 Contao\CoreBundle\Fragment\FragmentHandler->render() at vendor\contao\contao\core-bundle\src\Resources\contao\elements\ContentProxy.php:61
 Contao\ContentProxy->generate() at vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Controller.php:617
 Contao\Controller::getContentElement() at vendor\contao\contao\news-bundle\src\Resources\contao\classes\News.php:211
 Contao\News->generateFiles() at vendor\contao\contao\news-bundle\src\Resources\contao\classes\News.php:79
 Contao\News->generateFeeds() at vendor\contao\contao\core-bundle\src\Cron\LegacyCron.php:91
 Contao\CoreBundle\Cron\LegacyCron->runLegacyCrons() at vendor\contao\contao\core-bundle\src\Cron\LegacyCron.php:52
 Contao\CoreBundle\Cron\LegacyCron->onDaily() at vendor\contao\contao\core-bundle\src\Cron\CronJob.php:60
 Contao\CoreBundle\Cron\CronJob->__invoke() at vendor\contao\contao\core-bundle\src\Cron\Cron.php:119
 Contao\CoreBundle\Cron\Cron->run() at vendor\contao\contao\core-bundle\src\Command\CronCommand.php:45
 Contao\CoreBundle\Command\CronCommand->execute() at vendor\symfony\console\Command\Command.php:255
 Symfony\Component\Console\Command\Command->run() at vendor\symfony\console\Application.php:1027
 Symfony\Component\Console\Application->doRunCommand() at vendor\symfony\framework-bundle\Console\Application.php:97
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at vendor\symfony\console\Application.php:273
 Symfony\Component\Console\Application->doRun() at vendor\symfony\framework-bundle\Console\Application.php:83
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at vendor\symfony\console\Application.php:149
 Symfony\Component\Console\Application->run() at vendor\contao\contao\manager-bundle\bin\contao-console:37
 include() at vendor\bin\contao-console:92
```

### Possible solution

`Contao\News::generateFiles` already creates a fake request context here:

https://github.com/contao/contao/blob/effd5447f8072607f7dffc94c33241d154971ee2/news-bundle/src/Resources/contao/classes/News.php#L205-L207

However this will only work for the legacy content elements etc. - but not for fragments. I think we need to create a fake `Request` instance and push that to the request stack.

fritzmg:
Fixes #3856

This PR now always pushes a new request to the request stack for each individual news article. I left the original `Environment::set('request', …)` bit for the legacy framework untouched though.

### DIFF
--- a/calendar-bundle/src/Resources/contao/classes/Calendar.php
+++ b/calendar-bundle/src/Resources/contao/classes/Calendar.php
@@ -209,6 +209,7 @@ class Calendar extends Frontend
 		ksort($this->arrEvents);
 
 		$container = System::getContainer();
+
 		/** @var RequestStack $requestStack */
 		$requestStack = System::getContainer()->get('request_stack');
 
--- a/calendar-bundle/src/Resources/contao/classes/Calendar.php
+++ b/calendar-bundle/src/Resources/contao/classes/Calendar.php
@@ -10,6 +10,9 @@
 
 namespace Contao;
 
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+
 /**
  * Provide methods regarding calendars.
  *
@@ -205,13 +208,9 @@ class Calendar extends Frontend
 		$count = 0;
 		ksort($this->arrEvents);
 
-		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
-
-		if ($request)
-		{
-			$origScope = $request->attributes->get('_scope');
-			$request->attributes->set('_scope', 'frontend');
-		}
+		$container = System::getContainer();
+		/** @var RequestStack $requestStack */
+		$requestStack = System::getContainer()->get('request_stack');
 
 		$origObjPage = $GLOBALS['objPage'] ?? null;
 
@@ -229,6 +228,11 @@ class Calendar extends Frontend
 
 					// Override the global page object (#2946)
 					$GLOBALS['objPage'] = $this->getPageWithDetails(CalendarModel::findByPk($event['pid'])->jumpTo);
+
+					// Push a new request to the request stack (#3856)
+					$request = Request::create($event['link']);
+					$request->attributes->set('_scope', 'frontend');
+					$requestStack->push($request);
 
 					$objItem = new FeedItem();
 					$objItem->title = $event['title'];
@@ -292,18 +296,15 @@ class Calendar extends Frontend
 					}
 
 					$objFeed->addItem($objItem);
+
+					$requestStack->pop();
 				}
 			}
 		}
 
-		if ($request)
-		{
-			$request->attributes->set('_scope', $origScope);
-		}
-
 		$GLOBALS['objPage'] = $origObjPage;
 
-		$webDir = StringUtil::stripRootDir(System::getContainer()->getParameter('contao.web_dir'));
+		$webDir = StringUtil::stripRootDir($container->getParameter('contao.web_dir'));
 
 		// Create the file
 		File::putContent($webDir . '/share/' . $strFile . '.xml', $this->replaceInsertTags($objFeed->$strType(), false));
--- a/news-bundle/src/Resources/contao/classes/News.php
+++ b/news-bundle/src/Resources/contao/classes/News.php
@@ -10,6 +10,9 @@
 
 namespace Contao;
 
+use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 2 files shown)


Toflar (dismissed):

aschempp (changes requested):
That's exactly why the RSS feed should be its own controller that actually handles a request. Since I don't think this can be changed in 4.9, we should only create one fake request per RSS feed file (with the path of that file) to somehow be forward-compatible?

fritzmg:
> That's exactly why the RSS feed should be its own controller that actually handles a request.

Of course, but that would be a new feature. This PR is the only way to fix this in 4.9 I think.

> I don't think this can be changed in 4.9, we should only create one fake request per RSS feed file (with the path of that file) to somehow be forward-compatible?

Currently we create a fake request per feed item, with the feed item's URL (but only for the legacy `Environment`). I don't think we should change that in a bugfix.

aschempp:
> Currently we create a fake request per feed item, with the feed item's URL (but only for the legacy `Environment`). I don't think we should change that in a bugfix.

Can you point to where we do that? The only thing I can see in your PR is that we manipulate the scope, but not create a request.

fritzmg:
See the [original issue](https://github.com/contao/contao/issues/3856), I have referenced the code there.

https://github.com/contao/contao/blob/effd5447f8072607f7dffc94c33241d154971ee2/news-bundle/src/Resources/contao/classes/News.php#L205-L207

aschempp:
🤦  The original issue (https://github.com/contao/core/issues/7756) would also be solved by a front end controller 🙊

fritzmg:
Yes, but as I said, that's a new feature for Contao 5.0.

This PR fixes fragments for RSS feeds in Contao 4.9.

aschempp (commented):
I'm unsure whether we should create one request for each article or one request for the RSS file. One per article is similar to the current behaviour, but the current behaviour is wrong and would change once we have a real front end controller. Since both solutions would solve the initial problem (not having a request at all) I feel like we should "do it right"® already here 🤔

fritzmg:
But the content of a news article depends on a correct request. If you have a download content element in your news article, the download link will be wrong, if the request does not contain the correct information. Even if we generate the RSS feed via a controller in the future, we need to accommodate that.

aschempp:
hmm, again not a problem the Controller would have, since the query parameter would re-render the content and actually output the file. But I agree that currently wouldn't work and we need that fix 😢

aschempp (dismissed):

ausi (dismissed):

leofeyer:
What about RSS feeds in the calendar module? Do they not have the same problem?

fritzmg:
True, I will adjust the PR accordingly.

aschempp (dismissed):

leofeyer:
Thank you @fritzmg.

PR #3862, opened 2021-12-22
Add an "overview page" field

### LINKED ISSUE #3586: Remove javascript:history URLs
From https://github.com/contao/contao/pull/3424#issuecomment-931334841

> We also want to remove all usages of `javascript:history.go(-1)` and other back links from templates (e.g. reader modules).

I think this change should go into 4.13 as it is not a bug.

leofeyer:
Fixes #3586

There is still one occurrence of `javascript:history.go(-1)` left:

https://github.com/contao/contao/blob/2c151c6a23650fd2e993cd49833eb148f0dabc19/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L191-L192

Unfortunately, we cannot set an overview page in the module configuration, as the article module is a "virtual module". And we cannot just strip the `/articles/alias` part from the URL, because https://github.com/contao/core/issues/6955.

@contao/developers Any idea how to fix this?


listing-bundle/src/Resources/contao/modules/ModuleListing.php
```diff
@@ -359,9 +359,12 @@ protected function compile()
 	 */
 	protected function listSingleRecord($id)
 	{
+		/** @var PageModel $objPage */
+		global $objPage;
+
 		$this->Template = new FrontendTemplate($this->list_info_layout ?: 'info_default');
 		$this->Template->record = array();
-		$this->Template->referer = 'javascript:history.go(-1)';
+		$this->Template->referer = $objPage->getFrontendUrl();
```
leofeyer: Not sure about this change as there could be other GET parameters besides `/show/42` that need to be preserved (e.g. pagination). @contao/developers Any idea how to improve this?

aschempp:
Shouldn't the article know which page it is coming from and be able to render that page URL (if routable)?

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -87,8 +87,18 @@ protected function compile()
...
-		$this->Template->back = $GLOBALS['TL_LANG']['MSC']['goBack'];
+
+		if ($this->overviewPage)
+		{
+			$this->Template->referer = PageModel::findById($this->overviewPage)->getFrontendUrl();
+			$this->Template->back = $GLOBALS['TL_LANG']['MSC']['eventOverview'];
+		}
+		else
+		{
+			// Backwards compatibility
```
aschempp: should we add a deprecation warning that this back link will be removed in Contao 5 if no overview link is set?

calendar-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -65,6 +65,9 @@
       <trans-unit id="MSC.eventPicker">
         <source>Events</source>
       </trans-unit>
+      <trans-unit id="MSC.eventOverview">
+        <source>Back to the event overview</source>
```
aschempp, line 69: I don't think this label makes sense. Whenever I have received a design from an agency that does not design for Contao, they often have such links and all simply are "Back". Since we don't know if the back link points to "an overview", I think we should keep calling it "back" and anyone with more specific need can still change the template.

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -289,6 +289,15 @@
 			'sql'                     => "int(10) unsigned NOT NULL default 0",
 			'relation'                => array('type'=>'hasOne', 'load'=>'lazy')
 		),
+		'overviewPage' => array
```
aschempp, line 292: maybe call this `jumpToOverview` or `jumpBack` to follow the existing `jumpTo` field?

ausi:
> Shouldn't the article know which page it is coming from and be able to render that page URL (if routable)?

Think so too. A link to the URL of `$this->pid` should suffice IMO.

leofeyer:
No, it does not. See contao/core#6955.

ausi:
I’m confused by contao/core#6955. If I use an article teaser from a different page and I want a back link that goes somwhere else, I have to add a link content element after the included article teaser IMO. This would then also work if the user arrived via a Google search or the like.

calendar-bundle/src/Resources/contao/languages/en/default.xlf
leofeyer: > Since we don't know if the back link points to "an overview",

I disagree. We know exactly that it is a "back to the overview" link, because that is the main point of the PR. 😉

aschempp: The target module is called "event list". How do you know this is an "overview"? It could be the archive as well? I also know a lot of users that "abuse" the news module to list anything (e.g. projects of an architecture company). I sure don't want that link to be "back to news overview" if they are not news. "Go back" simply works for any and all cases 🤷

fritzmg: I agree. It might also be called a "blog" for example.

leofeyer: I hope you do realize that `MSC.eventOverview` is not a generic label? It is a very specific label that is only used in the event module where we exactly know that we are going back to the event overview.

aschempp: I know. Would you like me to write the same on the `newsOverview` label, just because our examples were about the news module? 😂

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
leofeyer: Added in 198e15bed06b31dcfb8abc466c75fb1307a9c7d9.

leofeyer:
I have adjusted `ModuleArticle` in 9538545be75b567d0da88f4b05b1ccbf6ee093a1.

listing-bundle/src/Resources/contao/modules/ModuleListing.php
fritzmg: That's the case with news and events as well. Your news list could be paginated and you might have been on page 3 when you clicked on a link to a news detail page. The back link would then show the news list on page 1 again. I don't think this can be solved without using the _actual_ URL the visitor came from.

We could check the referrer URL and see if it matches the overview URL and then use the whole referrer URL instead or something?

ausi: > We could check the referrer URL and see if it matches the overview URL and then use the whole referrer URL instead or something?

I’m not a a fan of that, IMO the link should always point to the same URL, no matter from where you came. (Also important to make it cachable)

ausi (dismissed):

listing-bundle/src/Resources/contao/modules/ModuleListing.php
leofeyer: As briefly discussed with @ausi, we will leave the `javascript:history.go(-1)` here for the time being. We can still fix this in Contao 5 by either storing the filter settings in the session storage of the browser or by reworking the module entirely.

fritzmg: Why do we need this special handling just for the listing module? What's the difference to a paginated news list for example?

ausi (approved):

listing-bundle/src/Resources/contao/modules/ModuleListing.php
leofeyer: The news module takes you back to the canonical URL, which is fine. The listing module, however, has a search field and a "Results per page" dropdown, which should not be reset when going back.

calendar-bundle/src/Resources/contao/languages/en/default.xlf
ausi: Personally I always find “Back” links hard to use on many websites because they often don’t tell me “Back to where exactly?”. I often look at which URL it points to, or use links from a breadcrumb navigation instead.

For me, “Back to the event overview” is way better than “Back”. And in cases where it doesn’t match with the terminology of the project, it gets changed to something like “Back to our Blog” or so, which is also way more useful than just “Back”.

So for me, tha fact that it has to be changed in some projects is a benefit (for the end user).

leofeyer: > I also know a lot of users that "abuse" the news module to list anything (e.g. projects of an architecture company). I sure don't want that link to be "back to news overview" if they are not news.

Well, in this case it is their responsibility to also adjust the label, isn't it?

listing-bundle/src/Resources/contao/modules/ModuleListing.php
fritzmg: > which should not be reset when going back.

Why not? Why is it okay to remove the query parameter for a paginated news list, but not for a filtered listing module? I do not see the difference.

PR #3865, opened 2021-12-23
Use generic image format labels

leofeyer:
We could save our translators some tedious work this way.


bytehead (approved):
I like.

aschempp (approved):
assuming they don't translate PNG to Japanese or so? @taca

Toflar (approved):

leofeyer:
> assuming they don't translate PNG to Japanese or so? @taca

We could make the labels even more generic and use `.png → .webp` then, which should be valid in all languages.

taca:
> > assuming they don't translate PNG to Japanese or so? @taca
> 
> We could make the labels even more generic and use `.png → .webp` then, which should be valid in all languages.

I did not understand what's going on.

Anyway, these name of file format would not be translated, so no problem.

fritzmg (approved):

aschempp:
Thank you @taca, your answer was exactly what I was looking for 👍

PR #3866, opened 2021-12-23
Fix some minor issues

leofeyer:
… such as typos, missing phpDoc comments, leftover Symfony 4.4 workarounds and the like.

### DIFF
--- a/core-bundle/src/Controller/InsertTagsController.php
+++ b/core-bundle/src/Controller/InsertTagsController.php
@@ -12,8 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Controller;
 
-use Contao\CoreBundle\Framework\ContaoFramework;
-use Contao\InsertTags;
+use Contao\CoreBundle\InsertTag\InsertTagParser;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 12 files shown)


core-bundle/src/Controller/InsertTagsController.php
```diff
@@ -26,20 +25,16 @@
...
     }
 
     public function renderAction(Request $request, string $insertTag): Response
     {
-        $this->framework->initialize();
-
-        $it = $this->framework->createInstance(InsertTags::class);
-
-        $response = new Response($it->replace($insertTag, false));
+        $response = new Response($this->insertTagParser->replaceInline($insertTag));
```
leofeyer, line 37: @ausi Just double checking: If `$blnCache` is `false`, we have to use the `replaceInline()` method, right?

ausi: ✅

ausi (dismissed):

installation-bundle/src/Database/Installer.php
```diff
@@ -110,7 +111,7 @@ public function compileCommands(): void
         $fromSchema = $this->connection->createSchemaManager()->createSchema();
         $toSchema = $this->schemaProvider->createSchema();
 
-        $diff = $fromSchema->getMigrateToSql($toSchema, $this->connection->getDatabasePlatform());
+        $diff = (new Comparator())->compareSchemas($fromSchema, $toSchema)->toSql($this->connection->getDatabasePlatform());
```
Toflar: This should take `createComparator()` on the platform if DBAL 3.x in order to fix the wrong lenght comparisons.

Toflar (changes requested):

installation-bundle/src/Database/Installer.php
leofeyer: Changed in 39895e4c1ad86ceedc986e8b559c0d9a186de700.

Toflar (dismissed):
Thanks!

PR #3867, opened 2021-12-26
Serialize array value before saving

bytehead:
It actually should return the serialized array:

https://github.com/contao/contao/blob/effd5447f8072607f7dffc94c33241d154971ee2/core-bundle/src/Resources/contao/dca/tl_layout.php#L565-L572

Bildschirmfoto 2021-12-26 um 14 30 42

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_layout.php
+++ b/core-bundle/src/Resources/contao/dca/tl_layout.php
@@ -588,6 +588,6 @@ class tl_layout extends Contao\Backend
 			array_insert($array, $i, 'layout.css');
 		}
 
-		return $array;
+		return serialize($array);
 	}
 }


fritzmg (approved):

leofeyer:
Thank you @bytehead.

aschempp:
This looks suspicious to me, I'm pretty sure this worked before. Our database connection automatically serializes array values (see https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php#L374-L377). @ausi could this be related to switching to DBAL in our database class?

ausi:
> I'm pretty sure this worked before.

And it still works I think if you silence the warnings.

The error level for “Array to string conversion” got raised from notice to warning in PHP 8.0

bytehead:
> I'm pretty sure this worked before

Sure it worked but it was wrong all the time 🙃

aschempp:
Well to me that was a feature. And not correctly saving array values certainly is a BC break to me.

> And it still works I think if you silence the warnings.
> 
> The error level for “Array to string conversion” got raised from notice to warning in PHP 8.0

Pretty sure thats not true. This warning means we're converting an array to string, not an array to serialized data.

ausi:
I think the warning gets triggered on this line: https://github.com/contao/contao/blob/3ce5a595d30a4cb26361f03f368f9bf38faa560e/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3105
by the `(string) $varValue !== ''` check, see https://3v4l.org/hK9Pl

> And not correctly saving array values certainly is a BC break to me.

Nothing has changed there I think. But the `save_callback` has to return the same data type as it retrieves because another `save_callback` that gets called later expects that same format. So this pull request is correct IMO.

bytehead:
https://github.com/contao/contao/blob/3ffe7e183bca7460d1f619d669d3a64d3e5a969a/core-bundle/src/Resources/contao/dca/tl_layout.php#L581-L584

Could be theoretically still wrong in case of an empty array. @ausi should we fix that or ignore?

aschempp:
> Nothing has changed there I think. But the `save_callback` has to return the same data type as it retrieves because another `save_callback` that gets called later expects that same format. So this pull request is correct IMO.

Thanks for debugging. I always saw it as a feature that load and save callbacks don't need to serialize and unserialise each. Simply because `deserialize` on an array does nothing, so this works both ways without issues.

This change is absolutely ok. The warning should be fixed by https://github.com/contao/contao/pull/3876

PR #3869, opened 2021-12-26
Add contao:mailer:send command

fritzmg:
A `mailer:send` command will be added to Symfony only in 6.1 (or higher), if at all (see https://github.com/symfony/symfony/pull/43687). In order to also have a console command for sending emails for our last `4.x` LTS, we should integrate the command into the `contao/core-bundle`. Later on we can deprecate and forward it to the Symfony command again.

The command largely works the same way as the `swiftmailer:email:send` command previously.

### DIFF
--- a/core-bundle/src/Command/SendMailCommand.php
+++ b/core-bundle/src/Command/SendMailCommand.php
@@ -1,0 +1,130 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Mailer\MailerInterface;
+use Symfony\Component\Mime\Email;
+
+/**
+ * A console command to send an email via Symfony Mailer.
+ */
+final class SendMailCommand extends Command
+{
+    protected static $defaultName = 'contao:sendmail';
+
+    private SymfonyStyle $io;
+    private MailerInterface $mailer;
+
+    public function __construct(MailerInterface $mailer)
+    {
+        parent::__construct();
+
+        $this->mailer = $mailer;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Sends an email message')
+            ->addOption('from', null, InputOption::VALUE_REQUIRED, 'The sender of the message')
+            ->addOption('to', null, InputOption::VALUE_REQUIRED, 'The recipient of the message')
+            ->addOption('subject', null, InputOption::VALUE_REQUIRED, 'The subject of the message')
+            ->addOption('body', null, InputOption::VALUE_REQUIRED, 'The body of the message')
+            ->addOption('transport', null, InputOption::VALUE_OPTIONAL, 'The transport to be used')
+            ->addOption('content-type', null, InputOption::VALUE_REQUIRED, 'The body content type of the message', 'text/plain')
+            ->addOption('charset', null, InputOption::VALUE_REQUIRED, 'The body charset of the message', 'utf-8')
+            ->addOption('body-source', null, InputOption::VALUE_REQUIRED, 'The source of the body [stdin|file]', 'stdin')
+            ->setHelp(
+                <<<'EOF'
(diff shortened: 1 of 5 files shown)


Toflar (commented):
👍 on merging this, however I would rename the command to `contao:mailer:send`. It allows for possible new commands in the future and does not imply any connection to the `sendmail` binary.

core-bundle/src/Command/SendMailCommand.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        parent::__construct();
+
+        $this->mailer = $mailer;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Sends an email message')
```
94noni: you can leverage https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description

Toflar (approved):

bytehead (approved):

leofeyer (changes requested):
I would like to wait and see if Symfony really merges the PR. Depending on that, we should consider providing an extension that backports the functionality to Contao.

aschempp:
> I would like to wait and see if Symfony really merges the PR. Depending on that, we should consider providing an extension that backports the functionality to Contao.

I agree with this. We should add a command with the same name and characteristic (if possible). Which makes me wonder if this should not be in Contao itself but a bundle that can be included if you want to command-line-send in Contao 4.13 🤷

fritzmg:
> I would like to wait and see if Symfony really merges the PR.

Symfony moved it to the `6.1` milestone. We could in theory allow the installation of `symfony/mailer: ^5.4 || ^6.1` though.

fritzmg:
> Depending on that, we should consider providing an extension that backports the functionality to Contao.

I agree too though. I wanted to create a back-port extension anyway, if this didn't make it into Symfony or Contao 4.13 🙃

m-vo (approved):
Very nice. :heart: 

Could we maybe add something like a `--test` option that takes the recipient and otherwise uses some opinionated defaults (or even uses the system admin email default for the recipient as well).

 I often find myself needing this kind of quick way to test a SMTP config and could not care less about the actual contents. I just want to verify mails can be send. :smile:

aschempp:
> Could we maybe add something like a `--test` option that takes the recipient and otherwise uses some opinionated defaults (or even uses the system admin email default for the recipient as well).
> 
> I often find myself needing this kind of quick way to test a SMTP config and could not care less about the actual contents. I just want to verify mails can be send. 😄

I understand that requirement, but I think it contradicts with the fact that we want to drop this again as soon as it is supported in Symfony?

m-vo:
> as soon as it is supported in Symfony

which could be anywhere between "in quite a while" and "never" :smile:

ausi (approved):

aschempp:
As discussed in Slack (and agreed to by most core devs), this might as well be an external bundle. The advantages would be that it can be installed in versions prior to 4.13, it could be kept up-to-date with the latest Symfony PR features (without being related to a Contao version) and we could still require it in a Contao Managed Edition by default (until if Symfony ships this feature).

fritzmg:
Do we want it to be part of `contao` (not the mono repo though)?

aschempp:
> Do we want it to be part of `contao` (not the mono repo though)?

I don't see a reason to. It could be @fritzmg same as we have bundles from @terminal42 and @ausi. Being in the `contao` organisation has no benefits except enforcing it to be maintained by the core team, possibly even after we drop it from Contao.

fenepedia:
@fritzmg Will this feature be shipped with Contao 5?

fritzmg:
@fenepedia see the discussion above - we decided to not ship it with Contao. We might create a `inspiredminds/symfony-mailer-send-command-bundle`.

leofeyer:
Maybe we should close the PR then?

fritzmg:
You can now install [`inspiredminds/contao-mailer-command`](https://github.com/inspiredminds/contao-mailer-command) in the Contao Managed Edition or [`inspiredminds/symfony-mailer-command-bundle`](https://github.com/inspiredminds/symfony-mailer-command-bundle) for standard Symfony projects in order to get a basic `mailer:send` command.

aschempp:
Should we make [inspiredminds/contao-mailer-command](https://github.com/inspiredminds/contao-mailer-command) a default in Contao (5)? Would also allow to implement https://github.com/contao/contao-manager/issues/254

leofeyer:
We can use `mailer:send` from Symfony 6.1 in Contao 5.

Toflar:
Where did you read that Symfony 6.1 provides this command? @fritzmg's PR https://github.com/symfony/symfony/pull/43687 is still open.

leofeyer:
You are right, the PR is now targeted on Symfony 6.2.

PR #3871, opened 2021-12-28
Rename the token value method

aschempp:
The `getFrontendTokenValue` method was added in #3620

I however noticed the token is not only used in the front end but also the back end, therefore I suggest to rename the method.

### DIFF
--- a/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
+++ b/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
@@ -66,7 +66,7 @@ class ContaoCsrfTokenManager extends CsrfTokenManager
     public function getDefaultTokenValue(): string
     {
         if (null === $this->defaultTokenName) {
-            throw new \RuntimeException('The Contao CSRF token manager was not initialized with a frontend token name.');
+            throw new \RuntimeException('The Contao CSRF token manager was not initialized with a default token name.');
         }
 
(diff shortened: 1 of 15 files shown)


Toflar (dismissed):

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
```diff
@@ -63,12 +63,12 @@ public function canSkipTokenValidation(Request $request, string $tokenCookieName
         ;
     }
 
-    public function getFrontendTokenValue(): string
+    public function getDefaultTokenValue(): string
     {
-        if (null === $this->frontendTokenName) {
+        if (null === $this->defaultTokenName) {
             throw new \RuntimeException('The Contao CSRF token manager was not initialized with a frontend token name.');
```
bytehead: We should adjust the exception message as well.

bytehead (changes requested):

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
aschempp: good catch! fixed in 8a08cd587f6b592ee9aa95c76ec3618e36919ae8

bytehead (approved):

fritzmg (approved):

PR #3872, opened 2021-12-28
Add front end preview links

aschempp:
This PR replaces the maintenance login feature _not_ introduced in #3729. As mentioned by @fritzmg in https://github.com/contao/contao/pull/3729#issuecomment-999468916, a regular use-case is not only to bypass the maintenance mode, but also to show unpublished pages to users that do not have a back end account (e.g. management to approve a new website or page).

This PR introduces _frontend preview links_. It work using a signed (= secure) link that logs in the front end user with our existing `FrontendPreviewToken`, which then allows to see pages in maintenance and optionally unpublished elements, even if not logged in to the back end.

I have discussed the feature with @Toflar and @netzarbeiter and this is the best approach we could think of. For a user, you simply navigate to the desired page in the front end, and click the _Share URL_ button in the preview bar. This opens the back end to share that page. 

We initially discussed a checkbox to show or hide the preview bar in the front end as well. I however noticed that the preview bar only ever works if a backend user is logged in. Therefore this additional check/option is unnecessary, the preview bar is simply not available for regular visitors through preview links. If they are for any reason logged in to the back end, the preview bar will appear as usual.

Here's a quick demo:

https://user-images.githubusercontent.com/1073273/147571764-87962ace-f0f6-4b08-b928-17268359a593.mov

### Security implications:
As you might notice in the PR, there are a few places where we currently check for `isGranted('ROLE_USER')` **in addition to** checking for `FrontendPreviewToken`. This is (a) inconsistent because we don't always to that and (b) obviously wouldn't work with this PR.
 - the `FrontendPreviewToken` is currently only ever set in the preview bar controller, and that controller double-checks on `ROLE_USER` permissions.
 - theoretically, a user could currently "loose" its backend firewall token and would then not have frontend preview either. I don't really see this ever happening though, even a logout of the back end invalidates the session and logs out any front end users.

### TODO:
- [x] Move DCA callbacks to event listener
- [x] Implement sharing link (copy to clipboard) in the DCA list view
- [x] Update the `share.svg` icon to reflect adding to clipboard
- [x] Unit tests


core-bundle/src/Controller/FrontendPreviewController.php
```diff
@@ -0,0 +1,69 @@
...
+
+        if (false === $link) {
+            throw new NotFoundHttpException('Preview link not found.');
+        }
+
+        if ($link['showUnpublished']) {
+            $this->previewAuthenticator->authenticateFrontendGuest((bool) $link['showUnpublished']);
+        } else {
+            $this->previewAuthenticator->removeFrontendAuthentication();
+        }
```
ausi: I’m confused by these lines. If `showUnpublished` is `false` wouldn’t calling `removeFrontendAuthentication()` make the URL not viewable by the current session?

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -185,7 +185,7 @@ private function setConstants(): void
         if ('FE' === $this->getMode() && ($session = $this->getSession()) && $this->request->hasPreviousSession()) {
             $session->start();
 
-            \define('BE_USER_LOGGED_IN', $this->tokenChecker->hasBackendUser() && $this->tokenChecker->isPreviewMode());
+            \define('BE_USER_LOGGED_IN', $this->tokenChecker->isPreviewMode());
```
ausi, line 190: I think we should deprecate `BE_USER_LOGGED_IN` (`FE_USER_LOGGED_IN` is already deprecated). And not change this check here as the name BE_USER_LOGGED_IN suggests that there has to be a logged in backend user.

The usages of `BE_USER_LOGGED_IN` in the core have to be changed then to use the token checker.

aschempp: sounds good, do you think it should be part of this PR? Also, how do we deprecated a constant?

core-bundle/src/Controller/FrontendPreviewController.php
aschempp: you're absolutely right, good catch 😂 
fixed in 91d4c5a7a6f8635f2e5fb24e4f058e4c02080507

core-bundle/src/Framework/ContaoFramework.php
aschempp: see https://github.com/contao/contao/pull/3873

ausi, line 190: IMO this is not resolved yet because this change hast to be reverted in this pull request.

```suggestion
            \define('BE_USER_LOGGED_IN', $this->tokenChecker->hasBackendUser() && $this->tokenChecker->isPreviewMode());
```

aschempp: jup, this is just a draft atm 🙃

aschempp: I reviewed this again, and I think this change needs to stay. Otherwise all legacy extensions relying on `BE_USER_LOGGED_IN` will incorrectly not show stuff in the new preview link mode. Again, the constant was never meant to represent whether a back end user was actually logged in … 🤷

ausi: The preview link mode is new (also doesn’t show a preview toolbar for example), so I think it is OK if it behaves different. For security reasons and to get extension developers to use the new method I would revert this.

aschempp: It makes the preview links likely unusable in my opinion …

core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
```diff
@@ -9,6 +9,9 @@
             <input type="hidden" name="REQUEST_TOKEN" value="{{ request_token }}">
             <div>
                 <button type="button" id="copyPublishedPath" class="tl_submit">{{ 'MSC.copyURL'|trans([], 'contao_default') }}</button>
+                {% if share is not empty %}
+                <a href="{{ share }}" target="_blank" class="tl_submit" onclick="window.open(this.href+'&url='+location.href); return false">{{ 'MSC.shareURL'|trans([], 'contao_default') }}</a>
```
Toflar:
```suggestion
                <a href="{{ share }}" target="_blank" class="tl_submit" onclick="window.open(this.href+'&amp;url='+location.href); return false">{{ 'MSC.shareURL'|trans([], 'contao_default') }}</a>
```

core-bundle/src/Framework/ContaoFramework.php
fritzmg: Yes, without this change the preview URL will not work with any extension that still uses `BE_USER_LOGGED_IN`. Imho it is fine to change this because there is no actual behaviour change anyway, is there? Except for the new feature.

ausi: > there is no actual behaviour change anyway

Technically there is IMO. Because now `BE_USER_LOGGED_IN` in no longer secured to a backend login. For example if you force 2FA, the preview link still works without 2FA.

But I agree, this “issue” is most probably only a theoretical possibility. So I think we can keep it as it is currently in this pull request 👍

@fritzmg can/should we document this? That A) `BE_USER_LOGGED_IN` in now deprecated and B) can also be true for the new preview links.

fritzmg: Yes https://github.com/contao/docs/issues/923 :)

core-bundle/src/Resources/views/Frontend/preview_toolbar_base.html.twig
leofeyer: Are you sure that the ampersand has to be encoded in the `onclick` attribute, too?

aschempp: pretty sure not. This is JS code, not HTML. `this.href` is already decoded when retrieving it from the DOM. But I didn't test it manually to be fair.

ausi: > Are you sure that the ampersand has to be encoded in the `onclick` attribute, too?

Yes it has to be encoded.

leofeyer:
I like this PR a lot. There are a few open questions though:

1. What about bypassing the maintenance mode? You are talking about a "maintenance login feature" in your initial comment, however, I don't see such a feature in the code.

2. Do we really have to show a "The preview link was copied to your clipboard" JavaScript alert? I find it quite inconvenient getting the alert and having to close it before I can continue with my workflow. Github does not show any kind of confirmation when copying commit hashes in the history view, so I'm asking myself if we really have to?

3. The edit screen has a lot of disabled fields. Are they all really necessary? For example, why does the editor see who created the preview URL if they cannot change the creator?

fritzmg:
> 2\. Do we really have to show a "The preview link was copied to your clipboard" JavaScript alert? I find it quite inconvenient getting the alert and having to close it before I can continue with my workflow. Github does not show any kind of confirmation when copying commit hashes in the history view, so I'm asking myself if we really have to?

In most places where you can use something like that a JavaScript Tooltip is used I think. I'd prefer that as well over an `alert()`. I'd also prefer nothing over the `alert()`.

Toflar:
Maybe we can introduce a general JS "flash message". Some box message that appears and then disappears again automatically. But styling and location etc. should probably go into a separate PR in order not to block this one. I mean, I'd rather have this feature with a simple confirm message than no preview feature at all just because of the JS hint :)

aschempp:
> I like this PR a lot. There are a few open questions though:
> 
> 1. What about bypassing the maintenance mode? You are talking about a "maintenance login feature" in your initial comment, however, I don't see such a feature in the code.

The maintenance mode is automatically (already) bypassed in the preview endpoint. You currently just couldn't use preview.php without a back end login.

> 2. Do we really have to show a "The preview link was copied to your clipboard" JavaScript alert? I find it quite inconvenient getting the alert and having to close it before I can continue with my workflow. Github does not show any kind of confirmation when copying commit hashes in the history view, so I'm asking myself if we really have to?

I think we need some sort of confirmation, but yes, a notification would be more convenient.

> 3. The edit screen has a lot of disabled fields. Are they all really necessary? For example, why does the editor see who created the preview URL if they cannot change the creator?

Good question. Yes the `createdBy` can probably be skipped. But the other ones provide relevant information to me, or not?

leofeyer:
> Good question. Yes the `createdBy` can probably be skipped. But the other ones provide relevant information to me, or not?

The "valid until" field is also kind of redundant, because "created at" + "expires after" = "valid until".

But I am not saying that we must not store these fields. I am only questioning if we must show them in the edit screen if they cannot be edited anyway. Maybe they should only be shown in the "show" view?

aschempp:
_Valid until_ is relevant information to me, and reading a date is easier than calculating it. But yeah, ultimately its all available in the _Show_ dialog 🤷

leofeyer:
The JavaScript alert mentioned above is still blocking this PR from getting merged. Shall we simply remove it for the time being or do you want to replace it with a better implementation?

aschempp:
I would not remove it because we require feedback. I would keep it and push a bugfix to change it later 🤷

leofeyer:
We already have the feedback that the JS alert is too obtrusive:



I am going to remove the JS alert then and you add a better solution in a follow-up PR.

leofeyer (dismissed):

leofeyer (approved):

PR #3876, opened 2021-12-29
Correctly handle array values in DCs

aschempp:
see https://github.com/contao/contao/pull/3867

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2317,7 +2317,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			}
 
 			// Make sure unique fields are unique
-			if ((string) $varValue !== '' && $arrData['eval']['unique'] && !$this->Database->isUniqueValue($this->strTable, $this->strField, $varValue, $this->objActiveRecord->id))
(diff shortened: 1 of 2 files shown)


bytehead (dismissed):

leofeyer (changes requested):
I'm not sure about this. We explicitly chose `(string) $varValue === ''` so `0` is handled correctly in PHP 8. This PR changes the behavior (see https://3v4l.org/E3WXe).

aschempp:
Ok, but what's the point of converting the value at all then? `0 === ''` will be `false` the same as `(string) 0 === ''`. The only difference I could make out is that `(string) false === ''` is empty/true but `false === ''` is not. But `false` is not really a valid database value, or is it?

Also, shouldn't `0` be considered empty for an integer field anyway?

fritzmg:
> But false is not really a valid database value, or is it?

It is, as it is equivalent to `0` for tinyint fields for example.

> Also, shouldn't 0 be considered empty for an integer field anyway?

Only if the field does not allow `null`.

aschempp:
I've update the implementation to check if value is scalar first. The `empty()` check was wrong anyway since it also meant an empty array was empty (see https://3v4l.org/n30EQ)

leofeyer (commented):
I think the new code works as expected. Let's wait for @ausi's review though.

ausi (approved):

PR #3873, opened 2021-12-28
Stop using BE_USER_LOGGED_IN constant

aschempp:
see https://github.com/contao/contao/pull/3872#discussion_r775934853

PR #3877, opened 2021-12-29
Enforce final newline in HTML5 templates

aschempp:
see https://github.com/contao/contao/pull/3873#discussion_r776347491


bytehead (approved):

bytehead:
Do we need to check every template now for this change?

aschempp:
> Do we need to check every template now for this change?

I quickly checked a few using regex and they seem to be ok. No idea if the CS fixer could check for something like that.

PR #3878, opened 2021-12-29
Enhance the service ID validation

leofeyer:
Validates sorting and warns if an FQCN service ID is used for something other than a controller.


ausi (approved):
❤️

bytehead (approved):

Toflar (approved):

Issue #3687, opened 2021-11-17
Proposal: add a DcaReader to access DCA data in a structured way


SeverinGloeckle:
I would like to discuss a possibility to standardize and ease the reading of configuration data from a DCA.

While the definition of DCAs via `$GLOBALS['TL_DCA']` cannot be changed at the moment, it would be nice if one could read values from the global array in a more structured way.

## Feature proposal: `DcaReader`
I'd suggest something like a `DcaReader`, which provides a method to get the value of any node within a DCA using dot notation:

````php
$mode = $dcaReader->get('list.sorting.mode');
```` 

**Caveats of dot notation**
While the dot notation is very convenient, there can be problems with nodes that contain a dot in their key. This could be handled explicitly in the code (and should probably be deprecated). Symfony's [PropertyAcccess](https://symfony.com/doc/current/components/property_access.html) therefore uses an array-like index notation, but in my opinion a dot notation is way easier to write and read.

**Utility methods**
Additionally, various utility methods could provide ways to query common data from the DCA:
- `field($field)` to get a field
- `field($field, 'key')` to get a subset of a field
- `isRelatedField($field)` to check whether a given field has a relation
- `isClosed()` for `config.closed`
- `getOperations` for `list.operations`
- ...

## Schema assets for DCAs
Even better than utility methods would be a definition of various subset schemas for a DCA, similar to Doctrine's [Schema Assets](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#schema-assets). 
As a quick thought, the following types come to mind:
- `DCA\Schema\Config`
- `DCA\Schema\CallbackCollection`
- `DCA\Schema\Callback`
- `DCA\Schema\Sorting`
- `DCA\Schema\Label`
- `DCA\Schema\GlobalOperation`
- `DCA\Schema\Operation`
- `DCA\Schema\Palette`
- `DCA\Schema\Field`
- `DCA\Schema\SQL` (which probably would extend Doctrine's `Column` asset)

The various assets then can provide dedicated utility methods.

**Custom DCA fields**
Since the various schema classes can only handle standard DCA fields, custom DCA fields would have to be accessed via some `get()` method that reads from the data source.

## Exemplary pseudo-code
An implementation could be used similarly to the following:

````php
// Get a reader from an injected factory or from a service locator / collection
$dcaReader = $this->dcaReaderCollection->get('tl_page');
````

````php
// Current:
[...]

PR #3879, opened 2021-12-29
Add DCA schemas

SeverinGloeckle:
Implements #3687 

The current draft is a working concept for an object-oriented and standardised access of DCA data.

The basic approach is as follows:
- A new service `DcaFactory` allows the creation of a Dca schema for a given resource (e.g. `tl_content`).
- Within the factory, a suitable driver for the resource is determined from a collection of appropriately tagged services.
	- Initially there is only one `GlobalArrayDriver`.
- Based on the data provided by the driver and a config-tree from the `DcaConfiguration`, an array with the configuration of the resource is created.
	- The `DcaConfiguration` is based on Symfony's configuration bundle and its config tree.
- The array with the configuration data is passed to a `Dca` schema.
- This schema allows querying the individual data nodes of the configuration via a fluid interface.

````php
$dca = $this->dcaFactory->get('tl_content');
$dca->list()->operations()->operation('edit')->get('href');
````

Note: When accessing the `Dca`, the individual nodes are created (and cached) dynamically at runtime. This way only the schemas of accessed nodes are initialized.

## Customization

For developers there are possibilities for customization:
- custom DCA sources can be queried via custom drivers
- configuration data can be added via `ConfigurationSourceEvent`
- schemas can be adjusted via `SchemaCreatedEvent`

## Mutable `Data`:
Internally, all schemas are based on a new `Data` class which wraps the original configuration array. Depending on the driver, the `Data` can be configured with observers so that all schemas are able to react to changes in the source data at runtime:

````php
$dca = $this->dcaFactory->get('tl_content');
$config = $dca->config();

$dca->config()->usesVersioning(); // true
$config()->usesVersioning(); // true

$GLOBALS['TL_DCA']['tl_content']['config']['enableVersioning'] = false;

$dca->config()->usesVersioning(); // false
$config()->usesVersioning(); // false
````

## Unit-testing
For a comfortable use of schemas within tests, a `DcaMockBuilder` is available, which makes the mocking of a DCA for unit tests very easy:

````php
$factory = $this->getDcaMockBuilder()
  ->addDcaData('tl_foo', [
    'config' => [
      'bar' => 'baz',
    ]
  ])
  ->addMocks('tl_foo', [
    'config.callback[onload]' => $onloadCallbacksMock
  ])
  ->addSpies('tl_foo', [
    'config.isClosed' => $this->exactly(1),
  ])
  ->getMock()
;

// Inject $factory to the subject of the test.
````

## Next steps
Since there is only little time left until Contao 4.13, I would suggest the following procedure (assuming the feature is appreciated for Contao 4.13):
- Integrate the DCA schemas for Contao 4.13 as an experimental feature.
- Use the schemas step-by-step within the core during Contao 4.13
	- thereby possibly also preparing/simplifying the DataContainers for a rework in the future
- Extend and finalize the schemas and the configuration tree during Contao 4.13

## Open issues
The following issues are still outstanding, but should be doable until Contao 4.13 RC:
- Agree onthe concept and fine-tune the implementation
- Clean up todos
- Create more unit tests
- Add additional nodes to the DcaConfiguration
- Add additional schemas and utility methods for schemas

## Assistance needed
- Maybe someone has an idea how to correctly mark up the individual callbacks for static code analysis with templates for a various number of parameters (e.g. `onload` with `DataContainer` and `button_callback` with `array, string, string etc.` instead of a generic `Callback<mixed>`).
- The individual schemas are not services, which is why regular dependency injection is not possible. Therefore, the draft currently includes a `ServiceSubscriberSchemaInterface`, which prepares the services for individual schemas. However, each subscribing schema has to be provided manually to the corresponding compiler pass. There may be a simpler way to allow schemas (= non-service classes) to access specific dependencies?

## Discuss this
I' m looking forward to input on this draft.

Ideally, the initial discussion should be limited to the concept and the basic features while the concrete schemas and nodes can be discussed in more detail at a later time.

### DIFF
--- a/core-bundle/config/commands.yaml
+++ b/core-bundle/config/commands.yaml
@@ -52,6 +52,7 @@ services:
         class: Contao\CoreBundle\Command\DebugDcaCommand
         arguments:
             - '@contao.framework'
+            - '@contao.dca.configuration_provider'
 
     contao.command.debug_fragments:
         class: Contao\CoreBundle\Command\DebugFragmentsCommand
--- a/core-bundle/config/commands.yaml
+++ b/core-bundle/config/commands.yaml
@@ -72,6 +72,9 @@ services:
             - '%kernel.project_dir%'
             - '@contao.twig.inspector'
 
+    contao.command.dump_dca_configuration:
+        class: Contao\CoreBundle\Command\DumpDcaConfigurationCommand
+
     contao.command.filesync:
         class: Contao\CoreBundle\Command\FilesyncCommand
         arguments:
--- a/core-bundle/config/listener.yaml
+++ b/core-bundle/config/listener.yaml
@@ -176,6 +176,22 @@ services:
         arguments:
             - '@database_connection'
 
+    contao.listener.data_container.schema.dc_file_exclude:
+        class: Contao\CoreBundle\EventListener\DataContainer\Schema\DcFileExcludeListener
+
+    contao.listener.data_container.schema.input_type_exclude:
+        class: Contao\CoreBundle\EventListener\DataContainer\Schema\InputTypeExcludeListener
+
+    contao.listener.data_container.schema.toggle_operation:
+        class: Contao\CoreBundle\EventListener\DataContainer\Schema\ToggleOperationListener
+        arguments:
+            - '@security.helper'
+
+    contao.listener.data_container.schema_data_validation:
+        class: Contao\CoreBundle\EventListener\DataContainer\Schema\SchemaDataValidationListener
+        arguments:
+            - '@contao.dca.validation.configuration_validator'
+
     contao.listener.data_container.start_stop_validation:
         class: Contao\CoreBundle\EventListener\DataContainer\StartStopValidationListener
         arguments:
--- a/core-bundle/config/services.yaml
+++ b/core-bundle/config/services.yaml
@@ -193,6 +193,9 @@ services:
         tags:
             - contao.dca.schema_dependency_provider
 
+    contao.dca.validation.configuration_validator:
+        class: Contao\CoreBundle\Dca\Validation\ConfigurationValidator
+
     contao.doctrine.backup.dumper:
         class: Contao\CoreBundle\Doctrine\Backup\Dumper
 
--- a/core-bundle/config/services.yaml
+++ b/core-bundle/config/services.yaml
@@ -186,7 +186,7 @@ services:
     contao.dca.schema_factory:
         class: Contao\CoreBundle\Dca\SchemaFactory
         tags:
-            - { name: 'container.service_subscriber', key: 'service_container', id: 'service_container' }
+            - { name: container.service_subscriber, key: service_container, id: service_container }
 
     contao.dca.schema_provider:
         class: Contao\CoreBundle\Dca\Provider\SchemaProvider
--- a/core-bundle/config/services.yaml
+++ b/core-bundle/config/services.yaml
@@ -157,6 +157,42 @@ services:
         tags:
             - { name: data_collector, template: '@ContaoCore/Collector/contao.html.twig', id: contao }
 
+    contao.dca.configuration_provider:
+        class: Contao\CoreBundle\Dca\Provider\ConfigurationProvider
+        arguments:
+            - '@contao.dca.driver_collection'
+            - '@event_dispatcher'
+
+    contao.dca.driver_collection:
+        class: Contao\CoreBundle\Dca\Driver\DriverCollection
+        arguments:
+            - !tagged_iterator contao.dca.driver
+
+    contao.dca.factory:
+        class: Contao\CoreBundle\Dca\DcaFactory
+        public: true
+        arguments:
+            - '@contao.dca.schema_factory'
+            - '@contao.dca.driver_collection'
+            - '@contao.dca.configuration_provider'
+
+    contao.dca.global_array_driver:
+        class: Contao\CoreBundle\Dca\Driver\GlobalArrayDriver
+        arguments:
+            - '@contao.framework'
+        tags:
+            - 'contao.dca.driver'
+
+    contao.dca.schema_factory:
+        class: Contao\CoreBundle\Dca\SchemaFactory
+        tags:
+            - { name: 'container.service_subscriber', key: 'service_container', id: 'service_container' }
+
+    contao.dca.schema_provider:
+        class: Contao\CoreBundle\Dca\Provider\SchemaProvider
+        tags:
+            - contao.dca.schema_dependency_provider
+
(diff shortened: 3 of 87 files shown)

### CODE REVIEW

**Comment by SeverinGloeckle** [comment]
To provide a starting point for discussing the concept and to show some of the benefits, I have provided an example on how to use the new schemas in `DataContainer::generateButtons` (see https://github.com/contao/contao/pull/3879/commits/3dd12c9e7bc057408e98024011e68bf73a554d9c):
- The method is much smaller and more readable.
	- This helps simplifying the dreaded legacy classes in the long run.
- A lot of the logic is split into small utility methods in the `Operation` schema.
	- These  are now easily testable in isolation (see [OperationTest](https://github.com/contao/contao/blob/0a8258606f01a00b7d7849f63b0afe2956b04c41/core-bundle/tests/Dca/Schema/OperationTest.php)).
	- These are reusable.

**Comment by SeverinGloeckle** [comment]
Since the topic of DCA came up several times at the Contao conference, I would like to take up this subject once more.

For a start, I would appreciate some feedback on the current draft:
- Is the basic concept still interesting?
    - The concept is described in detail in the original comment https://github.com/contao/contao/pull/3879#issue-1090686705. More background can be found in the issue https://github.com/contao/contao/issues/3687.
    - A simple use case with the advantages and the desired simplification – especially to `DC_` drivers – is outlined here: https://github.com/contao/contao/pull/3879#issuecomment-1010833925
- Is the current implementation the right way to go?

A few notes after checking the draft again myself:
- Instead of a single driver for a DCA, some kind of `DriverChain` should probably be used, to allow the composition of the final data structure from different source types.
  - For now, however, only the `GlobalArrayDriver` as `MutableDataDriverInterface` should be relevant for implementation.
- Autocompletion and type-hints for the custom schemas should be improved.
- The `DcaMockBuilder` is very shaky and should be ignored for now.

**Comment by SeverinGloeckle** [comment]
I have updated the initial PR from two years ago to be based on the current 5.x state. Additionally, the following changes and new functions are included to provide a better evaluation of the concept:

### Example usage in DataContainer::generateButtons

Since there were multiple changes, the original example from https://github.com/contao/contao/pull/3879#issuecomment-1010833925 didn’t hold up.

Thus, I have rewritten the example keeping the benefits and additionally moving the recently added logic outside of `DataContainer`:

- The check wether a toggle field should be hidden if the user has no access to the target field is extracted to [ToggleOperationListener](https://github.com/contao/contao/blob/34d2c40b0cde52d18647bcae7375f65969b1c387/core-bundle/src/EventListener/DataContainer/Schema/ToggleOperationListener.php).
- The “non generating” button callback checking the access for default operations has been moved to a new `permission_callback`.
  - I have kept the `DataContainerOperation` to minimize changes to the original code. However, this utility class may even be removed  and the permission check can be based on the Operation schema.
  - The reflection logic for cheating the `button_callback` into changing the operation data is no longer necessary and has been removed.

The advantages described in https://github.com/contao/contao/pull/3879#issuecomment-1010833925 therefore continue to apply. In addition, further logic is removed from the `DataContainer` via listeners (`SchemaCreatedEvent`) and thus integrated in a cleaner and testable way.



### Example usage in DataContainer::isFieldExcluded

A field schema knows whether it is excluded or not and provides this information.

The data container thus can simply query the field:

```php
public static function isFieldExcluded(string $table, string $field): bool
{
    return static::getDca($table)->fields()->field($field)->isExcluded();
}
```


Any special case regarding the exclusion of a field can be handled outside the data container:

- The special handling of fields in DC_File is extracted to [DcFileExcludeListener](https://github.com/contao/contao/blob/34d2c40b0cde52d18647bcae7375f65969b1c387/core-bundle/src/EventListener/DataContainer/Schema/DcFileExcludeListener.php)
- The default exclusion of fields with an input type is extracted to [InputTypeExcludeListener](https://github.com/contao/contao/blob/34d2c40b0cde52d18647bcae7375f65969b1c387/core-bundle/src/EventListener/DataContainer/Schema/InputTypeExcludeListener.php)

Both listeners use the `SchemaCreatingEvent` to adjust the exclude data of field schemas (either if their DCA uses DC_File or if an inputType or input_field_callback has been provided).

The whole `DataContainer::isFieldExcluded` method could even be deprecated now to reduce the public interface.


### Handling & deprecation of invalid DCA values

By default, Symfony will immediately throw exceptions for invalid configuration values. This behaviour is changed via two custom definitions to only trigger a deprecation for invalid values.

**1. Variable nodes**

`DcaNodeDefinition` wraps a regular definition and ensures that the configuration value matches the inner definition.

Example:

```php
->dcaNode('enableVersioning', 'boolean')->defaultFalse()
```

ensures that the configuration value for `enableVersioning` is always a boolean. If the value is not boolean, a deprecation is triggered and the provided default value is used instead.

> User Deprecated: Since contao/core-bundle 5.3: Setting an invalid DCA value has been deprecated and will no longer work in Contao 6. Invalid type for path "closed". Expected "bool", but got "string" at path "tl_page.config.closed". Falling back to value "false" until a valid value is provided.

**2. Array nodes**

`DcaArrayNodeDefinition` on the one hand allows additional array keys (e.g. tl_news.config.foo) and on the other hand replaces required nodes (`->isRequired()`) with `NULL` if no value was specified. In this case, a deprecation is triggered:

> User Deprecated: Since contao/core-bundle 5.3: Setting an invalid DCA value has been deprecated and will no longer work in Contao 6. A value for "dataContainer" must be configured at path "tl_opt_in_related.config.dataContainer". Falling back to value "NULL" until a valid value is provided.


**Benefits:**
This modelling should allow a graceful transition to predefined DCA structures. In Contao 6, for example, all `dcaNode('...', 'boolean')` could simply be replaced with real `boolean('...')` definitions and so on.



### Validation of schema data

In addition to the initial validation of data for an entire DCA when the configuration is created, any changes to the data at runtime are also validated against the `DcaConfiguration` (see [SchemaDataValidationListener](https://github.com/contao/contao/blob/34d2c40b0cde52d18647bcae7375f65969b1c387/core-bundle/src/EventListener/DataContainer/Schema/SchemaDataValidationListener.php) and [ConfigurationValidator](https://github.com/contao/contao/blob/34d2c40b0cde52d18647bcae7375f65969b1c387/core-bundle/src/Dca/Validation/ConfigurationValidator.php)).

This ensures that no invalid DCA data can be generated at a later point in time (as long as this data is run through a Schema).


### Dump the DCA configuration reference

The `config:dump-dca` command dumps the DCA configuration (similar to `config:dump-reference`).

```
contao-console config:dump-dca config.enableVersioning

# Contao DCA configuration at path "config.enableVersioning"

# If true, Contao saves the old version of a record when a new version is created.
enableVersioning:     false
```

Note that there is only a small subset of the DCA configuration defined and described at this point.

PR #3883, opened 2021-12-30
Improve the maintenance mode command

aschempp:
this adds the following features to the `contao:maintenance-mode` command

1. outputs the current state if no state (enable/disable) is given (or state value is invalid, see https://github.com/contao/contao/pull/3737#discussion_r774683461)
2. adds a format option to get the current maintenance state from JSON
3. ~restores backwards compatibility with previous versions by aliasing the `lexik:maintenance:lock` and `lexik:maintenance:unlock` commands 😎~


manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -43,26 +43,60 @@ protected function configure(): void
     {
         $this
             ->setName('contao:maintenance-mode')
-            ->addArgument('state', InputArgument::REQUIRED, 'Use "enable" to enable and "disable" to disable the maintenance mode. If the state is already the desired one, nothing happens. You can also use "on" and "off".')
+            ->setAliases(['lexik:maintenance:lock', 'lexik:maintenance:unlock'])
```
bytehead: What happens if someone has `LexikMaintenanceBundle` still installed?

aschempp: no idea what Symfony would do in that case, but add added handling for it in 85e2d1ccfc6b37e9a49231ddf15742257465a2d4

bytehead: Nice 👍

leofeyer (changes requested):
I would really prefer to add a conflict with the Lexik maintenance bundle instead of remapping the commands.

```json
"conflicts": {
    "contao/maintenance-bundle-deprecated": "*",
    "lexik/maintenance-bundle": "*"
}
```

aschempp:
Good idea, d1c266f1a0a612a22158e054f2b908381774d132 now conflicts with lexik/maintenance-bundle. We still need to add the aliases though for BC 😉

bytehead:
> I would really prefer to add a conflict with the Lexik maintenance bundle instead of remapping the commands.
> 
> ```json
> "conflicts": {
>     "contao/maintenance-bundle-deprecated": "*",
>     "lexik/maintenance-bundle": "*"
> }
> ```

I don't know if I'm in with that. Thinking about applications with the bundle and Contao installed...

leofeyer:
IIRC, the conflict is only present in the manager bundle, which you do not have in this case, do you?

aschempp:
yes, because the maintenance command is in the manager-bundle only as well, since it relies on our custom entry points.

bytehead:
You're right. Did totally miss that 🤦.

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -43,26 +43,62 @@ protected function configure(): void
...
```
leofeyer: > We still need to add the aliases though for BC 😉

Besides that there is no BC promise for command names, why do you want to provide a BC layer here? In case someone has added `vendor/bin/contao-console lexik:maintenance:lock` to their deploy script?

aschempp: exactly, deploy scripts or any other automation tools or CLI aliases.

Toflar: Not sure this is a good idea. People like @richardhj are testing on command existance to see which version is used. Something along the lines of

- list all commands
- if it contains `lexik:maintenance` then do this
- otherwise this

with an alias, it might possibly break.

Also, if you're updating the version, you are also updating the deployment, so you probably have to adjust deployment logic anyway.

composer.json
```diff
@@ -155,9 +155,11 @@
     "conflict": {
```
Toflar: I don't see any reason to add a conflict. If someone wants to use this on top for whatever reason, why should we care? Conflicts are here to notify people about things that cannot work together. We sure can because we don't care about those bundles anymore.

Toflar (changes requested):
I think the `conflicts` should be removed, hence requesting changes. For the alias, I really don't care but I'd love to hear from @richardhj.

composer.json
aschempp: This is for `contao/manager-bundle` only, where it never makes sense to manually require the bundle in my opinion.

Toflar: You're just creating additional rules for Composer to resolve for no reason. There is no conflict so there's no need to add one.

aschempp: Are you sure there is none? We're aliasing a command that potentially already exists

Toflar: As I wrote, we shouldn't alias the command either.

manager-bundle/src/Command/MaintenanceModeCommand.php
aschempp: the point of BC is that you don't need to adjust, but you do get a deprecation warning 🤷

leofeyer: I am with @Toflar on this one. If the command exists, people will falsely assume that the bundle is installed.

composer.json
aschempp: Well I think we should, thats why I added it in this PR 😂

Toflar: Then you should probably extract that part into its own PR. Otherwise nothing is going to make it into 4.x.

manager-bundle/src/Command/MaintenanceModeCommand.php
richardhj: 🤷 I won't be affected by either change. I understand both positions. But having a lexik: command while having a conflict on the lexik bundle feels wrong.

composer.json
leofeyer: I am with @Toflar on this one. We should not add a conflict and we should not alias the command.

Toflar (dismissed):

manager-bundle/src/Command/MaintenanceModeCommand.php
```diff
@@ -42,27 +44,55 @@ public function __construct(string $maintenanceFilePath, Environment $twig, File
...
+                    'maintenanceFilePath' => $this->maintenanceFilePath,
+                ],
+                JSON_THROW_ON_ERROR
+            ));
+
+            return 0;
+        }
+
+        if ($isEnabled) {
+            $io->note('Maintenance mode is enabled.');
```
leofeyer: Why is this a note and not an info like in line 94? And why do you return an error when there is nothing wrong? IMHO the code should be:

```php
if ($isEnabled) {
    $io->info('Maintenance mode is enabled.');
} else {
    $io->info('Maintenance mode is disabled.');
}

return 0;
```

aschempp: in my perspective, the enabled maintenance mode should be some sort of warning in the console. That includes a different message design and an exit code. Same as the back end previously added a "warning" bell that your system was in maintenance.

leofeyer: But `return 1` means "error", which means that a potential script will exit even though nothing is wrong.

ausi: Also makes the exit code dependent on the `format` which seams weird.

ausi (changes requested):
`contao:maintenance-mode off --format=json` output invalid JSON.

manager-bundle/src/Command/MaintenanceModeCommand.php
aschempp: well what's the exit code for "this is enabled" then? 😂

fritzmg: If you return anything other than zero, you cannot use this as part of your deployments, since your deployment script will likely abort. You can only use non-zero exit codes to signify specific errors, but successfully enabling the maintenance mode is not an error.

aschempp: Well, the exit code `1` is only returned if you _query_ for the status. And the most likely case if I query the status on deployment is that I _want_ it to fail if the mode is already enabled? Because for example I don't want to do another deployment that disables the maintenance mode?

ausi: If you really need that we should add a parameter like `--fail-if-enabled` (that works in every combination of format etc.)

This way you can use something like this:

```sh
$ contao:maintenance-mode --fail-if-enabled && do-something-only-if-not-in-maintenance foo bar
```

m-vo: From working on the command line in linux I, would expect an error code of 0 when successfully querying information. We should IMHO conform with that. 

The [Advanced Bash-Scripting Guide](https://tldp.org/LDP/abs/html/exit-status.html) says:
> 	A successful command returns a 0, while an unsuccessful one returns a non-zero value that usually can be interpreted as an error code. Well-behaved UNIX commands, programs, and utilities return a 0 exit code upon successful completion, though there are some exceptions.

leofeyer: I like the `--fail-if-enabled` idea if it is really necessary to return something other than `0` (which I still don't think it is).

aschempp: Changed in ba015aa126d342e9113844f56244316f20c9762b

ausi (dismissed):

leofeyer (approved):

PR #3884, opened 2021-12-30
Check for unpublished elements when generating the RSS feed

leofeyer:
The previous attempt was checking `TL_MODE` in the model, which leads to incorrect results if the model is used in a different context.

### DIFF
--- a/news-bundle/src/Resources/contao/models/NewsModel.php
+++ b/news-bundle/src/Resources/contao/models/NewsModel.php
@@ -244,8 +244,7 @@ class NewsModel extends Model
 			$arrColumns[] = "$t.featured=''";
 		}
 
-		// Never return unpublished elements in the back end, so they don't end up in the RSS feed
-		if (!BE_USER_LOGGED_IN || TL_MODE == 'BE')
+		if (!BE_USER_LOGGED_IN)
 		{
 			$time = Date::floorToMinute();
 			$arrColumns[] = "$t.published='1' AND ($t.start='' OR $t.start<='$time') AND ($t.stop='' OR $t.stop>'$time')";
--- a/calendar-bundle/src/Resources/contao/classes/Calendar.php
+++ b/calendar-bundle/src/Resources/contao/classes/Calendar.php
@@ -144,6 +144,12 @@ class Calendar extends Frontend
 		{
 			while ($objArticle->next())
 			{
+				// Never add unpublished elements to the RSS feeds
+				if (!$objArticle->published || ($objArticle->start && $objArticle->start > $time) || ($objArticle->stop && $objArticle->stop <= $time))
+				{
+					continue;
+				}
+
 				$jumpTo = $objArticle->getRelated('pid')->jumpTo;
 
 				// No jumpTo page set (see #4784)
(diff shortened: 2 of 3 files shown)


aschempp:
Wouldn't this prevent the entries from being shown in front end preview mode?

leofeyer:
No, the front end preview is not affected from this change. The check has been moved to the `generateFiles()` method, which is never called in the front end anyway.

news-bundle/src/Resources/contao/models/NewsModel.php
```diff
@@ -244,8 +244,7 @@ public static function findPublishedByPids($arrPids, $blnFeatured=null, $intLimi
 			$arrColumns[] = "$t.featured=''";
 		}
 
-		// Never return unpublished elements in the back end, so they don't end up in the RSS feed
-		if (!static::isPreviewMode($arrOptions) || TL_MODE == 'BE')
+		if (!static::isPreviewMode($arrOptions))
```
fritzmg, line 247: Instead of duplicating the published attributes check multiple times, why not use the scope matcher here?

fritzmg (changes requested):

news-bundle/src/Resources/contao/models/NewsModel.php
leofeyer: Because checking the scope is wrong here. The `findPublishedByPids()` method should always return the same results in the back end and in the front end. Just like the other `findPublishedByPids()` methods do.

fritzmg: I agree in principle, but it still seems wrong to me to duplicate this logic into several places now.

aschempp: we _could_ have a helper method in the model, but then again our models don't contain logic, right?

leofeyer: As I said, the model is the **wrong** place to check this! It should be checked where the RSS feed is generated.

Also, I don't see a problem with duplicating one single if condition into two separate extensions. Each extension has its own `generateFiles()` method with its own logic, so I don't understand why we are discussing this at all.

ausi: This check is already everywhere in the codebase I tink. 

I think streamlining this might be a good idea but should probably not be part of this PR.

ausi (approved):

news-bundle/src/Resources/contao/models/NewsModel.php
fritzmg: > I don't see a problem with duplicating one single if condition into two separate extensions.

@leofeyer I do - the condition is fairly complex and copying it and its logic over and over again is cumbersome and error prone. Case in point: the implementation in this PR is wrong: it is missing `Date::floorToMinute()`.

> This check is already everywhere in the codebase I tink.

Mostly in models, but there are some exceptions, yes.

ausi: > it is missing `Date::floorToMinute()`

This is only used for database queries AFAIK.

fritzmg: > This is only used for database queries AFAIK.

Yes - but that's wrong. The PHP check needs to use `Date::floorToMinute()` as well - otherwise you will have different results, i.e. without flooring the element in question would be invisible up to 59 seconds sooner, depending on where you check. We should probably fix this in the other places as well (`SitemapController` and `AbstractContentElementController` for instance).

PR #3885, opened 2021-12-30
Unify the command output format

aschempp:
`txt` is the default we use in other commands and Symfony uses as well.

### DIFF
--- a/core-bundle/src/Command/UserListCommand.php
+++ b/core-bundle/src/Command/UserListCommand.php
@@ -61,9 +61,9 @@ class UserListCommand extends Command
         $columns = $input->getOption('column');
 
         switch ($input->getOption('format')) {
-            case 'text': // Backwards compatibility
+            case 'text':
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

core-bundle/src/Command/UserListCommand.php
```diff
@@ -61,7 +61,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $columns = $input->getOption('column');
 
         switch ($input->getOption('format')) {
-            case 'text':
+            case 'txt':
+            case 'text': // Backwards compatibility
```
leofeyer: No `trigger_deprecation()` here?

aschempp: can do 😆  381ca36f79751235ef9344d73d59298b9343364c

leofeyer (approved):

PR #3886, opened 2021-12-30
Add the pagination info to the page title

### LINKED ISSUE #2523: Adding "page number" to "Page title" in pagination
For SEO reason, it's better to add page number to Page title in pagination. Because, currently, all pagination pages has same Title.

for example:
https://contao.org/en/news.html                         -> Read the official Contao announcements - Contao
https://contao.org/en/news.html?page_a104=2 -> Read the official Contao announcements - Contao 
be "Read the official Contao announcements - Page 2 of 37 - Contao"

leofeyer:
Implements #2523

@contao/developers To prevent that the pagination information is added multiple times (e.g. if there are multiple pagination menus on the same page), I am adding the menu with `sprintf()`. This also implies that people will have to add `%s` to the page title of the page to use the feature.

Is this a good approach or would you prefer to add the pagination info automatically?

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Pagination.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Pagination.php
@@ -10,6 +10,8 @@
 
 namespace Contao;
 
+use Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag\HtmlHeadBag;
+
 /**
  * Provide methods to render a pagination menu.
  *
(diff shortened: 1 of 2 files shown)


aschempp (changes requested):
How about just add an insert tag for the pagination info? Since one has to modify the page title anyway, this way I could actually add the info if and where I want to have it.
I think using `%s` is troublesome, we don't know if that placeholder is intended for the pagination.

fritzmg:
I find both using an insert tag and `%s` not a good idea. This means you basically have to add `%s` at the end of every page title in the back end where there could be a pagination. I find this rather confusing - and it would look weird in the back end. I think it should be appended automatically, but you should be able to configure it somehow (enable/disable it in the config or even as a page setting).

Toflar:
I'm not sure this is a good idea from a SEO perspective. I disagree with the author of the original issue.
Indexing a paginated list seems wrong to me in the first place because it changes all the time. Also, as of 4.13 we automatically set the `canonical` link of all `page=` links to the one without. So it wouldn't even be indexed.
If we want to adjust the title then imho it should only happen if the current response context also wants to keep `page` (or whatever) for the current canonical. Totally doable. So either don't do it at all or make it dependent on the canonical settings. The current state does not make sense though :)

leofeyer:
> Also, as of 4.13 we automatically set the canonical link of all page= links to the one without. So it wouldn't even be indexed.

That's true and it renders the PR redundant. 👍

PR #3887, opened 2021-12-31
Make sure the parent directory exists when writing files

### LINKED ISSUE #3582: Check if system is installed when loading language files
Btw. booting the Contao Framework and then using a Model method will cause the `loadLanguageFile` Hook to be executed, which in turn can have other unintended consequences that may even cause `cache:warmup --env=prod` on the console to fail. For example if your Contao instance uses `madeyourday/contao-rocksolid-custom-elements` and the `system/` folder is missing (as it usually is during a deployment), the following error will occur:

```
$ vendor/bin/contao-console cache:warmup -vvv

In Files.php line 159:

  [ErrorException]
  Warning: fopen(system/tmp/7ba9a9e97772f43b365094323d38535d): failed to open stream: No such file or directory  


Exception trace:
  at vendor\contao\core-bundle\src\Resources\contao\library\Contao\Files.php:159
 Contao\Files->fopen() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\File.php:829
 Contao\File->fputs() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\File.php:461
 Contao\File->write() at vendor\madeyourday\contao-rocksolid-custom-elements\src\CustomElements.php:1556
 MadeYourDay\RockSolidCustomElements\CustomElements::loadConfig() at vendor\madeyourday\contao-rocksolid-custom-elements\src\CustomElements.php:1633
 MadeYourDay\RockSolidCustomElements\CustomElements->loadLanguageFileHook() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\System.php:545
 Contao\System::loadLanguageFile() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\DcaExtractor.php:376
 Contao\DcaExtractor->createExtract() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\DcaExtractor.php:128        
 Contao\DcaExtractor->__construct() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\DcaExtractor.php:150
 Contao\DcaExtractor::getInstance() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\Model\QueryBuilder.php:33     
 Contao\Model\QueryBuilder::find() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\Model.php:1215
 Contao\Model::buildFindQuery() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\Model.php:1053
 Contao\Model::find() at vendor\contao\core-bundle\src\Resources\contao\library\Contao\Model.php:967
 Contao\Model::findAll() at vendor\contao\core-bundle\src\Framework\Adapter.php:44
 Contao\CoreBundle\Framework\Adapter->__call() at vendor\contao\core-bundle\src\Twig\Loader\TemplateLocator.php:76
 Contao\CoreBundle\Twig\Loader\TemplateLocator->findThemeDirectories() at vendor\contao\core-bundle\src\Twig\Loader\ContaoFilesystemLoaderWarmer.php:55
 Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoaderWarmer->warmUp() at vendor\symfony\http-kernel\CacheWarmer\CacheWarmerAggregate.php:98
 Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerAggregate->warmUp() at vendor\symfony\http-kernel\Kernel.php:584
 Symfony\Component\HttpKernel\Kernel->initializeContainer() at vendor\contao\manager-bundle\src\HttpKernel\ContaoKernel.php:323    
 Contao\ManagerBundle\HttpKernel\ContaoKernel->initializeContainer() at vendor\symfony\http-kernel\Kernel.php:786
 Symfony\Component\HttpKernel\Kernel->preBoot() at vendor\symfony\http-kernel\Kernel.php:125
 Symfony\Component\HttpKernel\Kernel->boot() at vendor\symfony\framework-bundle\Console\Application.php:168
 Symfony\Bundle\FrameworkBundle\Console\Application->registerCommands() at vendor\symfony\framework-bundle\Console\Application.php:74
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at vendor\symfony\console\Application.php:167
 Symfony\Component\Console\Application->run() at vendor\bin\contao-console(21) : eval()'d code:36
 eval() at vendor\bin\contao-console:21
```

_Originally posted by @fritzmg in https://github.com/contao/contao/issues/3576#issuecomment-943415123_

aschempp:
Fixes https://github.com/contao/contao/issues/3582

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -156,9 +156,9 @@ class Files
 	{
 		$this->validate($strFile);
 
-		if ($strMode[0] !== 'r' && \dirname($strFile) && \dirname($strFile) !== '.')
+		if ($strMode[0] != 'r' && ($strPath = \dirname($strFile)) && $strPath != '.')
 		{
-			$this->mkdir(\dirname($strFile));
+			$this->mkdir($strPath);
 		}
 
 		return fopen($this->strRootDir . '/' . $strFile, $strMode);
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -10,8 +10,6 @@
 
 namespace Contao;
 
-use Symfony\Component\Filesystem\Filesystem;
-
 /**
  * A class to access the file system
  *
@@ -91,7 +89,7 @@ class Files
 			return true;
 		}
 
-		return mkdir($this->strRootDir . '/' . $strDirectory);
+		return mkdir($this->strRootDir . '/' . $strDirectory, 0777, true);
 	}
 
 	/**
@@ -158,7 +156,10 @@ class Files
 	{
 		$this->validate($strFile);
 
-		(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
+		if ($strMode[0] !== 'r')
+		{
+			$this->mkdir(\dirname($strFile));
+		}
 
 		return fopen($this->strRootDir . '/' . $strFile, $strMode);
 	}
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -10,6 +10,8 @@
 
 namespace Contao;
 
+use Symfony\Component\Filesystem\Filesystem;
+
 /**
  * A class to access the file system
  *
@@ -156,13 +158,7 @@ class Files
 	{
 		$this->validate($strFile);
 
-		$filesystem = System::getContainer()->get('filesystem');
-		$parentDir = \dirname($this->strRootDir . '/' . $strFile);
-
-		if (!$filesystem->exists($parentDir))
-		{
-			$filesystem->mkdir($parentDir);
-		}
+		(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
 
 		return fopen($this->strRootDir . '/' . $strFile, $strMode);
 	}
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -156,7 +156,7 @@ class Files
 	{
 		$this->validate($strFile);
 
-		if ($strMode[0] !== 'r')
+		if ($strMode[0] !== 'r' && \dirname($strFile) && \dirname($strFile) !== '.')
 		{
 			$this->mkdir(\dirname($strFile));
 		}
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -10,9 +10,6 @@
 
 namespace Contao;
 
-use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\Filesystem\Filesystem;
-
 /**
  * A class to access the file system
  *
@@ -159,7 +156,13 @@ class Files
 	{
 		$this->validate($strFile);
 
-		$this->mkdir($this->strRootDir);
+		$filesystem = System::getContainer()->get('filesystem');
+		$parentDir = \dirname($this->strRootDir . '/' . $strFile);
+
+		if (!$filesystem->exists($parentDir))
+		{
+			$filesystem->mkdir($parentDir);
+		}
 
 		return fopen($this->strRootDir . '/' . $strFile, $strMode);
 	}
--- a/core-bundle/src/Resources/contao/library/Contao/Files.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Files.php
@@ -10,6 +10,9 @@
 
 namespace Contao;
 
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
 /**
  * A class to access the file system
  *
@@ -155,6 +158,8 @@ class Files
 	public function fopen($strFile, $strMode)
 	{
 		$this->validate($strFile);
+
+		$this->mkdir($this->strRootDir);
 
 		return fopen($this->strRootDir . '/' . $strFile, $strMode);
 	}


core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -156,6 +159,8 @@ public function fopen($strFile, $strMode)
 	{
 		$this->validate($strFile);
 
+		$this->mkdir($this->strRootDir);
```
ausi:
```suggestion
		$this->mkdir(\dirname($strFile));
```
Did you mean something like this?

ausi: But `$this->mkdir()` is not recursive so maybe better to use something like this?
```suggestion
		(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
```

fritzmg: Yes, definitely use the `Filesystem` component for mkdir.

core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -156,6 +156,14 @@ public function fopen($strFile, $strMode)
...
 		$this->validate($strFile);
 
+		$filesystem = System::getContainer()->get('filesystem');
+		$parentDir = \dirname($this->strRootDir . '/' . $strFile);
+
+		if (!$filesystem->exists($parentDir))
+		{
+			$filesystem->mkdir($parentDir);
+		}
+
```
ausi: No need to check for existence or ask the container IMO.

```suggestion
		(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
```

ausi (changes requested):

core-bundle/src/Resources/contao/library/Contao/Files.php
aschempp: but that will make it impossible to mock …

ausi: Mocking is not needed here, see the `fopen()` in the next line.

aschempp: Changed in 0e37183f2689cd33c2c6b7df34e02c57b197423e

ausi (dismissed):

leofeyer (changes requested):
I think we should check `$strMode` to actually know if a file is being written or just read. In the latter case, we should not create the folder.

Also, you should probably use the `Files::mkdir()` method instead of the Symfony `Filesystem` class to generate the folder.

fritzmg:
> Also, you should probably use the `Files::mkdir()` method instead of the Symfony `Filesystem` class to generate the folder.

`Files::mkdir()` does not create the directories recursively though.

leofeyer:
We should set this flag to `true` in line 92.

aschempp:
> I think we should check `$strMode` to actually know if a file is being written or just read. In the latter case, we should not create the folder.

If we don't create the parent directory, wouldn't still break if we are running a `fopen` call on a non-existing file & directory?

leofeyer:
> If we don't create the parent directory, wouldn't still break if we are running a `fopen` call on a non-existing file & directory?

Yes, probably. Just like when you enter a wrong filename, therefore I’d expect this behavior.

The point is that I don’t want `fopen('i/do/not/exist.pdf', 'r')` to create an orphan `i/do/not` folder in case the path is invalid. And your changes would do exactly this.

core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -156,6 +158,8 @@ public function fopen($strFile, $strMode)
 	{
 		$this->validate($strFile);
 
+		(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
```
ausi:
```suggestion
		if ($strMode[0] !== 'r')
		{
			(new Filesystem())->mkdir(\dirname($this->strRootDir . '/' . $strFile));
		}
```

core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -156,6 +156,11 @@ public function fopen($strFile, $strMode)
 	{
 		$this->validate($strFile);
 
+		if ($strMode[0] !== 'r')
```
ausi:
```suggestion
		if ($strMode[0] !== 'r' && \dirname($strFile) && \dirname($strFile) !== '.')
```

Otherwise files from the root directory could cause problems I think `->fopen('foo.txt')`.
_(A reason why I’d prefer using Symfony Filesystem ☺️)_

aschempp: what would that "attack" look like? `./file.php`?

aschempp: also, wouldn't that be catched by `is_dir` in `mkdir`?

ausi: `Files::getInstance()->fopen('test.txt', 'w')` results in `$this->mkdir('.')` and `'.'` is an insecure path (I hope).

aschempp: changed in 8209e66dbc84d36c41db05dafebb797a00980c55

aschempp: changed in 8209e66dbc84d36c41db05dafebb797a00980c55

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Files.php
```diff
@@ -156,6 +156,11 @@ public function fopen($strFile, $strMode)
 	{
 		$this->validate($strFile);
 
+		if ($strMode[0] !== 'r' && \dirname($strFile) && \dirname($strFile) !== '.')
+		{
+			$this->mkdir(\dirname($strFile));
+		}
+
```
leofeyer:
```suggestion
		if ($strMode[0] != 'r' && ($strPath = \dirname($strFile)) && $strPath != '.')
		{
			$this->mkdir($strPath);
		}

```

This might not be a problem at all due to the stat cache, but since `dirname()` calls are potentially expensive, it seems better to store the result in a variable.

Toflar: They are free in case of disabled open_basedir but we should change it anyway 👍

leofeyer (approved):

Toflar (approved):

PR #3888, opened 2021-12-31
Deprecate System::getTimeZones()

### LINKED ISSUE #3661: Refactor `System::getTimeZones()`
Refactor `System::getTimeZones()` similar to #3138

ausi:
Fixes #3661

While looking at this code, I realized, that we currently set the default time zone when initializing the framework in https://github.com/contao/contao/blob/6a815a472a46c44c4dbf3885f44c229a26fb8510/core-bundle/src/Framework/ContaoFramework.php#L361-L367 I’m not sure if this is a good idea as Symfony code that runs before our framework potentially has a different time zone set then.

I think we should deprecate the `timeZone` setting and remove it in Contao 5.0

### DIFF
--- a/core-bundle/src/Resources/contao/config/default.php
+++ b/core-bundle/src/Resources/contao/config/default.php
@@ -19,7 +19,7 @@ $GLOBALS['TL_CONFIG']['folderUrl']      = true;
 $GLOBALS['TL_CONFIG']['datimFormat'] = 'Y-m-d H:i';
 $GLOBALS['TL_CONFIG']['dateFormat']  = 'Y-m-d';
 $GLOBALS['TL_CONFIG']['timeFormat']  = 'H:i';
-$GLOBALS['TL_CONFIG']['timeZone']    = ini_get('date.timezone') ?: 'GMT';
+$GLOBALS['TL_CONFIG']['timeZone']    = date_default_timezone_get();
 
 // Input and security
 $GLOBALS['TL_CONFIG']['allowedTags']
--- a/core-bundle/src/Resources/contao/config/timezones.php
+++ b/core-bundle/src/Resources/contao/config/timezones.php
@@ -8,6 +8,8 @@
  * @license LGPL-3.0-or-later
  */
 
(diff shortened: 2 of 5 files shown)


ausi:
Funny side note: This method hasn’t changed since the initial GIT commit ☺️ https://github.com/contao/core/blob/4a2809d34ce08193c26b1c5815ecad2c1cdb909b/system/libraries/Controller.php#L573-L593

fritzmg:
> I think we should deprecate the timeZone setting and remove it in Contao 5.0

We will still need a way to override the server's time zone though.

ausi:
> We will still need a way to override the server's time zone though.

Via environment variable in the `ContaoKernel`?

Or via configuration in the `ContaoCoreExtension`?

ausi:
Do we even need to support hosting environments where you cannot set the time zone via php.ini?

fritzmg:
> Via environment variable in the `ContaoKernel`?
> 
> Or via configuration in the `ContaoCoreExtension`?

I'd be fine with either option. Since this is another configuration that is currently present in the back end system settings: moving it from there to the Contao Manager (i.e. a configuration of the Contao CoreBundle) (once we have the UI in place) probably makes the most sense?

fritzmg:
> Do we even need to support hosting environments where you cannot set the time zone via php.ini?

I think this should generally be a configuration of the application, not a configuration of the environment. You should be able to control to which time zone a date string relates to that you enter in the back end. Independent of the server's or platform's settings.

If we spin this further - in theory your application might serve back end users from all around the world and those users may independently want to enter dates relating to different time zones. Let's say you have a multi-domain instance where you serve `example.com` and `example.au` and you have multiple back end user groups, one for each of these domains. The group managing the content for `example.com` might want to enter their dates (for news or events for example) in one specific American time zone while the group managing the content for `example.au` might want to enter dates in one specific Australian time zone.

leofeyer (approved):

leofeyer:
> I think we should deprecate the `timeZone` setting and remove it in Contao 5.0

@ausi Do you want to add this in this PR or in a separate PR?

ausi:
separate PR IMO: #3900

leofeyer:
Thank you @ausi.

PR #3889, opened 2022-01-03
Add generic toggle operation handling

aschempp:
since I needed the toggle operation _again_ in https://github.com/contao/contao/pull/3872 I had to copy the same stuff all over again.  And since we don't have [Haste](https://github.com/codefog/contao-haste/blob/main/docs/Dca/index.md#easy-ajax-toggle-operation) in the core, this adds a generic toggle handling to DC_Table.

I'm aware this is not by far as powerful as Haste is, but I think it serves the 99% case. If merged we might also rewrite the existing callbacks, but I think that should not be part of this PR.


Toflar:
> If merged we might also rewrite the existing callbacks, but I think that should not be part of this PR.

That would help to see how it should be used though. Maybe just do it for one file as reference?

fritzmg:
> I'm aware this is not by far as powerful as Haste is, but I think it serves the 99% case.

It won't work for Content Elements though - at least it will not show the correct icon.

aschempp:
> That would help to see how it should be used though. Maybe just do it for one file as reference?

Added in 0c8d2cf78366742326216eb7b22ee38103eaff27

> It won't work for Content Elements though - at least it will not show the correct icon.

It does, but it requires a button callback (for the icon). We could add a "reverse" option to the operation, but since we need to permission-check for the operation anyway, it would not help in this case.

Unfortunately all core toggle operations are either inverted (tl_member.disable) or require additional permission checks, so we can't get rid of the button callback. But we sure can unify the operation handling 😊

Toflar:
> Unfortunately all core toggle operations are either inverted (tl_member.disable) or require additional permission checks, so we can't get rid of the button callback. But we sure can unify the operation handling 😊

I think we might still add the `reverse` option. If in future we foreach over all operations to ask the voters for permission, we don't need a callback anymore :)

Toflar (dismissed):
<3

core-bundle/src/Resources/contao/dca/tl_member.php
```diff
@@ -584,158 +580,4 @@ public function storeDateAdded($dc)
...
-	public function toggleIcon($row, $href, $label, $title, $icon, $attributes)
-	{
-		if (Input::get('tid'))
-		{
-			$this->toggleVisibility(Input::get('tid'), (Input::get('state') == 1), (func_num_args() <= 12 ? null : func_get_arg(12)));
-			$this->redirect($this->getReferer());
-		}
-
-		// Check permissions AFTER checking the tid, so hacking attempts are logged
-		if (!System::getContainer()->get('security.helper')->isGranted(ContaoCorePermissions::USER_CAN_EDIT_FIELD_OF_TABLE, 'tl_member::disable'))
```
leofeyer, line 609: You have kept this check in the `tl_content::toggleIcon()` method. Why did you remove it here?

aschempp: This security check (basic field access) is done by the generic implementation as well, since it's simple. `tl_content` needs to additionally check for access to the content element though (allow the operation), so it needs a button callback. Access to `act=toggle` (through URL manipulation) is then again done by `checkPermission` callback.

leofeyer: Does this mean that we can remove this check?

https://github.com/contao/contao/blob/b5ffcb1e816ce920c5b22966985fdc45cc638c1a/core-bundle/src/Resources/contao/dca/tl_content.php#L1983-L1986

Because it is already checked in the generic method:

https://github.com/contao/contao/blob/b5ffcb1e816ce920c5b22966985fdc45cc638c1a/core-bundle/src/Resources/contao/drivers/DC_Table.php#L2703-L2707

aschempp: The generic method is not executed if you have your own `button_callback`! You need to do everything on your own.
https://github.com/contao/contao/blob/b5ffcb1e816ce920c5b22966985fdc45cc638c1a/core-bundle/src/Resources/contao/classes/DataContainer.php#L911-L922

leofeyer: I see. 👍

leofeyer (changes requested):
There is still a problem with the permission check of `tl_content`. If I copy the toggle URL as a regular user and paste the ID of an element that I am not allowed to access, I can still toggle the field.

aschempp:
> There is still a problem with the permission check. If I copy the toggle URL as a regular user and paste the ID of an element that I am not allowed to access, I can still toggle the field.

Did you test that? Because I don't think you can, the default permission check should already prevent that: https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/dca/tl_content.php#L967-L969

leofeyer:
I have just sent you a video in Slack. 😉

leofeyer (approved):

Issue #3688, opened 2021-11-17
Separate the FE preview toolbar more from the FE page


Aybee:
Contao 4.9.22

The Preview Toolbar is too strongly influenced by the CSS of the site so that it is not forgotten for customers to control the toolbar and then set up additional CSS into the FE.

There would be some ways to prevent|reduce these problems.

- Toolbar in a frame
- Don't use the same CSS classes as in the FE templates
- Reset the CSS with `all:revert;`,  put an ID attribute into the HTML tag via JS and use it with the CSS to have a very strong selector

Example for last point with ID `preview` within the HTML tag:
```css
#preview .cto-toolbar * {
  all: revert;
}
#preview .cto-toolbar {
  font-family: -apple-system, system-ui, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
  font-weight: 400;
  font-size: 14px;
  line-height: 1;
  color: #444;
}
@media (-webkit-min-device-pixel-ratio: 2), (min-resolution: 192dpi) {
  #preview .cto-toolbar {
    font-weight: 300;
    color: #222;
  }
}
/* ... */
```
What do you think?

I also would like to revise the toolbar. E.g. why not simply `position:fixed;` the whole bar instead of some inner elements? So you also can get rid of the empty `div.cto-toolbar__clear`.

bytehead:
Yep, sometimes it's a little buggy 😬 

<img width="690" alt="Bildschirmfoto 2021-11-17 um 15 49 33" src="https://user-images.githubusercontent.com/754921/142223160-6d70d21b-c57a-43eb-81f4-f883f0b4c597.png">

Aybee:
Contao 4.9.22
closed
![preview_1](https://user-images.githubusercontent.com/1143013/142255121-2080bac1-7361-433b-b5de-84976b810936.jpg)
opened
![preview_2](https://user-images.githubusercontent.com/1143013/142255123-9f6b5153-23fc-422a-93b2-155a16381865.jpg)

Modified
closed
![preview_3](https://user-images.githubusercontent.com/1143013/142255125-1ef6b9e5-0885-4c79-bbea-cece559023d2.jpg)
opened
![preview_4](https://user-images.githubusercontent.com/1143013/142255126-e9ef840f-ec66-48d9-98b1-cf5a7e238530.jpg)

Modified with JS with button to directly open the real URL
closed
[...]

PR #3890, opened 2022-01-04
Reworked preview toolbar using Shadow DOM

bennyborn:
Using the magic of Shadow DOM it should be possible to make the styling of the preview toolbar truly independent of any CSS that might come with the users theme.

Quote from my comment in #3688:

- Created a wrapper element to encapsulate the whole thing
- Created the actual Shadow DOM with mode `open` so the contents can be modified by the rest of the JS and attached it to the wrapper (needed for the `copyButton` to work)
- Moved all the toolbar related styles inside the Shadow DOM
- Attaching the "original" toolbar-DIV to the Shadow DOM

Feel free to play around with it. 

I tested it by adding

```css
.formbody button {
    background: red !important;
    width: 500px;
    height: 500px;
}
```

to the main stylesheet of the page which would of course totally destroy the styling of the buttons but with Shadow DOM this doesn't bother the toolbar at all 😁

### DIFF
--- a/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
+++ b/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
@@ -198,7 +198,7 @@
 
         const toolbar = document.createElement('div');
         toolbar.classList.add('cto-toolbar');
-        shadow.appendChild(toolbar);
+        toolbarWrapper.shadowRoot.appendChild(toolbar);
 
         function request(method, uri, body, callback, addClass = true) {
             body = body || null;
--- a/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
+++ b/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
@@ -1,4 +1,11 @@
-<style>
+<style cto-toolbar>
+    cto-toolbar {
+        display: block;
+        position: fixed;
+        top: 0;
+        left: 0;
+        z-index: 99999;
+    }
     .cto-toolbar {
         font-family: -apple-system,system-ui,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif;
         font-weight: 400;
@@ -180,9 +187,19 @@
 </style>
 <script>
     (function () {
+
+        const toolbarWrapper = document.createElement('cto-toolbar');
+        document.querySelector('body').prepend(toolbarWrapper);
+
+        const shadow = toolbarWrapper.attachShadow({mode: 'open'});
+        toolbarWrapper.appendChild(shadow);
+
+        const css = document.querySelector('style[cto-toolbar]');
+        shadow.innerHTML = css.outerHTML;
+
         const toolbar = document.createElement('div');
         toolbar.classList.add('cto-toolbar');
-        document.querySelector('body').prepend(toolbar);
+        shadow.appendChild(toolbar);
 
         function request(method, uri, body, callback, addClass = true) {
             body = body || null;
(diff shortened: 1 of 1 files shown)


fritzmg:
As we are past the cut off date for Contao 4.13 this would only go into Contao 5.0. However, I would consider this more a bugfix, thus we could merge this into `4.9`? @contao/developers wdyt?

aschempp:
I think we should implement some sort of fallback logic if the browser does not support ShadowDOM (e.g. any IE version). It should be pretty simple, we could just fall back to a regular element instead. That would obviously not fix the current issue for these browsers, but that shouldn't really be a problem. What do you think @bennyborn ?

bennyborn:
Is IE support still relevant for the backend users? I was under the impression that this ship has already sailed since the current version of the preview toolbar already makes use of some ES6 features which breaks the compatibility with IE11.

But if absolutely necessary I guess we could rewrite the toolbar to have at least a working fallback without Shadow DOM.

aschempp:
> Is IE support still relevant for the backend users? I was under the impression that this ship has already sailed since the current version of the preview toolbar already makes use of some ES6 features which breaks the compatibility with IE11.

If it already does not work (I'm not aware of that) I think we can ignore it. Generally I try to add basic support far back if I can, just thinking of other countries or corporate environments where modern computers are not always available 😊 

If we add this to Contao 5, I would be totally fine with no IE support. Not sure if we _should_ break IE support (if there was) if this is a bugfix for Contao 4.9.

bennyborn:
For the record: It seems IE support was broken with c87c6a2. Just verified it using the online demo 😁

aschempp:
You're right, I can confirm that. `Array.from` is not supported in IE11 either. So just ignore what I said 😂

aschempp:
@leofeyer since you just assigned this to a milestone: I would agree with @fritzmg that this should be considered a bugfix for the existing preview bar (4.9+).

leofeyer:
The target branch is `4.x`, so the PR needs to be rebased first.

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -1,4 +1,11 @@
-<style>
+<style cto-toolbar>
+    cto-toolbar {
```
ausi, line 2:
```suggestion
<style>
    :host {
```

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -180,9 +187,19 @@
 </style>
 <script>
     (function () {
+
+        const toolbarWrapper = document.createElement('cto-toolbar');
+        document.querySelector('body').prepend(toolbarWrapper);
+
+        const shadow = toolbarWrapper.attachShadow({mode: 'open'});
+        toolbarWrapper.appendChild(shadow);
```
```suggestion
        toolbarWrapper.attachShadow({mode: 'open'});
```

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -180,9 +187,19 @@
...
+
+        const css = document.querySelector('style[cto-toolbar]');
+        shadow.innerHTML = css.outerHTML;
```
ausi, line 197:
```suggestion
        toolbarWrapper.shadowRoot.appendChild(Array.from(document.querySelectorAll('style')).at(-1));
```

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -180,9 +187,19 @@
...
+
         const toolbar = document.createElement('div');
         toolbar.classList.add('cto-toolbar');
-        document.querySelector('body').prepend(toolbar);
+        shadow.appendChild(toolbar);
```
```suggestion
        toolbarWrapper.shadowRoot.appendChild(toolbar);
```

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
```diff
@@ -248,29 +265,32 @@
...
-                            userList.innerHTML = '';
-                            JSON.parse(this.response).forEach(item => {
-                                const option = document.createElement('option');
-                                option.value = item;
-                                userList.appendChild(option);
-                            });
-                        }, false);
+                        if (timer) {
+                            clearTimeout(timer);
+                        }
```
ausi, line 276:
```suggestion
                        clearTimeout(timer);
```
No need to check for `timer` as `clearTimeout()` allows its parameter to be undefined.

ausi (commented):
Some minor tweaks. 
Apart from that ✅

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
bennyborn: I guess that could go wrong if there is anything else injecting a `<style>` into the body?

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
bennyborn: True. But I'm not even sure this should be still in here. Since my commit #3790 has been reviewed, even if not merged yet. I guess I'd wait until it is merged and then modify my PR accordingly.

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
bennyborn: Seems like a cleaner solution, will definitely give it a try!

core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig
ausi: At the point where this script gets executed the last `<style>` element is always the one directly before our script tag.

leofeyer:
@bennyborn Are you still working on this one?

bennyborn:
> @bennyborn Are you still working on this one?

Currently not. I'll give it another try in the upcoming weeks.

leofeyer:

PR #3896, opened 2022-01-06
Improve the toggle operation

aschempp:
Improvement on https://github.com/contao/contao/pull/3889 

1. for BC security it must not be possible to toggle any field unless it is explicitly enabled in the DCA field config
2. Also implemented toggle icon handling for header fields (parent sorting mode / 4)
3. update `tl_news` to show a DCA with multiple toggle fields


core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -101,7 +101,7 @@
 			),
 			'toggle' => array
 			(
-				'href'                => 'act=toggle&amp;field=invisible',
```
Toflar: I would expect a `toggleField => 'invisible'` or something like thtat still. Otherwise, if you add a second toggle field to the same DCA, you would also have to adjust the `href` of the previous toggle operation. That makes no sense imho.

Toflar (changes requested):

core-bundle/src/Resources/contao/dca/tl_content.php
aschempp: if you have multiple fields, you need to tell what field to toggle. I thought if you only have one field you might not want to add that to the config?

Toflar: I'll call you.

aschempp:
questions for @contao/developers 
1. should we also update the other core DCAs?
2. do we need to keep the `toggleIcon` and `toggleVisibility` methods for BC?

leofeyer:
> 1. should we also update the other core DCAs?

If you have time to do it, that would be great!

> 2. do we need to keep the `toggleIcon` and `toggleVisibility` methods for BC?

I am not sure if our BC promise covers DCA classes, but I lean towards "no".

fritzmg:
> 2. do we need to keep the toggleIcon and toggleVisibility methods for BC?

No, these callback classes are not part of our BC promise. We have already removed methods and whole classes in past feature and bugfix releases.

Toflar (commented):
Looks good, just wondering if the deprecated JS methods should forward to `toggleField` internally?

aschempp:
> Looks good, just wondering if the deprecated JS methods should forward to `toggleField` internally?

no, they behave differently and have a BC layer, which the new method does not.

aschempp:
> > 1. should we also update the other core DCAs?
> 
> If you have time to do it, that would be great!

done in 08988a82fb5b1c3a53b76bdc6fd01610d93bb147 😎

Toflar (dismissed):
Minus 2k LOC looks like a long overdue improvement 😉 Thanks!

leofeyer (changes requested):
I don't like the `rowIcon` concept. The `toggleVisibility` method did not have the argument and automatically toggled the icon if there was a `data-icon` and a `data-icon-disabled` attribute. Why do we have to add a new DCA setting, which, moreover, is not backwards compatible?

leofeyer:
I am struggling to understand your changes to the DCA files. You have completely removed the `toggleIcon()` method in:

* tl_calendar_events
* tl_faq
* tl_image_size_item
* tl_member_group
* tl_news
* tl_newsletter_recipients
* tl_style
* tl_user_group

However, you have kept the `toggleIcon()` method in:

* tl_article
* tl_comments
* tl_content
* tl_form_field
* tl_page
* tl_user

I see why we have to keep it in the `tl_user` file, because there is an additional "Protect admin accounts and own account" check. But why have you kept it in the other files?

aschempp:
> Why do we have to add a new DCA setting, which, moreover, is not backwards compatible?

Because the same method handles **all** toggle operations. And we don't want the feature-toggle to also enable/disable the row icon. Also, there is no BC break since this is a new method and new DC configuration.


> I see why we have to keep it in the `tl_user` file, because there is an additional "Protect admin accounts and own account" check. But why have you kept it in the other files?

They are all needed because they have custom permission checks on the icon.

- **tl_article** can only toggle according to page access config
- **tl_comments** can only toggle if you have access to the source element
- **tl_content** can only toggle if you have access to the element type
- **tl_form_field** can only toggle if you have access to the form field type
- **tl_page** can only toggle according to page access config 
- **tl_user** regular users cannot toggle admin accounts and you cannot toggle yourself

It's pretty easy to spot, all of these operations have additional checks 😉

aschempp:
@leofeyer honestly I disagree with the CS in 8ab038ced24f64867304c5eaf48bcc868432e456. I know we use `null` as a fallback for the existing legacy behaviour. But this is new code, and the fallback should be what I expect as default. Similar to `$request->get()` where I can pass in the fallback value to be something else than `null` 🤷

leofeyer:
I can revert the changes, but if an array property is not set, it is `null` by definition. This has been the case since I started working with PHP 4, therefore it seems wrong to me to use `false` or something else. 🤷‍♂️

aschempp:
The point of `??` is to shorten `if (isset($foo) ? $foo : false)`, and you wouldn't use `null` here? 😉

leofeyer:
No, your case is this:

```php
$foo = isset($bar['foo']) ? $bar['foo'] : false;
```

However, the case we are talking about is this:

```php
if (true === (isset($bar['foo']) ? $bar['foo'] : false)) {
}
```

And this case translates to "if foo is set and equals true". And in PHP, "is set" is defined as "not null". 🤷‍♂️ 

```php
if (true === ($bar['foo'] ?? null)) {
}
```

m-vo:
This...

> if (true === ($bar['foo'] ?? null)) {
}

... is mixing two types: bool and null. In case of an assignment, this would already trigger a code style warning in some tools (incompatible types). A fallback value should IMHO always be of the same type and our refactoring of the legacy code be the exception (that, being a bit coarse, reduces the likelihood of introducing errors).

core-bundle/src/Resources/public/core.js
```diff
@@ -431,6 +431,127 @@ var AjaxRequest =
...
+	 * @returns {boolean}
+	 */
+	toggleField: function(el, rowIcon) {
+		el.blur();
+
+		var img = null,
+			image = $(el).getFirst('img'),
+			published = (image.get('data-state') == 1),
+			div = el.getParent('div'),
+			index, next, icon, icond, pa, params;
```
leofeyer:
```suggestion
			next, icon, icond, pa;
```

It seems you are not using `index` and `params`?

leofeyer (changes requested):
The `rowIcon` concept remains a show stopper for me. I think we can do without it if we compare the icon name with the `data-icon` and `data-icon-disabled` attributes. In order to make it work, we have to remove the old fallback here:

```js
icon = image.get('data-icon') || AjaxRequest.themePath + 'icons/visible.svg';
icond = image.get('data-icon-disabled') || AjaxRequest.themePath + 'icons/invisible.svg';
```

@contao/developers WDYT?

core-bundle/src/Resources/public/core.js
```diff
@@ -431,6 +431,127 @@ var AjaxRequest =
...
+						icond = AjaxRequest.themePath + (icond.match(/\.svg$/) ? 'icons/' : 'images/') + icond;
+					}
+
+					img.setStyle('background-image', 'url(' + (!published ? icon : icond) + ')');
+				}
+			}
+		}
+
+		icon = image.get('data-icon') || AjaxRequest.themePath + 'icons/visible.svg';
+		icond = image.get('data-icon-disabled') || AjaxRequest.themePath + 'icons/invisible.svg';
```
leofeyer: The ` || AjaxRequest.themePath + 'icons/visible.svg'` part was a BC layer for the `toggleVisibility` method.

https://github.com/contao/contao/blob/4166f4eeccafb9cdcbd15e5e89728bbb4f1f3a81/core-bundle/src/Resources/public/core.js#L453-L456

It would prefer if we didn't include that in the new method as well.

core-bundle/src/Resources/public/core.js
```diff
@@ -581,20 +706,20 @@ var AjaxRequest =
 			div.getParent('div').getElement('pre').toggleClass('disabled');
 		}
 
-		icon = image.get('data-icon') || AjaxRequest.themePath + 'icons/visible.svg';
-		icond = image.get('data-icon-disabled') || AjaxRequest.themePath + 'icons/invisible.svg';
```
leofeyer: Why did you change the existing `toggleVisibility` method? Shouldn't it remain untouched for BC?

core-bundle/src/Resources/public/core.js
aschempp: Changed in f041bb08e3f284830623cbdde1e3d7da43f02e9b and also removed relative-path handling

core-bundle/src/Resources/public/core.js
aschempp: Fixed in 818a5f18c87cf97383fdac47406feecb1626e553

core-bundle/src/Resources/public/core.js
aschempp: Fixed in 818a5f18c87cf97383fdac47406feecb1626e553

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -953,7 +953,7 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 					if (($params['act'] ?? null) == 'toggle' && isset($params['field']))
 					{
 						// Hide the toggle icon if the user does not have access to the field
-						if (!System::getContainer()->get('security.helper')->isGranted(ContaoCorePermissions::USER_CAN_EDIT_FIELD_OF_TABLE, $this->strTable . '::' . $params['field']))
+						if (($GLOBALS['TL_DCA'][$strTable]['fields'][$params['field']]['toggle'] ?? false) !== true || !System::getContainer()->get('security.helper')->isGranted(ContaoCorePermissions::USER_CAN_EDIT_FIELD_OF_TABLE, $strTable . '::' . $params['field']))
```
leofeyer, line 956: Isn't this a BC break? Because the condition will also be true if the field has no `toggle` property, which is the case in Contao <4.13.

leofeyer (approved):

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer, line 956: Isn't this a BC break? Because the condition will also be true if the field has no `toggle` property, which is the case in Contao <4.13. So all toggles will disappear until the developers have added `'toggle' => true` in their DCAs.

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp: there was no `act=toggle` in previous versions

leofeyer: Apparently there was, because there is a `($params['act'] ?? null) == 'toggle'` in line 953 and you did not add this line in this PR.

aschempp: this PR **improves** the toggle operation added in Contao 4.13. I think you have too many PRs on your plate 😂🤗

leofeyer: This is very possible. 😄

leofeyer (approved):

PR #3904, opened 2022-01-07
Fix boolean empty value for doctrine 3

fritzmg:
If you have a field like this:

```php
$GLOBALS['TL_DCA']['tl_page']['fields']['addNewsMenuItems'] = [
    'exclude' => true,
    'inputType' => 'checkbox',
    'eval' => ['submitOnChange' => true],
    'sql' => ['type' => 'boolean', 'default' => false],
];
```

the following error will occur in the back end when saving the unchecked checkbox:

```
Doctrine\DBAL\Driver\PDO\Exception:
SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect integer value: '' for column `c412dev`.`tl_page`.`addNewsMenuItems` at row 1

  at vendor\doctrine\dbal\src\Driver\PDO\Exception.php:26
  at Doctrine\DBAL\Driver\PDO\Exception::new(object(PDOException))
     (vendor\doctrine\dbal\src\Driver\PDO\Statement.php:97)
  at Doctrine\DBAL\Driver\PDO\Statement->execute(array(false, '43'))
     (vendor\doctrine\dbal\src\Connection.php:1031)
  at Doctrine\DBAL\Connection->executeQuery('UPDATE tl_page SET `addNewsMenuItems`=? WHERE id=?', array(false, '43'))
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Database\Statement.php:286)
  at Contao\Database\Statement->query('', array(false, '43'))
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Database\Statement.php:242)
  at Contao\Database\Statement->execute(array(false, '43'))
     (vendor\contao\contao\core-bundle\src\Resources\contao\drivers\DC_Table.php:3248)
  at Contao\DC_Table->save(false)
     (vendor\contao\contao\core-bundle\src\Resources\contao\classes\DataContainer.php:515)
  at Contao\DataContainer->row('{title_legend},type,title,alias;{meta_legend},pageTitle,robots,description,serpPreview;{canonical_legend:hide},canonicalLink,canonicalKeepParams;{protected_legend:hide},protected;{layout_legend:hide},includeLayout;{cache_legend:hide},includeCache;{chmod_legend:hide},includeChmod;{expert_legend:hide},cssClass,sitemap,hide,noSearch,guests,requireItem;{newsmenu_legend},addNewsMenuItems;{tabnav_legend:hide},tabindex,accesskey;{publish_legend},published,start,stop')
     (vendor\contao\contao\core-bundle\src\Resources\contao\drivers\DC_Table.php:1948)
  at Contao\DC_Table->edit()
     (vendor\contao\contao\core-bundle\src\Resources\contao\classes\Backend.php:653)
  at Contao\Backend->getBackendModule('page', null)
     (vendor\contao\contao\core-bundle\src\Resources\contao\controllers\BackendMain.php:163)
  at Contao\BackendMain->run()
     (vendor\contao\contao\core-bundle\src\Controller\BackendController.php:49)
  at Contao\CoreBundle\Controller\BackendController->mainAction()
     (vendor\symfony\http-kernel\HttpKernel.php:152)
     (vendor\symfony\http-kernel\HttpKernel.php:74)
     (vendor\symfony\http-kernel\Kernel.php:202)
     (public\index.php:44)
  at require('public\\index.php')
     (public\app.php:13)   
```

The error is quite odd. The Symfony profiler shows the following database query being executed:

image

The stack trace also shows that the value `false` is used, and not an empty string. However, for some reason this is causing this exception somewhere inside Doctrine. Providing `0` instead of `false` works - and so this PR changes the empty field value for boolean types also to `0`.

This error does not occur in Contao 4.12.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Widget.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Widget.php
@@ -1491,14 +1491,9 @@ abstract class Widget extends Controller
 					return null;
 				}
 
-				if (\in_array($sql['type'], array(Types::BIGINT, Types::DECIMAL, Types::INTEGER, Types::SMALLINT, Types::FLOAT)))
+				if (\in_array($sql['type'], array(Types::BIGINT, Types::DECIMAL, Types::INTEGER, Types::SMALLINT, Types::FLOAT, Types::BOOLEAN)))
 				{
 					return 0;
-				}
-
-				if ($sql['type'] === Types::BOOLEAN)
-				{
-					return false;
 				}
 
 				return '';


aschempp:
has support for boolean been added in 4.13? Because I had this issue in Contao 4.9 as well, but I think it was never supported 🤔

fritzmg:
> has support for boolean been added in 4.13? Because I had this issue in Contao 4.9 as well, but I think it was never supported 🤔

No, that was added in Contao 4.9.11: https://github.com/contao/contao/commit/5cfdc9846fca5b58bd337ceeb8d9af5dc04802d0

aschempp:
so we should fix it in 4.9 and not 4.13?

fritzmg:
> so we should fix it in 4.9 and not 4.13?

Contao 4.9 uses Doctrine 2.x where this issue does not occur. It might be a bug in Doctrine too of course.

aschempp:
or maybe related to enabling/enforcing the strict mode in MySQL?

fritzmg:
> or maybe related to enabling/enforcing the strict mode in MySQL?

Yes and no. The error would probably not occur without strict mode - however as I said the error also does not occur in Contao 4.9 with Doctrine 2.x in strict mode. But again: I do not quite understand the error - according to the profiler the correct value (`0`) is already used for the query, and not an empty string as the error message suggests.

aschempp (changes requested):
This fix cannot be right, even though I understand it fixes the problem. Boolean is only stored as TINYINT in MySQL but likely not in other platforms. Apparently Doctrine needs to know a parameter type, otherwise it will not convert it. I'm further investigating this and will hopefully come up with another solution.

fritzmg:
Closing in favor of #3909

PR #3905, opened 2022-01-08
Deprecate the PackageUtil class

leofeyer:
This replaces the deprecated "package-versions" package with the `InstalledVersions` class from Composer v2. Intentionally not backported to Contao 4.9 to allow using Composer v1 for existing installations.


core-bundle/src/Util/PackageUtil.php
```diff
@@ -21,16 +21,21 @@ class PackageUtil
      */
     public static function getVersion(string $packageName): string
     {
-        $version = Versions::getVersion($packageName);
+        $version = InstalledVersions::getPrettyVersion($packageName) ?? '';
```
Toflar: That's incorrect. You should simply use `getPrettyVersion()` and not modify at all here. This is the one you would want to use when showing a version to the user.

core-bundle/src/Util/PackageUtil.php
```diff
@@ -21,16 +21,21 @@ class PackageUtil
...
 
     /**
      * Returns the version number as "major.minor.patch".
+     *
+     * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0;
+     *             use the getVersion() method instead
      */
     public static function getNormalizedVersion(string $packageName): string
     {
+        trigger_deprecation('contao/core-bundle', '4.13', 'The PackageUtil::getNormalizedVersion() method has been deprecated and will no longer work in Contao 5.0. Use the PackageUtil::getVersion() method instead.');
```
Toflar: This should not be deprecated but instead use `getVersion()` of Composer's `InstalledVersions`. This is the one you should use for `version_compare()` calls.

Toflar (changes requested):

core-bundle/src/Util/PackageUtil.php
```diff
@@ -12,7 +12,7 @@
 
 namespace Contao\CoreBundle\Util;
 
-use PackageVersions\Versions;
+use Composer\InstalledVersions;
 
 class PackageUtil
```
aschempp, line 21: shouldn't we deprecate our custom class as well and just use the Composer version?

core-bundle/src/Util/PackageUtil.php
leofeyer: That's wrong, because `getPrettyVersion()` can return `null` and `getVersion()` cannot. We could use a string cast instead of `?? ''` although the result will be the same. 🤷‍♂️

core-bundle/src/Util/PackageUtil.php
leofeyer: If you read the phpDoc comment, the method is supposed to return the version as `major.minor.patch`, which is not what `InstalledVersions::getVersion()` does. We have never used the method in the core and I don't want to maintain it in the future, therefore I have deprecated it.

core-bundle/src/Util/PackageUtil.php
leofeyer: I would be fine with that as well, however, then we would have to duplicate the code of the `getContaoVersion()` method at several places.

core-bundle/src/Util/PackageUtil.php
Toflar: Want to explain to me how Composer works? :) Yes, the pretty version is optional, we can fall back to the normalized one if there is none. But we always need 2 methods and not just one because the normalized and pretty versions are not the same. Right now you deprecated the normalized one which is almost always the one you want to use.

Toflar: There are just two representations. One of which is meant to be used for the users to improve readability and the other one is the one used for comparisons. What third options do we need then?

leofeyer: > Want to explain to me how Composer works? :)

No, but obviously I have to explain to you how the PackageUtil class works. We only need two methods in the future: `getVersion()` and `getContaoVersion()`. Both are supposed to return the pretty version without a leading `v` as string, because that is how we need the version string. For anything beyond that, please directly use the InstalledVersions class.

core-bundle/src/Util/PackageUtil.php
m-vo: As it's a small static method, it could live somewhere else, couldn't it? Maybe in `ContaoCoreBundle`?

aschempp: Maybe simply `ContaoFramework::getVersion()`?

leofeyer: I have deprecated the whole `PackageUtil` class in 6489dba43faed73f92d3e7589f45d28eb43cd7ac now.

core-bundle/src/DependencyInjection/Compiler/AddPackagesPass.php
```diff
@@ -12,24 +12,30 @@
...
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * Adds the composer packages and version numbers to the container.
  *
  * @internal
+ *
+ * @deprecated Deprecated since Contao 4.5, to be removed in Contao 5.0; use
+ *             the PackageUtil class instead
```
aschempp: that's probably wrong?

fritzmg (dismissed):
👍 for `ContaoCoreBundle::getVersion()`.

core-bundle/src/Util/PackageUtil.php
```diff
@@ -14,13 +14,20 @@
...
+ *             the Composer\InstalledVersions class instead
+ */
 class PackageUtil
 {
     /**
      * Returns the version number of a package.
      */
     public static function getVersion(string $packageName): string
     {
+        trigger_deprecation('contao/core-bundle', '4.13', 'Using the PackageUtil::getVersion() method has been deprecated and will no longer work in Contao 5.0. Use the Composer\InstalledVersions class instead.');
```
aschempp, line 28: shouldn't the deprecation be triggered in the class itself and not in every method?

aschempp (changes requested):

core-bundle/src/Util/PackageUtil.php
leofeyer: These are all static methods…

fritzmg: Yes, but the deprecation is still triggered once the autoloader loads the class, is it not?

leofeyer: That is not what I want. I want the deprecations to be triggered if a method is used, not if the class is autoloaded. Also, the deprecations have different messages, so…

core-bundle/src/DependencyInjection/Compiler/AddPackagesPass.php
leofeyer: Good catch. Fixed in 415922e3d5f5be3a6b259ebfa653a5a154e6e6bc.

core-bundle/src/Util/PackageUtil.php
fritzmg: > I don't want that. I want the deprecations to be triggered if a method is used, not if the class is autoloaded.

Why not, we have already used that strategy for a lot of classes. The deprecation should be triggered as soon as anyone uses the class, regardless of which method.

aschempp: but that's what we usually do if the whole class is deprecated …
https://github.com/contao/contao/blob/27f248764c86bd9929ec1b961c9241b5e593c9e2/core-bundle/src/EventListener/UserAwareTrait.php
https://github.com/contao/contao/blob/89833ae383bda4d86d60d909b01c3fe7d13edbb9/core-bundle/src/Resources/contao/controllers/FrontendCron.php
https://github.com/contao/contao/blob/ef97b65f6965e78c8cf6a824dbcd0dc28e9e4f72/installation-bundle/src/Database/AbstractVersionUpdate.php
https://github.com/contao/contao/blob/d66886594592f41a11084286e20bebf314a1c1f6/core-bundle/src/Resources/contao/elements/ContentMarkdown.php
https://github.com/contao/contao/blob/b8ce2592bd15750c0adbffcf5d829c7c107328b3/core-bundle/src/Framework/ScopeAwareTrait.php

leofeyer: Jeez… Why are we even discussing this? We are just waisting time.

Some of the deprecations say "use Composer\InstalledVersions" and others say "use ContaoCoreBundle::getVersion()". How am I supposed to distinguish between the two at autoloading time?

m-vo (dismissed):
LGTM

Regarding the removed `AddPackagesPassTest` - was this accidental? IMHO this should stay as long as the code is there (even though deprecated), but maybe I'm missing something,

aschempp (dismissed):
> Some of the deprecations say "use Composer\InstalledVersions" and others say "use ContaoCoreBundle::getVersion()".

Thats a valid argument 😊

Toflar (dismissed):

leofeyer:
> Regarding the removed `AddPackagesPassTest` - was this accidental? IMHO this should stay as long as the code is there (even though deprecated), but maybe I'm missing something,

Yes, sorry. Restored in ed0d75a311cebb279bdaacb363be9e8ebae7dbd1.

m-vo (approved):

PR #3909, opened 2022-01-10
Correctly handle types and empty values in DC_Table::save()

aschempp:
Fixes https://github.com/contao/contao/pull/3904

I was unable to figure out why this did work in Doctrine 2, but Doctrine 3 can only convert parameters based on their types, and we're not giving it the type information, even though it's there. So I extended our database statement and added the type to DC_Table if available.

Also noticed that the fallback empty value seems incorrect to me and should be inferred from the type/SQL as well.

FYI I also tried to apply this to Contao 4.9 but our database abstraction has changed significantly (rewrite on DBAL by @ausi) and it would require a lot more changes in 4.9 that I would be afraid to apply in a patch version.


fritzmg:
Hm, but this should actually automatically work: https://github.com/doctrine/dbal/blob/66bcb0bc86bdfb980420a4d19eec39289401ff96/src/Platforms/MySQLPlatform.php#L1075

It should not be necessary to specify the type in this case.

// ah nvm, this is actually only used for the schema manager 🤔

fritzmg (dismissed):

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -283,7 +283,7 @@ static function ($varParam)
 		// TODO: remove the try/catch block in Contao 5.0
 		try
 		{
-			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams);
+			$this->statement = $this->resConnection->executeQuery($this->strQuery, $arrParams, $arrTypes);
```
ausi, line 286: Needs to be added on line 303 too I think.

And additionally to passing the types we should probably cast boolen values to integers on line 270.

aschempp: added to second statement in 602d55b7e1432ee78a143d0732be23fc1e13a0ba. And no we must not cast, that's exactly the point of this PR. The database platform is supposed to know to convert the type.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3221,31 +3221,35 @@ protected function save($varValue)
 		// Save the value if there was no error
 		if (((string) $varValue !== '' || !($arrData['eval']['doNotSaveEmpty'] ?? null)) && ($this->varValue !== $varValue || ($arrData['eval']['alwaysSave'] ?? null)))
 		{
+			$varEmpty = Widget::getEmptyValueByFieldType($GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['sql'] ?? array());
+			$strType = $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['sql']['type'] ?? null;
```
leofeyer: Should we apply the `array_filter()` here already instead of three times in the code below?

aschempp: I thought about this as well, but I think this is more explicit. The types array is related to the parameters given to the SQL method. This **string** is just the type of the current field.

leofeyer: Yeah, but `$strType` is not used anywhere else, so why declare the variable at all?

aschempp: not sure what you mean. The array is created "on the fly" to match the other argument in the specific place. How else would you get `$strType`?

Maybe to make this more clear: The values and types passed to `->query()` **must** have the same array keys to work. Reading line 3248 to 3252 is all I need to understand that this is correct. If we build the array before, one has to analyse the whole method to mentally make sure that's correct. I even had to look up in the manual what array keys `array_unshift` produces for the types to work correctly 🙃

leofeyer: Not sure about that. There are some basic programming principles and one of it is DRY – don't repeat yourself. And right now, you are repeating yourself, because your are executing `array_filter(array($strType)` three times in the same method instead of just once. 🤷‍♂️

aschempp: https://softwareengineering.stackexchange.com/a/194507 🤷 
ultimately I don't care, feel free to CS it if you have to.

ausi:
```suggestion
			$arrTypes = array_filter(array($GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['sql']['type'] ?? null));
```

leofeyer: See cac1c988a72ebeb13f3107d8fc1b996fcf27ab1d

ausi (commented):
I still think that casting the boolean values to integers would be correct so that boolean parameters work for all cases (as they did with doctrine 2) as noted in https://github.com/contao/contao/pull/3909#discussion_r781225408

Apart from that ✅

aschempp:
If we cast the values ourselves, the whole PR becomes obsolete. The Doctrine type does the casting for us, thats why we need to provide it in the first place.

ausi:
> The Doctrine type does the casting for us, thats why we need to provide it in the first place.

Then we should pass the boolean type for all values that are `is_bool()` IMO (if no explicit type array was passed) to keep BC.

m-vo (approved):

aschempp:
> > The Doctrine type does the casting for us, thats why we need to provide it in the first place.
> 
> Then we should pass the boolean type for all values that are `is_bool()` IMO (if no explicit type array was passed) to keep BC.

I thought about that as well actually, but Contao never supported boolean types so far, and I didn't wanna risk changing the behaviour.

ausi:
> and I didn't wanna risk changing the behaviour.

But the bahavior is changed now, isn’t it? With doctrine 2 boolean values worked and with doctrine 3 they do not work anymore.

I will investigate and create a new PR if necessary

PR #3910, opened 2022-01-10
Fix the ScriptHandlers for Composer 2.3

fritzmg:
Composer 2.3 will currently break our ScriptHandlers, because the PHP and `symfony/process` requirement of Composer are finally bumped up (https://github.com/composer/composer/commit/d2ede370be20aedfc74749bc87394a79ed4c1e52). 

```
Fatal error: Uncaught TypeError: Argument 1 passed to Symfony\Component\Process\Process::__construct() must be of the type array, string given, called in vendor\contao\contao\manager-bundle\src\Composer\ScriptHandler.php on line 69
```

This PR backports the changes from Contao 4.10 to 4.9, where the `symfony/process` cross-compatibility was already introduced.

### DIFF
--- a/core-bundle/src/Composer/ScriptHandler.php
+++ b/core-bundle/src/Composer/ScriptHandler.php
@@ -15,6 +15,7 @@ namespace Contao\CoreBundle\Composer;
 use Composer\Script\Event;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
+use Webmozart\PathUtil\Path;
 
 class ScriptHandler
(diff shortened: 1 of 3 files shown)


Toflar (approved):

m-vo (approved):
I wonder if we should migrate 4.9 to use the `contao-setup` binary as well. These are the kind of errors that motivated it in the first place and 4.9 will still live quite a while.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #3913, opened 2022-01-10
Dispatch events when compiling and rendering Twig templates

m-vo:
Next attempt after #3852. This one takes 'modules' into accout, i.e. each traversed template, no matter if called directly via `Environment#render()` or if it was embedded/included.

The PR introduces two events:

1) A `RenderTemplateEvent` that is dispatched from within the template when it is rendered/displayed. You can use this event to alter the context.
    ```php
    public function __invoke(RenderTemplateEvent $event): void
    {
         if ('@Foo/bar.html.twig' === $event->getName()) {
            $event->setValue('foo', 42); // context now contains 'foo' that is 42
        }
    }
    ```
    It's accomplished by injecting a custom `RenderEventNode` to the head of each module.

2) A `CompileTemplateEvent` that is dispatched when the Twig template is compiled (source to AST to dumped PHP). You can use this event to alter/inject nodes. For now, I added two helpers `prepend` and `append`, which can be used like this (other things could follow, like `include($template)` for instance):
    ```php
    public function __invoke(CompileTemplateEvent $event): void
    {
        if ($event->matchType('html', 'html5') && $event->isContaoTemplate()) {
            $name = $event->getName();

            $event->prepend("\n<!-- TWIG TEMPLATE START: $name -->");
            $event->append("<!-- TWIG TEMPLATE END: $name -->\n");
        }
    }
    ```
    This is accomplished by dispatching an event from our `EventsNodeVisitor` when walking the AST. If you want to have a closer look: The `AddTemplateDebugInformationListener` actually contains the full example.

There is a notable difference to how the current template hooks work: There is no HTML that you can alter. Neither during compilation, nor during render do we know anything about the output. For this you would still need to overwrite the `Environment` which is probably a bad idea. 

Advanced users can probably do pretty much everything they need by implementing their own node visitor, but I see the `CompileTemplateEvent` as an extension point that is way easier to use.

### DIFF
--- a/core-bundle/src/Event/CompileTemplateEvent.php
+++ b/core-bundle/src/Event/CompileTemplateEvent.php
@@ -144,17 +144,17 @@ class CompileTemplateEvent
 
     private function prependNode(Node $node): void
     {
-        $this->node->setNode(
-            'display_start',
-            new Node([$node, ...$this->node->getNode('display_start')])
-        );
+        $inner = iterator_to_array($this->node->getNode('display_start'), true);
+        array_unshift($inner, $node);
+
+        $this->node->setNode('display_start', new Node($inner));
     }
 
     private function appendNode(Node $node): void
     {
-        $this->node->setNode(
-            'display_end',
-            new Node([...$this->node->getNode('display_end'), $node])
-        );
+        $inner = iterator_to_array($this->node->getNode('display_end'), true);
+        $inner[] = $node;
+
+        $this->node->setNode('display_end', new Node($inner));
     }
 }
--- a/core-bundle/src/Event/CompileTemplateEvent.php
+++ b/core-bundle/src/Event/CompileTemplateEvent.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Event;
+
+use Contao\CoreBundle\Twig\Event\TemplateNameTrait;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\ModuleNode;
+use Twig\Node\Node;
+use Twig\Node\PrintNode;
+
+/**
+ * @experimental
+ */
+class CompileTemplateEvent
+{
+    use TemplateNameTrait;
+
+    private ModuleNode $node;
+
+    /**
+     * @internal
+     */
+    public function __construct(ModuleNode $node)
+    {
+        $this->node = $node;
+
+        $sourceContext = $node->getSourceContext();
+        $this->name = null !== $sourceContext ? $sourceContext->getName() : '';
+    }
+
+    /**
+     * Add raw content that will be output before the template body.
+     */
+    public function prepend(string $content): void
+    {
+        $this->node->setNode(
+            'display_start',
+            new Node([
+                $this->node->getNode('display_start'),
+                new PrintNode(new ConstantExpression($content, 0), 0),
+            ])
+        );
+    }
+
+    /**
+     * Add raw content that will be output after the template body.
+     */
+    public function append(string $content): void
+    {
+        $this->node->setNode(
+            'display_end',
+            new Node([
+                new PrintNode(new ConstantExpression($content, 0), 0),
+                $this->node->getNode('display_end'),
+            ])
+        );
+    }
+}
--- a/core-bundle/src/Event/CompileTemplateEvent.php
+++ b/core-bundle/src/Event/CompileTemplateEvent.php
@@ -13,7 +13,10 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Event;
 
 use Contao\CoreBundle\Twig\Event\TemplateNameTrait;
+use Twig\Node\Expression\ArrayExpression;
 use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Expression\NameExpression;
+use Twig\Node\IncludeNode;
 use Twig\Node\ModuleNode;
 use Twig\Node\Node;
 use Twig\Node\PrintNode;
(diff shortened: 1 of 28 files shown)


fritzmg:
https://giphy.com/gifs/primevideo-2020-borat-subsequent-moviefilm-Od0QRnzwRBYmDU3eEO

Toflar:
Super smart! But you knew that already 😉 I like the concept and I would approve the implementation like that 👍

m-vo:
I added the possibility to prepend/append other templates like you would with `{% include '_my_template' %}` and included all tests. This PR is IMHO ready to be reviewed.

Note, that you need to manually clear the Twig cache when you want to test changes of your `CompileTemplateEvent` listener as it will only be called once when compiling; `rm -rf var/cache/dev/twig` does the trick, there is no (easy) way round.

fritzmg:
> Note, that you need to manually clear the Twig cache when you want to test changes of your CompileTemplateEvent listener as it will only be called once when compiling;

Hm, so there is no way to dynamically append or prepend additional output similar to the `outputFrontendTemplate` hook? This would still be needed, otherwise extensions cannot add additional HTML attributes for example for arbitrary content elements (unless we will have a different solution for that particular use case).

SeverinGloeckle:
One issue unrelated to the functionality: In my opinion, the events would be better off in their own sub-namespace (e.g. `Contao\CoreBundle\Event\Twig`, `Contao\CoreBundle\Event\Template` or similar).

Since more and more events will be added, we should start structuring the namespace rather sooner than later.

m-vo:
> Hm, so there is no way to dynamically append or prepend additional output similar to the outputFrontendTemplate hook? This would still be needed, otherwise extensions cannot add additional HTML attributes for example for arbitrary content elements (unless we will have a different solution for that particular use case).

I think it's possible: with the `CompileTemplateEvent` you can both prepend/append static strings as well as whole templates to a template. If you need dynamic context you can inject it via the `RenderTemplateEvent`. But maybe I am missing your point. Could you give a more specific use case?

m-vo:
> One issue unrelated to the functionality: In my opinion, the events would be better off in their own sub-namespace (e.g. Contao\CoreBundle\Event\Twig, Contao\CoreBundle\Event\Template or similar).
> 
> Since more and more events will be added, we should start structuring the namespace rather sooner than later.

This holds true for more things besides events: listeners, exceptions, entities, repositories, migrations. service annotations, … I get your point and I would generally rather group by logic than type as well. But it's IMHO not easy to generalize. Some of these need a shared location to simplify configuration (e.g. entities, …) and so on. At the end consistency is probably more important and moving not really possible. But I'm up for a change…

What I want to say with this: Let's discuss this, but not in this PR. :smile:

fritzmg:
> Could you give a more specific use case?

Something like this: https://github.com/inspiredminds/contao-wowjs/blob/9cbabf521f1af523329ed7e32ccbdfc06d1a4b55/src/EventListener/HookListener.php#L51-L72 - but I am actually using the `getContentElement` hook there anyway, which is better suited.

m-vo:
Manipulating HTML does not work. For this we would would for instance need an event that is dispatched with the subrequest response. But I'm not sure we want that. :grimacing: 

For your use case, couldn't you just prepend and append a template in which you add another wrapper?

```php
    public function onCompile(CompileTemplateEvent $event): void {
        // test for applicable templates …
        
        $event->prependInclude('@Contao/inspiredminds_wowjs_wrapper_start.html.twig');
        $event->appendInclude('@Contao/inspiredminds_wowjs_wrapper_end.html.twig');
    }
```
with
```twig
{# inspiredminds_wowjs_wrapper_start.html.twig #}
<div data-wow-duration="{{ wow_duration }}" data-wow-delay="{{ wow_delay }}" …>
```
and 
```twig
{# inspiredminds_wowjs_wrapper_end.html.twig #}
</div>
```

fritzmg:
> For your use case, couldn't you just prepend and append a template in which you add another wrapper?

That's a poor-man's solution in my opinion - ideally the content element itself should receive the appropriate attributes. Otherwise introducing another wrapper would influence existing CSS :). But the `getContentElement` hook is better suited for this anyway.

m-vo:
> That's a poor-man's solution in my opinion - ideally the content element itself should receive the appropriate attributes. Otherwise introducing another wrapper would influence existing CSS :). But the getContentElement hook is better suited for this anyway.

Well, the 'correct' way in my opinion for this would be to have a generic base template for content elements. Then you can just alter that one template and it could even solve things like attributes by expanding an array/object from the context. Then you could also simply add things to the context and you're done.

We can probably do these things when introducing the new templates. Wdyt?

fritzmg:
> Well, the 'correct' way in my opinion for this would be to have a generic base template for content elements.

Yep, although that still would not work if you created a custom template and e.g. removed the attributes output (but that's on yourself then anyway I would say) or if an extension provides a template that does not output all attributes. We _kinda_ can do it currently by abusing the `cssID` variable that is present in all (?) content element and front end module templates. With the hook I can make sure that the attributes are output no matter what.

Toflar:
The typical use case would be to add/change an attribute on the very first HTML element, imho. Maybe we can have a helper for that, that would cover 99% of the use cases if you ask me.

m-vo:
> With the hook I can make sure that the attributes are output no matter what.

But you are still relying on a certain structure then, otherwise your regex will fail. :upside_down_face: And we should not promote editing rendered HTML, I guess.


> The typical use case would be to add/change an attribute on the very first HTML element, imho.  Maybe we can have a helper for that, that would cover 99% of the use cases if you ask me.

I agree an the 99%. We should IMHO introduce a base template for modules + content elements to solve that and use structured attributes in there. Trying to search and alter the AST for this would be error prone and difficult to achieve.

m-vo:
As discussed on the Contao call on 13th of january, the events should be moved to `Twig/Event`.

m-vo:
> As discussed on the Contao call on 13th of january, the events should be moved to Twig/Event.

Done.

Toflar:
Just FYI: We had a use case where we needed to adjust the `$this->classes` property of every content element which was previously done using the `parseTemplate` hook. You can still do this in Contao 5 with proper fragment content elements by registering a `kernel.request` event listener. All those requests share one thing: the `ContentElementReference`. It allows to pass on CSS classes but also arbitrary `templateProperties`.

So in your listener, you can do something along the lines of this:

```php
#[AsEventListener]
class MyListener
{
    public function __invoke(RequestEvent $event): void
    {

        if (!$request->attributes->has('contentModel')) {
            return;
        }

        $contentModel = $request->attributes->get('contentModel');

        if (!$contentModel instanceof ContentModel) {
            $contentModel = ContentModel::findByPk($contentModel);
        }

        // Get current classes ($request->attributes->get('classes'))
        // Modify classes

        $request->attributes->set('classes', $modified);
    }
}
```

m-vo:
> Just FYI: We had a use case where we needed to adjust the $this->classes property of every content element

You should also be able to adjust the `content_element/_base.html.twig` template in this case, btw. If you need custom processing, this would need a custom Twig function, though. 🙂

xprojects-de:
see https://github.com/contao/contao/issues/6823 
Is there a possibility of taking up the PR? :-)

PR #3914, opened 2022-01-11
Show route path with regexp in page settings

aschempp:
This is just an idea for improvement on the new routing priority feature.

**Before:**
Bildschirmfoto 2022-01-11 um 07 27 31

**After:**
Bildschirmfoto 2022-01-11 um 07 26 51


Toflar:
I definitely like the idea of showing what values the placeholders can get. However, I don't think this is very readable to a regular user. Regular expressions are really not well understood for users and this might get complex here.
Look at one of your test examples:

`foo/{bar}/{baz}.html` with `['bar' => '.+', 'baz' => '\d+']` becomes `foo/{.+}/{\d+}.html` in the back end. I don't think any user can understand this.
I think for them, `foo/{bar}/{baz}.html` is more understandable. However, this obviously lacks the information of the requirements for the ones that actually do understand regex. That's why I think we should provide the information but not incorporate directly.

Not sure if we have enough space for yet another column but if not, then what about something like this?

| Seitenalias | Routing Informationen                                               |
|-------------|---------------------------------------------------------------------|
| foo         | foo/{bar}/{baz}.html<br><br>* {bar} -> .+<br>* {baz} -> \d+ |

That way, one that does not understand regex can understand why it may collide (because there are the same parameters). And one that does understand regular expressions gets the additional information required.

core-bundle/tests/EventListener/DataContainer/PageRoutingListenerTest.php
```diff
@@ -38,10 +42,17 @@ public function testGetsPathFromPageRoute(): void
         $framework = $this->mockContaoFramework([PageModel::class => $pageAdapter]);
 
         $pageRoute = $this->createMock(PageRoute::class);
+
```
leofeyer: This seems to be an unrelated change that also does not match the code style.

aschempp:
I like @Toflar's idea, but that wouldn't work with the `Route path` field. The route path is also visible if there are no conflicts. Not sure how that could be solved?

leofeyer:
I don't think we can (and should) solve this at the moment. And in my opinion `foo/{.+}/{\d+}.html` is still more helpful to the users than `foo{!parameters}.html` which can really be anything.

Toflar:
> The route path is also visible if there are no conflicts. Not sure how that could be solved?

Can you elaborate on that? I don't know what you mean.

aschempp:
Bildschirmfoto 2022-01-11 um 13 32 49

Toflar:
I see. However, as the user cannot influence that anyway, I think it's less important. It's more important when showing conflicts because that's when they possibly have to understand and do something about it.

What if we did this in the conflicts block:

```html
<td>foo/<span title=".+">{bar}</span>/<span title="\d+">{baz}</span>.html</td>
```
It would not take up more space and show all the information we need on hover :) We can also invert, if you think the regex is more important than the placeholder. But I do think the placeholder is more important. It makes seeing the conflicts more straight forward for the untrained eye. Sure, for us it's easy to tell the difference but for the regular user, these are just strange signs with no meaning.

m-vo:
Was just about to write the same as @Toflar did. The `routePath` does not have to be an input field as it is disabled all the time, using an `input_field_callback` might be the preferred way anyways. :slightly_smiling_face: 

We could even go one step further and add "translations" for the most common placeholders.

aschempp:
It looks like this now, and you can hover over the parameters in both fields to get the details.
Bildschirmfoto 2022-01-12 um 09 18 47

Toflar:
Yes! I love that! Now the CI chain and then we can approve this :)

core-bundle/tests/EventListener/DataContainer/PageRoutingListenerTest.php
aschempp: so we're discussing empty newline codestyle now? To me it's a newline because we have more than one call on the object. But feel free to CS it 🤷

Toflar (dismissed):

core-bundle/tests/EventListener/DataContainer/PageRoutingListenerTest.php
leofeyer: No, we are just discussing unrelated changes. You should make them in a separate PR. 🤷‍♂️

core-bundle/tests/EventListener/DataContainer/PageRoutingListenerTest.php
```diff
@@ -484,4 +546,23 @@ public function testSkipsSimilarPageIfUrlDoesNotMatch(): void
 
         $this->assertSame('', $listener->generateRouteConflicts($dc));
     }
+
+    /**
+     * @return PageRoute&MockObject
+     */
+    private function mockPageRoute(string $path, array $requirements = []): PageRoute
```
leofeyer, line 541: The `$requirements` argument is never used. Are the tests perhaps incomplete?

leofeyer (changes requested):

core-bundle/tests/EventListener/DataContainer/PageRoutingListenerTest.php
aschempp: Good catch. The tests were actually complete but because I duplicated the method instead of using it 😂 fixed in 3540ce501714be444c7397759c84b83af35e30c6

leofeyer (dismissed):

leofeyer:
087fd60eb5006a1cd1853c252cb08796d840ed6e ensures that the tip is correctly formatted:

leofeyer (dismissed):

PR #3915, opened 2022-01-11
Do not generate URLs for insert tags that don’t need it

aschempp:
It makes no sense to call routing and generate a URL if the insert tag does not need it. It also unnecessarily prevents these insert tags from working in Contao 4.12/4.13 if the page URL cannot be generated (non-routable / requires parameter).

FYI I intentionally did not remove the `case` for the same tags in the later method because it is required for non-regular pages (the new switch-case is inside an if-else handling).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -436,8 +436,9 @@ class InsertTags extends Controller
 							break;
 						}
 
-						// Do not generate URL for insert tags that don't need it
 						$strUrl = '';
+
+						// Do not generate URL for insert tags that don't need it
 						if (\in_array(strtolower($elements[0]), array('link', 'link_open', 'link_url'), true))
 						{
 							switch ($objNextPage->type)
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -436,6 +436,23 @@ class InsertTags extends Controller
 							break;
 						}
 
+						$strName = $objNextPage->title;
+						$strTarget = $objNextPage->target ? ' target="_blank" rel="noreferrer noopener"' : '';
+						$strClass = $objNextPage->cssClass ? sprintf(' class="%s"', $objNextPage->cssClass) : '';
+						$strTitle = $objNextPage->pageTitle ?: $objNextPage->title;
+
+						// Early return for tags that do not need the URL
+						switch (strtolower($elements[0]))
+						{
+							case 'link_title':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strTitle);
+								break 2;
+
+							case 'link_name':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strName);
+								break 2;
+						}
+
 						// Page type specific settings (thanks to Andreas Schempp)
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):

fritzmg (dismissed):

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -436,6 +436,23 @@ protected function doReplace($strBuffer, $blnCache)
...
+						{
+							case 'link_title':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strTitle);
+								break 2;
+
+							case 'link_name':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strName);
+								break 2;
+						}
+
```
ausi: I think we could get rid of the duplicated code (and the `break 2`s) if we instead use something like this:

```php
						// Do not generate URLs for tags that do not need it
						if (in_array(strtolower($elements[0]), array('link_title', 'link_name')))
						{
							$strUrl = null;
						}
						else
						{
							// The existing switch ($objNextPage->type) block
						}

```

aschempp: I think early return is better and easier to read than yet another level of [cyclomatic_complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity). The method/case is already long enough and hard to understand whats happening where …

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -436,6 +436,23 @@ protected function doReplace($strBuffer, $blnCache)
...
+						switch (strtolower($elements[0]))
+						{
+							case 'link_title':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strTitle);
+								break 2;
+
+							case 'link_name':
+								$arrCache[$strTag] = StringUtil::specialcharsAttribute($strName);
+								break 2;
+						}
```
leofeyer:
```suggestion
						if ($elements[0] == 'link_title')
						{
							$arrCache[$strTag] = StringUtil::specialcharsAttribute($strTitle);
							break 2;
						}

						if ($elements[0] == 'link_name')
						{
							$arrCache[$strTag] = StringUtil::specialcharsAttribute($strName);
							break 2;
						}
```

Please don't use a `switch` construct here and do not use `strtolower()`, either. Insert tags are not case-insensitive.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: IMO `break 2` and the duplicated code is way harder to unserstand.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
aschempp: you know this is a 1:1 copy of the lines below, right? Including the `lowercase` and switch-case?

leofeyer: The lines below are no early returns, so you cannot compare the two cases.

Are you aware that you have used complexity as an argument against @ausi's suggestion above while at the same time you refuse to reduce the complexity here? That does not make too much sense IMHO. 🤷‍♂️

ausi (approved):

PR #3916, opened 2022-01-11
Fixed tl_page permissions for routing fields

aschempp:
I forgot to add `exclude => true` to the new routing fields, they would always be available to all users.

The label change is necessary for the field name to correctly show in permissions, otherwise only the first character of the translation would be shown.

Regarding `type` and `requireItem` fields, they are only reordered for the permission to closely match to the field order in the default palettes.

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/tl_page.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/tl_page.xlf
@@ -8,17 +8,23 @@
       <trans-unit id="tl_page.title.1">
         <source>Please enter the page name.</source>
       </trans-unit>
+      <trans-unit id="tl_page.type.0">
+        <source>Page type</source>
+      </trans-unit>
+      <trans-unit id="tl_page.type.1">
+        <source>Please choose the page type.</source>
+      </trans-unit>
       <trans-unit id="tl_page.alias.0">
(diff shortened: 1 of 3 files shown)


Toflar (dismissed):

fritzmg (dismissed):

core-bundle/src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -473,7 +473,7 @@
       <trans-unit id="tl_page.legacyRouting">
         <source>Disable legacy routing mode to enable this configuration.</source>
       </trans-unit>
-      <trans-unit id="tl_page.routeConflicts">
+      <trans-unit id="tl_page.routeConflicts.0">
         <source>Route conflicts</source>
       </trans-unit>
       <trans-unit id="tl_page.conflictingAlias">
```
leofeyer:
```suggestion
      <trans-unit id="tl_page.routeConflicts.1">
```

Should we adjust the ID here, too, then?

aschempp: Changed in d72f3aadd100b1fff875a91acad929bf4421eb5b

leofeyer (approved):

PR #3917, opened 2022-01-11
Update maintenance response and add to preview endpoint

aschempp:
Fixes https://github.com/contao/contao/pull/3737#discussion_r774685376

I have also added the maintenance response to the preview endpoint. I've already run into situations where I was deploying an update to an installation, and the front end as well as the back end is "locked" through the maintenance file. That's correct, since I'm updating Composer dependencies and the application most likely will not work. But the preview endpoint to the front end was still working, which makes no sense to me.


Toflar (dismissed):

fritzmg (approved):

Toflar (approved):

leofeyer (changes requested):
How am I supposed to preview my website now if the maintenance mode is enabled?

Toflar:
You cannot. This is the system maintenance mode!

leofeyer (approved):

PR #3918, opened 2022-01-11
Fix root permission checks (including paste button)

aschempp:
This reverts some invalid changes from https://github.com/contao/contao/pull/3391

`$this->root` is always initialized in tree mode, therefore we must check the original value not the internal one.
I've gone through all changes of `$this->root` in #3391 and reverted what I saw unrelated to the parent trail feature.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -1718,7 +1718,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 	public function move()
 	{
 		// Proceed only if all mandatory variables are set
-		if ($this->intId && Input::get('sid') && (!$this->root || !\in_array($this->intId, $this->root)))
+		if ($this->intId && Input::get('sid') && (!($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'] ?? null) || !\in_array($this->intId, $this->root)))
 		{
 			$objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=? OR id=?")
 									 ->limit(2)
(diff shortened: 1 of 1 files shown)


Toflar (changes requested):
I don't like the changes even if they fix the problem.

The reason why I removed all `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']` occurrences and used `$this->root` is because I did not understand why at some occasions we use one and at some the other. With this PR you bring back this mess and in a few months, nobody will remember why it was again.

If we were to build the `DC_Table` from scratch, the config would come from the outside and not change somewhere in between based on superglobals. That's why I wanted to have only one occurrence of `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']`.

We can keep this by introducing another variable on `$dc` but please, let's not merge this as is. It's absolutely horrible.

aschempp:
We're accessing a million superglobals inside DC_Table… why would this be any other?

Toflar:
Because we're improving things step by step. This must be improved. It took me hours to figure out where the stuff was coming from and we're not going back there.

aschempp:
🤷 feel free to provide a fix then, I wouldn't know where to start

Toflar:
I cannot fix it as I did not debug the original issue. Can you explain why it works when you access the super globals but not when `$this->root`? It should be equal?

aschempp:
> Can you explain why it works when you access the super globals but not when `$this->root`? It should be equal?

Initial issue description:

> `$this->root` is always initialized in tree mode, therefore we must check the original value not the internal one.

Toflar:
> `$this->root` is always initialized in tree mode, therefore we must check the original value not the internal one.

See, and this is exactly the problem! `$this->root` **is meant to be the representation** of `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']`. Why would checking for `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']` instead of `$this->root` make any difference?
If it does - which is apparently the case - then this must be fixed instead of going back to accessing `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']`.

fritzmg:
There is a much simpler solution: https://github.com/contao/contao/pull/3919 😁

aschempp:
`$this->root` holds the nodes the tree is build from. If you do not have limited permission, it holds the root nodes of the database (`pid = 0`). It is required to build the tree from. It holds the array to start the tree from (therefore `root`!), not the permissions you have. https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/drivers/DC_Table.php#L6196-L6210

Toflar:
That's what I meant by "then we need a second variable". I've extracted all the logic properly into `initRoots()` so it's handled in one single place. Not in 20. `initRoots()` is supposed to handle all the settings from the DCA in order for the rest of the methods to check on the variables rather than any `$GLOBALS` array.

aschempp:
`initRoots` contains 90% code that was previously in `__construct`, where it is called. The other 10% are new. Nothing was spread across DC_Table. But that's not really the points, right=
As I said, feel free to do a PR that _improves_ something on the general concept. This PR _fixes_ the existing bug by reverting to what previously worked and was actually unrelated to #3391. I think introducing a local variable will possibly break because the superglobal can be changed at runtime, but thats a problem of the future PR 😉

Toflar:
Marking up for discussion because I don't agree with the "I fix stuff by ignoring whatever concept anybody else came up with, please do it yourself if you want a different fix" mentality. That's not how things work around here.

fritzmg:
Just to recap for my own understanding: `$GLOBALS['TL_DCA'][$table]['list']['sorting']['root']` works as a filter (dynamically set according to permissions/mounts) and `$this->root` will be the final array of the roots that the tree view should actually build upon, correct? (in <=4.12 I mean)

ausi:
As discussed with myself just now 😅 I want to take a look at this 👍

ausi:
> I want to take a look at this

Done: #3927

leofeyer:
See #3927

PR #3919, opened 2022-01-12
Fix rootPaste not being enabled for tl_page

fritzmg:
This fixes not being able to insert new website roots in `4.x` (supersedes #3918).

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_page.php
+++ b/core-bundle/src/Resources/contao/dca/tl_page.php
@@ -83,7 +83,8 @@ $GLOBALS['TL_DCA']['tl_page'] = array
 			'showRootTrails'          => true,
 			'icon'                    => 'pagemounts.svg',
 			'paste_button_callback'   => array('tl_page', 'pastePage'),
-			'panelLayout'             => 'filter;search'
+			'panelLayout'             => 'filter;search',
+			'rootPaste'               => true,
 		),
 		'label' => array
 		(


aschempp (changes requested):
This is not the correct solution. It would mean we need `rootPaste` for **every single DCA** in TREE mode, which basically makes the setting obsolete. It also enables root paste regardless of permissions as far as I know.

fritzmg:
> It also enables root paste regardless of permissions as far as I know.

No, the permissions for the paste button itself are still checked by the `paste_button_callback`.

> It would mean we need rootPaste for every single DCA in TREE mode

No, only for ones that allow you to have multiple roots (like the site structure). Inserting the first root always works.

I think this is the originally intended use case and the fact that you didn't have to enable it in the past was a bug. That's how I interpret the current code base (by current I mean also 4.9/4.12). May be @leofeyer can shed some light on that.

aschempp:
The `paste_button_callback` checks the visibility of the button, not the permission for the DCA operation.

Even [the documetation](https://docs.contao.org/dev/reference/dca/list/#sorting) says exactly that:

> Override disabling paste buttons to root, if root is set. (default: false)

**if root is set**. In our case, root is not set if you're an admin, so `rootPaste` must not be set for it to work.

leofeyer:
I can only tell you that the `rootPaste` flag exists so you can allow pasting at root level even if `$this->root` is empty. I cannot say whether it is more a correct fix than #3918 or not.

fritzmg:
> I can only tell you that the rootPaste flag exists so you can allow pasting at root level even if $this->root is empty.

It allows you to paste into the root node, if `$this->root` is _not_ empty.

aschempp:
> It allows you to paste into the root node, if `$this->root` is _not_ empty.

No. It allows you to paste into the root node, if `$GLOBALS['TL_DCA']['table']['list']['sorting']['root']` is not empty! The DCA does not know about `$this->root`.

PR #3924, opened 2022-01-13
Show page name in duplicate alias error

### LINKED ISSUE #3903: Should we show the conflicting pages?
### Description

https://github.com/contao/contao/blob/df10500be0123729bcd59736a870229df00c6ce9/core-bundle/src/EventListener/DataContainer/PageUrlListener.php#L270

The return value is the which page conflicts, should we display that? (Follow-up on #3843)

aschempp:
Fixes https://github.com/contao/contao/issues/3903

Bildschirmfoto 2022-01-13 um 17 14 20


leofeyer:
According to #3903, the error message should be _This alias already exists in page "Foobar" (ID 12)_. The absolute URL will hide the page name and ID if it gets longer.

aschempp:
hmm, but that message might be incorrect. Page ID 12 might have a different alias, but the route path compiled from the page type generates the same URL 🤔

aschempp:
@contao/developers any other opinions on the message?

Toflar:
Marking this `up for discussion` because it looks like something which can be resolved a lot quicker in a call.

leofeyer:
As discussed in the Contao call, we want to use `This alias leads to a conflict with page "News NEU" (ID 123)`.

Toflar (dismissed):
Fine for me. Just as a sidenote for the next time. For things like this you would usually add static methods on the exception. E.g. `DuplicateAliasException::createWithPageModel()`. But that's not stopping this one from being merged as is :)

aschempp:
> Fine for me. Just as a sidenote for the next time. For things like this you would usually add static methods on the exception. E.g. `DuplicateAliasException::createWithPageModel()`. But that's not stopping this one from being merged as is :)

Yeah I've seen that. But Symfony does not (always) do that either, e.g. for `AccountStatusException` 🤷

leofeyer (dismissed):

m-vo (approved):

PR #3927, opened 2022-01-13
Explicitly set rootPaste, deprecate implicit rootPaste

ausi:
Alternative to #3918

The current behavior is:
- If `list.sorting.root` is not set, pasting (and creating) at the root level is allowed
- If `list.sorting.root` is set, pasting at the root level is **NOT** allowed
- Unless `list.sorting.rootPaste` is set to `true` in which case pasting at the root level is always allowed

AFAIK the reason for this is that `list.sorting.root` was used to differentiate between admins (not set) and users (set to the configured mounts array).
`rootPaste` was only added as an afterthought recently (😉) in https://github.com/contao/core/issues/1312

This pull request changes the behavior by making `rootPaste` always take precedence and keeping BC by still enabling it if `list.sorting.root` is missing (together with a deprecation notice).

_BTW, `rootPaste` was only added for @aschempp it seems and was not used by the core until now ☺️_

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -185,6 +185,12 @@ abstract class DataContainer extends Backend
 	protected $visibleRootTrails = array();
 
 	/**
+	 * If pasting at root level is allowed (permissions)
+	 * @var bool
+	 */
+	protected $rootPaste = false;
+
+	/**
 	 * WHERE clause of the database query
 	 * @var array
 	 */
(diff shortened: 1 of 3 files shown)


leofeyer:
> _BTW, `rootPaste` was only added for @aschempp it seems and was not used by the core until now ☺️_

Do the changes from this PR allow to accomplish the same without `rootPaste`? Then we could deprecate the option.

ausi:
> Do the changes from this PR allow to accomplish the same without `rootPaste`?

No. This PR makes `rootPaste` more explicit so that it can be disconnected from the `root` setting in the future.
This is correct IMO because if you are allowed to create an entry at the root level should not depend on the fact that the root nodes are filtered or not.

m-vo (approved):
Code looks correct to me, I cannot say too much about the concept.

leofeyer:
Thank you @ausi.

fritzmg:
@ausi does the wording in the [documentation](https://docs.contao.org/dev/reference/dca/list/#sorting) need to be adjusted too?

ausi:
Yes: https://github.com/contao/docs/pull/932

PR #3929, opened 2022-01-14
Adjust the SERP preview formatting

leofeyer:
This PR adjusts the SERP preview formatting according to the new info field style added in #3926.

**Before**

**After**


Toflar (dismissed):

aschempp:
I would actually like the dotted line better for the other field 😂 also because it does not lower the contrast. Not sure if it works well for the small field though 🤷

fritzmg:
I also preferred the original version, since it's supposed to preview how a Google search result would look like. It's imho not related to the other field at all and therefore does not need any adjustments.

bytehead:
I agree with @fritzmg.

PR #3930, opened 2022-01-14
Resolve page model in request

aschempp:
Fixes the discussion in https://github.com/contao/contao/pull/3838#discussion_r780548609

The listener also automatically checks access to the page, which makes the whole implementation in https://github.com/contao/contao/pull/2015/files#diff-84c54a7b63df1afc1990d4aea951a7a696d445b16a2c4937591831a96f5d1ea4R47-R86 obsolete (and much smaller thanks to our security voter 😎).


richardhj:
Nah that's smart 👍

fritzmg (commented):
So henceforth `$request->attributes->get('pageModel')` would be the official way of retrieving the `PageModel` of the current page?

What about

https://github.com/contao/contao/blob/6d19cb4cf23502b05e4b878146dca722f2e1367a/core-bundle/src/Asset/ContaoContext.php#L83-L112

This should be changed as well, shouldn't it?

Also, should we give the listener a default higher priority to ensure the `pageModel` attribute is resolved for any other request listener?

aschempp:
> What about
> 
> https://github.com/contao/contao/blob/6d19cb4cf23502b05e4b878146dca722f2e1367a/core-bundle/src/Asset/ContaoContext.php#L83-L112
> 
> This should be changed as well, shouldn't it?

Yes, I missed that.
 
> Also, should we give the listener a default higher priority to ensure the `pageModel` attribute is resolved for any other request listener?

You can't get that much higher, because we need to be after Routing and Security (prio 8). So we could go somewhere between 0 and 7, and I don't really see a point because it should be relevant for after-request only.

fritzmg:
> and I don't really see a point because it should be relevant for after-request only.

These `RequestEvent` listeners of the Contao Core Bundle need access to the page model:

* https://github.com/contao/contao/blob/724fd43e33dc8bcaa720a71142bb521688a1c71b/core-bundle/src/EventListener/ServiceUnavailableListener.php#L44-L54
* https://github.com/contao/contao/blob/724fd43e33dc8bcaa720a71142bb521688a1c71b/core-bundle/src/EventListener/Security/TwoFactorFrontendListener.php#L65-L70

So this new listener needs to have a higher priority than these two at least.

m-vo:
Just a thought (and no idea about BC): would it be feasible that one key only ever contains the `PageModel` and a different one the ID? So that - no matter you are running before the listener - there won't ever be the subtle bug of it not being resolved yet and therefore having the wrong type?

aschempp:
> Just a thought (and no idea about BC): would it be feasible that one key only ever contains the `PageModel` and a different one the ID? So that - no matter you are running before the listener - there won't ever be the subtle bug of it not being resolved yet and therefore having the wrong type?

not possible without a BC break, and imho not necessary.

> > and I don't really see a point because it should be relevant for after-request only.
> 
> These `RequestEvent` listeners of the Contao Core Bundle need access to the page model:
> 
> * https://github.com/contao/contao/blob/724fd43e33dc8bcaa720a71142bb521688a1c71b/core-bundle/src/EventListener/ServiceUnavailableListener.php#L44-L54
> * https://github.com/contao/contao/blob/724fd43e33dc8bcaa720a71142bb521688a1c71b/core-bundle/src/EventListener/Security/TwoFactorFrontendListener.php#L65-L70
> 
> So this new listener needs to have a higher priority than these two at least.

Technically right. Appart from the fact that both these listeners only ever run on the main request, and the main request already has a page model. But it won't harm to change it.

core-bundle/src/Asset/ContaoContext.php
```diff
@@ -23,14 +22,12 @@
 class ContaoContext implements ContextInterface
 {
     private RequestStack $requestStack;
-    private ContaoFramework $framework;
     private string $field;
     private bool $debug;
 
-    public function __construct(RequestStack $requestStack, ContaoFramework $framework, string $field, bool $debug = false)
```
m-vo, line 30: IMHO this is a BC break, the constructor is not marked as `@internal` and the class not excluded from our BC promise.

m-vo (changes requested):
LGTM besides one remark.

core-bundle/src/Asset/ContaoContext.php
aschempp: it is `@internal`?
https://github.com/contao/contao/blob/14bae73113064e76592035231799b562f29381b6/core-bundle/src/Asset/ContaoContext.php#L20

bytehead (dismissed):

core-bundle/src/Asset/ContaoContext.php
m-vo: Ah sorry, you are right. I missed the class doc block for some reason.

m-vo (dismissed):

fritzmg:
Since the listener checks the permissions now, we can remove

https://github.com/contao/contao/blob/2f1827b8a2cac27d2b86d7ba6337470217d6c60d/core-bundle/src/Resources/contao/controllers/FrontendIndex.php#L139-L160

core-bundle/src/EventListener/PageAccessListener.php
```diff
@@ -0,0 +1,97 @@
...
+            return $GLOBALS['objPage'];
+        }
+
+        if ($pageModel instanceof PageModel) {
+            return $pageModel;
+        }
+
+        $this->framework->initialize();
+
+        return $this->framework->getAdapter(PageModel::class)->findByPk((int) $pageModel);
```
fritzmg, line 95: Btw. the original logic in `AbstractFragmentController` and `ContaoContext` is a little different:

If no request is available or no `pageModel` attribute is available, then `$GLOBALS['objPage']` will be returned, if available (otherwise null).

The new version now always returns null, if no `pageModel` attribute is available. Not sure if that makes a significant difference though.

aschempp: good point. The _request's_ page model is actually empty. Maybe you should get the (global) page model from the master request if you want that?

leofeyer (approved):

fritzmg (approved):

PR #3931, opened 2022-01-14
Use target URL from redirect in route

aschempp:
Fixes the discussion in https://github.com/contao/contao/pull/3838/files#r775749282 that we need to duplicate the code handling for target page type in multiple places, yet again in the KnpMenu. Superseeds https://github.com/contao/contao/pull/2031 with a much simpler solution by just leveraging the Symfony redirect controller.

This means 2 things:
 - A page still has its route, as it previously did. E.g. a redirect page to `https://example.com/` can still be called in Contao through its domain and alias, and will then redirect. The redirect however happens directly through the Symfony RedirectController, because the target URL is already known. If the target URL cannot be generated (e.g. if a root page does not have a subpage), the original controller can still handle that case and throw an exception.

Since we now have the target URL in the route information, we can use it when **generating** a route. Which allows the route to generate non-internal links to pages! Instead of handling the two known types in every place (see initial link), we can simply to `$route->generate(RouteObjectInterface::OBJECT_BASED_ROUTE_NAME, [RouteObjectInterface::CONTENT_OBJECT => $pageModel]);`

**TODO:**
 - [ ] Requires https://github.com/contao/contao/pull/3930 to check for access permission to the redirect page
 - [ ] Not yet implemented the `ForwardPageController` just to get some feedback first

### DIFF
--- a/core-bundle/src/Controller/Page/RedirectPageController.php
+++ b/core-bundle/src/Controller/Page/RedirectPageController.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Controller\Page;
+
+use Contao\CoreBundle\InsertTag\InsertTagParser;
+use Contao\CoreBundle\Routing\Page\DynamicRouteInterface;
+use Contao\CoreBundle\Routing\Page\PageRoute;
+use Contao\CoreBundle\ServiceAnnotation\Page;
+use Contao\PageModel;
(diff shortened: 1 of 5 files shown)


Toflar (commented):
This looks really good to me!

aschempp:
Unfortunately, I realized (once again) that this concept does not work. The target URL when generating a navigation item is not necessarily the same a controller would render. For example the [`PageForward`](https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/pages/PageForward.php#L74) keeps query and URL parameters when redirecting, but we do not keep them in the navigation menu. Even more problematic is the fact that we might skip a page from appearing in the navigation (https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/modules/Module.php#L366), which would not work anymore with this concept. I guess we need a different solution 😞

PR #3934, opened 2022-01-15
Fix the permission check for preview links

aschempp:
There is no `$dc->activeRecord` in the `onload_callback` 😇

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
@@ -112,7 +112,8 @@ class PreviewLinkListener
                 case 'toggle':
                 case 'delete':
                 default:
-                    if ($dc->activeRecord->createdBy !== $userId) {
+                    $createdBy = $this->connection->fetchOne("SELECT createdBy FROM tl_preview_link WHERE id=?", [$dc->id]);
+                    if ($createdBy !== $userId) {
                         throw new AccessDeniedException(sprintf('Preview link ID %s was not created by user ID %s', $dc->id, $userId));
                     }
                     break;
--- a/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
@@ -113,6 +113,7 @@ class PreviewLinkListener
                 case 'delete':
                 default:
                     $createdBy = $this->connection->fetchOne('SELECT createdBy FROM tl_preview_link WHERE id=?', [$dc->id]);
+
                     if ($createdBy !== $userId) {
                         throw new AccessDeniedException(sprintf('Preview link ID %s was not created by user ID %s', $dc->id, $userId));
                     }
--- a/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
@@ -112,7 +112,7 @@ class PreviewLinkListener
                 case 'toggle':
                 case 'delete':
                 default:
-                    $createdBy = $this->connection->fetchOne("SELECT createdBy FROM tl_preview_link WHERE id=?", [$dc->id]);
+                    $createdBy = $this->connection->fetchOne('SELECT createdBy FROM tl_preview_link WHERE id=?', [$dc->id]);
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
```diff
@@ -112,7 +112,8 @@ public function createFromUrl(DataContainer $dc): void
                 case 'toggle':
                 case 'delete':
                 default:
-                    if ($dc->activeRecord->createdBy !== $userId) {
+                    $createdBy = $this->connection->fetchOne('SELECT createdBy FROM tl_preview_link WHERE id=?', [$dc->id]);
```
m-vo: Without casting, the comparison below will always be true. I'm not sure what should happen if `$userId` was set to `0` in L84.

```suggestion
                    $createdBy = (int) $this->connection->fetchOne('SELECT createdBy FROM tl_preview_link WHERE id=?', [$dc->id]);
                    
```

m-vo (changes requested):
Is there a reason why we did not use an entity for `tl_preview_links`? :grimacing:

aschempp:
> Is there a reason why we did not use an entity for `tl_preview_links`? 😬

Because it's not a table anyone should read or write from/to, especially not in the front end (where we usually use entities). There's no advantage in using an entity here.

m-vo:
> Because it's not a table anyone should read or write from/to, especially not in the front end (where we usually use entities). There's no advantage in using an entity here.

I don't get that point. Why can't we use entities for internal things? Type safety when getting values would be one of the benefits compared to [here](https://github.com/contao/contao/pull/3934#discussion_r785445810). :see_no_evil:

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
aschempp: I don't think that's true, because `$user->id` is a string as well. Obviously a type-safe comparison would be better, but we'd need to cast the values everywhere.

leofeyer:
I don't get the point, either. We have created entities for every new table so far (`tl_crawl_queue`, `tl_cron_job`, `tl_remember_me`, `tl_trusted_device`), so we should do the same for `tl_preview_links`.

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
m-vo: If `$user->id` is always a string, two things should probably be adjusted:

*  the `@property` annotation in `BackendUser` (or more specific `User`) says `string|integer`
* L84 of this class assigns an integer as fallback value

fritzmg: `$user->id` can be an integer - for instance if you manually created the object instance with a custom model.

In any case I would cast to `int` here for defensive programming.

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
```diff
@@ -112,7 +112,9 @@ public function createFromUrl(DataContainer $dc): void
...
+
+                    if ($createdBy !== $userId) {
```
leofeyer, line 117: The strict comparison will not work here, because the DB result will be a string and `$userId` might be an integer.

leofeyer (changes requested):

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
m-vo: We should IMHO cast both the DB result as outlined [here](https://github.com/contao/contao/pull/3934/files#r785445810) and the assignment in L84:
 `$userId = (int) ($user instanceof BackendUser ? $user->id : 0);`

leofeyer: See 8116179273ebda6ffca7ab6b316992d6e98624ff

leofeyer (approved):

core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
leofeyer: See See 8116179273ebda6ffca7ab6b316992d6e98624ff

PR #3935, opened 2022-01-15
Allow Flysystem v3

m-vo:
This PR introduces the ability to check if directories exist in the filesystem implementation. This was a limitation with the not-that-old Flysystem 2, but Flysystem 3 was released shortly because this was considered a [design flaw](https://blog.frankdejonge.nl/flysystem-3-0-0-is-released/). I probably should have went with my gut feeling and added these things no matter what in #3774, but then trusted them on this decision. :smile: 

Ok, not too late. Here are the changes:

* The `VirtualFilesystemInterface`, `DbafsManager` and `MountManager` now allow to differentiate between `fileExists()` and `directoryExists()`.
* The `VirtualFilesystemInterface` and `DbafsManager` also got a `has()` method as a shortcut to check if any resource exists. It's a convenient method, so it would only be 'necessary' for the virtual filesystem, but it allows for a more effective evaluation in the DBAFS manager, so that's why it's there as well.
* I explicitly required `league/flysystem` in `^2.0 || ^3.0` because the Symfony bundle with the transient dependency was only released with a new minor version which completely hides the real dependency we're using.

Installing v3 is completely optional and the reason for that is, that it unfortunately [requires PHP8](https://github.com/thephpleague/flysystem/issues/1405). For users of v2 there is a workaround in place: we list contents of the parent directory of the requested resource and - if an item is returned where the path matches - check the type of the item. Not as effective, but fully compatible.

In Contao 5 we can then require v3, remove the fallback and also get rid of some code in the `Dbafs` class that currently also uses the workaround to determine if a resource is a directory.


m-vo:
Code coverage is only reported for a single set of dependencies, but actually you would need a prefer lowest + edge run to see that in fact everything's covered. I don't know how to fix this. Or in general: how to properly test things that are discriminated by `method_exists` calls.

composer.json
```diff
@@ -30,9 +30,9 @@
         "ext-pcre": "*",
         "ext-pdo": "*",
         "ext-zlib": "*",
-        "composer-runtime-api": "^2.0",
         "ausi/slug-generator": "^1.1",
         "bacon/bacon-qr-code": "^2.0",
+        "composer-runtime-api": "^2.0",
```
leofeyer: Unrelated, please revert.

bytehead: Why was it in false order anyway?

leofeyer: The order is correct.

1. PHP version
2. PHP extensions
3. Composer runtime API
4. Packages

See https://github.com/ergebnis/composer-normalize.

m-vo: This is the output produced by the `composer-json` helper.

m-vo: see https://github.com/contao/monorepo-tools/issues/18

m-vo: reverted in d5729bdee2c95a3c376b3d1be395e05b4831330d

ausi (dismissed):

bytehead (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3937, opened 2022-01-16
Various small filesystem tweaks

m-vo:
This PR tweaks and fixes some small things in the filesystem implementation (followup to #3774):

* Services instantiated from the `Dbafs` class are now tagged with `kernel.reset` (the interface was already there, but the tag was missing).
* The VFS now enhances filesystem items read from the DBAFS with on-demand file metadata from the filesystem.<sup>*)</sup>
* `VirtualFilesystemInterface#listContents()` got a generic return type hint (docbloc only).

<sup>*)</sup> This means you will always get all information (like file size, mime type) when listing contents no matter what's the source. We do this by enhancing the `FilesystemItem` with closures that retrieve the data on access if they haven't been set already.

### DIFF
--- a/core-bundle/src/Filesystem/VirtualFilesystem.php
+++ b/core-bundle/src/Filesystem/VirtualFilesystem.php
@@ -257,7 +257,7 @@ class VirtualFilesystem implements VirtualFilesystemInterface
      */
     private function doListContents(string $path, bool $deep, int $accessFlags): \Generator
     {
-        // Read from DBAFS, but enhance result with file metadata on demand
+        // Read from DBAFS but enhance result with file metadata on demand
         if (!($accessFlags & self::BYPASS_DBAFS) && $this->dbafsManager->match($path)) {
             /** @var FilesystemItem $item */
             foreach ($this->dbafsManager->listContents($path, $deep) as $item) {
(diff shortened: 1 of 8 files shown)


core-bundle/src/Filesystem/VirtualFilesystem.php
```diff
@@ -269,10 +285,12 @@ private function doListContents(string $path, bool $deep, int $accessFlags): \Ge
...
         /** @var FilesystemItem $item */
         foreach ($this->mountManager->listContents($path, $deep) as $item) {
+            $path = $item->getPath();
+
             yield $item
-                ->withPath(Path::makeRelative($item->getPath(), $this->prefix))
+                ->withPath(Path::makeRelative($path, $this->prefix))
                 ->withExtraMetadata(
-                    fn () => $this->dbafsManager->getExtraMetadata($item->getPath())
+                    fn () => $this->dbafsManager->getExtraMetadata($path)
```
m-vo: In case you're wondering: These lines are pure refactoring (no change in what they do). This way `getPath()` is not executed twice and it's easier to understand which path we are meaning, namely the original one before adjusting it.

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

Issue #686, opened 2015-03-31
Drop support of empty ptable for content element.


dmolineus:
Contao 3 introduced the dymanic parent table feature for content elements. For backwards compatibility an empty ptable column of tl_content is threated as `tl_article`, so additional conditions to check this were included in the core.

I think Contao 4 would be a good point to drop this backward compatibility. 3rd party extensions would have to set the ptable always.

An migration script (from Cto 3 to 4) could change any empty ptable value of `tl_content` to `tl_article`.

leofeyer:
Changed in #4653.

PR #3939, opened 2022-01-17
Fix duplicating child records with dynamic ptable

dmolineus:
Use current table for parent table statement. The configured dynamic ptable might differ from the current table.

## Scenario

1. I have a data container setup where a data container `tl_child` has a `dynamicPtable = (tl_root|tl_category)`
2. The hierarchy of the data containers is `tl_root => tl_category => tl_child` 
3. I want to duplicate a row of `tl_root`

## Expected result

Each children in `tl_category` and `tl_child` should be copied no matter which parent table the entry in `tl_child` has

## Current result

Only rows of `tl_child` matching the configured `ptable` are copied. Furthermore wrong children might be assigned if `pid`s are matching for different `ptable`s.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -1057,7 +1057,7 @@ class DC_Table extends DataContainer implements \listable, \editable
 					$cond = ($ptable == 'tl_article') ? "(ptable=? OR ptable='')" : "ptable=?";
 
 					$objCTable = $this->Database->prepare("SELECT * FROM $v WHERE pid=? AND $cond" . ($this->Database->fieldExists('sorting', $v) ? " ORDER BY sorting" : ""))
-												->execute($id, $ptable);
+												->execute($id, $table);
 				}
 				else
 				{
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -1629,11 +1629,10 @@ class DC_Table extends DataContainer implements \listable, \editable
 			// Consider the dynamic parent table (see #4867)
 			if ($GLOBALS['TL_DCA'][$v]['config']['dynamicPtable'])
 			{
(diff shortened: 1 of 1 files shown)


aschempp:
the described configuration is confusing to me. Are the children related to the categories at all? If not, how can the described hierarchy be correct?

dmolineus:
> the described configuration is confusing to me. Are the children related to the categories at all? If not, how can the described hierarchy be correct?

Children may belong to directly to `tl_root` or to `tl_category`

aschempp:
thanks for clarifying, I think point 1 was confusing then (was `dynamicPtable = (tl_root|tl_child)`, should be `dynamicPtable = (tl_root|tl_category)` right?)

Regarding your code, I think this makes sense even though it's confusing. But you should also remove the `$ptable` assignment and maybe add a comment why this approach works (recursive calls etc.)

fritzmg:
The change generally makes sense to me. `$table` should already always be the correct parent table. An additional comment would be good though.

> But you should also remove the $ptable assignment and maybe add a comment why this approach works (recursive calls etc.)

The variable is still used for the `tl_article` backwards compatibility (where `tl_content.ptable` can be empty, when upgrading from Contao 2 or 3, apparently).

dmolineus:
> thanks for clarifying, I think point 1 was confusing then (was dynamicPtable = (tl_root|tl_child), should be dynamicPtable = (tl_root|tl_category) right?)

Yeah, you're right.

> But you should also remove the $ptable assignment and maybe add a comment why this approach works (recursive calls etc.)

It was deferred to be removed in Contao 5 and not Contao 4, see https://github.com/contao/contao/issues/686

aschempp:
The check for `tl_article` should use `$table` as well, shouldn't it? Otherwise the SQL query _might_ use a different value than the check.

dmolineus:
> The check for tl_article should use $table as well, shouldn't it? Otherwise the SQL query might use a different value than the check.

You're right, fixed with b78230d.

aschempp (approved):
I'm only missing a comment to describe why we're not using `ptable` 🙃

fritzmg (approved):

leofeyer:
Thank you @dmolineus.

PR #3942, opened 2022-01-17
Allow to embed fragment controllers in Twig templates

m-vo:
This PR introduces an easy way to embed fragments controllers in your Twig templates as a better alternative to insert tags:

```twig
    {{ render(contao_module(42)) }}
    {{ render(contao_content_element(5)) }}
```

We reuse the functionality that Symfony already provides with the `render` [functions](https://symfony.com/doc/current/reference/twig_reference.html#render), but instead of using the `path` or `controller` functions, you can use the `contao_module` and `contao_content_element` functions to get a `ControllerReference` from a model ID.

This is an early draft, that showcases the general idea. This would currently only work for fragment controllers. We might also want a fail tolerant mode, because exceptions cannot be worked with inside templates.

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -945,6 +945,12 @@ services:
         arguments:
             - '@contao.image.studio.figure_renderer'
 
+    contao.twig.http_kernel_runtime:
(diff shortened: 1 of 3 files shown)


aschempp:
As I already said to @m-vo, this is wrong because a fragment decides its own renderer, it must not be done in the template. There is no advantage on using the built in render functions.

PR #3947, opened 2022-01-18
Remove the routing error log in breadcrumb module

aschempp:
This error message is pretty annoying and useless. I have a breadcrumb module on error pages, so I get an error message whenever the current URL is build. It correctly generates an empty URL though and does not link the breadcrumb item.

I think the error message does not make sense for the breadcrumb, because it might also appear on non-routable pages and pages with parameters. It still makes sense in the navigation and other places that should never be generated.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleBreadcrumb.php
@@ -255,8 +255,6 @@ class ModuleBreadcrumb extends Module
 		}
 		catch (ExceptionInterface $exception)
 		{
-			System::log('Unable to generate URL for page ID ' . $pageModel->id . ': ' . $exception->getMessage(), __METHOD__, TL_ERROR);
-
 			return '';
 		}
 	}


bytehead (approved):

fritzmg (approved):

Toflar:
What about replacing it with a comment so we don't re-add in the future?

SeverinGloeckle:
There's probably more to it than just removing the error message and more than just error pages are affected - see https://github.com/contao/contao/pull/3540

olli1770:
The error is still present in 4.13.2, so breadcrumb is is unuseable at the moment.
Is there still somebody working on it?

leofeyer:
Which error are you referring to?

olli1770:
This one: Breadcrumb on 404 page leads to system log error entry #4194
https://github.com/contao/contao/issues/4194
If I activate an error-page 404 the log is floaded with entries, if I use the breadcrumb-feature.

leofeyer:
Then please comment in #4194 and not here.

PR #3953, opened 2022-01-18
Fix the PackageUtil class

ausi:
This reverts #3946

### DIFF
--- a/core-bundle/src/Util/PackageUtil.php
+++ b/core-bundle/src/Util/PackageUtil.php
@@ -12,7 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Util;
 
-use PackageVersions\Versions;
+use Composer\InstalledVersions;
 
 /**
  * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0; use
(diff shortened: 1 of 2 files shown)


core-bundle/src/Util/PackageUtil.php
```diff
@@ -60,7 +59,7 @@ public static function parseVersion(string $version): string
     {
         trigger_deprecation('contao/core-bundle', '4.13', 'Using the PackageUtil::parseVersion() method has been deprecated and will no longer work in Contao 5.0. Use the Composer\InstalledVersions class instead.');
 
-        return ltrim(strstr($version, '@', true), 'v');
+        return ltrim(strstr($version.'@', '@', true), 'v');
```
leofeyer, line 62:
```suggestion
        return ltrim($version, 'v');
```

We could simplify this, because `InstalledVersions::getPrettyVersion()` will never return the `@<hash>` part.

ausi: But the method is public so it has to support inputs that previously worked.

leofeyer: You are right of course. 🦊

m-vo (approved):

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #3957, opened 2022-01-18
Upgrade to symfony/filesystem 5.4 and migrate webmozart/path-util usage

m-vo:
This PR requires the `symfony/filesystem` component in version 5.4 and drops the abandoned webmozart/path-util for Contao 4.9.

see https://github.com/contao/image/issues/86#issuecomment-1002958667

### DIFF
--- a/core-bundle/composer.json
+++ b/core-bundle/composer.json
@@ -129,10 +129,7 @@
         "lexik/maintenance-bundle": "To put the application into maintenance mode"
     },
     "extra": {
-        "contao-manager-plugin": "Contao\\CoreBundle\\ContaoManager\\Plugin",
-        "symfony": {
-            "require": "^4.4"
-        }
+        "contao-manager-plugin": "Contao\\CoreBundle\\ContaoManager\\Plugin"
     },
     "autoload": {
         "psr-4": {
--- a/core-bundle/composer.json
+++ b/core-bundle/composer.json
@@ -81,7 +81,7 @@
         "symfony/dom-crawler": "4.4.*",
         "symfony/event-dispatcher": "4.4.*",
         "symfony/expression-language": "4.4.*",
-        "symfony/filesystem": "4.4.*",
+        "symfony/filesystem": "^5.4",
         "symfony/finder": "4.4.*",
         "symfony/framework-bundle": "4.4.*",
         "symfony/http-foundation": "4.4.*",
@@ -103,7 +103,6 @@
         "twig/twig": "^2.7",
         "ua-parser/uap-php": "^3.9",
         "webignition/robots-txt-file": "^3.0",
-        "webmozart/path-util": "^2.2",
         "wikimedia/less.php": "^1.7"
     },
     "conflict": {
--- a/calendar-bundle/composer.json
+++ b/calendar-bundle/composer.json
@@ -37,10 +37,7 @@
         "symfony/phpunit-bridge": "4.4.*"
     },
     "extra": {
-        "contao-manager-plugin": "Contao\\CalendarBundle\\ContaoManager\\Plugin",
-        "symfony": {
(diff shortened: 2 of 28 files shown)


bytehead:
I don't know if 4.9 should use Symfony 5 components?

bytehead (commented):
I don't know if 4.9 should use Symfony 5 components?

m-vo:
> I don't know if 4.9 should use Symfony 5 components?

The filesystem component is a small component that does not introduce any dependencies (besides polyfills), so it's IMHO the best option. For reference, also see the discussion in https://github.com/contao/image/issues/86#issuecomment-1002958667.

m-vo:
I had to remove the `extra.symfony` sections from the `composer.json` files, because they were restricting the Symfony version to 4.4 for all components. IIRC this was only there for the flex plugin to get faster installs, which would now be obsolete anyways thanks to Composer 2.

core-bundle/composer.json
```diff
@@ -81,7 +81,7 @@
         "symfony/dom-crawler": "4.4.*",
         "symfony/event-dispatcher": "4.4.*",
         "symfony/expression-language": "4.4.*",
-        "symfony/filesystem": "4.4.*",
+        "symfony/filesystem": "^5.4",
```
ausi, line 84:
```suggestion
        "symfony/filesystem": "4.4.* || 5.4.*",
```
I would have just widened this requirement and kept using the webmozart lib, to maximize compatibility with extensions. But I’m fine either way, @contao/developers WDYT?

m-vo: IMHO we must get rid of the abandoned library anyways in the remaining lifetime of 4.9 (and so do all extension developers), so I'd vote for removing it.

ausi (approved):

bytehead (approved):

leofeyer (changes requested):
We cannot do this, I'm afraid, since there have been breaking changes between version 4 and 5 of the component.

* `Filesystem::dumpFile()` and `appendToFile()` don't accept arrays anymore.
* The minimum PHP version has been raised from PHP 7.1 to PHP 7.2.
* The method arguments now use type hints.

Also, everyone requiring `"symfony/filesystem": "4.4*"` would not be able to update their installation anymore.

core-bundle/composer.json
aschempp: we never got rid of abandoned libraries in any LTS. We still use lexik and other stuff. There's no reason to change this 🤷

leofeyer: Optionally allowing version 5.4 in this special case would be OK for me. Then our users could require `symfony/filesystem` in version 5 if no other extension explicitly requires version 4.

m-vo:
> Filesystem::dumpFile() and appendToFile() don't accept arrays anymore.
> The method arguments now use type hints.

Is there anything that needs to be adjusted in this PR? 

> Also, everyone requiring "symfony/filesystem": "4.4*" would not be able to update their installation anymore.

This is correct. But it's the same as anyone requiring dbal in `2.*` with 4.13. 

It's not about pushing the boundaries but taking steps where needed.

ausi:
> The minimum PHP version has been raised from PHP 7.1 to PHP 7.2.

Contao 4.9 is already PHP 7.2 https://github.com/contao/contao/blob/7b8aada0c4ea930baba6c78dab4bcd08ee52debf/composer.json#L18

m-vo:
> Contao 4.9 is already PHP 7.2

Just realized this as well. This change only affects `contao/image`.

ausi:
Pull request for widening the constraint for `contao/filesystem`: #3966

I would like to keep this pull request open as we have to merge this anyway as soon as the path-util library is incompatible with a new PHP version or has a security issue.

PR #3960, opened 2022-01-18
Refactor HTML into templates in PageRoutingListener

m-vo:
The `PageRoutingListener` already uses templates, so we should IMHO avoid composing HTML in it.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PageRoutingListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PageRoutingListener.php
@@ -19,7 +19,6 @@ use Contao\CoreBundle\Routing\Page\PageRoute;
 use Contao\CoreBundle\ServiceAnnotation\Callback;
 use Contao\DataContainer;
 use Contao\PageModel;
-use Contao\StringUtil;
 use Twig\Environment;
 
 class PageRoutingListener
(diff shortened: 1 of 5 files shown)


leofeyer:
TBH, this seems a little over-engineered to me. 🙈

bytehead:
> TBH, this seems a little over-engineered to me.

In this little narrow case maybe. But if we don't use templates everywhere, we will have a lot of classes, listeners or services containing html code in the future. 🤷‍♂️ 

Strictly separating between controller and view is not over-engineering imho.

leofeyer:
That's right.

SeverinGloeckle:
Extracting the HTML from the listener is a good optimisation, in my opinion.

Adding an extra template just for the composition of the route path however might actually be a bit excessive. 
Maybe you could simply compose the route path from the chunks directly within the existing `be_route_conflicts.html.twig` template?

Even if a route path is be rendered anywhere else in the future, it's likely that the markup will be different there (e.g. no `tl_tip` class).

m-vo:
> Even if a route path is be rendered anywhere else in the future, it's likely that the markup will be different there (e.g. no tl_tip class).

That's exactly the point why extracting 'partials' or component templates makes sense to me. It forces you to think about generalizing your UI.

Because we expect route paths to be rendered identical everywhere they are used, it's a single template and not duplicated markup within `be_route_conflicts.html.twig` and `be_route_path.html.twig`.

SeverinGloeckle:
The concept certainly makes sense and points towards a good direction!

Whether it should really be broken down that far in this case seems debatable. 

However, in fact I overlooked the second template which also uses `_route_path.html.twig`. Thus, I tend to agree here as well, especially if a good solution is found for #3967, so that a larger number of templates remains manageable.

PR #3961, opened 2022-01-19
Gray out expired preview links

leofeyer:
Because the line-through with the red date looks like a misconfiguration that needs to be fixed.

**Before** 

**After**

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php
@@ -21,7 +21,6 @@ use Contao\DataContainer;
 use Contao\Image;
 use Contao\Input;
 use Contao\StringUtil;
-use Contao\UserModel;
 use Doctrine\DBAL\Connection;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpKernel\UriSigner;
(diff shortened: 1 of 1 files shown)


aschempp (changes requested):
you need to keep the handling for `createdBy`, otherwise it will only show the user ID

fritzmg:
The custom formatting shouldn't actually be necessary when using the proper DCA settings.

aschempp:
> The custom formatting shouldn't actually be necessary when using the proper DCA settings.

I couldn't get it to work 🤷

leofeyer:
> you need to keep the handling for `createdBy`, otherwise it will only show the user ID

The field is not even shown in the list?

leofeyer:
> The custom formatting shouldn't actually be necessary when using the proper DCA settings.

@fritzmg What exactly do you mean?

fritzmg:
> @fritzmg What exactly do you mean?

I mainly meant fetching the user from the database to get their name. However your PR removes that anyway (since the field isn't actually added to the list labels config).

m-vo:
> Because the line-through with the red date looks like a misconfiguration that needs to be fixed.

Fully agree on that and I like the new looks, but the grayed-out version will now trigger accessibility warnings due to low contrast, unfortunately.

leofeyer:
That's a different issue. If the CSS class `tl_gray` has to little contrast, we should fix it in a separate PR.

m-vo (approved):
In my opinion, this should too render a template. For example a reusable `label.html.twig` with semantic attributes `['title' => $arg, 'disabled' => true]`:
```twig
<span{% if disabled|default(false) %} class="tl_gray"{% endif %}>{{ title }}</span>
```

But not blocking merging as this is already an improvement.

aschempp:
You can sort the list by name, and then it is shown 🤷

PR #3969, opened 2022-01-19
Do not enable the maintenance mode for new pages

### LINKED ISSUE #3958: Maintenance mode
### Affected version(s)

4.13

### Description

If I create a new "Website root" the maintenance mode is enabled by default.

Is that intentional, I would prefer if it was disabled.

The Alert bell no longer shows the maintenance mode.

I would add the hint again and write the list of domains in maintenance mode in the alert window.

leofeyer:
Fixes #3958


aschempp (approved):

aschempp:
does not fix the alert bell though 🤔

m-vo (approved):

leofeyer:
I know. See https://github.com/contao/contao/issues/3958#issuecomment-1016611499.

PR #3971, opened 2022-01-19
Fix `Input::post()` for form data

### LINKED ISSUE #3433: insert tag {{post::*}} funktioniert manchmal nicht
**Affected version(s)**

Contao 4.9.15, 4.9.17, 4.9.20

**Description**

In den Contao Formular-Einstellungen habe ich einen insert tag {{post::betreff}} unter Formulardaten versenden -> Per E-Mail versenden -> in der Betreff Zeile stehen. Das Insert Tag soll das Feld aus dem Formular "betreff" mit in den Betreff der E-Mail eintragen. Wenn ich im Backend eingeloggt bin und ich teste das Formular im Frontend, ist der Betreff in der versendeten E-Mail enthalten. Wenn ich einen frischen Browser Chrome aufmache, dort das Formular versende, erscheint der Betreff nicht mehr in der E-Mail. Bleibe ich dann in Chrome, und versende das Formular nocheinmal, ist der Betreff plötzlich wieder in der E-Mail vorhanden. Das gleiche habe ich auch in den Browsern Edge und Opera getestet. Dort ist das gleiche Problem. LEider ist das auch so, das Bestätigungsmails von dem Modul Notification Center dann nicht versendet werden, wenn das insert tag {{post::betreff}} in der E-Mail fehlt.

fritzmg:
Fixes #3433

This issue was introduced in Contao 4.6.0 in https://github.com/contao/contao/commit/b2244f07a1d55f1d4e6d00103e9211b6b421a70e.

`Input::findPost` currently checks for `!$request->hasPreviousSession()` and if true, no value will be returned. This in turn however means that `Input::post(…)` or `{{post::*}}` will not return anything for the _current_ request, because at that point the `$_POST` variable for the form field has already been removed here

https://github.com/contao/contao/blob/f9b628281f9e3f6adfb3df8cb2d903a7c89ab269/core-bundle/src/Resources/contao/forms/Form.php#L225-L231

and thus only the `$_SESSION['FORM_DATA']` variable remains. And without a previously existing session (for which there will be none, when visiting the page with the form for the first time and no session was started otherwise), said data will not be returned though, because of the `!$request->hasPreviousSession()` check.

This PR fixes that by removing the `hasPreviousSession` check - as it breaks the logic used by `Contao\Form::compile` and `Contao\Input::findPost`.

// [edit](https://github.com/contao/contao/pull/3971#issuecomment-1016711132):

The check is already done (in the correct way) in our `LazySessionAccess` anyway, so no need to duplicate that for any code that accesses `$_SESSION`:

https://github.com/contao/contao/blob/f9b628281f9e3f6adfb3df8cb2d903a7c89ab269/core-bundle/src/Session/LazySessionAccess.php#L41-L50

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -985,7 +985,7 @@ class Input
 			return $_POST[$strKey];
 		}
 
-		// $_SESSION access is governed by LazySessionAccess (#3971)
+		// Do not check for $request->hasPreviousSession() and early return here (see #3971)
 		if (isset($_SESSION['FORM_DATA'][$strKey]))
 		{
 			return ($strKey == 'FORM_SUBMIT') ? preg_replace('/^auto_/i', '', $_SESSION['FORM_DATA'][$strKey]) : $_SESSION['FORM_DATA'][$strKey];
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -985,7 +985,7 @@ class Input
 			return $_POST[$strKey];
 		}
 
-		// $_SESSION access is governed by LazySessionAccess
+		// $_SESSION access is governed by LazySessionAccess (#3971)
 		if (isset($_SESSION['FORM_DATA'][$strKey]))
 		{
 			return ($strKey == 'FORM_SUBMIT') ? preg_replace('/^auto_/i', '', $_SESSION['FORM_DATA'][$strKey]) : $_SESSION['FORM_DATA'][$strKey];
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -985,14 +985,6 @@ class Input
 			return $_POST[$strKey];
 		}
 
-		$request = System::getContainer()->get('request_stack')->getMasterRequest();
-
-		// Return if the session has not been started before
-		if ($request === null || !$request->hasPreviousSession())
-		{
-			return null;
-		}
-
 		if (isset($_SESSION['FORM_DATA'][$strKey]))
 		{
 			return ($strKey == 'FORM_SUBMIT') ? preg_replace('/^auto_/i', '', $_SESSION['FORM_DATA'][$strKey]) : $_SESSION['FORM_DATA'][$strKey];
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -985,6 +985,7 @@ class Input
 			return $_POST[$strKey];
 		}
 
+		// $_SESSION access is governed by LazySessionAccess
 		if (isset($_SESSION['FORM_DATA'][$strKey]))
 		{
 			return ($strKey == 'FORM_SUBMIT') ? preg_replace('/^auto_/i', '', $_SESSION['FORM_DATA'][$strKey]) : $_SESSION['FORM_DATA'][$strKey];
--- a/core-bundle/src/EventListener/ClearSessionDataListener.php
+++ b/core-bundle/src/EventListener/ClearSessionDataListener.php
@@ -52,6 +52,7 @@ class ClearSessionDataListener
 
     private function clearLegacyAttributeBags(string $key): void
     {
+        // $_SESSION access is governed by LazySessionAccess
         if (!isset($_SESSION[$key])) {
             return;
         }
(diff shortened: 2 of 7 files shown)


aschempp:
I feel like this can't be right. We don't want to start/initiate a session if the user does not have one. If she does not have a previous session, there can't be any post data to look up. But maybe we need to additionally check for whether the current session was started?

fritzmg:
`Input::findPost` does not start/initiate a session, as far as I tested.

aschempp:
No, but wasn't your point that it does not work if the session was just started in this request (by submitting the form)? If the session was started in this very request, the request will not have a previous session but it will have session data.

fritzmg:
Yes, but the check is superfluous in any case.

fritzmg:
The check is already done (in the correct way) in our `LazySessionAccess` anyway, so no need to duplicate that for any code that accesses `$_SESSION`:

https://github.com/contao/contao/blob/f9b628281f9e3f6adfb3df8cb2d903a7c89ab269/core-bundle/src/Session/LazySessionAccess.php#L41-L50

aschempp (dismissed):

ausi (dismissed):

Toflar (changes requested):
Blocking this from an early merge because I would like to double check this when I'm back at work too.

Toflar (changes requested):
Changes look okay but in order to not re-add this issue (as it cannot really be covered by a unit test), can we at least replace it with a comment? Something along the lines of

```php
// Do not check for $request->hasPreviousSession() and early return here, see https://github.com/contao/contao/pull/3971
```

fritzmg:
> Changes look okay but in order to not re-add this issue (as it cannot really be covered by a unit test), can we at least replace it with a comment?

Sure, however we access `$_SESSION` in more places and the same applies there. Should we add a comment everywhere? Wouldn't hurt - because I forgot about our `LazySessionAccess` when initially working on this :)

ausi:
> Should we add a comment everywhere?

👍 these comments can then also be used in the 5.x branch to replace the direct access to `$_SESSION` with the session service.

fritzmg:
Indeed. What wording should we use? e.g. simply

```
// $_SESSION access is governed by LazySessionAccess
```

fritzmg:
I have added the comments now. However, I did not add the comments to _all_ instances of `$_SESSION` usage, as I found that to be too excessive. I more or less only added it to the first occurrence in a particular section of code.

ausi (dismissed):

Toflar (dismissed):

leofeyer (changes requested):
Please remove the repeated comments. They do not provide any added value IMHO.

Also, I think the fix is wrong. The point of b2244f07a1d55f1d4e6d00103e9211b6b421a70e was to not access `$_SESSION['FORM_DATA']` if the session has not been started before. Now the session will be started upon every call to `Input::post()` if `$_POST[$strKey]` is not set, which is the exact opposite of what we wanted back then. Why do we want it now?

fritzmg:
> Also, I think the fix is wrong. The point of b2244f0 was to not access $_SESSION['FORM_DATA'] if the session has not been started before. Now the session will be started upon every call to Input::post() if $_POST[$strKey] is not set

No, it will not, see the previous discussion or the intial (updated) description. Access to `$_SESSION` is governed by our `LazySessionAccess` which already makes the necessary (and correct) checks.

This misconception is exactly why we have dicussed that we should may be add these comments.

leofeyer:
Ok, I understand why the check can be removed. I don't agree with the comments, though.

fritzmg:
Do you want all comments removed? Except for the one where we removed the check (though I find that inconsistent).

leofeyer:
Yes please. The comment is really not very helpful and only at some occurrences of `$_SESSION`, so no added value here.

fritzmg:
> Yes please. The comment is really not very helpful and only at some occurrences of `$_SESSION`, so no added value here.

Done.

core-bundle/src/Resources/contao/library/Contao/Input.php
```diff
@@ -985,14 +985,7 @@ public static function findPost($strKey)
...
 
-		$request = System::getContainer()->get('request_stack')->getMasterRequest();
-
-		// Return if the session has not been started before
-		if ($request === null || !$request->hasPreviousSession())
-		{
-			return null;
-		}
-
+		// $_SESSION access is governed by LazySessionAccess (#3971)
```
leofeyer:
```suggestion
		// Do not check for $request->hasPreviousSession() and early return here (see #3971)
```

As suggested by @Toflar, so we find this issue again.

fritzmg: I already referenced the PR in my comment, but ok ;). Also this still applies to any `$_SESSION` access, not just this particular one in `Input::findPost`.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #3972, opened 2022-01-19
Add the require-checker tool

ausi:
I think this tool might not be so bad as I was able to find some bugs ☺️


bytehead (dismissed):
This is nice!

Toflar (dismissed):
Yey! That was on my list too :)

fritzmg (dismissed):

leofeyer (commented):
I am absolutely OK with this PR. I am just wondering about these cases:







The bundle will always include the library in a matching version, so why do we have to add a dependency as well? Are we only doing it for the tool or is it really wrong to omit it (because "don't rely on your dependencies' dependencies")?

m-vo:
> The bundle will always include the library in a matching version

Thought this as well, but actually that's not always the case. For Flysystem for instance, now a minor release of the bundle allows a new major release of the library (that's why I also added it in #3935). So it's not safe to assume these transitive requirements always match.

ausi:
> Are we only doing it for the tool or is it really wrong to omit it

IMO it is wrong to omit them, as they are not the same version contraints most of the time.
I was also a bit surprised how far off these version constraints were for some bundle/library combination.

bytehead (approved):

Toflar:
@leofeyer to answer your question, just think doctrine/doctrine-bundle and doctrine/dbal ;)

leofeyer:
Thank you @ausi.

Issue #3967, opened 2022-01-19
How should we organize our new templates?


m-vo:
### Description

We should discuss how we want to structure our Twig templates in the future. 

### Background info
Currently there are some old bundle templates and there will be Twig variants for all Contao templates. The latter is the way to go if we want to profit from our dynamic inheritance logic (which we do :smile:).

At this point, structuring in directories is not supported<sup>*)</sup>, or rather: they are collapsed as if they do not exist because our legacy template system behaved like this ever since. This also means that their names must be unique. I'll have a go at https://github.com/contao/contao/issues/3487 to see if we can optionally allow nested structures still. 

<sup>*)</sup> A draft PR for this functionality can now be found here: #3973.

### Open questions
* Should we migrate the old bundle templates and how?
* Where should new templates, partials, macros, … go? (e.g. for the back end)
* What is our preferred naming scheme/directory structure? 
(with or without https://github.com/contao/contao/issues/3487)

ausi:
As discussed in the Contao call, our first draft for naming conventions for the new template system:

* `content_element/_block_searchable.html.twig` partial template
* `content_element/_block_unsearchable.html.twig` partial template
* `content_element/_somesing/foo.html.twig` partial template
* `content_element/text.html.twig` main template für ce_text
* `content_element/text.json.twig` main template für ce_text
* `content_element/text/foobar.html.twig` varianten-template
* `content_element/text/foobar.json.twig` varianten-template
* `content_element/text/_samsing.html.twig` partial (wird nicht als variante aufgelistet)
* `content_element/text/subfolder/bar.html.twig` wird nicht als variante aufgelistet
* `frontend_module/navigation.html.twig`
* `backend/layout.html.twig`
* `layout/page.html.twig`

Not meiselt in stone yet ☺️

m-vo:
Sure 😀

Issue #3487, opened 2021-09-21
Support Directory Structure in Namespaced Twig Paths


richardhj:
Currently, the registered Twig templates do not support hierarchy. For example:

| File                                              | Fully-Qualified Twig Path                    |
|---------------------------------------------------|----------------------------------------------|
| `/templates/FooBar/fe_page.html.twig`             | `@Contao_Theme_FooBar/fe_page.html.twig`     |
| `/templates/FooBar/partial/headline.html.twig`    | `@Contao_Theme_FooBar/headline.html.twig`    |
| `/templates/FooBar/partial/container/w33.html.twig`    | `@Contao_Theme_FooBar/w33.html.twig`    |
| `/templates/FooBar/modules/mod_article.html.twig` | `@Contao_Theme_FooBar/mod_article.html.twig` |

While this was always the case for HTML5-Templates (that did not know about hierarchy), we must acknowledge that this behavior might be cumbersome in current practice.

Current practice IMHO involves using "atomic" templates (I borrowed the term from "Atomic Design"). To organize the bulk of templates, we need to support structure in Twig paths. Further, we should make the Twig experience in Contao as seamless as possible for existing Twig developers.

fritzmg:
iirc this was already discussed - but @m-vo will know for sure :)

richardhj:
I think it was voted out. My idea was to simultaneously register the hierarchical and flat template path, but in fact, there are problems of ambiguity.

I first noticed that this will be an issue when I wanted to continue work on #2697. Let's say, I create a template called `contao/templates/Backend/Layout/sidebar.html.twig`. You can only reference it as `@Contao/sidebar.html.twig`, which is super ambiguous. Also, I cannot use the Twig folder (`src/Resources/views`) because those templates won't be extendable by any bundle.

m-vo:
> I think it was voted out.

>  Also, I cannot use the Twig folder

Yes, this was discussed in a call while I was midway implementing it. Initially the design looked like this: You would put your Twig templates where you would normally put them in Symfony land. Then, there is a way to mark them as being "managed" either by moving them in a special directory (like `bundles`, so e.g. `contao`) or by config. The templates would then include the full name with sub directories.

[...]

PR #3973, opened 2022-01-19
Support nested template paths in Twig

m-vo:
This PR allows using nested template paths as an opt-in-solution (for reference see #3487).

image

To enable it, put a `.root` file (thinking of the "tree root") into a directory that is covered by the template locator. So. e.g. `/contao/templates/some_folder/.root` or `/templates/.root`. All files in these directories or any subdirectories will then be made available under a name that includes the path relative to the `.root` file.

**Implication 1**: You cannot overwrite such templates without replicating the directories and adding a `.root` file yourself, because otherwise you would not get the name to match.

**Implication 2**: By using this, the used directory structure is part of your template name (or your "template BC promise" should you have one), so this ideally comes with some sort of best-practice naming guide. Questions to ask: 
* Should there be a vendor folder?
* Are there common folders everyone should be using (like for content elements, modules, …)? 
* Should files in these folders have a vendor prefix? 
* …

Edit: see #3967 for a discussion about a naming convention.

~~This is draft-state without additional unit-tests and only rudimentary manual testing. But it should help to get a discussion going.~~

### DIFF
--- a/core-bundle/src/Twig/Loader/TemplateLocator.php
+++ b/core-bundle/src/Twig/Loader/TemplateLocator.php
@@ -143,8 +143,7 @@ class TemplateLocator
         $templates = [];
 
         foreach ($finder as $file) {
-            $filePath = $file->getPathname();
-            $templates[Path::makeRelative($filePath, $path)] = Path::canonicalize($filePath);
+            $templates[$file->getRelativePathname()] = Path::canonicalize($file->getPathname());
         }
 
         return $templates;
--- a/core-bundle/src/Twig/Loader/TemplateLocator.php
+++ b/core-bundle/src/Twig/Loader/TemplateLocator.php
@@ -16,6 +16,7 @@ use Contao\CoreBundle\Exception\InvalidThemePathException;
 use Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Exception\DriverException;
+use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Filesystem\Path;
 use Symfony\Component\Finder\Finder;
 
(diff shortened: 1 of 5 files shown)


m-vo:
I assigned this issue to the 5.0 milestone now. If we want, we can backport specific things later on, but I feel this way it's much easier to get the individual parts to work together than in the current 4.13 codebase.

ausi (dismissed):

core-bundle/src/Twig/Loader/TemplateLocator.php
```diff
@@ -127,26 +132,55 @@ public function findTemplates(string $path): array
...
+        if (!$this->isNamespaceRoot($path)) {
+            $finder = $finder->depth('< 1');
+        }
+
         $templates = [];
 
         foreach ($finder as $file) {
-            $templates[$file->getFilename()] = Path::canonicalize($file->getPathname());
+            $filePath = $file->getPathname();
+            $templates[Path::makeRelative($filePath, $path)] = Path::canonicalize($filePath);
```
leofeyer:
```suggestion
            $templates[$file->getRelativePathname()] = Path::canonicalize($file->getPathname());
```

(untested)

core-bundle/src/Twig/Loader/TemplateLocator.php
```diff
@@ -127,26 +132,55 @@ public function findTemplates(string $path): array
...
+                    foreach ($namespaceRoots as $directory) {
+                        if (Path::isBasePath($directory, $path)) {
+                            return false;
+                        }
+                    }
+
+                    if ($this->isNamespaceRoot($path)) {
+                        $namespaceRoots[] = $path;
+
+                        return true;
```
leofeyer:
```suggestion
```

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3975, opened 2022-01-20
Allow custom labels for the overview links

leofeyer:
Why is this a bugfix? Because our initial plan to add a custom "go back" link with an additional content element is ruined by the BC layer that automatically adds a `javascript:history.go(-1)` link if no overview page has been specified. So we would end up with two "go back" links.


aschempp:
I thought one would just remove the original link (or hide it with CSS) if I want a custom one?

aschempp:
I don't really understand this. If the idea was to have a custom go-back link through a content element, why did we add the overviewPage field at all?

leofeyer:
No, we would just not set an overview page. It will work this way as soon as we have removed the BC layer. But given the fact that there is a `customLabel` column already, we might as well use it.

leofeyer:
You should have asked this in the PR in which we have added the feature. It's too late now. 🤷‍♂️

aschempp:
> You should have asked this in the PR in which we have added the feature. 

I didn't question the feature? But in Contao 5 there will be no back link if you do not set `overviewPage`. We agreed to keep it BC in 4.13 with `javascript:back()`, since everyone that didn't want the original link should already have removed it from the template.

ausi (approved):

aschempp:
What I actually tried to say is that the initial issue description doesn't really mean anything to me. But _having_ the option to set a custom label certainly doesn't harm 😂 maybe we could apply that in other places as well?

SeverinGloeckle:
I don't think this is the correct solution to the problem.

As mentioned when discussing the initial problem, I would suggest the following change:

Instead of an overview page and a custom label, the editor should be able to enable an option "Add link" which shows a subpalette of:
- URL
- link title
- title text
- target

In this way, the editor is given the full configuration option for the link and the option and sub-palette can be used universally for all kinds of modules.

This solves the problem during the transition period and is also a convenient solution afterwards with no need for additional content elements.

Is it too late to switch this because the `overviewPage` has been added to 4.13 RC already?

m-vo:
> Is it too late to switch this because the overviewPage has been added to 4.13 RC already?

No, we can adjust things in the RC phase. :+1:
 But you might want to open a new issue for this.

leofeyer:
> Instead of an overview page and a custom label, the editor should be able to enable an option "Add link" which shows a subpalette of:
> 
> URL
> link title
> title text
> target
> 
> In this way, the editor is given the full configuration option for the link and the option and sub-palette can be used universally for all kinds of modules.

Definitely not. If you need this kind of flexibility, please use a separate link element after the reader element.

PR #3977, opened 2022-01-20
Revert 'Add an "overview page" field'

leofeyer:
Reverts #3862 and #3975. As discussed, it does not make sense to configure the overview page in the reader module settings. A link element should be added after the reader module instead.


leofeyer:
On the other hand, the content element with the custom link will be shown **after** the comments and the "back to the overview" link is currently shown **before** the comments (where I think it belongs). That's just stupid! 🙃

fritzmg:
> As discussed, it does not make sense to configure the overview page in the reader module settings.

Why? We agreed on the originally proposed solution. It should be reverted to the state prior to #3975.

PR #3979, opened 2022-01-21
Sort the root IDs if there is a `sorting` column

leofeyer:


fritzmg (approved):

Toflar (approved):

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6209,7 +6209,7 @@ protected function initRoots()
 			// Unless there are any root records specified, use all records with parent ID 0
 			if (!isset($GLOBALS['TL_DCA'][$table]['list']['sorting']['root']) || $GLOBALS['TL_DCA'][$table]['list']['sorting']['root'] === false)
 			{
-				$objIds = $this->Database->execute("SELECT id FROM $table WHERE pid=0");
+				$objIds = $this->Database->execute("SELECT id FROM $table WHERE pid=0" . ($this->Database->fieldExists('sorting', $table) ? ' ORDER BY sorting' : ''));
```
aschempp, line 6212: I think this database field check is unnecessary, since a TreeView always has a sorting field.

leofeyer: This check is present several times in the tree view methods. I don‘t think it can be omitted.

PR #3982, opened 2022-01-21
Fix undo preview for fragments

aschempp:
The new undo preview feature in 4.13 does not work with fragments, since they require the database record to be present. This PR contains three changes (see 3 commits).

 1. it try-catches any exception to make sure the back end undo section works if anything goes wrong in the child_record/label callback
 2. It enforces the inline renderer in the back end, since you never want an ESI tag there
 3. It changes the fragment handling to always use the original model for inline fragments. This might be a behaviour change, not sure if someone has any doubts.

We might want to fix 2 and 3 in Contao 4.9 and merge upstream instead of just here, since they seem like an inconsistency to me to change just in this version.

### DIFF
--- a/core-bundle/src/Fragment/FragmentHandler.php
+++ b/core-bundle/src/Fragment/FragmentHandler.php
@@ -53,6 +53,10 @@ class FragmentHandler extends BaseFragmentHandler
             throw new UnknownFragmentException(sprintf('Invalid fragment identifier "%s"', $uri->controller));
         }
 
+        if ($uri->isBackendScope()) {
+            $config->setRenderer('inline');
+        }
+
         $this->preHandleFragment($uri, $config);
 
         $renderer = $config->getRenderer();
--- a/core-bundle/src/Fragment/FragmentHandler.php
+++ b/core-bundle/src/Fragment/FragmentHandler.php
@@ -14,6 +14,7 @@ namespace Contao\CoreBundle\Fragment;
 
 use Contao\CoreBundle\Exception\ResponseException;
 use Contao\CoreBundle\Fragment\Reference\FragmentReference;
+use Contao\Model;
 use Contao\PageModel;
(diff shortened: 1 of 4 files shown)


core-bundle/src/Fragment/FragmentHandler.php
```diff
@@ -53,6 +54,10 @@ public function render($uri, string $renderer = 'inline', array $options = []):
             throw new UnknownFragmentException(sprintf('Invalid fragment identifier "%s"', $uri->controller));
         }
 
+        if ($uri->isBackendScope()) {
```
Toflar, line 57: I know that it never makes sense to have an ESI fragment in the back end because you're always logged in = there's always a cookie = there's never a cache. But I still don't think this is the right way to fix the undo problem. We might still keep this of course because as I said, I don't see any valid use case where separate fragments (not even hincluce, really) would make sense in the back end. 
Fixing the undo preview problem can imho only be fixed by rendering the HTML **before it's deleted** and store that along the undo record. Because there are so many things that might not work anymore: relations, API calls, etc. pp. //cc @contao/developers

bezin: See https://github.com/contao/contao/pull/3994

leofeyer:
Can this be closed in favor of #3994 or do I need to merge both?

bezin:
> Can this be closed in favor of https://github.com/contao/contao/pull/3994 or do I need to merge both?
IMO this can be closed

bezin:
The changes made do the `LabelListener` are irrelevant, if #3994 is merged. I don't know about the general changes made regarding fragments.

leofeyer:
#3994 has been merged now. @aschempp Please review the PR and resolve the conflicts if necessary.

aschempp:
I think we should not mess with the fragment system if we don't need to, so I'll close this for now 🙃

PR #3986, opened 2022-01-21
Do not instantiate real data container for undo process

aschempp:
Unfortunately, instantiating a DataContainer will call the `onload_callback` (and a lot of other stuff), which will check permissions and fail for non-existing records and when not being in the "assumed" backend scope.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
+++ b/core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
@@ -62,7 +62,11 @@ class LabelListener
     private function getTemplateData(string $table, array $row, array $originalRow): array
     {
         $dataContainer = $this->framework->getAdapter(DataContainer::class)->getDriverForTable($table);
-        $originalDC = new $dataContainer($table);
+        $ref = new \ReflectionClass($dataContainer);
+        $originalDC = $ref->newInstanceWithoutConstructor();
+        $originalDC->table = $table;
+        $originalDC->id = $originalRow['id'];
+        $originalDC->activeRecord = (object) $originalRow;
 
         $user = $this->framework->getAdapter(UserModel::class)->findById($row['pid']);
         $config = $this->framework->getAdapter(Config::class);


core-bundle/src/EventListener/DataContainer/Undo/LabelListener.php
```diff
@@ -62,7 +62,11 @@ public function __invoke(array $row, string $label, DataContainer $dc): string
     private function getTemplateData(string $table, array $row, array $originalRow): array
     {
         $dataContainer = $this->framework->getAdapter(DataContainer::class)->getDriverForTable($table);
-        $originalDC = new $dataContainer($table);
+        $ref = new \ReflectionClass($dataContainer);
+        $originalDC = $ref->newInstanceWithoutConstructor();
+        $originalDC->table = $table;
```
m-vo, line 67: This should IMHO set the protected field `strTable` instead, because there is no magic setter for `table` in the `DataContainer` class. 

You can also see this with the currently failing 8.2 test:

> Creation of dynamic property Contao\CoreBundle\Fixtures\Contao\DC_NewsTableStub::$table is deprecated

Should we maybe also use `intId` and `objActiveRecord`? These feel more part of the API to me, but on the other hand, custom data containers can do pretty much anything…

m-vo (changes requested):
This is probably the best fix for the symptom. But ultimately we might want to keep the context or the whole rendered preview when storing undo records, so that we do not have to 'boot the system' for a zombie record, that might induce all sorts of things. :neutral_face:

bezin:
Good catch. 

> This is probably the best fix for the symptom. But ultimately we might want to keep the context or the whole rendered preview when storing undo records, so that we do not have to 'boot the system' for a zombie record, that might induce all sorts of things. 😐

Yes, my thoughts, too. Store the rendered preview `ondelete` and output this instead of dynamically rendering anything.

bezin:
I'll give that a try and make another PR

bezin:
See #3994

PR #3987, opened 2022-01-22
Fix the available transports service

fritzmg:
Currently in Contao 4.13 you cannot select any available mailer transports that you have defined. This is because the service name of the `AvailableTransports` service was changed in 4.13 - but no adjustments to the code had been made accordingly (or rather no alias was created).

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -407,6 +407,7 @@ services:
 
     contao.mailer.available_transports:
         class: Contao\CoreBundle\Mailer\AvailableTransports
+        public: true
         arguments:
             - '@?translator'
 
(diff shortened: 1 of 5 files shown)


m-vo (dismissed):

m-vo (approved):

core-bundle/src/Resources/config/services.yml
```diff
@@ -407,6 +407,7 @@ services:
 
     contao.mailer.available_transports:
         class: Contao\CoreBundle\Mailer\AvailableTransports
+        public: true
```
fritzmg, line 410: _Note:_ this isn't strictly necessary as the currently used annotation will make the service public in any case. However, since we use the service directly in a DCA in a completely different bundle, I think it makes sense to specifically define the service as being public, instead of relying on the annotation (as the annotation only works within the `core-bundle`).

bytehead (approved):

leofeyer:
Thank you @fritzmg.

core-bundle/src/Resources/config/services.yml
```diff
@@ -1020,6 +1021,7 @@ services:
     Contao\CoreBundle\Intl\Locales:
         alias: contao.intl.locales
         public: true # backwards compatibility
+    Contao\CoreBundle\Mailer\AvailableTransports: '@contao.mailer.available_transports'
```
aschempp, line 1024: shouldn't the alias also be public (if it was previously)?

leofeyer: See #4027

PR #3990, opened 2022-01-22
Fast manual file sync for the back end

m-vo:
This PR makes the manually file sync in the back end (basically `DC_Folder::sync()`) use the new DBAFS service.

With this, there is no difference anymore between using the back end or the `contao:filesync` command on the CLI. Technically, we could deprecate the old `\Contao\Dbafs::syncFiles()` now, but we might want to leave this until the back end is fully migrated and we consider the DBAFS manager stable.

I simplified the report a bit to fit the new `ChangeSet` (for instance removed "unchanged") but reused styles and labels where possible:

image

As the label that's summing up  the results previously contained more placeholders, I adjusted our translator to gracefully handle the transition in https://github.com/contao/contao/commit/18b5535fbfcf8a6385b6f1d5d4dda38181986512. The solution is not perfect but probably enough. If we want that, I could also extract this part into another PR.


ausi (commented):
✅ Approved apart from the Translator changes. I think we should discuss them in a different PR and you can add the two extra parameters fixed to `0`. Or create a new label for the new thing in this PR.

ausi:
I reverted the Translator changes in 689f34869a01ded4773bb424943339d6141fe6b0, fixed BC with the old label in 8dd3ec1e2fa205dc18ac9951f5cfcb9f911cad5e and fixed a style bug in 80d7e37dc5884ba9c36546256f868d1762cd7dde

The old labels now look like this (until the new translation is added)
Bildschirmfoto 2022-01-28 um 21 48 00

New label:
Bildschirmfoto 2022-01-28 um 21 48 56

Toflar (dismissed):

ausi (dismissed):

m-vo:
👍 to the commits by @ausi. Thanks for adjusting!

core-bundle/src/Resources/contao/languages/en/tl_files.xlf
```diff
@@ -129,7 +129,7 @@
         <source>The synchronization has been completed</source>
       </trans-unit>
       <trans-unit id="tl_files.syncResult">
-        <source>&lt;strong&gt;Result:&lt;/strong&gt; %s added | %s modified | %s unchanged | %s moved | %s deleted</source>
+        <source>&lt;strong&gt;Result:&lt;/strong&gt; %s added | %s updated | %5$s deleted</source>
```
leofeyer: `%5$s`?

ausi: That means use the 5th parameter instead of the 3rd. This is to make the label compatible with older translations. (because the 5th parameter was used for “deleted” files previously).

https://www.php.net/manual/en/function.printf.php

> An integer followed by a dollar sign $, to specify which number argument to treat in the conversion.

leofeyer: So does this mean that we are no longer tracking modified, unchanged and moved files?

ausi: Modified and moved is consolidated into updated AFAIK. The count of unchanged files is probably not that valuable as it is just the count of all files minus updated and added.

leofeyer (changes requested):
There are two issues with this PR:

1. `.DS_Store` files are now added. They haven‘t been before.



2. The "Show unchanged" option is no longer there.

Toflar:
I never understood "Show unchanged". It's just the rest and I don't care about them because they are unchanged :D

leofeyer:
This is not about personal preferences. We cannot remove a feature just because you never understood it. 🤷‍♂️

Toflar:
This is not about personal preference? The information itself is objectively pointless.

m-vo:
> .DS_Store files are now added. They haven‘t been before.

Then that was either wrong before or should be adjusted in the DBAFS service in general. But that's unrelated to this PR in my perspective.

> The "Show unchanged" option is no longer there.

The unchanged files (basically all files minus changed) aren't listed anymore, so this option was removed. Listing unchanged files is not helpful in my eyes and the option a good indicator that you probably wanted to hide them anyway. 🙃 The new change set also does not contain this information.

> Modified and moved is consolidated into updated AFAIK.

Yes. We can differentiate between path changed/hash changed/both changed, but as these could overlap (if both change) the individual counts would be misleading.

leofeyer (changes requested):
As I said, we cannot remove existing features just because they are "not helpful in your eyes" or you "don't care about it". I cannot merge the PR this way.

ausi:
> > .DS_Store files are now added. They haven‘t been before.
> 
> Then that was either wrong before or should be adjusted in the DBAFS service in general.

Needs to be adjusted in DBAFS IMO. A configurable pattern of ignored files maybe? Currently we skip all files that begin with a dot: https://github.com/contao/contao/blob/f577f1ebadebefd88eb44177ab4115a80fcad55a/core-bundle/src/Resources/contao/library/Contao/Filter/SyncExclude.php#L34

> As I said, we cannot remove existing features just because they are "not helpful in your eyes"

We do this all the time, e.g. removing Google Fonts from the layout.

So I think we should discuss the usefulness of the counts and implement what we agree on.

My personal favorite would be `N added | N modified | N moved | N deleted` as it can be helpful, if for example I do some renames on the server, that I can see that nothing was “modified” but only „moved“.

m-vo:
> My personal favorite would be N added | N modified | N moved | N deleted as it can be helpful, if for example I do some renames on the server, that I can see that nothing was “modified” but only „moved“.

Would you then *not* count a file as modified that was also moved or vice versa? Otherwise the individual sums would not add up.

ausi:
> Would you then _not_ count a file as modified that was also moved or vice versa?

If a file was moved AND modified I’d count them in both categories. I would not expect the sums to have to add up based on the shown counts.

ausi (approved):

ausi:
Voting might help us out here ☺️

@contao/developers please vote:
- 🎉 for `N added | N modified | N moved | N deleted` the current state of this PR.
- 🚀 for `N added | N modified | N unchanged | N moved | N deleted` as it is now in 4.12 including unchanged files.

ausi:
My voting questions were not optimal 🙈

@contao/developers the real question is:

- 🎉 for removing the “Show unchanged” option (as currently in this PR)
- 🚀 for keeping/reimplementing the “Show unchanged” option with the list of all files

leofeyer:
I‘m still not a fan of removing the option but at least I can blame you now if it upsets the users. 😛

core-bundle/src/Resources/views/Backend/be_filesync_report.html.twig
```diff
@@ -0,0 +1,44 @@
...
+
+    <p>{{ 'tl_files.syncResult'|trans([change_set.itemsToCreate|length, modified|length, '-', moved|length, change_set.itemsToDelete|length], 'contao_tl_files')|raw }}</p>
+
+    {# Added resources #}
+    {% for added in change_set.itemsToCreate %}
+        <p class="tl_new">{{ 'tl_files.syncAdded'|trans([added[constant('ATTR_PATH', change_set)]]) }}</p>
+    {% endfor %}
+
+    {# Modified resources #}
+    {% for path, updated in modified %}
```
leofeyer, line 24: It seems you are not using the `updated` variable?

core-bundle/src/Resources/views/Backend/be_filesync_report.html.twig
```diff
@@ -0,0 +1,44 @@
...
+        <p class="tl_info">{{ 'tl_files.syncChanged'|trans([path]) }}</p>
+    {% endfor %}
+
+    {# Moved resources #}
+    {% for path, updated in moved %}
+        <p class="tl_info">{{ 'tl_files.syncMoved'|trans([path, updated[constant('ATTR_PATH', change_set)]]) }}</p>
+    {% endfor %}
+
+    {# Deleted resources #}
+    {% for deleted, type in change_set.itemsToDelete %}
```
leofeyer, line 34: It seems you are not using the `type` variable?

leofeyer:
This still needs to be fixed:

> 1. `.DS_Store` files are now added. They haven‘t been before.
> 
> 

These are the options we are using now:

https://github.com/contao/contao/blob/be0d1bbff682506f0f9ed9d71425fc91c45c5cea/core-bundle/src/Resources/contao/library/Contao/Dbafs.php#L129

We should not change the behavior IMHO.

core-bundle/src/Resources/views/Backend/be_filesync_report.html.twig
m-vo: Would you prefer using `modified|keys`?

m-vo:
> .DS_Store files are now added. They haven‘t been before.

see #4111

core-bundle/src/Resources/views/Backend/be_filesync_report.html.twig
leofeyer: Not sure. Isn‘t it `{% for path in modified %}` then?

leofeyer: Ah, I see. Associative array. Never mind then. 👍

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #3994, opened 2022-01-24
Pre-render record preview for undo view on delete

bezin:
As @aschempp found an issue with the current implementation of the undo view, I now pre-render the record preview on delete and store it alongside the other data in `tl_undo`. `LabelListener` than just outputs this alongside the undo records metadata.

Things to do:
* [x] Dynamically register the `StoreRecordPreviewListener` for all data containers that need it. IMO this must be all data containers that are deletable. 
* [x] Get correct parameters for `DataContainer::generateRecordLabel`
* [x] Update tests
* ~~Replace link hrefs in the preview HTML and replace them with `#`~~ Will do a follow-up

WDYT?


core-bundle/src/EventListener/DataContainer/StoreRecordPreviewListener.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Contao\DataContainer;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @Callback(table="tl_article", target="config.ondelete")
+ * @Callback(table="tl_content", target="config.ondelete")
+ * @Callback(table="tl_page", target="config.ondelete")
+ * @Callback(table="tl_page", target="config.ondelete")
```
fritzmg: Instead of using the annotation I think in this case the callbacks should be registered dynamically, e.g. via a `loadDataContainer` hook. As you said, it needs to be done for all deletable DCAs anyway.

fritzmg (commented):
I wonder if this has security implications. The preview could show content that a certain user might not have access to anymore.

core-bundle/src/EventListener/DataContainer/StoreRecordPreviewListener.php
bezin: Yes, that is the first todo in the opening comment ;-)

bezin:
> I wonder if this has security implications. The preview could show content that a certain user might not have access to anymore.

That is true, but also not a new implication as you could have looked at the than restricted data via the show action. 
With the new undo view it would be more easily noticeable though.

bezin:
Except updating and fixing tests and special link treatment (which is actually another issue), the current state would be a possible alternative. What do you all think?

core-bundle/src/EventListener/DataContainer/RecordPreviewListener.php
```diff
@@ -0,0 +1,87 @@
...
+        $system = $this->framework->getAdapter(System::class);
+
+        if (\is_array($GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback'] ?? null)) {
+            $class = $GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback'][0];
+            $method = $GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback'][1];
+
+            $instance = $system->importStatic($class);
+            $preview = $instance->{$method}($row);
+        } elseif (\is_callable($GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback'] ?? null)) {
+            $preview = $GLOBALS['TL_DCA'][$dc->table]['list']['sorting']['child_record_callback']($row);
```
fritzmg: Technically you should check for the DC mode here, as the child record callback is only executed for the parent modes and is otherwise ignored.

core-bundle/src/Resources/contao/dca/tl_undo.php
```diff
@@ -113,6 +113,10 @@
 		(
 			'search'                  => true,
 			'sql'                     => "mediumblob NULL"
+		),
+		'preview' => array
+		(
+			'sql'					  => "text NULL"
```
fritzmg: Should we may be use a `mediumtext` or  `longtext` here? The complete preview output could in theory be rather long.

fritzmg (changes requested):

core-bundle/src/EventListener/DataContainer/RecordPreviewListener.php
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            if (\is_callable($callback)) {
+                return $callback($row);
+            }
+        }
+
+        if ($GLOBALS['TL_DCA'][$dc->table]['list']['label']['showColumns'] ?? false) {
+            $args = $dc->generateRecordLabel($row, $dc->table);
+            return '<table><tr>'.join('', array_map(static function($arg) { return '<td>'.$arg.'</td>'; }, $args)).'</tr></table>';
```
fritzmg: Should probably be a template /cc @m-vo 😁 🙈

m-vo: I think it might be better to store the actual data and compose it when outputting. For example serialized in a blob.
Then you can still adjust markup and styling and it's not baked in more than it has to.

bezin: That would require another column for only the cases where `showColumns === true`. Not sure, if we want that?

m-vo: No, I would store everything in a single blob column, also the content currently going into `preview`.

bezin: See 5b48adf24

core-bundle/src/Resources/contao/dca/tl_undo.php
```diff
@@ -113,6 +113,10 @@
 		(
 			'search'                  => true,
 			'sql'                     => "mediumblob NULL"
+		),
+		'preview' => array
+		(
+			'sql'					  => "mediumblob NULL"
```
fritzmg:
```suggestion
			'sql'                     => "mediumblob NULL"
```

Minor code style issue.

fritzmg (changes requested):

fritzmg (dismissed):

core-bundle/src/EventListener/DataContainer/RecordPreviewListener.php
```diff
@@ -0,0 +1,94 @@
...
+    public function registerDeleteCallbacks(string $table): void
+    {
+        if ($GLOBALS['TL_DCA'][$table]['config']['notDeletable'] ?? false) {
+            return;
+        }
+
+        if (!isset($GLOBALS['TL_DCA'][$table]['config']['ondelete_callback'])) {
+            $GLOBALS['TL_DCA'][$table]['config']['ondelete_callback'] = [];
+        }
+
```
ausi:
```suggestion
```
This is not needed as PHP automatically sets the array when calling `$GLOBALS['TL_DCA'][$table]['config']['ondelete_callback'][] = …`

ausi (dismissed):

ausi (dismissed):

fritzmg (dismissed):

core-bundle/src/EventListener/DataContainer/RecordPreviewListener.php
```diff
@@ -0,0 +1,90 @@
...
+            $preview = $this->renderPreview($dc, $row);
+        } catch (\Exception $exception) {
+            $preview = '';
+        }
+
+        $this->connection->update('tl_undo', ['preview' => $preview], ['id' => $undoId]);
+    }
+
+    private function renderPreview(DataContainer $dc, array $row): string
```
m-vo: This should IMHO be called `compilePreview`, otherwise you could think the output is readily rendered and escaped HTML.

m-vo (dismissed):

bezin:
I only added the popup=1 parameter to get rid of the menu bar inside of the popup. Thanks @xprojects-de

ausi (dismissed):

core-bundle/src/Resources/contao/dca/tl_undo.php
```diff
@@ -113,6 +113,10 @@
 		(
 			'search'                  => true,
 			'sql'                     => "mediumblob NULL"
+		),
+		'preview' => array
+		(
+			'sql'                     => "mediumblob NULL"
```
leofeyer, line 120:
```suggestion
			'eval'                    => array('doNotShow'=>true),
			'sql'                     => "mediumblob NULL"
```

This will hide the field in the "show" view.

leofeyer:
I‘m not sure about this TBH. We now store the undo data twice, once as a serialized array and once as pre-rendered HTML? Is there no better way to generate a preview from the already existing data?

m-vo (dismissed):

Toflar:
> I‘m not sure about this TBH. We now store the undo data twice, once as a serialized array and once as pre-rendered HTML? Is there no better way to generate a preview from the already existing data?

I don't think there is. Imagine an API content element. The data might contain the API endpoint but the data might be long gone and I'm pretty sure you also wouldn't want the undo list view to execute API requests ;)

leofeyer (approved):

leofeyer:
Thank you @bezin.

PR #3995, opened 2022-01-24
Replace old back end paths

aschempp:
We should fix this 4.13 so extensions still using the old paths will get deprecation warnings. We can then remove the old BE url rewrite in Contao 5.

- [x] requires https://github.com/contao/contao/pull/4091 to not trigger the deprecation when an old path does not exist


Toflar (dismissed):
Funny, I put the same thing on my list this morning :D I love when problems resolve themselves :P

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1180,6 +1180,8 @@ protected static function replaceOldBePaths($strContext)
 
 		$generate = static function ($route) use ($router)
 		{
+			trigger_deprecation('contao/core-bundle', '4.0', 'Using old backend paths has been deprecated in Contao 4.0 and will be removed in Contao 5. Use the backend routes instead.');
+
```
ausi: This will always trigger the deprecation even if no old backend path was used.

Toflar: You sould use the router service directly so you shouldn't even land here.

ausi: But the core itself calls the `replaceOldBePaths()` method. IMO we have to check if there is actually an old backend path present before triggering the deprecation...

aschempp: I would think this only triggers if the regex callback is executed for a match?

ausi: There is no regex or regex callback anywhere in this method? I’m confused now 🙈

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -190,7 +190,7 @@ protected static function queryString()
 	}
 
 	/**
-	 * Return the request URI [path]?[query] (e.g. /contao/index.php?id=2)
+	 * Return the request URI [path]?[query] (e.g. /index.php?id=2)
```
leofeyer: This is wrong. The method does return the path.

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: I think @aschempp is referring to the anonymous function. However, as we are using `str_replace()`, the anonymous function is triggered for every array member. https://3v4l.org/8jZIK

So @ausi is right here.

core-bundle/src/Resources/contao/library/Contao/Environment.php
aschempp: no idea what it returns, but it certainly does not return `/contao/index.php?id=2` 😂

core-bundle/src/Resources/contao/library/Contao/Controller.php
aschempp: Then the whole method is very stupid because we are generating all routes for no reason. They should only be generated if the key is found, so we must replace this with `preg_replace_callback`. Imho that's a patch worth for Contao 4.9

core-bundle/src/Resources/contao/library/Contao/Environment.php
fritzmg: Are you sure? It would only do that, if you access Contao via `localhost` - or may be when aliasing the `public` directory.

core-bundle/src/Resources/contao/library/Contao/Controller.php
ausi: I’d prefer not using words like “stupid”, especially in public code reviews.

core-bundle/src/Resources/contao/library/Contao/Environment.php
aschempp: you cannot/should not run Contao in a subfolder. This return example refers to our old entry point, certainly not a setup in subfolder `contao`. Also, this is just an example what the method returns, it really does not matter 😂

core-bundle/src/Resources/contao/library/Contao/Controller.php
aschempp: see https://github.com/contao/contao/pull/4091 for a fix that will make this deprecation usable 🙃

core-bundle/src/Resources/contao/library/Contao/Environment.php
fritzmg: > you cannot/should not run Contao in a subfolder.

We specifically allow it for `localhost`.

aschempp: I don't get this discussion. The changed line is clearly a **comment example** of an entry point script that no longer exists. 🤷 Yes you can technically create a local folder `contao` and run the `index.php` from there, but this is **obviously** not what the comment was about.

calendar-bundle/src/Resources/contao/modules/ModuleCalendar.php
```diff
@@ -56,7 +56,7 @@ public function generate()
 			$objTemplate->title = $this->headline;
 			$objTemplate->id = $this->id;
 			$objTemplate->link = $this->name;
-			$objTemplate->href = 'contao/main.php?do=themes&amp;table=tl_module&amp;act=edit&amp;id=' . $this->id;
+			$objTemplate->href = StringUtil::specialcharsUrl(System::getContainer()->get('router')->generate('contao_backend', array('do'=>'themes', 'table'=>'tl_module', 'act'=>'edit', 'id'=>$this->id)));
```
leofeyer, line 59: Why did you wrap all the routes with `StringUtil::specialcharsUrl()`? The `replaceOldBePaths()` method does not do this, so either the call is unnecessary or it should be added to the method as well, shouldn‘t it?

aschempp: the old URLs all encoded `&` to `&amp;` and I guessed we should continue to do that. The `replaceOldBePaths` method only changed the prefix, not the query parameters.

leofeyer: @ausi WDYT?

ausi: Yes, the URL has to be HTML-encoded IMO as the template expects that.

leofeyer (approved):

PR #3997, opened 2022-01-24
Fix several array to string conversions

leofeyer:
Fixes the following errors, among others:

* `Warning: strpos() expects parameter 1 to be string, array given`
* `Warning: strncmp() expects parameter 1 to be string, array given`
* `Warning: explode() expects parameter 2 to be string, array given`

This happens if an attacker manipulates the POST request to submit arrays instead of strings.


core-bundle/src/Resources/contao/controllers/FrontendShare.php
```diff
@@ -26,19 +26,29 @@ class FrontendShare extends Frontend
 	 */
 	public function run()
 	{
+		if (($url = Input::get('u', true)) && !\is_string($url))
+		{
+			return new RedirectResponse('../');
```
aschempp: can you please explain what this redirect is supposed to do? Maybe add a comment?

leofeyer: This is an early return. The redirect already exists at the end of the method. :)

aschempp: still a very strange response, isn't it? Should it point to the root instead?

ausi (dismissed):

core-bundle/src/Resources/contao/controllers/FrontendShare.php
leofeyer: Do you mean `new RedirectResponse('/');`?

ausi (dismissed):

core-bundle/src/Resources/contao/controllers/FrontendShare.php
```diff
@@ -26,20 +26,28 @@ class FrontendShare extends Frontend
...
+		{
+			return new RedirectResponse('https://www.facebook.com/sharer/sharer.php?p[url]=' . rawurlencode($url));
+		}
+
+		$text = Input::get('t', true);
+
+		if (!$text || !\is_string($text))
+		{
+			return new RedirectResponse('../');
+		}
```
aschempp: is a text really mandatory to share this on twitter? Can't I just share the URL and keep text empty?

leofeyer: Changed in 3e909d1645305bb73907c55e75b9bc13f377796a.

leofeyer:
Exceptions added in 2f6295f4b3f728aaba89d60552da57f25fa36279.

aschempp (approved):

m-vo (commented):
I cannot say much regarding the chanegs in `Frontend#run()`, but LGTM otherwise.

PR #4001, opened 2022-01-24
Check `$objPage` in `Controller::getTemplate()`

### LINKED ISSUE #3928: $objPage could be null
### Affected version(s)

4.x

### Description

Hello,

using Contao as API-Framework the $objPage could be null at 
https://github.com/contao/contao/blob/4166f4eeccafb9cdcbd15e5e89728bbb4f1f3a81/core-bundle/src/Resources/contao/library/Contao/Controller.php#L82

in my case i´m using the notificationCenter to send emails during an API-Request and there is a warning because $objPage is null. 

So maybe we can check if $objPage !== null

Thanks and Greetings

xprojects-de:
Fixes https://github.com/contao/contao/issues/3928

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -74,7 +74,7 @@ abstract class Controller extends System
 		// Check for a theme folder
 		if (\defined('TL_MODE') && TL_MODE == 'FE')
 		{
-			/** @var PageModel $objPage */
+			/** @var PageModel|null $objPage */
 			global $objPage;
 
 			if (null !== $objPage && $objPage->templateGroup)
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -77,7 +77,7 @@ abstract class Controller extends System
 			/** @var PageModel $objPage */
 			global $objPage;
 
-			if ($objPage->templateGroup)
+			if (null !== $objPage && $objPage->templateGroup)
 			{
 				if (Validator::isInsecurePath($objPage->templateGroup))
 				{
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -77,7 +77,7 @@ abstract class Controller extends System
 			/** @var PageModel|null $objPage */
 			global $objPage;
 
-			if (null !== $objPage && $objPage->templateGroup)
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -77,7 +77,7 @@ public static function getTemplate($strTemplate)
 			/** @var PageModel $objPage */
 			global $objPage;
 
-			if ($objPage->templateGroup)
+			if (null !== $objPage && $objPage->templateGroup)
```
m-vo: The typehint three lines above should also be adjusted to `/** @var PageModel|null $objPage */`.

m-vo (changes requested):

core-bundle/src/Resources/contao/library/Contao/Controller.php
fritzmg: That would need to be done everywhere though (there are 40 more occurrences). Imho that should be a separate PR.

m-vo (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
xprojects-de: Done. It also makes sense there...

bytehead (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -74,10 +74,10 @@ public static function getTemplate($strTemplate)
 		// Check for a theme folder
 		if (\defined('TL_MODE') && TL_MODE == 'FE')
 		{
-			/** @var PageModel $objPage */
+			/** @var PageModel|null $objPage */
 			global $objPage;
 
-			if ($objPage->templateGroup)
+			if (null !== $objPage && $objPage->templateGroup)
```
leofeyer:
```suggestion
			if ($objPage->templateGroup ?? null)
```

xprojects-de: Does in this case the null coalescing Operator not crash if objPage itself is null?

fritzmg: No it won't crash: https://3v4l.org/hlVR1 :)

xprojects-de: Crazy but does what it's supposed to ... Never used the Coalescing Operator in this way :-) "Again what learned" (as Lothar would say ). Thanks

leofeyer (approved):

leofeyer:
Thank you @xprojects-de.

PR #4002, opened 2022-01-25
Add a better exception message if a page is unroutable

### LINKED ISSUE #3996: Route 'cmf_routing_object' not found
### Affected version(s)

4.13.0-RC1

### Description

If a `link` insert tag points to a non-routable page, the following exception will be triggered here:

https://github.com/contao/contao/blob/ffa016c820987d6887332bdf2366a816626c3e7e/core-bundle/src/Resources/contao/models/PageModel.php#L1351-L1352

```
Symfony\Component\Routing\Exception\RouteNotFoundException
None of the chained routers were able to generate route: Route 'cmf_routing_object' not found, Route 'Route 'cmf_routing_object' not found' not found
```

You can basically trigger the exception with `PageModel::findOneByAlias('error-401')->getFrontendUrl()`. We might want to throw a custom exception instead, e.g. "this page type is not routable" or something alike.

leofeyer:
Fixes #3996

### DIFF
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -1324,6 +1324,7 @@ class PageModel extends Model
 	 * @param string $strForceLang Force a certain language
 	 *
 	 * @throws RouteNotFoundException
+	 * @throws ResourceNotFoundException
 	 *
 	 * @return string A URL that can be used in the front end
 	 */
@@ -1382,6 +1383,7 @@ class PageModel extends Model
 	 * @param string $strParams An optional string of URL parameters
 	 *
 	 * @throws RouteNotFoundException
+	 * @throws ResourceNotFoundException
 	 *
 	 * @return string An absolute URL that can be used in the front end
 	 */
@@ -1416,6 +1418,7 @@ class PageModel extends Model
 	 * @param string $strParams An optional string of URL parameters
 	 *
 	 * @throws RouteNotFoundException
+	 * @throws ResourceNotFoundException
 	 *
 	 * @return string The front end preview URL
 	 */
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -1361,7 +1361,7 @@ class PageModel extends Model
 
 			if (!$pageRegistry->isRoutable($this))
 			{
-				throw new ResourceNotFoundException(sprintf('Page ID %s is not routable', $this->id));
+				throw new ResourceNotFoundException(sprintf('Page ID %s is not routable', $this->id), 0, $e);
 			}
 
 			throw $e;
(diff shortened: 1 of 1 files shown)


bytehead (dismissed):

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -1349,7 +1350,22 @@ public function getFrontendUrl($strParams=null, $strForceLang=null)
...
+		{
+			$strUrl = $objRouter->generate(RouteObjectInterface::OBJECT_BASED_ROUTE_NAME, array(RouteObjectInterface::CONTENT_OBJECT => $this, 'parameters' => $strParams));
+		}
+		catch (RouteNotFoundException $e)
+		{
+			$pageRegistry = System::getContainer()->get('contao.routing.page_registry');
+
+			if (!$pageRegistry->isRoutable($this))
+			{
+				throw new ResourceNotFoundException(sprintf('Page ID %s is not routable', $this->id));
```
m-vo: When rethrowing, the inner exception should IMHO always get passed on.

```suggestion
				throw new ResourceNotFoundException(sprintf('Page ID %s is not routable', $this->id), 0, $e);
```

m-vo (changes requested):

fritzmg (changes requested):
The same would need to be done in `PageModel::getAbsoluteUrl`, `PageModel::getPreviewUrl` and `Frontend::addToUrl`.

However, I am not sure this is the right approach - may be @aschempp has an idea how we can improve the exception message of the router for pages?

aschempp:
Can't we just throw that exception in the UnroutableRouteCompiler?

fritzmg:
The `UnroutablePageRouteCompiler` currently throws `RouteNotFoundException('This route is not routable')`. However that is not the exception that will be thrown by the chained router.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -1374,7 +1390,22 @@ public function getAbsoluteUrl($strParams=null)
 		$this->loadDetails();
 
 		$objRouter = System::getContainer()->get('router');
-		$strUrl = $objRouter->generate(RouteObjectInterface::OBJECT_BASED_ROUTE_NAME, array(RouteObjectInterface::CONTENT_OBJECT => $this, 'parameters' => $strParams), UrlGeneratorInterface::ABSOLUTE_URL);
+
```
m-vo, line 1395: The docbloc currently says `@throws RouteNotFoundException` - this would be a breaking change because now we're throwing a `ResourceNotFoundExeception` instead. Or the docbloc is wrong :stuck_out_tongue:. 

(same for `getPreviewUrl()`)

m-vo (commented):
What about moving these three identical blocks of code to a shared `private generateUrl($strParams)`?

core-bundle/src/Resources/contao/models/PageModel.php
leofeyer: Fixed in 9c336f12caa4200814403bbfcaed70b9229ce965.

leofeyer:
> What about moving these three identical blocks of code to a shared `private generateUrl($strParams)`?

That does not make sense, because the `$objRouter->generate()` calls are not identical.

m-vo (approved):

PR #4003, opened 2022-01-25
Fix invalid array access if labels are not defined

m-vo:
Fixes some more places where an array access will fail if labels are not defined.


aschempp (dismissed):

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -516,7 +516,7 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 			}
 
 			// Add the name of the parent elements
-			if ($strTable && \in_array($strTable, $arrTables) && $strTable != $arrTables[0])
+			if ($strTable && \in_array($strTable, $arrTables) && $strTable != ($arrTables[0] ?? null))
```
leofeyer: How can `\in_array($strTable, $arrTables)` be `true` if `$arrTables[0]` is not set?

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -457,7 +457,7 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 			$this->Template->main .= $response;
 
 			// Add the name of the parent element
-			if (isset($_GET['table']) && !empty($GLOBALS['TL_DCA'][$strTable]['config']['ptable']) && \in_array(Input::get('table'), $arrTables) && Input::get('table') != $arrTables[0])
+			if (isset($_GET['table']) && !empty($GLOBALS['TL_DCA'][$strTable]['config']['ptable']) && \in_array(Input::get('table'), $arrTables) && Input::get('table') != ($arrTables[0] ?? null))
```
leofeyer, line 460: `$arrTables` is not related to label, so this change seems wrong.

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/Backend.php
m-vo: Yes, true. I skimmed through the file and fixed other parts as well. In this case `$arrTables` can be empty, see L339.

core-bundle/src/Resources/contao/classes/Backend.php
m-vo: You're right, wrongly parsed the parentheses in my head.

m-vo:
```suggestion
			if ($strTable && \in_array($strTable, $arrTables) && $strTable != $arrTables[0])
```

m-vo: Reverted in 36e3bda

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -474,7 +474,7 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 			}
 
 			// Add the name of the submodule
-			$this->Template->headline .= ' <span>' . sprintf($GLOBALS['TL_LANG'][$strTable][Input::get('key')][1], Input::get('id')) . '</span>';
+			$this->Template->headline .= ' <span>' . sprintf($GLOBALS['TL_LANG'][$strTable][Input::get('key')][1] ?? '%s', Input::get('id')) . '</span>';
```
leofeyer: This does not make sense because it will turn this



into this:



Maybe we should wrap the line with an `if (isset($GLOBALS['TL_LANG'][$strTable][Input::get('key')][1]))` instead? Then it would not be added if there is no label.

leofeyer (changes requested):

core-bundle/src/Resources/contao/classes/Backend.php
m-vo: But then there wouldn't be any name? What about using `ID %s` instead?

leofeyer: That would be wrong. The label is supposed to say what the custom action does and not the ID of the record.

If you absolutely want to display something, how about using `Input::get('key')`?



```php
			// Add the name of the submodule
			if (isset($GLOBALS['TL_LANG'][$strTable][Input::get('key')][1]))
			{
				$this->Template->headline .= ' <span>' . sprintf($GLOBALS['TL_LANG'][$strTable][Input::get('key')][1] ?? '%s', Input::get('id')) . '</span>';
			}
			else
			{
				$this->Template->headline .= ' <span>' . Input::get('key') . '</span>';
			}
```

ausi: Fallback to `Input::get('key')` makes most sense IMO.

ausi (approved):

core-bundle/src/Resources/contao/classes/Backend.php
m-vo: adjusted in  e02372fcd9376dc91a9d4c48fb8c66155e06c957

bytehead (approved):

leofeyer:
Thank you @m-vo.

PR #4004, opened 2022-01-25
Support virtual filesystem in CLI backup management

Toflar:
This ~outlines~/implements the changes required to support the new VFS in our (also new) backup management.

It's a demonstration as to how easy it is to work with the new VFS, thank you @m-vo <3 

Implementing it also showed a few minor flaws in the current backup design. Like `Backup` knowing about it's physical location as well as size which should be the `BackupManager`'s responsibility.

Certainly, changing to the VFS is something for Contao 5 or later but we might think about fixing the responsiblity stuff and removing the configuration parameter for `var/backups` in Contao 4.13 before the release. I think it would be better to introduce VFS support where you can configure the mount to an S3 running on cloud computers on the moon if you like rather than having a configurable local path right now.

On the other hand, we can also leave 4.13 as 5 will allow us to break also configuration.
Also depends a bit on @aschempp if he wants to introduce backup support in the Contao Manager. In the current implementation the full local path to the backup is returned in `contao:backup:list --format=json` so he could offer the backup for download. Reading from a local path obviously won't be possible anymore if the backup directory is mounted to S3 for example. That's why I included a `contao:backup:stream-content <backup-name>` command in this PR. It allows to stream and thus offer the backup contents via CLI and then the Manager no matter where they reside.


core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -216,6 +217,12 @@ public function configureFilesystem(FilesystemConfiguration $config): void
             ->addDefaultDbafs($filesStorageName, 'tl_files')
             ->addMethodCall('setDatabasePathPrefix', [$uploadPath]) // Backwards compatibility
         ;
+
+        // Backups
+        $config
+            ->mountLocalAdapter('var/backups', 'var/backups', 'backups')
+            ->addVirtualFilesystem('backups', 'var/backups')
```
m-vo: I wonder if we should make backups available directly under `backups` and just set the path of the mount to `var/backups` by default?

```suggestion
            ->mountLocalAdapter('var/backups', 'backups', 'backups')
            ->addVirtualFilesystem('backups', 'backups')
```

Toflar: Changed 👍

core-bundle/src/Command/Backup/BackupStreamContentCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+    protected BackupManager $backupManager;
+
+    public function __construct(BackupManager $backupManager)
+    {
+        $this->backupManager = $backupManager;
+
+        parent::__construct();
+    }
+
```
m-vo, line 38: Should we hide this command in the list as it's rather technical?

```suggestion

    protected function configure(): void
    {
        $this->setHidden(true);
    }
    
```

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -27,22 +27,32 @@ class BackupManager
 
     private Connection $connection;
     private DumperInterface $dumper;
-    private string $backupDir;
+    private VirtualFilesystemInterface $vfs;
     private array $tablesToIgnore;
     private RetentionPolicyInterface $retentionPolicy;
 
-    public function __construct(Connection $connection, DumperInterface $dumper, string $backupDir, array $tablesToIgnore, RetentionPolicyInterface $retentionPolicy)
+    public function __construct(Connection $connection, DumperInterface $dumper, VirtualFilesystemInterface $vfs, array $tablesToIgnore, RetentionPolicyInterface $retentionPolicy)
```
m-vo: What do you think about naming the parameter `VirtualFilesystemInterface $backupsStorage` to be in line with how you would use it with autowiring?

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -72,20 +82,17 @@ public function getLatestBackup(): ?Backup
...
-            ->depth('== 0')
-            ->name(Backup::VALID_BACKUP_NAME_REGEX)
-        ;
-
         $backups = [];
 
-        foreach ($files as $file) {
-            $backups[] = new Backup($file->getPathname());
+        foreach ($this->vfs->listContents('', false, VirtualFilesystemInterface::BYPASS_DBAFS) as $file) {
+            try {
```
m-vo, line 88: We should probably ignore directories when listing:

```suggestion
            if (!$file->isFile()) {
                continue;
            }

            try {
```

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -135,9 +174,10 @@ private function doCreate(CreateConfig $config): void
         $backup = $config->getBackup();
 
         // Ensure the target file exists and is empty
-        (new Filesystem())->dumpFile($backup->getFilepath(), '');
+        $this->vfs->write($backup->getFilename(), '');
 
-        $fileHandle = fopen($backup->getFilepath(), 'w');
+        $tmpFile = (new Filesystem())->tempnam(sys_get_temp_dir(), 'ctobckupmgr');
```
m-vo, line 179: Is the idea of using this `$tmpFile` that local writes are faster in general? We could also mount a `tmp` storage then and just use `move()` when finished. But maybe I'm not fully understanding the implications. :relaxed:

m-vo (changes requested):

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Sounds like a great addition to the VFS itself? A flag I mean? Wdyt?

m-vo: You mean for only listing files? Sounds good to me.

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: No the problem is compression and decompression. To get rid of the intermediary local temp file, I think we would need some kind of `GzFilesystem` but it needs to be optional. I mean, you can give it a try if you want to, then you see the problem 👼

core-bundle/src/Command/Backup/BackupStreamContentCommand.php
Toflar: Agreed, done in fd99907f1e42f5a7b32a3331be5e58eccde49eeb

core-bundle/src/Doctrine/Backup/BackupManager.php
Toflar: Agreed, done :)

leofeyer:
#4012 has been merged. 🎉

Toflar:
Updated in https://github.com/contao/contao/pull/4004/commits/8e545a8a9b441bc6839f6614d3bc4c4a96e6bdf5 :)

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -27,22 +27,32 @@ class BackupManager
 
     private Connection $connection;
     private DumperInterface $dumper;
-    private string $backupDir;
+    private VirtualFilesystemInterface $backupsStorage;
```
leofeyer, line 30: Just a minor nitpick: I think it is called "backup storage" instead of "backups storage", even though there might be multiple backups in the same storage. 🤓

Toflar: It's because autowiring would name it `backupsStorage` because it's named `backups` and not `backup`. Same as `files` :)

m-vo (dismissed):

ausi (dismissed):

core-bundle/src/Doctrine/Backup/BackupManager.php
m-vo: Just for the record: If we want this (and the autowired variables) to be named `$backupStorage` but still use `/backups` as the mount path, we just need to set the name explicitly:

```php
// Backups
$config
    ->mountLocalAdapter('var/backups', 'backups', 'backup') // fs path, mount path, name
    ->addVirtualFilesystem('backup', 'backups') // name, upload path
;
```

But we should probably adjust this for `files` as well in this case.

core-bundle/src/Doctrine/Backup/BackupManager.php
```diff
@@ -27,22 +27,32 @@ class BackupManager
...
-        $this->backupDir = $backupDir;
+        $this->backupsStorage = $backupsStorage;
         $this->tablesToIgnore = $tablesToIgnore;
         $this->retentionPolicy = $retentionPolicy;
     }
 
+    public function createNewBackup(\DateTime $dateTime = null): Backup
+    {
+        $now = $dateTime ?? new \DateTime('now');
+        $now->setTimezone(new \DateTimeZone('UTC'));
```
leofeyer, line 46: What exactly does this imply? Shouldn't we use the configured timezone instead?

Toflar: The timezone is not part of the filename so it has to be UTC.

leofeyer: Although this is correct from a technical point of view, it will surely confuse the users.



Any other file I create on the server (via `sftp`, `vi`, `touch` or …) will have the correct atime, mtime and ctime, therefore the backup should not behave differently IMHO. @contao/developers WDYT?

fritzmg: Yes, imho we should show it in the configured timezone, just as the operating system would do.

ausi: Storing as UTC makes sense IMO. But we could display it in the local timezone in the list command I think.

leofeyer: The filename would still say `20220204121042` instead of `…131042` then.

Toflar: Filenames **MUST** be represented in a consistent way. The only valid choice is UTC.

leofeyer: Not sure about that, either. `mysqldump mydb > backup__$(date +%Y%m%d%H%M%S).sql` will also show the date according to the server time zone and not UTC (unless UTC is the server time zone of course).

But if you think that UTC is so important, let‘s just add the timezone offset: `date +%Y%m%d%H%M%S%z` Then the file would be called `backup__ 20220204131042+0100.sql`. Now the timestamp is showing the expected date and time to the user while also representing the timestamp in a consistent way.

(I still think you are overthinking this, though. I‘ve been a server admin for over 20 years and we never had a problem with timestamps in filenames not being UTC. 🤷‍♂️)

Toflar: I just don't think it makes sense to consider the timezone in a file. Why not just store it in UTC like we do ever since the first PR? I don't know why this is suddenly a discussion. It's not related to this PR at all.

leofeyer: I have only noticed it now, otherwise I would have wanted to discuss this in the first PR as well. You cannot seriously want people to run `contao:backup:create` and then have `contao:backup:list` tell them that the last backup is one hour old? What kind of UX is this supposed to be?

Toflar: We can adjust the list command of course. But adjusting the file name doesn't fix anything while actually making it worse because when you switch the timezone settings you have no way of finding out when exactly the backup was created.

Toflar: There are 2 options imho:

1. Leave the date untouched and adjust the table header from `Created` to `Created (UTC)`
2. Adjust the date to your system timezone and adjust the table header from `Created` to `Created (foobar)`

Which one do you prefer? :)

Toflar: 2) is going to be PITA for the unit tests.

leofeyer: Neither is a good solution. We might leave the UTC timestamp in the filename but we surely have to adjust the `Created` column to show the date and time according to the server timezone.

Toflar: Why neither? Both display the timezone, that's what you need?

ausi: I prefer the second option :)

leofeyer: As discussed in the Contao call, we want to keep the timestamp in the filename UTC and adjust the timezone in the table header to the server timezone.

Toflar:
Made `contao:backup:list` aware of the server timezone in https://github.com/contao/contao/pull/4004/commits/e8095afa8413aa2b53837a23e0c775f16be99945.

The table header now simply outputs the timezone offset, so you'd get something like `Created (+08:00)` or `Created (-01:00)` depending on your timezone. This sould account for all timezone offsets (some have half hour offsets or even 45 minutes, see https://en.wikipedia.org/wiki/List_of_UTC_time_offsets) and is very short.

bytehead (dismissed):

leofeyer (changes requested):
Something is not right yet:

Toflar:
What does `date_default_timezone_get()` return for you?

leofeyer:
My bad. 🤦‍♂️ I have fixed the `date.timezone` setting in PHP. The timezone offset is correct now, however, the time still says `11:44:38`. Shouldn‘t it say `12:44:38` now?

Toflar:
Of course, fixed in https://github.com/contao/contao/pull/4004/commits/575dc2f38a1fdc2e8db4c33ffa73de8cb4a4edbd

leofeyer (approved):

leofeyer:
Thank you @Toflar.

Issue #3965, opened 2022-01-19
Problem with deactivated search indexer


frontendschlampe:
### Affected version(s)

4.13.0-RC1

### Description

After updating to Contao 4.13.0-RC1 and disabled search I got the following message in Contao Manager:

```
In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:

The service "contao.crawl.escargot.search_index_subscriber" has a dependency on a non-existent service "contao.search.indexer".
```

I deactivated the search in `config.yml` like this:

```
contao:
    search:
        default_indexer:
            enable: false
```

After communication on [Slack](https://contao.slack.com/archives/CK4J0KNDB/p1642586135098200) I got the hint using this deactivation settings:

```
contao:
    search:
        listener:
            index: false
            delete: false
```
Now I get no message after update.

What will be the right way?

m-vo:
There is also `contao.localconfig.enableSearch`. We should probably take a look at this when resolving the issue and consider triggering a deprecation when using an outdated config (whichever is the right one).

aschempp:
I don't think the legacy config works anymore, it has become obsolete with the indexer. Same as e.g. the URL suffix.

Toflar:
I know what to do here and will provide a PR after my holidays.

Toflar:
I don't know what information you received on Slack but it was completely wrong. Both configurations are correct and they have a completely different function.

```yaml
contao:
    search:
        default_indexer:
            enable: false
```

This disables the default serach indexer that is shipped with Contao. It's the same as if you did `contao.localconfig.enableSearch = false` in the past (which by the way still works but is deprecated). However, compared to previous versions that does not mean, that search is entirely disabled. It just means the core indexer is disabled but as we have an abstraction for that now (that's what `contao.search.indexer` is, it's a collection of indexers that forwards to all of them), there might be `n` extensions that provide their own search indexers (if you want to index to ElasticSearch or whatever).
If you disabled the `default_indexer` and no other extension provides any indexer, **then** the search is completely disabled. In other words, the `contao.search.indexer` service is removed.
In that case, it also makes no sense to have a subscriber for the crawler. This needs to removed from the container then too which has always been the case but due to service renaming in https://github.com/contao/contao/pull/3706, the bug was introduced because the service name was not correctly adjusted everywhere. This is the bugfix over at: https://github.com/contao/contao/pull/4005

This configuration

```
contao:
    search:
        listener:
            index: false
            delete: false
```

[...]

PR #4006, opened 2022-01-26
Trigger a deprecation when setting a deprecated contao.localconfig key

m-vo:
This makes sure setting deprecated `localconfig` keys via the bundle configuration will actually trigger a deprecation.

see https://github.com/contao/contao/issues/3965#issuecomment-1021503954


core-bundle/src/Resources/contao/library/Contao/Config.php
```diff
@@ -408,10 +408,7 @@ public static function has($strKey)
 	 */
 	public static function get($strKey)
 	{
-		if (isset(self::$arrDeprecated[$strKey]) || isset(self::$arrDeprecatedMap[$strKey]))
-		{
-			trigger_deprecation('contao/core-bundle', '4.12', 'Using "%s(\'%s\')" has been deprecated. Use the "%s" parameter instead.', __METHOD__, $strKey, self::$arrDeprecated[$strKey] ?? self::$arrDeprecatedMap[$strKey]);
-		}
+		self::checkForDeprecation($strKey, sprintf('Using "%s(\'%s\', …)"', __METHOD__, $strKey));
```
ausi:
```suggestion
		self::checkForDeprecation($strKey, sprintf('Using "%s(\'%s\')"', __METHOD__, $strKey));
```

ausi (dismissed):

Toflar (dismissed):

bytehead (dismissed):

ausi (approved):

aschempp (approved):

leofeyer (changes requested):
I think this is over-engineered. IMHO we should

* keep the deprecation check in the `Config::get()` method;
* not trigger a deprecation in the `Config::set()` method;
* maybe trigger a deprecation in the config tree builder but without adding another public static method to the `Config` class.

ausi:
> not trigger a deprecation in the `Config::set()` method;

But we already do that in the current version. I don’t think we should remove that, should we?
Related: #3056

leofeyer:
My bad. Of course we should keep it then.

If we already trigger a deprecation when setting or getting the key, why do we need another deprecation in the config tree at all?

ausi:
> why do we need another deprecation in the config tree at all?

I think because it would get triggered at container build time and should therefore be seen much more likley by a developer.

leofeyer:
I see. However, I do not like it very much. The configuration tree should not have to know about expired config keys. Maybe we should use `Config::set()` here instead?

https://github.com/contao/contao/blob/2ba558f74e0183ed2daec842f01198dd89e89b1b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php#L82

ausi:
> Maybe we should use `Config::set()` here instead?

We could call `Config::get()` there, just to trigger the deprecation warning (without changing $GLOBALS).

m-vo:
> We could call Config::get() there, just to trigger the deprecation warning (without changing $GLOBALS).

I'm ok with that if you prefer, although I think not using the results of a non-void function and assuming some behavior is way less expressive than just using a function made for this scenario.

leofeyer:
Yes, if we really have to show the deprecations during the container build, this would be the least invasive change. I still don‘t see any added value though.

m-vo:
> I still don‘t see any added value though.

https://github.com/contao/contao/issues/3965#issuecomment-1021503954

You won't get any deprecation notices when setting deprecated config keys via the `contao.localconfig`, but IMHO you should. That we did not trigger deprecations like with the other config keys, is due to the fact we're using a variable node here. But that's only an implementation detail.

m-vo:
Just checked if we could use `Config::get()` but now I remember why I didn't use that. 

This will produce a message like so:

```
Since contao/core-bundle 4.12: Using "Contao\Config::get('enableSearch')" has been deprecated. Use the "contao.search.default_indexer.enable" parameter instead.
```

But it should rather contain the respective config key:
```
Since contao/core-bundle 4.12: Setting "contao.localconfig.enableSearch" has been deprecated. Use the "contao.search.default_indexer.enable" parameter instead.
```

leofeyer:
I have created another PR with a different approach: #4076

leofeyer:
Closed in favor of #4076.

PR #4009, opened 2022-01-26
Fix long content being cut off in the back end popup

qzminski:
When viewing the record information in a popup, it may happen that the data without whitespaces (such as JSON) is cut off at the right. This pull request fixes this problem and makes the popup content container scrollable horizontally.

2022-01-26 at 15

/cc @Toflar


Toflar (approved):

aschempp (approved):

leofeyer:
Thank you @qzminski.

Issue #4011, opened 2022-01-26
Add filter flags for VirtualFilesystemInterface::listContents()


Toflar:
### Description

Currently we have to manually check for `$fileItem->isFile()` or `->isDir()`. Looks like we could improve that by adding more flags to the `VirtualFileSystemInterface`.

See https://flysystem.thephpleague.com/v2/docs/usage/directory-listings/.

m-vo:
I can think of two possible ways to implement this:

* a new flag `$filesOnly = false` for the `VirtualFilesystemInterface#listContents()` method
* returning a custom iterator that has a method `files()` so that you could do `foreach($storage->listContents(…)->files() as $file) { … }`

Toflar:
Both are okay for me. But it should be consistent then. Either one or the other, not both. So the dbafs flags would also need to become part of the iterator.

Toflar:
Would also like to have @ausi's input here :)

m-vo:
IMHO the iterator would have the benefit to be extendable in the future without breaking anything. And too many optional parameters might get cumbersome. So I'm leaning towards this solution.

Toflar:
Same gut feeling over here 👍

m-vo:
see #4012 for a proposal

PR #4012, opened 2022-01-26
Allow filtering for files/directories when listing contents

m-vo:
Implements #4011 

This introduces a `FilesystemItemIterator` that allows filtering the result on the go. There is a shortcut `files()` and `directories()` that filters for, well, files and directories. :smile: Everything is on-demand, because it uses generators under the hood.

After implementing, I realized, that Flysystem uses somthing awfully similar with their [DirectoryListing](https://github.com/thephpleague/flysystem/blob/3.x/src/DirectoryListing.php) which to me is an indicator, that we're on the right track.

**Usage:**
```php
// Listing files
foreach ($this->myStorage->listContents('images')->files() as $file) {
    $item->isFile(); // will be true
}

// Custom filter
$textFilesFilter = static fn(FilesystemItem $item): bool =>
    $item->isFile() && 'txt' === Path::getExtension($item->getPath(), true)
;

foreach ($this->myStorage->listContents('images')->filter($textFilesFilter) as $textFiles) {
    // …
}
```

### DIFF
--- a/core-bundle/src/Filesystem/FilesystemItemIterator.php
+++ b/core-bundle/src/Filesystem/FilesystemItemIterator.php
@@ -63,6 +63,7 @@ class FilesystemItemIterator implements \IteratorAggregate
     {
         foreach ($this->listing as $item) {
             if (!$item instanceof FilesystemItem) {
+                /** @phpstan-ignore-next-line */
                 $type = \is_object($item) ? \get_class($item) : \gettype($item);
 
                 throw new \TypeError(sprintf('%s can only iterate over elements of type %s, got %s.', __CLASS__, FilesystemItem::class, $type));
--- a/core-bundle/src/Filesystem/FilesystemItemIterator.php
+++ b/core-bundle/src/Filesystem/FilesystemItemIterator.php
@@ -70,7 +70,6 @@ class FilesystemItemIterator implements \IteratorAggregate
      */
     public function toArray(): array
     {
-        return $this->listing instanceof \Traversable ?
-            iterator_to_array($this->listing, false) : (array) $this->listing;
+        return iterator_to_array($this, false);
     }
 }
--- a/core-bundle/src/Filesystem/FilesystemItemIterator.php
+++ b/core-bundle/src/Filesystem/FilesystemItemIterator.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Filesystem;
+
+/**
+ * @implements \IteratorAggregate<int, FilesystemItem>
+ */
+class FilesystemItemIterator implements \IteratorAggregate
+{
+    /**
+     * @var iterable<FilesystemItem>
+     */
+    private iterable $listing;
+
+    /**
+     * @param iterable<FilesystemItem> $listing
+     */
+    public function __construct(iterable $listing)
+    {
+        $this->listing = $listing;
+    }
+
+    public function filter(callable $filter): self
+    {
+        $listFiltered = static function (iterable $listing) use ($filter): \Generator {
+            foreach ($listing as $item) {
+                if ($filter($item)) {
+                    yield $item;
+                }
+            }
+        };
+
+        return new self($listFiltered($this->listing));
+    }
+
+    public function files(): self
+    {
+        return $this->filter(static fn (FilesystemItem $item) => $item->isFile());
+    }
+
+    public function directories(): self
+    {
+        return $this->filter(static fn (FilesystemItem $item) => !$item->isFile());
+    }
+
+    /**
+     * @return iterable<FilesystemItem>
+     */
+    public function getIterator(): iterable
+    {
+        return $this->listing instanceof \Traversable ?
+            $this->listing : new \ArrayIterator($this->listing);
+    }
+
+    /**
+     * @return array<FilesystemItem>
+     */
+    public function toArray(): array
+    {
+        return $this->listing instanceof \Traversable ?
+            iterator_to_array($this->listing, false) : (array) $this->listing;
+    }
+}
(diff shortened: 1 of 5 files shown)


Toflar (dismissed):

core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $listFiltered = static function (iterable $listing) use ($filter): \Generator {
+            foreach ($listing as $item) {
+                if ($filter($item)) {
+                    yield $item;
+                }
+            }
+        };
+
+        return new self($listFiltered($this->listing));
```
ausi, line 46: Specific reason no use a bound closure?

```suggestion
        $listFiltered = function () use ($filter): \Generator {
            foreach ($this->listing as $item) {
                if ($filter($item)) {
                    yield $item;
                }
            }
        };

        return new self($listFiltered());
```

core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -0,0 +1,76 @@
...
+        return $this->filter(static fn (FilesystemItem $item) => !$item->isFile());
+    }
+
+    /**
+     * @return \Traversable<FilesystemItem>
+     */
+    public function getIterator(): \Traversable
+    {
+        return $this->listing instanceof \Traversable ?
+            $this->listing : new \ArrayIterator($this->listing);
```
ausi:
```suggestion
        return $this->filter(static fn (FilesystemItem $item) => true);
```
Type check for `FilesystemItem` so that it is guaranteed that no other type is returned?

core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * @return array<FilesystemItem>
+     */
+    public function toArray(): array
+    {
+        return $this->listing instanceof \Traversable ?
+            iterator_to_array($this->listing, false) : (array) $this->listing;
```
ausi:
```suggestion
        return iterator_to_array($this, false);
```
For type checking. And also removes the keys from the array ☺️

ausi (dismissed):
Just nitpicks ☺️

core-bundle/src/Filesystem/FilesystemItemIterator.php
m-vo: > Specific reason no use a bound closure?

Kind of: a static function does not hold a reference to the object which would prevent garbage collection. It's probably not a huge deal here, because the object is quite small, but it's probably a good idea in general for chained iterators.

m-vo: I constructed an example where the difference is quite severe: https://3v4l.org/KSubB#v8.1.2 (switch to the commented version to see the difference). I had to make the object artificially large and store references to make the effect become visible, though. :slightly_smiling_face:

core-bundle/src/Filesystem/FilesystemItemIterator.php
m-vo: I like this! :heart:

core-bundle/src/Filesystem/FilesystemItemIterator.php
m-vo: This would end in an endless loop, though. Not sure if it's worth implementing a more sophisticated solution 'just' for the type check.

ausi: Endless loop? 🤔

core-bundle/src/Filesystem/FilesystemItemIterator.php
ausi: I just thought that because the object only consists of the `listing` member that the memory usage would be basically the same. But I agree, if other members get added later or as a general guidline making it static shows the dependencies to the developer directly in the code. 👍

ausi (dismissed):

core-bundle/src/Filesystem/FilesystemItemIterator.php
m-vo: https://3v4l.org/YNqUO#v8.1.2

ausi:
```suggestion
        foreach ($this->listing as $item){
            if (!$item instanceof FilesystemItem) {
                // Type error...
            }
            
            yield $item;
        }
```

m-vo: Nice! :sunglasses: 

Implemented in d5ca515a9b502952837f52232284511ff20849af

ausi (dismissed):

ausi (dismissed):

Toflar (dismissed):

core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @return \Traversable<FilesystemItem>
+     */
+    public function getIterator(): \Traversable
+    {
+        foreach ($this->listing as $item) {
+            if (!$item instanceof FilesystemItem) {
+                /** @phpstan-ignore-next-line */
+                $type = \is_object($item) ? \get_class($item) : \gettype($item);
```
leofeyer, line 67: Why this?

m-vo: You cannot use `get_class` on literals and using `gettype` on objects only results in "object". So this way you'll always get a meaningful message.

leofeyer: So why did you have to add `@phpstan-ignore-next-line`?

m-vo: Because trusting the annotated types, this case would never happen. But it's a long way until we'll eventually get real generics, so I followed @ausi's point in https://github.com/contao/contao/pull/4012#discussion_r793088830 and added a type check.

m-vo: failing run before ignoring: https://github.com/contao/contao/runs/4964774085?check_suite_focus=true

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4020, opened 2022-01-28
Fix compatibility with Composer 2.2.5

### LINKED ISSUE #4017: Composer version 2.2.5 might break Contao installations
### Affected version(s)

all

### Description

Due to the new handling regarding the composer/package-versions-deprecated plugin introduced in Composer version 2.2.5 (see Pull Request [#10458](https://github.com/composer/composer/pull/10458)), there's a chance that the ``vendor/composer/package-versions-deprecated/src/PackageVersions/Versions.php`` file contains an empty VERSIONS constant. That could lead to errors like ``There is no "contao-components/*" asset package``.

I just had two installations where composer was updated from 2.1.x to 2.2.5 which lead to this issue.

ausi:
Fixes #4017

For everyone experiencing the issue `There is no "contao-components/*" asset package`:  
Downgrading to Composer 2.2.4 (e.g. via `composer self-update 2.2.4`) can be a way to workaround this issue for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -232,7 +232,6 @@
     "config": {
         "allow-plugins": {
             "bamarni/composer-bin-plugin": true,
-            "composer/package-versions-deprecated": true,
             "contao-components/installer": true,
             "contao/manager-plugin": true
         }
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "ext-zlib": "*",
         "ausi/slug-generator": "^1.1",
         "bacon/bacon-qr-code": "^2.0",
-        "composer/package-versions-deprecated": "^1.8",
+        "composer/package-versions-deprecated": "^1.11.99.5",
         "contao-components/ace": "^1.2",
         "contao-components/chosen": "^1.2",
         "contao-components/colorbox": "^1.6",
--- a/core-bundle/composer.json
+++ b/core-bundle/composer.json
@@ -26,7 +26,7 @@
         "ext-zlib": "*",
         "ausi/slug-generator": "^1.1",
         "bacon/bacon-qr-code": "^2.0",
-        "composer/package-versions-deprecated": "^1.8",
+        "composer/package-versions-deprecated": "^1.11.99.5",
         "contao-components/ace": "^1.2",
         "contao-components/chosen": "^1.2",
         "contao-components/colorbox": "^1.6",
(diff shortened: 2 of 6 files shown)


Toflar (dismissed):

aschempp:
I imagine this would be a problem for the Contao Manager. If we ever ship Composer >= 2.2.5 people can no longer install a Contao version that does not include this bugfix, am I right? So it will not be possible to install e.g. Contao 4.11, not from an existing `composer.lock` either?

ausi:
With Composer 2.2.5+ allowing the package explicitly should solve the issue too:

```
"config": {
    "allow-plugins": {
        "composer/package-versions-deprecated": true
    }
},
```

But composer will not automatically ask you if you want to enable the plugin (as it does with other “unknown” plugins).

If the manager can automatically append this config using 2.2.5 should be possible.

aschempp:
so this should already be fixed (in dev) 🙃 
https://github.com/contao/contao-manager/commit/891e59ece5d4968a0778812acb51871d4febc290#diff-20de9bfda6f6cf4bf4e1062c4f2ed258244cdf5a179cb0cf39ff66d7aac966ceR40

ausi:
> so this should already be fixed (in dev) 🙃

Agree, the manager should not be affected then 💪

leofeyer (approved):

leofeyer:
Thank you @ausi.

fritzmg:
> But composer will not automatically ask you if you want to enable the plugin (as it does with other “unknown” plugins).

Why though?

ausi:
> Why though?

Because it is not needed. The package versions code would use the new `InstalledVersions` instead anyway.

We should never have used the constant in the first place because it is meant to be internal.

fritzmg:
But composer definitely asked at some point (not sure which pre 2.2.5 snapshot that was).

ausi:
Yes, from version 2.2.0 until 2.2.4, see https://github.com/composer/composer/commit/6a70161c0175ce178a5abd9edea891ab03298e03

fritzmg:
I am getting this error again with Contao `4.9.27`, Composer `2.2.9` and `"composer/package-versions-deprecated": true` _not_ being present in the `composer.json` under `config.allow-plugins`.

@ausi may be I misunderstood, but this PR is supposed to not cause this error even if you do not have `"composer/package-versions-deprecated": true` in your `composer.json`, correct?

ausi:
You need `composer/package-versions-deprecated` in at least version `1.11.99.5` to fix this error AFAIK. Which version do you have installed in your project?

fritzmg:
```
$ composer show composer/package-versions-deprecated
name     : composer/package-versions-deprecated
descrip. : Composer plugin that provides efficient querying for installed package versions (no runtime IO)
keywords : 
versions : * 1.11.99.5
```

PR #4021, opened 2022-01-28
Fix time sensitive test

ausi:
See <a href="https://github.com/contao/contao/runs/4986268345?check_suite_focus=true#step:6:85" target="_blank">contao/contao/runs/4986268345?check_suite_focus=true#step:6:85</a>

I would guess that this was a copy pase error @Toflar ?

### DIFF
--- a/core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
+++ b/core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
@@ -206,7 +206,8 @@ class ContentElementControllerTest extends TestCase
 
         $response = $controller(new Request(), $model, 'main');
 
-        $this->assertSame($expires, $response->getMaxAge());
+        $this->assertGreaterThan($expires - 10, $response->getMaxAge());
+        $this->assertLessThan($expires + 10, $response->getMaxAge());
     }
 
     public function testSetsTheSharedMaxAgeIfTheElementHasAStopDate(): void
@@ -225,7 +226,8 @@ class ContentElementControllerTest extends TestCase
 
         $response = $controller(new Request(), $model, 'main');
 
-        $this->assertSame($expires, $response->getMaxAge());
+        $this->assertGreaterThan($expires - 10, $response->getMaxAge());
+        $this->assertLessThan($expires + 10, $response->getMaxAge());
     }
 
     public function testDoesNotSetTheSharedMaxAgeIfTheElementHasNeitherAStartNorAStopDate(): void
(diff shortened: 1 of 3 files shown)


ausi:
Fixed another one in 1958cd6ed5d500e9beb08a21a23de481a693c909 see <https://github.com/contao/contao/runs/4987440507?check_suite_focus=true#step:5:70>

ausi:
We should also check if some of these need to be backported 🙈

Toflar (dismissed):
Probably yes 🙈

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
```diff
@@ -206,7 +206,8 @@ public function testSetsTheSharedMaxAgeIfTheElementHasAStartDate(): void
 
         $response = $controller(new Request(), $model, 'main');
 
-        $this->assertSame($expires, $response->getMaxAge());
+        $this->assertGreaterThan($expires - 10, $response->getMaxAge());
```
m-vo: Can't we clock-mock these?

ausi: I think it should be possible, yes. Because the Controller uses the `time()` function: https://github.com/ausi/contao-contao/blob/927a912b96d98499be6b589c32623434d7a30b17/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php#L46

m-vo:
Here is another one: https://github.com/contao/contao/runs/5037584975?check_suite_focus=true

```
1) Contao\CoreBundle\Tests\EventListener\DataContainer\ContentCompositionListenerTest::testUsesTheLayoutColumnForNewArticle with data set #1 (array(array(1, 'foo'), array(0, 'main')), 'main')
Expectation failed for method name is "insert" when invoked 1 time(s)
Parameter 1 for invocation Doctrine\DBAL\Connection::insert('tl_article', Array (...), Array ()) does not match expected value.
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'pid' => 17
     'sorting' => 128
-    'tstamp' => 1643810724
+    'tstamp' => 1643810725
     'author' => 1
     'inColumn' => 'main'
     'title' => 'foo'
```

core-bundle/tests/Controller/ContentElement/ContentElementControllerTest.php
leofeyer: Are you going to adjust the PR accordingly?

ausi: Yes

ausi: Done in e7c2f0e9ca5cf4c641b31ea4939ce929c770a0a4

ausi:
> Here is another one: [contao/contao/runs/5037584975?check_suite_focus=true](https://github.com/contao/contao/runs/5037584975?check_suite_focus=true)

Done in 27673f8ae4d6f0965640d6764a1c51929026440c

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4022, opened 2022-01-28
Add missing option showFilePreview to fileTree widget

ausi:
Followup to #3848

With `'showFilePreview' => true` regular files will also be shown as images in the picker widget.


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4033, opened 2022-01-31
Fix some more PHP8 issues

aschempp:


Toflar (approved):

ausi (approved):

bytehead (approved):

core-bundle/src/Resources/contao/widgets/ImageSize.php
```diff
@@ -75,7 +75,7 @@ public function __set($strKey, $varValue)
 	 */
 	protected function validator($varInput)
 	{
-		$varInput[2] = preg_replace('/[^a-z0-9_]+/', '', $varInput[2]);
+		$varInput[2] = preg_replace('/[^a-z0-9_]+/', '', $varInput[2] ?? '');
```
m-vo, line 78: There is more unsafe array access in the following lines. Should we adjust this as well? Also the method's type hint seems wrong.

aschempp: the other props didn't popup in my situation. I think the error was related to `options` not being filled (see https://github.com/codefog/contao-news_categories/pull/206), which resulted in an empty dropdown. The input fields are still there and submitted.

PR #4034, opened 2022-02-01
Fix 'Warning: Undefined array key 1' in insert tags

xprojects-de:
Fixes "Warning: Undefined array key 1" on PHP 8


aschempp (changes requested):
This code does not do the same as before. The only necessary change should be to use `??` instead of `?:`

xprojects-de:
Changed and tested. Now it´s working as expected...
(looking at first commit the value of element[1] would never have been taken 👍)

aschempp (approved):

leofeyer:
Thank you @xprojects-de.

Issue #3345, opened 2021-08-19
Find a better way to set the doctrine server_version


ausi:
**Affected version(s)**

Contao 4.9

**Description**

I want to find a better solution for https://github.com/contao/contao/blob/5fdcb347db6e2f18e999ee5e673c7d0c916bcee9/manager-bundle/src/ContaoManager/Plugin.php#L311-L328 as this currently also gets triggered for Contao (outside of the install-tool) and therefore can cause issues like in https://github.com/contao/contao/issues/3305#issuecomment-899730840

Related: https://github.com/contao/core-bundle/issues/1422 https://github.com/contao/installation-bundle/issues/54 #1696

PR #4042, opened 2022-02-01
Enable SQL strict mode by default

### LINKED ISSUE #3584: Add warning to enable SQL strict mode for commandline contao-console contao:migrate
Shouldn't `contao-console contao:migrate` correctly display a hint about the MySQL strict mode similar to the install tool?

As discussed in slack, maybe Contao 4.13 should enable strict mode by default.
see: https://contao.slack.com/archives/CJT61P8R1/p1634553592351600?thread_ts=1634552259.350700&cid=CJT61P8R1

m-vo:
This PR enables SQL strict mode by default in the managed edition as discussed in https://github.com/contao/contao/issues/3584#issuecomment-950926414.

If an init command was already defined, nothing will be set. So to change the setting add the following to your config file (use `3` as the key for `mysqli`):

```
doctrine:
  dbal:
    connections:
      default:
        options:
          # Define a MYSQL_ATTR_INIT_COMMAND
          1002: "SET SESSION sql_mode=<your desired mode>"
```

I did not remove the warnings in the install tool as they will only show up if you configured another (non-strict) mode. So it's still valid there.


manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -407,6 +408,52 @@ private function addDefaultDoctrineMapping(array $extensionConfigs, ContainerBui
...
+     */
+    private function enableStrictMode(array $extensionConfigs): array
+    {
+        $driver = '';
+        $options = [];
+
+        foreach ($extensionConfigs as $config) {
+            if (null !== ($driverConfig = $config['dbal']['connections']['default']['driver'] ?? null)) {
+                $driver = $driverConfig;
+            }
```
ausi: We also need to handle the `url` key I think, see line 314

m-vo: Adjusted in fb2cd9e5b89e7ab388f80cc28e3cd9f03089460d - could you have another look?

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -407,6 +408,62 @@ private function addDefaultDoctrineMapping(array $extensionConfigs, ContainerBui
...
+                $options = array_replace($options, $optionsConfig);
+            }
+        }
+
+        if (null !== $url) {
+            $driver = str_replace('-', '_', parse_url($url, PHP_URL_SCHEME));
+        }
+
+        // Skip if driver is not supported
+        if (null === ($key = ['pdo_mysql' => 1002, 'mysqli' => 3][$driver] ?? null)) {
```
ausi:
```suggestion
        if (null === ($key = ['pdo_mysql' => 1002, 'mysql' => 1002, 'mysql2' => 1002, 'mysqli' => 3][$driver] ?? null)) {
```

`mysql` and `mysql2` are aliases for `pdo_mysql`

aschempp (commented):
The whole driver parsing is already done in `addDefaultPdoDriverOptions`, maybe we could make that reuseable?

aschempp:
Also on https://github.com/contao/contao/issues/3584#issuecomment-950926414

> and add a setting so users can disable it in case they have to.

Can this already be done? Can I set an empty `MYSQL_ATTR_INIT_COMMAND` or do I have to explicitly set the old mode?

manager-bundle/src/ContaoManager/Plugin.php
m-vo: see c0f041cd97dc5ab54352c470f1dd1a918c1d3344

m-vo:
> The whole driver parsing is already done in addDefaultPdoDriverOptions, maybe we could make that reuseable?

Good point. I refactored the code in c0f041cd97dc5ab54352c470f1dd1a918c1d3344.

m-vo:
> Can this already be done? Can I set an empty MYSQL_ATTR_INIT_COMMAND or do I have to explicitly set the old mode?

Maybe there is a no-op for init commands but I think it's fine to set the explicit mode in this case. You would not want to do this anyways in the 99% case :smile:.

ausi (dismissed):
❤️

fritzmg (dismissed):

Toflar (dismissed):

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -291,32 +292,15 @@ private function addDefaultPdoDriverOptions(array $extensionConfigs, ContainerBu
...
-        if (null !== $url) {
-            $driver = str_replace('-', '_', parse_url($url, PHP_URL_SCHEME));
+        // Do not add PDO options if custom options have been defined
+        if (isset($options[\PDO::MYSQL_ATTR_MULTI_STATEMENTS])) {
+            return $extensionConfigs;
         }
 
         // Do not add PDO options if the selected driver is not mysql
-        if (null !== $driver && !\in_array($driver, ['pdo_mysql', 'mysql', 'mysql2'], true)) {
+        if (null !== $driver && 'mysql' !== $driver) {
```
leofeyer, line 303: Why did you remove the array? Can `$driver` no longer be `pdo_mysql` or `mysql2`?

leofeyer: Ah, I see. You are normalizing it in the `parseDbalDriverAndOptions()` method. @ausi Is this safe?

ausi: IMO yes.

aschempp (dismissed):

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -464,6 +484,42 @@ private function addDefaultMailer(array $extensionConfigs, ContainerBuilder $con
...
+            if (null !== ($driverConfig = $config['dbal']['connections']['default']['driver'] ?? null)) {
+                $driver = $driverConfig;
+            }
+
+            if (null !== ($urlConfig = $config['dbal']['connections']['default']['url'] ?? null)) {
+                $url = $container->resolveEnvPlaceholders($urlConfig, true);
+            }
+
+            if (null !== ($optionsConfig = $config['dbal']['connections']['default']['options'] ?? null)) {
+                $options = array_replace($options, $optionsConfig);
```
leofeyer: Can we optimize this?

leofeyer (dismissed):

manager-bundle/src/ContaoManager/Plugin.php
ausi:
```suggestion
                $options = $optionsConfig + $options;
```
Would this one be more performant? (And does it work the same? ☺️)

leofeyer: For `array_merge()`, this can usually be fixed like this: https://3v4l.org/Gl8Op

```php
$a = [1];
$n = [[2, 3, 4], [5, 6, 7]];

// Multiple calls to array_merge
foreach ($n as $m) {
    $a = array_merge($a, $m);
}

// ------------------------------

$a = [1];
$b = [$a];

foreach ($n as $m) {
    $b[] = $m;
}

// Single call to array_merge
$a = array_merge(...$b);
```

Maybe the same can be done for `array_replace()`?

m-vo: There is literally no performance loss by using `array_replace` in this case. Feel free to change it, but this is IMHO nothing more than code style.

To the different options: Array merging does not work, because keys are reassigned, spread doesn't work with keys except for recent PHP versions, + means the same, but most of the time people argue that it could be confused with an arithmetic operation.

leofeyer: > There is literally no performance loss by using `array_replace` in this case.

Why not in this case?

m-vo: There a way too few iterations.

leofeyer: At this point, the discussion is becoming pointless.



Any discussion beyond the scope of this decision process is most likely just a waste of time.

m-vo: The inspector cannot know about iterations and the impact. Blindly trusting inspections is not helpful at all. They warn you about possible pitfalls, but you, as a developer, must ultimately be able to judge if this is relevant in your case or not.

If you _really_ believe this must be implemented differently, please go ahead. As I said, for me it's code style at this point.

leofeyer: This is exactly how technical debt is created!

On your system, there are only two or three iterations, so you think "this is not relevant" and ignore it. But out in the wild, there might be systems with fifty or sixty iterations or very huge arrays and suddenly tools like Blackfire identify this exact function call as a bottleneck.

I don‘t understand why a developer would not want to prevent such a problem in the first place? Especially if they have already been made aware of it.

We usually care about defensive programming but in this case it is OK not to do it? IMHO, this is not "judgement" but ignorance. No hard feelings.

ausi (dismissed):

manager-bundle/src/ContaoManager/Plugin.php
ausi:
```suggestion
                $options[] = $optionsConfig;
```

and at the end:

```php
$options = array_replace([], ...$options);
```

should work IMO.

But I would not say it’s ignorance as `$config['dbal']['connections']['default']['options']` wil never be set in fifty configs and will not be huge itself. So I would be happy with both versions 🥰

ausi (dismissed):
🥰

leofeyer (changes requested):
I‘ve just realized that the code does not check if the strict mode is already active. My sql_mode is `ONLY_FULL_GROUP_BY, STRICT_TRANS_TABLES, ERROR_FOR_DIVISION_BY_ZERO, NO_ENGINE_SUBSTITUTION` and this PR still appends `, TRADITIONAL`. Am I missing something?

m-vo:
> I‘ve just realized that the code does not check if the strict mode is already active. My sql_mode is ONLY_FULL_GROUP_BY, STRICT_TRANS_TABLES, ERROR_FOR_DIVISION_BY_ZERO, NO_ENGINE_SUBSTITUTION and this PR still appends , TRADITIONAL. Am I missing something?

That is intentional. `TRADITIONAL` already includes these other modes: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_traditional

Additional modes, not in the list, would still be there, because we're concatenating: `SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))`

leofeyer:
> `TRADITIONAL` already includes these other modes

I know. And our servers use `STRICT_TRANS_TABLES` instead of `TRADITIONAL` for a reason. I don‘t want Contao to change this by default.

m-vo:
We could check for `SELECT @@sql_mode` but we would need to create the database connection ourselves at this point. 

> And our servers use STRICT_TRANS_TABLES instead of TRADITIONAL for a reason.

Why that?

manager-bundle/src/ContaoManager/Plugin.php
```diff
@@ -407,6 +391,41 @@ private function addDefaultDoctrineMapping(array $extensionConfigs, ContainerBui
...
+            return $extensionConfigs;
+        }
+
+        // Enable strict mode
+        $extensionConfigs[] = [
+            'dbal' => [
+                'connections' => [
+                    'default' => [
+                        'options' => [
+                            $key => "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',TRADITIONAL'))",
```
ausi:
```suggestion
                            $key => "SET SESSION sql_mode=IF(FIND_IN_SET('STRICT_TRANS_TABLES', @@sql_mode) + FIND_IN_SET('STRICT_ALL_TABLES', @@sql_mode), @@sql_mode, CONCAT(@@sql_mode, ',TRADITIONAL'))",
```

How about something like this? _(untested)_

ausi: The alternative would be to just add `STRICT_TRANS_TABLES` only (as this is the minimum requirement).
But I would prefer the one from above.

```suggestion
                            $key => "SET SESSION sql_mode=(SELECT CONCAT(@@sql_mode, ',STRICT_TRANS_TABLES'))",
```

leofeyer: Maybe we can reuse the existing check in the install tool? 

https://github.com/contao/contao/blob/c823a49d627cc40712b339fba198ded55a081f2a/installation-bundle/src/InstallTool.php#L277-L284

leofeyer: Not sure if it is a good idea to send a `SET SESSION` query with two `FIND_IN_SET`s upon every request? I actually don‘t want Contao to send any `SET SESSION` query at all if my server is configured correctly.

m-vo: I would go with @ausi's proposal but switch `+` with `OR` to allow short circuiting. 

Init commands are not performance critical. They are bound to the connection, only evaluated once and - in this case - very cheap.

leofeyer: Still I don‘t want Contao to automatically send a `SET SESSION` query. My server is configured correctly, please do not mess around with it.

m-vo: To be honest, I don't know why you're so opposed to this. We discussed and decided that Contao should enforce strict mode, we tested, that setting the session in init commands is no problem on several shared servers and I already invested way too much time in this PR and its predecessor. This is frustrating.

ausi:
```suggestion
                            $key => "SET SESSION sql_mode=CONCAT(@@sql_mode, IF(INSTR(@@sql_mode, 'STRICT_'), '', ',TRADITIONAL'))",
```

This would be shorter and only use a single `INSTR()` check. Takes about 6 microseconds on my system.

ausi (dismissed):

bytehead (approved):

ausi (approved):

aschempp (changes requested):
I fully agree with @leofeyer that Contao should not set any init command if my server is correctly configured. Since we already iterate over the configuration and do have access to the DB credentials, I would prefer to check the server config and only add something if necessary.

fritzmg:
You mean when creating the container?

aschempp:
yes, that's where this config is stored

ausi:
> I would prefer to check the server config and only add something if necessary.

This is already the case, see the `INSTR()` method. If the server is already in strict mode, the init query is a no-op.

>> You mean when creating the container?
>
> yes, that's where this config is stored

Accessing the DB on container build time is not possible, see #3345

fritzmg (approved):

manager-bundle/src/ContaoManager/Plugin.php
ausi: > Takes about 6 microseconds

That was directly measured on the DB btw.

When creating a real PDO connection, I was unable to measure any time difference (probably because PDO sends an init command like `SET NAMES utf8` anyways...)

Toflar (approved):

ausi:
For clarification: this change sets the [native `MYSQL_INIT_COMMAND` option](https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html) for the connection to the server. This is not an additional query that is sent to the database server separately.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4045, opened 2022-02-01
Correctly strip all CSRF token occurrences

### LINKED ISSUE #4038: Use CsrfTokenManager::randomize() everywhere
### Affected version(s)

4.12 (maybe 4.9 too)

### Description

In https://github.com/symfony/symfony/commit/8b01095933e5073688b9bde2842974f33cc88412 something was changed for generating the CSRF tokens. They get randomized now and have a different format.

We have to change https://github.com/contao/contao/blob/64952eea57830f86ff95928f3810b5e5aac6d169/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php#L111 to not use the token storage but the manager instead.

I’m not sure if we can update https://github.com/contao/contao/blob/64952eea57830f86ff95928f3810b5e5aac6d169/core-bundle/src/EventListener/CsrfTokenCookieSubscriber.php#L127 too or if we can get rid of the removal of the (unused) token.

ausi:
Fixes #4038

If my tests are correct, only stripping unnecessary tokens did not work as expected. 
The validation itself still worked.

### DIFF
--- a/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
+++ b/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
@@ -53,7 +53,7 @@ class ContaoCsrfTokenManager extends CsrfTokenManager implements ResetInterface
         return $this->usedTokenValues;
     }
 
-    public function getToken(string $tokenId): CsrfToken
+    public function getToken($tokenId): CsrfToken
     {
         $token = parent::getToken($tokenId);
         $this->usedTokenValues[] = $token->getValue();
@@ -61,7 +61,7 @@ class ContaoCsrfTokenManager extends CsrfTokenManager implements ResetInterface
         return $token;
     }
 
-    public function refreshToken(string $tokenId): CsrfToken
+    public function refreshToken($tokenId): CsrfToken
     {
         $token = parent::refreshToken($tokenId);
         $this->usedTokenValues[] = $token->getValue();
--- a/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
+++ b/core-bundle/src/Csrf/ContaoCsrfTokenManager.php
@@ -18,8 +18,9 @@ use Symfony\Component\Security\Csrf\CsrfToken;
 use Symfony\Component\Security\Csrf\CsrfTokenManager;
 use Symfony\Component\Security\Csrf\TokenGenerator\TokenGeneratorInterface;
 use Symfony\Component\Security\Csrf\TokenStorage\TokenStorageInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
-class ContaoCsrfTokenManager extends CsrfTokenManager
+class ContaoCsrfTokenManager extends CsrfTokenManager implements ResetInterface
 {
     /**
      * @var RequestStack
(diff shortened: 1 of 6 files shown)


core-bundle/src/Csrf/ContaoCsrfTokenManager.php
```diff
@@ -18,8 +18,9 @@
 use Symfony\Component\Security\Csrf\CsrfTokenManager;
 use Symfony\Component\Security\Csrf\TokenGenerator\TokenGeneratorInterface;
 use Symfony\Component\Security\Csrf\TokenStorage\TokenStorageInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
-class ContaoCsrfTokenManager extends CsrfTokenManager
+class ContaoCsrfTokenManager extends CsrfTokenManager implements ResetInterface
```
aschempp, line 23: you must tag the service with `kernel.reset`, there is no autoconfig on `ResetInterface` afair

aschempp (dismissed):
It feels strange that our event listener has both the manager and storage and knows how they internally handle each other. Maybe we should have the manager initialize the storage (via event listener)?

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
ausi: Wanted to do that, but forgot 🙈

fritzmg: But it should work with autoconfigure? https://github.com/symfony/symfony/blob/582f572f5975d05b1dedbfcc55bc09abba463e5c/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L420-L421

ausi:
> It feels strange that our event listener has both the manager and storage and knows how they internally handle each other.

`replaceTokenOccurrences()` needs the manager because it has to remove the occurrences of randomized tokens from the HTML.

The rest of the listener “stores” the tokens from the storage inside a cookie.

How the storage interacts with the manager is not relevant here and the listener doesn’t have to know anything about it (that is why this part still worked when Symfony changed their code).

> Maybe we should have the manager initialize the storage (via event listener)?

I had startet with this approach but came to the conclusion that it doesn’t make sense and would just redirect several method calls to the storage.

ausi:
Need to rebase on 4.9 because of https://github.com/contao/contao/issues/4038#issuecomment-1027728953

bytehead (commented):
This should be rebased against 4.9, as `symfony/security-bundle` requires `symfony/security-csrf` in `4.x` or `5.x`: 

```
composer show symfony/security-bundle
name     : symfony/security-bundle
descrip. : Provides a tight integration of the Security component into the Symfony full-stack framework
keywords : 
versions : * v4.4.36
type     : symfony-bundle
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : https://symfony.com
source   : [git] https://github.com/symfony/security-bundle.git ea53c7fc357a330e7d865ed1bf2d3583f462e3f5
dist     : [zip] https://api.github.com/repos/symfony/security-bundle/zipball/ea53c7fc357a330e7d865ed1bf2d3583f462e3f5 ea53c7fc357a330e7d865ed1bf2d3583f462e3f5
names    : symfony/security-bundle

support
source : https://github.com/symfony/security-bundle/tree/v4.4.36

autoload
psr-4
Symfony\Bundle\SecurityBundle\ => .
exclude-from-classmap

requires
ext-xml *
php >=7.1.3
symfony/config ^4.2|^5.0
symfony/dependency-injection ^4.4|^5.0
symfony/http-kernel ^4.4
symfony/polyfill-php80 ^1.16
symfony/security-core ^4.4
symfony/security-csrf ^4.2|^5.0
symfony/security-guard ^4.2|^5.0
symfony/security-http ^4.4.5
```

fritzmg:
> Need to rebase on 4.9 because of #4038 (comment)

Not necessarily. We can require `symfony/security-csrf: ^4.4` instead.

ausi:
> Not necessarily. We can require `symfony/security-csrf: ^4.4` instead.

@contao/developers WDYT?

👍  for rebase 👎  for adding the dependency.

core-bundle/src/Csrf/ContaoCsrfTokenManager.php
ausi: Done in e02daa5e937f9c17a002bce79319c7fb1e06f6c9

ausi:
This PR is RTM IMO, but it is currently blocked by @aschempp in https://github.com/contao/contao/pull/4045#pullrequestreview-870133977

Toflar (dismissed):

bytehead (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4046, opened 2022-02-01
Encode binary data as hex literal in backup dump

ausi:
This should improve compatibility with SQL tools like SequelPro or SequelAce.


aschempp:
I understand why to use `bin2hex` but I don't understand why we should stop to use DBAL (to create view and detect the type)?

ausi:
> I don't understand why we should stop to use DBAL (to create view…

`CREATE VIEW` does not drop the view before (re-)creating it. That’s why I changed it to `CREATE OR REPLACE VIEW`. DBAL doesn’t support that.

> …and detect the type)?

The type of the column is not relevant actually. Every non-UTF-8 column has to be encoded if it contains non-ASCII data in order to be compatible with several SQL tools. And every UTF-8 column needs to be encoded if it contains non-UTF8 data.

core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -61,7 +60,7 @@ private function dumpViews(AbstractSchemaManager $schemaManager, AbstractPlatfor
     {
         foreach ($schemaManager->listViews() as $view) {
             yield sprintf('-- BEGIN VIEW %s', $view->getName());
-            yield $platform->getCreateViewSQL($view->getQuotedName($platform), $view->getSql()).';';
```
Toflar, line 64: We can just make 2 queries out of it, as it's non persistent data anyway? Like so?

```php
            yield $platform->getDropViewSQL($view->getQuotedName($platform)).';';
            yield $platform->getCreateViewSQL($view->getQuotedName($platform), $view->getSql()).';';
```

core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -117,8 +112,16 @@ private function formatValueForDump($value, Column $column, Connection $connecti
             return 'NULL';
         }
 
-        if ($column->getType() instanceof BinaryType) {
-            return sprintf('UNHEX(%s)', $connection->quote($value));
+        // non-ASCII values
+        if (
+            \is_string($value)
```
Toflar, line 117: Can we have a test for that in `DumperTest` maybe? ☺️

core-bundle/src/Doctrine/Backup/Dumper.php
ausi: This will fail if the view does not exist.

Toflar: Right :) 👍 then

core-bundle/src/Doctrine/Backup/Dumper.php
ausi: Done in 0733330e9b7bb40bacfdc1ef19815fe785673de8

Toflar (approved):
Thanks for the tests <3

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #4049, opened 2022-02-02
Support \Attribute::TARGET_METHOD for our DI attributes

### LINKED ISSUE #4048: Support \Attribute::METHOD for our DI attributes
### Affected version(s)

4.13

### Description

It's possible to use attributes on methods as well. See https://github.com/symfony/symfony/pull/42039 which was merged in 5.4.

We should allow this as well for things like callbacks or hooks, currently only annotations work on a method level, but attributes don't.

m-vo:
Fixes #4048 

With this PR it's possible to use annotations on methods for hooks, callbacks and cronjobs.

Example:
```php
use Contao\CoreBundle\DependencyInjection\Attribute\AsCallback;

class Things
{
    #[AsCallback(table: 'tl_project', target:'list.label.label')]
    public function listThings(array $row): string
    {
        // …
    }

    // …
}
```

I also added tests to the existing code which took forever in comparison to the actual changes. :neutral_face:

### DIFF
--- a/core-bundle/src/DependencyInjection/Attribute/AsCallback.php
+++ b/core-bundle/src/DependencyInjection/Attribute/AsCallback.php
@@ -15,7 +15,7 @@ namespace Contao\CoreBundle\DependencyInjection\Attribute;
 /**
  * An attribute to register a DCA callback.
  */
-#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 class AsCallback
 {
     public function __construct(public string $table, public string $target, public ?string $method = null, public ?int $priority = null)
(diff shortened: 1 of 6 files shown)


m-vo:
Any reason why our coverage CI job is running with PHP7.4?

Edit: see https://github.com/contao/contao/pull/4051

aschempp (commented):
Do we need to handle `AsPage` and `AsPickerProvider` differently? They simply cannot be added to a method (because of the attribute definition), so `instanceof \ReflectionMethod` would just never be true?

m-vo:
> Do we need to handle AsPage and AsPickerProvider differently? They simply cannot be added to a method (because of the attribute definition), so instanceof \ReflectionMethod would just never be true?

Agreed, I simplified the logic in 006688590df7faf724c6c539085b7185fb1be752.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4052, opened 2022-02-02
Do not fetch similar pages with empty alias

aschempp:
Otherwise the user will get a list of all pages with empty alias that potentially conflict with this one – which isn't true.


core-bundle/src/Resources/views/Backend/be_route_path.html.twig
```diff
@@ -1,5 +1,5 @@
 {% trans_default_domain 'contao_tl_page' %}
-<div class="w50 widget">
+<div class="clr w50 widget">
```
aschempp: this fixes alignment issues if there is an error message in the alias field (e.g. if you enter a number)

Toflar (dismissed):

m-vo (dismissed):

core-bundle/src/Resources/views/Backend/be_route_path.html.twig
leofeyer: I cannot reproduce. Can you post a screenshot please?

leofeyer: I will revert this change, so I can merge the fix into the RC2. Feel free to open an issue if you can reproduce, then I‘ll take care of it.

leofeyer (approved):

PR #4065, opened 2022-02-03
Fix order of parameters in AsContentElement and AsFrontendModule constructors

### LINKED ISSUE #4055: Problematic order of optional constructor parameters in AsContentElement/AsFrontendModule
### Affected version(s)

4.13

### Description

Our `AsContentElement` attribute currently looks like this:

https://github.com/contao/contao/blob/3691e1d7f9ee4eb35a9f5e05a3537dc633f94ffc/core-bundle/src/DependencyInjection/Attribute/AsContentElement.php#L23

Having the first non-optional parameter at the second place forces you to define at least two parameters (https://3v4l.org/Y7N4G#v8.1.2), but with the content element attribute, you typically only define the `category`, which then results in:

```
Contao\CoreBundle\DependencyInjection\Attribute\AsContentElement::__construct(): Argument #1 ($type) not passed
```

(Similar reasoning for `AsFrontendModule`.)

@aschempp What's the reason for this unusual order? Can we change this?

m-vo:
Fixes #4055

This PR swaps the first two parameters of the `AsContentElement` and `AsFrontendModule` constructors. This was not released yet, so no BC issue.

Background info: As of PHP8.0 having optional parameters before mandatory ones is deprecated. Here is an explanation why the current code did not trigger a deprecation warning and behaved just like two mandatory parameters: https://github.com/php/php-src/pull/5067#issuecomment-573872903


fritzmg (dismissed):

ausi (dismissed):

aschempp:
I would kinda prefer @ausi's solution in https://github.com/contao/contao/issues/4055#issuecomment-1028927805
If I ever write an attribute with a type, I would never write 
```
#AsContentElement("texts", "foobar")
```

The goal should be to work the same as annotatinos did, so to use named paramters (I know we can't enforce that). Either write
```
#AsContentElement(category: 'texts')
```
or write
```
#AsContentElement("foobar", category: 'texts')
```
if you have a type. Making category default null and throw an exception would solve that.

leofeyer:
@m-vo Will you update this PR accordingly?

ausi:
I’d prefer something that works so that my IDE shows me if something is missing or what will happen.

IMO only the following two possibilities accomplish that:

```
__construct(string $category, string $type = null, …
__construct(string $type = null, string $category = 'texts',…
```

With `string $category = null` my IDE would suggest that this parameter is nullable and not required (which it is not at runtime then 😕)

leofeyer:
I‘d be fine with `string $category = 'texts'` as well.

m-vo:
But then you could omit all parameters and the element would appear in the text category (why text?). I'm ok with that, but I think I would prefer a solution without adding magic.

leofeyer:
Why don‘t we just make `$type` a required parameter, too?

```php
__construct(string $type, string $category, …
```

The maker bundle will always set the type anyway:

https://github.com/contao/contao/blob/f577f1ebadebefd88eb44177ab4115a80fcad55a/maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php#L18-L20

m-vo:
> Why don‘t we just make $type a required parameter, too?

In my opinion it's a good thing, that the type is inferred by default, because it promotes a common naming scheme. And I don't want to type it when creating fragments myself. :slightly_smiling_face:

m-vo:
We should also discuss the usage of the variadic parameter.

fritzmg:
I usually always end up using

```php
/**
 * @ContentElement(MyContentElement::TYPE, category="texts")
 */
class MyContentElement extends AbstractContentElementController
{
    public const TYPE = 'my_content_element';
}
```

so that I can reference the type in the DCA for example.

m-vo:
> I usually always end up using
> 
> /**
>  * @ContentElement(MyContentElement::TYPE, category="texts")
>  */

:+1: for constants, that can be automatically refactored. Maybe the FQCN should get the default type in the future? 

Other aspect: We currently use the type to identify templates, but with Twig there is no point in injecting a `Template` in `getResponse()`. So either we build a special render method that, again, uses the type to identify a template or you would explicitly reference a template in `$this->render()` anyways and not need the type at all. This is also coupled to the discussion about template selection in the back end. 

If we want to alter the behavior in Contao 5, we could maybe already prepare this in 4.13 when using attributes.

leofeyer (changes requested):
As discussed in the Contao call, we want to use `miscellaneous` as default category for both (!) content elements and front end modules. Also, we want to change `array ...$attributes` to `mixed ...$attributes`.

m-vo:
The changes are now implemented as discussed.

Because we're now in a situation, where it's required to use named arguments to set arbitrary keys, I had to use some trickery in the tests, so that jobs < PHP8 would not result in a syntax error. If there is a better solution, let me know.

ausi:
> If there is a better solution, let me know.

```php
new AsContentElement(...[
    'type' => 'content_element/text',
    'template' => 'a_template',
    'method' => 'aMethod',
    'renderer' => 'inline',
    'foo' => 'bar',
    'baz' => 42,
]);
```

https://3v4l.org/3mVUg

core-bundle/tests/DependencyInjection/ContaoCoreExtensionTest.php
```diff
@@ -969,4 +969,11 @@ private function getContainerBuilder(array $params = null): ContainerBuilder
 
         return $container;
     }
+
+    private function skipTestIfAttributesAreNotSupported(): void
+    {
+        if (\PHP_VERSION_ID <= 80000) {
```
ausi:
```suggestion
        if (\PHP_VERSION_ID < 80000) {
```

PHP 8.0.0 should be supported I guess?

m-vo: Ah yes, my bad.

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4067, opened 2022-02-04
Correctly handle null values in DCs

ausi:
With #3876 we introduced a bug with `null` values because `is_scalar(null) === false` in PHP. 🙈

This means that `eval.doNotSaveEmpty` does not work anymore for `null` values and now tries to save them to the database.

This breaks our custom elements extension because it uses `doNotSaveEmpty` for its “virtual” fields. 🥺

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2317,7 +2317,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			}
 
 			// Make sure unique fields are unique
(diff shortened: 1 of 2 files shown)


aschempp:
What if the value is an object? Wouldn't that trigger a warning as well? (see https://github.com/contao/contao/pull/3867#issuecomment-1002215200). We could agree that's not supported though 😊

ausi:
> We could agree that's not supported though 😊

As we do not allow objects in our serialized strings, they probably don’t work now anyway, right?

leofeyer:
Thank you @ausi.

PR #4070, opened 2022-02-04
Fix monorepo-tools

ausi:


leofeyer:
Too late. 😄 https://github.com/contao/monorepo-tools/commit/3c319a2e344f4f7c95187e8b2eb0a813efe5db80

PR #4072, opened 2022-02-04
Reduce memory consumption when warming up cache

### LINKED ISSUE #4069: High memory consumption with new nikic/php-parser implementation
### Affected version(s)

4.9.25

### Description

The new PHP parser implementation based on `nikic/php-parser` that was recently backported to Contao 4.9.25 in https://github.com/contao/contao/pull/4031 causes a much higher memory consumption than the previous implementation based on `phpunit/token-stream`. This scales with the amount of DCA, config and PHP language files that have to be parsed. In one instance on the community forum the usage was bursting through the `128M` `memory_limit` that they happened to have.

In my Contao 4.9 dev instance the _peak_ memory consumption delta (peak memory after minus peak memory before) after `ContaoCacheWarmer::generateConfigCache`, `generateDcaCache` & `generateLanguageCache` is about **8 MiB** in Contao 4.9.24 and about **45 MiB** in Contao 4.9.25.

In the Contao 4.9.25 instance of the community forum user the memory consumption jumps from **40 MiB** to **104 MiB**. Most of it caused by `generateDcaCache` and `generateLanguageCache`.

Toflar:
Attempts to fix https://github.com/contao/contao/issues/4069

### DIFF
--- a/core-bundle/src/Config/Loader/PhpFileLoader.php
+++ b/core-bundle/src/Config/Loader/PhpFileLoader.php
@@ -136,7 +136,7 @@ class PhpFileLoader extends Loader
         $namespaceNode = $namespaceResolver->getNameContext()->getNamespace();
         $namespace = null !== $namespaceNode ? $namespaceNode->toString() : '';
 
-        // Force GC collection in order to reduce total memory required when building the cache // see https://github.com/contao/contao/issues/4069
+        // Force GC collection to reduce the total memory required when building the cache (see #4069)
         gc_collect_cycles();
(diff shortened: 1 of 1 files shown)


fritzmg (dismissed):

aschempp:
Should you maybe unset unused variables so they can be collected as well?

Toflar:
They are because they are unused^^

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #4073, opened 2022-02-05
Replace the getSearchablePages hook in the news bundle

fritzmg:
The `getSearchablePages` hook has been deprecated and will be removed in Contao 5.0. This PR will instead implement a `contao.sitemap` event for news, faqs, newsletters and events.


Toflar:
Let's remove the hook and everything associated with it too then :)

fritzmg:
Yes, though for better PR readability I prefer to separate these tasks. First remove the self deprecating implementations, then remove the hook and everything related once nothing in the core bundles uses it anymore.

fritzmg:
Let's review and merge this first, before I do the other bundles. Failing tests are unrelated.

Toflar (dismissed):

news-bundle/src/EventListener/SitemapListener.php
```diff
@@ -0,0 +1,127 @@
...
+use Contao\CoreBundle\Event\SitemapEvent;
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\News;
+use Contao\NewsArchiveModel;
+use Contao\NewsModel;
+use Contao\PageModel;
+use Terminal42\ServiceAnnotationBundle\Annotation\ServiceTag;
+
+/**
+ * @ServiceTag("kernel.event_listener", event=ContaoCoreEvents::SITEMAP)
```
aschempp: I think we're not using annotations but define our service tags inside the `.yml` file.
(also, we _should_ use PHP8 attributes for Contao 5 😆)

news-bundle/src/EventListener/SitemapListener.php
```diff
@@ -0,0 +1,127 @@
...
+        $newsArchiveModelAdapter = $this->framework->getAdapter(NewsArchiveModel::class);
+
+        // Get all news archives
+        $archives = $newsArchiveModelAdapter->findAll();
+
+        if (null === $archives) {
+            return;
+        }
+
+        /** @var PageModel $pageModelAdapter */
```
aschempp: No need to declare the types, this should be automatically inferred in our latest return type comments.

news-bundle/src/EventListener/SitemapListener.php
```diff
@@ -0,0 +1,127 @@
...
+
+        /** @var NewsArchiveModel $archive */
+        foreach ($archives as $archive) {
+            // Skip archives without a target page
+            if (empty($archive->jumpTo)) {
+                continue;
+            }
+
+            // Skip archive if protected and not accessible
+            if (!$controllerAdapter->isVisibleElement($archive)) {
```
aschempp: this method was not meant to be used here. We could use `ModuleNews::sortOutProtected()` instead (maybe make it `public static`) or use the security voter instead.

news-bundle/src/EventListener/SitemapListener.php
```diff
@@ -0,0 +1,127 @@
...
+
+            $targetPage->loadDetails();
+
+            // Skip news outside the current page root IDs
+            if (!\in_array((int) $targetPage->rootId, $rootPageIds, true)) {
+                continue;
+            }
+
+            // Skip target page if protected and cannot be accessed
+            if (!$controllerAdapter->isVisibleElement($targetPage)) {
```
aschempp: don't reuse this method here. use the security voter instead

news-bundle/src/EventListener/SitemapListener.php
fritzmg: What do you mean? We are using annotations for service tagging in countless services already. But yes, I can change it to PHP 8 attributes instead. A separate PR should then may be change _all_ the annotations to attributes for consistency.

news-bundle/src/EventListener/SitemapListener.php
fritzmg: True, it's just my IDE that's unable to parse that.

news-bundle/src/EventListener/SitemapListener.php
fritzmg: > this method was not meant to be used here.

Why not? `Controller::isVisibleElement` is not restricted to any particular model. It is applied, and can be applied, to any model that contains the following fields:

* `protected`
* `groups`

Currently it is applied to instances of `ArticleModel`, `ContentModel` and `ModuleModel` and I do not see any reason not to apply it here as well, as it already covers everything that is necessary and also implements the `isVisibleElement` hook, which is still not deprecated.

We might of course decide to deprecate this method and its hook - but that is unrelated to this PR and should be done in a separate PR. This PR can then be updated accordingly before or after the fact.

news-bundle/src/EventListener/SitemapListener.php
fritzmg: See above.

leofeyer:
Now that #4548 has been merged, we should probably align this PR with it.

fritzmg:
> Now that #4548 has been merged, we should probably align this PR with it.

#4548 does not include protected calenders in the feed. Since the feed is generated by a Controller in 4.9+ we now have the opportunity to also include protected items in the feed. The current state of my PR would allow this (although my usage of `Controller::isVisibleElement` to achieve this without code duplication is contested by @aschempp).

Toflar:
Let's migrate to the event first and add new features later.

news-bundle/tests/EventListener/SitemapListenerTest.php
```diff
@@ -0,0 +1,103 @@
...
+        ];
+
+        $sitemapEvent = $this->createSitemapEvent([]);
+        $listener = $this->createListener([], $adapters);
+        $listener($sitemapEvent);
+
+        $this->assertStringNotContainsString('<url><loc>', (string) $sitemapEvent->getDocument()->saveXML());
+    }
+
+    public function testCalendarEventIsAdded(): void
```
Toflar: A bit too much copy&paste ;)

fritzmg: 🙈

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

leofeyer:
Thank you @fritzmg.

PR #4074, opened 2022-02-05
Fix `contao:user:list` with empty database

AlexanderWillner:
When executing `contao-console contao:user:list` on an empty database we'd receive the following error otherwise:
`Return value of Contao\CoreBundle\Command\UserListCommand::getUsers() must be an instance of Contao\Model\Collection, null returned`.

With this change instead we receive a nice warning: ` ! [NOTE] No accounts found.`


aschempp:
nice catch! I think we should use an array instead of returning a collection, then this problem would have happened in the first place. This also does not fix the problem when checking for admin only accounts, as that might still return `null`.

AlexanderWillner:
> nice catch! I think we should use an array instead of returning a collection, then this problem would have happened in the first place. 

I've refactored the code in d2d4236 - have you had something like this in mind?

> This also does not fix the problem when checking for admin only accounts, as that might still return `null`.

right - also fixed in d2d4236

leofeyer:
Why not allow `null` as return type?

```php
private function getUsers(bool $onlyAdmins = false): ?Collection
```

AlexanderWillner:
> Why not allow `null` as return type?
> 
> ```
> private function getUsers(bool $onlyAdmins = false): ?Collection
> ```

That's an alternative. In this case we'd need to change line 69 from

```
if (0 === $users->count()) {
```

to something like

```
if (! $users || 0 === $users->count()) {
```

as well. Otherwise we get a the error `Call to a member function count() on null`.

core-bundle/src/Command/UserListCommand.php
```diff
@@ -133,10 +133,10 @@ static function (string $field) use ($user) {
         return $rows;
     }
 
-    private function formatJson(Collection $users, array $columns): array
+    private function formatJson(?Collection $users, array $columns): array
     {
         if ([] === $columns) {
-            return $users->fetchAll();
+            return $users ? $users->fetchAll() : [];
```
leofeyer: There should be an early return at the beginning of the method, because line 144 might still trigger the error.

```php
private function formatJson(?Collection $users, array $columns): array
{
    if (!$users) {
        return [];
    }

    if ([] === $columns) {
        return $users->fetchAll();
    }

    // …
```

leofeyer (changes requested):

leofeyer (approved):

aschempp (approved):
Thanks @AlexanderWillner and sorry for having to refactor this twice 😅

bytehead (approved):

leofeyer:
Thank you @AlexanderWillner.

AlexanderWillner:
> Thanks @AlexanderWillner and sorry for having to refactor this twice 😅

refactoring is fun ;)

PR #4075, opened 2022-02-07
Remove the `Controller::$Template` property

### LINKED ISSUE #3894: Possible BC break with protected `$Template` property
### Affected version(s)

4.9.24, 4.12.5

### Description

The new `protected $Template` property added in https://github.com/contao/contao/pull/3822 breaks existing code which accesses or modifies the template property of an object which didn't have the property before.

Before the change, the `Template` was added to the object's `arrData` dynamically and thus accessible via its magic methods.

For example, modifying the template of a form via the `compileFormFields` hook is not possible anymore:

````php
/**
 * @Hook("compileFormFields")
 */
public function enhanceForms(array $fields, string $formId, Form $form): array
{
  // This worked before, now the template is protected and thus inaccessible from the outside.
  $form->Template->foo = 'bar';
}
```` 

Whether this was a supported use case is debatable for sure, but I want to raise awareness for the issue.

A possible fix could be handling the `Template` property explicitly via magic methods while probably triggering appropriate deprecations.

leofeyer:
Fixes #3894

### DIFF
--- a/core-bundle/src/Resources/contao/controllers/BackendMain.php
+++ b/core-bundle/src/Resources/contao/controllers/BackendMain.php
@@ -25,6 +25,13 @@ use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 class BackendMain extends Backend
 {
 	/**
+	 * @var Template
+	 *
+	 * @todo Remove in Contao 5.0
+	 */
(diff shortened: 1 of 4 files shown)


bytehead (approved):
LGTM

PR #4076, opened 2022-02-07
Trigger a deprecation if an old `contao.localconfig` key is used

leofeyer:
This is an alternative approach to #4006.

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -12,6 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\DependencyInjection;
 
+use Contao\Config;
 use Contao\CoreBundle\Util\LocaleUtil;
 use Contao\Image\ResizeConfiguration;
 use Imagine\Image\ImageInterface;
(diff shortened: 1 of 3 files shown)


core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -64,6 +65,19 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->end()
                 ->variableNode('localconfig')
                     ->info('Allows to set TL_CONFIG variables, overriding settings stored in localconfig.php. Changes in the Contao back end will not have any effect.')
+                    ->validate()
+                        ->always(
+                            static function (array $options): array {
+                                foreach (array_keys($options) as $option) {
+                                    if ($newKey = Config::getNewKey($option)) {
+                                        trigger_deprecation('contao/core-bundle', '4.12', 'Setting "contao.localconfig.%s" has been deprecated. Use "%s" instead.', $option, $newKey);
```
leofeyer, line 73: We could even set the new key here instead of the old one.

ausi: This would not work (easily) because `Config::getNewKey()` also returns non-settable keys (from `$arrDeprecated`).

core-bundle/src/Resources/contao/library/Contao/Config.php
```diff
@@ -424,14 +424,26 @@ public static function get($strKey)
...
-			trigger_deprecation('contao/core-bundle', '4.12', 'Using "%s(\'%s\', …)" has been deprecated. Use the "%s" parameter instead.', __METHOD__, $strKey, self::$arrDeprecated[$strKey] ?? self::$arrDeprecatedMap[$strKey]);
+			trigger_deprecation('contao/core-bundle', '4.12', 'Using "%s(\'%s\', …)" has been deprecated. Use the "%s" parameter instead.', __METHOD__, $strKey, $newKey);
 		}
 
 		$GLOBALS['TL_CONFIG'][$strKey] = $varValue;
 	}
 
+	/**
+	 * Return the new key if the existing one is deprecated
+	 *
```
ausi, line 437:
```suggestion
	 *
	 * @internal
	 *
```

core-bundle/src/DependencyInjection/Configuration.php
leofeyer: True. 🦊

ausi (approved):

PR #4077, opened 2022-02-07
POC: add-on domains

aschempp:
After running again into the issue of copying a product DB to dev and having to change the host name, I had another attempt at fixing this. I first discussed with @Toflar we could use the Symfony reverse proxy to change the host name on-the-fly. However, this does not really work, because Contao will generate all page URLs with the original host name. But I found a pretty simple solution to the problem I think 😎 

There's still a lot to do, and I intend to rewrite our static routes for `/favicon.ico` etc. before this is useable. But feel free to leave a comment about your thoughts 🙃


Toflar:
I honestly don't think any of this makes sense. Rewriting "dev-domain.com" to "domain.com" internally is exactly what a proxy is meant to do. I think this approach opens up Pandora's Box and will result in performance degradations as we need to check for additional domains in many places.
Also, generating URLs won't be possible from e.g. the CLI which is a mandatory requirement for several use cases such as the crawler and more.
I wouldn't want to further investigate this approach.

aschempp:
is a proxy meant to rewrite e.g. a `RedirectResponse` `Location` and the content of all (html) documents as well? The CLI won't change, with or without this PR it would still default to `tl_page.dns` (since there is no proxy or request info to rewrite anything).

The only place that _should_ be affected is routing (URL matching and generating), which is mostly solved here. 
The hardcoded place is our hardcoded controllers that try to find a root page for a domain, which is wrong to do in the database (see my initial comment), since that is what routing is supposed to do!

Toflar:
> is a proxy meant to rewrite e.g. a RedirectResponse Location and the content of all (html) documents as well?

Yes.

fritzmg:
Currently I use a migration to update the database depending on an environment variable. This would work for an arbitrary amount of environments whereas `tl_page.addonDns` would only support one additional domain.

m-vo:
> Currently I use a migration to update the database depending on an environment variable. This would work for an arbitrary amount of environments whereas tl_page.addonDns would only support one additional domain.

Same here, I'm using a custom command, that migrates the "dns" entries which gets executed on deploy/transfer.

bytehead:
> Same here, I'm using a custom command, that migrates the "dns" entries which gets executed on deploy/transfer.

Same same, also using a custom command 👍

aschempp:
> whereas `tl_page.addonDns` would only support one additional domain.

Nope, it supports a comma-separated list of add-on domains 😊

richardhj:
> > Same here, I'm using a custom command, that migrates the "dns" entries which gets executed on deploy/transfer.
> 
> Same same, also using a custom command 👍

Wow, there is clearly a need for such a feature, as almost every core member is using a workaround already 😆 

Maybe this is naïve but I would like to use _environment_ variables for such an _environment_ issue.
Instead of `tl_page.addonDns`, you could allow using an environment variable, e.g., `env(APP_DOMAIN)`. This also works for multi-domain projects where you can have multiple env vars.
The additional benefit is that env vars can easily be used in different parts of the app, for example, when you use custom controllers in your project and want to restrict them to the current domain (I used to use this approach).

EDIT:
Something like this could work:
```php
if ($container->hasParameter($pageModel->dns)) {
    $pageModel->domain = $container->getParameter($pageModel->dns);
}
```
One thing we must be aware of: Allowing backend users to use env vars must not allow backend users to read env vars. We won't be allowed to output the contents of the env vars, otherwise users might enter `env(DATABASE_URL)` and become access to secrets.

aschempp:
closing in favor of https://github.com/contao/contao/pull/6529

PR #4083, opened 2022-02-08
Unlock flatten option for imagine configuration

rabauss:
Fix: Support for animated gifs 

With the `flatten: false` option you can keep animation of GIFs.
By default if the option is not set it is equivalent to `flatten: true`
**Attention:** The `flatten` option is only supported by imagine via Imagick or Gmagick and obviously not by GD - see: 
* https://github.com/php-imagine/Imagine/blob/1.2.4/src/Imagick/Image.php#L422
* https://github.com/php-imagine/Imagine/blob/1.2.4/src/Gmagick/Image.php#L446

This PR unlocks the support for the `flatten` option, so you can disable it globally:
```yml
// config/config.yml
contao:
    image:
        imagine_options:
            flatten: false
```

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -160,6 +160,8 @@ class Configuration implements ConfigurationInterface
                         ->end()
                         ->booleanNode('webp_lossless')
                         ->end()
+                        ->booleanNode('flatten')
+                        ->end()
                         ->scalarNode('interlace')
                             ->defaultValue(ImageInterface::INTERLACE_PLANE)
                         ->end()
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -161,6 +161,7 @@ class Configuration implements ConfigurationInterface
                         ->booleanNode('webp_lossless')
                         ->end()
                         ->booleanNode('flatten')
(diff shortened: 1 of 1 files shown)


m-vo:
Nice! Though, I assume this would be something you would rather want to enable for a single image size not globally.

rabauss:
The option in image sizes was also my thought in the slack discussion - see: https://contao.slack.com/archives/CK4J0KNDB/p1644268875948709.
But a fast way to enable it in a first step would be the global configuration. 
I'm not sure if GIFs are used so often because of better support with videos or let's say more than 256 colors 🌈

ausi (dismissed):

ausi:
> rather want to enable for a single image size not globally.

IMO a global setting should be enough, as this basically means something like “keep animations in GIFs”.

m-vo (dismissed):

m-vo:
> IMO a global setting should be enough, as this basically means something like “keep animations in GIFs”.

My reasoning was: This only works "well enough" for images that aren't too big of a size. And you do not want to have animations playing in your hero image for example. But on the other hand, this only works if the output format is capable, so probably ok.

fritzmg (dismissed):

bytehead (dismissed):

rabauss:
I wonder if we need some hints in the docs or is it generated by the automatic config export?

ausi:
You could add an `->info('')` and describe that this is most useful for animated images and doesn’t have any affect for the `GD` Imagine driver.

ausi (approved):

leofeyer:
Thank you @rabauss.

PR #4051, opened 2022-02-02
Run the coverage CI job with PHP8.0

m-vo:
With attributes and directory checks in the VFS we now have (optional) PHP8 features in the code base. IMHO, we should therefore run our coverage job with PHP8 as well.

PR #4085, opened 2022-02-08
Update the CI chain

leofeyer:
This is not going to be fun. 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,409 +34,405 @@ jobs:
             - name: Generate the coverage report
               run: php -d pcov.enabled=1 vendor/bin/phpunit --testsuite=coverage --coverage-clover=clover.xml --colors=always
 
-            - name: List
-              run: ls -la
-
             - name: Upload the coverage report
               uses: codecov/codecov-action@v2
               with:
                   files: ./clover.xml
-                  verbose: true
                   fail_ci_if_error: true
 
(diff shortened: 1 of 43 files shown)


.github/workflows/ci.yml
```diff
@@ -168,13 +159,13 @@ jobs:
...
         name: Require Checker
-        runs-on: ubuntu-18.04
+        runs-on: ubuntu-latest
         if: github.event_name == 'pull_request'
         steps:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                  php-version: 8.1
+                  php-version: 8.0
```
m-vo: Intentional downgrade?

leofeyer:
Nice, it actually went smooth. Seems the MySQL issues in `ubuntu-latest` have been resolved meanwhile. 🎉 

Only Codecov still does not like PHP 8 (also mentioned in #4051).

leofeyer:
Could be this: 😄 

```
PHPUnit 8.5.23 by Sebastian Bergmann and contributors.
Error: This version of PHPUnit does not support code coverage on PHP 8
```

m-vo:
:see_no_evil: We should probably upgrade to 9.3 anyways… 

I ran into PHP8 related issues with 8.5, too, when adjusting the attributes implementation. And there will probably more in the future.

m-vo (approved):

PR #4091, opened 2022-02-09
Lazy call router when replacing old backend paths

aschempp:
In #3995 we noticed the router was called for every old backend path even if it did not exist in the replaced URL. See https://github.com/contao/contao/pull/3995/files#r795530334 for the discussion.

This PR changes the behaviour to use the callback only when necessary, happily covered with some unit tests 😎


Toflar (dismissed):

ausi (dismissed):
❤️

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1210,27 +1210,29 @@ public static function redirect($strLocation, $intStatus=303)
...
+			'contao/password.php'  => 'contao_backend_password',
+			'contao/popup.php'     => 'contao_backend_popup',
+			'contao/preview.php'   => 'contao_backend_preview',
+		);
 
-		$generate = static function ($route) use ($router)
+		$replace = static function ($matches) use ($arrMapper)
 		{
-			return substr($router->generate($route), \strlen(Environment::get('path')) + 1);
+			$router = System::getContainer()->get('router');
```
leofeyer: Just a nitpick: Wouldn‘t it be more performant if we retrieved the router service outside the callable?

aschempp: no, because then we would always retrieve it even if the callback is not called. The callback will never be called multiple times (you usually don't have more than one old entry point in one URL 😉 )

Toflar: Not exactly true, it happens for every single back end template :) But it still doesn't matter^^

aschempp: hmmm, I actually didn't think about that. That means we will generate the same URL multiple times if it appears multiple times. Guess I need to change the PR again 🙈

PR #4093, opened 2022-02-09
Allow autowiring of preview factory

m-vo:
You should IMHO be able to use the new preview factory in your services using autowiring:

```php
public function __construct(private PreviewFactory $previewFactory)
{
}
```

This PR adds an autowiring alias.


ausi (approved):

fritzmg (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #4094, opened 2022-02-09
Fix missing fallback for densities in preview factory

m-vo:
If an image config specified `items` but omitted the `densities` key in these `items`, an invalid array access would occur when generating previews.

This PR fixes that by adding a fallback.


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #4096, opened 2022-02-09
Fix an undefiend array key

richardhj:
This fixes the array key "href" being undefined<sup>1</sup> and throwing an ErrorException/Warning.

<sup>1</sup> This is the case when using the "route" param instead of the "href" param (see 3 lines above the change).

Issue introduced in #3889.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -954,7 +954,7 @@ abstract class DataContainer extends Backend
 						$href = $this->addToUrl($v['href'] . '&amp;id=' . $arrRow['id'] . (Input::get('nb') ? '&amp;nc=1' : ''));
 					}
 
-					parse_str(StringUtil::decodeEntities($v['href']), $params);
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -954,7 +954,7 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 						$href = $this->addToUrl($v['href'] . '&amp;id=' . $arrRow['id'] . (Input::get('nb') ? '&amp;nc=1' : ''));
 					}
 
-					parse_str(StringUtil::decodeEntities($v['href']), $params);
+					parse_str(StringUtil::decodeEntities($v['href'] ?? ''), $params);
```
richardhj, line 957: Intentionally not using `$href` here because when using a route, we won't have any query params in the href anyways.

m-vo: What about moving the original line up two lines and setting `$params = [];` in the `!empty($v['route'])` case?

leofeyer (approved):

leofeyer:
Thank you @richardhj.

PR #4099, opened 2022-02-10
Do not store record preview for DC_Folder instances

### LINKED ISSUE #4098: Cannot delete files
### Affected version(s)

4.13.0-RC2

### Description

When deleting a file, the following exception is thrown: 

`Contao\CoreBundle\EventListener\DataContainer\RecordPreviewListener::storePrecompiledRecordPreview(): Argument #1 ($dc) must be of type Contao\DataContainer, string given`

@bezin /cc

bezin:
Fixes #4098, however I am not sure if this is the way to go. After all, custom data containers could also call a ondelete_callback with other arguments and than we would run into the same issue.

WDYT?


core-bundle/src/EventListener/DataContainer/RecordPreviewListener.php
```diff
@@ -41,6 +41,10 @@ public function registerDeleteCallbacks(string $table): void
             return;
         }
 
+        if ($GLOBALS['TL_DCA'][$table]['config']['dataContainer'] === 'Folder') {
```
ausi:
```suggestion
        if ($GLOBALS['TL_DCA'][$table]['config']['dataContainer'] !== 'Table') {
```

I would skip all data containers that are not `DC_Table`

bezin: That would also exclude `DC_General`, which could be supported.

m-vo: Everyone with a custom DC could still register the callbacks themselves. IMHO we should be defensive here.

bezin: See b9f16f7

aschempp:
I would expect the `ondelete_callback` to always behave the same, and I sure do want to apply the callback to any other DC (e.g. DC_ProductData or DC_Multilingual) which are all based on DC_Table. I would prefer to just not add it to DC_Folder 😊

ausi (dismissed):

leofeyer:
As briefly discussed in Slack, we should use an `instanceof` check:

```php
if (!$dc instanceof DC_Table) {
    return;
}
```

This way the hook will work for every DC extending `DC_Table`.

leofeyer (changes requested):
As briefly discussed in Slack, we should use an `instanceof` check:

```php
if (!DataContainer::getDriverForTable($table) instanceof DC_Table) {
    return;
}
```

This way the listener works for all DCs extending `DC_Table`.

fritzmg (changes requested):
If it's just about the parameters of the callback - both `DC_Folder` and `DC_Table` pass the `DataContainer` instance to the `ondelete` callback, just not in the same spot. `RecordPreviewListener::storePrecompiledRecordPreview` could instead not use any arguments and analyse the arguments with `func_get_args()`.

leofeyer:
DC_Folder does not support undo, so the listener does not need to handle it anyway, does it?

bezin:
> If it's just about the parameters of the callback - both `DC_Folder` and `DC_Table` pass the `DataContainer` instance to the `ondelete` callback, just not in the same spot. `RecordPreviewListener::storePrecompiledRecordPreview` could instead not use any arguments and analyse the arguments with `func_get_args()`.

I thought about this, too, at first, but there is actually nothing to do for DC_Folder anyways, as @leofeyer mentioned. So it is all about not storing a preview for Folder DCs at all – the arguments are not relevant here.

ausi (dismissed):

core-bundle/tests/EventListener/DataContainer/RecordPreviewListenerTest.php
```diff
@@ -32,19 +32,21 @@ public static function tearDownAfterClass(): void
...
+        $GLOBALS['TL_DCA'][$table]['config']['dataContainer'] = $dataContainer;
 
         $framework = $this->mockContaoFramework();
         $connection = $this->createMock(Connection::class);
 
         $listener = new RecordPreviewListener($framework, $connection);
         $listener->registerDeleteCallbacks($table);
 
-        if ($notDeletable) {
+        if ($notDeletable || 'Table' !== $dataContainer) {
```
leofeyer: This line still says `'Table' !== $dataContainer`. Intentionally?

bezin: I was still working on the tests, please see the d3a988a

bezin:
I just saw that there is a DC_NewsTableStub, that was from the original undo view PR. It is actually not needed anywhere and is a left-over so we may delete it. Should I commit this here or make a separate PR?

ausi (dismissed):

core-bundle/tests/EventListener/DataContainer/RecordPreviewListenerTest.php
```diff
@@ -29,22 +31,39 @@ public static function tearDownAfterClass(): void
         unset($GLOBALS['TL_DCA']);
     }
 
+    protected function tearDown(): void
+    {
+        unset($GLOBALS['TL_DCA']);
```
ausi: This line might have killed the test suite (e.g. `PreviewLinkListenerTest`) in php 8.1 and 8.2.

and a call to `parent::tearDown()` is missing.

leofeyer:
> I just saw that there is a DC_NewsTableStub, that was from the original undo view PR. It is actually not needed anywhere and is a left-over so we may delete it. Should I commit this here or make a separate PR?

I‘d say commit the change here.

core-bundle/tests/EventListener/DataContainer/RecordPreviewListenerTest.php
ausi: To clarify: I think `unset($GLOBALS['TL_DCA']);` is correct here, but probable some other tests have to be fixed now 🙈

bezin: Yeah I added this, because it cost me like half an hour to find out why a test that was supposed to fail, did not fail at all and that was due to the global registry. However, the line is not necessary here anymore for this test case. Even though I think it would be better, if one has a clean slate after every test run, maybe we should handle this separately.

bezin: NOW this line is not needed anymore ^^ see e85e216

leofeyer (approved):

leofeyer:
I guess the tests are failing because there is no `unset($GLOBALS['TL_DCA'])` at the end of the `testRegistersDeleteCallbackOnDeletableDataContainers()` method?

leofeyer:
The tests are failing because there is no data container here:

https://github.com/contao/contao/blob/c964ea313972bd060e00f7a064d42bfb2ab72a63/core-bundle/src/Resources/contao/classes/DataContainer.php#L1618

See #4113

leofeyer (approved):

leofeyer:
Thank you @bezin.

PR #4103, opened 2022-02-10
Fix the color of bold strings inside error messages

leofeyer:


ausi:
Should we remove it from here instead? https://github.com/contao/contao/blob/4eb7e909cf7a9f1dceed6692bdc2cdafef2f0f16/core-bundle/src/Resources/contao/themes/flexible/main.css#L153

or where does the color come from?

fritzmg:
I like the contrast of before better 🙃

leofeyer:
> Should we remove it from here instead?

I‘d prefer not to. The definition is likely there for a reason.

ausi (approved):

PR #4113, opened 2022-02-11
Return an empty string if there is no driver

leofeyer:
Otherwise the method would return `DC_` if `$GLOBALS['TL_DCA'][$table]['config']['dataContainer']` is not set.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -1615,6 +1615,11 @@ abstract class DataContainer extends Backend
 	 */
 	public static function getDriverForTable(string $table): string
 	{
+		if (!isset($GLOBALS['TL_DCA'][$table]['config']['dataContainer']))
+		{
+			return '';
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1612,9 +1612,16 @@ abstract protected function save($varValue);
...
 	 *
 	 * @return string
+	 *
+	 * @todo Change the return type to ?string in Contao 5.0
 	 */
 	public static function getDriverForTable(string $table): string
 	{
+		if (!isset($GLOBALS['TL_DCA'][$table]['config']['dataContainer']))
+		{
+			return '';
```
ausi, line 1622: Should we throw an invalid argument exception instead?

leofeyer: Risky but of course it would be correct. It might break extensions though.

ausi (approved):

PR #4117, opened 2022-02-11
Do not add the element name to the PHP attribute in the maker bundle

### LINKED ISSUE #4090: Adding type in maker-bundle is wrong
Why don‘t we just make `$type` a required parameter, too?

```php
__construct(string $type, string $category, …
```

The maker bundle will always set the type anyway:

https://github.com/contao/contao/blob/f577f1ebadebefd88eb44177ab4115a80fcad55a/maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php#L18-L20

_Originally posted by @leofeyer in https://github.com/contao/contao/issues/4065#issuecomment-1030807027_

leofeyer:
Fixes #4090

### DIFF
--- a/maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php
+++ b/maker-bundle/src/Resources/skeleton/content-element/ContentElement.tpl.php
@@ -16,7 +16,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
 <?php if ($use_attributes): ?>
-#[AsContentElement("<?= $elementName ?>", category: "<?= $category ?>")]
+#[AsContentElement(category: "<?= $category ?>")]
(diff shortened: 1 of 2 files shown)


m-vo:
Looks, good. I think we should omit the type for the annotation as well, though.

And, one other thing I noticed when testing: Should we add `miscellaneous` here, too?
image

aschempp (changes requested):
Just requesting changes for what @m-vo already said, the type should also be removed from annotations 😊

leofeyer:
Types removed in c60bda221faadbc943e340a30f17ea7c46ceed25. The suggested categories are taken from the `$GLOBALS['TL_CTE']` array, so as long as there are no elements in the `miscellaneous` category, it will not show up here.

aschempp (approved):

PR #4119, opened 2022-02-11
Fix error for parameter "_locale" for route "contao_frontend"

### LINKED ISSUE #4041: Parameter "_locale" for route "contao_frontend" must match "[a-z]{2}(?:\-[A-Z]{2})?" ("en_DE" given) to generate a corresponding URL.
### Affected version(s)

4.9.24

### Description

We have added support for localised languages such as de-AT to MetaModels. However, the routing throws an error because a locale is fixed with underscore see 

https://github.com/contao/contao/blob/6af1e0f1a670b84d1c4d1d788695b63d279ed21e/core-bundle/src/EventListener/LocaleSubscriber.php#L104

although it should be hyphenated.

The specification in the method is "**Returns** the locale from the request or the HTTP header." - however, a change is made

Why is `_locale` set to an underscore?

Stacktrace:

```
Symfony\Component\Routing\Exception\InvalidParameterException:
Parameter "_locale" for route "contao_frontend" must match "[a-z]{2}(?:\-[A-Z]{2})?" ("en_CA" given) to generate a corresponding URL.

  at vendor/symfony/routing/Generator/UrlGenerator.php:193
  at Symfony\Component\Routing\Generator\UrlGenerator->doGenerate()
     (vendor/symfony-cmf/routing/src/ProviderBasedGenerator.php:78)
  at Symfony\Cmf\Component\Routing\ProviderBasedGenerator->generate()
     (vendor/symfony-cmf/routing/src/DynamicRouter.php:180)
  at Symfony\Cmf\Component\Routing\DynamicRouter->generate()
     (vendor/symfony-cmf/routing/src/ChainRouter.php:248)
  at Symfony\Cmf\Component\Routing\ChainRouter->generate()
     (vendor/contao/core-bundle/src/Routing/UrlGenerator.php:81)
  at Contao\CoreBundle\Routing\UrlGenerator->generate()
     (vendor/metamodels/core/src/Filter/FilterUrlBuilder.php:142)
  at MetaModels\Filter\FilterUrlBuilder->generate()
     (vendor/metamodels/core/src/Helper/PaginationGenerator.php:298)
  at MetaModels\Helper\PaginationGenerator->linkToPage()
     (vendor/metamodels/core/src/Helper/PaginationGenerator.php:267)
  at MetaModels\Helper\PaginationGenerator->getItemsAsArray()
     (vendor/metamodels/core/src/Helper/PaginationGenerator.php:147)
  at MetaModels\Helper\PaginationGenerator->generateForFilterUrl()
     (vendor/metamodels/core/src/Helper/PaginationLimitCalculator.php:424)
  at MetaModels\Helper\PaginationLimitCalculator->getPaginationString()
     (vendor/metamodels/core/src/ItemList.php:867)
  at MetaModels\ItemList->getPagination()
     (vendor/metamodels/core/src/CoreBundle/Controller/ListControllerTrait.php:192)
  at MetaModels\CoreBundle\Controller\FrontendModule\ItemListController->getResponseInternal()
     (vendor/metamodels/core/src/CoreBundle/Controller/FrontendModule/ItemListController.php:96)
  at MetaModels\CoreBundle\Controller\FrontendModule\ItemListController->getResponse()
     (vendor/contao/core-bundle/src/Controller/FrontendModule/AbstractFrontendModuleController.php:40)
  at Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController->__invoke()
     (vendor/metamodels/core/src/CoreBundle/Controller/FrontendModule/ItemListController.php:64)
  at MetaModels\CoreBundle\Controller\FrontendModule\ItemListController->__invoke()
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (vendor/symfony/http-kernel/HttpCache/SubRequestHandler.php:85)
  at Symfony\Component\HttpKernel\HttpCache\SubRequestHandler::handle()
     (vendor/symfony/http-kernel/Fragment/InlineFragmentRenderer.php:81)
  at Symfony\Component\HttpKernel\Fragment\InlineFragmentRenderer->render()
     (vendor/symfony/http-kernel/Fragment/FragmentHandler.php:85)
  at Symfony\Component\HttpKernel\Fragment\FragmentHandler->render()
     (vendor/contao/core-bundle/src/Fragment/FragmentHandler.php:84)
  at Contao\CoreBundle\Fragment\FragmentHandler->render()
     (vendor/contao/core-bundle/src/Resources/contao/modules/ModuleProxy.php:61)
  at Contao\ModuleProxy->generate()
     (vendor/contao/core-bundle/src/Resources/contao/elements/ContentModule.php:101)
  at Contao\ContentModule->generate()
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:617)
  at Contao\Controller::getContentElement()
     (vendor/contao/core-bundle/src/Resources/contao/modules/ModuleArticle.php:218)
  at Contao\ModuleArticle->compile()
     (vendor/contao/core-bundle/src/Resources/contao/modules/Module.php:214)
  at Contao\Module->generate()
     (vendor/contao/core-bundle/src/Resources/contao/modules/ModuleArticle.php:71)
  at Contao\ModuleArticle->generate()
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:545)
  at Contao\Controller::getArticle()
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:386)
  at Contao\Controller::getFrontendModule()
     (vendor/contao/core-bundle/src/Resources/contao/pages/PageRegular.php:168)
  at Contao\PageRegular->prepare()
     (vendor/contao/core-bundle/src/Resources/contao/pages/PageRegular.php:48)
  at Contao\PageRegular->getResponse()
     (vendor/contao/core-bundle/src/Resources/contao/controllers/FrontendIndex.php:337)
  at Contao\FrontendIndex->renderPage()
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (web/index.php:31)
```

zonky2:
Fix error for parameter "_locale" for route "contao_frontend"

Fixes #4041

Thx at @discordier

### DIFF
--- a/core-bundle/src/Routing/UrlGenerator.php
+++ b/core-bundle/src/Routing/UrlGenerator.php
@@ -70,8 +70,9 @@ class UrlGenerator implements UrlGeneratorInterface
         $scheme = $context->getScheme();
         $httpPort = $context->getHttpPort();
         $httpsPort = $context->getHttpsPort();
+        $locale = $context->getParameter('_locale');
 
-        $this->prepareLocale($parameters);
+        $this->prepareLocale($parameters, $context);
         $this->prepareAlias($name, $parameters);
         $this->prepareDomain($context, $parameters, $referenceType);
 
@@ -88,6 +89,7 @@ class UrlGenerator implements UrlGeneratorInterface
         $context->setScheme($scheme);
         $context->setHttpPort($httpPort);
         $context->setHttpsPort($httpsPort);
+        $context->setParameter('_locale', $locale);
 
         return $url;
     }
@@ -95,10 +97,14 @@ class UrlGenerator implements UrlGeneratorInterface
     /**
      * Removes the locale parameter if it is disabled.
      */
-    private function prepareLocale(array &$parameters): void
+    private function prepareLocale(array &$parameters, RequestContext $context): void
     {
         if (!$this->prependLocale && \array_key_exists('_locale', $parameters)) {
             unset($parameters['_locale']);
+            return;
+        }
+        if ($this->prependLocale && $context->hasParameter('_locale')) {
+            $context->setParameter('_locale', str_replace('_', '-', $context->getParameter('_locale')));
         }
     }
 
--- a/core-bundle/src/Routing/UrlGenerator.php
+++ b/core-bundle/src/Routing/UrlGenerator.php
@@ -70,8 +70,9 @@ class UrlGenerator implements UrlGeneratorInterface
         $scheme = $context->getScheme();
         $httpPort = $context->getHttpPort();
         $httpsPort = $context->getHttpsPort();
+        $locale = $context->getParameter('_locale');
 
-        $this->prepareLocale($parameters);
+        $this->prepareLocale($parameters, $context);
         $this->prepareAlias($name, $parameters);
         $this->prepareDomain($context, $parameters, $referenceType);
 
@@ -88,6 +89,7 @@ class UrlGenerator implements UrlGeneratorInterface
         $context->setScheme($scheme);
         $context->setHttpPort($httpPort);
         $context->setHttpsPort($httpsPort);
+        $context->setParameter('_locale', $locale);
 
         return $url;
     }
@@ -95,10 +97,16 @@ class UrlGenerator implements UrlGeneratorInterface
     /**
      * Removes the locale parameter if it is disabled.
      */
-    private function prepareLocale(array &$parameters): void
+    private function prepareLocale(array &$parameters, RequestContext $context): void
     {
         if (!$this->prependLocale && \array_key_exists('_locale', $parameters)) {
             unset($parameters['_locale']);
+
+            return;
+        }
+
+        if ($this->prependLocale && $context->hasParameter('_locale')) {
+            $context->setParameter('_locale', str_replace('_', '-', $context->getParameter('_locale')));
         }
     }
 


zonky2 (commented):
fixed

aschempp (commented):
This feels strange. The original issue says the locale is not correctly formatted, so we discussed applying formatting. This PR now applies the locale from the request page to any URL being generated. How is that related and why would that make sense?

discordier:
> This feels strange. The original issue says the locale is not correctly formatted, so we discussed applying formatting. This PR now applies the locale from the request page to any URL being generated. How is that related and why would that make sense?

I try to clarify this, we have two kinds of locale formatting here.
1. Contao URL Format: `de-DE` (as specified by the URL parameter regex `[a-z]{2}(?:\-[A-Z]{2})?`).
2. Contao locale format:  `de_DE` as enforced in [LocaleSubscriber.php#104](https://github.com/contao/contao/blob/6af1e0f1a670b84d1c4d1d788695b63d279ed21e/core-bundle/src/EventListener/LocaleSubscriber.php#L104) by ensuring we only have an underscore.

The consumers in the application, like the translator component, should use the locale "as is", which makes sense.
However, for URL generating (Symfony URL generator) this will fail, as the fallback locale in the context is always using an underscore (as written in the original ticket).
So we now have 2 options:
1. also use the underscore locale notation in URLs and be done (unlikely to happen, all URLs will change).
2. fixup the locale right prior URL generating and reset it later on.
    We could implement this by enforcing an `_locale` parameter, however this will then append an query parameter to non-locale aware URLs and therefore I implemented it via context manipulation.

fritzmg:
But the context should have the locale already in the correct format, so the issue must be somewhere else.

aschempp:
>    We could implement this by enforcing an `_locale` parameter, however this will then append an query parameter to non-locale aware URLs and therefore I implemented it via context manipulation.

This is already how it works. That's why we unset the `_locale` parameter if it is not present in the URL. I think we should just do what @fritzmg said in https://github.com/contao/contao/issues/4041#issuecomment-1027988215, normalize the `_locale` parameter in the method instead of using it from the context.

discordier:
> > We could implement this by enforcing an `_locale` parameter, however this will then append an query parameter to non-locale aware URLs and therefore I implemented it via context manipulation.
> 
> This is already how it works. That's why we unset the `_locale` parameter if it is not present in the URL.

No, this is not how it works, trace the reproducer in the ticket, the `_locale` parameter is not present at all and therefore the value from the context is used (implicit fallback) in url generating which violates the constraint.
The problem is that the locale in the context is formatted `de_DE` and the url generator expects it to be `de-DE` (as that is the parameter constraint of the frontend route).

> I think we should just do what @fritzmg said in [#4041 (comment)](https://github.com/contao/contao/issues/4041#issuecomment-1027988215), normalize the `_locale` parameter in the method instead of using it from the context.

In that case, you would have to obtain the parameter from the context non the less, because that is the final fallback that get's used if no locale has been set in the parameters.
So the diff then becomes:
```diff
-    private function prepareLocale(array &$parameters): void
+    private function prepareLocale(array &$parameters, RequestContext $context): void
     {
         if (!$this->prependLocale && \array_key_exists('_locale', $parameters)) {
             unset($parameters['_locale']);
+
+            return;
+        }
+
+        if ($this->prependLocale && !\array_key_exists('_locale', $parameters)) {
+            $parameters['_locale'] = str_replace('_', '-', $context->getParameter('_locale'));
         }
     }

```

On a final note, I wonder how we shall fix this at all, as the `Contao\CoreBundle\Routing\UrlGenerator` has been deprecated and using the "normal" Symfony router will only use the value from the context when generating urls.
Therefore this will never work unless you always provide a locale, hence making `_locale` an always mandatory parameter, even when not needed at all in the final route/url.

aschempp:
I was not aware routes use the context for fallback parameters. But as far as I can tell, it was always intentional that you MUST pass the `_locale` parameter and Contao will remove it if not used. The `_locale` parameter depends on the URL you generate, never the URL the user is coming from (contrary to e.g. the host if that's not configured in the root settings).

Regarding the new routing, as far as I remember, we do not use the `_locale` parameter anymore - or better said the `PageRegistry` is already taking care of that. May I ask why you're not simply using `PageModel::getFrontendUrl` (which automatically does everything right already)?

discordier:
> I was not aware routes use the context for fallback parameters. But as far as I can tell, it was always intentional that you MUST pass the _locale parameter and Contao will remove it if not used. The _locale parameter depends on the URL you generate, never the URL the user is coming from (contrary to e.g. the host if that's not configured in the root settings).

In that sense, IMO it is wrong to name it `_locale`, as that is inferred from the context and breaks then - it must become a mandatory parameter then (not internal - hence the underscore prefix).

> Regarding the new routing, as far as I remember, we do not use the _locale parameter anymore - or better said the PageRegistry is already taking care of that.

Might be, I have no idea what is going on in there.

> May I ask why you're not simply using `PageModel::getFrontendUrl` (which automatically does everything right already)?

The real application is generating a combination of urls with slug and query parameters per URL, I fail to see how `PageModel::getFrontendUrl()` can do that for us?

It might be that we will have to revisit all of this for Contao 4.13 as the Contao route generator is deprecated - It would be nice to have a single way that will remain for 4.9, 4.13 and 5.0 and that supports:
1. slug parameters
2. query parameters
3. locale handling

I fail to see such a way currently.

fritzmg:
In theory you should be able to use the regular Symfony `router` for any circumstances, in Contao 4.9 as well as 4.13.

aschempp:
> > I was not aware routes use the context for fallback parameters. But as far as I can tell, it was always intentional that you MUST pass the _locale parameter and Contao will remove it if not used. The _locale parameter depends on the URL you generate, never the URL the user is coming from (contrary to e.g. the host if that's not configured in the root settings).
> 
> In that sense, IMO it is wrong to name it `_locale`, as that is inferred from the context and breaks then - it must become a mandatory parameter then (not internal - hence the underscore prefix).

I would guess _any_ parameter that existing in the query route is inferred from the context, nothing "hardcoded" to `_locale`? We use `_locale` because that's a default in Symfony. And be aware, the same parameters are used to generate AND match the route, so if we want `_locale` to work in the request, we must use it to generate a URL.


> The real application is generating a combination of urls with slug and query parameters per URL, I fail to see how `PageModel::getFrontendUrl()` can do that for us?
>  It would be nice to have a single way that will remain for 4.9, 4.13 and 5.0 

The best I can think of is to use `PageModel::getFrontendUrl()` and manually append your query parameters.

> In theory you should be able to use the regular Symfony `router` for any circumstances, in Contao 4.9 as well as 4.13.

I'm not sure, but I would say `router` does not (easily) work in Contao 4.9, because it does not have modern routing (`PageRoute` etc.).

pagestyles:
Hi, I have a customer who is affected by the situation, somehow there seems to be no agreement here?

leofeyer:
@pagestyles Does the code fix your customer‘s issue?

pagestyles:
sorry - yes it did

aschempp:
the code changes the behaviour of the UrlGenerator to fix an issue introduced by MetaModels (as far as I understand). But the behaviour change is wrong imho, the issue should be fixed in MetaModels. I'm open to discussing (or explaining) this though (preferably in our voice calls) 😊

zonky2:
The problem comes from Contao and was **not** introduced by MetaModels - it was only noticed via MetaModels...

see https://github.com/contao/contao/issues/4041#issuecomment-1027977285 and https://github.com/contao/contao/issues/4041#issuecomment-1027988215

If Contao doesn't want to or can't fix this, we would have to implement a workaround in MetaModels by necessity. :(

But then Contao would still have the problem!

aschempp:
The problem I understand is that MetaModels is not passing the current page language to the `UrlGenerator`, and the side effect is that Symfony then takes the data from the request context, which has a wrong format. But this side effect is not supported, you must always pass the language of the target page! Even more so because the target page language is absolutely not related to the current request language.

leofeyer:
As discussed in the Contao call, the fix in this PR (always setting the `_locale` parameter)  is wrong.

@discordier Can you fix this in MetaModels? If not, can you join the next Contao call so we work on a solution together?

dmolineus:
> But this side effect is not supported, you must always pass the language of the target page! 

So there should be at least a check of required parameters so that a developer get a chance to be informed the missing parameter. Furthermore a phpdoc for it would help also. Especially for Symfony developer who are aware of the usage of the context parameters for routing the current situation is IMHO really misleading.

zonky2:
> So there should be at least a check of required parameters so that a developer get a chance to be informed the missing parameter. Furthermore a phpdoc for it would help also. Especially for Symfony developer who are aware of the usage of the context parameters for routing the current situation is IMHO really misleading.

@leofeyer:
How do we proceed with the PR? The statement that the PR would "destroy" the functionality of Contao I could not find after the installation. At which point does something break or no longer work?

Will Contao include this PR or another fix or do we have to by necessity circumvent this in MM?

cc @discordier

aschempp:
> The statement that the PR would "destroy" the functionality of Contao I could not find after the installation. At which point does something break or no longer work?

Did you try a multilingual website and link from one language tree to another language? In a system that actually uses this class (not sure if Contao 4.9 does)?

> Will Contao include this PR or another fix or do we have to by necessity circumvent this in MM?

You don't "circument", you fix the bug in MM 😉

zonky2:
We add a workaround in MM 2.2

PR #4120, opened 2022-02-11
Fix some tests for order-by=reverse and repeat=2

### LINKED ISSUE #2316: Tests should not rely on execution order

ausi:
Small step in the direction of #2316

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,7 +249,7 @@ jobs:
                   DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
     reverse:
-        name: PHP 8.1 reverse order
+        name: PHP 8.1 Reverse Order
         runs-on: ubuntu-latest
         if: github.event_name != 'push'
         steps:
@@ -434,9 +434,9 @@ jobs:
               run: tools/monorepo/vendor/bin/monorepo-tools composer-json --validate --ansi
 
     inspect:
-        name: Inspect unit tests
+        name: Inspect Unit Tests
         runs-on: ubuntu-latest
-        #if: github.event_name == 'schedule'
+        if: github.event_name == 'schedule'
         continue-on-error: true
         steps:
             - name: Setup PHP
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -262,11 +262,6 @@ jobs:
               env:
                   COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-            - name: Initialize the database
-              run: |
-                  sudo /etc/init.d/mysql start
-                  mysql -uroot -proot -e "CREATE database contao_test"
-
             - name: Checkout
               uses: actions/checkout@v1
 
@@ -275,11 +270,6 @@ jobs:
 
             - name: Run the unit tests in reverse order
               run: vendor/bin/phpunit --order-by=reverse --colors=always
-
-            - name: Run the functional tests in reverse order
-              run: vendor/bin/phpunit --testsuite=functional --order-by=reverse --colors=always
-              env:
-                  DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
     nightly:
         name: PHP 8.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,6 +248,9 @@ jobs:
               env:
                   DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
+            - name: Run the unit tests in reverse order
+              run: vendor/bin/phpunit --order-by=reverse --colors=always
+
     nightly:
         name: PHP 8.2
         runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,8 +248,38 @@ jobs:
               env:
                   DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
+    reverse:
+        name: PHP 8.1 reverse order
+        runs-on: ubuntu-latest
+        if: github.event_name != 'push'
+        steps:
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 8.1
+                  extensions: dom, fileinfo, filter, gd, hash, intl, json, mbstring, mysqli, pcre, pdo_mysql, zlib
+                  coverage: none
+              env:
+                  COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Initialize the database
+              run: |
+                  sudo /etc/init.d/mysql start
+                  mysql -uroot -proot -e "CREATE database contao_test"
+
+            - name: Checkout
+              uses: actions/checkout@v1
+
+            - name: Install the dependencies
+              run: composer install --no-interaction --no-progress
+
             - name: Run the unit tests in reverse order
               run: vendor/bin/phpunit --order-by=reverse --colors=always
+
+            - name: Run the functional tests in reverse order
+              run: vendor/bin/phpunit --testsuite=functional --order-by=reverse --colors=always
+              env:
+                  DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
     nightly:
         name: PHP 8.2
@@ -412,6 +442,35 @@ jobs:
 
             - name: Validate the composer.json files
               run: tools/monorepo/vendor/bin/monorepo-tools composer-json --validate --ansi
+
+    inspect:
+        name: Inspect unit tests
+        runs-on: ubuntu-latest
+        #if: github.event_name == 'schedule'
+        continue-on-error: true
+        steps:
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 8.1
+                  extensions: dom, fileinfo, filter, gd, hash, intl, json, mbstring, mysqli, pcre, pdo_mysql, zlib
+                  coverage: none
+              env:
+                  COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Checkout
+              uses: actions/checkout@v1
+
+            - name: Install the dependencies
+              run: |
+                  composer install --no-interaction --no-progress
+                  composer bin isolated-tests install --no-interaction --no-progress
+
+            - name: Run the unit tests
+              run: vendor/bin/phpunit --extensions Contao\\CoreBundle\\Tests\\PhpunitExtension\\GlobalStateWatcher --colors=always
+
+            - name: Run the unit tests isolated
+              run: tools/isolated-tests/bin/run-tests-isolated --ansi
 
     monorepo-split:
         name: Monorepo Split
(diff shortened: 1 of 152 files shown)


.github/workflows/ci.yml
```diff
@@ -248,6 +248,9 @@ jobs:
               env:
                   DATABASE_URL: mysql://root:root@127.0.0.1:3306/contao_test
 
+            - name: Run the unit tests in reverse order
+              run: vendor/bin/phpunit --order-by=reverse --colors=always
+
```
ausi, line 273: We should probably create an extra job that only runs in the newest php version.

core-bundle/tests/TestCase.php
```diff
@@ -22,6 +25,60 @@
...
+        $prop->setAccessible(true);
+        $prop->setValue(null);
+
+        $prop = (new \ReflectionClass(Model::class))->getProperty('arrClassNames');
+        $prop->setAccessible(true);
+        $prop->setValue([]);
+
+        unset($_SESSION, $GLOBALS['TL_HEAD'], $GLOBALS['TL_CONFIG'], $GLOBALS['TL_LANG'], $GLOBALS['TL_LANGUAGE'], $GLOBALS['TL_MIME'], $GLOBALS['TL_DCA'], $GLOBALS['TL_HOOKS'], $GLOBALS['BE_MOD'], $GLOBALS['FE_MOD'], $GLOBALS['TL_USERNAME'], $GLOBALS['objPage']);
+
+        static::resetStaticProperties();
```
ausi: This now resets all static properties of all classes after each test.

I’d like to add this call to every `tearDown()` that needs it, and pass the classes into the method that should get reset.

core-bundle/tests/PhpunitExtension/GlobalStateWatcher.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Contao\CoreBundle\Tests\PhpunitExtension;
+
+use PHPUnit\Framework\ExpectationFailedException;
+use PHPUnit\Runner\AfterTestHook;
+use PHPUnit\Runner\BeforeTestHook;
+use SebastianBergmann\Diff\Differ;
+use SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder;
+
+final class GlobalStateWatcher implements AfterTestHook, BeforeTestHook
```
ausi, line 21: With this extension you can execute

```sh
vendor/bin/phpunit --extensions=Contao\\CoreBundle\\Tests\\PhpunitExtension\\GlobalStateWatcher
```

and you will get an error if a test leaks global state via `$GLOBALS` or static class properties.

core-bundle/tests/TestCase.php
ausi: Done.

ausi:
This pull request is in pretty good shape now I think. 
`vendor/bin/phpunit --order-by=random` seems fine 🎉

I added a PHPUnit extension that watches for modifications to the global state in tests (`$GLOBALS`, `static $members`, `CONSTANTS`, php.ini, the file system and ENV). It can be executed by calling `vendor/bin/phpunit --extensions=Contao\\CoreBundle\\Tests\\PhpunitExtension\\GlobalStateWatcher` and fails if a unit test doesn’t clean up its changes correctly.

There is also a `tools/isolated-tests/bin/run-tests-isolated` command now that runs every test class, every test method and every single test data set isolated in a single execution of `phpunit`. This is great to find tests that accidentally depend on others (but is very slow).

The new `resetStaticProperties()` helper method can be used in unit tests to reset static properties of classes in the `tearDown()` method.

If you’re happy with these changes I would move `resetStaticProperties()` to `contao/test-case` (this will resolve the failing CI job “Bundles”).

leofeyer:
> If you’re happy with these changes I would move `resetStaticProperties()` to `contao/test-case` (this will resolve the failing CI job “Bundles”).

Yes. 👍

tools/isolated-tests/src/RunTestsIsolatedCommand.php
```diff
@@ -0,0 +1,172 @@
...
+            $output->writeln("> vendor/bin/phpunit --filter '".str_replace("'", '\'"\'"\'', $filter)."'");
+
+            try {
+                $this->executeCommand(array_merge($php, [$phpunit, '--extensions', GlobalStateWatcher::class, '--filter', $filter], $commandFlags), $buffer = new BufferedOutput());
+
+                // Clear previously written line
+                $output->write("\e[1A\e[K");
+            } catch (\Throwable $e) {
+                $failedTests[] = $test;
+                $output->writeln($buffer->fetch());
```
leofeyer, line 131: Do we need to handle this?

leofeyer (dismissed):

core-bundle/tests/TestCase.php
```diff
@@ -15,13 +15,95 @@
...
+        $_SERVER = $this->backupServerEnvGetPost['_SERVER'] ?? $_SERVER;
+        $_ENV = $this->backupServerEnvGetPost['_ENV'] ?? $_ENV;
+        $_GET = $this->backupServerEnvGetPost['_GET'] ?? [];
+        $_POST = $this->backupServerEnvGetPost['_POST'] ?? [];
+    }
+
+    /**
+     * @param array<int, class-string|array{0: class-string, 1: array<int, string>}> $classNames
+     */
+    protected function resetStaticProperties(array $classNames): void
```
aschempp: This looks suspicious to me. Most of these classes have a `reset` method that is called by the `ContaoFramework`. We should call that method instead, and add `reset` methods if some are missing (e.g. the `System` class) to make sure it works consistently.

aschempp (changes requested):

tools/isolated-tests/src/RunTestsIsolatedCommand.php
ausi: 👍
Fixed in 8b9242045b9fdf7e0bc4f3f347b4d44412eda5a1

core-bundle/tests/TestCase.php
ausi: As this is only used in the unit tests, using reflection should be OK here IMO.
Also, several tests need to reset third party classes like `Request`, `Terminal`, `Process` and so on.

But we could implement a `method_exists($class, 'reset')` and call `$class::reset()` if it exists here, WDYT?

m-vo: By the way: I tried this in #2518, but it got a convoluted complicated thing and probably not worth for the legacy code.  :see_no_evil: While I generally agree with @aschempp, I think it's probably most helpful adjusting it in the tests with reflection here. :+1:

m-vo: > But we could implement a method_exists($class, 'reset') and call $class::reset() if it exists here, WDYT?

It should be sufficient to call `ContaoFramework#reset()` but unfortunately this is a class method.

tools/isolated-tests/src/RunTestsIsolatedCommand.php
```diff
@@ -0,0 +1,174 @@
...
+use Contao\CoreBundle\Tests\PhpunitExtension\GlobalStateWatcher;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\BufferedOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\PhpExecutableFinder;
+use Symfony\Component\Process\Process;
+
+class RunTestsIsolatedCommand extends Command
```
Toflar, line 24: What's the difference to `phpunit --process-isolation`?

ausi: That no state is shared at all. I tried `--process-isolation` but it delivered different results.

Toflar: Okay, then I think the command description or anything in this whole tool should say so :) So that other devs know why we have our own tool too :)

core-bundle/tests/TestCase.php
ausi: How would we reset the three properties from here https://github.com/symfony/symfony/blob/44db49fb8f6d70c4341120a3f8f1b1a1a0dfa014/src/Symfony/Component/Console/Terminal.php#L16-L18 ?

tools/isolated-tests/src/RunTestsIsolatedCommand.php
ausi: Done in f75f185617c7decb606895815f09aa85f29625fe

Toflar (dismissed):
❤️

tools/isolated-tests/src/RunTestsIsolatedCommand.php
```diff
@@ -0,0 +1,179 @@
...
+
+    protected function configure(): void
+    {
+        $this->setDescription(
+            'Runs every test completely isolated from the other by starting a '
+            .'new phpunit process for every test class, method or data set. '
+            .'This gets us “real” isolation and no shared state at all in '
+            .'contrast to the phpunit --process-isolation option.'
+        );
+        $this->addOption('depth', null, InputOption::VALUE_REQUIRED, '1 for test classes, 2 for test methods, 3 for every single provider data set', '3');
```
leofeyer:
```suggestion
        $this->setDescription('Runs the unit tests isolated from each other.');
        $this->addOption('depth', null, InputOption::VALUE_REQUIRED, '1 for test classes, 2 for test methods, 3 for every single provider data set', '3');

        $this->setHelp(<<<EOT
            The command runs each unit test completely isolated from the others, starting
            a new PHPUnit process for each test class, method, or data set. This gives us
            "real" isolation rather than shared state, unlike the PHPUnit option
            --process-isolation does.
            EOT
        );
```

core-bundle/tests/TestCase.php
ausi: > But we could implement a `method_exists($class, 'reset')` and call `$class::reset()` if it exists

Implemented in 042e1a647d543a33efbeb3a001348f2528b718ba

m-vo (dismissed):
:heart_eyes_cat:

m-vo (dismissed):
:heart_eyes_cat:

leofeyer (dismissed):

ausi:
https://github.com/contao/test-case/pull/16 has to be merged and released as version 4.6.0 first. After that, all tests here should be green 🤞

leofeyer (approved):

leofeyer:
Great work! Thank you very much @ausi.

Issue #2491, opened 2020-11-10
Link to Contao Manual (Documentation) within Backend


fkaminski:
Maybe it would be helpful to provide a link to the documentation in the Contao backend ...

fritzmg:
There already is a link on the dashboard (for the keyboard shortcuts). However, may be it would be good to overhaul the dashboard in that regards in general?

bennyborn:
> However, may be it would be good to overhaul the dashboard in that regards in general?

+1 from me. We've had quite some projects were we added custom listings to the dashboard by simply overwriting `be_welcome.html5` since there is no way to add more content to it in an update-safe way.

ameotoko:
> There already is a link on the dashboard (for the keyboard shortcuts). However, may be it would be good to overhaul the dashboard in that regards in general?

BTW, this link is broken in https://github.com/contao/contao/blob/7e6bacd4e334bdd1a32996dab12d3b05e1e161cc/core-bundle/src/Resources/contao/languages/en/default.xlf#L1148-L1153

Currently working link is https://docs.contao.org/manual/en/administrationsbereich/keyboard-shortcuts/, but it doesn't look right either, may be **it** should be updated first?

fritzmg:
@ameotoko that's actually an error in the Contao docs which I have now rectified: https://github.com/contao/docs/commit/094febbd58089f2a0e905e15085d54589cd706e3

Thanks for reporting :)

leofeyer:
As discussed in Mumble on December 17th, we want to add a `?` icon in the upper right corner that links to to.contao.org.

MDevster:
@leofeyer should the icon link to support section (https://to.contao.org/support => https://contao.org/de/support.html) or direct to manual (https://to.contao.org/manual => https://docs.contao.org/manual/)?

leofeyer:
Directly to the manual I think.

MDevster:
I would take this

aschempp:
see https://github.com/contao/contao/pull/4123

PR #4124, opened 2022-02-13
use contao short url service for keyboard shortcuts

MDevster:
Fixes https://github.com/contao/contao/issues/2491#issuecomment-738877348 and facilitates the change for future alias adjustments.

Requires that contao/to.contao.org#46 is merged!
The link should also be adapted in the German language on Transifex.


aschempp:
Nice catch! Since the link no longer needs to be translated, I usually go with `%s` in the translation and insert the link where the label is generated. Not sure what others think about that though @contao/developers ?

fritzmg (commented):
Does this link also exist in Contao 4.9? If yes, then this should be targeted against the `4.9` branch.

leofeyer:
See #4229 for a pull for Contao 4.9.

PR #4133, opened 2022-02-14
Improve the preview links back end

### LINKED ISSUE #3974: Preview link improvements
Currently when clicking _Share URL_ in the preview bar, the back end opens in a new tab in order to create a new `tl_preview_link` entry. However, after saving, you will always have to manually click on the _Preview links_ main menu entry in the back end, in order to actually be able to share this URL (_Save and close_ will likely send you somewhere else, unless you happened to have clicked on _Preview links_ before clicking _Share URL_). Furthermore, once more preview link entries are in that view, it will not be easy to find out which of these preview links you just generated and thus want to share.

I wanted to discuss how we could improve this. May be an additional button _Save and copy preview link_?

### LINKED ISSUE #4007: Copy preview link: navigator.clipboard is undefined
### Affected version(s)

4.13.x-dev

### Description

We currently use `navigator.clipboard` in https://github.com/contao/contao/blob/862755b972e2225fce3237da8f154a5299ff5a63/core-bundle/src/EventListener/DataContainer/PreviewLinkListener.php#L172 but this is not available in all browsers and not in all contexts (HTTPS only).

I think we should use something like what we have in the preview toolbar instead: https://github.com/contao/contao/blob/64952eea57830f86ff95928f3810b5e5aac6d169/core-bundle/src/Resources/views/Frontend/preview_toolbar_base_js.html.twig#L330-L346

### LINKED ISSUE #4100: Preview links should auto-update "Valid until"
### Affected version(s)

4.13-RC2

### Description

When I create a new link and change the days for the expiration in the select field, the `valid until field` is not updated dynamically. It only would be updated, when I click on "save".

Is this how it should be?
Can this be made more user-friendly?
screen-20220210_i6tUE6zn

### LINKED ISSUE #4101: Preview links - Add hint for new entries
### Affected version(s)

4.13-RC2

### Description

Would it be possible to add a small hint that new links could be created via the front preview? From the user's point of view, I'm missing the "New" button here.

Similar to Template hints
screen-20220210_J7d0D3Xk

aschempp:
1. fixes https://github.com/contao/contao/issues/4007

    The clipboard javascript is now a separate script triggered by `data-to-clipboard` attribute and uses the clipboard fallback from our preview bar.

2. fixes https://github.com/contao/contao/issues/4100

    `submitOnChange` added to the expiration select menu

3. fixes https://github.com/contao/contao/issues/3974

    Using `nb=1` hides the back button if the user clicks from the front end. That's a standard Contao feature we also use in the popups. Imho "Save & Close" should still be there if a user navigates from the back end module. I didn't add a "Save & Add to clipboard" button because _next point_
 
4. fixes https://github.com/contao/contao/issues/4101

    Added several hints to improve UX. 

    - in the list view
        Bildschirmfoto 2022-02-14 um 15 52 20

    - for new links or when not enabled in edit mode
        Bildschirmfoto 2022-02-14 um 15 53 15

    - when link has expired
        Bildschirmfoto 2022-02-14 um 15 53 57

    - when link is enabled and available, the share link is displayed in the hint and can be clicked. Clicking it uses the `data-to-clipboard` again to add the link to your clipboard.
        Bildschirmfoto 2022-02-14 um 15 54 11

5. Additionally disabled the clipboard links if the link has expired
    Bildschirmfoto 2022-02-14 um 15 55 42

6. Also implemented an ajax confirmation message based on our existing simple modal
    Bildschirmfoto 2022-02-14 um 15 55 48


core-bundle/src/Resources/public/clipboard.js
```diff
@@ -0,0 +1,35 @@
...
+        link.addEventListener('click', function (event) {
+            event.preventDefault();
+
+            if (navigator.clipboard && navigator.clipboard.writeText) {
+                navigator.clipboard.writeText(data.content).catch(clipboardFallback);
+            } else {
+                clipboardFallback();
+            }
+
+            Backend.openModalWindow(380, data.title, '<div style="padding: 20px 12px">'+data.message+'</div>');
```
ausi:
```suggestion
                navigator.clipboard.writeText(data.content).catch(clipboardFallback).finally(function () {
                    Backend.openModalWindow(380, data.title, '<div style="padding: 20px 12px">'+data.message+'</div>');
                });
            } else {
                clipboardFallback();
                Backend.openModalWindow(380, data.title, '<div style="padding: 20px 12px">'+data.message+'</div>');
            }
```
The modal should not be opened before the promise from `navigator.clipboard.writeText` was resolved.

aschempp: I agree that's technically right, but do you really think we need that for the split-second where the data is asynchronously added to the clipboard? I also inflicts on the opposite direction, the user has no screen action when clicking the button _until_ the data is added to the clipboard. I think that's more annoying than the likeliness of trying to paste before async is finished?

ausi: Don’t know all browser implementations, but some may be asking the user if they allow access to the clipboard. Showing a success message before the user clicked on “allow” is confusing IMO.

I’m fine either way. I just noticed this when reviewing the code...

leofeyer (changes requested):
I still do not like the modal window. We need something non permanent that does not require to be closed manually. Either a flash message or an inline confirmation.

aschempp:
> I still do not like the modal window. We need something non permanent that does not require to be closed manually. Either a flash message or an inline confirmation.

I agree this could be improved, but I think the current state is already much better than what it was before. We might want to add toast messages to the back end, but I didn't want to open that discussion in this PR and stick to the existing implementations 🤷

leofeyer:
Since Contao 4.13.0 is scheduled to be released today, I have reverted the changes in question for now so I can merge the other fixes. Sorry about that.

I have also created a new issue (see #4147) in which we can work on a solution.

aschempp:
I don't understand why having no modal should be better than having a "not ideal" modal we can improve on 😞

PR #4134, opened 2022-02-14
Remove custom template option

bytehead:
The module does not use a template at all, so no custom option is needed here.


leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #4135, opened 2022-02-14
Do not configure the default clickjacking paths in the skeleton

fritzmg:
Suppose you want to allow a certain URL (or URLs) to be used as an `<iframe>` on another domain. What you would normally do is the following in your `config/config.yaml`:

```yaml
nelmio_security:
    clickjacking:
        paths:
            '^/foobar/.+': ALLOW
            '^/.*': SAMEORIGIN
```

However, this does not actually work. The reason is the loading of the `skeleton/config.yml` of the Manager Bundle. It messes up the order of paths, causing `'^/.*': SAMEORIGIN` to always be the first element in the resulting config array after the configurations were processed by Symfony. Therefore the `X-Frame-Options` response header will always contain `SAMEORIGIN` and changing that header for a specific URL will have no effect - you can only set it to `ALLOW` globally.

To fix this, the default `'^/.*': SAMEORIGIN` value (needed for the Contao back end) should only be set via `ExtensionPluginInterface`, if the rule for that path was not defined yet.


fritzmg:
For anyone needing a workaround right now (other than allowing any URL to be used in an `<iframe>`):

```php
// src/ContaoManager/Plugin.php
namespace App\ContaoManager;

use Contao\ManagerPlugin\Config\ExtensionPluginInterface;

class Plugin implements ExtensionPluginInterface
{
    public function getExtensionConfig($extensionName, array $extensionConfigs, ContainerBuilder $container)
    {
        switch ($extensionName) {
            case 'nelmio_security': return $this->addClickJackingPaths($extensionConfigs);
        }

        return $extensionConfigs;
    }

    private function addClickJackingPaths(array $extensionConfigs): array
    {
        array_unshift($extensionConfigs, [
            'clickjacking' => [
                'paths' => [
                    '^/foobar/.+' => 'ALLOW',
                ],
            ],
        ]);

        return $extensionConfigs;
    }
}
```

ausi (dismissed):

aschempp (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

Issue #4131, opened 2022-02-14
Installation of 4.13.0-RC3 not possible?


bennyborn:
### Affected version(s)

4.13.0-RC3

### Description

We've upgraded our local test installation from 4.12 to 4.13.0-RC3 without any problems but currently it seems to be impossible to install 4.13 from scratch using `composer create-project contao/managed-edition . 4.13.x-dev`. 

Composer is doing his job so far but trying to use the `contao-console` or the Install-Tool fails because there is no way to configure the database connection. 

I know how the `parameters.yml` should look like but is this really intended or just a bug?

fritzmg:
What exactly is the problem?

bennyborn:
Created the project using `composer create-project contao/managed-edition . 4.13.x-dev`, went through without errors. Tried to open `/contao/install` and failed with the following log entry

```
[2022-02-14T15:22:11.335471+01:00] request.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Exception\ConnectionException: "An exception occurred in the driver: SQLSTATE[HY000] [2002] No such file or directory" at /home/benny/www/413test/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php line 103 {"exception":"[object] (Doctrine\\DBAL\\Exception\\ConnectionException(code: 2002): An exception occurred in the driver: SQLSTATE[HY000] [2002] No such file or directory at /home/benny/www/413test/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:103)\n[previous exception] [object] (Doctrine\\DBAL\\Driver\\PDO\\Exception(code: 2002): SQLSTATE[HY000] [2002] No such file or directory at /home/benny/www/413test/vendor/doctrine/dbal/src/Driver/PDO/Exception.php:30)\n[previous exception] [object] (PDOException(code: 2002): SQLSTATE[HY000] [2002] No such file or directory at /home/benny/www/413test/vendor/doctrine/dbal/src/Driver/PDO/MySQL/Driver.php:28)"} []
```

Also trying to use the `contao-console` command

```benny@devel:~/www/413test$ vendor/bin/contao-console
Contao Managed Edition 4.13.0-RC3 (env: prod, debug: false)

Usage:
  command [options] [arguments]

Options:
  -h, --help            Display help for the given command. When no command is given display help for the list command
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
[...]

PR #4149, opened 2022-02-17
Use static description for commands

m-vo:
From Symfony 5 on, commands should have a static `$defaultDescription` similar to the `$defaultName`, so that they do not have to get instantiated when using the `list` command.

### DIFF
--- a/core-bundle/src/Command/CrawlCommand.php
+++ b/core-bundle/src/Command/CrawlCommand.php
@@ -45,7 +45,7 @@ use Terminal42\Escargot\Subscriber\SubscriberInterface;
 class CrawlCommand extends Command
 {
     protected static $defaultName = 'contao:crawl';
-    protected static $defaultDescription = 'Crawls the Contao root pages with the desired subscribers';
+    protected static $defaultDescription = 'Crawls the Contao root pages with the desired subscribers.';
 
     private Factory $escargotFactory;
     private Filesystem $filesystem;
--- a/core-bundle/src/Command/CrawlCommand.php
+++ b/core-bundle/src/Command/CrawlCommand.php
@@ -45,6 +45,7 @@ use Terminal42\Escargot\Subscriber\SubscriberInterface;
 class CrawlCommand extends Command
 {
     protected static $defaultName = 'contao:crawl';
+    protected static $defaultDescription = 'Crawls the Contao root pages with the desired subscribers';
 
     private Factory $escargotFactory;
     private Filesystem $filesystem;
@@ -75,7 +76,6 @@ class CrawlCommand extends Command
             ->addOption('no-progress', null, InputOption::VALUE_NONE, 'Disables the progress bar output')
             ->addOption('enable-debug-csv', null, InputOption::VALUE_NONE, 'Writes the crawl debug log into a separate CSV file')
             ->addOption('debug-csv-path', null, InputOption::VALUE_REQUIRED, 'The path of the debug log CSV file', Path::join(getcwd(), 'crawl_debug_log.csv'))
-            ->setDescription('Crawls the Contao root pages with the desired subscribers')
             ->setHelp('You can add additional URIs via the <info>contao.crawl.additional_uris</info> parameter.')
         ;
     }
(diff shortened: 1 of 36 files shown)


core-bundle/src/Command/CrawlCommand.php
```diff
@@ -45,6 +45,7 @@
 class CrawlCommand extends Command
 {
     protected static $defaultName = 'contao:crawl';
+    protected static $defaultDescription = 'Crawls the Contao root pages with the desired subscribers';
```
fritzmg:
```suggestion
    protected static $defaultDescription = 'Crawls the Contao root pages with the desired subscribers.';
```

core-bundle/src/Command/InstallCommand.php
```diff
@@ -28,6 +28,7 @@
 class InstallCommand extends Command
 {
     protected static $defaultName = 'contao:install';
+    protected static $defaultDescription = 'Installs the required Contao directories';
```
fritzmg:
```suggestion
    protected static $defaultDescription = 'Installs the required Contao directories.';
```

aschempp:
If I run `console list --format=json` I also get all arguments and options. Can this really work with this?

manager-bundle/src/Command/DebugPluginsCommand.php
```diff
@@ -40,6 +40,7 @@
 class DebugPluginsCommand extends Command
 {
     protected static $defaultName = 'debug:plugins';
+    protected static $defaultDescription = 'Displays the Contao Manager plugin configurations';
```
fritzmg:
```suggestion
    protected static $defaultDescription = 'Displays the Contao Manager plugin configurations.';
```

manager-bundle/src/Command/InstallWebDirCommand.php
```diff
@@ -27,6 +27,9 @@
  */
 class InstallWebDirCommand extends Command
 {
+    protected static $defaultName = 'contao:install-web-dir';
+    protected static $defaultDescription = 'Installs the files in the public directory';
```
fritzmg:
```suggestion
    protected static $defaultDescription = 'Installs the files in the public directory.';
```

fritzmg (dismissed):
👍 

We could use this opportunity to unify the descriptions a bit (i.e. add missing periods).

m-vo:
> If I run console list --format=json I also get all arguments and options. Can this really work with this?

I don't think using `--format=json` will use that. It's primarily for humans, see https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description.

bytehead (dismissed):

ausi:
> If I run `console list --format=json` I also get all arguments and options. Can this really work with this?

`--format=json` is still broken in my test, but without `--format=json` it works now with wrong DB credentials.

ausi (dismissed):

fritzmg:
> `--format=json` is still broken in my test, but without `--format=json` it works now with wrong DB credentials.

Yes, this is not a fix for #4131 anyway.

ausi (dismissed):

fritzmg (dismissed):

leofeyer (dismissed):

fritzmg:
@leofeyer why the parent call? The parent configure() method is empty.

leofeyer:
Consistency. The call is already present in about 50% of the classes.

I‘d be fine with removing it everywhere, too.

m-vo:
> 

But it could have content in the future. IMHO it's correct to have the call.

fritzmg:
I see, I haven't realised. Yes, I would remove them (when just extending from `Symfony\Component\Console\Command\Command` and not another command). The [Symfony Docs](https://symfony.com/doc/current/console.html) also do not use the parent call for configure.

m-vo:
> The [Symfony Docs](https://symfony.com/doc/current/console.html) also do not use the parent call for configure.

Interesting, that might have changed. 😊 All good for removing then 👍

leofeyer:
Removed in 4cc1e093a44dbec64a6fde92fa6ea050b5e47b1f.

leofeyer (dismissed):

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4150, opened 2022-02-17
Improve sorting in the system log

### LINKED ISSUE #4148: Improve sorting in System-Log
### Description

By default the System-Log is sorted and grouped by `tstamp`. This is all fine for the usual workflow when using Contao.

In a couple of our projects we've implemented imports (or other routines) which do some logging into the System-Log and might result in 10 or 20 concurrent entries within the same second. Being sorted by `tstamp` rather than `id` this significantly messes with the order of the entries.

Any chance to improve this? I don't see any easy solution because altering the sorting also always messes up the grouping.

bennyborn:
Fixes #4148


core-bundle/src/Resources/contao/dca/tl_log.php
```diff
@@ -35,7 +35,7 @@
 		'sorting' => array
 		(
 			'mode'                    => DataContainer::MODE_SORTABLE,
-			'fields'                  => array('tstamp'),
+			'fields'                  => array('tstamp','id'),
```
fritzmg:
```suggestion
			'fields'                  => array('tstamp', 'id'),
```

fritzmg:
If we do this as a bugfix, it should be targeted against `4.9`.

bennyborn:
> If we do this as a bugfix, it should be targeted against `4.9`.

I actually considered this more as a feature than a bug. I mean, it's not broken but it could be better 🤷‍♂️

m-vo:
> I actually considered this more as a feature than a bug. I mean, it's not broken but it could be better man_shrugging

Then, this should be targeted at 5.x. :see_no_evil:

bennyborn:
> Then, this should be targeted at 5.x. 🙈

Hmpf 🤔 Is there nothing in between? 
As small as this change is I really wouln't want to wait until the next LTS for this to be included but I also see no benefit adding this back into 4.9 with the 4.13 being released (possibly) today?

fritzmg:
> Hmpf 🤔 Is there nothing in between?

Contao 4.13 already has a feature freeze. So if we consider this a feature, it must go into 5.x. If we consider this a bug, it must go into 4.9.

> I also see no benefit adding this back into 4.9

Contao 4.9's (bugfix) support ends in almost exactly one year (see https://contao.org/en/release-plan.html). Why do you see no benefit in adding this to Contao 4.9?

> As small as this change is I really wouln't want to wait until the next LTS

Not sure what you mean. You don't have to wait for the next LTS, you can just use Contao 5.0. And you can use the DCA adjustment in _any_ Contao version.

bennyborn:
> Contao 4.9's (bugfix) support ends in almost exactly one year (see https://contao.org/en/release-plan.html). Why do you see no benefit in adding this to Contao 4.9?
[...]
Not sure what you mean. You don't have to wait for the next LTS,

I guess it all worked out a bit different in my head.
So considering this a bugfix now, rebased to 4.9 it should be implemented in 4.9, 4.13 and of course in upcoming versions right?

> you can just use Contao 5.0. And you can use the DCA adjustment in _any_ Contao version.

For clarification: We only use LTS versions, that's why didn't want to wait for 5.0

fritzmg:
> So considering this a bugfix now, rebased to 4.9 it should be implemented in 4.9, 4.13 and of course in upcoming versions right?

Everything will always be merged upstream, yes (if applicable).

aschempp (approved):

leofeyer:
Thank you @bennyborn.

PR #4151, opened 2022-02-17
Make the `crontao.cron` service lazy

aschempp:
Fixes https://github.com/contao/contao/issues/4131

Making `crontao.cron` lazy will result in the Doctrine repository not being created. Creating it if the database connection is not configured would fail.

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -98,6 +98,7 @@ services:
     contao.cron:
         class: Contao\CoreBundle\Cron\Cron
         public: true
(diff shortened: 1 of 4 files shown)


core-bundle/src/Resources/config/services.yml
```diff
@@ -98,6 +98,7 @@ services:
     contao.cron:
         class: Contao\CoreBundle\Cron\Cron
         public: true
+        lazy: true # defer creating the doctrine repository for when the database connection is not configured
```
```suggestion
        lazy: true # Defer creating the Doctrine repository in case the DB connection is not yet configured
```

ausi (approved):

PR #4155, opened 2022-02-17
Allow `<summary>` and `<details>` tags by default

### LINKED ISSUE #4156: add <details> and <summary> to Allowed HTML tags
### Description

add `<details>` and `<summary>` to Allowed HTML tags  
so in content-element 'markdown' it's possible to add (in basic settings) a collapsible section like

```
<details>
<summary>click to expand</summary>
<p>my awsome information</p>
</details>
```

Toflar:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details


bytehead (dismissed):

ausi:
We should probably add the following line to the allowed attributes too I think:

```php
	array('key' => 'details', 'value' => 'open'),
```

core-bundle/src/Resources/contao/config/default.php
```diff
@@ -60,6 +60,7 @@
 	array('key' => 'th', 'value' => 'rowspan,colspan,width,height'),
 	array('key' => 'style', 'value' => 'media'),
 	array('key' => 'time', 'value' => 'datetime'),
+    array('key' => 'details', 'value' => 'open'),
```
ausi:
```suggestion
	array('key' => 'details', 'value' => 'open'),
```

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #4158, opened 2022-02-17
Set DB server version in install tool

ausi:
To reduce the likelihood of issues with the Doctrine server detection mechanism and to improve performance we should IMO always set the MySQL server version if possible.

Not sure if this needs/should get a backport for Contao 4.9.


leofeyer (approved):

leofeyer:
Thank you @ausi.

aschempp:
Wouldn't this mean the server version is hardcoded and possibly never updated even if the server changes?

ausi:
It is stored in the parameters.yml, the same way as the other values.

aschempp:
yeah but as a user I will not know to update this value. My access details don't randomly change, but the server version does. For any shared hosting I fear this might lead to issues...

ausi:
> My access details don't randomly change, but the server version does.

The MySQL server version of shared hostings change randomly with backwards incompatible versions?

fritzmg:
There have been some instances in the community where this did indeed lead to problems. In most cases it was a switch from MySQL 5.7 to MariaDB 10.x (where the credentials stayed the same), which then leads to the following error:

```
Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'rows rows VARCHAR(8) DEFAULT '2rwh' NOT NULL'
```

for the query

```
ALTER TABLE tl_layout CHANGE rows rows VARCHAR(8) DEFAULT '2rwh' NOT NULL
```

because `rows` needs to be quoted then - which Doctrine won't do because `server_version` was still pinned to MySQL 5.7 by the Install Tool.

ausi:
I think we should do the following: Implement a check in `contao:migrate` that verifies that the serverVersion is correct. This way the user can be informed of the incorrect configuration and the Contao Manager will propt you to reenter the database credentials (in the future) which would fix the serverVersion.

Issue #2868, opened 2021-03-11
Remove the "first", "even", "odd" and "last" CSS classes


asaage:
**Affected version(s)**

I think that's been forever.

**Description**

The generated classes of news_list items are mixed up
![image](https://user-images.githubusercontent.com/4138336/110792181-5b53ca80-8273-11eb-93ba-ed0d56dba044.png)

Not a big deal - i just wodered why i was getting strange results while using css like such:
```css
mod_newslist>.even { }
mod_newslist>.odd { }
/* vs */
mod_newslist>:nth-child(even) { }
mod_newslist>:nth-child(odd) { }
```
One/first is an odd number but i think somebody startet counting from zero 😂
FunFact: ```:nth-child(even/odd)``` behaves like ```.even``` if your mod_newslist has a headline.

fritzmg:
Yes, it uses the zero-based index. This is the case everywhere in Contao, not just the newslist. It cannot be changed for BC reasons, at least not in Contao 4. If anything it should be removed (just like we did for articles).

m-vo:
Fun fact: I tripped over the fact `nth-of-type()` uses a one-based index multiple times. I mean, come on, who starts counting at 1?! :smile:

fritzmg:
Same. I am always unsure when using `nth-…(even/odd)` 😁 (but not because of Contao).

aschempp:
we're using Math© (1 is an odd number), looks like CSS got it wrong 😆

leofeyer:
As briefly discussed in Mumble on March 11th, we want to remove these classes in Contao 5.

leofeyer:
Removed in #4178.

PR #4178, opened 2022-02-19
Remove the "first", "even", "odd" and "last" CSS classes

fritzmg:
Implements #2868

Discussion points:

1. Should we keep things like
    ```php
    <li class="foobar<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?>>
    ```
    and
    ```php
    <li<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?>>
    ```
    in the templates, even if the module or content element itself does not provide any CSS classes any more at all? People are still able to add custom CSS classes via a `parseTemplate` hook for example - but with this template change they will not be output any more.

2. Likewise should we transform things like
    ```php
    <li class="<?= $item['class'] ?>">
    ```
    into
    ```php
    <li<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?>>
    ```
    instead of removing it?

3. Should we also remove the `row_0` etc. classes (used for front end form fields mostly)? Technically we only talked about removing first/last/even/odd.


ausi:
> 1. Should we keep things like … and … in the templates, even if the module or content element itself does not provide any CSS classes any more at all?
> 2. Likewise should we transform things like … into … instead of removing it?

I think we should remove them.

> 3. Should we also remove the `row_0` etc. classes (used for front end form fields mostly)?

Yes, as selecting the `nth-child` is easy in CSS nowadays.

aschempp:
I would be fine with keeping `echo $class` in templates, but certainly remove `row_0` etc., that should also apply to the gallery content element.

fritzmg:
> I would be fine with keeping `echo $class` in templates

Expanding on that, should we then keep empty `class` array items - or use `$item['class'] ?? …` in the template?

fritzmg:
I have also removed `row_*` and `col_*`. I also decided to remove the template variables completely, if the core no longer sets any class.

Toflar (dismissed):

leofeyer:
> Expanding on that, should we then keep empty `class` array items - or use `$item['class'] ?? …` in the template?

I‘d replace every

```php
<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?>
```

with

```php
class="<?= $item['class'] ?? '' ?>"
```

just because it is shorter and improves the readability of the template. But that is a highly biased opinion. 😉

m-vo:
Just my two cents: With new Twig templates coming for all content elements (that do not contain any of these classes of course) we could be a bit on the "easier side" in regards of breaking templates for the old content elements. But all in all I'm fine with any of the proposed changes :relaxed:.

leofeyer:
How will we handle the case above in Twig? Do we have to use `is defined` or `is not empty` or will we just set an empty string to be the default value?

m-vo:
> How will we handle the case above in Twig? Do we have to use is defined or is not empty or will we just set an empty string to be the default value?

For this specific case you can use the new `HtmlAttributes` class (would also work in the legacy templates btw.)<sup>1)</sup>:
```twig
<div{{ attrs().addClass(item.class) }}>
or
<div{{ attrs().setIfExists('class', item.class) }}>
```

<sup>1) the first example would currently produce an empty class attribute, but that is a bug/edgecase that I'm going to fix anyway</sup>

fritzmg:
But what if item.class is not set (which would be the case currently)? Doesn't that result in an error?

m-vo:
> But what if item.class is not set (which would be the case currently)? Doesn't that result in an error?

In case its an empty string it would be ignored (so no error). In case the variable could possible not exist, you would need to check with `if defined` or - preferably - use the default filter: `item.class|default('')`. In PHP templates this would probably be `?: ''` then.

By the way having undefined variables instead of just setting them to null/false/empty if not applicable is something I generally avoid in Twig as it makes it unnecessarily complicated to deal with in templates.

fritzmg:
> By the way having undefined variables instead of just setting them to null/false/empty if not applicable is something I generally avoid in Twig as it makes it unnecessarily complicated to deal with in templates.

Yes, but would you rather set a variable that will be _always_ empty now?

m-vo:
> Yes, but would you rather set a variable that will be always empty now?

After a short discussion on Slack I now get the issue. IMHO our templates are tied to a controller. If the controller does not set a context, it doesn't make any sense to me to try outputting it in a template, so I would remove this altogether. 

But like I said, if we think this imposes too big of a break, we should consider not changing anything at all for the legacy templates and instead motivate people to migrate to the new fragment controllers and Twig templates.

aschempp:
Please also consider a lot of tools might use these to add css classes using an `outputFrontendTemplate` or similar hook. There are multiple extensions to ease management of CSS classes on content elements and other stuff, like https://extensions.contao.org/?p=craffft%2Fcss-style-selector-bundle or https://extensions.contao.org/?p=oveleon%2Fcontao-component-style-manager.

leofeyer:
> In case its an empty string it would be ignored (so no error). In case the variable could possible not exist, you would need to check with `if defined` or - preferably - use the default filter: `item.class|default('')`. In PHP templates this would probably be `?: ''` then.

The question is whether we will use `{% if item.class is defined %}class="{{ item.class }}"{% endif %}` or just `class="{{ item.class|default('') }}"`. Is it OK to have an empty `class=""` attribute or do we want to avoid that if possible?

fritzmg:
> Is it OK to have an empty `class=""` attribute or do we want to avoid that if possible?

`class=""` would be valid - just seems erroneous to me if a CMS outputs that by default.

leofeyer:
I have created a follow-up PR because there were just too many changes to request them here. I can make a pull on your pull once #4524 has been approved, so you don‘t lose the credits for the work you have already done.

leofeyer (changes requested):
I have created a follow-up PR because there were just too many changes to request them here. I can make a pull on your pull once #4524 has been approved, so you don‘t lose the credits for the work you have already done.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

Issue #4159, opened 2022-02-17
MySQL Row size too large


ausi:
### Affected version(s)

4.9+

### Description

I was finally able to reproduce the too large row size error after it was [reported on Slack](https://contao.slack.com/archives/CK9GW3ARL/p1645126965059009) today.

It seems that there are two very similar but different error messages at work here:

1. `ERROR 1118 (42000): Row size too large (> 8126). Changing some columns to TEXT or BLOB may help.`  
   This is caused by the InnoDB storage engine.
  
2. `ERROR 1118 (42000): Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs`  
   This limit is the absolute maximum row size for all tables regardless of the storage engine.

See <https://mariadb.com/kb/en/innodb-row-formats-overview/#maximum-row-size> for more details.

`VARCHAR(255)` does **not** add more to the row size than `VARCHAR(2048)` for InnoDB (max 8126) but it does add to the row size for the Database itself (max 65535).

The following query runs into the general size limit (second error from above) because 16383 * 4 = 65532 and the remaining 3 bytes are not enough for the `int` column:

```sql
CREATE TABLE test (col1 varchar(16383), col2 int) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

On the other hand, InnoDB uses a maximum of 40 data bytes per `VARCHAR()` one byte for the length and one additional bit (makes 41.125 bytes per column). So it can store a maximum of 197 `VARCHAR()` columns (whose size is 10 characters (40 bytes) or more).

<details>
<summary>Reproduction:</summary>

```sql
CREATE TABLE test (
   col1 varchar(10) NOT NULL,
   col2 varchar(10) NOT NULL,
   col3 varchar(10) NOT NULL,
   col4 varchar(10) NOT NULL,
   col5 varchar(10) NOT NULL,
   col6 varchar(10) NOT NULL,
   col7 varchar(10) NOT NULL,
   col8 varchar(10) NOT NULL,
   col9 varchar(10) NOT NULL,
   col10 varchar(10) NOT NULL,
   col11 varchar(10) NOT NULL,
   col12 varchar(10) NOT NULL,
   col13 varchar(10) NOT NULL,
   col14 varchar(10) NOT NULL,
   col15 varchar(10) NOT NULL,
   col16 varchar(10) NOT NULL,
   col17 varchar(10) NOT NULL,
[...]

PR #4179, opened 2022-02-19
Add a warning for too large database row sizes

ausi:
To improve the situation for #4159 I created a small tool that is able to measure and calculate row sizes.

Not sure if this should ever get merged, but I needed a way to write the (somewhat confusing) rules for row sizes down and be able to validate them.

`getMysqlColumnSizeBits()` and `getInnodbColumnSizeBits()` calculates the amout of bits the column contributes to the row size.

`measureMysqlColumnSizeBits()` and `measureInnodbColumnSizeBits()` return the exact same information, but evaluate the amout of bits needed by trail and error against the database. (slow, only used for testing)

Bildschirmfoto 2022-02-19 um 23 46 42


ausi:
As an example: in d4d4858c57643cd1c915683f40202a20505e0138 I reduced the row size of `tl_content` from 34698 down to 2663 (MySQL) and from 2313 down to 1998 (InnoDB).

**Side note:** Since MySQL 8.0.13 and MariaDB 10.2.1, default expressions are supported including for `TEXT` and `BLOB` columns. Are these versions too high as a system requirement for Contao 5.0?

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -194,7 +194,7 @@
 			'options_callback'        => array('tl_content', 'getContentElements'),
 			'reference'               => &$GLOBALS['TL_LANG']['CTE'],
 			'eval'                    => array('helpwizard'=>true, 'chosen'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50'),
-			'sql'                     => array('name'=>'type', 'type'=>'string', 'length'=>64, 'default'=>'text')
+			'sql'                     => array('name'=>'type', 'type'=>'string', 'length'=>64, 'default'=>'text', 'platformOptions'=>array('collation'=>'ascii_bin'))
```
fritzmg:
```suggestion
			'sql'                     => array('name'=>'type', 'type'=>'string', 'length'=>64, 'default'=>'text', 'customSchemaOptions'=>array('collation'=>'ascii_bin'))
```
As discussed in Slack :) https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#vendor-specific-options

fritzmg (changes requested):

ausi:
Briefly discussed with @fritzmg we should probably:

1. Apply all `COLLATE ascii_bin` changes
2. Change `varchar`s that are lager than 255 to `text`
3. Don’t touch all other `varchar`s as the change of the default value from `''` to `null` could cause backwards compatibility issues.

aschempp:
could this _tool_ be useful to solve https://github.com/isotope/core/issues/1802 as well? 🙃

ausi:
> could this _tool_ be useful to solve [isotope/core#1802](https://github.com/isotope/core/issues/1802) as well? 🙃

Don’t think so. The 64 key limit seems to be pretty straightforward. You should probably not create such a huge amount of indexes in the first place I would guess. Maybe some of them could be combined and some removed?

ausi:
> 1. Apply all `COLLATE ascii_bin` changes
> 2. Change `varchar`s that are lager than 255 to `text`
> 3. Don’t touch all other `varchar`s as the change of the default value from `''` to `null` could cause backwards compatibility issues.

Done in 8db72fa4293aca0a2cf13f129710cf604e0ff494

Before this PR:

- `tl_content`: MySQL 34698 (53%) InnoDB 2313 (28%)
- `tl_module`: MySQL 25333 (39%) InnoDB 2726 (34%)

Now:

- `tl_content`: MySQL 14679 (22%) InnoDB 1963 (24%)
- `tl_module`: MySQL 5290 (8%) InnoDB 2427 (30%)

ausi:
The changes from `tl_content` and `tl_module` should be backported to Contao 4.9.
@leofeyer should I make an extra PR after this one was approved?

core-bundle/src/Resources/contao/dca/tl_content.php
```diff
@@ -277,30 +277,30 @@
 			'inputType'               => 'imageSize',
 			'reference'               => &$GLOBALS['TL_LANG']['MSC'],
 			'eval'                    => array('rgxp'=>'natural', 'includeBlankOption'=>true, 'nospace'=>true, 'helpwizard'=>true, 'tl_class'=>'w50'),
-			'sql'                     => "varchar(255) NOT NULL default ''"
+			'sql'                     => "varchar(64) COLLATE ascii_bin NOT NULL default ''"
```
fritzmg: I am not sure about this change. If you have an image size called `this_lorem_ipsum_foobar_image_size` then the resulting serialized array

```
a:3:{i:0;i:0;i:1;i:0;i:2;s:23:"_this_lorem_ipsum_foobar_image_size";}
```

would be 69 characters long. (nice)

ausi:
```suggestion
			'sql'                     => "varchar(128) COLLATE ascii_bin NOT NULL default ''"
```

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -435,21 +435,21 @@
 			'inputType'               => 'imageSize',
 			'reference'               => &$GLOBALS['TL_LANG']['MSC'],
 			'eval'                    => array('rgxp'=>'natural', 'includeBlankOption'=>true, 'nospace'=>true, 'helpwizard'=>true, 'tl_class'=>'w50'),
-			'sql'                     => "varchar(255) NOT NULL default ''"
+			'sql'                     => "varchar(64) COLLATE ascii_bin NOT NULL default ''"
```
ausi:
```suggestion
			'sql'                     => "varchar(128) COLLATE ascii_bin NOT NULL default ''"
```

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
> The changes from `tl_content` and `tl_module` should be backported to Contao 4.9.
> @leofeyer should I make an extra PR after this one was approved?

Yes, please.

leofeyer:
The changes do not work for me without https://github.com/ausi/contao-contao/pull/3. Contao will adjust the columns, however, since the `DcaSchemaProvider` does not handle the collation, the migration is never done.

leofeyer (changes requested):
The changes do not work for me without https://github.com/ausi/contao-contao/pull/3. Contao will adjust the columns, however, since our `DcaSchemaProvider` does not handle collations, the migration is repeated over and over again.

leofeyer (approved):

leofeyer:
Thank you @ausi.

Issue #501, opened 2019-05-31
Page visibility in XML-Sitemap and frontend module should be separate settings


tabcontao:
If I decide not to include a page in "the" XML sitemap used for search engines, it will not appear in the frontend module sitemap either. There should be a possibility to choose the visibility in both sitemaps separately, because XML sitemap and sitemap frontend module are different matters IMHO.

In a project we decided to set the Robots-Tag for the privacy statement page to "noindex, nofollow", but at the same time we chose to include the page in the sitemap, because we want to show it in the sitemap page which uses the sitemap frontend module. Now, Google sends me mails complaining about the page being included in the XML sitemap while being set to noindex in the Robots-Tag. Therefor, I need to exclude the page from XML sitemap but keeping it included in the frontend module.

leofeyer:
> If I decide not to include a page in "the" XML sitemap

How exactly did you do this?

tabcontao:
Actually, I didn't. If I would do it, then I would do it in the properties of the page by setting "In der Sitemap anzeigen" to "Nie anzeigen" or "Standard". But then it wouldn't show up any more in the frontend module as well.

My settings at the moment are:
Im Menü verstecken: Activated/Checked
In der Sitemap anzeigen: Immer anzeigen

The privacy statement page is not to be included in the main navigation ("Navigationsmenü"), only in a modul "Individuelle Navigation" which is shown in the footer. It should also be included in the sitemap frontend module. So I think my settings are pretty much forced. If I set the page to "Standard" or "Nie anzeigen" it is not shown in the frontend module.

leofeyer:
How exactly do I reproduce this?

tabcontao:
I'm not sure, if the "Google Search Console Team" still sees this as a problem. I haven't got another mail about this "problem" since. I also didn't get one during the years before. So my case may be a purely academic problem at the moment.
Still I think, that the sitemap module and the XML-Sitemap are separate things and pages should have separate settings for both. Anyway, here goes ...

1. Login to the backend of the official demo
2. Go to the root page in the site structure and enable the XML-sitemap, save.
3. Select another page, e.g. the text elements page, edit its properties, set noindex, nofollow for robots, hide it in navigation, always show in sitemap and save it. 
4. Go to the system maintenance, purge the page cache and recreate the XML-Sitemap.

Now, the text elements page is not visible anymore in the navigation, but is still shown on the sitemap page and, unfortunately, also contained in the XML-Sitemap while its head section shows 
[...]

PR #4186, opened 2022-02-21
Set regular pages regarding XML sitemap

Total-Reality:
Regarding this issue:
https://github.com/contao/contao/issues/501

@leofeyer wrote
> We want to distinguish between "XML sitemap" and "HTML sitemap" in the future, so we need to adjust the following

Sorry if sth. is wrong, this is my first Contao pull request. This is a first suggestion.

**Explanation**
- Nearly the same settings like the settings of the HTML sitemap, but this setting is only for regular pages
- map_default = no changes in sitemap.xml
- map_always = it always show this page in the sitemap.xml - independent from the robot settings `noindex,nofollow`
- map_never = page won't be added to sitemap.xml

**Current problems of this PR**
- We need to change the tl_class of the backend fields below or the position in the palette has to be changed - I think it would be nice to show both sidemap settings side by side, however it is a problem that the XML sitemap field is only visible in the regular page type
- The new Backend labels are only available in _en_ and _de_ currently

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -141,7 +141,7 @@ class SitemapController extends AbstractController
             if (
                 $isPublished
                 && !$pageModel->requireItem
-                && 'noindex,nofollow' !== $pageModel->robots
+                && ('noindex,nofollow' !== $pageModel->robots && 'map_never' !== $pageModel->xml || 'map_always' === $pageModel->xml)
                 && $this->pageRegistry->supportsContentComposition($pageModel)
                 && $this->pageRegistry->isRoutable($pageModel)
                 && 'html' === $this->pageRegistry->getRoute($pageModel)->getDefault('_format')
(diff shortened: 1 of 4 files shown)


ausi:
> we want to keep URLs to `noindex,follow` pages in the XML sitemap.

With this pull request, `noindex,follow` pages would get removed from the XML sitemap in some cases, so it is against what we have discussed.

Total-Reality:
> > we want to keep URLs to `noindex,follow` pages in the XML sitemap.
> 
> With this pull request, `noindex,follow` pages would get removed from the XML sitemap in some cases, so it is against what we have discussed.

How do you mean this? A page with `noindex,follow` will be removed from the XML sitemap only if you set the **new** XML sitemap setting to `map_never`. So I don't understand the problem?

asaage:
With this i think you would be able to have a site ```index,follow```+```map_never``` and it would be in sitemap.xml.
That's ridiculous isn't it?
If the goal is to have no warnings or errors in google - google simply should get a different sitemap without ambiguous options in the page.

Total-Reality:
> With this i think you would be able to have a site `index,follow`+`map_never` and it would be in sitemap.xml. That's ridiculous isn't it? If the goal is to have no warnings or errors in google - google simply should get a different sitemap without ambiguous options in the page.

That is not true. If you set `index,follow`+`map_never` -> this site **won't** be in the XML sitemap. I tested this already...

asaage:
yep your probably right.
I allways struggle to resolve ```( X && Y || Z )``` the right way without plenty of ```(``` and ```)``` 🤦‍♂️

leofeyer:
You can use the [contao.sitemap](https://docs.contao.org/dev/reference/events/#contao-sitemap) event to modify the XML according to your wishes, so there is no need to add another field to the core. We would only do this if there was no other way to accomplish it or if the majority of users would need it. Both is not the case IMHO.

PR #4190, opened 2022-02-22
Remove left-over url field in tl_module

bytehead:
This is a left-over of the flash player integration from ealier days 😬


fritzmg (approved):

m-vo:
Isn't this a bug then, that should also land in 4.9/4.13?

fritzmg:
> Isn't this a bug then, that should also land in 4.9/4.13?

I would not remove this field in Contao 4. People might use it and then suddenly "lose" it when they update to a newer bugfix version.

bytehead:
IMO it's a bug, but also a possibly break.

m-vo:
> IMO it's a bug, but also a possibly break.

:+1: Because this is another database field (which unfortunately is a limited resource), I'd vote for bug though.

fritzmg:
> 👍 Because this is another database field (which unfortunately is a limited resource), I'd vote for bug though.

Yes, but with https://github.com/contao/contao/pull/4179#issuecomment-1047658782 this field will be changed to `text` anyway (in 4.9+) and thus it will be less of a problem.

ausi:
I’d also vote for keeping the field in Contao 4, the likelihood that some extension or application reuses this field is very high I think.

m-vo (approved):

ausi (approved):

aschempp (approved):

Toflar (approved):

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #4192, opened 2022-02-22
Shorten the version description to the length of the DB column

### LINKED ISSUE #4188: Creating a new version may result in an exception
### Affected version(s)

4.9.26

### Description

The automatic creation of a new version may result in an exception. This is the case, if the DCA has an attribute `title`, `name`, `firstname`, `headline`, `selector` or `subject` which will be used as description for the version and this field allows more than 255 characters.

This is the part, where the attribute is read: https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/classes/Versions.php#L187-L223

The description field only allows 255 chars: https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/dca/tl_version.php#L60

I would suggest to shorten the description value before inserting it into the database.

I can create a PR if my suggestion is okay.

cliffparnitzky:
closes #4188

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -216,6 +216,7 @@ class Versions extends Controller
 		{
 			$strDescription = $data['subject'];
 		}
+
 		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
 
 		$intId = $this->Database->prepare("INSERT INTO tl_version (pid, tstamp, version, fromTable, username, userid, description, editUrl, active, data) VALUES (?, ?, IFNULL((SELECT MAX(version) FROM (SELECT version FROM tl_version WHERE pid=? AND fromTable=?) v), 0) + 1, ?, ?, ?, ?, ?, 1, ?)")
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -216,7 +216,7 @@ class Versions extends Controller
 		{
 			$strDescription = $data['subject'];
 		}
-		$strDescription = StringUtil::substr($strDescription, 260);
+		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
 
 		$intId = $this->Database->prepare("INSERT INTO tl_version (pid, tstamp, version, fromTable, username, userid, description, editUrl, active, data) VALUES (?, ?, IFNULL((SELECT MAX(version) FROM (SELECT version FROM tl_version WHERE pid=? AND fromTable=?) v), 0) + 1, ?, ?, ?, ?, ?, 1, ?)")
 								->execute($this->intPid, time(), $this->intPid, $this->strTable, $this->strTable, $blnHideUser ? null : $this->getUsername(), $blnHideUser ? 0 : $this->getUserId(), $strDescription, $this->getEditUrl(), serialize($data))
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -217,7 +217,7 @@ class Versions extends Controller
 			$strDescription = $data['subject'];
 		}
 
-		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
+		$strDescription = mb_substr($strDescription, 0, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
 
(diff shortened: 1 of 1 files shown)


aschempp (commented):
why 260 characters if the DB field is 255? Also, can we somehow dynamically determine the length if someone did increase it through a DCA change? maybe @ausi has an idea?

ausi:
Something like this?  
`System::getContainer()->get('database_connection')->createSchemaManager()->listTableColumns('tl_version')['description']->getLength() ?? 255`

cliffparnitzky:
> why 260 characters if the DB field is 255?
I've tested all possible values around this "255" ... and what should I say: "260" was the maximum possible.

cliffparnitzky:
I will try the suggestion of @ausi and give feedback here.

ausi:
> I've tested all possible values around this "255" ... and what should I say: "260" was the maximum possible.

This is probably because `StringUtil::substr()` does not separate words.

cliffparnitzky:
Just for documentation. It has to be:
`System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength()`.

Fixed it. Works fine.

core-bundle/src/Resources/contao/classes/Versions.php
```diff
@@ -216,6 +216,7 @@ public function create($blnHideUser=false)
 		{
 			$strDescription = $data['subject'];
 		}
+		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
```
leofeyer: Why not read the length from the DCA? Would save us the DB query.

cliffparnitzky: :-/ any suggestion?

cliffparnitzky: @leofeyer How should I read the length from the DCA, without String parsing the `sql` field? Or ist that the idea, you had? We cannot rely on the `maxlength` value of the `eval` array. I have no clue how to get the length value directly than via the code @ausi proposed.

Toflar (dismissed):

core-bundle/src/Resources/contao/classes/Versions.php
cliffparnitzky: @leofeyer Any proposal to get the conversation resolved?

aschempp (dismissed):

core-bundle/src/Resources/contao/classes/Versions.php
leofeyer: > String parsing the `sql` field? Or ist that the idea, you had?

Yes.

Toflar: We have a schema listener that does this and extracts the information to add it to the schema. It's okay to access it over the schema even if that means queries.

leofeyer:
```suggestion

		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
```

leofeyer (changes requested):
I couldn‘t fix this myself because you did not allow edits by maintainers.

core-bundle/src/Resources/contao/classes/Versions.php
```diff
@@ -217,6 +217,8 @@ public function create($blnHideUser=false)
 			$strDescription = $data['subject'];
 		}
 
+		$strDescription = StringUtil::substr($strDescription, System::getContainer()->get('database_connection')->getSchemaManager()->listTableColumns('tl_version')['description']->getLength());
```
m-vo: Judging from the doc block `StringUtil::substr()` strips tags and may return something slightly longer than the given `$intNumberOfChars`. Is this the intended outcome here?

cliffparnitzky: > The function preserves words, so the result might be a bit shorter or longer than the number of characters given. It strips all tags.

I think, the length must not be longer than the detected length of the DB field.

@m-vo Any suggestion to get the ticket finally done?

ausi: > and may return something slightly longer than the given `$intNumberOfChars`.

If this is the case we should fix `StringUtil::substr()` IMO to never return something longer than `$intNumberOfChars`.

cliffparnitzky: Thank you @ausi . That was my thought too.

leofeyer: If you don‘t need the special functionality of the `StringUtil::substr()` method, please just use `mb_substr()` directly.

cliffparnitzky: Seems the version title is currently not displayed anywhere. So, the special functionality of the `StringUtil::substr()` method is not needed. I changed to code to `mb_substr()`.

leofeyer (approved):

leofeyer:
Thank you @cliffparnitzky.

PR #4195, opened 2022-02-22
Set a session for the fake request

### LINKED ISSUE #4185: There is currently no session available.
### Affected version(s)

4.13.0, 5.x

### Description

After creating and saving a news element, I can no longer access the "News" section in the backend. The following error occurs:

```
Symfony\Component\HttpFoundation\Exception\SessionNotFoundException:
There is currently no session available.

  at vendor/symfony/http-foundation/RequestStack.php:126
  at Symfony\Component\HttpFoundation\RequestStack->getSession()
     (vendor/symfony/security-core/Authentication/Token/Storage/UsageTrackingTokenStorage.php:89)
  at Symfony\Component\Security\Core\Authentication\Token\Storage\UsageTrackingTokenStorage->getSession()
     (vendor/symfony/security-core/Authentication/Token/Storage/UsageTrackingTokenStorage.php:44)
  at Symfony\Component\Security\Core\Authentication\Token\Storage\UsageTrackingTokenStorage->getToken()
     (vendor/symfony/security-core/Authorization/AuthorizationChecker.php:70)
  at Symfony\Component\Security\Core\Authorization\AuthorizationChecker->isGranted('ROLE_USER', null)
     (vendor/symfony/security-core/Security.php:63)
  at Symfony\Component\Security\Core\Security->isGranted('ROLE_USER')
     (vendor/contao/core-bundle/src/EventListener/PrettyErrorScreenListener.php:87)
  at Contao\CoreBundle\EventListener\PrettyErrorScreenListener->handleException(object(ExceptionEvent))
     (vendor/contao/core-bundle/src/EventListener/PrettyErrorScreenListener.php:79)
  at Contao\CoreBundle\EventListener\PrettyErrorScreenListener->__invoke(object(ExceptionEvent), 'kernel.exception', object(TraceableEventDispatcher))
     (vendor/symfony/event-dispatcher/Debug/WrappedListener.php:117)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(object(ExceptionEvent), 'kernel.exception', object(TraceableEventDispatcher))
     (vendor/symfony/event-dispatcher/EventDispatcher.php:230)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners(array(object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener)), 'kernel.exception', object(ExceptionEvent))
     (vendor/symfony/event-dispatcher/EventDispatcher.php:59)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch(object(ExceptionEvent), 'kernel.exception')
     (vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:154)
  at Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch(object(ExceptionEvent), 'kernel.exception')
     (vendor/symfony/http-kernel/HttpKernel.php:213)
  at Symfony\Component\HttpKernel\HttpKernel->handleThrowable(object(SessionNotFoundException), object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:106)
  at Symfony\Component\HttpKernel\HttpKernel->terminateWithException(object(SessionNotFoundException), object(Request))
     (vendor/symfony/http-kernel/EventListener/DebugHandlersListener.php:131)
  at Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::Symfony\Component\HttpKernel\EventListener\{closure}(object(SessionNotFoundException))
     (vendor/symfony/error-handler/ErrorHandler.php:607)
  at Symfony\Component\ErrorHandler\ErrorHandler->handleException(object(SessionNotFoundException))
```

All other sections work fine. Then if I create and save a new event in a calendar, the same thing happens. I will no longer be able to access the "Events" section in the back end (i.e. I can no longer acess "News" and "Events" at that point). This happens under PHP 7.4 as well as 8.1.

The error is unfortunately not reproducible in the Contao Online Demo.

fritzmg:
Fixes #4185

Having `BackendUser::getInstance()->id` in the DCA is fine. However, in Symfony 5.4+, the `UsageTrackingTokenStorage` always retrieves the session from the request, if available. Since there is a main request available in the back end, this will fail, because in this case the last request in the request stack is one of the faux requests for generating the RSS feed items (see #3857).

May be this is more of an oversight on Symfony's part though. `UsageTrackingTokenStorage::shouldTrackUsage` assumes that if a main request is available, there also must be a session available - I am not sure if that is a fair assumption.

This error does not happen in Contao 4.9, because first of all `UsageTrackingTokenStorage::$enableUsageTracking` will be `false` there and second of all Symfony 4.4 still uses the `session` service, instead of retrieving the session from the request.


aschempp (dismissed):
I don't think setting a fake session can be the right fix. The actual issue is probably related to that the `BackendUser` class is trying to access the session to retrieve the current firewall. 

Also, I'm confused by your initial description that combines _session is retrieved from main request_ but _our fake request does not have one_. Those are separate request objects?

fritzmg:
> I don't think setting a fake session can be the right fix.

I am not sure either. The alternative would be to copy the session of the current request, if available.

> The actual issue is probably related to that the `BackendUser` class is trying to access the session to retrieve the current firewall.

imho that should not be an issue. See below

> Also, I'm confused by your initial description that combines _session is retrieved from main request_ but _our fake request does not have one_. Those are separate request objects?

The issue is as follows:

1. `BackendUser::getInstance()` retrieves an authentication token from `security.token_storage`.
    https://github.com/contao/contao/blob/85505c87670ac7488ceb10bc349548e7fffcae60/core-bundle/src/Resources/contao/classes/BackendUser.php#L137
2. `security.token_storage` will be an instance of `UsageTrackingTokenStorage`.
3. `UsageTrackingTokenStorage::getToken` will [check whether it should track session usage](https://github.com/symfony/symfony/blob/49b6d7f308ddc85833cd626302ff3238dbabe800/src/Symfony/Component/Security/Core/Authentication/Token/Storage/UsageTrackingTokenStorage.php#L42). <sup>1</sup>
4. [`UsageTrackingTokenStorage::shouldTrackUsage`](https://github.com/symfony/symfony/blob/49b6d7f308ddc85833cd626302ff3238dbabe800/src/Symfony/Component/Security/Core/Authentication/Token/Storage/UsageTrackingTokenStorage.php#L92-L108) will return true because:
    1. `$this->enableUsageTracking` will not be false.
    2. `$this->container->get('request_stack')->getMainRequest()` will return a request.
5. Since `shouldTrackUsage` will be true, `getToken` will [retrieve the session](https://github.com/symfony/symfony/blob/49b6d7f308ddc85833cd626302ff3238dbabe800/src/Symfony/Component/Security/Core/Authentication/Token/Storage/UsageTrackingTokenStorage.php#L44).
6. `UsageTrackingTokenStorage::getSession` will [retrieve the session from the request stack](https://github.com/symfony/symfony/blob/49b6d7f308ddc85833cd626302ff3238dbabe800/src/Symfony/Component/Security/Core/Authentication/Token/Storage/UsageTrackingTokenStorage.php#L89).
7. `RequestStack::getSession` will [return the session from the last request in the request stack](https://github.com/symfony/symfony/blob/49b6d7f308ddc85833cd626302ff3238dbabe800/src/Symfony/Component/HttpFoundation/RequestStack.php#L122-L124). <sup>2</sup>

And the last request in the request stack will be the faux request, containing no session.

<sup>1</sup> In Contao 4.9 / Symfony 4.4, session tracking will not be enabled.
<sup>2</sup> In Contao 4.9  / Symfony 4.4, the session will be retrieved from the `session` service, not from the last request in the request stack. So even if session tracking was enabled, this issue would not occur there.

aschempp:
So maybe we should clone the current request instead of creating a fake one for the fragments? It appears that's how Symfony does subrequests, otherwise they would suffer the same problem.

fritzmg:
May be yes. But in my eyes it's also kind of wrong to use the current back end session for these faux requests for generating RSS feed items 🤷.

fritzmg:
After all, these requests get the __frontend__ scope. Therefore I still think it needs to be an empty session without any storage.

aschempp:
What if I generate the feeds on the command line? It wouldn't get a request at all, wouldn't it?

fritzmg:
> What if I generate the feeds on the command line? It wouldn't get a request at all, wouldn't it?

On the command line `UsageTrackingTokenStorage::shouldTrackUsage` will return false, because `enableUsageTracking` will be false.

fritzmg:
Alternatively, we could also disable session usage tracking on the fly during RSS feed generation via 

```php
$tokenChecker = System::getContainer()->get('contao.security.token_checker');

if ($tokenChecker instanceof UsageTrackingTokenStorage) {
    $tokenChecker->disableUsageTracking();
}
```

However, there is no way of knowing whether it was enabled or disabled before.

aschempp:
> But in my eyes it's also kind of wrong to use the current back end session for these faux requests for generating RSS feed items 🤷.

It's not a backend request. It's a clone of the current/main request with adjusted properties. The only thing making it a backend request is the `_scope` attribute, which we change. There is no "distinction" of request "types" in Symfony. We should behave exactly like any subrequest in the kernel does.

fritzmg:
> It's not a backend request.

But it is. Regardless of the `frontend` `_scope` that we set after the fact, `$security->getUser()` will be an instance of `BackendUser`, if we copy the back end session over.

aschempp:
> But it is. Regardless of the `frontend` `_scope` that we set after the fact, `$security->getUser()` will be an instance of `BackendUser`, if we copy the back end session over.

Yes, because the firewall is tied to your main request. The main request has backend scope. You probably mean that the current request is a subrequest of a backend request, and there's really nothing we can do to change that except using a real controller imho.

fritzmg:
You are right, I assumed the firewall would use the session of the current request, but that is not the case. So we might as well use the same session for the faux request. Implemented in 8bc37ed.

ausi (dismissed):

fritzmg:
As discussed in the Contao call, we want to integrate the changes into `4.9` as well and also check whether we should create the new request just like in `InlineFragmentRenderer::createSubRequest`.

aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #4197, opened 2022-02-23
Fix some dynamic routes handling

### LINKED ISSUE #4191: Custom backend route prefix not used in robots.txt
### Affected version(s)

4.13

### Description

If you set your contao to have a custom backend route prefix, as described here https://docs.contao.org/manual/en/system/settings/#additional-back-end-settings

The disallow entry in the robots.txt is not adjusted correctly to the new route. As you can see here `contao` is hardcoded in there
https://github.com/contao/contao/blob/c904c5624e7f813980942cbf9dc5386405c0ee27/core-bundle/src/EventListener/RobotsTxtListener.php#L59

aschempp:
Fixes https://github.com/contao/contao/issues/4191

Also found a few other issues by chance. Instead of injecting the router into the `ContaoFramework`, we could also throw an appropriate exception and handle it in the `PrettyErrorScreenHandler`. But that would disable the redirect `dev` mode, but we could also render a template to explain instead of a plain redirect.


leofeyer (dismissed):

leofeyer:
You also have to adjust the tests, though. 🙈

core-bundle/src/Resources/config/services.yml
```diff
@@ -1119,7 +1120,12 @@ services:
             version: 4.13
             message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "contao.csrf.token_manager" instead.
 
-    Contao\CoreBundle\Framework\ContaoFrameworkInterface: '@contao.framework'
+    Contao\CoreBundle\Framework\ContaoFrameworkInterface:
+        alias: contao.framework
+        deprecated:
+            package: contao/core-bundle
+            version: 4.7
+            message: Using the "%alias_id%" service ID has been deprecated and will no longer work in Contao 5.0. Please use "contao.framework" instead.
```
fritzmg: This will need to be backported to Contao 4.9 as well.

aschempp: see https://github.com/contao/contao/pull/4199

aschempp:
no idea why tests fail, maybe @m-vo can tell?

m-vo:
> no idea why tests fail, maybe @m-vo can tell?

The error happens when the code is accessing `imageSize` via the legacy `File` class. If you add the following to the test, the error disappears:

```php
// FigureRendererTest#testRendersContaoTemplate()`
// …
        $imageFactory = $this->createMock(ImageFactoryInterface::class);
        $imageFactory
            ->method('create')
            ->willReturn($this->createMock(ImageInterface::class))
        ;

        $container->set('contao.image.factory', $imageFactory);
```

This side effect is likely unrelated to this PR but only appears with it.

m-vo:
https://github.com/aschempp/contao/pull/8

core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -358,7 +361,7 @@ private function validateInstallation(): void
         }
 
         if (!$this->getAdapter(Config::class)->isComplete()) {
-            throw new RedirectResponseException('/contao/install');
+            throw new RedirectResponseException($this->urlGenerator->generate('contao_install', [], UrlGeneratorInterface::ABSOLUTE_URL));
```
leofeyer, line 364: Why not inject the route prefix like in the `RobotsTxtListener` and redirect to `$this->routePrefix.'/install'`?

Toflar: What, why? The router is responsible to generate URLs?

leofeyer: We did not use the router before, either.

fritzmg: > We did not use the router before, either.

Which is is why this issue occurs in the first place.

leofeyer (approved):

PR #4203, opened 2022-02-23
Add a HtmlAttributes class and Twig function

m-vo:
This PR introduces a new `HtmlAttributes` class that eases building and outputting HTML attributes in a structured way. Its primary use case is within Twig templates, but it could of course be used everywhere else as well.

```php
// object that is stringable, iterable and implements array access 
$attributes = new HtmlAttributes(['class' => 'foo', 'baz' => 42]);
$attributes = HtmlAttributes::fromString('class=>"foo" baz=42');

$attributes
  ->addClass('bar')
  ->set('style', 'color: red;')
  ->unset('baz')
;

echo "<div$attributes>🥗</div>"; // outputs '<div class="foo bar" style="color: red;">🥗</div>'
```

In Twig there is an additional `attrs()` function that allows instantiating an `HtmlAttributes` object on the fly, optionally passing in an array of properties. Because the class is registered as a "safe class" in HTML, we can use chaining as well and let the escaper convert the value to string for us without the need to add `|raw`.

```twig
{# Outputs <div>🍦</div> #}
<div{{ attrs() }}>🍦</div>

{# Outputs '<div data-foo="bar">🥕</div>' #}
<div{{ attrs({'data-foo': 'bar'}) }}>🥕</div>

{# Outputs '<div class="foo bar">🌮</div>' because we left the style variable empty #}
<div{{ attrs(attributes).addClass('bar').setIfExists('style', style) }}>🌮</div>
```

Note, that attribute names are constrained when being set (must start with a letter, no double hyphen/underscore, no hyphen/underscore at the end) and are only escaped when being output as a string. By default a leading space is returned when converting to a string if at least one attribute exists, otherwise this would get cumbersome to use at places where you cannot be sure if attributes will be present or not.


m-vo:
Should we also allow to pass in a string in the `html_attributes()` Twig function that then calls `HtmlAttributes::fromString()`?

Edit: done in [4c778a6](https://github.com/contao/contao/pull/4203/commits/4c778a6e225a11bfb657091afb918d7efec14bd5)

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        return new self($attributes);
+    }
+
+    public function set(string $name, string|int|bool|\Stringable|null $value): self
+    {
+        $name = strtolower($name);
+
+        if (1 !== preg_match('/^[a-z](?:-?[a-z0-9])*$/', $name)) {
```
ausi:
```suggestion
        if (1 !== preg_match('/^[a-z](?:[_-]?[a-z0-9])*$/', $name)) {
```
It seems that unserscores are used for custom attributes too in the wild 🙈, e.g.: https://github.com/contao/contao/pull/4204

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,248 @@
...
+{
+    /**
+     * @var array<string, string|null>
+     */
+    private array $attributes = [];
+
+    /**
+     * @param array<string, string|int|bool|\Stringable|null> $attributes
+     */
+    public function __construct(array $attributes = [])
```
ausi:
```suggestion
    public function __construct(iterable $attributes = [])
```

WDYT?

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,248 @@
...
+    {
+        $attributes = [];
+
+        foreach ($this->attributes as $name => $value) {
+            $attributes[] = null !== $value ? sprintf('%s="%s"', $name, $this->escapeValue($name, $value)) : $name;
+        }
+
+        $string = implode(' ', $attributes);
+
+        return $leadingSpace && !empty($string) ? " $string" : $string;
```
ausi:
```suggestion
        return $leadingSpace && $string ? " $string" : $string;
```
Nitpick ☺️

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -117,6 +121,10 @@ function (Environment $env, $context, $template, $variables = [], $withContext =
                 },
                 ['needs_environment' => true, 'needs_context' => true, 'is_safe' => ['all']]
             ),
+            new TwigFunction(
+                'html_attributes',
+                static fn (array $attributes = []): HtmlAttributes => new HtmlAttributes($attributes),
```
ausi:
```suggestion
                static fn (iterable $attributes = []): HtmlAttributes => new HtmlAttributes($attributes),
```

core-bundle/src/String/HtmlAttributes.php
m-vo: adjusted in 4bf6a9c8750ca54ce1f3821935a9046364f57650

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,248 @@
...
+
+    public function set(string $name, string|int|bool|\Stringable|null $value): self
+    {
+        $name = strtolower($name);
+
+        if (1 !== preg_match('/^[a-z](?:-?[a-z0-9])*$/', $name)) {
+            throw new \InvalidArgumentException(sprintf('A HTML attribute name must only consist of the characters [a-z0-9-], must start with a letter, must not end with a hyphen and must not contain two hyphens in a row, got "%s".', $name));
+        }
+
+        $this->attributes[$name] = null !== $value ? (string) $value : null;
```
ausi:
```suggestion
        if (false === $value) {
            unset($this->attributes[$name]);
        } else {
            $this->attributes[$name] = true === $value ? '' : (string) $value;
        }
```

As in HTML, boolean attributes are set by existing or not existing, we should unset `false` values IMO.

This way something like this should work as intended:

```twig
<video{{ html_attributes({ 'src': 'video.mp4', 'loop': true, 'muted': false, 'controls': true}) }}>
```

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,248 @@
...
+     * Outputs the attributes as a string that is safe to be placed inside HTML
+     * tags. The output will contain a leading space if $leadingSpace is set to
+     * true and there is at least one property set: e.g. ' foo="bar" bar="42"'.
+     */
+    public function toString(bool $leadingSpace = true): string
+    {
+        $attributes = [];
+
+        foreach ($this->attributes as $name => $value) {
+            $attributes[] = null !== $value ? sprintf('%s="%s"', $name, $this->escapeValue($name, $value)) : $name;
```
ausi:
```suggestion
            $attributes[] = '' !== $value ? sprintf('%s="%s"', $name, $this->escapeValue($name, $value)) : $name;
```

ausi (commented):
❤️

core-bundle/src/String/HtmlAttributes.php
m-vo: see 545992fbdb5b7cc859fa9181e829ae37ecc3517f

core-bundle/src/String/HtmlAttributes.php
m-vo: see 545992fbdb5b7cc859fa9181e829ae37ecc3517f

core-bundle/src/String/HtmlAttributes.php
m-vo: see e86a8130490c9181c3de0023d39c0e75fb74b8b8

core-bundle/src/Twig/Extension/ContaoExtension.php
m-vo: see e86a8130490c9181c3de0023d39c0e75fb74b8b8

m-vo:
@ausi Thanks for the remarks and the enduring support on slack :heart:. 
Feel free to swap out the parser to a regex based on :relaxed:.

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,260 @@
...
+    public function __construct(iterable $attributes = [])
+    {
+        foreach ($attributes as $name => $value) {
+            $this->set($name, $value);
+        }
+    }
+
+    /**
+     * Outputs the attributes as a string that is safe to be placed inside HTML
+     * tags. The output will contain a leading space i there is at least one
```
m-vo:
```suggestion
     * tags. The output will contain a leading space if there is at least one
```

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -58,6 +59,9 @@ public function __construct(Environment $environment, TemplateHierarchyInterface
 
         // Use our escaper on all templates in the `@Contao` and `@Contao_*` namespaces
         $this->addContaoEscaperRule('%^@Contao(_[a-zA-Z0-9_-]*)?/%');
+
+        // Mark HtmlAttributes class as safe as it escapes it output itself
```
m-vo:
```suggestion
        // Mark HtmlAttributes class as safe for HTML as it escapes its output itself
```

m-vo (commented):
:heavy_check_mark: to https://github.com/contao/contao/pull/4203/commits/e62f6a20b903989d0fc1d8a2babb12bc1f8a21e9, https://github.com/contao/contao/pull/4203/commits/23a42c568706b0b6a9c46326b1e05b90acdb74d2, https://github.com/contao/contao/pull/4203/commits/1a50420da3904d920b18b9ea286ac58054b37bc9 by @ausi

m-vo (commented):
:heavy_check_mark: to https://github.com/contao/contao/pull/4203/commits/cce269dea477b36c862a07827c8ff53ee3a8adb5 as well

ausi (dismissed):

ausi (dismissed):

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -132,10 +129,7 @@ public function addClass(string ...$classes): self
             ' ',
             array_unique(
                 array_filter(
-                    array_merge(
-                        explode(' ', $this->attributes['class'] ?? ''),
-                        array_map('trim', $classes)
-                    )
+                    explode(' ', ($this->attributes['class'] ?? '').' '.implode(' ', $classes)),
```
ausi:
```suggestion
                    preg_split('/[ \t\n\r\f]+/', ($this->attributes['class'] ?? '').' '.implode(' ', $classes)),
```

This would be all “supported” whitespace if we want to support this? 
https://drafts.csswg.org/selectors/#whitespace

_(all `explode()`s would have to be changed then I think)_

m-vo: If we're already here. Why not go the full way. :relaxed: 
See f8e25174e14819d03f017a0a1920a21b8a52ff02.

ausi (dismissed):

ausi (dismissed):

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -0,0 +1,250 @@
...
+    {
+        unset($this->attributes[$offset]);
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function split(string $value): array
+    {
+        return array_filter(preg_split('/[\s]+/', $value));
```
ausi:
```suggestion
        return array_filter(preg_split('/\s+/', $value));
```

m-vo: Of course :+1: :man_facepalming:

ausi (dismissed):

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4207, opened 2022-02-23
Fix undefined array key warning when creating new layout sections

fritzmg:
If you add a new custom layout section i a page layout under PHP 8 and then save, the following warning will occur:

```
ErrorException:
Warning: Undefined array key "foobar"

  at vendor\contao\core-bundle\src\Resources\contao\classes\Backend.php:1120
  at Contao\Backend::convertLayoutSectionIdsToAssociativeArray(array('main' => 'Main column', 'header' => 'Header', 'footer' => 'Footer', 'menu' => 'Menu', 'foobar' => 'foobar'))
     (vendor\contao\core-bundle\src\Resources\contao\widgets\ModuleWizard.php:135)
  at Contao\ModuleWizard->generate()
```

This is because the `$GLOBALS['TL_LANG']['COLS']` reference for the custom layout section is added via an `onload` callback. However, the `onload` callback is executed _before_ the record is saved, while the `ModuleWizard` widget is executed _after_ the record is saved. Thus the `ModuleWizard` already tries to render the new custom layout section - but the label for the layout section will be missing in `$GLOBALS['TL_LANG']['COLS']`.

This PR fixes that by ensuring in `ModuleWizard` that the label is always present in `$GLOBALS['TL_LANG']['COLS']`.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/ModuleWizard.php
+++ b/core-bundle/src/Resources/contao/widgets/ModuleWizard.php
@@ -124,6 +124,9 @@ class ModuleWizard extends Widget
 					{
 						$cols[] = $v['id'];
 						$positions[$v['id']] = $v['position'];
+
+						// Add custom layout section reference
+						$GLOBALS['TL_LANG']['COLS'][$v['id']] = $v['title'];
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


aschempp:
so `$GLOBALS['TL_LANG']['COLS']` is no added twice? also, that `$GLOBAL` looks suuper strange to me 😆

leofeyer:
See #4274 for what I think is the correct fix.

leofeyer:
See #4274 for what I think is the correct fix.

PR #4208, opened 2022-02-23
Lower max file size in Dbafs service

### LINKED ISSUE #4200: Dbafs.php, int $maxFile causes Fatal Error in php7.4 32bit
### Affected version(s)

4.13.0

### Description

Local System with  MampPro4 (php7.4, 32bit)
Path/contao/install causes a Fatal Error:

Compile Error: Default value for property of type int can only be int

Symfony\Component\ErrorHandler\Error\ FatalError
in PFAD\vendor\contao\core-bundle\src\Filesystem\Dbafs\Dbafs.php (line 54)

private EventDispatcherInterface $eventDispatcher;
private string $table;
private string $dbPathPrefix = '';
**private int $maxFileSize = 2147483648; // 2 GiB**
private int $bulkInsertSize = 100;
private bool $useLastModified = true;
/**
* @var array<string, array|null>

Solution:

I changed

private int $maxFileSize = 2147483648; // 2 GiB

to

private int $maxFileSize = 2147483647; // 2 GiB

Recording to the [PHP documentation](https://www.php.net/manual/de/language.types.integer.php)  this is the highest possible int number in the 32bit version.

m-vo:
Fixes #4200 

This PR reduces the DBAFS max file size by one so that it also fits in a integer on 32-bit PHP versions. The value is opinionated anyways, so the effect is negligible. Without this change an error would occur because property initializers do no coerce or clamp values.

### DIFF
--- a/core-bundle/src/Filesystem/Dbafs/Dbafs.php
+++ b/core-bundle/src/Filesystem/Dbafs/Dbafs.php
@@ -51,7 +51,7 @@ class Dbafs implements DbafsInterface, ResetInterface
 
     private string $table;
     private string $dbPathPrefix = '';
-    private int $maxFileSize = 2147483647; // 2 GiB - 1 byte (see #?)
+    private int $maxFileSize = 2147483647; // 2 GiB - 1 byte (see #4208)
     private int $bulkInsertSize = 100;
     private bool $useLastModified = true;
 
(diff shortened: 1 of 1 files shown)


bytehead (approved):

fritzmg (approved):

Toflar (approved):

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #4210, opened 2022-02-24
Fix warning in ModuleArticle if page object is null

### LINKED ISSUE #4205: ModuleArticle: Attempt to read property "datimFormat" on null
### Affected version(s)

4.13.0

### Description

If there is a content element of type `article` referencing another article, there will be an error in dev-mode when showing the list view of the content elements.

The problem occurs because contao tries to render / compile the content element.
But `vendor/contao/core-bundle/src/Resources/contao/modules/ModuleArticle.php` tries to access some properties of the global `$objPage`. But the page is null (backend view).

As a workaround the `compile` function could check if `$objPage` is given:

```php
$this->Template->date = $objPage ? Date::parse($objPage->datimFormat, $this->tstamp) : (new \DateTimeImmutable())->format('d.m.Y H:i');
```

image

bezin:
Fixes #4205 

$objPage is null in back-end requests. In this case I fall back to the default datim format.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleArticle.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleArticle.php
@@ -121,7 +121,7 @@ class ModuleArticle extends Module
 
 		// Add the modification date
 		$this->Template->timestamp = $this->tstamp;
-		$this->Template->date = Date::parse($objPage->datimFormat, $this->tstamp);
+		$this->Template->date = Date::parse($objPage->datimFormat ?? Config::get('datimFormat'), $this->tstamp);
 
 		// Clean the RTE output
 		$this->teaser = StringUtil::toHtml5($this->teaser ?? '');


bytehead (dismissed):

fritzmg (dismissed):

fritzmg:
Shouldn't we fix this in 4.9 as well though?

bytehead (approved):

aschempp:
shouldn't we check for `$objPage` instead of accessing a property of a non-existing object?

fritzmg:
> shouldn't we check for `$objPage` instead of accessing a property of a non-existing object?

That's not necessary due to the null coalescing assignment operator.

fritzmg (approved):

aschempp (approved):

leofeyer:
Thank you @bezin.

PR #4212, opened 2022-02-24
Unset key in URL when switching edit mode

aschempp:
I'm using a custom global operation and backend module (`$GLOBALS['BE_MOD'][...]['foo'] = ['controller', 'method')` to create a database record with specific data. I can leverage the `DataContainer::create` method for that, which explicitly allows to pass additional data to the new records.

When creating the data, `DataContainer::create` redirects to the new record and in the process removes known "invalid" parameters from the URL. But since it does not remove `key`, it results in an endless redirect (because my module is called again). I think we can safely remove `key` as well since thats a well supported pattern imho.


leofeyer (approved):
I guess so, yes.

PR #4214, opened 2022-02-24
Adding the option `data-only` to the backup create command, to onyl export the data

cliffparnitzky:
To have the option to onyl export the data, a new option for the backup create command is added.

### DIFF
--- a/core-bundle/src/Command/Backup/BackupCreateCommand.php
+++ b/core-bundle/src/Command/Backup/BackupCreateCommand.php
@@ -29,7 +29,7 @@ class BackupCreateCommand extends AbstractBackupCommand
     protected function configure(): void
     {
         parent::configure();
-        
+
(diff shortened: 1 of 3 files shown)


Toflar:
I don't really see the point of creating a backup without the schema. But I'm not against it either, however, the target branch for features has to be 5.x :)

cliffparnitzky:
@Toflar The idea was born when maintaining the official demo. The exported SQL only contains the data. With this command, maintaining will be a little bit easier ;-).

aschempp:
> The idea was born when maintaining the official demo. 

Instead of adding this for Contao 5, we should adjust our SQL import to also import schema data! This is always a PITA for a demo, because the SQL import needs to match the database structure. This would no longer be the case if we lift the restriction 🙃

leofeyer:
As discussed in the Contao call, we do not want to add this feature because maintaining SQL dumps for themes is not what the feature was designed for.

PR #4216, opened 2022-02-26
Skip non-UTF-8 resources when syncing the DBAFS

### LINKED ISSUE #4215: Problem beim Synchronisieren
### Affected version(s)

4.13

### Description

Contao 4.13 Update problemlos
PHP 7.4.27
BE und FE OK
Im BE Klick auf Synchronisieren und Internal Server Error 
` An exception occurred while executing a query: SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xE4hrige...' for column `txbittel_`.`tl_files`.`path` at row 1`
Was kann man tun?

m-vo:
Fixes #4215

This PR adds a check for invalid directory/file paths and skips over them when syncing. In Contao 5 we can upgrade this to throwing a custom `VirtualFilesystemException` directly in `VirtualFilesystem#listContents()`. If needed, this exception can then be catched and converted into a meaningful error message (e.g. in the back end).


m-vo:
Here is a fun fact:

When performing a query similar to the one that caused the original issue (with a `\0xE` character in the path/name)…

```mysql
INSERT INTO tl_files (uuid, pid, path, hash, type, name, extension, tstamp, lastModified) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
```
image,

… the reported MySQL error will include something about "invalid datetime", which doesn't make any sense at all to me:
```
SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xE4bc.tx...' for column `<db>`.`tl_files`.`path` at row 1
```

fritzmg:
Would it be possible to simply list the paths with malformed characters? Would make it easy to get rid of them afterwards.

m-vo:
> Would it be possible to simply list the paths with malformed characters? Would make it easy to get rid of them afterwards.

I'm not sure. What would you expect if you simply listed affected files/driectories (in your code, so outside the DBAFS implementation)?

fritzmg:
Not sure what you mean? If the file synchronisation fails, because I copied a file or directory from a filesystem that does not use UTF-8, I would want to know which file it is, so I can fix it.

m-vo:
Yes, that would be possible with a custom exception like I suggested in the initial description?

ausi (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4217, opened 2022-02-26
Clean-up

leofeyer:
Mostly anonymous functions that are now fn() functions and inconsistencies after merging over 260 pull requests. 😄


m-vo:
Tbh, I found nearly all multi to single line conversions much more readable before. 🙈 Esp. for methods like `array_map` or queries.

PR #4218, opened 2022-02-27
Do not process redirected URLs outside base domains

### LINKED ISSUE #4213: Crawler also shows broken links on subpages of banner destination URL
### Affected version(s)

4.13.0

### Description

When crawling my links, the Broken Link Checker log shows not only broken links on my own home page, but also all the broken links that are on the pages pointed to by the banner target URLs in my Bugbuster banner module .
In the forum, Spooky said: "Possibly the 301 forwarding of the banner URLs is the problem. Since the banner URL could change at any time, maybe 302 or 307 would be better here (you could suggest that in the extension on GitHub)."

Additionally, on GitHub at Contao, you could suggest that the crawler (or broken link checker) not analyze other links that are on a page after a 301 redirect if the 301 target page is no longer in the same domain.

Since I don't understand enough of the subtleties myself, I'll post the link to this thread here so that you can understand what we discussed in detail. I have already written the above-mentioned message to Bugbuster.

[https://community.contao.org/de/showthread.php?82581-Crawler-also-shows-broken-links-on-third-party-pages-of-landing-pages-of-the-banner-module-&p=555552&posted=1#post555552](url)

Maybe you can take care of it. Thanks very much,

grashalm

fritzmg:
Fixes #4213

See https://github.com/contao/contao/issues/4213#issuecomment-1052022561 for a detailed explanation of the problem.

This PR fixes this by comparing the host of the actual URL of the response with the base URI collection and returning a negative decision in `needsContent`, so that any internal URL that was redirected to an external URL is not getting processed.

### DIFF
--- a/core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
+++ b/core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
@@ -96,7 +96,7 @@ class BrokenLinkCheckerSubscriber implements EscargotSubscriberInterface, Escarg
             return SubscriberInterface::DECISION_NEGATIVE;
         }
 
-        // Skip any redirected URLs that are now outside our base hosts
+        // Skip any redirected URLs that are now outside our base hosts (#4213)
         $actualHost = parse_url($response->getInfo('url'), PHP_URL_HOST);
 
         if ($crawlUri->getUri()->getHost() !== $actualHost && !$this->escargot->getBaseUris()->containsHost($actualHost)) {
--- a/core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
+++ b/core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
@@ -12,6 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Crawl\Escargot\Subscriber;
 
+use Nyholm\Psr7\Uri;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerAwareTrait;
 use Psr\Log\LogLevel;
@@ -99,7 +100,7 @@ class BrokenLinkCheckerSubscriber implements EscargotSubscriberInterface, Escarg
         ++$this->stats['ok'];
 
         // Skip any redirected URLs that are now outside our base hosts (#4213)
-        $actualHost = parse_url($response->getInfo('url'), PHP_URL_HOST);
+        $actualHost = (new Uri($response->getInfo('url')))->getHost();
 
         if ($crawlUri->getUri()->getHost() !== $actualHost && !$this->escargot->getBaseUris()->containsHost($actualHost)) {
             return SubscriberInterface::DECISION_NEGATIVE;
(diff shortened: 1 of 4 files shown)


fritzmg:
_Note:_ just to be clear, the broken link checker will still report on any internal links that redirect to non-existing external links. In the following example `http://c49.local/en/external` responds with a `302` redirect to `https://asdfjklasdfjkl.com/` - which is a host name that cannot be resolved and the broken link checker reports accordingly on it:

```
Crawling…
[Contao\CoreBundle\Crawl\Escargot\Subscriber\BrokenLinkCheckerSubscriber] [[object] (Terminal42\Escargot\CrawlUri: URI: http://c49.local/en/external (Level: 4, Processed: yes, Found on: http://c49.local/en/, Tags: none))] Broken link! Could not request properly: Couldn't resolve host name for "https://asdfjklasdfjkl.com/"..
```

Toflar (commented):
Looks good to me, I just wonder if we could facilitate this in Escargot but that's another discussion maybe.

core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
```diff
@@ -96,6 +96,13 @@ public function needsContent(CrawlUri $crawlUri, ResponseInterface $response, Ch
...
         }
 
+        // Skip any redirected URLs that are now outside our base hosts (#4213)
+        $actualHost = parse_url($response->getInfo('url'), PHP_URL_HOST);
+
+        if ($crawlUri->getUri()->getHost() !== $actualHost && !$this->escargot->getBaseUris()->containsHost($actualHost)) {
+            return SubscriberInterface::DECISION_NEGATIVE;
+        }
+
         ++$this->stats['ok'];
```
ausi: Shouldn’t `++$this->stats['ok'];` go above your new code block?

fritzmg: @ausi no, it would then report, that x additional broken links have been checked. But in reality they were discarded.

fritzmg:
> I just wonder if we could facilitate this in Escargot but that's another discussion maybe.

Yes, I have an idea actually. In Escargot we could set `max_redirects` to `0` and then catch every `RedirectException`. You could then update the `CrawlUri` instance with the URL - or even track every redirect there additionally. Then it would work seamlessly for every subscriber and the subscriber would not have to check for the actual URL via `->info` on the response.

This might constitute a BC break though.

Toflar:
I also don't think that's the right way to do it but we might provide the info to the subscriber in a simpler fashion. But let's discuss that over at the Escargot repo.

ausi (dismissed):

core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
```diff
@@ -96,6 +96,13 @@ public function needsContent(CrawlUri $crawlUri, ResponseInterface $response, Ch
             return SubscriberInterface::DECISION_NEGATIVE;
         }
 
+        // Skip any redirected URLs that are now outside our base hosts (#4213)
+        $actualHost = parse_url($response->getInfo('url'), PHP_URL_HOST);
```
Toflar: You should use `new Uri()` here.

Toflar (changes requested):
Should use the Uri library instead of `parse_url`.

core-bundle/src/Crawl/Escargot/Subscriber/BrokenLinkCheckerSubscriber.php
fritzmg: Hm, actually got confused there I think. Changed in 2a7072c48 .

fritzmg:
> Should use the Uri library instead of `parse_url`.

Done in c1d25a538

Toflar (approved):
👍 ❤️

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4219, opened 2022-02-27
Make the contao:install:lock command idempotent

richardhj:
Deployment-wise it would be great if the `contao:install:lock` command would be idempotent so that you can call the command on every deployment without working around the exception.

This change is also an easy way to increase coverage 🤣

### DIFF
--- a/installation-bundle/src/Command/LockCommand.php
+++ b/installation-bundle/src/Command/LockCommand.php
@@ -46,12 +46,6 @@ class LockCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        if (file_exists($this->lockFile)) {
-            $output->writeln('<comment>The install tool has been locked already.</comment>');
-
-            return 1;
-        }
-
         $fs = new Filesystem();
         $fs->dumpFile($this->lockFile, 3);
 


aschempp (dismissed):
👍  it would also be consistent with the `contao:maintenance-mode enable` command

m-vo (dismissed):

bytehead (dismissed):

leofeyer:
Isn‘t this a bugfix for Contao 4?

richardhj:
> Isn‘t this a bugfix for Contao 4?

It is an API change, though

aschempp:
ah, I didn't notice that. It could be a bugfix to me 😊

richardhj:
Rebased to 4.9 :)

installation-bundle/src/Command/LockCommand.php
```diff
@@ -46,12 +46,6 @@ protected function configure(): void
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        if (file_exists($this->lockFile)) {
```
leofeyer, line 49: We could still return early if the lock file already exists. Maybe even add a different message like @aschempp did here?

https://github.com/contao/contao/blob/dba2cf227e062bfc06e6b8c1a2c52c8df4fbb21b/manager-bundle/src/Command/MaintenanceModeCommand.php#L115-L124

richardhj: I thought about this as well. But what if the lock file is a) compromised or b) we increase the lock count to (e.g.) 5 in a future Contao version? Then you won't be able to lock the Install Tool with the command at all.

aschempp: I agree with @richardhj. The `contao:maintenance-mode enable` command always writes the maintenance template again, even if the file exists. This command should do the same, we don't know if the existing file contains the correct data.

Toflar (approved):

installation-bundle/src/Command/LockCommand.php
leofeyer: What about adding different messages as described above?

m-vo (approved):

installation-bundle/src/Command/LockCommand.php
aschempp: A message that says it was already enabled? We would need to check the file content for that, and we alao don't do that for maintenance mode 🤷‍♂️

leofeyer:
Thank you @richardhj.

PR #4224, opened 2022-03-01
Execute BackendTemplate#compile() when using the AbstractBackendController

m-vo:
Until we refactored the legacy templates' compile methods, we need to call them when rendering inside the `AbstractBackendController`. Otherwise things like the custom back end attributes (badge, …) and ua string are missing.

In Contao 5 we should refactor these things and move them into controllers/listeners.


ausi (dismissed):

bytehead (dismissed):

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #4225, opened 2022-03-02
Remove the wizard if a DCA field is not editable

### LINKED ISSUE #4173: Disable picker for disabled field
### Affected version(s)

4.13.0

### Description

If a field is disabled, a page picker for the field can still be used and is not disabled.

E. g. edit page -> ``Canonical URL`` -> ``Custom URL`` if ``Enable rel="canonical"`` is disabled in root page.

aschempp:
Fixes https://github.com/contao/contao/issues/4173

I've decided against handling each case, but just globally disable wizard if a field is disabled or readonly. I don't think any custom wizard should be available in that case either.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -516,7 +516,7 @@ abstract class DataContainer extends Backend
 			}
 		}
 
-		if ($wizard)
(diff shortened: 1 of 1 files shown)


bytehead (approved):

Toflar (approved):
It's a behavior change but I think it's the correct UX decision. I tested it out myself and set e.g. regular `start` and `stop` fields to `disabled` and I was actually surprised the datepicker still worked.

PR #4228, opened 2022-03-02
Improve the performance of contao:backup:create

Toflar:
This PR improves the performance of `contao:backup:create` dramatically for both, memory consumption as well as runtime.

* Memory consumption for large databases increases steadily if MySQL query buffering is active for PDO which seems to be the default. The solution is to disable this because it's not needed here anyway. This reduced 130 MB RAM to 44 MB RAM in my tests.
* Runtime optimizations are quite a few here but here are the most important ones (25 seconds -> 9 seconds):
    * `$table->getColumn($columnName)` was executed for every single value (so millions of times for bigger databases)
    * Introduced shortcuts for empty strings and integer types (which were quoted as strings for no reason), which reduced the number of `$connection->quote()` calls by like 80%
    * Removed lots of `sprintf()` calls


core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -78,46 +96,61 @@ private function dumpData(Connection $connection, Table $table): \Generator
...
+
+            yield "INSERT INTO `$tableName` ($insertColumns) VALUES ($insertValues);";
         }
     }
 
-    /**
-     * @param mixed $value
-     */
-    private function formatValueForDump($value, Column $column, Connection $connection): string
+    private function formatValueForDump(?string $value, int $columnBindingType, string $columnCharset, Connection $connection): string
```
Toflar: `$value` can only ever be `null` or `string` coming from the DB so we can typehint for that :)

core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -78,46 +96,61 @@ private function dumpData(Connection $connection, Table $table): \Generator
...
     {
         if (null === $value) {
             return 'NULL';
         }
 
+        if ('' === $value) {
+            return '';
```
ausi:
```suggestion
            return "''";
```
Should this be something like this?

aschempp: yes, @ausi is correct

aschempp (changes requested):

core-bundle/src/Doctrine/Backup/Dumper.php
Toflar: Of course, I've fixed and covered with tests in https://github.com/contao/contao/pull/4228/commits/962955401ef9361f518afdc089a6e0af1975a534.

ausi (dismissed):

core-bundle/src/Doctrine/Backup/Dumper.php
```diff
@@ -78,46 +96,61 @@ private function dumpData(Connection $connection, Table $table): \Generator
...
+        }
+
         // non-ASCII values
         if (
-            \is_string($value)
-            && preg_match('/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\xFF]/', $value)
+            preg_match('/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\xFF]/', $value)
             && (
-                !\in_array(strtolower($column->getPlatformOptions()['charset'] ?? ''), ['utf8', 'utf8mb4'], true)
+                !\in_array($columnCharset, ['utf8', 'utf8mb4'], true)
```
aschempp: since the `$columnCharset` seems only to be used here, could we do a `$isUtf8` instead and build a `$columnIsUtf8` array instead in the initial loop?

Toflar: Done in https://github.com/contao/contao/pull/4228/commits/53910cef9c6ed8a7599e308d52782c97e22fbc01. I've also added a test for boolean handling. Not sure if we could also optimize the non-ASCII values check there @ausi ? I mean, an early return maybe if `$isUtf8Charset` is `true/false`. Also, I noticed that there is a `$columnBindingType` for both, binary as well as ASCII stuff. Maybe our checks can even be simplified more by relying on them? https://github.com/doctrine/dbal/blob/3.3.x/src/ParameterType.php#L42-L47
(they are somehow mapped driver specifly https://github.com/doctrine/dbal/blob/e839cecc812fd04ec6c39fc865302b6a322d967d/src/Driver/PDO/Statement.php#L21-L29)

ausi (dismissed):

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4229, opened 2022-03-02
Use the Contao short URL service for keyboard shortcuts

leofeyer:
Follow-up on #4124.


aschempp (dismissed):

leofeyer:
I have reviewed the other links as well and I found quite a few other things that needed to be fixed, too. 🙈

Toflar (approved):
❤️

bytehead (approved):

aschempp:
why did you revert moving the link out of the language file?

leofeyer:
Because it only worked for this one link. The others cannot be removed, therefore we might as well leave them all in the language files. Transifex handles links very nicely, so it should not matter:

PR #4230, opened 2022-03-02
Fix undefined array index warnings for content elements and forms

fritzmg:
If you use

```php
/**
 * @ContentElement(category="Foobar")
 */
```

and you do not have a translation for that content element category yet (or don't want to use one), an undefined index warning will occur. This PR fixes that. It also fixes another instance related to the content element's name that was inconsistent.

// Update: just added another fix for an undefined index warning when you submit a front end form without file uploads.


aschempp (approved):

m-vo (approved):

leofeyer:
Thank you @fritzmg.

PR #4231, opened 2022-03-03
Remove the `$Template` property from the PageRegular class again

### LINKED ISSUE #4086: BC break with protected $Template property
### Affected version(s)

4.9.26

### Description

The fix from #4075 has still a BC break in `PageRegular`:

Bildschirmfoto 2022-02-08 um 14 14 25

leofeyer:
Fixes #4086

### DIFF
--- a/core-bundle/src/Resources/contao/pages/PageRegular.php
+++ b/core-bundle/src/Resources/contao/pages/PageRegular.php
@@ -20,15 +20,9 @@ use Symfony\Component\HttpFoundation\Response;
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
+#[AllowDynamicProperties]
 class PageRegular extends Frontend
 {
-	/**
-	 * @var Template
-	 *
-	 * @todo Remove in Contao 5.0
-	 */
-	protected $Template;
-
 	/**
 	 * Generate a regular page
 	 *
--- a/core-bundle/src/Resources/contao/pages/PageRegular.php
+++ b/core-bundle/src/Resources/contao/pages/PageRegular.php
@@ -20,7 +20,7 @@ use Symfony\Component\HttpFoundation\Response;
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-#[AllowDynamicProperties]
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/pages/PageRegular.php
```diff
@@ -20,15 +20,9 @@
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
+#[AllowDynamicProperties]
```
ausi:
```suggestion
 * @property Template $Template Will be made protected in Contao 5.0
 *
 * @author Leo Feyer <https://github.com/leofeyer>
 */
#[AllowDynamicProperties]
```

Should we add an info that this will change in Contao 5.0?

leofeyer: Since there is nothing to do in this file (the `$Template` property will be added to the `Controller` class), I did not add a `@todo`. But we might as well add one.

ausi: > the `$Template` property will be added to the `Controller` class

We sure about that? _(But I guess I should proably create a separate issue for that...)_

leofeyer: Yes, I guess:

https://github.com/contao/contao/blob/dba2cf227e062bfc06e6b8c1a2c52c8df4fbb21b/core-bundle/src/Resources/contao/library/Contao/Controller.php#L52-L57

But of course we can discuss this. (Preferably in a separate issue.)

leofeyer:
What is more problematic, is that `#[AllowDynamicProperties]` does not seem to fix the issue: https://github.com/contao/contao/runs/5403654666?check_suite_focus=true 😞

core-bundle/src/Resources/contao/pages/PageRegular.php
ausi:
```suggestion
#[\AllowDynamicProperties]
```

m-vo (approved):

PR #4232, opened 2022-03-03
Limit the number of suggestions in the front end preview

### LINKED ISSUE #3985: Limit the number of members in the preview bar
I would also add `LIMIT 20` to the db query, but it might be too opinionated, what do you think?

The first request (with two initial letters of the username) already returns 3000 results for me; does it make sense for browser to download, `JSON.parse` and insert 3000 elements into the DOM at this point? I find `select`s unusable anyway, when options list is higher then screen height. Or would you rather consider it an edge-case?

_Originally posted by @ameotoko in https://github.com/contao/contao/issues/3790#issuecomment-991067385_

leofeyer:
Fixes #3985

### DIFF
--- a/core-bundle/src/Controller/BackendPreviewSwitchController.php
+++ b/core-bundle/src/Controller/BackendPreviewSwitchController.php
@@ -177,30 +177,32 @@ class BackendPreviewSwitchController
                 $user->amg
             );
 
-            $andWhereGroups = "AND (groups LIKE '".implode("' OR GROUPS LIKE '", $groups)."')";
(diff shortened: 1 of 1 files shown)


aschempp (approved):
Could have used the query builder but I guess this works as well 😅

bytehead (approved):

leofeyer:
I have tried the query builder first, however, it does not handle our special cases such as `login='1'`. It will convert the string to an int, so the query ends up being `login=1`.

m-vo:
> It will convert the string to an int, so the query ends up being login=1.

IIRC the `setParameter()` method accepts a `$type` argument that allows you to control how it should get mapped.

Issue #3246, opened 2021-08-02
SubmitOnChange not triggering save_callback


koertho:
**Affected version(s)**

Contao 4.9 (occured in 4.9.16) and 4.11 (4.11.5)

**Description**

A field with submitOnChange, for example tl_member.login does not trigger the save_callback, when submitted async (e.g. the login checkbox is triggered and the subpalette ist loaded, the save_callback is not executed).

This was discussed in Slack (https://contao.slack.com/archives/CK4J0KNDB/p1627907859023700) with, among others, @ausi and @m-vo :)

leofeyer:
What exactly does your save_callback do?

koertho:
I wrote this issue some months ago, don't know the exact reason anymore, but from the slack-thread I think I wanted to manipulate the palette on palette change? Maybe :)

But to be honest, this is not important, as this Issue can be summarized to a single question: should the save_callback be triggered on submitOnChange events?

koertho:
I just read the old slack thread a little more carefully and I think I wanted to evaluate if the value of the login field was changed in tl_member.

leofeyer:
See #4236.

bytehead:
This issue still exists on 4.13.15.

PR #4236, opened 2022-03-03
Also create a new version if a checkbox is auto-submitted

### LINKED ISSUE #3547: Behaviour of submitOnChange
When toggling a field which has `submitOnChange` set, the record gets updated without creating a new version or adding any entry to `tl_log` - this can lead to some confusion...

**Story:**
We just ran into the problem that the site of a customer has been offline due to the root page not having assigned a layout anymore. Looking through the logs we saw nothing that could've explained how this happened. The customer sweared he didn't do anything and we knew for sure we did nothing so how did this happen?

I know how `submitOnChange` works and that a new version is only created once the user actually saves a modified record but having had this situation begs the question if the Core should at least create a new entry in the `System log` so one can understand what happened and why.

This isn't about blaming anyone for doing something wrong rather than making sure the change was done by someone privileged to do so instead of an unknown third party which might have gained access otherwise.

leofeyer:
Fixes #3547 #3246

This pull fixes the issue, however, the version drop-down menu will only show the new version after a full page reload. If this is not a problem, we can merge it.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Ajax.php
+++ b/core-bundle/src/Resources/contao/classes/Ajax.php
@@ -417,7 +417,13 @@ class Ajax extends Backend
 					if (Input::get('act') == 'editAll')
 					{
 						$this->strAjaxId = preg_replace('/.*_([0-9a-zA-Z]+)$/', '$1', Input::post('id'));
+
+						$objVersions = new Versions($dc->table, $this->strAjaxId);
+						$objVersions->initialize();
+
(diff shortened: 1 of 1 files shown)


fritzmg (approved):
> however, the version drop-down menu will only show the new version after a full page reload. If this is not a problem, we can merge it.

I am fine with that.

ausi (approved):

leofeyer:
As discussed in the Contao call, we want to merge this PR and create a new issue so we can discuss if auto-submit versions should be merged.

leofeyer:
See #4390 for the follow-up issue.

PR #4241, opened 2022-03-04
Do not use the `.tl_confirm` CSS class in the front end

### LINKED ISSUE #4239: js_autofocus no longer works for comments
### Affected version(s)

4.13.0

### Description

The CSS class no longer is named "confirm" now it's named "tl_confirm".

https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/templates/js/js_autofocus.html5#L4

You can extend this line in above template this way, then it will work again.
```js
    var p = document.querySelector('p.error') || document.querySelector('p.confirm') || document.querySelector('p.tl_confirm');
```

leofeyer:
Fixes #4239

### DIFF
--- a/comments-bundle/src/Resources/contao/templates/modules/mod_comment_form.html5
+++ b/comments-bundle/src/Resources/contao/templates/modules/mod_comment_form.html5
@@ -5,7 +5,7 @@
     <p class="info"><?= $this->login ?></p>
   <?php else: ?>
     <?php if ($this->confirm): ?>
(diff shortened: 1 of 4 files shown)


fritzmg (approved):

Toflar (approved):

leofeyer:
I think this change was wrong. I recall using `tl_confirm` in 15eed8462da160035afb019da026204eb2b23977 because that is what `Message::generateUnwrapped()` generates, too. Now the output is inconsistent. 😞

PR #4244, opened 2022-03-07
Make sure tl_content.type has an index

Toflar:
Thanks to APM with Tideways I found that the back end has become quite a bit slower in Contao 4.13 than it was the case previously. This seems to be due to more alias check queries in `tl_content` which are not present in 4.9. I couldn't believe we have no DB index on `tl_content.type`? `WHERE type='xy'` seems to be a very common query?!

I don't think we need an index on `cteAlias` because usually you won't have hundreds of alias elements but I'm fine adding one there too, wdyt @contao/developers?

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_content.php
+++ b/core-bundle/src/Resources/contao/dca/tl_content.php
@@ -47,7 +47,8 @@ $GLOBALS['TL_DCA']['tl_content'] = array
 			'keys' => array
 			(
 				'id' => 'primary',
-				'pid,ptable,invisible,start,stop' => 'index'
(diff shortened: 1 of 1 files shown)


aschempp:
wouldn't harm to optimize this in 4.9 as well?

Toflar:
> which are not present in 4.9

We can add the index but the rest is irrelevant.

m-vo (approved):
Good find! I always thought there was an index on `type`. :smile:

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #4248, opened 2022-03-08
Correctly encode redirect URLs between 401 pages and login forms

### LINKED ISSUE #3782: Query parameters are lost during redirect back after login
We want to use GA campaigns on a website that is behind a login. If the visitor isn't login already he will be redirected to the login page here:

https://github.com/contao/contao/blob/27f248764c86bd9929ec1b961c9241b5e593c9e2/core-bundle/src/Resources/contao/pages/PageError401.php#L137

But since the Environment::get('request') is not urlencoded everything after the first & in Environment::get('request')  is lost on the redirect back. If i change this line to
```
$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . urlencode(Environment::get('base') . Environment::get('request'));
```
it seems to work.

leonexcc:
Fixes #3782

Since the Environment::get('request') is not urlencoded everything after the first & in Environment::get('request') was lost on the redirect back.

Now if a 401 page redirects to a login page the originating url is now encoded correctly if they contain query strings.

### DIFF
--- a/core-bundle/src/Resources/contao/pages/PageError401.php
+++ b/core-bundle/src/Resources/contao/pages/PageError401.php
@@ -122,7 +122,7 @@ class PageError401 extends Frontend
 			}
 
 			// Add the referer so the login module can redirect back
-			$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . urlencode(Environment::get('base') . Environment::get('request'));
+			$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . rawurlencode(Environment::get('base') . Environment::get('request'));
 
 			/** @var UriSigner $uriSigner */
 			$uriSigner = System::getContainer()->get('uri_signer');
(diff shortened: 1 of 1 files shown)


fritzmg (dismissed):

core-bundle/src/Resources/contao/pages/PageError401.php
```diff
@@ -122,7 +122,7 @@ protected function prepare($objRootPage=null)
 			}
 
 			// Add the referer so the login module can redirect back
-			$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . Environment::get('base') . Environment::get('request');
+			$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . urlencode(Environment::get('base') . Environment::get('request'));
```
leofeyer: Are you sure that `urldecode()` instead of `rawurldecode()` is correct here?

ausi:
```suggestion
			$url = $objNextPage->getAbsoluteUrl() . '?redirect=' . rawurlencode(Environment::get('base') . Environment::get('request'));
```

leofeyer:
Thank you @leonexcc.

PR #4252, opened 2022-03-09
Fix the trigger_error() calls

leofeyer:
Follow-up on #4226

### DIFF
--- a/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
@@ -222,6 +222,7 @@ class FrontendIndex extends Frontend
 		{
 			/** @var PageRoot $objHandler */
 			$objHandler = new $GLOBALS['TL_PTY']['root']();
+
 			throw new ResponseException($objHandler->getResponse($objPage->id));
 		}
 
--- a/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
+++ b/core-bundle/src/Resources/contao/controllers/FrontendIndex.php
@@ -13,6 +13,7 @@ namespace Contao;
 use Contao\CoreBundle\Exception\AccessDeniedException;
 use Contao\CoreBundle\Exception\InsufficientAuthenticationException;
 use Contao\CoreBundle\Exception\PageNotFoundException;
+use Contao\CoreBundle\Exception\ResponseException;
 use Contao\Model\Collection;
 use Symfony\Component\HttpFoundation\Response;
 
@@ -221,7 +222,7 @@ class FrontendIndex extends Frontend
 		{
 			/** @var PageRoot $objHandler */
 			$objHandler = new $GLOBALS['TL_PTY']['root']();
-			$objHandler->generate($objPage->id);
+			throw new ResponseException($objHandler->getResponse($objPage->id));
 		}
 
 		// Set the admin e-mail address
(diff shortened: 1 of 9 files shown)


Toflar (dismissed):

bytehead (dismissed):

m-vo (dismissed):

leofeyer:
I had to remove the `trigger_error()` calls because there is no `$this->expectDeprecation()` method in the Symfony PHPUnit bridge in version 4.4 and thus I cannot fix the unit tests otherwise.

ausi:
> I had to remove the `trigger_error()` calls because there is no `$this->expectDeprecation()` method in the Symfony PHPUnit bridge in version 4.4

There is the `@expectedDeprecation` annotation instead. But this might require more changes in the tests so I’m happy either way.

ausi (dismissed):

core-bundle/src/Resources/contao/pages/PageError401.php
```diff
@@ -31,8 +31,6 @@ class PageError401 extends Frontend
 	 */
 	public function generate($objRootPage=null)
 	{
-		@trigger_error('Using PageError401::generate() has been deprecated in Contao 4.9 and will be removed in Contao 5.0. Use the PageError401::getResponse() method instead.');
```
aschempp, line 34: Looks like I forgot the `E_USER_DEPRECATED` constant, maybe that's the original issue? 😆

aschempp (changes requested):

core-bundle/src/Resources/contao/pages/PageError401.php
ausi: See 4cec36c79219b1c2efe10b69c6f4f2f30744bd5a

aschempp: how did you find that commit, since it doesn't belong to this branch? Was it a previous commit/attempt Github isn't showing?

ausi: Bildschirmfoto 2022-03-28 um 09 16 44

It is the first commit in this PR?

bytehead (dismissed):

core-bundle/src/Resources/contao/pages/PageError401.php
leofeyer: See https://github.com/contao/contao/pull/4252#issuecomment-1079157780

> I had to remove the `trigger_error()` calls because there is no `$this->expectDeprecation()` method in the Symfony PHPUnit bridge in version 4.4 and thus I cannot fix the unit tests otherwise.

core-bundle/src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -221,7 +222,7 @@ public function renderPage($pageModel)
 		{
 			/** @var PageRoot $objHandler */
 			$objHandler = new $GLOBALS['TL_PTY']['root']();
-			$objHandler->generate($objPage->id);
+			throw new ResponseException($objHandler->getResponse($objPage->id));
```
aschempp, line 226: ⬆️  this was the actual problem 😂

leofeyer: Thanks for debugging this. ❤️

PR #4255, opened 2022-03-10
Added similar news list hook to all other modules

baumannsven:
These hooks are used to implement your own logic and accordingly to obtain more information for evaluation, which is currently not available from the modules.

When programming a filter, I noticed that this is currently only reasonably possible with the news list module. I missed this in the other modules and therefore added it accordingly.

I would be pleased if these changes could be incorporated into the 4.13. Because in order to implement my logic completely, I have to make a not very nice hack.

### DIFF
--- a/calendar-bundle/src/Resources/contao/classes/Events.php
+++ b/calendar-bundle/src/Resources/contao/classes/Events.php
@@ -103,7 +103,49 @@ abstract class Events extends Module
 			return array();
 		}
 
+		$this->collectCalendars($arrCalendars, $intStart, $intEnd, $blnFeatured);
+
+		// HOOK: modify the result set
+		if (isset($GLOBALS['TL_HOOKS']['getAllEvents']) && \is_array($GLOBALS['TL_HOOKS']['getAllEvents']))
+		{
+			foreach ($GLOBALS['TL_HOOKS']['getAllEvents'] as $callback)
+			{
+				$this->import($callback[0]);
+				$this->arrEvents = $this->{$callback[0]}->{$callback[1]}($this->arrEvents, $arrCalendars, $intStart, $intEnd, $this);
+			}
+		}
+
+		return $this->arrEvents;
+	}
+
+	/**
+	 * Collect the matching calendars.
+	 *
+	 * @param array   $arrCalendars
+	 * @param integer $intStart
+	 * @param integer $intEnd
+	 * @param boolean $blnFeatured
+	 */
+	protected function collectCalendars($arrCalendars, $intStart, $intEnd, $blnFeatured)
+	{
 		$this->arrEvents = array();
+
+		// HOOK: add custom logic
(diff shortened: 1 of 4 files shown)


bytehead (commented):
We should use proper events and not legacy hooks for feature stuff 🙃

baumannsven:
@bytehead 

I agree with you there. I had implemented this as a HOOK because I didn't want to mess anyone up.

I can also change this to EVENTS if you want.

bytehead:
As of SemVer 4.13 will not get any new features, so it should target at 5.x IMO. 
And you don't upset anyone with proper events 😉

news-bundle/src/Resources/contao/modules/ModuleNewsArchive.php
```diff
@@ -189,51 +192,107 @@ protected function compile()
...
+				{
+					continue;
+				}
+
+				if ($objCollection === null || $objCollection instanceof Collection)
+				{
+					return $objCollection;
+				}
+			}
+		}
```
fritzmg, line 268: The same hook should not be re-used for the news archive in my opinion.

calendar-bundle/src/Resources/contao/classes/Events.php
```diff
@@ -103,8 +103,50 @@ protected function getAllEvents($arrCalendars, $intStart, $intEnd, $blnFeatured
...
+				{
+					continue;
+				}
+
+				if (\is_array($this->arrEvents))
+				{
+					return;
+				}
+			}
+		}
```
fritzmg, line 148: imho it is difficult or even unnecessary to provide hooks/events for every conceivable use-case. May be in your case it's better to implement your own output modules instead?

baumannsven: It is not my intention to implement a HOOK for every application. My idea is that the HOOK is executed first in the message list module and that one can implement one's own logic through it. Because the existing HOOK is simply too late at this point.

news-bundle/src/Resources/contao/modules/ModuleNewsArchive.php
baumannsven: You can see via the supplied instance of the module whether you are in the archive module or in the list module. My thought here was that if, for example, you want to have a filter that serves both, then this does not have to be handled separately.

If the pull request should be accepted, be it as EVENT (5.x) or as HOOK. Then I could also separate them.

fritzmg: > My thought here was that if, for example, you want to have a filter that serves both, then this does not have to be handled separately.

You don't have to handle it separately. You can still use one service to handle both cases.

baumannsven: Okay then I will rename that to `newsArchiveFetchItems `. For now, I'll wait and see what the basic decision is on this PR.

PR #4256, opened 2022-03-10
Allow fragment reference with fake model instance

### LINKED ISSUE #4108: Create ContentElement or ModuleElement from "faked" model
### Affected version(s)

4.9.12

### Description

Sometimes we need or use the capability to render "fake" Content Elements / Frontend Modules.

For example our custom element / module will reuse the `ContentText` and `ContentDownloads` element to combine it into something new. This way we can also make use of the existing stylings and be more update safe.

For a "fake" Content Element we will create a new `ContentModel`, set the necessary fields and render it:

```php
$oModel = new ContentModel();
$oModel->type = 'image';
$oModel->singleSRC = $some->uuid;
$oModel->fullsize = 1;
...

$oElement = new ContentImage($oModel);

$template->image_element = $oElement->generate();
```

With all the "legacy" core elements this is not a problem so far. But as we transition to the newer Content Element Controller we noticed this is no longer possible.

We expected to see the same behaviour with the `ContentProxy` so we built up the `ContentModel` like above and go on like before:

```php
$oModel = new ContentModel();
$oModel->type = 'somethingawesome';
$oModel->customStuff = 'whatever';
...

$oElement = new ContentProxy($oModel);

$template->awesome_element = $oElement->generate();
```
But this will lead to the following error: `Could not resolve argument $model of "contao.fragment._contao.content_element.somethingawesome()", maybe you forgot to register the controller as a service or missed tagging it with the "controller.service_arguments"?`, on Contao 4.12.

Also setting a "fake" `id` in the `ContentModel` will not help.

From what I understand the problem lies in the way the `ContentProxy` stores and gets its reference.
https://github.com/contao/contao/blob/56e6e3b31bb4450842b4557122752b94fcf12fc4/core-bundle/src/Resources/contao/elements/ContentProxy.php#L31
The `ContentProxy` stores a reference using `ContentElementReference`
https://github.com/contao/contao/blob/56e6e3b31bb4450842b4557122752b94fcf12fc4/core-bundle/src/Resources/contao/elements/ContentProxy.php#L43

Within `ContentElementReference` only the `type` and `id` is really used, the rest of the given `ContentModel` gets thrown away.
https://github.com/contao/contao/blob/56e6e3b31bb4450842b4557122752b94fcf12fc4/core-bundle/src/Fragment/Reference/ContentElementReference.php#L24-L32

When the fragment handler tries to render this element the `ModelArgumentResolver` will kick in and tries to fetch the original model from the database:
https://github.com/contao/contao/blob/56e6e3b31bb4450842b4557122752b94fcf12fc4/core-bundle/src/HttpKernel/ModelArgumentResolver.php#L93-L96
This way we're completely dependant on an already existing `ContentElement` from the database and loose the capability to render "fake" Content Elements on the fly.

We're aware that the current implementation was done this way so the rendered elements can be cached for performance reasons but this is not applicable in such use case since our final custom element will be cached anyway 🤷‍♂️

One last thing: Everything mentioned above also applies to `ModuleProxy`.

aschempp:
Fixes https://github.com/contao/contao/issues/4108


core-bundle/src/Fragment/Reference/ContentElementReference.php
```diff
@@ -30,4 +30,12 @@ public function __construct(ContentModel $model, string $section = 'main', array
         $this->attributes['classes'] = $model->classes;
         $this->attributes['templateProperties'] = $templateProperties;
     }
+
+    public static function withModel(ContentModel $model, string $section = 'main', array $templateProperties = []): ContentElementReference
```
Toflar: I don't like the naming because the constructor also accepts a model. How about `createForInlineRenderer()` or so?

Toflar: I don't like the naming because the constructor also accepts a model. How about `createForInlineRenderer()` or so?

aschempp:
I have now updated the PR to be more clever. The `ContentProxy` and `ModuleProxy` now automatically work if the model is not registered, so the original code in the originating issue still works as expected 😎

Toflar (approved):

ausi (approved):

PR #4261, opened 2022-03-13
Fix SQL error in purge expired data cron

ausi:
Executing this cron runs into the following error: `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':tstamp' at line 1`

`Connection::prepare()` does not support named parameters AFAIK (or maybe that changed in a recent Doctrine version?).

### DIFF
--- a/core-bundle/src/Cron/PurgeExpiredDataCron.php
+++ b/core-bundle/src/Cron/PurgeExpiredDataCron.php
@@ -50,7 +50,7 @@ class PurgeExpiredDataCron
         }
 
         $this->connection->executeStatement(
-            sprintf('DELETE FROM %s WHERE tstamp<:tstamp', $table),
+            "DELETE FROM $table WHERE tstamp < :tstamp",
             ['tstamp' => time() - $period],
             ['tstamp' => Types::INTEGER],
         );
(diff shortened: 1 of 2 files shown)


ausi:
Not sure if this needs fixing in Contao 4.9 too? as it might be Doctrine 3.0 related?

fritzmg:
Seems odd, why would they drop support for that? 🤔

core-bundle/src/Cron/PurgeExpiredDataCron.php
```diff
@@ -48,7 +49,10 @@ private function cleanTable(string $table, int $period): void
             return;
         }
 
-        $stmt = $this->connection->prepare(sprintf('DELETE FROM %s WHERE tstamp<:tstamp', $table));
-        $stmt->executeStatement(['tstamp' => time() - $period]);
```
leofeyer, line 52: Isn‘t the actual problem that the array key has to be `':tstamp'` instead of `'tstamp'`?

Toflar: No, this changed in DBAL 3.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4262, opened 2022-03-13
Fix SQL commands not supported in prepared statements

ausi:
`Database::getInstance()->query("LOCK TABLES")` currently runs into the following error with MySQLi:
`SQLSTATE[HY000]: This command is not supported in the prepared statement protocol`


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4265, opened 2022-03-13
Use service_closure instead of lazy service

### LINKED ISSUE #4154: Lazy Services
As discussed in Slack, #4151 was a quickfix to make the Contao Manager work with Contao 4.13.0.

We should discuss if we want to keep the `lazy: true` there and therefore use `lazy: true` for other things as well in the future, or if we should go with a different solution instead (and decide to never use `lazy: true`).

What I don’t like about proxy classes is, that they are not visible to the developer who is using them. Writing code like the following will not work as expected if the service is lazy:

```php
function doSomethingRisky(MyService $myService) {
	try {
		$myService->somethingRisky();
	} catch (\Throwable $exception) {
		return false;
	}
	
	return true;
}
```

Such a lazy feature needs to be visible to the developer IMO.  
In an optimal world this would look like this:

```php
function doSomethingRisky(Lazy<MyService> $myServiceProxy) {
	$myService = $myServiceProxy->get();
	
	try {
		$myService->somethingRisky();
	} catch (\Throwable $exception) {
		return false;
	}
	
	return true;
}
```

Currently possible would be something like this:

```php
function doSomethingRisky(MyService&LazyLoadingInterface $myService) {
	$myService->initializeProxy();
	
	try {
		$myService->somethingRisky();
	} catch (\Throwable $exception) {
		return false;
	}
	
	return true;
}
```

ausi:
As discussed in https://github.com/contao/contao/issues/4154#issuecomment-1046668260

### DIFF
--- a/core-bundle/src/Cron/Cron.php
+++ b/core-bundle/src/Cron/Cron.php
@@ -23,8 +23,16 @@ class Cron
     public const SCOPE_WEB = 'web';
     public const SCOPE_CLI = 'cli';
 
-    private CronJobRepository $repository;
-    private EntityManagerInterface $entityManager;
+    /**
+     * @var \Closure():CronJobRepository
+     */
+    private \Closure $repository;
+
+    /**
+     * @var \Closure():EntityManagerInterface
+     */
+    private \Closure $entityManager;
+
     private ?LoggerInterface $logger;
 
     /**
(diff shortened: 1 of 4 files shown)


Toflar (dismissed):

fritzmg (commented):
Wait, I thought we only wanted to make the `@contao.cron` service lazy for the `contao.listener.command_scheduler`?

Toflar:
No, the opposite :)

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4271, opened 2022-03-14
Correctly show all breadcrumb items

### LINKED ISSUE #4128: Extend backend breadcrumb
### Description

With Contao 4.13 the backend breadcrumb was optimized.
see: #3727

This currently only works with content-elements under «Content > Articles».
Can the concept be implemented universally for Contao so that this can also be used under other areas such as layout, forms, etc.

Possibly, this makes no sense at all?
Understand this ticket as an idea. If this is crap, please feel free to close the ticket.

### LINKED ISSUE #4176: Breadcrumb Themes
### Affected version(s)

4.13.0

### Description

In breadcrumbs in the Themes section the theme title is missing.

E. g. instead of ``Themes › $ThemeTitle › Front end modules`` in previous versions, only ``Themes ›  Front end modules`` is shown.

leofeyer:
Fixes #4176

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Backend.php
+++ b/core-bundle/src/Resources/contao/classes/Backend.php
@@ -536,7 +536,10 @@ abstract class Backend extends Controller
 				$ptable = $act != 'edit' ? ($GLOBALS['TL_DCA'][$strTable]['config']['ptable'] ?? null) : $strTable;
 				$container = System::getContainer();
 
-				$this->loadDataContainer($ptable);
+				if ($ptable)
+				{
+					$this->loadDataContainer($ptable);
+				}
 
 				while ($ptable && !\in_array($GLOBALS['TL_DCA'][$table]['list']['sorting']['mode'] ?? null, array(DataContainer::MODE_TREE, DataContainer::MODE_TREE_EXTENDED)) && ($GLOBALS['TL_DCA'][$ptable]['config']['dataContainer'] ?? null) === 'Table')
 				{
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -534,9 +534,10 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 				$pid = $dc->id;
 				$table = $strTable;
 				$ptable = $act != 'edit' ? ($GLOBALS['TL_DCA'][$strTable]['config']['ptable'] ?? null) : $strTable;
-
 				$container = System::getContainer();
 
+				$this->loadDataContainer($ptable);
```
ausi: `$ptable` can be `null` here, should it fallback to the table instead?

```suggestion
				$this->loadDataContainer($ptable ?? $strTable);
```

fritzmg (dismissed):

core-bundle/src/Resources/contao/classes/Backend.php
leofeyer: I don‘t think we have to load the DC at all if `$ptable` is `null`, do we? So maybe we should just wrap it with `if ($ptable)` as in line 577?

ausi: 👍

leofeyer: Changed in 7c925bd2ec062900b0b8f39fa41f2acf4a17c251.

Toflar (approved):

Total-Reality:
Great work. Thank you very much!

PR #4272, opened 2022-03-15
Add a help wizard if the canonical URL fields are disabled

### LINKED ISSUE #4177: Hint if rel="canonical" is disabled in root page
### Affected version(s)

4.13.0

### Description

A hint at the disabled fields ``Custom URL`` and ``Query parameters`` would be helpful if `` Enable rel="canonical"`` is disabled in the root page.

(Similar to the hint at the field ``URL prefix`` and other if legacy routing is used.)

leofeyer:
Fixes #4177


core-bundle/src/EventListener/DataContainer/RelCanonicalListener.php
```diff
@@ -0,0 +1,57 @@
...
+            'show.svg',
+            '',
+            sprintf(
+                'title="%s"',
+                StringUtil::specialchars($this->translator->trans('tl_page.relCanonical', [], 'contao_tl_page'))
+            )
+        );
+
+        foreach (['canonicalLink', 'canonicalKeepParams'] as $field) {
+            $GLOBALS['TL_DCA']['tl_page']['fields'][$field]['eval']['disabled'] = true;
```
ausi: Does this field now get always diesabled?
And isn’t this handled already by https://github.com/contao/contao/blob/0ed2c1e25c49d8d6557201423e2be6976ba11ba9/core-bundle/src/EventListener/DataContainer/DisableCanonicalFieldsListener.php#L42

Toflar: Yeah we already have a `DisableCanonicalFieldsListener`, I'd prefer having the logic in the same one (even if that means renaming it but 2 separate seems strange).

leofeyer: You are right. I have just copied the `LegacyRoutingListener`, but the context is different there. 🙈

leofeyer: Please review again.

m-vo (approved):

fritzmg (approved):

Toflar (approved):

ausi (approved):
❤️

PR #1468, opened 2020-03-02
Use "isSortable" in file and page trees without orderField

ausi:
Instead of `'orderField' => 'orderSrc'` we can now use `'isSortable' => true`.

In cases where folders are selectable and the files can be sorted (`isGallery` and `isDownloads`) `orderField` is still required.

PR #4281, opened 2022-03-15
Add missing isSortable checks to the picker widget

MarkejN:
The `isSortable` checks were added in https://github.com/contao/contao/pull/1468 but I noticed that the `sortable` class is not set when I use the picker widget. I added two more checks, just like it was done in the pageTree and fileTree widgets:

https://github.com/contao/contao/blob/01d8f3ea25115aaf01297c9662ed59f7142a858e/core-bundle/src/Resources/contao/widgets/PageTree.php#L199


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @MarkejN.

PR #4283, opened 2022-03-15
Update the codebase to PHP 8.1

leofeyer:
As basis for the discussion on the upcoming developers meeting, this PR updates the codebase to PHP 8.1 using Rector. Every commit is one group of changes, so we can undo single commits if we want to.

- f1afde64d3b894a7833f104fcf0d7411837693b1: constructor property promotion
- b8f4519004ee3777be27c0c7d5eaf534bc00596a: union types
- d0b93e7a447e3462cc110431edbde804dad8fa54: final public class constants
- 2bb6bd03e1ab598ef396f3a9ba04d2fc9b567cb5: `::class` on objects
- 9fc55bcff08cc4b81e05145581d6a9fe57496440: `str_starts_with()`, `str_contains()` and `str_ends_with()`
- 5ccd71c5cb716e3fd76495c3dc9ae2f832a2828c: catching exceptions without capturing them
- <del>4bac38c12cd8d38dfb49d810fe8865ae1ea8570c: exceptions on JSON errors</del> (reverted, because it should be a separate pull)
- 258ddc12dbfa38c5b97efb83dc1e410b2280d4c7: return type `:never`
- f15df0eef6af888518e5e8e78491a69fb7fdb545: implement `\Stringable` if there is a `__toString()` method
- 3e51c4e7cb106e9b16a3ee6b86dca149a1991f43: spread operator instead of `array_merge()`
- 1a8d4e90908c6ecd99141d499e965b2895da8c64: misc

We should also talk about `?PageModel $page` vs `PageModel|null $page`.


m-vo:
Do we also want to make properties `readonly` where possible? (this also works in constructor property promotion)

core-bundle/src/Twig/Interop/ContextFactory.php
```diff
@@ -116,7 +116,7 @@ private function getAllMembers(object $object): \Generator
      */
     private function getCallableWrapper(callable $callable, string $name): object
     {
-        return new class($callable, $name implements \Stringable) {
```
m-vo: IMHO these should work if you write it like this:

```php
return new class($callable, $name) implements \Stringable {
    // …
};
```

PR #4290, opened 2022-03-16
Do no longer fall back to "web" if "public" does not exist

leofeyer:


bytehead (commented):
Needs to be changed as well: https://github.com/contao/contao/blob/5.x/core-bundle/src/DependencyInjection/ContaoCoreExtension.php#L102-L103

m-vo (approved):

PR #4293, opened 2022-03-16
Revert 'Do not use the `.tl_confirm` CSS class in the front end'

leofeyer:
Fixes https://github.com/contao/contao/pull/4241#issuecomment-1069371125

The rule is

- use `tl_confirm` and `tl_error` for messages that would have been generated by `Messages::generateUnwrapped()` (usually **above** the `<form>` tag)
- and use `confirm` and `error` for form related messages (usually **inside** the `<form>` tag).

Regarding the original issue #4239, I am not sure if we should autofocus both types of message. @contao/developers WDYT?


aschempp:
Honestly I don't think it matters. But you could also change `Messages::generateUnwrapped()` to not use `tl_` in the front end?

leofeyer:
Yes, I would like to do this in Contao 5 to guarantee that `tl_` classes are only used in the back end. But not in Contao 4.9 because that would force everyone to adjust their style sheets.

fritzmg (changes requested):
`js_autofocus` still uses `p.confirm`. Imho the change to

```js
var p = document.querySelector('p.error') || document.querySelector('p.tl_confirm');
```
or
```js
var p = document.querySelector('p.error') || document.querySelector('p.confirm')) || document.querySelector('p.tl_confirm');
```
needs to be part of this PR. I'd vote for the latter.

leofeyer:
First of all, your suggestion is missing `.tl_error`. Second of all, if you add `.tl_confirm` at the end and there is both a `.tl_confirm` element and a `.confirm` element, the script will not focus the message above the form but the one inside the form. Not sure if that is what we want.

fritzmg:
> First of all, your suggestion is missing .tl_error

But `tl_error` is not used this way in front end forms?

leofeyer:
If you use `Message::generate()`, it can be `tl_error`, `tl_confirm`, `tl_new` or `tl_info`.

fritzmg:
The purpose of `js_autofocus` is to scroll to form fields/sections that need attention. None of the front end modules or the form class use `Message::generate()`.

leofeyer:
I don‘t understand your logic. So the confirmation message in the `mod_changePassword` template needs attention but the error message in the `mod_two_factor` template doesn‘t? 😕

fritzmg:
What I mean is that this revert might be wrong in general. `js_autofocus` is supposed to focus on elements that need attention in the front end, like the aforementioned messages in the modules or form fields that failed server side validation.

I am not sure what the use case for `Message::generate()` is. While you can use `Message::generate()` in the front end if you want to, none of the core modules currently do and thus I also don't think `js_autofocus` needs to accomodate to that?

However - maybe `js_autofocus` is _only_ supposed to focus on form _fields_ and not on the general messages generated by these modules? I am now uncertain about its original use case.

leofeyer:
> However - maybe `js_autofocus` is _only_ supposed to focus on form _fields_ and not on the general messages generated by these modules? I am now uncertain about its original use case.

Yes, that was the original use case.

fritzmg (approved):
> Yes, that was the original use case.

👍

Issue #4257, opened 2022-03-10
Conflicting Controller::reset method


aschempp:
### Affected version(s)

4.9.27

### Description

in https://github.com/contao/contao/pull/4091 we added a new `public static reset()` method to `Controller` to reset the internal class properties on kernel shutdown. Unfortunately, this `reset` method might exist in a lot of other classes that _due to the Contao "default"_ inherit from `Controller`. Isotope for example does in https://github.com/isotope/core/blob/f22f6e3a27db8d870e8daeb6923b8930a9acee03/system/modules/isotope/library/Isotope/Backend/Product/Fallback.php#L26.

I wonder if we should rename that method to something "internal" like `public static function controller_cache_reset()`? Since the method name is irrelevant to Contao, it would sure be less conflicting than something like `reset`.

ausi:
How about `resetControllerCache()` or `resetContaoControllerCache()` ?

aschempp:
see https://github.com/contao/contao/pull/4294

PR #4294, opened 2022-03-17
Rename the Controller::reset() method

aschempp:
Fixes https://github.com/contao/contao/issues/4257


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1450,7 +1450,11 @@ public static function loadDataContainer($strTable, $blnNoCache=false)
 		$loader->load($blnNoCache);
 	}
 
-	public static function reset()
+	/**
+	 * Do not name this "reset" because it might result in conflicts with child classes
+	 * @see https://github.com/contao/contao/issues/4257
+	 */
+	public static function resetStaticCache()
```
Toflar: `@internal` maybe?

aschempp: I thought about that as well, but the we should make all our reset methods `@internal`? And maybe rename them to `resetStaticCache`?

ausi (dismissed):

Toflar (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
Toflar: I think we can discuss that at the devs meeting real quick. @leofeyer can you put that on the list? Should be a matter of a few minutes.

m-vo:
I get the intention, but now we're renaming a public method that is not internal. That doesn't feel right.

Toflar:
It's only been introduced in the latest version. It can be considered a bugfix imho.

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: Done. ✔️

ausi (dismissed):

ausi (approved):

Toflar (approved):

PR #4296, opened 2022-03-17
Improve version diff of recursive array data

aschempp:
I'm using https://github.com/m-vo/contao-group-widget here but I think it applies to any recursive array. Be aware that this only affects fields with `eval => multiple = true` set, otherwise the source data is not unserialized (e.g. it does not affect headline fields).

**Before:**
Bildschirmfoto 2022-03-17 um 14 15 14

**After:**
Bildschirmfoto 2022-03-17 um 14 11 36

_Screenshots are not exactly the same content but the same markup_ 😉

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -822,14 +822,14 @@ class Versions extends Controller
 	 *
 	 * @return string
 	 */
-	protected function implodeRecursive($var, $binary=false)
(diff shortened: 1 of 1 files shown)


Toflar (approved):

m-vo (approved):

leofeyer (changes requested):
You did not test this, did you?

leofeyer:
Also, the change has some strange side-effects:



This version diff of a table element looks like this with the changes:

PR #4299, opened 2022-03-18
Fix invalid news content element ID

### LINKED ISSUE #2201: Restricted users cannot use "save and new" in the news module
**Affected version(s)**

Contao 4.8+ (maybe older ones, too)

**Description**

Create a news item and then a CE 'text'. Write some text and then use the editor's 'insert link' feature to link to another news item. Try to save the CE with the button "Save and new" – works (if logged in as admin). Try the same but as restricted user – and it fails. 

What I've found out: 
- It fails because checkAccessToElement() fails (in class 'tl_content_news')
- the method fails because the methods first SQL query yields an empty result 
- result is empty because $id is strange
- $id resolves to CURRENT_ID
- CURRENT_ID gets filled with $objSession->get('CURRENT_ID')
- the session's CURRENT_ID gets changed as soon as you use the 'insert link' feature and choose a news archive (to choose another news item)

Resulting error (taken from the Contao demo installation):
Bildschirmfoto 2020-08-26 um 18 42 13

The content element doesn't have the ID 1, but the news archive does have this ID …

Tested in a Contao 4.8.8 installation and in the Contao demo installation.

ausi:
Fixes #2201

https://user-images.githubusercontent.com/367169/159057599-85424cf6-3b0d-43c4-8133-afca81f09660.mov

### DIFF
--- a/news-bundle/src/Resources/contao/dca/tl_content.php
+++ b/news-bundle/src/Resources/contao/dca/tl_content.php
@@ -59,10 +59,14 @@ class tl_content_news extends Contao\Backend
 		{
 			case '': // empty
 			case 'paste':
-			case 'create':
 			case 'select':
 				// Check access to the news item
 				$this->checkAccessToElement(CURRENT_ID, $root, true);
+				break;
+
+			case 'create':
+				// Check access to the parent element if a content element is created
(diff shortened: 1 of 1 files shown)


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4306, opened 2022-03-21
Drop the initialize.php BC layer

aschempp:


Toflar (changes requested):
The InitializeControllerListener can go too, no?

Toflar (dismissed):

fritzmg (dismissed):

bytehead (dismissed):

leofeyer (changes requested):
There is more to remove: both the `InitializeApplicationListener` and the `InstallCommandListener` contain references to `system/initialize.php`.

leofeyer (approved):

PR #4316, opened 2022-03-21
Drop legacy routing

aschempp:


aschempp:
The failing service linter error seems unrelated, and I'm unsure we can/should change that, but not sure how to "bypass" the fixer check?

leofeyer (approved):

PR #4319, opened 2022-03-21
Drop the user agent class and insert tag

aschempp:


Toflar (dismissed):

bytehead (dismissed):

aschempp:
@m-vo any idea why the controller test is failing?

fritzmg:
> @m-vo any idea why the controller test is failing?

https://github.com/aschempp/contao/pull/9 :)

fritzmg (changes requested):
As mentioned in https://github.com/aschempp/contao/pull/9#issuecomment-1074403788 I think all the back end templates now need to be a little bit refactored. Currently `be_login` looks like this for example:

```php
<body class="<?= $this->ua ?>"<?= $this->attributes ?>>
```

Since `$this->ua` will now always be an empty string (unless `Config::get('fullscreen')` is true) we will end up with

```html
<body class="">
```

So it should be changed to:

```php
<body<?php if ($this->ua): ?> class="<?= $this->ua ?>"<?php endif; ?><?= $this->attributes ?>>
```

Or may be the variable should be renamed to `class`:

```php
<body<?php if ($this->class): ?> class="<?= $this->class ?>"<?php endif; ?><?= $this->attributes ?>>
```

ausi (dismissed):

core-bundle/src/Resources/contao/templates/backend/be_confirm.html5
```diff
@@ -59,4 +59,4 @@
   </div>
 
 </body>
-</html>
\ No newline at end of file
+</html>
```
leofeyer: Please revert these changes.

leofeyer (changes requested):

core-bundle/src/Resources/contao/templates/backend/be_confirm.html5
aschempp: just had to merge upstream and now it's gone 🙃

core-bundle/src/Resources/contao/classes/BackendTemplate.php
```diff
@@ -68,8 +68,7 @@ public function getResponse()
 	 */
 	protected function compile()
 	{
-		// Backwards compatibility (see #3074 and #6277)
-		$this->ua = Environment::get('agent')->class;
+		$this->ua = '';
```
leofeyer, line 71: This should be removed entirely, shouldn‘t it?

leofeyer:
Maybe we should rename `ua` to `class` so it is the same as in the front end templates?

```php
<?php if ($this->class): ?> class="<?= $this->class ?>"<?php endif; ?>
```

I can create a pull request if you agree.

fritzmg:
Yes, see the previous comments. I am in favour of renaming it to `class`.

leofeyer:
See https://github.com/aschempp/contao/pull/10

leofeyer (approved):

PR #4321, opened 2022-03-21
Drop the redundant event names

aschempp:
The two remaining event names are not optional because the same event object is dispatched in different contexts.


fritzmg:
Should we really change this? How will you be able to subscribe to an event via service tagging in 4.13 and 5.x?

aschempp:
> Should we really change this? How will you be able to subscribe to an event via service tagging in 4.13 and 5.x?

I think it already works if you only use the event object? It currently just also works with the name. But maybe someone else knows for sure 🙃

fritzmg:
> I think it already works if you only use the event object?

You mean when using autoconfiguration? Yes, that will work. But if you define your event listener like this:

```yaml
# config/services.yaml
services:
    App\EventListener\BackendMenuBuildListener:
        tags:
            - { name: kernel.event_listener, event: 'contao.backend_menu_build' }
```

It won't work anymore (just tested it).

aschempp:
I think you would need to use the FQCN as event name then?

fritzmg:
> I think you would need to use the FQCN as event name then?

Yes, but that will only work in Contao 5.0, not 4.13.

core-bundle/src/Event/ContaoCoreEvents.php
```diff
@@ -41,39 +27,4 @@ final class ContaoCoreEvents
      * @see ImageSizesEvent
      */
     public const IMAGE_SIZES_USER = 'contao.image_sizes_user';
```
leofeyer, line 29: Do we really want to keep the `ContaoCoreEvents` class for this or can we use `contao.image_sizes_user` instead of the constant?

m-vo: IMHO we should deprecate the file and fully switch to FQCN as event names. That usage should also be possible with Symfony 4.4 when registering the events as aliases:  https://symfony.com/doc/current/event_dispatcher.html#event-aliases.

fritzmg (changes requested):
I still think we should not remove these. Even Symfony [still keeps all their event names](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/HttpKernel/KernelEvents.php). Instead we should create aliases and add them to the `%event_dispatcher.event_aliases%` parameter, just like Symfony does:

* https://github.com/symfony/symfony/blob/a51197bc4fa65183ccb07be4d32c64a360f39d2c/src/Symfony/Component/HttpKernel/KernelEvents.php#L113-L127
* https://github.com/symfony/symfony/blob/a51197bc4fa65183ccb07be4d32c64a360f39d2c/src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php#L53-L59

This way we can unify the current state as some events do have names (even recently introduced ones) and some don't.

aschempp:
> Instead we should create aliases and add them to the `%event_dispatcher.event_aliases%` parameter, just like Symfony does

We already do have these aliases (I removed them in this PR). I thought that event names are deprecated, but apparently they are not if Sf still has them?

fritzmg:
> I thought that event names are deprecated, but apparently they are not if Sf still has them?

Yes, I thought so too, but even Symfony's Event Dispatcher Component documentation specifically says to use an event name when dispatching: https://symfony.com/doc/current/components/event_dispatcher.html#dispatch-the-event

Within Symfony itself however, all "old" events are dispatched with their respective names, whereas newer events are dispatched without a specific name (i.e. only the FQCN is used).

My assumption is that Symfony still keeps the event names so that any event listener/subscriber still registered to the event's name does not have to be adjusted.

imho we should do the same.

aschempp:
Sounds good to me. Closing this then 🙃

PR #4323, opened 2022-03-21
Add the DefaultDcaVoter

Toflar:
This PR introduces the ability to register permission voters which are considered in our DC drivers.
It's foundational work required to finally get rid of our `checkPermission` `onload_callbacks` which are everything but reusable.

Future goals include having permission voters on all `global_operations` as well as `operations` and `buttons` in order to reduce callbacks step by step and make things reusable.


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -128,6 +130,41 @@ abstract class DataContainer extends Backend
...
+	/**
+	 * Move permission
+	 */
+	public const PERMISSION_MOVE = self::PERMISSION_PREFIX . 'move';
+
+	/**
+	 * Delete permission
+	 */
+	public const PERMISSION_DELETE = self::PERMISSION_PREFIX . 'delete';
+
```
aschempp: shouldn't this be part of `ContaoCorePermissions`

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +895,29 @@ protected function switchToEdit($id)
 		return $strUrl . (!empty($arrKeys) ? '&' : '') . (Input::get('table') ? 'table=' . Input::get('table') . '&amp;' : '') . 'act=edit&amp;id=' . rawurlencode($id);
 	}
 
+	/**
+	 * @throws AccessDeniedException
+	 */
+	public function denyAccessIfDisallowed($attributes, $subject): void
```
aschempp: should this really be `public`. Also, in Symfony this is `Controller::denyAccessUnlessGranted(…)`

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +895,29 @@ protected function switchToEdit($id)
...
+	{
+		/** @var Security $security */
+		$security = System::getContainer()->get('security.helper');
+
+		if ($security->isGranted($attributes, $subject))
+		{
+			return;
+		}
+
+		$message = 'Permission denied.';
```
```suggestion
		$message = 'Access denied.';
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +895,29 @@ protected function switchToEdit($id)
...
+
+		if ($subject instanceof DataContainerSubject)
+		{
+			$message = sprintf('Permission denied for %s [%s].', $subject, $attributes);
```
```suggestion
			$message = sprintf('Access denied for %s [%s].', $subject, $attributes);
```

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2476,6 +2498,8 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 		$objSessionBag = System::getContainer()->get('session')->getBag('contao_backend');
 		$session = $objSessionBag->all();
 
+		$this->denyAccessIfDisallowed(self::PERMISSION_VIEW, new DataContainerSubject($this->strTable, $path));
```
aschempp: I would not check permissions recursively for a tree, since child nodes cannot be "invisible". We _might_ add that as a new feature, but I fear it would have a lot of voter overhead.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1638,6 +1656,8 @@ public function editAll()
 			// Walk through each record
 			foreach ($ids as $id)
 			{
+				$this->denyAccessIfDisallowed(self::PERMISSION_EDIT, new DataContainerSubject($this->strTable, $id));
```
aschempp: should catch exception and ignore the `$id`

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1077,6 +1090,8 @@ public function deleteAll()
 
 			foreach ($ids as $id)
 			{
+				// No need for permission check here, it's done in delete()
```
aschempp: should try-catch and ignore undeletable `$id`s

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -982,6 +991,8 @@ protected function copyChilds($table, $insertID, $id, $parentId)
 												->execute($id);
 				}
 
+				$this->denyAccessIfDisallowed(self::PERMISSION_COPY, new DataContainerSubject($v, $objCTable->id));
```
aschempp: shouldn't this be in the while-loop in the line below?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1572,6 +1587,8 @@ public function deleteChilds($table, $id, &$delete)
 	 */
 	public function undo()
 	{
+		$this->denyAccessIfDisallowed(self::PERMISSION_MOVE, new DataContainerSubject($this->strTable, $this->intId));
```
aschempp: is this the correct permission?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2723,6 +2744,8 @@ public function overrideAll()
 			{
 				foreach ($ids as $id)
 				{
+					$this->denyAccessIfDisallowed(self::PERMISSION_EDIT, new DataContainerSubject($this->strTable, $id));
```
aschempp: this should try-catch and not edit unallowed records

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3830,6 +3853,8 @@ public function ajaxTreeView($id, $level)
 	 */
 	protected function generateTree($table, $id, $arrPrevNext, $blnHasSorting, $intMargin=0, $arrClipboard=null, $blnCircularReference=false, $protectedPage=false, $blnNoRecursion=false, $arrFound=array())
 	{
+		$this->denyAccessIfDisallowed(self::PERMISSION_VIEW, new DataContainerSubject($table, $id));
```
aschempp: shouldn't be done in tree, see `DC_Folder`

core-bundle/src/Security/DataContainer/DataContainerSubject.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $subject = [];
+        $subject[] = 'Table: '.$this->getTable();
+
+        if ($id = $this->getId()) {
+            $subject[] = 'ID: '.$id;
+        }
+
+        if ($attributes = $this->getAttributes()) {
+            $subject[] = 'Attributes: '.implode(', ', $attributes);
```
aschempp: this should also output the `$attributes` keys?

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
```diff
@@ -0,0 +1,35 @@
...
+use Contao\DataContainer;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+
+/**
+ * By default, the Contao back end is fully accessible unless a developer wants to have specific
+ * permissions. That's why this voter is implemented with a very low priority, so it allows everything
+ * in the back end as the last voter in case no other voter decided to deny access before.
+ */
+class DefaultDcaVoter extends Voter
```
aschempp: I would probably not implement the `Voter` class to simplify the implementation.

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    protected function supports(string $attribute, $subject)
+    {
+        return str_starts_with($attribute, DataContainer::PERMISSION_PREFIX);
```
aschempp: `str_starts_with ` 🎉

aschempp (changes requested):
Looks very good and simple to me! We need to improve the handling of "multiple" records as discussed, but thats should be pretty simple as well.

We might considering moving the check for `closed`, `notEditable` etc. to voters as well, since it would then be overridable with priority. But that could be a later step.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: I was thinking about this too. Wasn't sure but now done in 535bd13d1fab703af0ab22c2256f7f171f13ec0f.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: No idea why this was public :D Fixed and adjusted naming in 780f188c6041468981ef08158c0d2b87834005c6.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: Adjusted in 8475477b36b3db9591087cc76f69d8c8c9c5f570.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: Adjusted in 8475477b36b3db9591087cc76f69d8c8c9c5f570.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in edc788926a82ee22648165f81a3df11f66ec2b9a.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in edc788926a82ee22648165f81a3df11f66ec2b9a.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in edc788926a82ee22648165f81a3df11f66ec2b9a.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I guess so, and it should also ignore the exception. See 7042bcbf4f5ed3096c9be157e1551a8f5e3de0ea.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Not sure no. Undo is very special anyway because proooobably undoing is related to being able to edit a record. But if any relations are gone (e.g. a PID that's gone), voters are probably lost anyway. We could introduce a special `undo` permission if we need it but for now, I've removed it in e02e61e06737a0876795a538ee6cd1712308b752.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Done in bd590f567edd4c962bdb0933031a64afa28c1168.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Done in bd590f567edd4c962bdb0933031a64afa28c1168.

core-bundle/src/Security/DataContainer/DataContainerSubject.php
Toflar: What if they do not have keys? What if they are objects? Not sure if we can fix this really.

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
Toflar: Then I have to implement the foreach loop to vote on all `$attributes` myself. Why would I?

core-bundle/src/Security/DataContainer/DataContainerSubject.php
aschempp: if they are objects, we can't really implode it either?

Toflar: I know. The question is what to do.

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
aschempp: it would looks simpler to me with less parent overhead 🙃 

```
class DefaultDcaVoter implements VoterInterface
{
    public function vote(TokenInterface $token, $subject, array $attributes)
    {
        foreach ($attributes as $attribute) {
            if (str_starts_with($attribute, DataContainer::PERMISSION_PREFIX)) {
                return self::ACCESS_GRANTED
            }
        }

        return self::ACCESS_ABSTAIN
    }
}
```

core-bundle/src/Security/DataContainer/DataContainerSubject.php
aschempp: maybe dump JSON?

Toflar: Wouldn't work with objects either but maybe we can just ignore that. It's only used as debugging message anyway.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +860,29 @@ protected function switchToEdit($id)
...
+		}
+
+		$message = 'Access denied.';
+
+		if ($subject instanceof DataContainerSubject)
+		{
+			$message = sprintf('Access denied to %s [%s].', $subject, $attributes);
+		}
+
+		throw new AccessDeniedException($message);
```
aschempp: subject and attributes should be added to the exception (according to `Controller::denyAccessUnlessGranted()`

```
$exception->setAttributes($attribute);
$exception->setSubject($subject);
```

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -794,6 +800,8 @@ public function cutAll()
 		{
 			foreach ($arrClipboard[$this->strTable]['id'] as $id)
 			{
+				// No need for permission check here, it's done in cut()
+
```
aschempp: should try-catch and skip if a `cut` operation is denied

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -964,6 +974,8 @@ public function copyAll()
 		{
 			foreach ($arrClipboard[$this->strTable]['id'] as $id)
 			{
+				// No need for permission check here, it's done in copy()
```
aschempp: should try-catch and skip if a `copy` operation is denied

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4376,6 +4411,8 @@ protected function parentView()
 
 			for ($i=0, $c=\count($row); $i<$c; $i++)
 			{
+				$this->denyAccessUnlessGranted(ContaoCorePermissions::DCA_VIEW, new DataContainerSubject($this->strTable, $row[$i]['id']));
```
aschempp, line 4424: this means if I want to limit an element from being visible, I must not use this permission but also apply a `filter` or similar to the DCA, right? Otherwise we should catch the exception and not list the item.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4811,6 +4848,8 @@ protected function listView()
 
 			foreach ($result as $row)
 			{
+				$this->denyAccessUnlessGranted(ContaoCorePermissions::DCA_VIEW, new DataContainerSubject($this->strTable, $row['id']));
```
aschempp, line 4861: this means if I want to limit an element from being visible, I must not use this permission but also apply a `filter` or similar to the DCA, right? Otherwise we should catch the exception and not list the item.

core-bundle/src/Security/ContaoCorePermissions.php
```diff
@@ -186,4 +186,39 @@ final class ContaoCorePermissions
      * Subject must be a corresponding group ID or an array of group IDs.
      */
     public const MEMBER_IN_GROUPS = 'contao_member.groups';
+
+    /**
+     * Prefix for all DCA related permission attributes.
+     */
+    public const DCA_PREFIX = 'contao.dca.';
```
aschempp: I find that prefix constant a little superfluous if only used by the voter and tests. We didn't add one for `contao_member.` either. But I can live with it if you insist 🙃

core-bundle/src/Security/DataContainer/DataContainerSubject.php
Toflar: I will use JSON and mark this as resolved. Otherwise I'm losing track here.

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
Toflar: Done (but better ;)) in dfb2f3b8d6c0fc1bbad763e30ce300fe08659ed8.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: Done in 0a9dffbbc7c237bdbf09a8bc3f2bf22ff961462a.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in 596477a7322a89937cf775075b03c7521be9d120.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in 596477a7322a89937cf775075b03c7521be9d120.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Yes, this is only a security measure here.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Yes, this is only a security measure here.

core-bundle/src/Security/ContaoCorePermissions.php
Toflar: I would prefer having it because I want to use it also when looping over `operations`, `global_operations` and `buttons` which can have an arbitrary key but they should all run in this "namespace".

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +860,33 @@ protected function switchToEdit($id)
 		return $strUrl . (!empty($arrKeys) ? '&' : '') . (Input::get('table') ? 'table=' . Input::get('table') . '&amp;' : '') . 'act=edit&amp;id=' . rawurlencode($id);
 	}
 
+	/**
+	 * @throws AccessDeniedException
+	 */
+	protected function denyAccessUnlessGranted($attributes, $subject): void
```
```suggestion
	protected function throwUnlessGranted($attributes, $subject): void
```

Might make it easier to read the code that calls this method?

core-bundle/src/Security/ContaoCorePermissions.php
ausi:
```suggestion
    public const DCA_PREFIX = 'contao_dca.';
```
Would this be more consistent with the other prefixes `contao_member` and `contao_user`?

core-bundle/src/Security/DataContainer/DataContainerSubject.php
```diff
@@ -0,0 +1,58 @@
...
+    private string $table;
+    private int|string|null $id;
+    private array $attributes;
+
+    public function __construct(string $table, int|string|null $id = null, array $attributes = [])
+    {
+        $this->table = $table;
+        $this->id = $id;
+        $this->attributes = $attributes;
+    }
```
ausi:
```suggestion
    public function __construct(private string $table, private int|string|null $id = null, private array $attributes = [])
    {}
```
😎

ausi: Might even be a use case for `public readonly` 🤩

ausi (dismissed):
❤️❤️❤️!!

Just some minor notes/nitpicks ☺️

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp: `denyAccessUnlessGranted` is the method name Symfony already uses and is widely known and maybe even documented. I would stick with that.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -858,6 +860,33 @@ protected function switchToEdit($id)
...
+		if ($security->isGranted($attributes, $subject))
+		{
+			return;
+		}
+
+		$message = 'Access denied.';
+
+		if ($subject instanceof DataContainerSubject)
+		{
+			$message = sprintf('Access denied to %s [%s].', $subject, $attributes);
```
aschempp: wouldn't this generate `Access denied to XXX [Array()].`?

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1920,6 +1963,8 @@ public function editAll()
 	 */
 	public function source()
 	{
+		$this->denyAccessUnlessGranted(ContaoCorePermissions::DCA_EDIT, new DataContainerSubject($this->strTable, $this->intId));
```
aschempp: I think we should move the permission check after `isValid` just below, because we don't want permission checks on invalid files.

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    public function supportsAttribute(string $attribute): bool
+    {
+        return str_starts_with($attribute, ContaoCorePermissions::DCA_PREFIX);
+    }
+
+    public function supportsType(string $subjectType): bool
+    {
+        return true;
+    }
```
aschempp, line 34: not sure what these methods are for? The first one is still used but should either be private or the line just _integrated_ below?

aschempp (changes requested):
Just some minor nitpicks left 🎉 
I would agree with @ausi about the prefix as well.

core-bundle/src/Resources/contao/classes/DataContainer.php
ausi: `$attributes` is always a string in our cases. Maybe rename to `$attribute` as they did in Symfony? https://github.com/symfony/symfony/commit/91249ed0bfac5609e6bac4e4613b388256aba9cc

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
ausi: They are the implementation of the `CacheableVoterInterface`, no?

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp: aah, I was confusing this with `$attributes` of the `DataContainerSubject` 🙃

core-bundle/src/Security/Voter/DataContainer/DefaultDcaVoter.php
aschempp: aah, didn't notice that 👍

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Adjusted in 9705ea5111e1b53942ecbcb63fee0d49519bd017.

Toflar:
I think all comments have been addressed here. I hope I didn't miss any 😊

ausi (dismissed):

core-bundle/src/Security/DataContainer/DataContainerSubject.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Contao\CoreBundle\Security\DataContainer;
+
+class DataContainerSubject
+{
+    public function __construct(public readonly string $table, public readonly int|string|null $id = null, public readonly array $attributes = [])
+    {
+    }
+
+    public function __toString(): string
```
aschempp, line 24: from the PHP docs:

> As of PHP 8.0.0, any class that contains a [__toString()](dfile:///Users/aschempp/Library/Application%20Support/Dash/DocSets/PHP/PHP.docset/Contents/Resources/Documents/www.php.net/manual/en/language.oop5.magic.html#object.tostring) method will also implicitly implement the [Stringable](dfile:///Users/aschempp/Library/Application%20Support/Dash/DocSets/PHP/PHP.docset/Contents/Resources/Documents/www.php.net/manual/en/class.stringable.html) interface, and will thus pass type checks for that interface. Explicitly implementing the interface anyway is recommended.

aschempp (commented):
One final thing to fix I think.

Regarding the `contao_dca` prefix, I wonder if people might confuse this permission with the DCA _configuration_, whereas it implicitly checks permissions for the _Contao back end_ only. We use the DCA _configuration_ for the `PersonalData` front end modules, but the DCA permissions obviously (to us) do not apply here. Not sure if a prefik like `contao_backend.` would rather match to reflect these are back end (DCA) permissions?

core-bundle/src/Security/DataContainer/DataContainerSubject.php
Toflar: I know, but that's CS to me. We should do it everywhere in one specific commit.

Toflar:
I did name it `contao_backend` intially but then changed it because it's wrong. The back end does not only consist of DCA so I think it's correct the way it is.

aschempp (dismissed):

core-bundle/src/Resources/config/voters.yml
```diff
@@ -0,0 +1,18 @@
+services:
+    _defaults:
+        autoconfigure: true
+
+    contao.security.backend_access_voter:
```
leofeyer: I don‘t like these services to be in a separate file. If at all, they should be in a `security.yml` file because that is the name of the namespace.

Would you mind me moving the definitions back to the `services.yml` file?

Toflar: The number of voters will grow considerably in the future. That's why. I think it will become quite a mess in `services.yml`.

leofeyer: Then we should think of using a dedicated namespace (e.g. `contao.security.voter.`) for the voters, so we can move them to a separate `voters.yml` file in the future.

Toflar: Yeah but then the service name linter will fail. But anyway, you can totally move them to wherever you prefer. I'm just saying that the number will grow significantly and it's probably better to have those in their own file for easier lookup :)

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -793,7 +799,14 @@ public function cutAll()
...
 			{
-				$this->cut($id); // do not urldecode() here (see #6840)
+				try
+				{
+					$this->cut($id); // do not urldecode() here (see #6840)
+				}
+				catch (AccessDeniedException)
+				{
+					// noop
+				}
```
leofeyer, line 809: Why do we catch the exception here?

Toflar: We catch it in all places where multiple operations are happening. If you cut 30 IDs and you have access to 25, then the 25 should work and the other ones should be ignored.

leofeyer (approved):

core-bundle/src/Resources/config/voters.yml
leofeyer: As discussed in Slack, we want to keep the service definitions in the `services.yml` for now and discuss this again when the other voters are added.

leofeyer:
Thank you @Toflar.

PR #4326, opened 2022-03-21
Fix the remaining relative URLs

fritzmg:
This PR implements path absolute URLs for the front end (back end will still use relative URLs). i.e.:

* `PageModel::getFrontendUrl` will now return path absolute URLs, like:
    * `/page-alias.html`,
    * `/contao-subfolder/public/page-alias.html`,
    * `/contao-subfolder/public/index.php/page-alias.html` instead of `page-alias.html`
* Wherever applicable, `$request->getBaseUrl() . $request->getPathInfo()` from the _main_ request will be used when the request without the query string is needed.
* Wherever applicable, `$request->getRequestUri()` is used when the request with the query string is needed.


core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -1068,7 +1068,15 @@ static function ($allowedTag)
...
-							$arrCache[$strTag] = System::urlEncode($objFile->path);
+							$staticUrl = '';
+
+							// Prepend static URL if insert tag is used as URL attribute
+							if (\in_array('urlattr', $flags, true))
+							{
+								$staticUrl = $container->get('contao.assets.files_context')->getStaticUrl();
+							}
+
+							$arrCache[$strTag] = System::urlEncode($staticUrl . $objFile->path);
```
fritzmg: _Note:_ `ContaoContext::getStaticUrl()` already returns either a path absolute or an absolute URL.

_Note:_ This change ensures that things like `<a href="{{file::*}}">` or `` in the front end will produce a correct URL.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -1068,7 +1068,15 @@ static function ($allowedTag)
 
 						if ($objFile !== null)
 						{
-							$arrCache[$strTag] = System::urlEncode($objFile->path);
+							$staticUrl = '';
+
+							// Prepend static URL if insert tag is used as URL attribute
+							if (\in_array('urlattr', $flags, true))
```
ausi: I think we should always add the static URL/base URL, even if no `urlattr` flag is set.

fritzmg:
There is one issue I haven't thought about previously: the `tl_content.url` and `tl_news.url` etc. fields currently might have relative URLs in them. My plan is to migrate these (or rather any DCA field with `rgxp => url`) to path absolute URLs, i.e. prepending a `/` if:

* The URL does not start with an Insert Tag.
* The URL does not start with a scheme.

This also made me think that we should may be introduce [recorded migrations](https://github.com/contao/contao/pull/3108) after all 🙃.

fritzmg:
And speaking of `rgxp => url` fields in the DCA: should Contao 5.0 then automatically prepend `/` for URLs entered in that field? Or should the validation of `rgxp => url` be changed to only allow absolute URLs (or insert tags)?

ausi:
> should Contao 5.0 then automatically prepend `/` for URLs entered in that field?

IMO yes. But it would also have to add the base path and the script name, right?

fritzmg:
> IMO yes. But it would also have to add the base path and the script name, right?

Yes.

However, saving the base path and script name in the database means that you cannot easily switch anymore between `https://localhost/contao5/public/` and `https://contao5.local/`.

ausi:
> However, saving the base path and script name in the database means that you cannot easily switch anymore

So instead of prepending `/` we should prepend something like `{{env::path}}`?

fritzmg:
May be, yes. Or all content elements, modules etc. have to transform URLs from the database themselves from relative to absolute when they output them? 🤔

leofeyer:
As discussed in the Contao call, we want to add the `{{env::path}}` insert tag automatically in TinyMCE for relative URLs.

fritzmg:
This PR has now been updated after #4851 has been merged. It now fixes a few remaining relative URL issues.

_Note:_ to test all changes, #4891 needs to be merged upstream first.

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1048,7 +1048,7 @@ public static function redirect($strLocation, $intStatus=303)
 		$strLocation = str_replace('&amp;', '&', $strLocation);
 
 		// Make the location an absolute URL
-		if (!preg_match('@^https?://@i', $strLocation))
+		if (!preg_match('@^(https?://|/)@i', $strLocation))
 		{
 			if ($strLocation[0] == '/')
```
ausi: This seems to be wrong?
`$strLocation[0] == '/'` can never be true because of the `!preg_match()` now.

ausi (dismissed):

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
```diff
@@ -42,6 +42,6 @@ public function __invoke(RequestEvent $event): void
             return;
         }
 
-        $event->setResponse(new RedirectResponse($request->getSchemeAndHttpHost().$request->getPathInfo().(null !== ($qs = $request->server->get('QUERY_STRING')) ? '?'.$qs : '')));
+        $event->setResponse(new RedirectResponse($request->getSchemeAndHttpHost().$request->getBasePath().$request->getPathInfo().(null !== ($qs = $request->server->get('QUERY_STRING')) ? '?'.$qs : '')));
```
leofeyer, line 45: Please add a unit tests for this so we don‘t run into regressions.

Also, I have searched the code base and found the following variants:

* `$request->getSchemeAndHttpHost().$request->getRequestUri()` (1 time)
* `$request->getSchemeAndHttpHost().$request->getPathInfo()` (2 times)
* `$request->getSchemeAndHttpHost().$request->getBaseUrl().$request->getPathInfo()` (3 times)
* `$request->getSchemeAndHttpHost().$request->getBasePath()` (1 time)

`$request->getSchemeAndHttpHost().$request->getBasePath().$request->getPathInfo()` does not occur in the code base at all, therefore I‘m wondering if it is correct?

leofeyer (changes requested):

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -1166,7 +1166,7 @@ public function getPreviewUrl($strParams=null)
 		$baseUrl = $context->getBaseUrl();
 
 		// Add the preview script
-		$context->setBaseUrl($previewScript);
+		$context->setBaseUrl(\dirname($context->getBaseUrl()) . $previewScript);
```
leofeyer: What issue does this fix?

core-bundle/tests/EventListener/StoreRefererListenerTest.php
```diff
@@ -75,7 +75,7 @@ public function refererStoredOnKernelResponseProvider(): \Generator
             [
                 'newRefererId' => [
                     'last' => '',
-                    'current' => 'hi/I/am/your_current_referer.html',
+                    'current' => '/hi/I/am/your_current_referer.html',
```
leofeyer: This change seems unrelated to the changes above. How is this related?

core-bundle/src/Resources/contao/models/PageModel.php
fritzmg: If you are logged into the back end via `https://example.com/foobar/public/` the preview links in the back end will be wrong otherwise.

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
fritzmg: > therefore I‘m wondering if it is correct?

The change in this particular instance is correct, otherwise you will be redirected to the wrong URL if you open a back end preview URL without being logged in.

I'll check the other instances as well.

core-bundle/tests/EventListener/StoreRefererListenerTest.php
fritzmg: Not directly related - but since #4851 our `StoreRefererListener` stores _absolute_ URLs instead of relative ones. However, the tests were not changed to reflect that.

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
fritzmg: > * $request->getSchemeAndHttpHost().$request->getRequestUri()

This is ok, as `$request->getRequestUri()` already contains the base path.

> * $request->getSchemeAndHttpHost().$request->getBaseUrl().$request->getPathInfo()
> * $request->getSchemeAndHttpHost().$request->getBasePath()

These are also correct.

> * $request->getSchemeAndHttpHost().$request->getPathInfo()

I have fixed the second occurrence as well now.

core-bundle/tests/EventListener/StoreRefererListenerTest.php
fritzmg: But since the tests don't fail I suppose we do not really have to change this. Reverting back.

core-bundle/src/Resources/contao/models/PageModel.php
leofeyer: How do I reproduce the issue?

fritzmg: 1. Log into the Contao back end via `https://localhost/your-contao-install/public/contao`.
2. Go to _Content_ » _Pages_.
3. Click on a page icon to open the preview for that page.
4. You will be redirected to the wrong URL.

| | Redirect URL |
|---|---|
| Without this change | `https://localhost/preview.php/alias-of-page`
| With this change | `https://localhost/your-contao-install/public/preview.php/alias-of-page`

ausi (dismissed):

core-bundle/src/Resources/contao/models/PageModel.php
leofeyer: I cannot reproduce this issue.

1. I open `https://localhost/managed-edition/public/contao`.
2. The page preview link is `https://localhost/managed-edition/public/contao/preview?page=51`.
3. If I click the link, the `BackendPreviewController` redirects me to `https://localhost/managed-edition/public/preview.php/contao/preview?page=52`.
4. On this page in the `PageModel::getPreviewUrl()` method, `$context->getBaseUrl()` is the same as `\dirname($context->getBaseUrl()) . $previewScript`, so the whole `$context->setBaseUrl()` call is redundant.

How exactly do I reproduce your case?

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
leofeyer: How do I reproduce this?

core-bundle/src/Resources/contao/models/PageModel.php
fritzmg: In line 1156 we set `$previewScript`, which will always be `/preview.php` (when using the default config) and in line 1169 we override the base URL with `$previewScript`. So at that point `$context->getBaseUrl()` will be wrong as it will be missing the original base. I am not sure why you are not able to reproduce the issue, this issue is definitely present in `5.x` and I can still reproduce it as described.

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
fritzmg: 1. Log out of the back end.
2. Access a preview URL via localhost + subfolder, e.g. `https://localhost/your-contao-instance/public/preview.php/contao/preview?page=3`

You will be redirect to

```
https://localhost/contao/login?_hash=…&redirect=https%3A%2F%2Flocalhost%2Fyour-contao-instance%2Fpublic%2Fpreview.php%2Fcontao%2Fpreview%3Fpage%3D3
```

instead of

```
https://localhost/your-contao-instance/public/contao/login?_hash=…&redirect=https%3A%2F%2Flocalhost%2Fyour-contao-instance%2Fpublic%2Fpreview.php%2Fcontao%2Fpreview%3Fpage%3D3
```

core-bundle/src/Resources/contao/models/PageModel.php
ausi: I was also able to reproduce this, and this change fixed it.

core-bundle/src/EventListener/BackendPreviewRedirectListener.php
ausi: I was also able to reproduce this, and this change fixed it.

Simpler reproduction was just opening https://localhost/your-contao-instance/public/preview.php/contao (without the need to log out).

ausi (dismissed):

core-bundle/contao/models/PageModel.php
```diff
@@ -1165,7 +1165,7 @@ public function getPreviewUrl($strParams=null)
 		$baseUrl = $context->getBaseUrl();
 
 		// Add the preview script
-		$context->setBaseUrl($previewScript);
+		$context->setBaseUrl(\dirname($context->getBaseUrl()) . $previewScript);
```
ausi:
```suggestion
		$context->setBaseUrl(\dirname($baseUrl) . $previewScript);
```
nitpick ☺️

ausi (dismissed):

core-bundle/contao/classes/Versions.php
```diff
@@ -671,6 +671,7 @@ public static function addToTemplate(BackendTemplate $objTemplate)
 								   ->execute(...$params);
 
 		$security = System::getContainer()->get('security.helper');
+		$request = System::getContainer()->get('request_stack')->getMainRequest();
```
leofeyer: Should this be `getCurrentRequest()` based on #5006?

ausi: Agree.
```suggestion
		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
```

ausi (approved):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4329, opened 2022-03-21
Deprecated Cache lib

Toflar:
Marking `Contao/Cache` as deprecated. It's a very simple in-memory cache which is better implemented in your own classes. Persistence was never supported so there's no real point in maintaining this class.
If you need persistence, use `symfony/cache` instead.


aschempp:
I think we should rewrite the usage of Cache (from https://github.com/contao/contao/pull/4330) in 4.13 to already deprecated it.

Toflar:
Yeah, I didn't bother because 🤷‍♂️

ausi (dismissed):

comments-bundle/src/Resources/contao/dca/tl_comments.php
```diff
@@ -387,16 +386,18 @@ protected function isAllowedToEditComment($intParent, $strSource)
 			return true;
 		}
 
+		static $cache = array();
```
leofeyer, line 389: This is considered bad practice (see https://twitter.com/s_bergmann/status/1182282457011363842). If we go this way, it should be a static class property. However, I am not convinced about the change.

leofeyer, line 389: This is considered bad practice (see https://twitter.com/s_bergmann/status/1182282457011363842). If we go this way, it should be a static class property. However, I am not convinced about the change.

leofeyer (changes requested):

leofeyer (changes requested):
I don‘t see any advantage in removing the class.

aschempp (dismissed):

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4342, opened 2022-03-21
Rewrite the Backend.limitPreviewHeight to vanilla JS

aschempp:
The new combiner is intentional since the script is not used in any of the other backend templates.


core-bundle/src/Resources/contao/templates/backend/be_main.html5
```diff
@@ -39,6 +39,11 @@
       $objCombiner->add('system/themes/'.$this->theme.'/hover.min.js');
       echo $objCombiner->getCombinedFile();
     ?>"></script>
+    <script src="<?php
+      $objCombiner = new Contao\Combiner();
```
Toflar: Why do you need the combiner? What does it do?

core-bundle/src/Resources/contao/templates/backend/be_main.html5
```diff
@@ -117,4 +122,4 @@
   <script>Backend.initScrollOffset()</script>
 
 </body>
-</html>
```
Toflar: You should configure your IDE correctly, this happens in all of your PRs somehow :D

.editorconfig
```diff
@@ -13,7 +13,7 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
-[*.{svg,min.css,min.js}]
+[*.{svg,min.css,min.js,html5}]
```
ausi: Unrelated ☺️

ausi (approved):

core-bundle/src/Resources/public/limit-height.js
```diff
@@ -0,0 +1,47 @@
...
+
+        div.style.height = hgt+'px';
+
+        button.addEventListener('click', function () {
+            if (div.offsetHeight > hgt) {
+                div.style.height = hgt+'px';
+            } else {
+                div.style.height = 'auto';
+            }
+        });
```
leofeyer, line 43: Do we no longer need the `button.set('data-state', button.get('data-state') ? 0 : 1);` line here?

aschempp: I don't think we do, at least it has no technical benefit in this code. I can imagine we used it through CSS back in the days when we had that green down/up arrow to expand and collapse.

leofeyer: Guess so…

leofeyer (approved):

PR #4345, opened 2022-03-22
Rewrite the Backend.autoFocusFirstInputField to vanilla JS

aschempp:


Toflar (dismissed):

core-bundle/src/Resources/public/core.js
```diff
@@ -906,15 +906,6 @@ var Backend =
...
 
-	/**
-	 * Toggle the synchronization results
-	 */
-	toggleUnchanged: function() {
-		$$('#result-list .tl_confirm').each(function(el) {
-			el.toggleClass('hidden');
-		});
-	},
-
```
ausi, line 822: Seems to be unrelated? On purpose?

core-bundle/src/Resources/public/autofocus.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const inputs = edit.querySelectorAll('input, textarea');
+
+    for (let i = 0; i < inputs.length; i++) {
+        const input = inputs[i];
+
+        if (
+            !input.getAttribute('disabled')
+            && !input.getAttribute('readonly')
```
ausi:
```suggestion
            !input.disabled
            && !input.readonly
```

The mootools `get()` method is not synonymous with `getAttribute`.

core-bundle/src/Resources/public/autofocus.js
```diff
@@ -0,0 +1,37 @@
...
+            && !isHidden(input)
```
ausi:
```suggestion
            && input.offsetWidth
            && input.offsetHeight
```

core-bundle/src/Resources/public/core.js
aschempp: nah, I just forgot it in https://github.com/contao/contao/pull/4327 and this was the next PR related to JS 😆

core-bundle/src/Resources/public/autofocus.js
aschempp: That would not check for the `display: none` property (according to the Mootols code)?

core-bundle/src/Resources/public/autofocus.js
aschempp: According to the Mootols code, it actually is pretty synonymous. Even though `getAttribute` and DOM properties are _almost always_ synonymous (they are in this case), https://javascript.info/dom-attributes-and-properties suggests to use the DOM property since it would probably also be boolean in this case. I'll adjust all my PRs 👍

ausi: Side note: using `getAttribute()` for boolean attributes gives you the empty string for the `true` value, need to use `hasAttribute()` instead then to get a boolean.
But 👍 for using DOM properties instead.

core-bundle/src/Resources/public/autofocus.js
ausi: Elements with `display: none` have `offsetWidth` and `offsetHeight` of `0`.

core-bundle/src/Resources/public/autofocus.js
aschempp: Changed in b4b4b733ab915ced966e20ea3f073d28bc27d52f

core-bundle/src/Resources/public/autofocus.js
aschempp: Changed in b4b4b733ab915ced966e20ea3f073d28bc27d52f

leofeyer (approved):

PR #4346, opened 2022-03-22
Drop useless framework initialization

m-vo:
/cc @ausi

### DIFF
--- a/core-bundle/src/Controller/ContentElement/TemplateController.php
+++ b/core-bundle/src/Controller/ContentElement/TemplateController.php
@@ -22,8 +22,6 @@ class TemplateController extends AbstractContentElementController
 {
     protected function getResponse(Template $template, ContentModel $model, Request $request): ?Response
     {
-        $this->initializeContaoFramework();
-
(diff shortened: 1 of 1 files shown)


bytehead (approved):

aschempp (approved):

leofeyer:
Are you sure that the initialization is useless? The method gets a `Contao\Template` object which likely does not work without the Contao 3 framework.

fritzmg:
I think the idea is that the `getResponse` method gets object instances of legacy classes and thus the framework must be initialized already beforehand.

ausi (approved):

leofeyer:
Sounds like a "don‘t rely on your dependencies‘ dependencies" situation. 🤷‍♂️

fritzmg:
But then we would also need to inject and initialize the ContaoFramework in every Hook that uses any legacy class - which we currently do not do.

m-vo:
I don't think a method is responsible for initializing the arguments, the caller is. But up to you. In Contao 5 things will be a bit different anyway.

leofeyer:
Thank you @m-vo.

PR #4348, opened 2022-03-22
Fix an execute statement

ausi:
We should probably search for all usages of `$statement->executeStatement(…)`


leofeyer:
Thank you @ausi.

leofeyer (approved):

PR #4353, opened 2022-03-22
Ensure the decorated access decision manager shows up in profiler

Toflar:
Symfony decorates its own access decision manager in the `security-bundle`. See https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/SecurityBundle/Resources/config/security_debug.php#L20.

However, in the ME, our `core-bundle` is loaded after the `security-bundle` and thus we decorate the already decorated service. We need to duplicate the logic in order to bring back the AD logs in the profiler.

### DIFF
--- a/core-bundle/src/Resources/config/services_debug.yml
+++ b/core-bundle/src/Resources/config/services_debug.yml
@@ -1,0 +1,5 @@
+services:
+    contao.debug.security.access.decision_manager:
+        class: Symfony\Component\Security\Core\Authorization\TraceableAccessDecisionManager
+        arguments:
+            - '@.inner'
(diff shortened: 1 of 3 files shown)


Toflar:
Failing test unrelated.

bytehead (dismissed):
🥳

core-bundle/src/Resources/config/services_debug.yml
```diff
@@ -0,0 +1,5 @@
+services:
+    contao.debug.security.access.decision_manager:
+        class: Symfony\Component\Security\Core\Authorization\TraceableAccessDecisionManager
+        arguments:
```
aschempp, line 5: doesn't this miss the service to decorate?

aschempp (changes requested):

core-bundle/src/Resources/config/services_debug.yml
Toflar: Lol, of course! I tested, then noticed that I forgot the `decorates` info, added this but then forgot to commit that :D Fixed now, thanks!

aschempp (approved):

PR #4355, opened 2022-03-22
Add the SQL import to the backend

ausi:


core-bundle/src/Resources/contao/classes/Theme.php
```diff
@@ -282,6 +299,35 @@ protected function compareThemeFiles($arrFiles, $arrDbFields)
...
+			}
+
+			$return .= '</select>
+  <p class="tl_help tl_tip" title="">' . ($GLOBALS['TL_LANG']['tl_theme']['selectExampleWebsite'][1] ?? '') . '</p>
+</div><div class="w50 widget"' . ($count ? ' style="display: none"' : '') . ' id="ctrl_example_website_import">
+  <h3><label>' . ($GLOBALS['TL_LANG']['tl_theme']['exampleWebsiteImportType'][0] ?? '') . '</label></h3>
+  <select name="example_website_import" class="tl_select">
+    <option value="full">Full SQL import</option>
+    <option value="data">Data only</option>
+    <option value="data_no_truncate">Data only without truncating tables</option>
```
aschempp: missing translations

core-bundle/src/Resources/contao/classes/Theme.php
```diff
@@ -663,9 +723,56 @@ protected function extractThemeFiles($arrFiles, $arrDbFields)
...
+					$connection->executeStatement('TRUNCATE TABLE ' . $connection->quoteIdentifier($table));
+				}
+			}
+		}
+
+		$data = file(Path::join(System::getContainer()->getParameter('kernel.project_dir'), $exampleWebsite));
+
+		try
+		{
+			foreach ($insertOnly ? preg_grep('/^INSERT /', $data) : $data as $query)
```
aschempp, line 756: I think this would fail if there are no `INSERT` statements

core-bundle/src/Resources/contao/classes/Theme.php
```diff
@@ -663,9 +723,56 @@ protected function extractThemeFiles($arrFiles, $arrDbFields)
...
+			{
+				$connection->executeStatement($query);
+			}
+		}
+		finally
+		{
+			// Restore backend user
+			if ((int) $connection->fetchOne('SELECT COUNT(*) FROM tl_user') < 1)
```
aschempp, line 764: shouldn't we check for admins as well? Also, does it make sense to restore users if their groups were replaced (or truncated)?

core-bundle/src/Resources/contao/classes/Theme.php
ausi: Cannot confirm this. How can I reproduce that?

core-bundle/src/Resources/contao/classes/Theme.php
ausi: We should restore the current user so that he does not get logged out.

core-bundle/src/Resources/contao/classes/Theme.php
ausi: Fixed in f037b11056c2fd1cb82ba84fd91cc32eb61c3b20

core-bundle/src/Resources/contao/classes/Theme.php
m-vo: IMO this is pretty important for UX. Otherwise you would need to create a user to enter the back end, import the file and then, again, go back and create a user again. I don't think people would like this.

core-bundle/src/Resources/contao/classes/Theme.php
aschempp: ah I was confusing the method. But would that mean we do not support multiline SQL statements? Did we also not support that previously?

core-bundle/src/Resources/contao/classes/Theme.php
aschempp: yeah maybe SQL theme import should only be available to admins (maybe it already is)? and we should restore that admin user, then we don't need to know/verify about user groups.

core-bundle/src/Resources/contao/classes/Theme.php
ausi: Correct, multiline is not supported, same as it was in the install tool.

core-bundle/src/Resources/contao/classes/Theme.php
ausi: Users with permission to import themes are “implicit admins” 
Bildschirmfoto 2022-04-05 um 09 26 15

leofeyer (approved):

core-bundle/src/Resources/contao/classes/Theme.php
leofeyer: IMHO, the current logic is correct. If the SQL dump includes back end users, you will be logged out after the import and have to log in again with the user defined in the theme. Otherwise, you can keep working with the user you have already created.

leofeyer:
Thank you @ausi.

PR #4357, opened 2022-03-22
Fix srcToInsertTag() for URL encoded paths

fritzmg:
If you insert an image in a text element that contains spaces or other url encoded characters in its path, then `StringUtil::srcToInsertTag` will not convert the file path to a `{{file::*}}` insert tag because `FilesModel::findByPath` will not find the image in its url encoded format (TinyMCE stores the URL url encoded).


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4362, opened 2022-03-22
Use gulp-uglify-es

aschempp:


fritzmg:
Try

```javascript
uglify({comments: false})
```

to get the comments removed. Not sure if that's the right option though.

fritzmg (approved):

PR #4364, opened 2022-03-22
Remove the legacy session access

Toflar:
* The `{{post::*}}` insert tag has been removed. Accessing POST data is no longer possible with insert tags.
To access submitted form data on forward pages, use the new `{{form_session_data::*}}` insert tag instead.

* The use of `$_SESSION` is discouraged because it makes testing and configuring alternative storage
back ends hard. In Contao 4, access to `$_SESSION` has been transparently mapped to the Symfony Session.
This has been removed. Use `$request->getSession()` directly instead.

The core is now 100% free of `$_SESSION` and our `LazySessionAccess` monster is gone 🥳 

I did introduce a general concept to have "auto expiring" session attributes which we needed for the form data. The `ClearSessionDataListener` takes care of all of them so storing a key for a given TTL is now as easy as reuising the components provided.


core-bundle/src/Session/Attribute/AutoExpiringAttribute.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Contao\CoreBundle\Session\Attribute;
+
+class AutoExpiringAttribute
+{
+    private int $tstamp;
+    private int $ttl;
+    private mixed $value;
+
+    public function __construct(int $ttl, mixed $value, \DateTimeInterface|null $createdAt = null)
```
fritzmg:
```suggestion

    /**
     * @param int $ttl Time to live in seconds
     */
    public function __construct(int $ttl, mixed $value, \DateTimeInterface|null $createdAt = null)
```

fritzmg (dismissed):
Tested and für gut befunden 👍

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4367, opened 2022-03-22
Remove the orderField support in the picker widgets

Toflar:


aschempp (dismissed):

leofeyer (changes requested):
We cannot remove this yet, because the `orderField` is still needed for the image gallery. However, @ausi already has an idea how to solve this. 🦊

ausi:
> We cannot remove this yet, because the `orderField` is still needed for the image gallery.

It seems that the file tree order field is still here: https://github.com/contao/contao/blob/23b6373c8a52b576911ef4acc53de209f7ef7fe2/core-bundle/src/Resources/contao/widgets/FileTree.php#L74

So this pull request should be fine as it only removes the order field for the pageTree and the Picker, not the fileTree.

Toflar:
Yeah, I only removed the deprecated stuff.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4368, opened 2022-03-22
Remove all sorts of deprecated stuff in the Database namespace

Toflar:


core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -228,16 +220,7 @@ public function limit($intRows, $intOffset=0)
...
-
-		if (\count($arrParams) === 1 && \is_array($arrParams[0]))
-		{
-			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s()" with an array parameter has been deprecated and will no longer work in Contao 5.0. Use argument unpacking via ... instead."', __METHOD__);
-
-			$arrParams = array_values($arrParams[0]);
-		}
-
-		return $this->query('', array_merge($this->arrSetParams, $arrParams));
+		return $this->query('', array_merge($this->arrSetParams, \func_get_args()));
```
aschempp, line 228: we could use `...$args` in the method name and here?

Toflar: We could but I'm removing deprecated stuff here, I'm not using new language constructs.

aschempp (dismissed):

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -141,13 +140,6 @@ public function prepare($strQuery)
 	 */
 	public function set($arrParams)
 	{
-		if (substr_count($this->strQuery, '%s') !== 1 || !\in_array(strtoupper(substr($this->strQuery, 0, 6)), array('INSERT', 'UPDATE'), true))
-		{
-			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s()" is only supported for INSERT and UPDATE queries with the "%%s" placeholder. This will throw an exception in Contao 5.0.', __METHOD__);
```
leofeyer, line 146: Where is the exception thrown?

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -228,16 +220,7 @@ public function limit($intRows, $intOffset=0)
 	 */
 	public function execute()
 	{
-		$arrParams = \func_get_args();
-
-		if (\count($arrParams) === 1 && \is_array($arrParams[0]))
-		{
-			trigger_deprecation('contao/core-bundle', '4.13', 'Using "%s()" with an array parameter has been deprecated and will no longer work in Contao 5.0. Use argument unpacking via ... instead."', __METHOD__);
```
leofeyer, line 235: Did you check that we don‘t accidentally do this?

Toflar: I expect us not to if it's deprecated.

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
Toflar: I guess Doctrine will? @ausi ?

ausi: The `trigger_deprecation()` has to be replaced with a `throw` statement here.

core-bundle/tests/Contao/Database/StatementTest.php
```diff
@@ -1,333 +0,0 @@
...
-use Contao\System;
-use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\Mysqli\Exception\ConnectionError;
-use Doctrine\DBAL\Exception\DriverException;
-use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Result;
-use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
-use Symfony\Component\DependencyInjection\Container;
-
-class StatementTest extends TestCase
```
ausi: This should not be removed here IMO, there are tests for the correct behavior in here too (not only tests for deprecations).

core-bundle/tests/Functional/SearchTest.php
```diff
@@ -1,101 +0,0 @@
...
-    {
-        unset($GLOBALS['TL_LANGUAGE']);
-
-        parent::tearDown();
-    }
-
-    /**
-     * @group legacy
-     */
-    public function testSearchFor(): void
```
ausi, line 45: Should be kept and rewired to `Search::query()`

core-bundle/tests/Contao/Database/StatementTest.php
Toflar: Which ones? They all contain `$this->expectDeprecation()`.

core-bundle/tests/Functional/SearchTest.php
Toflar: Why are these tests so strange? We should be able to delete all methods of deprecated stuff and new ones should be tested separately, no?

core-bundle/tests/Contao/Database/StatementTest.php
```diff
@@ -1,333 +0,0 @@
...
-     */
-    public function testSetTriggersDeprecation(string $query): void
-    {
-        $statement = new Statement($this->createMock(Connection::class));
-
-        if ($query) {
-            $statement->prepare($query);
-        }
-
-        $this->expectDeprecation('%sUsing "%s()" is only supported for INSERT and UPDATE queries with the "%%s" placeholder%s');
```
ausi, line 49: Should be replaced with expecting an exception.

core-bundle/tests/Contao/Database/StatementTest.php
```diff
@@ -1,333 +0,0 @@
...
-        $statement = (new Statement($connection));
-        $statement->prepare($query);
-
-        if ($set) {
-            $statement->set($set);
-        }
-
-        $this->expectDeprecation('%sUsing "%s" with an array parameter has been deprecated%s');
-
-        $statement->execute($params ?? []);
```
ausi, line 211: Everything except these last lines from this test should stay IMO.

ausi (changes requested):

core-bundle/tests/Functional/SearchTest.php
ausi: Probably “should” yes. But AFAIK the legacy code is not 100% covered by tests.
This class is the only search test we have AFAIK, we should not remove it.

Toflar:
Sorry, I won't be able to finish this PR because tests are not clearly separated and I don't know how they have to be adjusted.

ausi:
Should I take over this one?

Toflar:
Sure, you can either PR against my PR or whatever is easier for you.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4369, opened 2022-03-22
Remove deprecated StringUtil stuff

Toflar:


fritzmg (dismissed):

aschempp (dismissed):

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -69,19 +69,6 @@ public static function substr($strString, $intNumberOfChars, $strEllipsis=' …'
...
-			$strEllipsis = '';
-		}
-
-		// Deprecated since Contao 4.0, to be removed in Contao 5.0
-		if ($strEllipsis === true)
-		{
-			trigger_deprecation('contao/core-bundle', '4.0', 'Passing "true" as third argument to "Contao\StringUtil::substr()" has been deprecated and will no longer work in Contao 5.0. Pass the ellipsis string instead.');
-			$strEllipsis = ' …';
-		}
-
```
leofeyer, line 84: Simply removing this without adding a type hint that enforces the type will not work IMHO. Maybe we should throw an `InvalidArgumentException` if the argument `!is_string()`?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -785,13 +741,6 @@ public static function resolveFlaggedUrl(&$url)
 	 */
 	public static function convertEncoding($str, $to, $from=null)
 	{
-		if ($str !== null && !\is_scalar($str) && !(\is_object($str) && method_exists($str, '__toString')))
-		{
-			trigger_deprecation('contao/core-bundle', '4.9', 'Passing a non-stringable argument to StringUtil::convertEncoding() has been deprecated an will no longer work in Contao 5.0.');
-
-			return '';
-		}
-
```
leofeyer, line 794: Again, we should either add a type hint or throw an exception to enforce the type.

Toflar:
I've added typehints at the places requested by @leofeyer. I think we should add typehints in the whole `StringUtil` class but this sounds like a task that is not related to this PR and better done with some tooling.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -777,21 +733,14 @@ public static function resolveFlaggedUrl(&$url)
...
 	 *
-	 * @param string $str  The input string
+	 * @param mixed  $str  The input string
 	 * @param string $to   The target character set
 	 * @param string $from An optional source character set
 	 *
 	 * @return string The converted string
 	 */
-	public static function convertEncoding($str, $to, $from=null)
+	public static function convertEncoding(mixed $str, string $to, string|null $from=null): string
```
leofeyer:
```suggestion
	public static function convertEncoding(string $str, string $to, string|null $from=null): string
```

`mixed` cannot be right if we want a stringable argument, can it?

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
Toflar: No we have tests that explicitly test this. So apparently it's a supported use case and the phpdoc was just wrong.

leofeyer: The deprecation message says _Passing a non-stringable argument to StringUtil::convertEncoding() has been deprecated an will no longer work in Contao 5.0._ @ausi /cc

ausi:
```suggestion
	public static function convertEncoding(bool|float|int|string|\Stringable|null $str, string $to, string|null $from=null): string
```

IMO we only want to allow stringable objects, scalars and null.

leofeyer: @Toflar https://github.com/Toflar/contao/pull/12

leofeyer (approved):

ausi (approved):

leofeyer:
Thank you @Toflar.

Issue #4147, opened 2022-02-17
Improve the "copy to clipboard" feature


leofeyer:
### Description

As discussed in #4133, we need a visual confirmation if someone copies a preview link to their clipboard. The confirmation must not be permanent and must not require to be closed manually. It could be either a flash message or an inline confirmation.

![](https://user-images.githubusercontent.com/1192057/154438692-4085e6e2-bc65-4dd1-ab81-4217cf36162a.png)

While we are on this, we might also merge the `clipboard.js` script into the `core.js` script to save one HTTP request.

aschempp:
> While we are on this, we might also merge the clipboard.js script into the core.js script to save one HTTP request.

I think we should split the features into separate files, and saving the request is a no-brainer with HTTP/2. Also, we don't need to load the script on 99% of the pages, only if a user has access to and accesses the preview link backend. 🙃

aschempp:
I'm happy to implement this using https://www.stimulus-components.com/docs/stimulus-clipboard once we have #5406 😎

m-vo:
I already implemented this as a stimulus controller + some CSS magic for my file manager prototype. Can contribute this 👍

m-vo:
[copytoclipboard.webm](https://user-images.githubusercontent.com/5305677/202914055-6e16a282-3878-4b37-95c1-6223a8ded1e6.webm)

When clicking three things happen:
1) the clipboard content is set
2) a CSS class is added to the element, which makes the icon change and plays a single shot animation (here: the green flash); if the class is already there, it is removed first, so that the animation is played again
3) all other instances are notified to remove the class

Pseudocode:
```twig
<i class="clipboard" {{ stimulus_controller('contao/clipboard-copy', [], {'copied': 'clipboard--copied'}) }}>
    the-thing-that-is-copied
</i>
```

```css
.clipboard {
    &:before {
        background-image: url("…");
    }

    &--copied {
        animation: flash 2 alternate 500ms;

        &:before {
            background-image: url("…");
        }
    }

    @keyframes flash {
        0% {}
        80% {background-color: green;}
        100% {background-color: green;}
    }
}
```


Wdyt?

leofeyer:
I like it. 👍

Issue #3708, opened 2021-11-19
Don't reset CSS for ul, ol and li in BE


Aybee:
**Affected version(s)**

Contao 4.9.22 or Contao 4.123

**Description**
If you reset margin, padding and list-style for ul, ol and li, we have no nice rendering of lists in several places in BE, e.g. within the help wizzard text. This values should never be reset. Only reset them where needed, e.g. within navigation lists.

[](https://github.com/contao/contao/blob/2cabec850046d20fd005e11501f581d980121323/core-bundle/src/Resources/contao/themes/flexible/basic.css#L19-L28)](https://github.com/contao/contao/blob/2cabec850046d20fd005e11501f581d980121323/core-bundle/src/Resources/contao/themes/flexible/basic.css#L19-L28)

BTW there's too much resetting anyway. E.g. for margins it should be enough to reset it for body and figure
```css
body,
figure {
  margin: 0;
}
```
For Padding you don't have to reset this at all.

aschempp:
Maybe this can be considered in https://github.com/contao/contao/pull/2697? I generally agree with you on the topic, but I don't think we can easily change this in the current backend, as people also rely on this reset.

Aybee:
> ... as people also rely on this reset.

Really, I just can not believe that. The browser style for paragraphs, lists, form elements, tables, buttons, headings ... is always better than none padding and/or margin at all. If you use an element for special purposes e.g. a list for a navigation, you style it anyway.

#2697 is too complicated for me. But I will follow it.

fritzmg:
> Really, I just can not believe that.

What @aschempp probably means is backwards compatibility. This cannot be changed in a feature version. If you did, an extension that styled its back end components according to the current state will suddenly look different when updating Contao.

Issue #3388, opened 2017-08-14
Standardise the edit icons


agoat:
We should standardise the use of all edit icons. Currently they are used in different case.

For example the use of the pencil icon in the **site structure** and **articles**:
![site](https://user-images.githubusercontent.com/1402434/29267800-985efece-80ea-11e7-84e4-881d9fc1ec12.jpg)
![articles](https://user-images.githubusercontent.com/1402434/29267802-9adb8f32-80ea-11e7-9ed6-d5787d802fa9.jpg)

Sometimes the pencil is for **editing the element itself** and sometimes for **editing child elements**. That's confusing. 

We should add a icon for editing child elements or consistently use the _header_ icon for editing elements itself.

I would like to see a new icon for editing child elements, despite I don't know how it should look like.

leofeyer:
> I would like to see a new icon for editing child elements, despite I don't know how it should look like.

I fully agree, however, I have no idea how such an icon should look. It is essential that we find one, though, otherwise I cannot continue to work on contao/contao#5391 and contao/core#4371. So @contao/developers and everyone else: how should an "edit child records" icon look like?

<img width="255" alt="" src="https://user-images.githubusercontent.com/1192057/170532320-59e6d40c-f855-48c6-8d78-9a5e4c83ec06.png">

asaage:
I'd probably use the [parent.svg](https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/themes/flexible/icons/parent.svg) icon and change the colors to grey (top/parent) and yellow (for editing/child-items)

aschempp:
I don't like to think about how an icon should look like, but look at it from UX perspective. If you look at a page tree to edit records, you use the edit icon. If I look at an article tree, 95% of the time I want to edit _the article content_ and therefore click the edit icon. It's not only a matter of icon but also of position. Moving the _edit child records_ operation to the second position would certainly be worse UX.

So _if_ we really have to change this, maybe we should use the parent.svg as the first one, and move the pencil to the second operation?

ausi:
Something in this direction maybe?
![Bildschirmfoto 2022-05-27 um 10 41 00](https://user-images.githubusercontent.com/367169/170664095-63144844-232b-48d4-bd9f-c861ca20f014.png)

fritzmg:
[...]

Issue #2218, opened 2020-09-01
user icons: mouseover + better lock icon


math-GH:
In Contao 4.9 the backen user icons looks like
![grafik](https://user-images.githubusercontent.com/1645099/91906382-62889300-eca8-11ea-9dd2-68a016c1b403.png)

I am honest: I am not really lucky with them in some little points:
1) there is a little orange lock on the corner if 2FA is used
![grafik](https://user-images.githubusercontent.com/1645099/91906514-982d7c00-eca8-11ea-845d-7ffd429ba9b8.png)
but in the list view the icon is so little, that it is difficult to see the lock: orange on red is very hard to see (also for me who is not red-green-"blind") and the edge of lock is mostly the same like the edge of the icon.

2) there is no mouseover text hint, that would describe the icon (red = administrator, blue = non-administrator, with lock = 2FA activated). It would help me to understand the icons and get a quick overview of my backend users.

What do you think about it?

aschempp:
what would be your proposed solution?

math-GH:
I won't mix up the user icon (user or admin) with the type of login (PW or PW+2FA). 
For the type of login I would use a second icon:
The second icon could look like:
no 2FA required or no 2FA set up: no icon or 3 points in a row (like a password input)
2FA required, but not set up: open lock or just the borders of a lock or a grey lock
2FA set up: closed lock

aschempp:
I like the idea of having two icons. I'm not sure we need a separate one for _2FA enforced but not enabled_. We could simply use a grey or colored lock. @contao/developers wdyt?

math-GH:
any news?

Toflar:
It has the `help wanted` label so it requires someone willing to take on that issue and work on it :)

fritzmg:
In Contao 5.5+ the icons look like this now:

<img src="https://github.com/user-attachments/assets/566356b9-a9d6-4d10-b67d-95f3c5dc8e0a" width="228">

\
@math-GH does the issue still exist for you there?

PR #4371, opened 2022-03-22
Tune the back end theme

leofeyer:
Apparently a relevant number of users prefers the back end without the blue sidebar. 😄


leofeyer:
We should probably improve the contrast of the menu section headlines and their icons.

leofeyer:
And we should also add a dark theme:



However, it is going to be **a lot of work** to adjust all the icons and every detail…

leofeyer:
And of course we should use CSS custom properties, so our extension developers can adjust their styles accordingly.

```css
:root {
  --bg: white;
  --text-color: #555;
  --link-color: #639A67;
  --link-hover: #205D67;
}

.dark-theme {
  --bg: #222;
  --text-color: white;
  --link-color: #B793E6;
  --link-hover: #3532A7;
}
```

fkaminski:
> We should probably improve the contrast of the menu section headlines and their icons.
... However, it is going to be a lot of work to adjust all the icons

If it stays with this implementation (white, black/grey), it would probably make sense if then the icons are colorless too. 
At least easier in terms of contrast for dark/light mode with corresponding only one custom property.

```
:root {
  --clr-icon: #333;
 }
 
.dark-theme {
  --clr-icon: #eee;
}

<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16">
  <path fill="var(--clr-icon)" d="M11.347 2.056L8.341 5.061a.49.49 0 01-.689 0l-3-2.984a.49.49 0 00-.689 0L2.056 3.985a.486.486 0 00.002.687l3.009 2.974c.19.189.191.499.001.688l-3.02 3.019a.49.49 0 000 .689l1.895 1.895a.49.49 0 00.689 0l3.02-3.02a.49.49 0 01.689 0l3.028 3.027c.188.189.5.189.688 0l1.896-1.895a.491.491 0 000-.689l-3.028-3.027a.488.488 0 010-.689l3.005-3.006a.49.49 0 000-.689l-1.895-1.895a.488.488 0 00-.688.002z"/>
</svg>
```
You could still set matching color values over several properties this way. Regardless of the actual color theme, it might still be beneficial to define separate custom properties for each icon. You would then be more flexible in the future.


```
:root {
  --clr-primary: red;

  --clr-icon-delete: var(--color-primary);
  /* or individual
  --clr-icon-delete: pink;
  */ 
  --clr-icon-edit: var(--color-primary);
 }
 
.dark-theme {
  --clr-primary: green;

  --clr-icon-delete: var(--color-primary);
    /* or individual
  --clr-icon-delete: lightblue;
  */ 
  --clr-icon-edit: var(--color-primary);
}

<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16">
  <path fill="var(--clr-icon-delete)" d="M11.347 2.056L8.341 5.061a.49.49 0 01-.689 0l-3-2.984a.49.49 0 00-.689 0L2.056 3.985a.486.486 0 00.002.687l3.009 2.974c.19.189.191.499.001.688l-3.02 3.019a.49.49 0 000 .689l1.895 1.895a.49.49 0 00.689 0l3.02-3.02a.49.49 0 01.689 0l3.028 3.027c.188.189.5.189.688 0l1.896-1.895a.491.491 0 000-.689l-3.028-3.027a.488.488 0 010-.689l3.005-3.006a.49.49 0 000-.689l-1.895-1.895a.488.488 0 00-.688.002z"/>
</svg>
```

aschempp:
I don't like the idea of colorless icons, just because they are currently also easily distinguishable through their color. Maybe that could only be applied to the "grey" versions that conflict with dark mode.

fkaminski:
> I don't like the idea of colorless icons, just because they are currently also easily distinguishable through their color. Maybe that could only be applied to the "grey" versions that conflict with dark mode.

You are not bound to it but only one custom property would be needed. Examples with color information and custom properties per icon see above.

The suggestion is only, regardless of the icon colors, not to set them manually within the svg definition but also to assign custom properties and thus control them via css.

leofeyer:
Related issues we could work: #3044

m-vo:
> You are not bound to it but only one custom property would be needed. Examples with color information and custom properties per icon see above.

For svgs it is IMHO best practice to use `currentColor` for fill or stroke. This will inherit the color of the parent HTML element that was set via CSS. :relaxed: 

Here are some more PRs & Issues that might be related:
- https://github.com/contao/contao/pull/2053
- https://github.com/contao/contao/issues/3388
- https://github.com/contao/contao/issues/3708
- https://github.com/contao/contao/issues/4147
- https://github.com/contao/contao/issues/2218

fkaminski:
> For svgs it is IMHO best practice to use `currentColor` for fill or stroke. This will inherit the color of the parent HTML element that was set via CSS.

As long as all svg's have only one color. 🤔

leofeyer:
This will be a two-step process. Step 1 is to remove the blue sidebar and the orange top bar, so all areas are bright:



The next step – most likely in a separate PR for Contao 5.1 or later – will be to add a dark mode and adjust all icons accordingly. This will be the hardest part.

fritzmg (dismissed):
Approving as I did not find any obvious issues while testing.

However I still strongly prefer this:

Screenshot 2022-07-20 at 08-38-09 News News Items c50 local

over this:

Screenshot 2022-07-20 at 08-37-24 News News Items c50 local

leofeyer:
Using a solid border for the header makes it look better in Chrome and Firefox.

leofeyer:
@fritzmg We really don‘t need to discuss this. Other CMS behave exactly the same way, so if you don‘t want to have a wide back end, simply resize your browser window accordingly. 🤷‍♂️ 

TYPO3



Wordpress

fritzmg:
> We really don‘t need to discuss this

We already did once and re-introduced it. I do not understand why you want to remove the max-width again. What is the benefit within the current layout of the Contao back end?

> so if you don‘t want to have a wide back end, simply resize your browser window accordingly. 

I am not gonna constantly resize my browser window just because there is one application behaving out of line. That's the _opposite_ of responsive web design.

> Other CMS behave exactly the same way

That's a moot point in my opinion. You cannot make this decision solely based on "others do it too". You should base it on usability of _your_ user interface alone. Also Typo3 for example utilises the additional space much more in various locations than the current Contao back end can. And if you still want to draw comparisons to other web applications:

* https://www.apple.com has a max-width between 1024px and 1560px for its single column content and a max-width of 2560px for its double column content.
* https://store.google.com has a max-width of 1440px.
* https://www.microsoft.com has a max-width of 1600px.
* https://www.amazon.de has a max-width of 1500px for its product lists and product details and 1000px for the account back end.
* https://www.aramco.com has a max-width of 1440px for its content.

fkaminski:
Leo's statements in this regard are not wrong to begin with.
However, I also find Fritz's statement justified.

Regarding "usability" I would find a future solution helpful which then optionally (if space is available) provides for example an additional column with context-related properties for editing. 
JM2C

bennyborn:
To be honest... I don't like it at all. Not enough contrast and way too bright - hurts my eyes. If a "dark mode" will only be available with 5.1 so should this change be 🤷‍♂️

PR #4375, opened 2022-03-23
Rewrite fieldset toggling to vanilla JS

aschempp:


Toflar (dismissed):
Perfetto 😊

core-bundle/src/Resources/public/toggle-fieldset.js
```diff
@@ -0,0 +1,68 @@
...
+        el.addEventListener('click', function (event) {
+            event.preventDefault();
+            toggleState(el, id, table);
+        })
+    });
+
+    AjaxRequest.toggleFieldset = function (el, id, table) {
+        window.console && console.warn('Using AjaxRequest.toggleFieldset is deprecated and will be removed in Contao 6. Add the data-toggle-fieldset attribute instead.');
+        toggleState(el, id, table)
+    };
```
leofeyer: Do we really need a BC layer here?

aschempp: I think so. Anyone who copied stuff from DC_Table will include this line and would otherwise break. Since we did not and cannot deprecate it in Contao 4, and since its so simple, I would keep this atm.

leofeyer: I would prefer to remove it and add it again in case it causes problems. I think this is also what have agreed on when it comes to removing BC layers.

aschempp: Sorry but I disagree. The BC layer is already written, and it does not harm. There was currently no deprecation to use the method, so we must not remove it. As I said, I _know_ this will cause problems because I did write multiple custom DCs 😉 
https://github.com/isotope/core/blob/2.7/system/modules/isotope/drivers/DC_ProductData.php#L516
https://github.com/terminal42/contao-DC_Multilingual/blob/main/src/Driver.php#L250

leofeyer: Ok. Please resolve the conflicts then, so I can merge the PR.

leofeyer (approved):

PR #4379, opened 2022-03-23
Remove the textarea toggleWrap feature

aschempp:
Not sure we need the BC method, since someone copying the code to their own DC would probably also have the existing class (and therefore it would already work automatically) 🙃


core-bundle/src/Resources/public/toggle-wrap.js
```diff
@@ -0,0 +1,56 @@
...
+                })
+            }
+        }
+    }).observe(document, {
+        attributes: false,
+        childList: true,
+        subtree: true
+    });
+
+    Backend.toggleWrap = function (id) {
```
ausi: Should we add some checks here?

```suggestion
    if (window.Backend && !Backend.toggleWrap) {
    Backend.toggleWrap = function (id) {
```

aschempp: I would not for now. Once we can switch to Encore it all changes anyway.

leofeyer:
I don‘t think we need the BC method. Honestly, I don‘t even think we need this feature at all. 😄

leofeyer:
@contao/developers Do you agree that we can remove this entirely?

aschempp:
Looks like this feature was present forever in Contao (it was in the initial commit of Contao 2.6 on GitHub 😆). I have honestly never used it, so I can also update the PR to remove it if we agree.

ausi:
> Do you agree that we can remove this entirely?

Yes.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -343,12 +343,6 @@ protected function row($strPalette=null)
 
 		$xlabel = '';
 
-		// Toggle line wrap (textarea)
-		if (($arrData['inputType'] ?? null) == 'textarea' && !isset($arrData['eval']['rte']))
-		{
-			$xlabel .= ' ' . Image::getHtml('wrap.svg', $GLOBALS['TL_LANG']['MSC']['wordWrap'], 'title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['wordWrap']) . '" class="toggleWrap" onclick="Backend.toggleWrap(\'ctrl_' . $this->strInputName . '\')"');
```
leofeyer, line 349: I think we should also remove the wrap.svg icon, shouldn‘t we?

leofeyer (dismissed):
Changes look good but you will have to resolve the conflict first.

leofeyer (approved):

PR #4383, opened 2022-03-24
Rewrite tooltips to vanilla JS

aschempp:
The new implementation should behave and look exactly the same, but has some subtle but useful improvements:

- it only ever uses one DOM element for all tips, and that element is much simpler (just a DIV with text content). Previously every tip (once used) created its own DOM element wich consisted of multiple DIVs.
- it keeps the `title` attribute on elements, until these are focussed by mouse pointer. I would assume this is better for accessibility.
- It _automatically_ updates to any changes of the `title` attribute (or the tip content) since it is retrieved whenever the element is mouse-hovered. That's also why we don't need `Backend.retrieveInteractiveHelp` anymore before copying an element in wizards.
- It _automatically_ hides the tip when clicking button or links, therefore we don't need `Backend.hideInteractiveHelp` anymore. This was previously only the case for buttons, but now also _magically_ works e.g. for our module wizard in layouts. 😎 

It is (currently) not supported for developers to add their own tips (= register their own element for tips). But since the core automatically uses tips for basically anything with a `title` attribute, I don't think that's really necessary.


core-bundle/src/Resources/public/tips.js
```diff
@@ -0,0 +1,127 @@
...
+            init(el, -9, el.classList.contains('gimage') ? 60 : 30);
+        });
+
+        ['a[title]', 'input[title]', 'button[title]', 'time[title]', 'span[title]'].forEach(function(selector) {
+            node.querySelectorAll(selector).forEach(function (el) {
+                init(el, -9, ((selector === 'time[title]' || selector === 'span[title]') ? 26 : 30));
+            });
+        });
+    }
+
```
Toflar, line 112: What you could do is collect those as an array, then dispatch an event for other devs to add their own elements, and then init the array?

aschempp: yeah there are many options. I'm just unsure whether we want to add an API for it since any API we add would then need to be supported. If we add an API that noone uses then we add ourselves some unnecessary overhead. We could still add an API on request 🤷 
On the other hand, once we have "real" JS modules (when I rewrite it again) we can export the necessary method so anyone can use that in their own code.

leofeyer:
Did you test that this also works with subpalettes that are loaded via Ajax?

aschempp:
> Did you test that this also works with subpalettes that are loaded via Ajax?

yes, that's why we use the `MutationObserver` 😎

leofeyer (approved):

PR #4381, opened 2022-03-24
Allow to json-serialize the HtmlAttributes class

m-vo:
This is a small upgrade to the `HtmlAttributes` class, that allows it to be serialized with `json_encode()`. I need this functionality for the fragment controller tests of my upcoming PR but figured it would be nice to have anyways.

PR #4393, opened 2022-03-26
Modern fragment foundation

m-vo:
This PR is the foundation for our new core fragment controllers (PR upcoming). Our new fragments will also have new templates and a cleaned up template context, but there are BC layers in place to ease the transition. 

#### Setup
* Fragments without a template specified now default to the new `content_element/<type>` and `frontend_module/<type>`. The template names will always be set directly in the `RegisterFragmentsPass`. If you are currently using fragment controllers and do not want to update your templates, just specify them explicitly. 
* CTE/FMD Definitions made via `config.php` now have precedence. This allows a by-element/module fallback to the old system. It will also automatically work if someone registered a replacement for a core element/module.

#### Context
* We operate in two modes, modern and legacy (BC), depending on the template name: if it contains a `/` and exists as a Twig template we assume modern templates.
* For legacy templates, nothing changes, for modern templates the default context will look somewhat like this:
    ```php
    $headlineData = StringUtil::deserialize($model->headline ?: '', true);
    $attributesData = StringUtil::deserialize($model->cssID ?: '', true);

    $template->setData([
        'type' => $this->getType(),
        'data' => $model->row(),
        'section' => $section,
        'properties' => $properties,
        'attributes' => (new HtmlAttributes())
            ->setIfExists('id', $attributesData[0] ?? null)
            ->addClass($attributesData[1] ?? '', ...$classes),
        'headline' => [
            'text' => $headlineData['value'] ?? '',
            'tagName' => $headlineData['unit'] ?? 'h1',
        ],
    ]);
    ```

#### Compiling a response
* You can now omit the template when calling `AbstractFragmentController#render()` (`$view` is now optional).
* In the abstract `getReponse()` method we  now pass a new `FragmentTemplate`<sup>*)</sup> (from the `Twig` namespace) instead of the legacy `FrontendTemplate`/`FragmentTemplate`. It's a simple container object that allows setting template data and calling `getResponse()` but the latter is simply a closure set by the `AbstractFragmentController`. Effectively we're then calling `$this->render()`. This is a huge plus, because now there is only one path no matter if you're using `$template->getResponse()` or `$this->render()`:
    ```php
    public function getResponse(FragmentTemplate $template, ContentModel $model, Request $request): Response {
        // variant using the template object
        $template->foo = 'bar';

        // variant using the render method
        return $this->render(parameters: ['foo' => 'bar']);
    } 
    ```
* Modern fragments are Twig only, so for them the legacy template engine won't be used anymore. :tada: 

<sup>*)</sup> This does not break existing type hints because the new class still extends from the old one. But it sneakily disables calling parent methods, annotates them as `@internal` and throws an exception when trying to access them. People can now slowly migrate to the new type hint and in Contao 6 the base class can be dropped.

#### Todo
* [x] rebase once #4381 is merged
* [x] ~~move unrelated but still needed things out of the old template classes (maybe in a followup PR)~~ see #4403 
* [x] add some more tests


core-bundle/src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -163,6 +143,8 @@ static function ($matches)
 	 */
 	private function setCacheHeaders(Response $response)
 	{
+		// TODO: Move this. This part only depends on the current page and must also work if not using the legacy framework.
+
```
m-vo: I'll need help with this one. Where should this logic go? 

~~I removed the call to this method in this PR so that it isn't called when rendering legacy templates in the `AbstractFragmentController`. I could also revert the change and use `new Response($legacyTemplate->parse())` for now.~~ 

Reverted the changes for now, but ultimately, this code has to move :slightly_smiling_face:.

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -75,45 +81,72 @@ protected function getPageModel(): ?PageModel
...
+            return $this->render(parameters: $context);
+        };
+
+        $template = new FragmentTemplate($templateName, $onGetResponse);
 
-        /** @var Template $template */
-        $template = $this->container->get('contao.framework')->createInstance($templateClass, [$templateName]);
-        $template->setData($model->row());
+        if ($isLegacyTemplate && null !== ($row = $model->row())) {
+            $template->setData($row);
```
ausi:
```suggestion
        if ($isLegacyTemplate) {
            $template->setData((array) $model->row());
```
Nitpick.

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -75,45 +81,72 @@ protected function getPageModel(): ?PageModel
...
         return $template;
     }
 
+    /**
+     * @internal the addHeadlineToTemplate() method is considered internal in
+     * Contao 5 and won't be accessible anymore in Contao 6. Headline data is
+     * always added to the context of modern fragment templates.
+     */
     protected function addHeadlineToTemplate(Template $template, string|array|null $headline): void
     {
```
ausi: Should we add a deprecation for all these `@internal` methods?

```php
if (debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['class'] !== self::class) {
    trigger_deprecation('contao/core-bundle', '5.0', 'The "%s" method is considered internal and won't be accessible anymore in Contao 6.', __METHOD__);
}
```

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -155,8 +202,50 @@ protected function getType(): string
...
+            'model' => $model->row(),
+            'section' => $section,
+            'properties' => $properties,
+            'attributes' => (new HtmlAttributes())
+                ->setIfExists('id', $attributesData[0] ?? null)
+                ->addClass($attributesData[1] ?? '', ...$classes),
+            'headline' => [
+                'value' => $headlineData['value'] ?? '',
+                'unit' => $headlineData['unit'] ?? 'h1',
+            ],
```
ausi: `value` and `unit` is not very intuitive IMO, how about

```suggestion
            'headline' => $headlineData['value'] ?? '',
            'headlineTag' => $headlineData['unit'] ?? 'h1',
```

or maybe even

```suggestion
            'headline' => $headlineData['value'] ?? '',
            'headlineLevel' => (int) substr($headlineData['unit'] ?? 'h1', 1),
```

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -155,8 +202,50 @@ protected function getType(): string
...
+
+            return;
+        }
+
+        $headlineData = StringUtil::deserialize($model->headline ?: '', true);
+        $attributesData = StringUtil::deserialize($model->cssID ?: '', true);
+
+        $template->setData([
+            'type' => $this->getType(),
+            'model' => $model->row(),
```
ausi: Should we use a more general name like `data` or so here?

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -62,6 +62,8 @@
  * @property boolean      $trustedDevicesEnabled
  * @property array        $trustedDevices
  * @property string       $currentDevice
+ *
+ * @deprecated this class will be removed in Contao 6 with no replacement
```
ausi:
```suggestion
 * @deprecated this class will be removed in Contao 6, use Twig templates instead.
```

core-bundle/tests/EventListener/GlobalsMapListenerTest.php
```diff
@@ -18,30 +18,44 @@
...
-        yield ['foo', 'bar', 'baz'];
-
-        yield ['TL_CTE', null, ['foo' => 'bar']];
-        yield ['TL_CTE', ['foo' => 'bar'], ['foo' => 'baz']];
+        yield 'add single' => [
+            [],
+            ['text' => 'HeadlineFragment'],
+            ['text' => 'HeadlineFragment'],
+        ];
+
```
ausi, line 42:
```php
        yield 'overwrite single' => [
            ['text' => 'Headline'],
            ['text' => 'HeadlineFragment'],
            ['text' => 'HeadlineFragment'],
        ];
```

This would fail now because of the change in `array_replace_recursive()`?

core-bundle/src/Twig/Interop/FragmentTemplate.php
```diff
@@ -0,0 +1,700 @@
...
+    private array $context = [];
+
+    /**
+     * @param \Closure(self):Response $onGetResponse
+     *
+     * @internal
+     * @noinspection MagicMethodsValidityInspection
+     */
+    public function __construct(private string $templateName, private \Closure $onGetResponse)
+    {
```
ausi:
```suggestion
    {
        // Do not call parent constructor
```

AFAIK we add a comment if not calling the parent constructor is on purpose.

core-bundle/src/Resources/contao/classes/FrontendTemplate.php
ausi: A response listener? But maybe create a new ticket for this topic?

ausi (commented):
Some minor notes/nitpicks/questions.
Concept LGTM 🚀

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: IMHO the value for the headline needs to be an array, so that you can easily pass it around - and then "headline" is already in the names: `headline.value`, `headline.unit`. 

Use it like so…

```twig
{% include "<your-headline-partial>" with {headline} %}
```

… with an element like so:
```twig
<{{ headline.unit }}>{{ headline.value }}</{{ headline.unit }}>
```

Toflar: I was wondering if having an `HtmlTag` object would make sense. You already provided us with the attributes class so maybe that would make the collection complete? :)

aschempp (commented):
Two initial thoughts:

1. I think in legacy mode, we should continue to pass a `FrontendTemplate` instance to the `getResponse()` method. Maybe someone is using template methods, and I don't see an advantage to build it lazily?

2. all the protected methods that are now supposed to be made private/be removed for Contao 6 are _intentionally_ splitting code/logic to be overridden by child controllers, for the case that e.g. a content element does not have a the default fields. So e.g. if my content element does not have headlines, I would override the `addHeadlineToTemplate` method to noop. Are we sure this _feature_ should be dropped?

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: I actually had thought about a class for tags as well. But when implementing the templates, I noticed that most of the times you benefit from helper objects when:
- there are multiple parts that belong together but should be passed along together (e.g. the `Figure`)
- operations building parts of it are expensive and should be lazy (e.g. a "Download" containing previews that are built on the fly)
- there are a lot of parameters/variants where IDE autocompletion is a big plus

IMHO grouping things into objects only makes sense if the templates have the same level of abstraction: e.g. a `_figure.html.twig` renders a `Figure` object, a `_headline.html.twig` renders a `Headline` object. Then it''s clear what to pass to this template if you're using it standalone. How the markup itself is looking is completely up to the template. 

I feel classes for HTML tags most of the time break this abstraction because they are too specific - here is an example for two versions of a headline that were built with the same context but a different template:

```html
<h1 class="headline">foo</h1>

<div class="headline"><h1><span class="first">f</span>oo</h1></div>
```

I think I would leave this part as is for now but go over the contexts of all templates again in the upcoming PR for the fragments. Then we can probably see best where abstractions are needed. Wdyt?

m-vo:
> all the protected methods that are now supposed to be made private/be removed for Contao 6 are intentionally splitting code/logic to be overridden by child controllers, for the case that e.g. a content element does not have a the default fields. So e.g. if my content element does not have headlines, I would override the addHeadlineToTemplate method to noop. Are we sure this feature should be dropped?

IMHO yes. I think the reasoning might depend on if the `AbstractFragmentController` is supposed to work for any table/model or if its limited to content elements and modules. 

- If it's limited, I'm not seeing any benefit of the granular approach: These two scenarios always have a headline, attributes and so on because they always use the same tables/models that include these features. You're free to not use the data in your templates and your controller can even overwrite/drop any of the default context if needed. Do you have an example where you adjusted `addHeadlineToTemplate()`?

- In case it's not limited, you would need to implement `addDefaultDataToTemplate()` and set the whole default context. This looks fine to me - it's only a few lines of code. In fact, then, I would be in favor of dropping the method altogether and copying the code to both `AbstractContentElementController` and `AbstractFrontendModuleController` - because then there should not be an abstraction at all.

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: very :fox_face:y - implemented in 8519acd7bd80fe87f488395bada30d2f2ca51f36

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: done in 69ff895d8287feac987105c3a1625942d4eb35fa

Toflar:
Also 👍 on the concept, except for the questions raised by @aschempp. Let's see if he has a use case.

aschempp:
> If it's limited, I'm not seeing any benefit of the granular approach: These two scenarios always have a headline, attributes and so on because they always use the same tables/models that include these features. You're free to not use the data in your templates and your controller can even overwrite/drop any of the default context if needed. Do you have an example where you adjusted addHeadlineToTemplate()?

The following core content elements do not have a headline in the palette definition for example:
- html
- accordionStart
- accordionStop
- accordionSingle
- sliderStop
- toplink 
- alias
- article
- teaser 
- module

m-vo:
> The following core content elements do not have a headline in the palette definition for example: […]

True, but you would not remove the headline data for those, you would just not include the headline fragment in the template (or use another base template), no? Leaving it in for these elements also has the benefit, that I, as a user, can simply add the headline to the palette and template if I need it without having to create my own controller.

m-vo:
> I think in legacy mode, we should continue to pass a FrontendTemplate instance to the getResponse() method. Maybe someone is using template methods, and I don't see an advantage to build it lazily?

This change is intentional: Now the template type inside the `getResponse()` methods can be narrowed to `FragmentTemplate` and people can start migrating away from the now deprecated `Template` to the new `FragmentTemplate` in their existing controllers. Otherwise there would not be a smooth upgrade path. 

Using template methods other than `getResponse()` sounds wrong to me. I would drop support for this in Contao 5. An exception could maybe be `parse()`, but we already discouraged the use of that because it did not set the correct headers.

core-bundle/tests/EventListener/GlobalsMapListenerTest.php
m-vo: IMHO this is intended to fail, but maybe I'm misunderstanding the test case? The idea is: If there is already a legacy element registered, adding a fragment won't overrule it. This way you can add a `$GLOBALS['TL_CTE']['texts]['text']` entry to disable the new `TextController` fragment.

core-bundle/src/Resources/contao/classes/FrontendTemplate.php
m-vo: see https://github.com/contao/contao/issues/4403

m-vo:
Before this PR `$request->attributes->get('templateProperties', [])` are merged into the template context before the abstract `getResponse()` is called. It's similar with the current state of the PR though it's conflict-free under a `properties` key. 

But maybe we are missing the point in both the old and the new implementation of what this functionality tries to achieve. If it is  setting (model) data, so that a controller will behave differently, I'd suggest the following changes (example for content elements, analogous for modules) in `AbstractContentElementController`:

- always detach the model before calling the abstract `getResponse()`
- merge the `templateProperties` into the model data
- maybe even allow `ContentModel|array` in `__invoke`

ausi (dismissed):
🎉

ausi:
> If it is setting (model) data, so that a controller will behave differently, I'd suggest the following changes

Agree. I think modifying the model data (after detaching it) is the correct way for this use case.
Type hint `ContentModel|array` is not possible for `__invoke` due to BC AFAIK.

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -75,47 +83,91 @@ protected function getPageModel(): ?PageModel
...
+            $template->setData((array) $model->row());
+        }
 
         return $template;
     }
 
+    /**
+     * @internal
+     */
+    protected function isLegacyTemplate(string $templateName): bool
```
Toflar, line 123: Did you make this `protected` on purpose?

core-bundle/src/EventListener/GlobalsMapListener.php
```diff
@@ -31,7 +31,7 @@ public function onInitializeSystem(): void
     {
         foreach ($this->globals as $key => $value) {
             if (\is_array($value) && isset($GLOBALS[$key]) && \is_array($GLOBALS[$key])) {
-                $GLOBALS[$key] = array_replace_recursive($GLOBALS[$key], $value);
+                $GLOBALS[$key] = array_replace_recursive($value, $GLOBALS[$key]);
```
Toflar, line 36: Is that a bug? Sorry this just feels wrong in this PR :D

Toflar (dismissed):
Looks excellent! Just some nitpicks but I'll approve the concept anyway.

core-bundle/src/EventListener/GlobalsMapListener.php
ausi: Maybe rename the listener so that it is clear it is only meant for `TL_CTE` and `TL_MOD`?

m-vo: This is to give content elements registered via the `config.php` precedence. The idea behind this is, that people can opt out of the new system by recreating the entry in their `config.php` file. Also, if someone or an extension registered an element instead of the core's element, this would still work.

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: Yes, the two implementers (`AbstractContentElementController` + `AbstractFrontendModuleController`) use this as well.

core-bundle/src/EventListener/GlobalsMapListener.php
m-vo: > Maybe rename the listener so that it is clear it is only meant for TL_CTE and TL_MOD?

:+1: Something like `MapContentElementAndFrontendModuleGlobalDefinitionsListener`? :stuck_out_tongue:

Toflar: Ah, well I don't mind the name but a comment would be great 😅

m-vo: Agreed, I added a comment in 9cdfa470c1b141986645e5bd4457e266f912b6d1

ausi (dismissed):

Toflar (dismissed):

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -75,47 +83,91 @@ protected function getPageModel(): ?PageModel
...
-    protected function createTemplate(Model $model, string $templateName): Template
+    protected function createTemplate(Model $model, string|null $fallbackTemplateName = null): FragmentTemplate
     {
-        if (isset($this->options['template'])) {
-            $templateName = $this->options['template'];
+        $templateName = $this->getTemplateName($model, $fallbackTemplateName);
+
+        if ($isLegacyTemplate = $this->isLegacyTemplate($templateName)) {
+            // TODO: enable deprecation once existing fragments have been adjusted
+            // trigger_deprecation('contao/core-bundle', '5.0', 'Creating fragments with legacy templates is deprecated and will not work anymore in Contao 6.');
```
leofeyer: Why don‘t we adjust the existing fragments directly? At least I would enable the deprecation, so we don‘t forget about it.

core-bundle/src/Twig/FragmentTemplate.php
```diff
@@ -0,0 +1,719 @@
...
+     */
+    public function __get(/* string */ $key)
+    {
+        return $this->get($key);
+    }
+
+    /**
+     * @param string $key
+     */
+    public function __isset(/* string */ $key): bool
```
leofeyer: Why don‘t we adjust the parent class instead?

core-bundle/src/Resources/contao/templates/_new/back_end/module_wildcard.html.twig
```diff
@@ -0,0 +1,12 @@
+{% trans_default_domain 'contao_modules' %}
+
+{% if title %}
+    <h1>{{ title }}</h1>
+{% endif %}
+
+<div class="tl_gray">
+    <span class="upper">### {{ ('FMD.' ~ type ~ '.0')|trans }} ###</span>
+    <br>
+    {% set href = path('contao_backend', {do: 'themes', table: 'tl_module', act: 'edit', id, rt: request_token}) %}
```
leofeyer: Why do we compile this in the template instead of passing the URL in the data?

leofeyer (changes requested):
The template structure needs to be adjusted before we merge this.

1. I don‘t like the `_new` subfolder. Is this only temporary?
2. We never use `back_end` (two words) in the code base. It is always `backend`.

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -75,47 +83,91 @@ protected function getPageModel(): ?PageModel
...
+            return $this->render($template->getName(), $context, $preBuiltResponse);
+        };
+
+        $template = new FragmentTemplate($templateName, $onGetResponse);
 
-        /** @var Template $template */
-        $template = $this->container->get('contao.framework')->createInstance($templateClass, [$templateName]);
-        $template->setData($model->row());
+        if ($isLegacyTemplate) {
+            $template->setData((array) $model->row());
```
leofeyer:
```suggestion
            $template->setData($model->row());
```

The type cast seems redundant, because `row()` always returns an array, doesn‘t it?

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -14,21 +14,27 @@
...
 {
     public function __invoke(Request $request, ContentModel $model, string $section, array $classes = null): Response
     {
-        $type = $this->getType();
-        $template = $this->createTemplate($model, 'ce_'.$type);
+        $template = $this->createTemplate($model, 'ce_'.$this->getType());
+
+        $this->addDefaultDataToTemplate(
+            $template,
+            (array) $model->row(),
```
leofeyer:
```suggestion
            $model->row(),
```

core-bundle/src/Resources/contao/templates/_new/back_end/module_wildcard.html.twig
m-vo: I usually try to generate routes in the templates; here is my  reasoning:

- They are only generate when needed.
- You have control over arguments.
- You might want to add additional links anyway and its consistent this way.

In this case both variants are fine, I guess, especially because the BE URLs are very specific. Would you rather move this to the controller, instead?

core-bundle/src/Twig/FragmentTemplate.php
m-vo: We could, but as the base class is `Template`, I'm not sure if this will unnecessarily break things in the legacy framework. Shall we? :grin: 

Btw., there are also new and typed methods `set` `get` and `has`, that I would prefer using anyways.

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: Agreed, I enabled the warning in 400216f8485edf5534f3777e089bfbde565a593e and created an [issue](https://github.com/contao/contao/issues/4461) as a reminder.

m-vo:
> I don‘t like the _new subfolder. Is this only temporary?

Yes, the templates can reside anywhere as long it is a Contao template path and the `.root` file exists. At a later point I would try to move the PHP templates e.g. in a `templates_legacy` folder and the new ones one level up. Maybe when we move the `contao` folder to the root anyway.

m-vo:
> We never use back_end (two words) in the code base. It is always backend.

see 7bfcd1821c82db82126deed5bd41c3e1971c7c7e

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: It's mainly to facilitate testing. In all our current unit test it would yield`null` instead of `[]`. I'm planning on building a test case for content element / module controllers in #4444. We can then refactor the old tests and get rid of this.

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: Added the expected deprecations in f6f303dff0dcf99c118c5f0238e2392620735326. For now this is IMHO the correct way to do it.

Even when we finally create new templates in #4393 we, we will likely keep the old templates for BC (and suppressed deprecation warnings) and include the new ones from there. There are several ways to achieve this, but let's please keep this out of this PR.

m-vo:
Todo:
* [x] revert https://github.com/contao/contao/pull/4393/commits/f6f303dff0dcf99c118c5f0238e2392620735326
* [x] remove triggering the deprecation for now

m-vo:
Rebased and resolved conflicts.

Toflar (dismissed):

ausi (dismissed):

core-bundle/src/Controller/AbstractFragmentController.php
leofeyer: > In all our current unit test it would yield`null` instead of `[]`.

Then we should fix the tests, shouldn‘t we?

core-bundle/src/Twig/FragmentTemplate.php
leofeyer: Yes. We are allowed to break BC in Contao 5.0.

core-bundle/src/Resources/contao/templates/_new/back_end/module_wildcard.html.twig
leofeyer: @contao/developers WDYT?

leofeyer:
> > I don‘t like the _new subfolder. Is this only temporary?
> 
> Yes, the templates can reside anywhere as long it is a Contao template path and the `.root` file exists. At a later point I would try to move the PHP templates e.g. in a `templates_legacy` folder and the new ones one level up. Maybe when we move the `contao` folder to the root anyway.

Since the new templates are Twig templates, why do we not store them in `core-bundle/templates` instead of `core-bundle/src/Resources/contao/templates`?

m-vo:
> Since the new templates are Twig templates, why do we not store them in core-bundle/templates instead of core-bundle/src/Resources/contao/templates?

Because that's not a Contao template directory. I also don't think we should do it, because contrary to Symfony bundle templates our templates are managed (`@Contao` dynamic inheritance).

core-bundle/src/Twig/FragmentTemplate.php
m-vo: see d0d73f1f8b37062f39d00acf6a64b95c6702ff82

leofeyer: I like. 👍

core-bundle/src/Resources/contao/templates/_new/back_end/module_wildcard.html.twig
ausi: For me, compiling the URL in the template is fine for these cases.

core-bundle/src/Controller/AbstractFragmentController.php
ausi: IMO, the array cast is good practise as the row method has no native return type. Adding an `array` return type to row method would also be fine with me as an alternative.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
ausi: See https://github.com/contao/contao/pull/4393#discussion_r855875710

ausi (dismissed):
❤️

core-bundle/src/Controller/AbstractFragmentController.php
leofeyer: I agree. Still it seems bad practice to yield `null` in the unit tests even though we know that it is an invalid return value. If we added `: array` as return type, the tests would fail immediately, therefore they should yield the correct return value from the beginning IMHO.

core-bundle/src/Resources/contao/library/Contao/System.php
```diff
@@ -125,9 +125,9 @@ protected function __construct()
 	 *
 	 * @param string $strKey The property name
 	 *
-	 * @return mixed|null The property value or null
+	 * @return mixed The property value or null
 	 */
-	public function __get($strKey)
+	public function __get(string $strKey): mixed
```
ausi: This breaks, as some subclasses like `User` are incompatible with their method signature.

core-bundle/src/Controller/AbstractFragmentController.php
ausi: > If we added `: array` as return type, the tests would fail immediately

I think the Mock implementation would return an empty array by default then (because it knows the return type of the method).

leofeyer: And that is all I‘m asking to be changed in this PR. 🙈

m-vo: Do you want to adjust these things in #4536 as well or do we merge your PR first?

m-vo: see f5bf5458ff5c3f4efd89b9d1c23088a85bba0e62

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
m-vo: see f5bf5458ff5c3f4efd89b9d1c23088a85bba0e62

Toflar (dismissed):

Toflar (dismissed):

ausi (dismissed):

core-bundle/tests/Twig/FragmentTemplateTest.php
```diff
@@ -0,0 +1,116 @@
...
+                continue;
+            }
+
+            $args = array_map(
+                function (\ReflectionParameter $parameter) {
+                    if (!($type = $parameter->getType()) instanceof \ReflectionType) {
+                        return null;
+                    }
+
+                    /** @phpstan-ignore-next-line because PHPStan doesn't think getName() exists */
```
leofeyer:
```suggestion
                    if (!($type = $parameter->getType()) instanceof \ReflectionNamedType) {
                        return null;
                    }

```

PHPStan is absolutely right here and should not be silenced.

leofeyer (changes requested):

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -62,6 +62,8 @@
  * @property boolean      $trustedDevicesEnabled
  * @property array        $trustedDevices
  * @property string       $currentDevice
+ *
+ * @deprecated this class will be removed in Contao 6, use Twig templates instead
```
leofeyer: Do we really want to deprecate the `Template` class? It will generate a lot of deprecation warnings in IDEs such as PHPStorm and people cannot really do anything about it as long as .html5 templates are in use.

m-vo: Yes, IMHO we should, because then you'll see that your old usage when extending the abstract is deprecated:
image

m-vo: >  as long as .html5 templates are in use.

But the plan was to deprecate them in 5 anyway, no?

leofeyer: Yes, they will be deprecated. But as long as they are still supported, people will have to use the `Template` class. It would be a deprecation you cannot get rid of and at least I find these very annoying.

leofeyer (dismissed):

core-bundle/src/Resources/contao/library/Contao/Template.php
m-vo: Well, currently it's only a soft deprecation, not one that would trigger an error. IMHO there are no cases where you need to instantiate a template yourself although the core does. But we could also postpone this discussion and see what's the state close to the release? Wdyt?

leofeyer: Agreed.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

bytehead:
This is great! ❤️

PR #4376, opened 2022-03-23
Increase the speed of the functional tests

ausi:

PR #4394, opened 2022-03-26
Speed up functional tests

ausi:
Followup to #4376

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -308,7 +308,7 @@
             "@ecs-legacy",
             "@ecs-templates"
         ],
-        "ecs": "@php -dxdebug.mode=off tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests tools/*/src --config tools/ecs/config/default.php --fix --ansi",
+        "ecs": "@php -dxdebug.mode=off tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests test-case/src tools/*/src --config tools/ecs/config/default.php --fix --ansi",
         "ecs-legacy": "@php -dxdebug.mode=off tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao --config tools/ecs/config/legacy.php --fix --ansi",
         "ecs-templates": "@php -dxdebug.mode=off tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao/templates --config tools/ecs/config/template.php --fix --ansi",
         "functional-tests": "@php -dxdebug.mode=off vendor/bin/phpunit --testsuite=functional --colors=always",
--- a/composer.json
+++ b/composer.json
@@ -308,15 +308,15 @@
             "@ecs-legacy",
             "@ecs-templates"
         ],
-        "ecs": "@php tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests tools/*/src --config tools/ecs/config/default.php --fix --ansi",
-        "ecs-legacy": "@php tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao --config tools/ecs/config/legacy.php --fix --ansi",
-        "ecs-templates": "@php tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao/templates --config tools/ecs/config/template.php --fix --ansi",
-        "functional-tests": "@php vendor/bin/phpunit --testsuite=functional --colors=always",
-        "monorepo-tools": "@php tools/monorepo/vendor/bin/monorepo-tools composer-json --validate --ansi",
-        "phpstan": "@php tools/phpstan/vendor/bin/phpstan analyze --ansi",
-        "psalm": "@php tools/psalm/vendor/bin/psalm --no-suggestions",
-        "require-checker": "@php tools/require-checker/vendor/bin/composer-require-checker check --config-file=tools/require-checker/config.json composer.json",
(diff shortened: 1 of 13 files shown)


m-vo (dismissed):
Amazing! :heart:

aschempp (changes requested):
It looks like an error that `Not Fond` is returned instead of `(404 Not Found)`. Most likely the new one is the Symfony Response and not our `tl_page` title, because some of the test data is wrong. Pretty sure the tests must not be changed!

composer.json
```diff
@@ -308,15 +308,15 @@
...
-        "ecs-legacy": "@php tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao --config tools/ecs/config/legacy.php --fix --ansi",
-        "ecs-templates": "@php tools/ecs/vendor/bin/ecs check *-bundle/src/Resources/contao/templates --config tools/ecs/config/template.php --fix --ansi",
-        "functional-tests": "@php vendor/bin/phpunit --testsuite=functional --colors=always",
-        "monorepo-tools": "@php tools/monorepo/vendor/bin/monorepo-tools composer-json --validate --ansi",
-        "phpstan": "@php tools/phpstan/vendor/bin/phpstan analyze --ansi",
-        "psalm": "@php tools/psalm/vendor/bin/psalm --no-suggestions",
-        "require-checker": "@php tools/require-checker/vendor/bin/composer-require-checker check --config-file=tools/require-checker/config.json composer.json",
-        "service-linter": "@php tools/service-linter/bin/lint-service-ids --ansi",
-        "unit-tests": "@php vendor/bin/phpunit --colors=always"
+        "ecs": "@php -dxdebug.mode=off tools/ecs/vendor/bin/ecs check *-bundle/bin *-bundle/src *-bundle/tests test-case/src tools/*/src --config tools/ecs/config/default.php --fix --ansi",
```
leofeyer: I like this PR a lot, but can we please revert the `-dxdebug.mode=off`? I have set up an alias `phpx` that runs PHP with Xdebug enabled, so the regular `php` binary is fast and I can enable Xdebug when I need it. (This is a pretty common setup as far as I know.) With the change, I cannot use `phpx` anymore because Xdebug will always be turned off.

ausi: Done in ed34fccb6a662c22a8b2aa8aad5b4f75cd48bb06

How about adding in a PhpUnit extension that prints a warning if xdebug is enabled?

ausi: Added in 2035ae3f2f288a65d97a6ff5cd73b0bb914eca07

m-vo: > This is a pretty common setup as far as I know. 
> With the change, I cannot use phpx anymore because Xdebug will always be turned off.

Is it? How are you supposed to debug requests sent from a browser with this setup? And why do you want to run the shortcuts *with* Xdebug enabled?

test-case/src/WarnXdebugPhpunitExtension.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Contao\TestCase;
+
+use PHPUnit\Runner\BeforeFirstTestHook;
+
+class WarnXdebugPhpunitExtension implements BeforeFirstTestHook
+{
+    public function executeBeforeFirstTest(): void
+    {
+        if (\is_callable('xdebug_info') && ([] === xdebug_info('mode') || xdebug_info('mode') === ['off'])) {
```
ausi:
```suggestion
        if (\is_callable('xdebug_info') && [] !== xdebug_info('mode') && xdebug_info('mode') !== ['off']) {
```

composer.json
m-vo: Unfortunately, when composer starts a new process with `@php` the arguments aren't passed thru. So there is no way for me to run the scripts from the CLI and disable Xdebug on the fly (if I do not want to parse the file and do all sorts of string manipulation magic - which I don't). With Xdebug 3 switching on the fly is finally possible, so I don't see a point of disabling the extension. For users of the symfony binary having two PHP binaries for the same version does not work either. 

Can we maybe find a solution that fits both our needs?

leofeyer: Yes. Preferably in a separate pull, so we can merge this one now.

test-case/src/WarnXdebugPhpunitExtension.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Contao\TestCase;
+
+use PHPUnit\Runner\BeforeFirstTestHook;
+
+class WarnXdebugPhpunitExtension implements BeforeFirstTestHook
+{
+    public function executeBeforeFirstTest(): void
+    {
+        if (\is_callable('xdebug_info') && [] !== xdebug_info('mode') && xdebug_info('mode') !== ['off']) {
```
leofeyer: According to https://php.watch/articles/xdebug_info, the return value of `xdebug_info()` is a string. 🤔

m-vo: Could this be the case for Xdebug 2? At least for me with version 3 it's an array.

ausi: It is `mixed`: https://xdebug.org/docs/all_functions#xdebug_info and an array for `'mode'`

leofeyer: Ah, I missed that. Sorry.

leofeyer (dismissed):

leofeyer (approved):

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #4396, opened 2022-03-27
Deprecate the StringUtil::toHtml5() method

m-vo:
While building the new fragment controllers I found this nice legacy logic that was used in nearly every request. Deprecating it here, so that it can finally go in Contao 5.


ausi (dismissed):

Toflar (dismissed):

fritzmg (dismissed):

leofeyer:
Are we sure that we want to remove this in Contao 4.13 without even a migration? Older websites (aka websites that already ran with Contao 3.5) might still have XHTML code stored in the database. IMHO we should deprecate the method but only remove its usages in Contao 5.0. Or at least provide a migration.

Toflar:
Thinking about it again, I might be with @leofeyer here. We don't gain anything by removing them now already.

ausi:
I’d vote for removing the usages, but I’m fine either way (as long as we can remove it in Contao 5 ☺️)

fritzmg:
I am also fine with deprecating it in 4.13 and removing it in 5.0.

m-vo:
> IMHO we should deprecate the method but only remove its usages in Contao 5.0.

Are we fine with this? This will trigger a load of self deprecations…

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -523,9 +523,13 @@ public static function toXhtml($strString)
 	 * @param string $strString The XHTML string
 	 *
 	 * @return string The HTML5 string
+	 *
+	 * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0
 	 */
 	public static function toHtml5($strString)
 	{
+		trigger_deprecation('contao/core-bundle', '4.13', 'The "StringUtil::toHtml5()" method has been deprecated and will no longer work in Contao 5.0.');
```
ausi, line 529:
```suggestion
		if (str_starts_with(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['class'] ?? '', 'Contao\\')) 
		{
			trigger_deprecation('contao/core-bundle', '4.13', 'The "StringUtil::toHtml5()" method has been deprecated and will no longer work in Contao 5.0.');
		}
```
To not throw self deprecations?

leofeyer: Triggering a deprecation is meant for things that people can change to get rid of the deprecation, however, in this case we have to continue using the method, so we should only add a `@deprecated` comment and no `trigger_deprecation()` at all in my opinion.

ausi: I’d be fine with that too. But wouldn’t this make it harder to notice for everyone using `StringUtil::toHtml5()` themself?
With my suggestion from above, extension developers would get deprecation notices while the core itself does’t trigger them.

Toflar:
Other question: When did we remove XHTML support? Does anybody remember?

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
leofeyer: Agreed.

But why do we not provide a migration for the core fields and stop using the method ourselves?

ausi: A migration would be hard to get right IMO (and hard to detect if it should be executed).

But a migration that only replaces the ` rel="lightbox` part might be suitable. The lightbox replacement is the only one necessary I think, the others are only “cosmetic” IMO.
The detection would then simly need to check for `LIKE '% rel="lightbox%'`

m-vo: I wouldn't want a migration changing any of my database fields tbh. And on top of detecting the content of the database field we would need to make sure the code in question is actually used (e.g. `$GLOBALS['TL_CTE']['texts']['text']` not overwritten by anyone and so on). 

Then, I'd rather favor a cut in Contao 5 and the :fox_face:-deprecation without changing usages in 4.13.

m-vo: Judging by https://github.com/contao/core/issues/4073 the `rel` attribute was set by tinyMCE. The lightbox option seems gone in the current demo, but maybe this is something that can be configured in there? (producing the wanted tag in the first place)

ausi: AFAIK `rel="lightbox…` does no longer get inserted by tinyMCE.

aschempp:
Regardless of whether we remove this in 4.13 or 5.0, what is the upgrade path for users? Looks like we need some sort of heavy migration that is only executed on request?

ausi:
> what is the upgrade path for users?

IMO a migration that replaces `rel="lightbox…` with `data-lightbox="…`, see https://github.com/contao/contao/pull/4396#discussion_r837249794

m-vo:
MIgration for the lightbox string added in 0e9d0b13a5c4b6b868672516da1ae8c0dd0f0b9e.

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+
+        foreach ($this->getTargets() as [$table, $column]) {
+            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)[$column])) {
+                continue;
+            }
+
+            $test = $this->connection->fetchOne(
+                "SELECT TRUE FROM $table WHERE `$column` LIKE '% rel=\"lightbox%\"%' LIMIT 1;"
```
ausi:
```suggestion
                "SELECT TRUE FROM $table WHERE `$column` REGEXP ' rel="lightbox(\\\\[([^\\\\]]+)\\\\])?\"' LIMIT 1;"
```

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return false;
+    }
+
+    public function run(): MigrationResult
+    {
+        foreach ($this->getTargets() as [$table, $column]) {
+            $values = $this->connection->fetchAllKeyValue(
+                "SELECT id, $column FROM $table WHERE `$column` LIKE '% rel=\"lightbox%\"%';"
```
ausi:
```suggestion
                "SELECT id, value FROM $table WHERE `$column` REGEXP ' rel="lightbox(\\\\[([^\\\\]]+)\\\\])?\"'"
```

ausi (dismissed):
I’d use the exact same regex for the query to prevent the risk of the migration never resolving itself.
Apart from that, this approach is the way to go IMO, because it is BC and stops us from doing werid string replacements on the output.

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+use Contao\CoreBundle\Migration\AbstractMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @internal
+ */
+class RelLightboxMigration extends AbstractMigration
+{
+    public static array $targets = [
```
m-vo:
```suggestion
    private static array $targets = [
```

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return false;
+    }
+
+    public function run(): MigrationResult
+    {
+        foreach ($this->getTargets() as [$table, $column]) {
+            $values = $this->connection->fetchAllKeyValue(
+                "SELECT id, value FROM $table WHERE `$column` REGEXP ' rel="lightbox(\\\\[([^\\\\]]+)\\\\])?\"'"
```
ausi:
```suggestion
                "SELECT id, value FROM $table WHERE `$column` REGEXP ' rel=\"lightbox(\\\\[([^\\\\]]+)\\\\])?\"'"
```

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+
+        foreach ($this->getTargets() as [$table, $column]) {
+            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)[$column])) {
+                continue;
+            }
+
+            $test = $this->connection->fetchOne(
+                "SELECT TRUE FROM $table WHERE `$column` REGEXP ' rel="lightbox(\\\\[([^\\\\]]+)\\\\])?\"' LIMIT 1;"
```
ausi:
```suggestion
                "SELECT TRUE FROM $table WHERE `$column` REGEXP ' rel=\"lightbox(\\\\[([^\\\\]]+)\\\\])?\"' LIMIT 1;"
```

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return false;
+    }
+
+    public function run(): MigrationResult
+    {
+        foreach ($this->getTargets() as [$table, $column]) {
+            $values = $this->connection->fetchAllKeyValue(
+                "SELECT id, value FROM $table WHERE `$column` REGEXP ' rel=\"lightbox(\\\\[([^\\\\]]+)\\\\])?\"'"
```
ausi:
```suggestion
                "SELECT id, `$column` FROM $table WHERE `$column` REGEXP ' rel=\"lightbox(\\\\[([^\\\\]]+)\\\\])?\"'"
```

My mistake from the previous suggestion 🙈

ausi (dismissed):

m-vo:
TLDR; the migration will transform text/teaser/comment fields like so (actual output from manual testing):

`<p>I'm inside a <a href="#" rel="lightbox[foo]">x</a> <a href="#" rel="lightbox">y</a> text field.</p>` &rarr;
`<p>I'm inside a <a href="#" data-lightbox="foo">x</a> <a href="#" data-lightbox="">y</a> text field.</p>`

aschempp:
Is the lightbox really the only XHML we are transforming? If I remember correctly, we also had some issues that TinyMCE was generating XHTML instead of HTML5 (like `<u>`) that got fixed this way. Are we just agreeing that thats fine since it will work anyway (but be "less perfect")? Would be fine with me, just wanted to bring it up 🙃

ausi:
> Are we just agreeing that thats fine since it will work anyway (but be "less perfect")?

Yes, the other replacements do not change the behavior, so this is fine IMO. 
As it is now in this pull request is not “less perfect” to me, quite the contrary ☺️

aschempp (dismissed):

leofeyer:
@m-vo Can you please fix the CI chain?

core-bundle/src/Resources/config/migrations.yml
```diff
@@ -109,6 +109,11 @@ services:
         arguments:
             - '@database_connection'
 
+    contao.migration.version_413.text_fields:
```
ausi:
```suggestion
    contao.migration.version_413.rel_lightbox:
```

ausi (dismissed):

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4397, opened 2022-03-27
Fix several VFS bugs

m-vo:
This PR fixes two bugs and a major shortcoming in the VFS implementation:

1) The `AbstractDbafsMetadataEvent` was still in the wrong namespace (contrary to the two child classes). I did not notice it until I wanted to add things in Contao 5. :see_no_evil: It's moved now. [eb11863cbcca142cb200b8865dec1faba4b99b74]
2) There was an issue with prefixed `Dbafs` implementations when retrieving items by UUID. At least the tests were wrong as well. [7906b75d8b4dd8deaad0c90f97d1b1fd8948a8ba]
3) There was currently no way to get a `FilesystemItem` except when listing directories, therefore there is now a `get()` method. As this should also be in the interface, we IMHO need to add this to here (4.13) as well. This is kind of an important part if you want to pass around `FilesystemItems` and use their lazy nature. [964ff70e239a6e58517d4225a726b3412dd93772]

### DIFF
--- a/core-bundle/tests/Filesystem/VirtualFilesystemTest.php
+++ b/core-bundle/tests/Filesystem/VirtualFilesystemTest.php
@@ -427,16 +427,12 @@ class VirtualFilesystemTest extends TestCase
         $mountManager = $this->createMock(MountManager::class);
         $mountManager
             ->method('fileExists')
-            ->willReturnCallback(
-                static fn (string $path): bool => 'foo/file_a' === $path
-            )
+            ->willReturnCallback(static fn (string $path): bool => 'foo/file_a' === $path)
         ;
 
         $mountManager
             ->method('directoryExists')
-            ->willReturnCallback(
-                static fn (string $path): bool => 'foo/dir_a' === $path
-            )
+            ->willReturnCallback(static fn (string $path): bool => 'foo/dir_a' === $path)
         ;
 
         $mountManager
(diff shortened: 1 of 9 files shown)


Toflar (dismissed):

ausi (dismissed):

core-bundle/tests/Filesystem/VirtualFilesystemTest.php
```diff
@@ -418,6 +420,159 @@ public function testMove(): void
...
+        $this->assertInstanceOf(FilesystemItem::class, $fileA);
+        $this->assertTrue($fileA->isFile());
+
+        $this->assertSame(0, $handlerInvocationCount);
+
+        $this->assertSame(54321, $fileA->getLastModified());
+        $this->assertSame(2048, $fileA->getFileSize());
+        $this->assertSame('text/csv', $fileA->getMimeType());
+
+        /** @phpstan-ignore-next-line */
```
leofeyer, line 544: Why do we have to ignore these lines (548, 565 and 573)?

m-vo: PHPStan thinks the compared variable is still `0` at this point (which it is not, because it is altered by the handlers). For the next calls (565, 573) it thinks its the never type. Not that I understand why this is the case tbh.

```
Error: Call to method PHPUnit\Framework\Assert::assertSame() with 3 and 0 will always evaluate to false.
Error: Call to method PHPUnit\Framework\Assert::assertSame() with 3 and *NEVER* will always evaluate to false.
Error: Call to method PHPUnit\Framework\Assert::assertSame() with 7 and *NEVER* will always evaluate to false.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4402, opened 2022-03-28
Fixed nested http client options not supported

### LINKED ISSUE #4387: Cannot set additional headers for crawler
### Affected version(s)

4.9, 4.13, 5.x

### Description

Currently it is only possible to set additional [HTTP Client Options](https://symfony.com/doc/current/reference/configuration/framework.html#reference-http-client) that are scalars. But you cannot set options that are arrays, like additional headers. For instance, the following will not work:

```yaml
contao:
    crawl:
        default_http_client_options:
            headers:
                'User-Agent': 'Firefox'
```

The error will be:

```
Invalid type for path "contao.crawl.default_http_client_options.headers". Expected "scalar", but got "array".
```

I think we should probably allow "everything" there? Or should we copy the configuration tree of the Symfony Framework Bundle?

/cc @Toflar

Toflar:
Fixes https://github.com/contao/contao/issues/4387

I tried getting the options from the framework bundle but it's too complicated and I don't care enough to pursue it.

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -414,7 +414,7 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->arrayNode('default_http_client_options')
                     ->info('Allows to configure the default HttpClient options (useful for proxy settings, SSL certificate validation and more).')
-                    ->prototype('scalar')->end()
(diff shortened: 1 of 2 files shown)


fritzmg (approved):

leofeyer:
Thank you @Toflar.

Issue #4386, opened 2022-03-24
Add shared max age to response by default


m-vo:
### Description

```diff
// AbstractContentElementController
-        return $this->getResponse($template, $model, $request);
+        $response = $this->getResponse($template, $model, $request);
+        $this->addSharedMaxAgeToResponse($response, $model);
+        return $response;
```

To disable you can still override the method with a no-op in your controller class.

aschempp:
This would break any existing cache information in the response.

m-vo:
> To disable you can still override the method with a no-op in your controller class

aschempp:
That is very bad UX. My controller returns a **custom** response, and I still have to override the method to not have my "already correct" response modified? That makes no sense.

m-vo:
Didn't we discuss this already and decided this was the way to go for 5? IIRC the reasoning was: in general there is no reason not to use the feature. With the current approach, everyone has to manually add this nearly every time.

leofeyer:
I‘m with @aschempp on this one. If someone send a response with a shared max age, we should not "correct" it. Maybe we can check if the header exists already?

m-vo:
Just to make sure, we're talking about the same thing: It's not about adding anything to the `AbstractFragmentController` but the `AbstractContentElementController`. And this base class implements certain things common to our content elements (class names, css ids, headlines, automatic model tagging and so on …). Why should the start/stop setting that is present in every core content element be an exception from this? We would need to add the call to every of our content element controllers and - fun fact - we already forgot to do so for the already existing ones.

If this is really a problem in a certain use case maybe the base class for this should rather be `AbstractFragmentController` (or we need another one).

m-vo:
Alternatively we could pass the pre-configured response in `getResponse()` as well. `$this->render()` supports passing in a response and the new fragment template could allow this as well.

leofeyer:
> Just to make sure, we're talking about the same thing: It's not about adding anything to the `AbstractFragmentController` but the `AbstractContentElementController`.

Ah, I missed that. Then it makes sense I guess.

m-vo:
I looked at this again, IMHO this is pretty easy to solve with the new architecture of #4393. :slightly_smiling_face:

See [`f520dc8` (#4393)](https://github.com/contao/contao/pull/4393/commits/f520dc808731eb7d7bfede54b4f18718036c7721).

[...]

PR #4407, opened 2022-03-28
Cache handling for start and stop dates

aschempp:
After some long discussions with @m-vo regarding https://github.com/contao/contao/issues/4386, here's a draft the solution we came up with.

It basically boils down to _it was always wrong_, but we somehow knew that. Some might remember https://github.com/aschempp/contao/commit/245abdd80024e79c2108c1f122146c8015b91fc9 which I did about a year ago during a public call. It was also a foundation for https://github.com/contao/contao/pull/4393, though that seems to be targeted at Contao 5.

However, this is a bugfix since it is currently wrong in Contao. @m-vo and I discussed and agreed that a content element must be responsibility for its cache time, the same as with cache tags.

The difficult thing is that an element (or article) with a start date in the future is not parsed by the system, but it should be to affect the cache time. And that's why `AbstractContentElementController::addSharedMaxAgeToResponse` was always wrong, because it (probably) adds a start time to an element that should not have been generated in the first place.

There are now two cases handled in the core modules:
 1. if an element has a start date in the future (and not a stop date in the past, since that would be invalid), a (empty/fake) response is added to the `ResponseCacheStrategy` to make sure the _main page_ cache is cleared at the point in time when the element should become visible

 2. alternatively, if an element has a stop date in the future, the stop date is added to also reset the page cache at that time.

 There is some special hacking magic regarding news and events, since they lazily generate their content elements if requested by the template. The element **count** (`->hasDetails`) is always the _currently available elements_, so we lazily check whether elements were generated and otherwise check to add a fake response for upcoming elements.
This does unfortunately not work for event lists (or it would be very hard to do), so I skipped that for now.

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
@@ -14,6 +14,8 @@ use Contao\CoreBundle\Exception\InternalServerErrorException;
 use Contao\CoreBundle\Exception\PageNotFoundException;
 use Contao\CoreBundle\Exception\RedirectResponseException;
 use Patchwork\Utf8;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\HttpCache\ResponseCacheStrategyInterface;
 
 /**
  * Front end module "event reader".
@@ -183,6 +185,7 @@ class ModuleEventReader extends Events
 
 		$until = '';
 		$recurring = '';
+		$needsCacheHeaders = false;
 		$arrRange = array();
 
 		// Recurring event
@@ -256,10 +259,14 @@ class ModuleEventReader extends Events
 		{
 			$id = $objEvent->id;
 
-			$objTemplate->details = function () use ($id)
+			// Enable cache headers if event has details but they are not generated
+			$needsCacheHeaders = true;
+
+			$objTemplate->details = function () use ($id, &$needsCacheHeaders)
 			{
+				$needsCacheHeaders = false;
 				$strDetails = '';
-				$objElement = ContentModel::findPublishedByPidAndTable($id, 'tl_calendar_events');
+				$objElement = ContentModel::findVisibleByPidAndTable($id, 'tl_calendar_events');
 
 				if ($objElement !== null)
 				{
@@ -391,6 +398,24 @@ class ModuleEventReader extends Events
 		};
 
 		$this->Template->event = $objTemplate->parse();
+
+		// Add cache headers for content elements in the future if they have not been generated.
+		if ($needsCacheHeaders)
+		{
+			$objElement = ContentModel::findUpcomingByPidAndTable($id, 'tl_calendar_events');
+
+			if ($objElement !== null)
+			{
+				while ($objElement->next())
+				{
+					$response = new Response();
+					$response->setPublic();
+					$response->setMaxAge($objElement->start - time());
+
+					System::getContainer()->get(ResponseCacheStrategyInterface::class)->add($response);
+				}
+			}
+		}
 
 		// Tag the event (see #2137)
 		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
--- a/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
@@ -14,6 +14,7 @@ namespace Contao\CoreBundle\Controller\ContentElement;
 
 use Contao\ContentModel;
 use Contao\CoreBundle\Controller\AbstractFragmentController;
+use Contao\CoreBundle\EventListener\SubrequestCacheSubscriber;
 use Contao\Template;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -22,6 +23,10 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
 {
     public function __invoke(Request $request, ContentModel $model, string $section, array $classes = null): Response
     {
+        if (null !== ($response = $this->getUpcomingResponse($request, $model))) {
+            return $response;
+        }
+
         $type = $this->getType();
         $template = $this->createTemplate($model, 'ce_'.$type);
 
@@ -35,6 +40,12 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
 
         if (null === $response) {
             $response = $template->getResponse();
+        }
+
+        $time = time();
+        if (!$response->headers->has('Cache-Control') && '' !== $model->stop && $model->stop > $time) {
+            $response->setPublic();
+            $response->setMaxAge((int) $model->stop - $time);
         }
 
         return $response;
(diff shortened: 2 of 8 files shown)


core-bundle/src/Resources/contao/models/ContentModel.php
```diff
@@ -475,6 +475,86 @@ public static function countPublishedByPidAndTable($intPid, $strParentTable, arr
...
+	/**
+	 * Find all visible content elements by their parent ID and parent table
+	 *
+	 * @param integer $intPid         The article ID
+	 * @param string  $strParentTable The parent table name
+	 * @param array   $arrOptions     An optional options array
+	 *
+	 * @return Collection|ContentModel[]|ContentModel|null A collection of models or null if there are no content elements
+	 */
+	public static function findVisibleByPidAndTable($intPid, $strParentTable, array $arrOptions=array())
```
ausi, line 488: The naming could be misinterpreted I think. How about something like `findVisibleAndUpcomingByPidAndTable`?

Toflar:
> And that's why AbstractContentElementController::addSharedMaxAgeToResponse was always wrong, because it (probably) adds a start time to an element that should not have been generated in the first place.

It's not wrong, is it? Just calling it by default is?

fritzmg:
> > And that's why AbstractContentElementController::addSharedMaxAgeToResponse was always wrong, because it (probably) adds a start time to an element that should not have been generated in the first place.
> 
> It's not wrong, is it? Just calling it by default is?

I think what @aschempp means is that a fragment with a start date in the future would never be invoked in the first place and thus the fragment would be unable to set appropriate response headers so that the cache invalidates at that point in time.

Toflar:
Which can also be handled by this method, can't it?

fritzmg:
But how? The fragment will never be instantiated (in the current implementation).

Toflar:
Which there's a PR for https://github.com/contao/contao/pull/2867. So this should be considered too then.

m-vo:
The start/stop setting is a bit different than the other ones. Most of the time it would be fine to not invoke a controller at all if it is not visible in the current scope. But with the "start time in the future" case, you need a max age header that expires when the start time is reached. 

Regarding #2867, we could also move the logic for general published state down to the controller as well, so that unpublished ESI fragments would not show. But you would not generate the ESI tag then as well and the - other way round - the cache would be invalidated when changing state, so IMHO not a problem.

>> And that's why AbstractContentElementController::addSharedMaxAgeToResponse was always wrong, because it (probably) >adds a start time to an element that should not have been generated in the first place.
>
>It's not wrong, is it? Just calling it by default is?

The point is: This method was only ever usable for ESI fragments, because all other ones wouldn't even be invoked due to the optimization 'further up'. But an ESI fragment using this would also need to mostly duplicate the logic to actually return an empty response if the start time is in the future/the stop time in the past. So not very helpful.

core-bundle/src/Resources/contao/models/ContentModel.php
m-vo: Or even `findPublishedAndUpcoming` and then we drop the original `findPublished`in 5? :see_no_evil:.

Toflar:
I see, so how do we solve it then? The current PR doesn't solve that issue.

m-vo:
> I see, so how do we solve it then? The current PR doesn't solve that issue.

Which issue are you referring to? IMHO it does solve the ESI start/stop issue as well as that upcoming inline fragments didn't set cache information.

core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -37,6 +42,12 @@ public function __invoke(Request $request, ContentModel $model, string $section,
             $response = $template->getResponse();
         }
 
+        $time = time();
+        if (!$response->headers->has('Cache-Control') && '' !== $model->stop && $model->stop > $time) {
+            $response->setPublic();
+            $response->setMaxAge((int) $model->stop - $time);
+        }
+
```
m-vo, line 50: IMHO we should either put this in a protected method (`getExpiredResponse`?) as well or also inline `getUpcomingResponse`. I don't think it's a typical use case to disable this logic because the design of Contao content elements implies a start/stop setting; so I would vote for inlining. This would also prevent a glitch if the two `time()` calls produce a different output right at the edge of the stop time. 

If needed we could add a property `protected bool $handleStartStop = true` and check for that then. Wdyt?

m-vo: Btw. we discussed this, but I forgot the outcome: Why don't we use the `ResponseCacheStrategy` to merge the max age? IMHO this should work as intended, because it's no different than adding another content element with a limiting max age to the same page. This way an element could still set cache headers (like "this response generated from an API is cacheable for 30min") and the stop setting would still work 10min before it is hit. Otherwise you always have to consider the stop time yourself when constraining the max age.

```suggestion
        $time = time();
        if($model->stop && $model->stop > $time) {
            // Merge headers to limit the max age to the stop time
            $limitResponse = new Response();
            $limitResponse->setPublic();
            $limitResponse->setMaxAge((int) $model->stop - $time);

            $strategy = new ResponseCacheStrategy();
            $strategy->add($limitResponse);
            $strategy->update($response);
        }

```

m-vo:
One thing that just came to my mind: Does an empty response generated with `new Response()` by default also prevent caching when merged? Might have to handle that here and in #4393 as well then…

aschempp:
Why would an empty response affect caching?

Regarding `addSharedMaxAgeToResponse`, it never made sense to add a start that to a **not empty** response. But that was still up to the dev, so she still had to check for start date. If the method was just blindly used on a template response, it would add a start cache header with no sense!

I'm also unsure what problem is not solved according to @Toflar ?

m-vo:
> Why would an empty response affect caching?

Because the `ResponseCacheStrategy` does not seem to handle this case:
```php
$response = new Response();
$response->setPublic();
$response->setMaxAge(100);

var_dump($response->isCacheable()); // true

$strategy = new ResponseCacheStrategy();
$strategy->add(new Response());
$strategy->update($response);

var_dump($response->isCacheable()); // false

**Comment by aschempp** [comment]
> > Why would an empty response affect caching?
> 
> Because the `ResponseCacheStrategy` does not seem to handle this case:

uhm yeah, but it **is** supposed to affect caching if there is a start/stop date, is it not?

**Comment by dmolineus** [comment]
> But how? The fragment will never be instantiated (in the current implementation).

If they are referenced by any other source (like insert tag etc.)

File: core-bundle/src/Controller/ContentElement/AbstractContentElementController.php [hunk 3]
```diff
@@ -60,5 +71,36 @@ protected function addSharedMaxAgeToResponse(Response $response, ContentModel $m
...
+            return null;
+        }
+
+        if ($this->container->get('contao.security.token_checker')->isPreviewMode()) {
+            return null;
+        }
+
+        $time = time();
+
+        if ('' !== $model->stop && $model->stop < $time) {
```
dmolineus, line 86:
```suggestion
        if ($model->invisible || ('' !== $model->stop && $model->stop < $time)) {
```

Checking `invisible` flag here would make https://github.com/contao/contao/pull/2867 superfluous.

calendar-bundle/src/Resources/contao/modules/ModuleEventReader.php
```diff
@@ -392,6 +399,24 @@ protected function compile()
 
 		$this->Template->event = $objTemplate->parse();
 
+		// Add cache headers for content elements in the future if they have not been generated.
+		if ($needsCacheHeaders)
+		{
+			$objElement = ContentModel::findUpcomingByPidAndTable($id, 'tl_calendar_events');
```
dmolineus, line 405: Imho it's enough fetch the first upcoming element (ordered by `start` date) to define the max age.

news-bundle/src/Resources/contao/modules/ModuleNews.php
```diff
@@ -231,6 +239,24 @@ protected function parseArticle($objArticle, $blnAddArchive=false, $strClass='',
 			$responseTagger->addTags(array('contao.db.tl_news.' . $objArticle->id));
 		}
 
+		// Add cache headers for content elements in the future if they have not been generated.
+		if ($needsCacheHeaders)
+		{
+			$objElement = ContentModel::findUpcomingByPidAndTable($id, 'tl_news');
```
dmolineus, line 245: Imho it's enough fetch the first upcoming element (ordered by `start` date) to define the max age.

dmolineus (commented):
After talking with @aschempp at the Hackaton I checked if the pull request #2867 is still required or can be integrated here as well.

leofeyer:
As discussed in the Contao call, we want to implement the logic only for article content elements at the moment, so the news and event related changes need to be reverted.

aschempp:
we also agreed to only implement this in the latest Contao version and not as a bug fix, right?

PR #4410, opened 2022-03-29
Use symfony/polyfill-intl-idn instead of true/punycode

leofeyer:
After waiting almost five years to get https://github.com/true/php-punycode/pull/25 merged, we can safely assume that `true/punycode` will not receive any more updates, so let‘s use the native PHP functions instead (together with the Symfony polyfill).


core-bundle/tests/Contao/IdnaTest.php
```diff
@@ -23,21 +23,13 @@ public function testEncodesUnicodeDomain(): void
         $this->assertSame('xn--fbar-5qaa.de', Idna::encode('fööbar.de'));
         $this->assertSame('', Idna::encode(''));
         $this->assertSame('', Idna::encode(sprintf('f%sbär.de', str_repeat('o', 53))));
-
-        $this->expectException(DomainOutOfBoundsException::class);
```
bytehead, line 27: Use of `DomainOutOfBoundsException` not used anymore.

bytehead (dismissed):

core-bundle/tests/Contao/IdnaTest.php
leofeyer: Fixed in ad6f9d55048fc2bff92afd61989b275430ef18fe.

bytehead (approved):

PR #4417, opened 2022-03-30
Feature HtmlAttributes in the image studio

m-vo:
This PR does two things:
- It's now possible to set link attributes in the figure builder as `HtmlAttributes`. [d078cb64de64a8a93acdb8fe83dfee3b3def8728]
- `Figure#getLinkAttributes()` returns `HtmlAttributes` instead of an array. Because we're implementing array access, the BC break for this should not be that big. [c3ac987c7626ea8216abcb2476cf9002f6af64ab]


core-bundle/src/Image/Studio/Figure.php
```diff
@@ -152,7 +153,7 @@ public function getSchemaOrgData(): array
      * Returns a key-value list of all link attributes. This excludes "href" by
      * default.
      */
-    public function getLinkAttributes(bool $includeHref = false): array
+    public function getLinkAttributes(bool $includeHref = false): HtmlAttributes
```
Toflar, line 156: That's a BC break, should be mentioned in the Upgrade.md :)

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -370,8 +371,14 @@ public function setLinkAttribute(string $attribute, ?string $value, bool $forceR
      * remove an auto-generated value from the results, set the respective
      * attribute to null.
      */
-    public function setLinkAttributes(array $attributes): self
+    public function setLinkAttributes(array|HtmlAttributes $attributes): self
```
Toflar, line 374: You sure you don't want to break here too?

Toflar: Or have it the other way around? `HtmlAttributes::createFromArray()` or so?

ausi: I’d vote for not breaking here, so still allowing the `array` type too.

Toflar: But could convert to `HtmlAttributes` internally ;)

m-vo: We need to keep this internal representation, because it also allows to mark which values of the default set will be removed after building. The `HtmlAttributes` class does not store falsy values.

core-bundle/src/Image/Studio/Figure.php
m-vo: Done in 24cf19c1941e28d16d83bb9d9d18dc51c82de999, thanks for the reminder. :slightly_smiling_face:

Toflar (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4425, opened 2022-03-30
Return the prefix-relative path when getting filesystem items from the VFS

m-vo:
Followup fix for #4397 

I'm sorry, I got one thing wrong in the last VFS PR. :see_no_evil: 

When calling `VirtualFilesystem#get()` we need to return prefix-relative paths, so that these two calls result in `FilesystemItem` objects with identical properties:

```php
// directory "files/foo" contains a single file "bar.txt", the $filesStorage is mounted to "files"

$item1 = iterator_to_array($filesStorage->listContents('foo'))[0];
$item2 = $filesStorage->get('foo/bar.txt');

var_dump($item2->getPath()); // "foo/bar.txt" (note, no prefix "files")
var_dump($item1->getPath() === $item2->getPath()); // true
```


leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4426, opened 2022-03-30
Add the service subscriber tag to the correct controller

m-vo:
The service tag `{ name: container.service_subscriber, id: contao.csrf.token_manager }` must not be set to Symfony's `AbstractController` but ours, otherwise extending from this class without specifiying the CSRF token manager in the `getSubscribedServices()` declaration would fail:

```
Service key "contao.csrf.token_manager" does not exist in the map returned by <service>.
```

(We still need to set the container there as well, so that's why I added another definition.)

### DIFF
--- a/core-bundle/src/Resources/config/controller.yml
+++ b/core-bundle/src/Resources/config/controller.yml
@@ -10,8 +10,6 @@ services:
         Contao\CoreBundle\Controller\AbstractController:
             tags:
                 - { name: container.service_subscriber, id: contao.csrf.token_manager }
-            calls:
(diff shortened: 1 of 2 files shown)


aschempp:
is the `setContainer` not inherited automatically?

m-vo:
> is the setContainer not inherited automatically?

Ah, yes, of course. Fixed now and also removed a superfluous definition from the `services.yml` that was likely forgotten when we moved the controllers to their own file.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4430, opened 2022-03-31
Update the tools

leofeyer:


core-bundle/tests/Filesystem/VirtualFilesystemTest.php
```diff
@@ -557,15 +557,13 @@ static function () use (&$handlerInvocationCount) {
         $this->assertInstanceOf(FilesystemItem::class, $fileB);
         $this->assertTrue($fileB->isFile());
 
-        /** @phpstan-ignore-next-line */
```
leofeyer, line 560: @m-vo I keeps getting stranger. Now PHPStan said _No error to ignore is reported on line 561_ and neither on line 567, but the `@phpstan-ignore-next-line` on line 544 is still necessary. 🤔 Just FYI.

m-vo: Very strange indeed. 🤔

PR #4431, opened 2022-03-31
Allow to purge the preview cache in the user profile

### LINKED ISSUE #4175: Purge the preview cache in user profile
### Affected version(s)

4.13.0

### Description

In the user profile (via user menu) an option for purging the new preview [image] cache is missing (like the existing option to clear the image cache).

leofeyer:
Fixes #4175


ausi (approved):

aschempp:
I find this label confusing, I had to look at the original issue to understand its about the image cache and not a "cache of the preview.php". Maybe we should use something like "preview image cache"?

leofeyer:
I have used the existing name, so if we decide to change it, we will have to change it everywhere.



This should be a separate pull request, though.

PR #4433, opened 2022-03-31
Always create an article if page has no layout

aschempp:
When creating new pages, we do not create articles if the current page layout does not have an article module in its structure. However, if you start a new page tree in Contao 4.13, and do not assign a layout to the root/first page, no articles are created by default. This is pretty annoying since it _forces_ to create the page layout first. This PR changes the behaviour to allow articles if no layout can be found.

Per request of @fenepedia 😉


fritzmg (approved):

ausi (approved):

Toflar (approved):

m-vo (approved):

contaoacademy:
Danke!

PR #4438, opened 2022-04-01
Fix several errors that now appear due to removed BC layers

m-vo:
Mainly fixes this error:

> An exception occurred while executing a query: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

But also includes:

-  two `array_map` calls with static methods, that weren't specified with the FQCN (maybe worked before due to our autoloader) [81761027018fee7b157836e3151d23231cb3bea7]
- a missing `DC_Table::class` [70bf6afc591e7b8baea5dab51204379c6800fae1]


listing-bundle/src/Resources/contao/modules/ModuleListing.php
```diff
@@ -141,7 +141,7 @@ protected function compile()
 		}
 
 		$strQuery .= $strWhere;
-		$objTotal = $this->Database->prepare($strQuery)->execute($varKeyword);
+		$objTotal = $this->Database->prepare($strQuery)->execute(...($varKeyword ? array($varKeyword) : array()));
```
leofeyer: This looks strange. Isn't `$varKeyword` always a string?

m-vo: Yes, but could be `''` if the placeholder does not exist.

leofeyer: Maybe `$varKeyword` should be `array()` in line 102 and `array('%' . $strFor . '%')` in line 123? Then we could use just `...$varKeyword` here.

m-vo: done in a31ec9ed1ebf4e5e4e688684688665a9a40214b2

listing-bundle/src/Resources/contao/modules/ModuleListing.php
```diff
@@ -365,7 +365,7 @@ protected function listSingleRecord($id)
 		$this->list_info = StringUtil::deserialize($this->list_info);
 		$this->list_info_where = System::getContainer()->get('contao.insert_tag.parser')->replaceInline($this->list_info_where);
 
-		$objRecord = $this->Database->prepare("SELECT " . implode(', ', array_map('Database::quoteIdentifier', StringUtil::trimsplit(',', $this->list_info))) . " FROM " . $this->list_table . " WHERE " . ($this->list_info_where ? "(" . $this->list_info_where . ") AND " : "") . Database::quoteIdentifier($this->strPk) . "=?")
+		$objRecord = $this->Database->prepare("SELECT " . implode(', ', array_map(array(Database::class, 'quoteIdentifier'), StringUtil::trimsplit(',', $this->list_info))) . " FROM " . $this->list_table . " WHERE " . ($this->list_info_where ? "(" . $this->list_info_where . ") AND " : "") . Database::quoteIdentifier($this->strPk) . "=?")
```
ausi:
```suggestion
		$objRecord = $this->Database->prepare("SELECT " . implode(', ', array_map(Database::quoteIdentifier(...), StringUtil::trimsplit(',', $this->list_info))) . " FROM " . $this->list_table . " WHERE " . ($this->list_info_where ? "(" . $this->list_info_where . ") AND " : "") . Database::quoteIdentifier($this->strPk) . "=?")
```

ausi (dismissed):

listing-bundle/src/Resources/contao/modules/ModuleListing.php
m-vo: PHP8 meets legacy code :smile:

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -1082,13 +1082,9 @@ protected static function find(array $arrOptions)
...
 
-		if (!\array_key_exists('value', $arrOptions))
-		{
-			$arrOptions['value'] = array();
-		}
-
 		$objStatement = static::preFind($objStatement);
-		$objResult = $objStatement->execute(...(\is_array($arrOptions['value']) ? $arrOptions['value'] : array($arrOptions['value'])));
+
+		$objResult = $objStatement->execute(...((array) ($arrOptions['value'] ?? array())));
```
leofeyer:
```php
if (!isset($arrOptions['value']))
{
	$arrOptions['value'] = array();
}

$objStatement = static::preFind($objStatement);
$objResult = $objStatement->execute(...((array) $arrOptions['value']));
```

It is the same as your variant, of course, but maybe less complex?

m-vo: Changed in 75031d349c9f2bd780bf15999599684be1bb9d04. This is CS right? Because it's IMHO functionally equivalent.

ausi (approved):

core-bundle/src/Resources/contao/library/Contao/Model.php
leofeyer: Yes, see my comment above.

> It is the same as your variant, of course, but maybe less complex?

leofeyer:
Thank you @m-vo.

PR #4441, opened 2022-04-01
Add a bag for file metadata

m-vo:
This PR introduces a `Contao\CoreBundle\File\MetadataBag` - a simple collection for multiple `Metadata` objects of different locales for the same entity. The bag can be constructed with a list of default/fallback locales that will be used when calling `getDefault()`, otherwise it's not much more than a convenience object wrapped around an array.

The `DbafsMetadataSubscriber` will enhance metadata to be in the form of `MetadataBag`s now. When now passing `FilesystemItem` objects to templates it gets much more convenient to access and output metadata.


core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -98,4 +105,25 @@ private function supports(AbstractDbafsMetadataEvent $event): bool
...
+        }
+
+        $locales = [LocaleUtil::formatAsLocale($page->language)];
+
+        if (null !== $page->rootFallbackLanguage) {
+            $locales[] = LocaleUtil::formatAsLocale($page->rootFallbackLanguage);
+        }
+
+        return $this->defaultLocales = array_unique(array_filter($locales));
+    }
```
m-vo: This is pretty much copied from the image studio code. Should we change the globals access to reading from the request attributes instead?

Toflar: I still don't exactly know myself. Guess @aschempp needs to clarify this once and for all and then we need to change it everywhere (if :D).

leofeyer: IMHO, we should merge this PR as is. There will be a separate PR dealing with `$GLOBALS['objPage']`.

fritzmg: I agree, we should merge as is and then later on deal with all `$GLOBALS['objPage']` usages. Now that https://github.com/contao/contao/pull/3930 is merged it could be changed to 

```diff
-$page = $GLOBALS['objPage'] ?? null;
+$page = $request->attributes->get('pageModel') ?? null;
```

for instance.

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -98,4 +107,25 @@ private function supports(AbstractDbafsMetadataEvent $event): bool
...
         return 'tl_files' === $event->getTable();
     }
+
+    private function getDefaultLocales(): array
+    {
+        if (null !== $this->defaultLocales) {
+            return $this->defaultLocales;
+        }
+
+        $page = $GLOBALS['objPage'] ?? null;
```
Toflar: I also think we should use https://github.com/contao/contao/pull/3930 now :)

Toflar (changes requested):

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
m-vo: see fb80c6d653017a787dd64119d6e3d4a86451a35a

Toflar: PHP 8 ❤️

Toflar (dismissed):

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
m-vo: So much easier :smile:

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -114,7 +119,7 @@ private function getDefaultLocales(): array
             return $this->defaultLocales;
         }
 
-        $page = $GLOBALS['objPage'] ?? null;
+        $page = $this->requestStack->getMainRequest()?->attributes?->get('pageModel');
```
leofeyer, line 122:
```suggestion
        $page = $this->requestStack->getMainRequest()?->attributes->get('pageModel');
```

If there is a request object, it will have an `attributes` property.

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -98,4 +112,25 @@ private function supports(AbstractDbafsMetadataEvent $event): bool
...
         return 'tl_files' === $event->getTable();
     }
+
+    private function getDefaultLocales(): array
+    {
+        if (null !== $this->defaultLocales) {
+            return $this->defaultLocales;
+        }
+
+        $page = $this->requestStack->getMainRequest()?->attributes?->get('pageModel');
```
leofeyer:
```suggestion
        $page = $this->requestStack->getMainRequest()?->attributes->get('pageModel');
```

If there is a request object, it will have an `attributes` property.

m-vo: True ☺️

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4443, opened 2022-04-01
Expose public URIs in the VFS

m-vo:
Our current VFS implementation does not provide a way to get a public URI for a resource. 

The files storage does not know anything about the fact, that the file `images/zebra.jpg` is publicly available (i.e. served by the web server) under `example.com/files/images/zebra.jpg`. One solution to this would be to serve all files via a route and stream the output, but this isn't feasible for all things.

The VFS now got a `generatePublicUri()` function, that delegates to the `MountManager` that again dispatches an `PublicUriEvent`. In Contao we listen for this event and - if it targets the adapter for the local filesystem - concatenate the public dir based on the configured upload path. If you are using S3, this would allow you to create a listener that returns S3 signed urls for instance and so on (see https://github.com/contao/contao/pull/4443#issuecomment-1124841692 for an example implementation).

~~Draft state to get some input. I'll polish and add tests if we're happy with the design.~~ Done
/cc @Toflar @ausi


core-bundle/src/EventListener/PublicFileUriListener.php
```diff
@@ -0,0 +1,46 @@
...
+     * provided directly by the web server.
+     */
+    public function __invoke(PublicFileUriEvent $event): void
+    {
+        if ($event->getAdapter() !== $this->localFilesAdapter) {
+            return;
+        }
+
+        // todo: should this be an absolute path?
+        $event->setPublicPath(Path::join($this->uploadDir, $event->getPath()));
```
Toflar: Yes, always go for absolute URIs.

core-bundle/src/Filesystem/FilesystemItem.php
```diff
@@ -44,20 +44,32 @@ class FilesystemItem
...
         $this->lastModified = $lastModified;
         $this->fileSize = $fileSize;
         $this->mimeType = $mimeType;
         $this->extraMetadata = $extraMetadata;
+
+        // todo: We should probably implement this as a callback instead of a
+        //       parameter, because it will never be known in advance; would
+        //       also allow to pass the options. If options are serializable,
+        //       results could still be cached here.
+        $this->publicUri = $publicUri;
```
Toflar: I don't think this is correct here. The public url should not be in this value object. Just call `$this->vfs->getPublicUri($filesystemItem)`. The `Route` also doesn't know about it's full URI. You call the router for that :)

Toflar (commented):
What's the "public file uri" for you? The URI as to how one can access it from the web? So it might even be something else than http(s), right? In which case we must have absolute URIs and "URI" is the correct term (including scheme etc.) but I'd suggest that the vfs returns a PSR-7 `UriInterface` then :)
Also, I'm unsure about the event or if this should use tagged services. As a rule of thumb to decide between the two I usually try to see if there must be at least one listener/service to make things work. Like for example the new file preview stuff by @ausi. There must be a provider and even if it's just the fallback one. If there must be one, I choose tagged services. If it's purely optional, I choose events. So in this very case here I'm not sure whether it's purely optional. Most files are gonna be publicly accessible, no? Just feels like it should be tagged services with an interface like so:

```php
interface PublicFileUriProviderInterface
{
    public function supports(FilesystemItem $item): bool;
    public function getUri(FilesystemItem $item): UriInterface;
}
```

core-bundle/src/Filesystem/FilesystemItem.php
m-vo: I would call the method(s) `generatePublicUri()` everywhere and then it would effectively just be a callback. The only reason for it being in the `FilesystemItem` would be, because the item is something that can nicely be passed around. But we can still implement this later on if needed, I suppose. :+1:

m-vo:
> What's the "public file uri" for you? The URI as to how one can access it from the web? So it might even be something else than http(s), right?

Yes, that was the idea. But maybe limiting it to `https` would make sense anyway?

> In which case we must have absolute URIs and "URI" is the correct term (including scheme etc.) but I'd suggest that the vfs returns a PSR-7 UriInterface then :)

Good idea, I like!

> Also, I'm unsure about the event or if this should use tagged services

I had exactly that thought as well after pushing :smile:. IMHO having the core implementation is not strictly necessary but it's very unlikely you won't have it. So, I'm leaning towards tagged services as well.

Toflar:
I don't think limiting to http makes any sense. If I have a company-internal protocol supported by company-internal networks - why not? :)

core-bundle/src/Filesystem/FilesystemItem.php
m-vo: I removed the usage in c9bd99b14c7f66c574a95c7a6d89ca917e2ef0d8

core-bundle/src/EventListener/PublicFileUriListener.php
m-vo: see b85752b7ad360058901be05b328270f7ce4062da

core-bundle/src/Filesystem/PublicUri/Options.php
```diff
@@ -0,0 +1,18 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Filesystem\PublicUri;
+
+final class Options
```
Toflar: Wouldn't those be provder-specific? It looks like at the end of the day it will be some kind of array with unknown options because any provider can accept and support anything. But well, I'm fine having a data object here nevertheless. Maybe we actually can standardize one or two options here which are interesting for multiple providers :) But I think we should already implement setting and getting arbitrary options. Otherwise I cannot pass them on to my own provider (I would need to extend `Options` in order to fulfil the type hint but I cannot because it's final - maybe you're looking for an interface here?).

core-bundle/src/Filesystem/PublicUri/PublicUriProviderInterface.php
```diff
@@ -0,0 +1,25 @@
...
+use League\Flysystem\FilesystemAdapter;
+use Psr\Http\Message\UriInterface;
+
+interface PublicUriProviderInterface
+{
+    /**
+     * Return a public URI for this resource or null if this provider does not
+     * support the given adapter/path/options.
+     */
+    public function getUri(FilesystemAdapter $adapter, string $adapterPath, Options $options): ?UriInterface;
```
Toflar: See my other comment, maybe `Options` should be an interface here.

Toflar (commented):
Just a conceptual issue with `Options` imho but other than that this is slowly becoming a beauty ❤️

core-bundle/src/Filesystem/PublicUri/Options.php
m-vo: Yes, definitively on my list as well. IMHO it should be an object because you can easily test against features then (`if $options instanceof Foo`, then use them). But not sure whether it should be a base class or an interface. This was only a placeholder for now. :slightly_smiling_face: 

Maybe we should try to get an S3 mount going (not saying that we need to ship it by default) and see what we actually need. Not sure if you can create a free test account there, though.

Toflar: Can just try locally using a https://min.io container ;)

core-bundle/src/Filesystem/PublicUri/PublicUriProviderInterface.php
m-vo: see 31b535260b06297e14e8fbb45f8df92ebb05d3f5

Default settings (or maybe arbitrary ones in an array) are still to be figured out. :slightly_smiling_face:

m-vo:
Here is a simplified demo implementation for S3 in the ME:

### Step 1 - Create an S3 client
`composer require league/flysystem-aws-s3-v3`

```yaml
services:
  Aws\S3\S3Client:
    arguments:
      -
        version: 'latest'
        region: '<my-region>' # e.g. eu-central-1
        credentials:
          key: '<my-key>'
          secret: '<my-secret>'
```

### Step 2 - Mount an AWS adapter
```php
<?php
declare(strict_types=1);

namespace App\ContaoManager;

use Aws\S3\S3Client;
use Contao\CoreBundle\DependencyInjection\Filesystem\FilesystemConfiguration;
use Contao\ManagerPlugin\Config\ConfigPluginInterface;
use Symfony\Component\Config\Loader\LoaderInterface;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;

class Plugin implements ConfigPluginInterface
{
    public function registerContainerConfiguration(LoaderInterface $loader, array $managerConfig): void
    {
        $compilerPass = new class implements CompilerPassInterface {
            public function process(ContainerBuilder $container)
            {
                $config = new FilesystemConfiguration($container);
                $config->mountAdapter(
                    'aws',
                    [
                        'client' => S3Client::class, // the S3 service name
                        'bucket' => '<my-bucket>',
                        // 'prefix' => 'optional/path/prefix',
                    ],
                    'files/images',
                    's3demo'
                );
            }
        };

        $loader->load(static function (ContainerBuilder $container) use ($compilerPass) {
            $container->addCompilerPass($compilerPass);
        });
    }
}
```
### Step 3 - Add a custom public path provider
```yaml
services:
  App\Filesystem\PublicUri\S3SignedUriProvider:
    arguments:
      - '@contao.filesystem.adapter.s3demo'
      - '@Aws\S3\S3Client'
      - '<my-bucket>'
    tags:
      - { name: contao.public_file_uri_provider }
```
```php
<?php
declare(strict_types=1);

namespace App\Filesystem\PublicUri;

use Aws\S3\S3Client;
use Contao\CoreBundle\Filesystem\PublicUri\OptionsInterface;
use Contao\CoreBundle\Filesystem\PublicUri\PublicUriProviderInterface;
use League\Flysystem\FilesystemAdapter;
use Psr\Http\Message\UriInterface;

class S3SignedUriProvider implements PublicUriProviderInterface
{
    public function __construct(
        private readonly FilesystemAdapter $adapter,
        private readonly S3Client $client,
        private readonly string $bucket)
    {
    }

    public function getUri(FilesystemAdapter $adapter, string $adapterPath, ?OptionsInterface $options): ?UriInterface
    {
        if ($adapter !== $this->adapter) {
            return null;
        }

        $command = $this->client->getCommand(
            'GetObject',
            [
                'Bucket' => $this->bucket,
                'Key' => $adapterPath,
            ]
        );

        $expires = $options instanceof S3Options ? $options->expires : '+20 minutes';

        return $this->client->createPresignedRequest($command, $expires)->getUri();
    }
}
```
…we can also introduce some custom options:
```php
<?php
declare(strict_types=1);

namespace App\Filesystem\PublicUri;

use Contao\CoreBundle\Filesystem\PublicUri\OptionsInterface;

class S3Options implements OptionsInterface
{
    public function __construct(public readonly string $expires)
    {
    }
}
```
### Step 4 - Use it
```php
// […]
foreach($this->filesStorage->listContents('files/images', true, VirtualFilesystemInterface::BYPASS_DBAFS)->files() as $item) {
	$item->generatePublicUri();
	$item->generatePublicUri(new S3Options('+5 minutes'));
}

```

Toflar (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4447, opened 2022-04-02
Fix simple token parser default value for unknown variables

m-vo:
We automatically detect missing variables in the simple token parser and provide them (initialized with `null`). In Contao 5, the used expression language component now got type hints and now a problem appears when using `match`, because this relies on the subject being a `string`.

This PR therefore changes the default value for unknown values to `''` (empty string) which is a small BC break that will only affect people that explicitly wrote things like `{if foo == null}bar{endif}` and had a scenario where `foo` only existed in some cases.

### DIFF
--- a/core-bundle/src/String/SimpleTokenParser.php
+++ b/core-bundle/src/String/SimpleTokenParser.php
@@ -167,7 +167,7 @@ class SimpleTokenParser implements LoggerAwareInterface
 
         try {
             return (bool) $this->expressionLanguage->evaluate($expression, $data);
-        } catch (SyntaxError|\TypeError $e) {
+        } catch (\Throwable $e) {
             throw new \InvalidArgumentException($e->getMessage(), 0, $e);
         }
     }
--- a/core-bundle/src/String/SimpleTokenParser.php
+++ b/core-bundle/src/String/SimpleTokenParser.php
@@ -160,7 +160,7 @@ class SimpleTokenParser implements LoggerAwareInterface
 
             // Define variables that weren't provided with the value 'null'
             $data = array_merge(
-                array_combine($unmatchedVariables, array_fill(0, \count($unmatchedVariables), null)),
+                array_combine($unmatchedVariables, array_fill(0, \count($unmatchedVariables), '')),
                 $data
             );
         }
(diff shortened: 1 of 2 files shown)


Toflar (commented):
Can you provide an example of the problem? Because checking for `null` seems totally legit to me...

m-vo:
See this failing test:

```
1) Contao\CoreBundle\Tests\String\SimpleTokenParserTest::testHandlesUnknownTokens with data set "Test single unknown token (regex test)" ('foo matches "/whatever/"', 'Tried to evaluate unknown sim..."foo".', false)
TypeError: Symfony\Component\ExpressionLanguage\Node\BinaryNode::evaluateMatches(): Argument #2 ($str) must be of type string, null given, called in /home/runner/work/contao/contao/core-bundle/vendor/symfony/expression-language/Node/BinaryNode.php on line 167
```

Basically if you write `foo matches "/whatever/"` and `foo` is `null`, you'll get an error.

Toflar:
Yeah, that's expected? Then you have to check for null as well?

m-vo:
> Yeah, that's expected? Then you have to check for null as well?

Adjusted in c6c90438eae5998ba8d37104439ed4cea95a2f3c. Now, we're catching the `TypeError` instead (which we should do anyway).

ausi (dismissed):

ausi (commented):
Not sure about this anymore as the prefer lowest tests fail now.

I think this behavior change was introduced in https://github.com/symfony/expression-language/commit/a7e81d14dfe475130c9fbcc3bc269c1d42afd422 

So maybe we should suggest to fix this in Symfony by removing the type hint in https://github.com/symfony/expression-language/blob/2ab2550b48ee6e88137f69967a5ded0852741549/Node/BinaryNode.php#L176

ausi:
See https://github.com/symfony/symfony/issues/45926

m-vo:
> Not sure about this anymore as the prefer lowest tests fail now.

IMHO we should raise the minimum version so that the behavior is consistent.

m-vo:
Rebased onto 4.13 - because there the error will appear as well. 

I removed the test and went for a broader catching of exceptions when evaluating. This way we are better decoupling our implementation from the exact behavior of the expression language. If we want to make sure behavior is consistent, we can raise the requirement in Contao 5. But we should IMHO only do this there, because the EL is used everywhere and the inconsistency not that big.

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #4448, opened 2022-04-03
Skip row size calculation for MyISAM

### LINKED ISSUE #4442: Handle non-InnoDB tables
### Affected version(s)

4.13.2

### Description

If there are both InnoDB and MyISAM tables, the migrations fail:

ausi:
Fixes #4442

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -485,6 +485,10 @@ class MigrateCommand extends Command
     {
         $warnings = [];
 
+        if ($table->hasOption('engine') && 'innodb' !== strtolower($table->getOption('engine'))) {
+            return $warnings;
+        }
+
         $mysqlSize = $this->rowSizeCalculator->getMysqlRowSize($table);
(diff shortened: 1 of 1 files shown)


core-bundle/src/Command/MigrateCommand.php
```diff
@@ -485,6 +485,10 @@ private function compileTableWarnings(Table $table): array
     {
         $warnings = [];
 
+        if ($table->hasOption('engine') && 'innodb' !== strtolower($table->getOption('engine'))) {
```
m-vo, line 488: Can the 'engine' option be empty and we're still dealing with anything else than innodb? 
If so should we turn the condition around?

```suggestion
        if (!$table->hasOption('engine') || 'innodb' === strtolower($table->getOption('engine'))) {
```

ausi: AFAIK the engine defaults to `InnoDB`

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #3829, opened 2021-12-16
Handle `$objPage` not being set in the InsertTags class

leofeyer:
Fixes #3827

PR #4450, opened 2022-04-03
Make replacing insert tags more granular

m-vo:
This PR reintroduces the reverted changes from #3622 for Contao 5.

With this change, automatic insert tag replacement on the rendered template output will only be applied to the legacy templates. In Twig we're using explicit replacing inside the templates (see #3606).

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
+++ b/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
@@ -141,6 +141,9 @@ trait TemplateInheritance
 			$blnDebug = System::getContainer()->getParameter('kernel.debug');
 		}
 
+		// Replace insert tags
+		$strBuffer = System::getContainer()->get('contao.insert_tag.parser')->replace($strBuffer);
+
 		// Add start and end markers in debug mode
 		if ($blnDebug)
 		{
--- a/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
+++ b/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
@@ -142,7 +142,8 @@ trait TemplateInheritance
 		}
 
 		// Replace insert tags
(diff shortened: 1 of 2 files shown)


ausi (dismissed):

fritzmg (commented):
This will again break the back end, will it not?

m-vo:
From https://github.com/contao/contao/pull/3829#issuecomment-995851949

> As discussed in the Contao call, we want to revert https://github.com/contao/contao/pull/3622 in Contao 4.13 (and add it again  in Contao 5.0). This also implies that we will not merge https://github.com/contao/contao/pull/3722 into Contao 4.13.
> 
> The changes from this PR should be merged anyway, otherwise the problem would reappear in Contao 5.0.

What exactly would you expect to break with this?

fritzmg:
See https://github.com/contao/contao/pull/3622#issuecomment-971806487

> This in general causes insert tags to be always replaced in the back end now […] - which causes various issues (e.g. the picker will not return an insert tag anymore, but the already replaced URL for a page).

core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -141,6 +141,9 @@ public function inherit()
 			$blnDebug = System::getContainer()->getParameter('kernel.debug');
 		}
 
+		// Replace insert tags
+		$strBuffer = System::getContainer()->get('contao.insert_tag.parser')->replace($strBuffer);
```
leofeyer: As discussed in the Contao call, we want to wrap this with `instanceof FrontendTemplate` so it is only executed for front end templates.

m-vo: Doen in d8a817b80c1813e9043358aee83cf25c33dd4d38

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4451, opened 2022-04-03
Quote all schema names, same as we do for inserts

### LINKED ISSUE #4420: Increase compatibility between MariaDB and MySQL 8 in contao:backup tasks
### Description

As discussed in Slack, `contao:backup:create` in a MariaDB instance creates a dump that is not importable in MySQL 8, because ` groups` is a reserved keyword in MySQL 8 and needs backticks like 
```
`groups`
```

ausi:
Fixes #4420

Our dumper already quotes all names for the `INSERT` statements.
This PR now does the same for the schema statements.


m-vo (approved):
After reviewing the code of `AbstractPlatform`, using reflection is IMHO the most sensible thing to do here. The `_keywords` property is protected and should therefore be covered by their BC promise. The alternative, creating a decorator, is way too complex for this use case.

Toflar:
Can't we just not use `$connection->quote()` and simply quote ourselves? 🤔

ausi:
> Can't we just not use `$connection->quote()` and simply quote ourselves? 🤔

I tried that first by replacing all column and table objects with quoted ones, but that was way harder and complex and in the end I didn’t reach every identifier that the mysql platform creates.

Toflar (approved):

leofeyer:
Thank you @ausi.

Issue #3575, opened 2021-10-14
Content Element HTML - change quotes after saving


berecont:
**Affected version(s)**

Contao 4.9.21, Contao 4.12.2
I didn't test other versions

Content Element HTML:
`<source src="files/layout/video/css/my_265.mp4" type='video/mp4; codecs="avc1"'> `
after saving the element it changes to:
`<source src="files/layout/video/css/my_265.mp4" type="video/mp4; codecs="avc1"">`
and notes error: Unexpected character after attribute value. Invalid character in attribute name

chanching the quotes and single quotes in their position it will work
`<source src="files/layout/video/css/my_265.mp4" type="video/mp4; codecs='avc1'"> `

ausi:
I would guess that this is a problem in the textarea output. We have to enable double encoding there so that `type="video/mp4; codecs=&quot;avc1&quot;"` gets output correctly in the browser.

ausi:
Tried today to find a fix for this but didn’t succeed 😕

Such problems no longer occur with the restructured input encoding in Contao 5.0, but I don’t know how we can fix this in 4.9 or 4.13.

leofeyer:
Guess it is a known limitation then.

PR #4453, opened 2022-04-03
Rework the input encoding

ausi:
What was previously true and this PR tries to keep (especially in terms of XSS and security):

- all `<` are encoded
- all null bytes stripped
- all `\r` bytes stripped
- all `\0` encoded to `&#92;0`
- all `{{` and `}}` are encoded
- most entities are decoded except `amp`, `lt`, `gt` and so on (but `&#x26;` still gets converted to &)
- basic entities... _this needs some more thoughts 😬_
- if decode entities is `false`, special chars `[#<>()\="']` get encoded to decimal entities like `&#35;`
- An input encoded string stored in the DB mostly passes a roundtrip through the backend (`StringUtil::specialchars` => `Input::post`)


aschempp:
To help with review, is this PR supposed to change something or just make stuff testable?

ausi:
> is this PR supposed to change something

Yes. My goal is to make the `Input` class way less complex while keeping its “security” features intact.

The biggest behavior change is IMO that we no longer decode HTML entities. So sending `fo&ouml;` via HTTP will not get converted to `foö` anymore but stays `fo&ouml;`.

How the `Input` behaves in Contao 4.13 is best described in the test cases from 85770dc704602f44d99bf94f206b60720fb84bb8.

ausi:
I distilled the different modes our `Input` class handles now down to four modes:

- `encodeAll` corresponds to `decodeEntities => false`
- `sanitizeHtml` corresponds to `allowHtml => true`
- `encodeLessThanSign` corresponds to `decodeEntities => true`
- `encodeNone` corresponds to `postRaw`

Additionally for all modes we do now:

- Ensure the value to be UTF-8
- Normalize newlines to `\n`
- Replace null bytes with the `U+FFFD` Replacement Character
- Encode `\0` to `&#92;0`
- Encode insert tags (unless we are in a backend context)

I added a `specialcharsValue` method to the widget class that better handles double encoding. Now only entities encoded by the input class are not double encoded. This fixes issues like #3575 and removes the need for “basic entities”.

Basic entities are now a DCA feature. With `['eval' => ['basicEntities' => true]]` the user still sees the basic entities, but the value stored in the database are the actual entities.

ToDo:
- [x] Create a migration for the basic entities
- [x] Remove basic entities handling everywhere else
- [ ] Decide which core fields should get basic entities enabled.

Rewiring the `Input` to the Symfony request is probably better done in another pull request.

core-bundle/src/Migration/Version500/BasicEntitiesMigration.php
```diff
@@ -0,0 +1,189 @@
...
+use Contao\CoreBundle\Migration\MigrationResult;
+use Contao\StringUtil;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @internal
+ */
+class BasicEntitiesMigration extends AbstractMigration
+{
+    private static array $targets = [
```
Toflar: Every extension will have to copy this. Can we come up with a better solution?

ausi: Like so? d9e668b5b192e861c3394061fcd1482d6b1b7da9

Toflar: Not sure if that's the way to go. We'll have the same problem in the future anyway. Especially with input/output encoding. But also recent PRs like e.g. https://github.com/contao/contao/pull/4396 provide migrations which contain hardcoded targets and we should somehow provide a common way. Not sure if always creating an abstract makes sense or if we should introduce some general concept? @contao/developers

ausi: I thought about iterating over the DCA, but modifying database columns with such a general string replacement seems problematic to me. So I think every extension should have to hard-code the columns in a migration in order to be safe. Also not replacing some `[&]`s would not be so bad IMO, but replacing it in a column that stores binary data or so would be very bad.

core-bundle/src/Resources/contao/library/Contao/Input.php
```diff
@@ -56,9 +56,52 @@ class Input
...
 	{
-		$_GET    = static::cleanKey($_GET);
-		$_POST   = static::cleanKey($_POST);
-		$_COOKIE = static::cleanKey($_COOKIE);
+		$_GET    = static::cleanKeyInternal($_GET);
+		$_POST   = static::cleanKeyInternal($_POST);
+		$_COOKIE = static::cleanKeyInternal($_COOKIE);
+	}
+
+	public static function encodeInput(array|string $value, InputEncodingMode $mode, bool $encodeInsertTags = true): array|string
```
m-vo: Should we maybe simplify this signature to `encodeInput(string $value, …): string` for the sake of static type safety?

IMHO any code using arrays can still do `array_map(fn($v) => Input::encodeInput($v), $values)` which is a reasonable overhead.

ausi: I was in too much influence from the old code it seems 😅

ausi: Done in 9ffc6fdb963ad41d73cf6a19bd36e21ae15ec11b

core-bundle/src/Migration/Version500/BasicEntitiesMigration.php
ausi: See #4505

m-vo (dismissed):
I really love this PR, finally overcoming some of the worst legacy burdens gets in reach again. :heart:

IMHO we should merge this one sooner than later, so that we have enough time to react in case some things break.

core-bundle/src/Resources/contao/widgets/SectionWizard.php
```diff
@@ -109,15 +109,15 @@ public function generate()
 		{
 			$return .= '
     <tr>
-      <td><input type="text" name="' . $this->strId . '[' . $i . '][title]" id="' . $this->strId . '_title_' . $i . '" class="tl_text" value="' . StringUtil::specialchars($this->varValue[$i]['title'] ?? '') . '"></td>
-      <td><input type="text" name="' . $this->strId . '[' . $i . '][id]" id="' . $this->strId . '_id_' . $i . '" class="tl_text" value="' . StringUtil::specialchars($this->varValue[$i]['id'] ?? '') . '"></td>';
+      <td><input type="text" name="' . $this->strId . '[' . $i . '][title]" id="' . $this->strId . '_title_' . $i . '" class="tl_text" value="' . self::specialcharsValuself::specialcharsValueself::specialcharsValuee($this->varValue[$i]['title'] ?? '') . '"></td>
```
leofeyer:
```suggestion
      <td><input type="text" name="' . $this->strId . '[' . $i . '][title]" id="' . $this->strId . '_title_' . $i . '" class="tl_text" value="' . self::specialcharsValue($this->varValue[$i]['title'] ?? '') . '"></td>
```

leofeyer (changes requested):
I‘m pretty sure that this is a search/replace error, isn‘t it?

ausi:
> I‘m pretty sure that this is a search/replace error, isn‘t it?

Yep 🙈

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4455, opened 2022-04-04
Fix Atom feed generation

### LINKED ISSUE #4439: RSS feed does not deliver well formed xml
### Affected version(s)

4.9.27

### Description

there are situations where contao delivers xml with minimized attributes

```
This page contains the following errors:
error on line 1 at column 6993: Opening and ending tag mismatch: source line 1 and picture
Below is a rendering of the page up to the first error.
```

I could fix it for my installation in `picture_default.html5:5` by adding an end tag to `<source>`

maybe this should be fixed in the core.

fritzmg:
Fixes #4439

See https://github.com/contao/contao/issues/4439#issuecomment-1087243402

This PR removes the usage of `StringUtil::toXhtml` and deprecates the method accordingly. According to the Atom spec you can set the content type of nodes like `title` and `content` to `html` - but you need to escape the markup. See https://datatracker.ietf.org/doc/html/rfc4287#section-3.1.1.2

The resulting feed works correctly in RSS readers like Mozilla Thunderbird, as far as I tested.

@leofeyer _Note:_ `StringUtil::toXhtml` was erroneously deprecated in Contao 4.13, without removing its usage. Do you need a separate PR for 4.13 that reverts that change, so that you can merge this change upstream without issues?

_Note:_ due to the deprecation in 4.13, the method `StringUtil::toXhtml` was erroneously removed in `5.x` and thus you get an exception currently in `5.x` when trying to generate an Atom RSS feed. This issue would be fixed if this PR is merged and then merged upstream into `5.x`.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Feed.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Feed.php
@@ -197,7 +197,7 @@ class Feed
 		{
 			$xml .= '<entry>';
 			$xml .= '<title>' . StringUtil::specialchars(strip_tags(StringUtil::stripInsertTags($objItem->title))) . '</title>';
-			$xml .= '<content type="html">' . preg_replace('/[\n\r]+/', ' ', StringUtil::specialchars($objItem->description)) . '</content>';
+			$xml .= '<content type="html">' . preg_replace('/[\n\r]+/', ' ', htmlspecialchars($objItem->description, ENT_XML1, 'UTF-8')) . '</content>';
 			$xml .= '<link rel="alternate" href="' . StringUtil::specialchars($objItem->link) . '" />';
 			$xml .= '<updated>' . preg_replace('/00$/', ':00', date('Y-m-d\TH:i:sO', $objItem->published)) . '</updated>';
 			$xml .= '<id>' . ($objItem->guid ?: StringUtil::specialchars($objItem->link)) . '</id>';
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/library/Contao/Feed.php
```diff
@@ -197,7 +197,7 @@ public function generateAtom()
 		{
 			$xml .= '<entry>';
 			$xml .= '<title>' . StringUtil::specialchars(strip_tags(StringUtil::stripInsertTags($objItem->title))) . '</title>';
-			$xml .= '<content type="xhtml"><div xmlns="http://www.w3.org/1999/xhtml">' . preg_replace('/[\n\r]+/', ' ', StringUtil::toXhtml($objItem->description)) . '</div></content>';
+			$xml .= '<content type="html">' . preg_replace('/[\n\r]+/', ' ', StringUtil::specialchars($objItem->description)) . '</content>';
```
ausi:
```suggestion
			$xml .= '<content type="html">' . preg_replace('/[\n\r]+/', ' ', htmlspecialchars($objItem->description, ENT_XML1, 'UTF-8')) . '</content>';
```

We need to use entity encoding for XML here I think.

We need to enable

ausi (approved):

bytehead (approved):

m-vo (approved):

leofeyer:
Thank you @fritzmg.

PR #4457, opened 2022-04-05
Run unit tests faster

ausi:
`ContaoCoreExtension::load()` takes quite some time, so I thought we could cache it 😎


m-vo (approved):

leofeyer:
Thank you @ausi.

PR #4459, opened 2022-04-05
Add feed controller

bezin:
Hi all,

this is my first take on a real feed controller: This is a rough draft that I wanted to discuss, before working on the many details still missing. 

It is a generic controller for `/share/*.xml` that dispatches an **feed event**, to which custom implementation can listen to. See `Contao\NewsBundle\EventListener\FeedListener` for a rough sketch of an implementation. 

Since Contao feeds are database managed, the event does not contain anything else except the alias of the feed and the current request. The Listener then decides, if it is supposed to return a feed for this, e.g. the news feed listener checks, if a feed with the given alias is configured in `tl_news_feed`. If yes, it created Feed and FeedItems from NewsModel and NewsFeedModel.

The generated Feed and FeedItem objects are then rendered in a proper template, so a little less HTML-in-PHP. 

Happy to hear your thoughts. Also, if you remember some related issues, that should be adressed with this PR, too, please feel free to link them here.

----
Partially inspired by [spatie/laravel-feed](https://github.com/spatie/laravel-feed) and [eko/feedbundle](https://packagist.org/packages/eko/feedbundle)

### DIFF
--- a/core-bundle/src/Controller/FeedController.php
+++ b/core-bundle/src/Controller/FeedController.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Controller;
+
+use Contao\CoreBundle\Event\ContaoCoreEvents;
+use Contao\CoreBundle\Event\FeedEvent;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Routing\Annotation\Route;
+use Twig\Environment;
+
+class FeedController
+{
+    private EventDispatcherInterface $dispatcher;
+    private Environment $twig;
+
+    // TODO: Add ATOM feed
+    // TODO: Add JSON feed (see https://www.jsonfeed.org/)
+    private array $types = [
+        'rss' => [
+            'template' => '@ContaoCore/Frontend/Feed/rss.xml.twig',
+            'content_type' => 'application/rss+xml',
+        ],
+    ];
+
+    public function __construct(EventDispatcherInterface $dispatcher, Environment $twig)
+    {
+        $this->dispatcher = $dispatcher;
+        $this->twig = $twig;
+    }
+
+    /**
+     * @Route("/share/{alias}.xml", name="feed", defaults={"_scope" = "frontend"})
+     */
+    public function __invoke(string $alias, Request $request): Response
+    {
+        $request->attributes->set('_feed', $alias);
+
+        $event = new FeedEvent($alias, $request);
+        $this->dispatcher->dispatch($event, ContaoCoreEvents::FEED);
+
+        if (null === $event->getFeed() || null === $event->getType()) {
+            throw new NotFoundHttpException();
+        }
+
+        $view = $this->types[$event->getType()] ?? null;
+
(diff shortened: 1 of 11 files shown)


fritzmg:
Interesting, I would have rather chosen a RSS feed creation library, though I guess using templates makes this way more flexible 👍

richardhj:
> Interesting, I would have rather chosen a RSS feed creation library, though I guess using templates makes this way more flexible 👍 

Would definitely simplify the PR but it also comes with the costs of a dependency. What were your considerations, @bezin?

fritzmg:
See also https://github.com/contao/core-bundle/issues/1243 for some notes on this.

Toflar (commented):
I don't think reinventing the wheel makes sense. I would definitely choose an existing RSS library. 

I also don't think it makes sense to abstract the different types of feeds in the same controller. This type of event doesn't seem like a useful abstraction to me. After all, everybody can just register their own controller. 

Or what exactly are you trying to provide with this? :)

bezin:
Hello all, thanks for your comments. 

My goal was to have a single entry point / controller for this to have everything under `/share` as this is where the RSS feeds used to be. 

As I wrote I checked some Feed Libraries but did not see much of an advantage in using these. But I will give it another try and also have a look at https://github.com/alexdebril/rss-atom-bundle mentioned in the referenced issue. 

If I understand correctly, you would prefer no base FeedController in `core-bundle`, but separate controllers in `news-bundle` and `event-bundle`?

Toflar:
I don't know if it is a use case to be able to extend a certain RSS feed with more entries. For example, the calendar bundle provides a feed for all its events. Why would it allow to dynamically add even more entries to the feed that do not exist? 
This sounds like a very specific use case. And if this use case does exist, then still the calendar bundle itself could dispatch its own event in order to support it. With exactly the details it needs.
Having a general route for multiple feed types doesn't really make a lot of sense either to me. As you can see, you added `.xml` in your route which would not make any sense for a JSON feed.
We can already dynamically add routes. So why have a single `/share/{alias}.xml` route that requires more manual matching by quering the DB in your `supports()` etc. if you could add the concrete route `/share/foobar.xml` directly? 

Don't get me wrong, I think having an abstraction does make sense! However, I would try to

- not mix up formats
- generate routes directly in order to reduce DB queries
- think about if we can use our `ResponseContext` (which needs to be extended for head tags, I know) to dynamically add said routes as `<link rel="alternate" type="application/rss+xml"...` meta tags - another reason for not mixing up formats

Imho, that would be far more useful than providing a general abstraction for all feed types.

aschempp:
I have successfully used [alexdebril/rss-atom-bundle](https://github.com/alexdebril/rss-atom-bundle) since Contao 4.0 in a custom project. It does come with some Symfony quirks that might not apply to Contao, but I'm pretty sure we can work around this.

What I would like to see is that RSS feeds are their own page type. It should **not** be hardcoded to `/share/…` but simply a page in your page tree.
 1. create a page of type "news RSS feed".
 2. select the appropriate news archives (like you previously did in the feed configuration)
 3. possibly select whether RSS or Atom should be generated.
 4. give it an alias of `share/news` if you want it to live in `/share/...`, or anything else that you might prefer

This is what page controllers are meant for. A page controller can have a custom suffix (`.rss`/`.atom`) for exactly this reason. It would also make the RSS feed bound to a domain (and possibly a language), which seems useful to me.

bezin:
~~Hello,~~

~~as you I know I am currently working on implementing this as a regular page controller. I like this so far, but have a follow-up question:~~

~~Page controllers can only have one URL Suffix. Hence I would need one controller per format. However with page controllers, this would mean three different page types `News RSS Feed`, `News JSON Feed` or `News Atom Feed`. I would rather prefer only one page type `News Feed` with a `format` drop-down as it is currently the case. Is there any way to achieve this?~~

DynamicRouteInterface FTW 🚀

PR #4474, opened 2022-04-07
Convert count() to empty array comparison

leofeyer:


m-vo:
Is this equivalent, performance wise? Because it's a completely different operation.

leofeyer:
IDK. I am trusting Rector to know what they are doing, but feel free to validate yourself.

m-vo:
Tested and confirmed that it has a better performance esp. for large arrays. :+1: 

But I still hate the looks of it, especially together with the yoda style. :see_no_evil:

m-vo:
Btw.: if you randomize the test data enough, no one really is a clear winner anymore: https://3v4l.org/TXU1V#v8.1.4 (run multiple times)

ausi:
I’d favor the following style:

If we need to make sure `$foo` is an array (and want to throw a type error if not) use:

```php
if (\count($foo))
if (!\count($foo))
```

If `$foo` is already type safe (via type hint or was defined inline) use:

```php
if ($foo)
if (!$foo)
```

They are very readable (to me at least ☺️) and have the same performance as `!== []` and `=== []`.

PR #4322, opened 2022-03-21
Use FQCN instead of aliased class name

bytehead:
Follow up for #4314

```
Symfony\Component\ErrorHandler\Error\ClassNotFoundError:
Attempted to load class "DC_Table" from the global namespace.
Did you forget a "use" statement?

  at /contao/core-bundle/src/Resources/contao/classes/Backend.php:417
  at Contao\Backend->getBackendModule('page', null)
     (/contao/core-bundle/src/Resources/contao/controllers/BackendMain.php:154)
  at Contao\BackendMain->run()
     (/contao/core-bundle/src/Controller/BackendController.php:49)
  at Contao\CoreBundle\Controller\BackendController->mainAction()
     (vendor/symfony/http-kernel/HttpKernel.php:152)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:74)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/http-kernel/Kernel.php:202)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (public/index.php:43)
```

PR #4481, opened 2022-04-07
Fix AbstractTablePickerProvider

bytehead:
Follow-up for #4322.

### DIFF
--- a/core-bundle/src/Picker/AbstractTablePickerProvider.php
+++ b/core-bundle/src/Picker/AbstractTablePickerProvider.php
@@ -137,12 +137,7 @@ abstract class AbstractTablePickerProvider implements PickerProviderInterface, D
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
-        if (!class_exists($GLOBALS['TL_DCA'][$table]['config']['dataContainer']) || !class_exists($this->getDataContainer())) {
-            trigger_deprecation('contao/core-bundle', '4.9', 'The usage of a non fully qualified class name as DataContainer name has been deprecated and will no longer work in Contao 5.0. Use the fully qualified class name instead, e.g. Contao\DC_Table::class.');
-        }
-
-        return ($this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
-                || is_a(DataContainer::getDriverForTable($table), $this->getDataContainer(), true))
+        return $this->getDataContainer() === DataContainer::getDriverForTable($table)
             && 0 !== \count($this->getModulesForTable($table));
     }
 
--- a/core-bundle/src/Picker/AbstractTablePickerProvider.php
+++ b/core-bundle/src/Picker/AbstractTablePickerProvider.php
@@ -137,6 +137,10 @@ abstract class AbstractTablePickerProvider implements PickerProviderInterface, D
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
+        if (!class_exists($GLOBALS['TL_DCA'][$table]['config']['dataContainer']) || !class_exists($this->getDataContainer())) {
+                trigger_deprecation('contao/core-bundle', '4.9', 'The usage of a non fully qualified class name as DataContainer name has been deprecated and will no longer work in Contao 5.0. Use the fully qualified class name instead, e.g. Contao\DC_Table::class.');
+        }
+
         return ($this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
                 || is_a(DataContainer::getDriverForTable($table), $this->getDataContainer(), true))
             && 0 !== \count($this->getModulesForTable($table));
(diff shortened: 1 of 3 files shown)


core-bundle/src/Picker/AbstractTablePickerProvider.php
```diff
@@ -136,7 +137,7 @@ public function supportsContext($context): bool
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
-        return $this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
+        return is_a(DataContainer::getDriverForTable($table), $this->getDataContainer(), true)
```
ausi: Should we allow both versions for better backwards compatibility?
```suggestion
        return ($this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer'] || is_a(DataContainer::getDriverForTable($table), $this->getDataContainer(), true))
```
And also update the description of the abstract `getDataContainer` on line 276?

bytehead: I don't get it how this would make a difference 🤔

ausi: If a custom provider class that extends this abstract provider returns `'Foobar'` in the `getDataContainer()` method and the `$GLOBALS['TL_DCA'][$table]['config']['dataContainer']` is set to `'Foobar'` your version would execute `is_a('DC_Foobar', 'Foobar')` and fail.

bytehead: I see your point now. Thanks!

bytehead: We may add a deprecation for this kind of style?

bytehead:
Failing tests seem unrelated.

ausi (dismissed):

core-bundle/src/Picker/AbstractTablePickerProvider.php
bytehead: I'll add one.

bytehead: Added in 0963b3dd7.

ausi (dismissed):

core-bundle/src/Picker/AbstractTablePickerProvider.php
```diff
@@ -136,7 +137,12 @@ public function supportsContext($context): bool
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
-        return $this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
+        if (!class_exists($GLOBALS['TL_DCA'][$table]['config']['dataContainer']) || !class_exists($this->getDataContainer())) {
+            trigger_deprecation('contao/core-bundle', '4.9', 'The usage of a non fully qualified class name as DataContainer name has been deprecated and will no longer work in Contao 5.0. Use the fully qualified class name instead, e.g. Contao\DC_Table::class.');
+        }
+
+        return ($this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
+                || is_a(DataContainer::getDriverForTable($table), $this->getDataContainer(), true))
```
aschempp:
```suggestion
        return ($this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
                || $this->getDataContainer() === DataContainer::getDriverForTable($table))
```

aschempp (changes requested):
I don't think we need the deprecation, since there are already deprecations when not using an FQCN for the DCA driver. Additionally, `is_a` would also match child classes, which was not previously the case (and probably is not intended).

core-bundle/src/Picker/AbstractTablePickerProvider.php
bytehead: Where are the deprecations when not using a FQCN for the DCA driver?

m-vo (dismissed):

core-bundle/src/Picker/AbstractTablePickerProvider.php
aschempp: uhm, apparently there is none, but it should be here? Otherwise people would not get notified in Contao 4? https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/classes/DataContainer.php#L1625-L1628

aschempp: Unfortunately, my patch does not work, because `DataContainer::getDriverForTable` returns `DC_Table` if the driver is `Table`, it does not return the actually FQCN 😞

aschempp (changes requested):
The problem with this implementation is that it changes the behaviour to include class inheritance, which imho is not desired. See https://github.com/bytehead/contao/pull/6 for a suggested fix that also includes `trigger_error` in `DataContainer::getDriverForTable`

aschempp (approved):

leofeyer:
Thank you @bytehead.

PR #4486, opened 2022-04-07
Make the properties readonly

leofeyer:


tools/rector/config/rector.php
```diff
@@ -35,7 +35,11 @@
             '*/src/Entity/*',
         ],
         ChangeSwitchToMatchRector::class,
-        ReadOnlyPropertyRector::class,
+        ReadOnlyPropertyRector::class => [
+            'core-bundle/src/DependencyInjection/Filesystem/FilesystemConfiguration.php',
+            'core-bundle/src/Security/Exception/LockedException.php',
+            'core-bundle/src/Security/Authentication/Token/FrontendPreviewToken.php',
```
leofeyer, line 41: I had to disable the rector for these files because it kept adding the `readonly` attribute even though the property cannot be readonly. 🙄

Toflar:
Why would this make any sense? I only see the point of having `readonly` for public properties?

ausi:
> Why would this make any sense?

Can help spot errors when modifying a property accidentally IMO.
And helps you see members where changes are expected faster.

Toflar:
Not sure if that doesn't cause a performance penalty.

ausi:
IMO, it might even enable some performance improvements by possible optimizations in the future.

leofeyer:
I‘m closing this until we have agreed on a concept.

PR #4487, opened 2022-04-07
Remove rel="nofollow" for internal pages

fritzmg:
As discussed on Slack, this removes `rel="nofollow"` for internal page links. This feature was implemented around Contao 2.7.0, probably under the assumption that it directs crawlers to "not follow" that URI. But actually:

> The [nofollow](https://html.spec.whatwg.org/multipage/links.html#link-type-nofollow) keyword indicates that the link is not endorsed by the original author or publisher of the page, or that the link to the referenced document was included primarily because of a commercial relationship between people affiliated with the two pages.

The attribute is not supposed to indicate that a page should not be crawled. Only that the target URI is not "endorsed" by the author of the originating document - and thus it can only serve as a hint for any crawler subscriber.

This PR removes the automatic output of `rel="nofollow"` in the navigation modules, if the target page has `noindex,nofollow` defined. It also removes any `rel="nofollow"` occurrences in templates for internal links, notably the mini calendar.

Links to external pages like the website of a comment's author or social media share URLs retain that attribute.

_Note:_ we briefly discussed whether we should introduce a `data-skip-search-index` so that we can still direct the search index to not index the `example.com/foobar?month=202204` URLs of the calendar for instance. However, this is not actually necessary, as `PageRegular` will automatically enable `noSearch` when these query parameters are present and thus such URLs are not put into the search index by Contao anyway.

Related: https://github.com/terminal42/escargot/pull/26

### DIFF
--- a/core-bundle/src/Resources/contao/templates/modules/mod_article.html5
+++ b/core-bundle/src/Resources/contao/templates/modules/mod_article.html5
@@ -27,16 +27,16 @@
         <!-- indexer::stop -->
         <div class="syndication">
           <?php if ($this->printButton): ?>
-            <a href="<?= $this->print ?>" class="print" title="<?= $this->printTitle ?>" onclick="window.print();return false"><?= Contao\Image::getHtml('assets/contao/images/print.svg') ?></a>
+            <a href="<?= $this->print ?>" class="print" title="<?= $this->printTitle ?>" onclick="window.print();return false" data-skip-search-index><?= Contao\Image::getHtml('assets/contao/images/print.svg') ?></a>
           <?php endif; ?>
           <?php if ($this->pdfButton): ?>
-            <a href="<?= $this->href ?>" class="pdf" title="<?= $this->pdfTitle ?>"><?= Contao\Image::getHtml('assets/contao/images/pdf.svg') ?></a>
+            <a href="<?= $this->href ?>" class="pdf" title="<?= $this->pdfTitle ?>" data-skip-search-index><?= Contao\Image::getHtml('assets/contao/images/pdf.svg') ?></a>
           <?php endif; ?>
           <?php if ($this->facebookButton): ?>
-            <a href="<?= $this->route('contao_frontend_share', ['p' => 'facebook', 'u' => $this->encUrl]) ?>" class="facebook" rel="nofollow" title="<?= $this->facebookTitle ?>" onclick="var w=window.open(this.href,'','width=640,height=380,modal=yes,left=100,top=50,location=no,menubar=no,resizable=yes,scrollbars=yes,status=no,toolbar=no');w.opener=null;return false"><?= Contao\Image::getHtml('assets/contao/images/facebook.svg') ?></a>
+            <a href="<?= $this->route('contao_frontend_share', ['p' => 'facebook', 'u' => $this->encUrl]) ?>" class="facebook" rel="nofollow" title="<?= $this->facebookTitle ?>" onclick="var w=window.open(this.href,'','width=640,height=380,modal=yes,left=100,top=50,location=no,menubar=no,resizable=yes,scrollbars=yes,status=no,toolbar=no');w.opener=null;return false" data-skip-search-index><?= Contao\Image::getHtml('assets/contao/images/facebook.svg') ?></a>
           <?php endif; ?>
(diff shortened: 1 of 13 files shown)


ausi (dismissed):

Toflar:
Can you triple-check what this means in terms of crawling? The crawler will now crawl those links too (e.g. calendar paging) which is essentially endless. So it will hit max depth at some point (if configured).

fritzmg:
> Can you triple-check what this means in terms of crawling? The crawler will now crawl those links too (e.g. calendar paging) which is essentially endless. So it will hit max depth at some point (if configured).

Will do, however I just want to note that `rel="nofollow"` was currently only implemented in the `cal_mini` template, and not the `cal_default` template for instance. And, as far as I know, the crawler won't run indefinitely anyway - unless you have infinite repeating events - as the calendar does not allow you to jump to the next/prev month, if there are no further events in the future/past.

ausi:
I just tested this and the crawler stops at a depth of 32, see https://github.com/contao/contao/blob/c045448275d899aefadbd88451c8faa516995331/core-bundle/src/Resources/contao/classes/Crawl.php#L164

> I just want to note that `rel="nofollow"` was currently only implemented in the `cal_mini` template, and not the `cal_default` template for instance.

I think this was on purpose as the `cal_mini` template is typically used in the sidebar and thus would add 32 URLs to the crawler for every single page where the calender is present. The search index would skip it correctly, crawling would just take longer in such cases, which is probably fine?

Toflar:
We can also lower the max depth. 32 never really makes any sense, does it?

ausi:
> We can also lower the max depth.

Reducing it to 16 or 10 should be fine for 99% I’d guess.

fritzmg:
Hm, idk. The amount of pages to be crawled ballooned from just 50 to over 1400 (with the default depth of 32) in my test setup. I also added a paginated element which further increased it to over 3600. Changing the max depth to 10 reduced it to 315.

fritzmg:
As discussed on Slack I added the `data-skip-search-index` attribute as a replacement for `rel="nofollow"` (where applicable).

core-bundle/src/Resources/contao/templates/modules/mod_article.html5
```diff
@@ -27,16 +27,16 @@
...
+            <a href="<?= $this->href ?>" class="pdf" title="<?= $this->pdfTitle ?>" data-skip-search-index><?= Contao\Image::getHtml('assets/contao/images/pdf.svg') ?></a>
           <?php endif; ?>
           <?php if ($this->facebookButton): ?>
-            <a href="<?= $this->route('contao_frontend_share', ['p' => 'facebook', 'u' => $this->encUrl]) ?>" class="facebook" rel="nofollow" title="<?= $this->facebookTitle ?>" onclick="var w=window.open(this.href,'','width=640,height=380,modal=yes,left=100,top=50,location=no,menubar=no,resizable=yes,scrollbars=yes,status=no,toolbar=no');w.opener=null;return false"><?= Contao\Image::getHtml('assets/contao/images/facebook.svg') ?></a>
+            <a href="<?= $this->route('contao_frontend_share', ['p' => 'facebook', 'u' => $this->encUrl]) ?>" class="facebook" rel="nofollow" title="<?= $this->facebookTitle ?>" onclick="var w=window.open(this.href,'','width=640,height=380,modal=yes,left=100,top=50,location=no,menubar=no,resizable=yes,scrollbars=yes,status=no,toolbar=no');w.opener=null;return false" data-skip-search-index><?= Contao\Image::getHtml('assets/contao/images/facebook.svg') ?></a>
```
fritzmg: _Note:_ I left `rel="nofollow"` here since the `contao_frontend_share` route will respond with a 302 redirect to the respective social network and thus not 'endorsing' that URI seems appropriate.

ausi (dismissed):

ausi (dismissed):

Toflar:
Do you think we should also default to 10 in `contao:crawl`?

fritzmg:
May be, yeah. Currently `contao:crawl` is unlimited by default.

ausi (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #4489, opened 2022-04-08
Fix the SendNewsletterEvent when sending as text only

fritzmg:
Fixes

```
Uncaught PHP Exception TypeError: "Argument 3 passed to Contao\NewsletterBundle\Event\SendNewsletterEvent::__construct() must be of the type string, null given
```

To reproduce send the newsletter as text only (no HTML).

/cc @SeverinGloeckle

### DIFF
--- a/newsletter-bundle/src/Resources/contao/classes/Newsletter.php
+++ b/newsletter-bundle/src/Resources/contao/classes/Newsletter.php
@@ -413,7 +413,7 @@ class Newsletter extends Backend
 			$objEmail->imageDir = System::getContainer()->getParameter('kernel.project_dir') . '/';
 		}
 
-		$event = (new SendNewsletterEvent($arrRecipient['email'], $objEmail->text, $objEmail->html))
+		$event = (new SendNewsletterEvent($arrRecipient['email'], $objEmail->text, $objEmail->html ?? ''))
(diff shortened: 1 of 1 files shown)


ausi (approved):

fritzmg:
Failing tests are unrelated (they will be fixed by https://github.com/contao/contao/pull/4487).

SeverinGloeckle:
Thanks for covering up that mistake.

I'd suggest to also update this line to set `$objEmail->html ` to `null` instead of an empty string if no HTML is allowed to keep in line with the nullable property:

https://github.com/contao/contao/blob/43d917c5b1e8f79c09aa9cacd189026bd7ed3f08/newsletter-bundle/src/Resources/contao/classes/Newsletter.php#L429

fritzmg:
Why do you want to change that? It should be kept as an empty string.

SeverinGloeckle:
That's the problem: It isn't an empty string if a newsletter has only plain text (`$objEmail->html` is null then, that's why the initial fix is necessary). 

However, if the event does not allow HTML, then `$objEmail->html` will be an empty string in the current implementation.

In practice, this will likely have the same effect. Nevertheless, the same behavior as for for an initial text-only newsletter would be correct (`$objEmail->html = null`).

leofeyer:
Thank you @fritzmg.

PR #4495, opened 2022-04-08
Add permission checks on global operations

Toflar:
This PR does 3 things:

- Rename the DC permissions and thus also the voter from `DCA` to `DC`. It's not voting on the "data container array" but on the "data container". So the A is incorrect.
- It also groups the permission attributes into
    * action permissions (essentially what our DC supports - `create`, `edit` etc.)
    * view permissions (essentially the operations ->  `global_operations`, `operations` and `buttons` which sometimes map to the action **but do not necessarily have to** which is why they must be separate permissions. Example: The `editheader` operation.
- It introduces permission checks in all those areas 🥳


core-bundle/src/Resources/config/services.yml
```diff
@@ -746,8 +746,8 @@ services:
             - Contao\BackendUser
             - '@?logger'
 
-    contao.security.data_container.default_dca_voter:
-        class: Contao\CoreBundle\Security\Voter\DataContainer\DefaultDcaVoter
+    contao.security.data_container.default_dc_voter:
+        class: Contao\CoreBundle\Security\Voter\DataContainer\DefaultDcVoter
```
ausi: Do we need the `DC` abbreviation here?
```suggestion
        class: Contao\CoreBundle\Security\Voter\DataContainer\DefaultDataContainerVoter
```

ausi (dismissed):
Just one question regarding the naming of the default voter class. ☺️

Toflar:
I don't mind either way :)

core-bundle/src/Resources/config/services.yml
Toflar: Done

Toflar:
Will have to work a bit more on this after discussing with @aschempp. Our operations are not strictly related to our permissions. They are for e.g. `edit`, `delete` etc. but they might not match. E.g. `editheader` needs to check for `edit` of a different table. So we need to have different permissions on the operations :)

Toflar:
This is ready for review now. Have a look at especially the `TODO` comments which I cannot really answer. I'd vote to merge this as is which means that in case you do not have access to an operation you would not have the icon anymore in the future. This would be a new UX which is totally fine for Contao 5 if we want this. If we still want to keep the UX of having disabled icons, then I'd still vote to merge this as is in order for us to be able to continue. Because what needs to follow would then be adjustments in the back end theme (e.g. automatically adding a CSS class that picks the "disabled" icon or whatever). The solution needed depends very much on the back end theme adjustments where I clearly cannot help out.

fritzmg (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -887,12 +887,24 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 
 		$return = '';
 
+		$security = System::getContainer()->get('security.helper');
+		$subject = new DataContainerSubject($strTable, $arrRow['id']);
```
ausi: I think we need a `rawurldecode()` here, see 5 lines below.
```suggestion
		$subject = new DataContainerSubject($strTable, rawurldecode($arrRow['id']));
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1134,6 +1146,9 @@ protected function generateHeaderButtons($arrRow, $strPtable)
 			return '';
 		}
 
+		$security = System::getContainer()->get('security.helper');
+		$subject = new DataContainerSubject($strPtable, $arrRow['id']);
```
ausi: Same here.
```suggestion
		$subject = new DataContainerSubject($strPtable, rawurldecode($arrRow['id']));
```

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1285,6 +1297,18 @@ public function move($blnIsAjax=false)
 			}
 		}
 
+		// Button permissions
+		$security = System::getContainer()->get('security.helper');
+		$subject = new DataContainerSubject($this->strTable); // Move view so we CAN provide an ID
```
ausi: The comment diverges from the code here. 
I think we should pass the id (which is the path) of the folder into which the user tries to upload to.

```suggestion
		$subject = new DataContainerSubject($this->strTable, $strFolder); // Move (upload) view so we CAN provide an ID
```

ausi (dismissed):
Some minor findings. Apart from that ❤️

Toflar (commented):
PR updated :)

ausi (dismissed):

fritzmg (dismissed):
> If we still want to keep the UX of having disabled icons, then I'd still vote to merge this as is in order for us to be able to continue. Because what needs to follow would then be adjustments in the back end theme (e.g. automatically adding a CSS class that picks the "disabled" icon or whatever). The solution needed depends very much on the back end theme adjustments where I clearly cannot help out.

In my opinion we should keep disabled icons (there is also a separate issue with a discussion on that already) - but yes, let's merge first and then adjust later 👍

Toflar:
Yes, I agree. But what to do if the permission is not granted is a whole different discussion than actually implementing the checks :) That's why I think we should continue here so we can finally start transforming onload_callbacks into proper voters. What to do with the result of the voter can then be addressed separately.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4233, opened 2022-03-03
Make `config.sql` optional in the DCA

leofeyer:
Fixes #4050

PR #4497, opened 2022-04-09
Revert the DcaExtractor changes

### LINKED ISSUE #4458: Tables are created if driver is not set in DCA
### Affected version(s)

4.13 - maybe also 4.9?

### Description

Since Contao 4.13 tables are also created if the driver is not set in the DCA.
According to @dmolineus, this is caused by https://github.com/contao/contao/pull/4285

Example: You define additional fields in an extension for the `tl_faq` but in a project you don't install the faq bundle. Then the table `tl_faq` is also created - although the dc driver is not defined!
see also: https://github.com/hofff/contao-social-tags/issues/25

This is a BC break, isn't it?

fritzmg:
Fixes #4458 

See https://github.com/contao/contao/issues/4458#issuecomment-1092826944

The changes introduced by https://github.com/contao/contao/pull/4233 & https://github.com/contao/contao/pull/4285 caused a lot of BC breaks that ultimately aren't worth it, just to support the original use case - i.e. being able to create a table without any keys. Which in the end is not a super valid use case anyway.

This PR reverts the relevant changes and reinstates the original behaviour completely.


m-vo:
👍 We should IMHO keep/adjust it for 5, though instead of removing.

fritzmg:
> 👍 We should IMHO keep/adjust it for 5, though instead of removing.

Well, I think first we would need to come up with a clear specification :)

fritzmg:
I also added a simple test - and fixed a PHP 8 issue in the DcaExtractor.

fritzmg:
Added more tests so that we can at least lock down the status quo.

core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -386,13 +386,13 @@ protected function createExtract()
...
-		if (is_a($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'], DC_File::class, true))
+		if (is_a($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'] ?? '', DC_File::class, true))
 		{
 			return;
 		}
 
 		// Return if the DC type is "Folder" and the DC is not database assisted
-		if (is_a($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'], DC_Folder::class, true) && empty($GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']))
+		if (is_a($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'] ?? '', DC_Folder::class, true) && empty($GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']))
 		{
```
aschempp, line 396: Shouldn't this use the DataContainer method to find the class from string?

bytehead (dismissed):

leofeyer:
@fritzmg Can you please fix the CI chain?

fritzmg:
@leofeyer I fixed the Code style - however there are still unhandled self deprecations left in the tests which are not related to this PR. These need to be fixed elsewhere.

fritzmg:
The self deprecations were caused by https://github.com/contao/contao/pull/4481 - I have added a fix in this PR (7faacb1).

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4500, opened 2022-04-12
Fix header fields for blank option

rabauss:
While debugging the contao leads I struggled with the missing blank option for the header fields.
The default for the blank option in the database is `0` but this is not included in the options callback.

Of course we could also only output `''` instead of `'---'`

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -4294,13 +4294,13 @@ class DC_Table extends DataContainer implements \listable, \editable
 					$this->import($strClass);
 					$options_callback = $this->$strClass->$strMethod($this);
 
-					$_v = $options_callback[$_v] ?? '---';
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4294,13 +4294,13 @@ protected function parentView()
 					$this->import($strClass);
 					$options_callback = $this->$strClass->$strMethod($this);
 
-					$_v = $options_callback[$_v];
+					$_v = $options_callback[$_v] ?? '---';
```
```suggestion
					$_v = $options_callback[$_v] ?? '-';
```

The fallback value seems to be `'-'` in most conditions above, so we should keep this consistent I think?

rabauss: okay, I solved that although I thought the blank option is `---` in some cases

ausi (approved):

fritzmg (approved):

leofeyer:
Thank you @rabauss.

PR #4501, opened 2022-04-12
Fix undefined sorting mode for group header

rabauss:
Another small fix for the sorting mode.

I missed that duplicate in my first fix in 
https://github.com/contao/contao/blob/b986083e001cfe4ce79f4ff4b2e5fc9858824a91/core-bundle/src/Resources/contao/drivers/DC_Table.php#L4968


ausi (approved):

leofeyer:
Thank you @rabauss.

PR #4510, opened 2022-04-14
Rewire the Input and Environment classes to use the Symfony request

ausi:
Based on #4453 (needs to be rebased once it’s merged)


calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
```diff
@@ -70,7 +70,7 @@ public function generate()
 		}
 
 		// Show the event reader if an item has been selected
-		if ($this->cal_readerModule > 0 && isset($_GET['auto_item']))
+		if ($this->cal_readerModule > 0 && Input::get('auto_item') !== null)
```
aschempp, line 73: This was probably used to not mark `auto_item` as _being used_, how will that work in the future?

ausi: As it is only executed if `cal_readerModule` is set, this should not be a problem here.

To ensure backwards compatibility we could add the following line here:

```php
Input::setUnusedRouteParameters(['auto_item', ...Input::getUnusedRouteParameters()]);
```

WDYT?

leofeyer: I don‘t think that we are using `isset($_GET[])` to work around unused route parameters here. The pattern is used quite often throughout the code base, e.g. `isset($_GET['table']) && \in_array(Input::get('table'), $arrTables)` in the back end (where there are no unused route parameters).

leofeyer: BTW, you don‘t need the workaround above. You can just set the third argument:

https://github.com/contao/contao/blob/6c5b17cf6a337ed1ea4a8091cba9f3cb29774fcc/core-bundle/src/Resources/contao/library/Contao/Input.php#L125

ausi: I deprecated the use of `$blnKeepUnused` in this PR as it has no sensible use case IMO and using `setUnusedRouteParameters()` directly makes the code better understandable to me.

leofeyer: I see. 😄

core-bundle/src/Controller/AbstractBackendController.php
```diff
@@ -34,7 +34,7 @@ public function __invoke(): array
                 $this->Template->version = $GLOBALS['TL_LANG']['MSC']['version'].' '.ContaoCoreBundle::getVersion();
 
                 // Handle ajax request
-                if ($_POST && Environment::get('isAjaxRequest')) {
+                if (Input::postKeys() && Environment::get('isAjaxRequest')) {
```
leofeyer: How about adding an `isPost()` method that only checks if the current request is a POST request? Most of the time, that‘s all we want to know and the `postKeys()` method seems to do a lot more.

ausi: Done in 9ef89021221b951c63fa6b0b26972c189a8e51d0

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -177,161 +174,78 @@ protected static function documentRoot()
...
-
-			// Language only, e.g. en or fr (see #29)
-			if (preg_match('/^[a-z]{2}$/', $locale))
-			{
-				$arrLanguages[] = $locale;
-			}
-		}
-
-		return \array_slice(array_unique($arrLanguages), 0, 8);
+		return str_replace('_', '-', $request->getLanguages());
```
leofeyer: Previously, we only returned the first eight languages because the header can be manipulated to be a really long string. Should we restore this behavior?

ausi: As the core never uses it, it might make sense to restore the original behavior for better BC :+1: I will update the PR

ausi: Done in e8d61e853ba7348c90637044d17b097f129551b3

core-bundle/src/Resources/contao/modules/ModuleTwoFactor.php
```diff
@@ -46,7 +46,7 @@ protected function compile()
 		$user = BackendUser::getInstance();
 
 		// Inform the user if 2FA is enforced
-		if (!$user->useTwoFactor && empty($_GET['act']) && $container->getParameter('contao.security.two_factor.enforce_backend'))
+		if (!$user->useTwoFactor && !Input::get('act') && $container->getParameter('contao.security.two_factor.enforce_backend'))
```
leofeyer, line 49:
```suggestion
		if (!$user->useTwoFactor && Input::get('act') === null && $container->getParameter('contao.security.two_factor.enforce_backend'))
```

Just a nitpick because this is how you have done it in the other files.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
```diff
@@ -56,19 +56,12 @@ public function generate()
 	 */
 	protected function compile()
 	{
-		// Mark the x and y parameter as used (see #4277)
-		if (isset($_GET['x']))
-		{
-			Input::get('x');
-			Input::get('y');
```
leofeyer, line 63: Using a `<form>` that sends a GET request will also add the `&x=` and `&y=` parameters to the query string. Did you remove this because we no longer check variables in the query string but only in the path?

core-bundle/src/Resources/contao/modules/ModuleTwoFactor.php
ausi: I replaced `isset()`s with null checks and `empty()`s with boolen checks to keep the behavior the same.

core-bundle/src/Resources/contao/modules/ModuleSearch.php
ausi: Yes, we only check unused route parameters, so marking the query parameters as used is no longer neccesary.
We could also remove this in Contao 4.13, not sure if worth the effort?

core-bundle/src/Resources/contao/modules/ModuleTwoFactor.php
leofeyer: I see. 👍

core-bundle/src/Resources/contao/modules/ModuleSearch.php
leofeyer: Not worth the effort IMHO.

leofeyer (approved):

leofeyer:
Thank you @ausi. Great work!

fritzmg:
> I deprecated the use of $blnKeepUnused in this PR as it has no sensible use case IMO

I use that all the time in order to check whether an `auto_item` is present, without marking the parameter used (so that the 404 page is still generated if not used) - and lots of other extensions do so too. How do we need to do that in the future?

ausi:
> How do we need to do that in the future?

I think `in_array('auto_item', Input::getKeys())` might make sense.
And maybe there is a way to get the route parameters from the request object?

> I use that all the time in order to check whether an `auto_item` is present, without marking the parameter used (so that the 404 page is still generated if not used) - and lots of other extensions do so too.

I wasn’t aware of that 😕
I’d still like to deprecate it if we can find a better alternative, something like `Input::hasGet()` maybe?

fritzmg:
But we still need to be able to to retrieve the value without marking it as used.

ausi:
Still possible, but it is more verbose now:

```php
$unusedRouteParameters = Input::getUnusedRouteParameters();
… Input::get('auto_item') …
Input::setUnusedRouteParameters($unusedRouteParameters);
```

Making that more verbose was acually a goal of marking the parameter as deprecated, as I think most developers (including myself) don’t recognize that the same happens with using `Input::get('auto_item', false, true)`.

PR #4514, opened 2022-04-19
Do not use head and attribute data in search context

aschempp:
The search result module currently shows a context of the page where the search topic was found. It is confusing for a user to see context data that does not actually appear on the page (e.g. link `title` or image `alt` attributes).
This PR is a simple approach of splitting "meta" and body data. It will result in a new index of all pages (because the checksum changes), but it is supposed to seamlessly work with the existing search data until reindexed.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -173,7 +173,7 @@ class Search
 		$strBody = strip_tags($strBody);
 
 		// Put everything together
-		$arrSet['text'] = $strBody . "\n" . $arrData['title'] . ' ' . $arrData['description'] . ' ' . $arrData['keywords'];
+		$arrSet['text'] = $strBody . ' ' . $arrData['description'] . "\n" . $arrData['title'] . "\n" . $arrData['keywords'];
 		$arrSet['text'] = trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($arrSet['text'])));
 
 		// Calculate the checksum
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -173,7 +173,7 @@ class Search
 		$strBody = strip_tags($strBody);
 
 		// Put everything together
-		$arrSet['text'] = $arrData['title'] . ' ' . $arrData['description'] . ' ' . $strBody . ' ' . $arrData['keywords'];
+		$arrSet['text'] = $strBody . "\n" . $arrData['title'] . ' ' . $arrData['description'] . ' ' . $arrData['keywords'];
 		$arrSet['text'] = trim(preg_replace('/ +/', ' ', StringUtil::decodeEntities($arrSet['text'])));
 
 		// Calculate the checksum
--- a/core-bundle/src/Resources/contao/modules/ModuleSearch.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleSearch.php
@@ -245,7 +245,7 @@ class ModuleSearch extends Module
 				$objTemplate->unit = $GLOBALS['TL_LANG']['UNITS'][1];
 
 				$arrContext = array();
-				$strText = StringUtil::stripInsertTags($arrResult[$i]['text']);
+				$strText = StringUtil::stripInsertTags(strtok($arrResult[$i]['text'], "\n"));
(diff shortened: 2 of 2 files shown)


Toflar:
I can definitely approve the idea of not considering the meta data in the context (would solve a problem I have as well) but I don't know about the implementation. Will leave that decision up to @ausi who worked a lot on the search lately.

Toflar:
I can definitely approve the idea of not considering the meta data in the context (would solve a problem I have as well) but I don't know about the implementation. Will leave that decision up to @ausi who worked a lot on the search lately.

qzminski:
If the keywords are not found in the context, then no context will be displayed in the frontend. Thus, I would suggest the following addition:

```diff
// Shorten the context and highlight all keywords
if (!empty($arrContext))
{
	$objTemplate->context = trim(StringUtil::substrHtml(implode('…', $arrContext), $totalLength));
	$objTemplate->context = preg_replace('((?<=^|\PL|\p{Hiragana}|\p{Katakana}|\p{Han}|\p{Myanmar}|\p{Khmer}|\p{Lao}|\p{Thai}|\p{Tibetan})(' . implode('|', array_map('preg_quote', $arrMatches)) . ')(?=\PL|\p{Hiragana}|\p{Katakana}|\p{Han}|\p{Myanmar}|\p{Khmer}|\p{Lao}|\p{Thai}|\p{Tibetan}|$))ui', '<mark class="highlight">$1</mark>', $objTemplate->context);

	$objTemplate->hasContext = true;
} 
+else
+{
+	// Fall back to the whole text if no keywords are found
+	$objTemplate->context = trim(StringUtil::substrHtml($strText, $totalLength));
+	$objTemplate->hasContext = true;
+}
```

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -173,7 +173,7 @@ public static function indexPage($arrData)
 		$strBody = strip_tags($strBody);
 
 		// Put everything together
-		$arrSet['text'] = $arrData['title'] . ' ' . $arrData['description'] . ' ' . $strBody . ' ' . $arrData['keywords'];
+		$arrSet['text'] = $strBody . "\n" . $arrData['title'] . ' ' . $arrData['description'] . ' ' . $arrData['keywords'];
```
ausi:
```suggestion
		$arrSet['text'] = $strBody . ' ' . $arrData['description'] . "\n" . $arrData['title'] . "\n" . $arrData['keywords'];
```
As google also sometimes shows the description in search results, I think we can do that too. The title will always be shown, so it does not need to be in the context I agree.

ausi:
```suggestion
		$arrSet['text'] = $strBody . ' ' . $arrData['description'] . "\n" . $arrData['title'] . "\n" . $arrData['keywords'];
```
As google also sometimes shows the description in search results, I think we can do that too. The title will always be shown, so it does not need to be in the context I agree.

aschempp: did you intentionally move title and keywords to separate lines?

aschempp:
> If the keywords are not found in the context, then no context will be displayed in the frontend. Thus, I would suggest the following addition:

@qzminski I think that's not an issue and a supported use case. The template already has an if-check whether there is a context.

core-bundle/src/Resources/contao/library/Contao/Search.php
ausi: Yes so that the structure of the column is `[text]\n[title]\n[keywords]` as it might be useful to be able to distinguish between title and keywords IMO.

qzminski:
> @qzminski I think that's not an issue and a supported use case. The template already has an if-check whether there is a context.

Yes, the problem is that if there is no context, nothing would be displayed: https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/templates/search/search_default.html5#L10-L12

I think we should always display some text and if context is missing, it should be just the indexed content.

fritzmg:
> As google also sometimes shows the description in search results, I think we can do that too.

I agree with this as well.

Toflar:
I also agree with that.

Toflar (changes requested):
As at least 3 people said the context should be shown anyway.

aschempp:
We agreed to show the page description, which was added?

fritzmg (approved):

Toflar (approved):

ausi (approved):

PR #4517, opened 2022-04-19
Fix miscellaneous minor issues

leofeyer:


core-bundle/src/Resources/contao/modules/ModulePersonalData.php
```diff
@@ -121,6 +121,9 @@ protected function compile()
 		$objVersions->setEditUrl('contao/main.php?do=member&act=edit&id=%s&rt=1');
 		$objVersions->initialize();
 
+		$arrSubmitted = array();
+		$arrFiles = array();
+
```
leofeyer, line 126: This one puzzles me, because it means that the `updatePersonalData` hook cannot have worked. `$arrSubmitted` only contained the last field before the change. 🤔

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
```diff
@@ -29,7 +29,7 @@ public function generate(): string
 		$rootPages = $framework->getAdapter(PageModel::class)->findByType('root', array('order' => 'sorting'));
 		$wizard = StringUtil::deserialize($this->wizard);
 
-		$this->blankOptionLabel = $translator->trans(sprintf('tl_module.%sBlankOptionLabel', $this->name), array(), 'contao_module');
+		$this->blankOptionLabel = $translator->trans(sprintf('tl_module.%sBlankOptionLabel', $this->name), array(), 'contao_tl_module');
```
leofeyer, line 32: This cannot have worked, either. We might have to backport the fix.

core-bundle/src/Resources/contao/modules/ModulePersonalData.php
Toflar: Might be related to https://github.com/contao/contao/pull/4364 which is very new :)

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
Toflar: Yes, that should go to 4.13 and then get merged upstream :)

bytehead: I can provide a fix PR for 4.13 if you want :)

bytehead: This works as intended in my 4.13 (`dump($this->blankOptionLabel);`) on line 34:

Bildschirmfoto 2022-04-20 um 07 06 55

bytehead (approved):

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
leofeyer: I wonder why, because `contao_module` will try to load a `module.xlf` file which definitely does not exist. Maybe the `tl_modules.xlf` file has been loaded beforehand by another module?

core-bundle/src/Resources/contao/modules/ModulePersonalData.php
leofeyer: Ah, you are right. Never mind then. 👍

core-bundle/src/Resources/contao/widgets/RootPageDependentSelect.php
bytehead: Is there somewhere a magic fallback if no translation is found on the given domain?

leofeyer: No, our translator simply reads from `$GLOBALS['TL_LANG']` if the domain begins with `contao_`. And apparently `$GLOBALS['TL_LANG']['tl_module']` exists already, therefore we did not notice this. Fixed in #4521 now.

PR #4518, opened 2022-04-20
Carry out the planned renaming

leofeyer:
This resolves several `@todo`s for Contao 5.0.


aschempp:
Thinking how Symfony would do this, should we really just rename the classes? It might make it impossible for an extension to be compatible with Contao 4 and 5. We _could_ add mapping classes for the old names, but I feel we don't really have/follow that strategy?

ausi:
> We _could_ add mapping classes for the old names

This should be done by the extension IMO, and only if it is really neccesary.

ausi (dismissed):

ausi (approved):

PR #4519, opened 2022-04-20
Fix ESI fragments with inline attributes

aschempp:
https://github.com/contao/contao/pull/4256 added support for inline-rendering of fragments, if the passed-in model is not in the registry. Unfortunately, we missed the fact that e.g. the `child_record_callback` in `tl_content` uses the existing data to create a fake model and render the element. If an element now uses the ESI renderer, it cannot be displayed in the back end (`AbstractSurrogateFragmentRenderer` will throw an exception to only use scalar values).

This PR automatically uses our forward renderer (the default for fragments) if the URL contains non-scalar attribtues.


Toflar (changes requested):
Requires a test in `FragmentHandlerTest` but otherwise looks good 👍

core-bundle/src/Fragment/FragmentHandler.php
```diff
@@ -113,4 +117,19 @@ private function hasGlobalPageObject(): bool
     {
         return isset($GLOBALS['objPage']) && $GLOBALS['objPage'] instanceof PageModel;
     }
+
+    private function containsNonScalars(array $values): bool
+    {
+        foreach ($values as $value) {
+            if (\is_array($value)) {
+                return $this->containsNonScalars($value);
```
m-vo, line 125: If `$values` looks like this `[], object`, this function will return `false`, although the array contains a "non scalar".

Did you mean to write this?
```suggestion
                if ($this->containsNonScalars($value)) {
                      return true;
                }
```

m-vo (changes requested):

core-bundle/src/Fragment/FragmentHandler.php
aschempp: well that's a copy of Symfony: https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/HttpKernel/Fragment/AbstractSurrogateFragmentRenderer.php#L89-L100
we should behave like they do – but maybe it is worth fixing in Sf?

m-vo: True, IMHO that's wrong in Symfony as well. I opened an issue: https://github.com/symfony/symfony/issues/46152

m-vo:
We should monitor https://github.com/symfony/symfony/issues/46152 and adjust our implementation as well if there is any movement. Unblocking this for now.

m-vo (dismissed):

Toflar:
Ping @aschempp - I can approve once we have a test :)

Toflar (approved):

PR #4522, opened 2022-04-20
Add the `Controller::$Template` property again

leofeyer:
Finally implements #3822


aschempp:
For the use case of people accessing the property (e.g. as described in https://github.com/contao/contao/issues/3894), shouldn't the property be public?

bytehead:
See https://github.com/contao/contao/issues/4086#issuecomment-1035109547

bytehead (changes requested):
I think https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/pages/PageRegular.php#L24 needs to be removed as well

ausi:
> For the use case of people accessing the property (e.g. as described in #3894), shouldn't the property be public?

Not allowing the public access is the desired goal AFAIR. We postponed it to Contao 5.0 because it is a BC break.

aschempp:
> Not allowing the public access is the desired goal AFAIR.

Why is it the goal to limit flexibility of Contao? We intentionally pass module instances and a lot of other stuff to hooks, so people have access to the stuff they need. Now we're removing things we _know_ they need? 😕

fritzmg:
@aschempp the data is available through the `parseTemplate` hook.

fritzmg:
And just to reiterate this here again: this discussion keeps coming up because we have an inconsistency here. For (legacy) content elements and front end modules the access to the `Template` variable is handled through `__set` and `__get` and is always public. `PageRegular` does not use the magic methods and the usage of `$this->Template` in it was likely an oversight as we are used to it from the legacy content elements and front end modules.

leofeyer:
> content elements and front end modules the access to the `Template` variable is handled through `__set` and `__get` and is always public.

And that is wrong, too. The magic setter/getter writes into `$this->arrData`, which should not include the template object. Therefore it should be a public property IMHO.

ausi:
> > content elements and front end modules the access to the `Template` variable is handled through `__set` and `__get` and is always public.
> 
> And that is wrong, too. The magic setter/getter writes into `$this->arrData`, which should not include the template object. Therefore it should be a public property IMHO.

I don’t agree I think. Removing `Template` from `$this->arrData` is a behavior change that introduces unnecessary BC breaks. If I remember correctly, the protected `$Template` should go into the `PageRegular` class. But I’d have to research that again.

leofeyer:
> If I remember correctly, the protected `$Template` should go into the `PageRegular` class. But I’d have to research that again.

I am pretty sure that this is wrong. The code already contained `@todo`s with what needed to be done and according to them, the property should be added to the `Controller` class.

leofeyer:
I have added the "up for discussion" label, so this issue does not end up with dozens of comments again. I think the main question is whether accessing the template object as `$page->Template->foo = 'bar'` is a supported use case or not. If yes, we either need a public property or a getter/setter; and if not, we can add a protected property.

leofeyer:
As discussed in the Contao call, we want to keep the `protected $Template` property in the `BackendMain` class, remove it again from the `Controller` class and add `public $Template` to the `PageRegular` class.

leofeyer:
Adjusted in a955cd1994e7b26ad754ae5e2201f58e7f3fbb52.

m-vo (approved):

ausi (approved):

fritzmg (approved):

PR #4523, opened 2022-04-20
Handle predefined image sizes when validating the ImageSize widget

qzminski:
The `Field "Image size" must not be empty!` error is thrown if the field is mandatory and one tries to select the predefined image size (the one from the app configuration).

/cc @Toflar

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/ImageSize.php
+++ b/core-bundle/src/Resources/contao/widgets/ImageSize.php
@@ -75,7 +75,7 @@ class ImageSize extends Widget
 	{
 		$varInput[2] = preg_replace('/[^a-z0-9_]+/', '', $varInput[2] ?? '');
 
-		if (!is_numeric($varInput[2]))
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/widgets/ImageSize.php
```diff
@@ -103,6 +103,10 @@ protected function validator($varInput)
 			$varInput[0] = parent::validator($varInput[0]);
 			$varInput[1] = parent::validator($varInput[1]);
 		}
+		elseif ($this->mandatory)
+		{
+			$this->mandatory = (bool) $varInput[2];
+		}
```
ausi: What does this code do? 
The issue seems to be fixed by the change in line 78, these lines here do not seem to be needed?

qzminski: You're right, I have changed it.

Toflar (approved):

leofeyer:
Thank you @qzminski.

PR #4524, opened 2022-04-20
Remove the first/last/even/odd/col_/row_ CSS classes

### LINKED ISSUE #2868: Remove the "first", "even", "odd" and "last" CSS classes

leofeyer:
Fixes #2868


m-vo:
Is this a duplicate of https://github.com/contao/contao/pull/4178?

Also: Do we really want to change this for the old modules? We won't have any of these classes in the new content elements as fragment controllers and we want people to migrate to these anyway. Maybe we should keep the old implementation as is and simply deprecate it.

leofeyer:
No, the classes are already deprecated in Contao 4, so let‘s remove them now.

leofeyer:
> Is this a duplicate of #4178?

No, rather a more complete version. See https://github.com/contao/contao/pull/4178#pullrequestreview-948253198.

leofeyer:
Closed in favor of #4178.

PR #4527, opened 2022-04-21
Deprecate Contao\Request

Toflar:
It only supports HTTP 1.0 and there's just no point in keeping it in our code base any longer.


bytehead (approved):

aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #4533, opened 2022-04-22
Fix an error when unpacking an associative array of model search criteria values

qzminski:
In Contao 4.13 we do get an error when using Isotope product filtering:

164628775-6b76e30f-f1cf-436f-bd68-47cbbfa5d939

After quick discussion with @aschempp we agreed it is a BC break. This PR should fix that.


ausi (approved):

leofeyer:
Thank you @qzminski.

PR #4534, opened 2022-04-22
Fix the FAQ page module throwing a warning in PHP 8 if author could not be fetched

qzminski:
I have basically copied the logic from the FAQ reader module: https://github.com/contao/contao/blob/5.x/faq-bundle/src/Resources/contao/modules/ModuleFaqReader.php#L155-L163


m-vo (approved):

leofeyer:
Thank you @qzminski.

PR #4535, opened 2022-04-22
Fix the registration module throwing a warning in PHP 8 if captcha is disabled

qzminski:
I took the solution from https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/modules/ModuleRegistration.php#L306


leofeyer (approved):

leofeyer:
Thank you @qzminski.

PR #4536, opened 2022-04-22
Allow partial mocks and row()/setRow() when mocking classes with properties

leofeyer:
This PR allows getting rid of `(new \ReflectionClass(PageModel::class))->newInstanceWithoutConstructor();` in our unit tests and using `$this->mockClassWithProperties()` instead.

Naming and PHP 8 syntax can still be discussed. 😄

### DIFF
--- a/core-bundle/tests/Cache/EntityCacheTagsTest.php
+++ b/core-bundle/tests/Cache/EntityCacheTagsTest.php
@@ -93,8 +93,7 @@ class EntityCacheTagsTest extends DoctrineTestCase
     {
         $entityCacheTags = $this->getEntityCacheTags();
 
-        /** @var PageModel $page */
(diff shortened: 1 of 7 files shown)


ausi (dismissed):

Toflar (dismissed):

m-vo (dismissed):
I'm fine with the changes. :+1: 

Currently our models are a mixture between POPOs and some setup/registry logic. On the long run, we should IMHO get rid of the latter and then not mock them at all.

aschempp (dismissed):

leofeyer:
I have renamed the parameter from `$preserve` to `$except` in accordance with PHPUnit‘s `setMethodsExcept()` method.

m-vo (approved):

PR #4537, opened 2022-04-23
Allow TinyMCE 6

leofeyer:
This PR allows to install TinyMCE 6.


core-bundle/src/Resources/contao/templates/backend/be_tinyMCE.html5
```diff
@@ -37,15 +38,15 @@ window.tinymce && tinymce.init({
         }
       });
     },
-    file_picker_types: <?= json_encode($this->fileBrowserTypes) ?>,
+    file_picker_types: '<?= implode(' ', $this->fileBrowserTypes) ?>',
```
m-vo: Was this wrong ever since? 

Judging by the [docs](https://www.tiny.cloud/docs-4x/configure/file-image-upload/#file_picker_types ), in v4 this should also have been "types of file pickers you need by a space or comma separated list of type names".

m-vo (dismissed):

core-bundle/src/Resources/contao/templates/backend/be_tinyMCE.html5
leofeyer: Ah, good find. Then we can make `$this->fileBrowserTypes` a space separated string and it will be compatible with both TinyMCE 5 and 6. 🎉

leofeyer: Adjusted in afb35bc10f026315e3cc3bcae5be8c2013b00d4a.

composer.json
```diff
@@ -50,7 +50,7 @@
         "contao-components/swipe": "^2.0.3",
         "contao-components/tablesort": "^4.0",
         "contao-components/tablesorter": "^2.1",
-        "contao-components/tinymce4": "^4.7 || ^5.0",
+        "contao-components/tinymce4": "^4.7 || ^5.0 || ^6.0",
```
```suggestion
        "contao-components/tinymce4": "^5.0 || ^6.0",
```
Should we drop tinyMCE 4 support in Contao 5.0?

core-bundle/composer.json
```diff
@@ -50,7 +50,7 @@
         "contao-components/swipe": "^2.0.3",
         "contao-components/tablesort": "^4.0",
         "contao-components/tablesorter": "^2.1",
-        "contao-components/tinymce4": "^4.7 || ^5.0",
+        "contao-components/tinymce4": "^4.7 || ^5.0 || ^6.0",
```
leofeyer:
```suggestion
        "contao-components/tinymce4": "^5.0 || ^6.0",
```

composer.json
leofeyer: Yes, good idea. Changed in 7d18822c062bab1b499a7cce26f219b24cf5d4c6.

ausi (approved):

PR #4545, opened 2022-04-24
Remove the deprecated Model.php stuff

m-vo:
I also removed the `$arrClassNames` property where class names where cached, because with the current code it doesn't provide any speedup. Should we annotate this with `@deprecated` in 4.13?

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Model.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Model.php
@@ -729,8 +729,8 @@ abstract class Model
 	/**
 	 * Find a single record by its primary key
 	 *
-	 * @param mixed $varValue   The property value
-	 * @param array $arrOptions An optional options array
+	 * @param int|string $varValue   The property value
+	 * @param array      $arrOptions An optional options array
 	 *
 	 * @return static The model or null if the result is empty
 	 */
@@ -738,7 +738,7 @@ abstract class Model
 	{
 		if ($varValue === null)
 		{
-			throw new \TypeError('Model::findByPk(): Argument #1 ($varValue) must not be of type null.');
+			throw new \TypeError('Model::findByPk(): Argument #1 ($varValue) must be of type int|string, got null');
 		}
 
 		// Try to load from the registry
@@ -920,6 +920,11 @@ abstract class Model
 		if (\count($arrColumn) == 1 && ($arrColumn[0] === static::getPk() || \in_array($arrColumn[0], static::getUniqueFields())))
 		{
 			$blnModel = true;
+
+			if ($varValue === null && $arrColumn[0] === static::getPk())
+			{
+				throw new \TypeError('Model::findBy(): Argument #2 ($varValue) must be of type int|string when querying for the primary key, got null');
+			}
 		}
 
 		$arrOptions = array_merge
--- a/core-bundle/src/Resources/contao/library/Contao/Model.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Model.php
@@ -1177,12 +1177,12 @@ abstract class Model
 	 */
 	public static function getClassFromTable($strTable)
 	{
-		if (null === ($strClass = $GLOBALS['TL_MODELS'][$strTable] ?? null))
+		if (!isset($GLOBALS['TL_MODELS'][$strTable]))
 		{
 			throw new \RuntimeException(sprintf('There is no class for table "%s" registered in $GLOBALS[\'TL_MODELS\'].', $strTable));
 		}
 
-		return $strClass;
+		return $GLOBALS['TL_MODELS'][$strTable];
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -750,9 +738,7 @@ public static function findByPk($varValue, array $arrOptions=array())
 	{
 		if ($varValue === null)
 		{
-			trigger_deprecation('contao/core-bundle', '4.13', 'Passing "null" as primary key has been deprecated and will no longer work in Contao 5.0.', __CLASS__);
-
-			return null;
+			throw new \TypeError('Model::findByPk(): Argument #1 ($varValue) must not be of type null.');
```
aschempp: shouldn't we just add `string|int` as type argument instead?

m-vo: Thought about this as well, but this is IMHO better be done in a separate PR because it also affects all child classes. Maybe we should decide which of the legacy classes should receive type hints and then try to automate those changes as much as possible with tooling.

Toflar: In my StringUtil PR we added only the type hints required and left out all the other stuff which can still be improved with tooling :)

ausi (dismissed):

bytehead (dismissed):

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -1193,30 +1172,12 @@ public static function countAll()
 	 */
 	public static function getClassFromTable($strTable)
 	{
-		if (isset(static::$arrClassNames[$strTable]))
-		{
-			return static::$arrClassNames[$strTable];
-		}
-
-		if (isset($GLOBALS['TL_MODELS'][$strTable]))
+		if (null === ($strClass = $GLOBALS['TL_MODELS'][$strTable] ?? null))
```
leofeyer: Why so complicated?

```php
if (!isset($GLOBALS['TL_MODELS'][$strTable])) {
    throw new …
}

return $GLOBALS['TL_MODELS'][$strTable];
```

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -934,13 +920,6 @@ public static function findBy($strColumn, $varValue, array $arrOptions=array())
 		if (\count($arrColumn) == 1 && ($arrColumn[0] === static::getPk() || \in_array($arrColumn[0], static::getUniqueFields())))
 		{
 			$blnModel = true;
-
-			if ($varValue === null && $arrColumn[0] === static::getPk())
-			{
-				trigger_deprecation('contao/core-bundle', '4.13', 'Passing "null" as primary key has been deprecated and will no longer work in Contao 5.0.', __CLASS__);
-
-				return null;
```
leofeyer, line 942: Shouldn‘t we throw an exception here instead of removing the if condition entirely?

core-bundle/src/Resources/contao/library/Contao/Model.php
```diff
@@ -409,12 +403,6 @@ public function current()
 	 */
 	public function save()
 	{
-		// Deprecated call
-		if (\func_num_args() > 0)
-		{
-			throw new \InvalidArgumentException('The $blnForceInsert argument has been removed (see system/docs/UPGRADE.md)');
```
leofeyer, line 415: Why remove this? It still fulfills its purpose even in Contao 5 IMHO.

core-bundle/src/Resources/contao/library/Contao/Model.php
m-vo: I prefer DRY for array access. But this is otherwise CS for me.

core-bundle/src/Resources/contao/library/Contao/Model.php
m-vo: We're already throwing a type error/constrain the type in `findByPk`. Do we really need this here as well?

core-bundle/src/Resources/contao/library/Contao/Model.php
m-vo: Fine for me as well, though I'd prefer cleaning this up now.

leofeyer: Yes, we probably should. 👍

core-bundle/src/Resources/contao/library/Contao/Model.php
leofeyer: I think so, otherwise `findByPk(null)` would throw an exception but `findBy('id', null)` not.

m-vo: Adjusted in 874235efaf32d7aa52d6ef6cd37d3b555374ef91. I also tweaked the messages a bit to follow the errors thrown by PHP more closely.

core-bundle/src/Resources/contao/library/Contao/Model.php
m-vo: @leofeyer Feel free to change if you like.

ausi (dismissed):

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4548, opened 2022-04-26
Replace the "getSearchablePages" hook in the calendar-bundle

Toflar:
This is a first step in order to get rid of the `getSearchablePages` hook which has been deprecated for a while but the core itself is still using it :/

- I also had to introduce an event alias (which might be backported to 4.13)
- I extended the test case because we also need to be able to mock `createInstance` calls
- I've introduced a new helper method in the `SitemapEvent` in order to ease adding URLs to the existing url set


fritzmg:
> This is a first step in order to get rid of the getSearchablePages hook

Well the first step would have been  https://github.com/contao/contao/pull/4073 which still has not been merged 🙈

Toflar:
I didn't know this existed but good I did calendar-bundle and you did news-bundle :)

fritzmg:
> I didn't know this existed

You approved it though 😁🐠

Toflar:
Yeah, if it takes this long to get merged, I will forget about it.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4552, opened 2022-04-27
Remove the "fullscreen" option in the back end

leofeyer:
This means that the back end width will always be adjusted to the browser window.


ausi (approved):

asaage:
Why?
Is there a UI-Draft for 5.0 yet?
Having the BE 100% of browser-width was considered to be bad because of possible long distance of icons from titles.
I use the fullscreen option anyway in some installations because of there are Table-Views or very long titles.

leofeyer:
We will discuss this again once #4371 is more advanced.

PR #4574, opened 2022-04-29
Fix the impersonation message

### LINKED ISSUE #4170: Logging leaving user impersonation
### Affected version(s)

4.13.0

### Description

Leaving user impersonation (user menu -> ``Back to A``) creates a missleading system log entry (``User "B" has switched to user "A"``). But in fact A has quit the impersonation to B.

cliffparnitzky:
Fixes #4170

Explanation of the behaviour:

## Simple case

- "A" is logged in
- switches to "B" ... the log entry is `User "A" has switched to user "A"`
- switches back to itself ... the log entry is `User "A" has quit the impersonation of user "B"`

## Complex case

- "A" is logged in
- switches to "B" ... the log entry is `User "A" has switched to user "A"`
- switches directly to "C" ... the log entries are `User "A" has quit the impersonation of user "B"` and `User "A" has switched to user "C"`
- switches back to itself ... the log entry is `User "A" has quit the impersonation of user "C"`

I think, that is the right way, because it is alway user "A" who is the origin user.


core-bundle/src/EventListener/SwitchUserListener.php
```diff
@@ -59,9 +60,25 @@ public function __invoke(SwitchUserEvent $event): void
...
+
+        if ($originalUser === $targetUser) {
+            $this->logger->info(
+                sprintf('User "%s" has quit the impersonation of user "%s"', $originalUser, $sourceUser),
+                ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS, $originalUser)]
+            );
+        } else {
+            if (!empty($originalUser)) {
+                $sourceUser = $originalUser;
+            }
```
leofeyer, line 72:
```suggestion
            }

```

@cliffparnitzky If you would allow edits from the maintainers, I could have fixed this myself. 🙈

cliffparnitzky: I have currently no clue how allow edits of maintainers. I've added the new line after line 77 from the suggestion. But this runs into an "Cody styling error" in the checks.

core-bundle/src/EventListener/SwitchUserListener.php
```diff
@@ -59,9 +60,25 @@ public function __invoke(SwitchUserEvent $event): void
...
+        if ($token instanceof SwitchUserToken) {
+            $originalUser = $token->getOriginalToken()->getUser()->getUsername();
+        }
+
+        if ($originalUser === $targetUser) {
+            $this->logger->info(
+                sprintf('User "%s" has quit the impersonation of user "%s"', $originalUser, $sourceUser),
+                ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS, $originalUser)]
+            );
+        } else {
```
bytehead, line 69: I personally would combine the `else` and following `if` here: 
```suggestion
        } elseif (!empty($originalUser)) {
```

bytehead (changes requested):

core-bundle/src/EventListener/SwitchUserListener.php
```diff
@@ -59,9 +60,25 @@ public function __invoke(SwitchUserEvent $event): void
...
+            if (!empty($originalUser)) {
```
bytehead, line 70: Is there a special reason to use `!empty` instead of checking for `null !== $originalUser`?

cliffparnitzky: No, I think both is good. The `empty` is more or less my general choice of sting comparing, because it includes `null` and empty sting.

core-bundle/src/EventListener/SwitchUserListener.php
cliffparnitzky: I think that is not good, because the `if` is only used to reduce the number of loggings. @bytehead How should this look?

bytehead: Sorry, nevermind. I got distracted with the code view and the conversations 🙈

core-bundle/src/EventListener/SwitchUserListener.php
bytehead: 👍

core-bundle/src/EventListener/SwitchUserListener.php
bytehead: There's a checkbox on the lower right. After the closing bracket of the `if` and before `$this->logger->info` there should be an empty line.

bytehead (changes requested):

bytehead (dismissed):
👍

core-bundle/src/EventListener/SwitchUserListener.php
```diff
@@ -59,9 +60,26 @@ public function __invoke(SwitchUserEvent $event): void
...
         }
 
-        $this->logger->info(
-            sprintf('User "%s" has switched to user "%s"', $sourceUser, $targetUser),
-            ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS, $sourceUser)]
-        );
+        $originalUser = null;
+
+        if ($token instanceof SwitchUserToken) {
+            $originalUser = $token->getOriginalToken()->getUser()->getUsername();
```
aschempp:
```suggestion
            $originalUser = $token->getOriginalToken()->getUsername();
```

core-bundle/src/EventListener/SwitchUserListener.php
```diff
@@ -59,9 +60,26 @@ public function __invoke(SwitchUserEvent $event): void
             $targetUser = $targetUser->getUsername();
```
aschempp: `$targetUser` is always `instanceof UserInterface`, not sure why this check is here.

aschempp (commented):
By looking at the Symfony code, I think this could be a bit simplified. If `$token instanceof SwitchUserToken`, we don't need the `$sourceUser` etc. at all. So maybe that could be the first check in the method and log appropriately? 😊

leofeyer:
@cliffparnitzky What about @aschempp‘s suggestions above? Can you please resolve the conversations, so we can eventually merge this PR?

core-bundle/src/EventListener/SwitchUserListener.php
cliffparnitzky: Yes, this seems to be right. I will remove the check.

core-bundle/src/EventListener/SwitchUserListener.php
cliffparnitzky: This works. I will change it.

leofeyer (approved):

leofeyer:
Thank you @cliffparnitzky.

PR #4581, opened 2022-04-30
Remove the leading `index/` from folder URLs

### LINKED ISSUE #1000: Remove leading 'index' from alias when using folderUrls
The home page of a website usually has the alias `index` (for SEO reasons), right?
When `useFolderUrl` is enabled and subpages are created, their alias always contains the leading `index/` inherited from the home page.

Wouldn't it be nice to remove this part of the alias in `generateAlias()` of `tl_page`?

```
if(strpos($objPage->folderUrl, 'index/') === 0) {
    $objPage->folderUrl = str_replace('index/', '', $objPage->folderUrl);
}
```

Wusch:
Fixes #1000

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
@@ -87,7 +87,7 @@ class PageUrlListener
         // Generate folder URL aliases (see #4933)
         if ($pageModel->useFolderUrl) {
             // Remove leading index/ from folder url (see #1000)
-            $value = str_replace('index/', '', $pageModel->folderUrl) .$value;
+            $value = str_replace('index/', '', $pageModel->folderUrl).$value;
         }
 
         return $value;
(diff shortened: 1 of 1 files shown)


fritzmg:
Detail page URLs will still be be `https://example.com/index/news-alias` though.

aschempp (changes requested):
I think this should rather be fixed in `PageModel` where the `folderurl` property is created from the page alias.

Toflar:
As discussed in our Contao Slack Call on July 28, 2022, we decided that this is actually a bug (it does not affect existing aliases), so we're re-assigning it to 4.9. @Wusch do you want to adjust that and rebase for 4.9 or should @aschempp do that? It should be moved to here: https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/models/PageModel.php#L939

Wusch:
> As discussed in our Contao Slack Call on July 28, 2022, we decided that this is actually a bug (it does not affect existing aliases), so we're re-assigning it to 4.9. @Wusch do you want to adjust that and rebase for 4.9 or should @aschempp do that? It should be moved to here: https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/models/PageModel.php#L939

@aschempp can go for it. I'm not available the next weeks.

aschempp:
see https://github.com/contao/contao/pull/5075

PR #4583, opened 2022-04-30
Allow to override dns and protocol of pages by configuration

### LINKED ISSUE #1517: Override page dns settings by configuration
When having a Contao project on multiple stages and in custom developer environments the page dns settings has always be adjusted. Automated processes require the extra database manipulation to adopt these settings. 

After discussing this limitation with a customer of mine, the idea came up to override the page dns settings by a configuration file. So it would be possible to change the dns entries depending on your environment.

A possible configuration structure might be:

```yaml
contao:
  routing:
    override_dns:
       # DNS entry defined in the root page
       example.org:
         dns: staging.example.org
         ssl: false
```

Now Contao would always use staging.example.org for the pages where the dns setting is set to `example.org`.

Wdyt?

dmolineus:
Fixes #1517

This pull request allow to override the dns and ssl settings of route pages. This feature is designed for multiple stages and local development. Therefore it overrides the dns globally and not root page specific (e.g. different setting for root pages with same dns).

### DIFF
--- a/core-bundle/src/DependencyInjection/Configuration.php
+++ b/core-bundle/src/DependencyInjection/Configuration.php
@@ -117,6 +117,7 @@ class Configuration implements ConfigurationInterface
                 ->append($this->addInsertTagsNode())
                 ->append($this->addBackupNode())
                 ->append($this->addSanitizerNode())
+                ->append($this->addRoutingNode())
             ->end()
         ;
 
@@ -685,6 +686,28 @@ class Configuration implements ConfigurationInterface
                                 return $protocols;
                             }
                         )
+                    ->end()
+                ->end()
+            ->end()
+        ;
+    }
+
+    private function addRoutingNode(): NodeDefinition
+    {
+        return (new TreeBuilder('routing'))
+            ->getRootNode()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('override_dns')
+                    ->arrayPrototype()
+                        ->children()
+                            ->scalarNode('dns')
+                                ->info('The domain name')
+                            ->end()
+                            ->scalarNode('protocol')
+                                ->info('The domain name')
+                            ->end()
+                        ->end()
                     ->end()
                 ->end()
             ->end()
--- a/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
+++ b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
@@ -130,6 +130,7 @@ class ContaoCoreExtension extends Extension implements PrependExtensionInterface
         $container->setParameter('contao.intl.countries', $config['intl']['countries']);
         $container->setParameter('contao.insert_tags.allowed_tags', $config['insert_tags']['allowed_tags']);
         $container->setParameter('contao.sanitizer.allowed_url_protocols', $config['sanitizer']['allowed_url_protocols']);
+        $container->setParameter('contao.routing.override_dns', $config['routing']['override_dns']);
 
         $this->handleSearchConfig($config, $container);
         $this->handleCrawlConfig($config, $container);
--- a/core-bundle/src/EventListener/OverrideDnsSettingsListener.php
+++ b/core-bundle/src/EventListener/OverrideDnsSettingsListener.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\ServiceAnnotation\Hook;
+use Contao\PageModel;
+
+/**
+ * @psalm-type TOverrideDnsConfiguration = array{dns?: string, protocol?: string}
+ */
+class OverrideDnsSettingsListener
+{
+    private array $configuration;
+
+    /**
+     * @param array<string,array<string,string>> $configuration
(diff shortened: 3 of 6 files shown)


fritzmg:
This won't really be usable for us at least, as we often have multi-domain setups and thus also need different domains in our local development environments for example. Also I am not really sure how to use this within different environments?

richardhj:
Related: #4077

richardhj:
> > Therefore it overrides the dns globally and not root page specific
> 
> This won't really be usable for us at least, as we often have multi-domain setups and thus also need different domains in our local development environments for example.

From what the configuration looks like, you can override multiple DNSes, see https://github.com/contao/contao/pull/4583/files#diff-9b94f76b99d33315fb56d2727409c02c2fa453692699d8579a60b650f55746efR702.

> Also I am not really sure how to use this within different environments?

In the config then, you can use ENV vars.

This PR and https://github.com/contao/contao/pull/4077#issuecomment-1035201680 are competing though.

fritzmg:
> From what the configuration looks like, you can override multiple DNSes, see https://github.com/contao/contao/pull/4583/files#diff-9b94f76b99d33315fb56d2727409c02c2fa453692699d8579a60b650f55746efR702.

Yeah you are right, I misread how the config works. Essentially the DNS name in the website root acts as an alias that you use in your config - and the config will hold the actual domain.

> In the config then, you can use ENV vars.

Yes, but then it's not optional depending on the environment. But that's fine in the end.

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -690,4 +691,26 @@ static function (array $protocols): array {
...
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('override_dns')
+                    ->arrayPrototype()
+                        ->children()
+                            ->scalarNode('dns')
+                                ->info('The domain name')
+                            ->end()
+                            ->scalarNode('protocol')
+                                ->info('The domain name')
```
fritzmg, line 708:
```suggestion
                                ->info('The protocol for this domain.')
```

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -690,4 +691,26 @@ static function (array $protocols): array {
...
+    private function addRoutingNode(): NodeDefinition
+    {
+        return (new TreeBuilder('routing'))
+            ->getRootNode()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('override_dns')
+                    ->arrayPrototype()
+                        ->children()
+                            ->scalarNode('dns')
```
fritzmg, line 704: DNS stands for "Domain Name System" and describes the overall "naming system used to identify computers reachable through IP networks". I would prefer either `domain` or `host` or `hostname`.

aschempp:
I think the concept can work, not sure its easier to use than #4077, but it has the advantage of being configuration and not DB stuff. However, it would be rather intransparent in the backend, I wonder if we should override the backend fields to show the mapped host? But how do you override it then? 😅

fritzmg:
> I wonder if we should override the backend fields to show the mapped host? But how do you override it then? 😅

I would handle it the same way as `contao.localconfig`. Show an ℹ️  symbol and make it read-only.

dmolineus:
>  This won't really be usable for us at least, as we often have multi-domain setups and thus also need different domains in our local development environments for example. 

You can override multiple domain settings based on domain level but not on page level. Imagine you have two root pages `example.org/de` and `example.org/en`. You will be able to map both to `example.localhost` (with configured url prefix) but not both domains to different domains (`de.example.localhost` and `en.example.localhost`).

> Also I am not really sure how to use this within different environments?

You can use env variables for the domain names.

dmolineus:
> I think the concept can work, not sure its easier to use than https://github.com/contao/contao/pull/4077, but it has the advantage of being configuration and not DB stuff.

I honestly missed #4077 and after having a look at it, I would still prefer an configuration based solution.

> However, it would be rather intransparent in the backend, I wonder if we should override the backend fields to show the mapped host? But how do you override it then? sweat_smile

I would add some messages in the page structure. The overridden fields still have to be editable I think. Working on a staging system which database will be uploaded to the live environment requires that domain names can be edited.

fritzmg:
I also prefer this approach 👍

core-bundle/src/DependencyInjection/Configuration.php
bezin: On the other hand, `dns` is the name of the page field, that is being overwritten eventually.  `dns` sounds reasonable to me.

aschempp:
> I would handle it the same way as `contao.localconfig`. Show an ℹ️ symbol and make it read-only.

Yeah, but as @dmolineus said it would still need to be editable, maybe you want to use something else than the configuration. I can't think of a user friendly interface atm, maybe include it in the page routing information?

fritzmg:
idk, in my opinion it should not be editable. Just like the localconfig is not editable anymore, if persisted in the config.

> Working on a staging system which database will be uploaded to the live environment requires that domain names can be edited.

Not sure I understand the use case. Isn't the point of this setting that you _don't_ have to do this anymore?

aschempp:
I think this is very different from localconfig. You enter a domain name, and if it matches any of the replace configuration, you will no longer be able to change it. For localconfig, the field cannot be defined in the back end if it is in config. But the DNS _can_ be defined **to match a config**. That's why I think it's confusing.

biowebfr:
Hi, is there any progress on this ticket ? :)

leofeyer:
As we have merged #6529, this PR is now redundant.

@dmolineus Does #6529 work for you? If yes, I think we can close this PR, can‘t we?

PR #4586, opened 2022-05-01
Also ignore the Resources folder

aschempp:
The _magic_ autoloading might fail, if the `Resources` folder contains invalid files.

/cc @zonky2

### DIFF
--- a/manager-bundle/src/Resources/skeleton/config/services.php
+++ b/manager-bundle/src/Resources/skeleton/config/services.php
@@ -28,7 +28,7 @@ return static function(ContainerConfigurator $configurator) use ($container) {
     try {
         $config
             ->load('App\\', $servicesDir.'/*')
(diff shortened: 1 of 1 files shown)


fritzmg (approved):
Should we also exclude `Model` while we are at it?

aschempp:
Imho not necessary, anything extending the `Contao\System` class will be ignored automatically.

Toflar (approved):

fritzmg:
> Imho not necessary, anything extending the `Contao\System` class will be ignored automatically.

Models do not extend from `Contao\System`.

m-vo (approved):

aschempp:
> > Imho not necessary, anything extending the `Contao\System` class will be ignored automatically.
> 
> Models do not extend from `Contao\System`.

https://github.com/contao/contao/blob/5.x/manager-bundle/src/Resources/skeleton/config/services.php#L61

PR #4562, opened 2022-04-28
Remove TL_CROP

ausi:
For a smoother upgrade path I kept support for the legacy resize modes (like `left_top`, `center_center`,…) but deprecated their usage and removed them from the default image size dropdown.

PR #4587, opened 2022-05-01
Add missing service argument for Version410Update

fritzmg:
Follow up for #4562. Fixes

```
Too few arguments to function Contao\CoreBundle\Migration\Version401\Version410Update::__construct(), 2 passed in var\cache\dev\ContainerKqpboko\getContao_Migration_CollectionService.php on line 49 and exactly 3 expected
```

### DIFF
--- a/core-bundle/src/Resources/config/migrations.yml
+++ b/core-bundle/src/Resources/config/migrations.yml
@@ -17,6 +17,7 @@ services:
         arguments:
             - '@database_connection'
             - '@contao.framework'
+            - '@contao.image.sizes'
 
     contao.migration.version_403.version_430_update:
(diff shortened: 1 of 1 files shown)


m-vo (approved):
I wonder why the CI did run through, shouldn't this have failed when building the container?

fritzmg:
It only occurs at runtime and we have no functional tests for Migrations.

leofeyer:
Why is this a bugfix for Contao 5 instead of 4.13?

fritzmg:
> Why is this a bugfix for Contao 5 instead of 4.13?

The bug only exists in Contao 5 because of #4562.

leofeyer:
Thank you @fritzmg.

PR #4589, opened 2022-05-02
Re-introduce autowiring aliases for subscribed services

fritzmg:
Currently the following error happens, if you try to implement a content element or module as a fragment when extending from the appropriate abstract classes:

```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException:
The service ".service_locator.t.IY1uE" has a dependency on a non-existent service "Contao\CoreBundle\Csrf\ContaoCsrfTokenManager".
```

This is because we are registering `contao.csrf.token_manager` as a subscribed service in our `AbstractController`. However autowiring is still used for service subscribers and due to the removal of the alias for `ContaoCsrfTokenManager` in #4554 autowiring is not possible anymore.

This PR fixes that by re-introducing an autowiring alias for the `ContaoCsrfTokenManager`. This makes sense independently from this issue anyway, as it is currently not possible to autowire the `ContaoCsrfTokenManager` in your `App` - which is important since we want to use its `getDefaultValue` method.

The same error also happens for `Contao\CoreBundle\Twig\Loader\ContaoFilesystemLoader` which is subscribed in the `AbstractFragmentController`.

~~This PR also removes a superfluous `getSubscribedServices` definition in `AbstractFrontendModuleController`. It only adds the `ContaoCsrfTokenManager` - which is already added in on of the parent classes.~~

### DIFF
--- a/core-bundle/src/Controller/FrontendModule/AbstractFrontendModuleController.php
+++ b/core-bundle/src/Controller/FrontendModule/AbstractFrontendModuleController.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\Controller\FrontendModule;
 
 use Contao\CoreBundle\Controller\AbstractFragmentController;
-use Contao\CoreBundle\Csrf\ContaoCsrfTokenManager;
 use Contao\CoreBundle\String\HtmlAttributes;
 use Contao\CoreBundle\Twig\FragmentTemplate;
 use Contao\ModuleModel;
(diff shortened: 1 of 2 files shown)


m-vo:
> This PR also removes a superfluous getSubscribedServices definition in AbstractFrontendModuleController. It only adds the ContaoCsrfTokenManager - which is already added in on of the parent classes.

IMHO this is correct, because a class should not assume any transitive dependencies in the subscribed services.

fritzmg:
> IMHO this is correct, because a class should not assume any transitive dependencies in the subscribed services.

Fair enough - reverted.

m-vo (approved):
This feels very odd. But not for the first time 😉. And unfortunately, I don't see any other solution for this problem.

leofeyer:
I think it‘s perfectly fine. Thank you @fritzmg.

PR #4590, opened 2022-05-02
Require symfony/deprecation-contracts

### LINKED ISSUE #4550: trigger_deprecation in Contao 4.9?
### Affected version(s)

4.9

### Description

https://github.com/contao/contao/blob/4.9/core-bundle/src/Resources/contao/library/Contao/StringUtil.php#L492

isn't this a bug and should use `trigger_error`?

leofeyer:
Fixes #4550

Turns out we can use `trigger_deprecation()` in Contao 4.9. All we have to do is to require `symfony/deprecation-contracts`. 🤷‍♂️

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,6 +90,7 @@
         "symfony/debug": "4.4.*",
         "symfony/debug-bundle": "4.4.*",
         "symfony/dependency-injection": "4.4.*",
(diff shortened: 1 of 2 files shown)


bytehead (approved):

aschempp:
nice find! 😆

PR #4596, opened 2022-05-02
Fix upload directory check in ImageFactory on Windows

fritzmg:
On Windows, `%kernel.project_dir%/%contao.upload_path%` will be something like this:

```
C:\Users\fmg\www\c49/files
```

However, the given image path might look like this, if it has been normalised beforehand somewhere:

```
C:/Users/fmg/www/c49/files/dts/DeathtoStock_Meticulous-04.jpg
```

In this case the check whether the given file is part of the upload directory (i.e. part of the file manager) will fail due to the string comparison, and thus things like the important part will not be taken into account when creating the image.

This PR fixes that by using `Path::isBasePath` instead, which handles proper path normalisation.

### DIFF
--- a/core-bundle/src/Image/ImageFactory.php
+++ b/core-bundle/src/Image/ImageFactory.php
@@ -26,6 +26,7 @@ use Contao\ImageSizeModel;
 use Imagine\Image\ImagineInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Webmozart\PathUtil\Path;
 
 class ImageFactory implements ImageFactoryInterface
 {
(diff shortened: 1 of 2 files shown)


ausi (dismissed):

Toflar (dismissed):

leofeyer (commented):
Can you please also add a unit test, so we don‘t run into regressions in the future?

fritzmg:
> Can you please also add a unit test, so we don‘t run into regressions in the future?

I have adjusted the tests accordingly. No separate test is needed for this, because the tests would have already detected this. They didn't since the `uploadDir` path was already normalized (which does not accurately reflect the real application, where the `uploadDir` ist just built from `%kernel.project_dir%/%contao.upload_path%` without normalisation) - and all image paths are normalised in the tests and thus it never failed. I have removed the `uploadDir` normalisation, so now the tests would fail on Windows without the adjustment in `ImageFactory::createImportantPart`.

m-vo (approved):

leofeyer:
Thank you @fritzmg.

PR #4598, opened 2022-05-03
Remove leftover contao.encryption_key occurences

### LINKED ISSUE #4597: Remove contao.encryption_key from configuration
### Affected version(s)

4.13.2

### Description

After the legacy encryption was removed in https://github.com/contao/contao/pull/2694 and finally dropped in https://github.com/contao/contao/pull/4307. There is no need for the parameter `contao.encryption_key`, also it will not be set into the container.

Therefore we could remove following block:
https://github.com/contao/contao/blob/bc07412dd5c439590247dc684b970957e48e0f87/core-bundle/src/DependencyInjection/Configuration.php#L49-L52

If one still relies on `contao.encryption_key` you should use `kernel.secret`

bytehead:
Fixes #4597


bytehead:
Failing PHPStan check does not seem to be related.

leofeyer:
Thank you @bytehead.

leofeyer:
Thank you @bytehead.

PR #4603, opened 2022-05-03
Correctly generate the edit URL of a version

### LINKED ISSUE #4460: Dashboard version compare does not work
### Affected version(s)

4.13.2

### Description

I noticed today, when i click on show differences, the window shows the dashboard again
grafik
grafik

after closing and reloading, another version was created, and so on.

grafik

I have this on 2 installations, I think that's not right.

leofeyer:
Fixes #4460

This is because the icon now uses the new `toggle()` method.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -759,32 +759,36 @@ class Versions extends Controller
 			return sprintf($this->strEditUrl, $this->intPid);
 		}
 
-		$strUrl = Environment::get('request');
+		$pairs = array();
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 
-		// Save the real edit URL if the visibility is toggled via Ajax
-		if (preg_match('/&(amp;)?state=/', $strUrl))
-		{
-			$strUrl = preg_replace
-			(
-				array('/&(amp;)?id=[^&]+/', '/(&(amp;)?)t(id=[^&]+)/', '/(&(amp;)?)state=[^&]*/'),
-				array('', '$1$3', '$1act=edit'),
-				$strUrl
-			);
-		}
+		parse_str($request->server->get('QUERY_STRING'), $pairs);
 
 		// Adjust the URL of the "personal data" module (see #7987)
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/Versions.php
```diff
@@ -759,32 +759,36 @@ protected function getEditUrl()
 			return sprintf($this->strEditUrl, $this->intPid);
 		}
 
-		$strUrl = Environment::get('request');
+		$pairs = array();
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 
-		// Save the real edit URL if the visibility is toggled via Ajax
-		if (preg_match('/&(amp;)?state=/', $strUrl))
+		parse_str($request->server->get('QUERY_STRING'), $pairs);
```
ausi, line 765: Is there a specific reason we don’t use `$request->query->all()` here?

ausi (approved):

core-bundle/src/Resources/contao/classes/Versions.php
leofeyer: Good point. We are doing it this way here:

https://github.com/contao/contao/blob/ab71fd99f6c7ce36595a0d8dd75b17710d98a2bb/core-bundle/src/Resources/contao/library/Contao/Controller.php#L1119

https://github.com/contao/contao/blob/ab71fd99f6c7ce36595a0d8dd75b17710d98a2bb/core-bundle/src/Resources/contao/modules/ModuleLogout.php#L73

https://github.com/contao/contao/blob/ab71fd99f6c7ce36595a0d8dd75b17710d98a2bb/core-bundle/src/Resources/contao/pages/PageLogout.php#L62

But maybe we should use `$request->query->all()` everywhere instead?

ausi: But not in a bugfix version then I think. I’d merge this as is and maybe change it in Contao 5.

Toflar: Yeah that looks really weird but I agree changing it in 5.

Toflar (approved):

PR #4604, opened 2022-05-03
Use host name without port in some controllers

### LINKED ISSUE #4602: Sitemap, Robots and Favicon controller always return 404 with domain and port
### Affected version(s)

4.13, 5.x

### Description

1. Configure a DNS entry for your root page, e.g. `contao.local`
2. Run the application on some port, e.hg. `8000`
3. Request contao.local:8000/sitemap.xml

The response will always be 404, because `$request->server->get('HTTP_HOST')` always includes the port. The port however is not configured in the DNS page settings, because it is supposed to be port agnostic – hence the matching fails. 

See [RobotsTxtController](https://github.com/contao/contao/blob/2f178ccfbaaa1db7eb1f3a9018d530e3338e7796/core-bundle/src/Controller/RobotsTxtController.php#L46), [SitemapController](https://github.com/contao/contao/blob/2f178ccfbaaa1db7eb1f3a9018d530e3338e7796/core-bundle/src/Controller/SitemapController.php#L45) and [FaviconController](https://github.com/contao/contao/blob/2f178ccfbaaa1db7eb1f3a9018d530e3338e7796/core-bundle/src/Controller/FaviconController.php#L46).

As the other issues with DNS settings and ports other than 80 it probably only affects some dev environments, but since it is supposed to be port agnostic, I think we should fix this.

bezin:
Fixes #4602

### DIFF
--- a/core-bundle/src/Controller/FaviconController.php
+++ b/core-bundle/src/Controller/FaviconController.php
@@ -50,7 +50,7 @@ class FaviconController
         $pageModel = $this->framework->getAdapter(PageModel::class);
 
         $rootPage = $pageModel->findPublishedFallbackByHostname(
-            $request->server->get('HTTP_HOST'),
+            $request->getHost(),
             ['fallbackToEmpty' => true]
         );
 
--- a/core-bundle/src/Controller/RobotsTxtController.php
+++ b/core-bundle/src/Controller/RobotsTxtController.php
@@ -48,7 +48,7 @@ class RobotsTxtController
         $pageModel = $this->framework->getAdapter(PageModel::class);
 
         $rootPage = $pageModel->findPublishedFallbackByHostname(
-            $request->server->get('HTTP_HOST'),
+            $request->getHost(),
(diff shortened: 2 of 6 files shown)


fritzmg (dismissed):

fritzmg:
@bezin can you make a separate PR for Contao 4.9 for the `FaviconController` and `RobotsTxtController`?

Toflar:
Looks good to me but I'd like to have tests for that in order to not have regression bugs :)

bezin:
I added a test draft for SitemapController. Not sure if this is the way to go tbh. Maybe one test case with domain and port is sufficient?

Toflar:
Oh :D Yeah I think one isolated testcase that just asserts that `findPublishedFallbackByHostname()` is called with the correct host whould be enough. The rest is not relevant for that imho.

Toflar (approved):
Perfect! ❤️

fritzmg (approved):

leofeyer:
Thank you @bezin.

PR #4605, opened 2022-05-04
Ignore port in Favicon and RobotsTxt controller

### LINKED ISSUE #4602: Sitemap, Robots and Favicon controller always return 404 with domain and port

bezin:
This is a follow-up to #4604 for Contao 4.9 as requested by @fritzmg

### DIFF
--- a/core-bundle/src/Controller/FaviconController.php
+++ b/core-bundle/src/Controller/FaviconController.php
@@ -63,7 +63,7 @@ class FaviconController
 
         /** @var PageModel|null $rootPage */
         $rootPage = $pageModel->findPublishedFallbackByHostname(
-            $request->server->get('HTTP_HOST'),
+            $request->getHost(),
             ['fallbackToEmpty' => true]
         );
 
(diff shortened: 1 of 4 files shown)


bezin:
Fixes #4602 in Contao 4.9

fritzmg (approved):

Toflar (approved):

leofeyer:
Thank you @bezin.

PR #4618, opened 2022-05-06
Fix the global state unit tests

leofeyer:
This should fix https://github.com/contao/contao/runs/6307113827?check_suite_focus=true. Note that the job is only run periodically, so even if the CI chain below is green, the tests might still fail. Run `vendor/bin/phpunit --extensions Contao\\CoreBundle\\Tests\\PhpunitExtension\\GlobalStateWatcher` on the CLI to verify.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -39,23 +39,23 @@ class Input
 	/**
 	 * Parameters set via setGet() are stored by request
 	 *
-	 * @var \WeakMap<Request,array<string,array|string>>|null
+	 * @var \WeakMap<Request,array<string,array|string>>
 	 */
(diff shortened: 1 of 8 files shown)


core-bundle/src/Resources/contao/library/Contao/Input.php
```diff
@@ -39,23 +39,23 @@ class Input
 	/**
 	 * Parameters set via setGet() are stored by request
 	 *
-	 * @var \WeakMap<Request,array<string,array|string>>
+	 * @var \WeakMap<Request,array<string,array|string>>|null
 	 */
-	private static \WeakMap $setGet;
+	private static \WeakMap|null $setGet = null;
```
ausi: I’d prefer not making them nullable, see https://github.com/leofeyer/contao/pull/4
WDYT?

ausi (approved):

PR #4620, opened 2022-05-06
Adjust name of Twig marker file that denotes roots of nested template paths

### LINKED ISSUE #4577: Re-think the .root file name for the Twig integration
### Description

As noticed at the Contao Camp 22, the `.root` filename seems to be confusing for many people. Maybe we can come up with a better name for it?

m-vo:
Fixes #4577 

 `.root` &rarr; `.twig-root`


Toflar (approved):

bytehead (approved):

leofeyer:
Thank you @m-vo.

PR #4623, opened 2022-05-07
Improve how the Contao Twig escaper works

### LINKED ISSUE #4551: Improve default escaping strategy for Contao Twig templates
### Affected version(s)

4.13

### Description

Assume there is a Contao template `fe_page.html.twig` that includes `_my_partial.html.twig` like this (i.e. without a `@Contao` namespace prefix):

```twig
{% include "/_my_partial.html.twig" %}
```

Now our default escaper rule, `^@Contao(_[a-zA-Z0-9_-]*)?/`, that rewrites the escaper filter to our non-double-encoding one, won't match anymore for the included module. The same is the case if you include/embed/import things of a Symfony bundle template (e.g. our current image studio macros). 

This is somewhat expected and you could work around this behavior by adding your own escaper rules, but I think we should improve some things.

1) We should probably explicitly add our own views, but we could also…
2) …try to inherit the property of being a Contao template down the tree.

With inheriting I mean, that only the name of the template, that contains the include/embed/import statements matters, not the included/embedded/imported module (and their children).

m-vo:
This implements fixes for #4551 but in a different way than anticipated. The current problem turned out to be caused mainly by indeterministic behavior. Consider these two examples:
```twig
{% include '@Contao/foo.html.twig' %}
{% include '/foo.html.twig' %}
```

vs

```twig
{% include '/bar.html.twig' %}
{% include '@Contao/bar.html.twig' %}
```

In the first example, the included file will be compiled as if it was a Contao template and then reused by the second statement because of a matching cache key. In the second example, however, things will be exactly the opposite. This is now fixed by applying a prefix to the cache key in our loader.

Inheriting the property of being a "Contao template" in the tree, didn't turn out to be feasible. The second change in this PR, therefore at least adds an escaper rule to include our already shipped bundle templates.

### DIFF
--- a/core-bundle/src/Twig/Extension/ContaoExtension.php
+++ b/core-bundle/src/Twig/Extension/ContaoExtension.php
@@ -56,8 +56,11 @@ final class ContaoExtension extends AbstractExtension
         $escaperExtension->setEscaper('contao_html', [$contaoEscaper, 'escapeHtml']);
         $escaperExtension->setEscaper('contao_html_attr', [$contaoEscaper, 'escapeHtmlAttr']);
 
-        // Use our escaper on all templates in the "@Contao" and "@Contao_*" namespaces
+        // Use our escaper on all templates in the "@Contao" and "@Contao_*"
+        // namespaces, as well as the existing bundle templates we're already
+        // shipping.
         $this->addContaoEscaperRule('%^@Contao(_[a-zA-Z0-9_-]*)?/%');
+        $this->addContaoEscaperRule('%^@Contao(Core|Installation)/%');
     }
 
     /**
--- a/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
+++ b/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
@@ -184,7 +184,7 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
         // We prefix the cache key to make sure templates from the default
         // Symfony loader won't be reused. Otherwise, we cannot reliably
         // differentiate when to apply our input encoding tolerant escaper
-        // filters (see #todo).
+        // filters (see #4623).
         return 'c'.parent::getCacheKey($templateName);
     }
 
--- a/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
+++ b/core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
@@ -181,7 +181,11 @@ class ContaoFilesystemLoader extends FilesystemLoader implements TemplateHierarc
     {
         $templateName = $this->getThemeTemplateName($name) ?? $name;
 
-        return parent::getCacheKey($templateName);
+        // We prefix the cache key to make sure templates from the default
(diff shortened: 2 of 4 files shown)


ausi (approved):

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -56,8 +56,11 @@ public function __construct(Environment $environment, TemplateHierarchyInterface
         $escaperExtension->setEscaper('contao_html', [$contaoEscaper, 'escapeHtml']);
         $escaperExtension->setEscaper('contao_html_attr', [$contaoEscaper, 'escapeHtmlAttr']);
 
-        // Use our escaper on all templates in the "@Contao" and "@Contao_*" namespaces
+        // Use our escaper on all templates in the "@Contao" and "@Contao_*"
+        // namespaces, as well as the existing bundle templates we're already
+        // shipping.
         $this->addContaoEscaperRule('%^@Contao(_[a-zA-Z0-9_-]*)?/%');
+        $this->addContaoEscaperRule('%^@Contao(Core|Installation)/%');
```
leofeyer, line 63: The regex now no longer supports `@Contao_*`. Is this intentional? If so, we should probably adjust the comment.

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
```diff
@@ -181,7 +181,11 @@ public function getCacheKey(string $name): string
     {
         $templateName = $this->getThemeTemplateName($name) ?? $name;
 
-        return parent::getCacheKey($templateName);
+        // We prefix the cache key to make sure templates from the default
+        // Symfony loader won't be reused. Otherwise, we cannot reliably
+        // differentiate when to apply our input encoding tolerant escaper
+        // filters (see #4623).
+        return 'c'.parent::getCacheKey($templateName);
```
leofeyer, line 188: Is this safe or should we use `contao` instead?

core-bundle/src/Twig/Extension/ContaoExtension.php
ausi: There are two regexes now, one for `@Contao_` and one for `@ContaoCore/`/`@ContaoInstallation`
I got confused with the diff view too ☺️

leofeyer: My bad. 🤦‍♂️

core-bundle/src/Twig/Loader/ContaoFilesystemLoader.php
m-vo: IMHO we just need to make sure there is a difference. And as the parent returns paths that start with a slash, a single char should be enough. 

But this gets hashed anyway, so I'm fine with any other prefix as well if you prefer. ☺️

leofeyer: No, it‘s fine then.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

Issue #689, opened 2017-10-08
Last custom section cannot be deleted


fritzmg:
If you have defined one ore more custom sections, but you want to delete all of them, the last one cannot be deleted by clicking on the delete button (×). Instead, you have to remove the text within the "ID" input field in order to delete that custom sections, which seems a little counter intuitive.

PR #4624, opened 2022-05-07
Implement sectionWizard in vanilla JS

fritzmg:
This implements the sectionWizard in vanilla JS and also implements #689 for the sectionWizard.

_Note:_ this implementation still uses the _Sortable_ mooTools JavaScript. Any suggestions for a replacement? What about https://github.com/SortableJS/Sortable for example?

_Note:_ I did not include this part of the previous sectionWizard: https://github.com/contao/contao/blob/720a4e134d8446bdb528bce3a0c0fbe1d22a8d49/core-bundle/src/Resources/public/core.js#L1907-L1929 According to the code it is supposed to reorder the items via the keyboard (arrow up and down) via the drag handle button - however I could not get this to work even with the original JavaScript. Is this working for anyone?

_Note:_ I implemented a MutationObserver for this. If we do this for all back end JavaScripts, we could switch to `async` for the combined file in the future. wdyt?


core-bundle/src/Resources/public/sectionwizard.js
```diff
@@ -0,0 +1,101 @@
+(function(){
+    'use strict';
+
+    const initialized = [];
```
ausi: We should use a `WeakMap` here so that we don’t create a memory leak.
_(Probably not really relevant for the section wizard, but if this code gets reused for other widgets too it would make the implementation more robust IMO)_
```suggestion
    const initialized = new WeakMap();
```

core-bundle/src/Resources/public/sectionwizard.js
```diff
@@ -0,0 +1,101 @@
...
+
+    const init = (node) => {
+        node.querySelectorAll('.tl_sectionwizard').forEach((table) => {
+            if (initialized.includes(table)) {
+                return;
+            }
+
+            initialized.push(table);
```
ausi:
```suggestion
            if (initialized.has(table)) {
                return;
            }

            initialized.set(table, true);
```

core-bundle/src/Resources/public/sectionwizard.js
```diff
@@ -0,0 +1,101 @@
...
+                                Array.from(tr.children).forEach((child) => {
+                                    const next = child.cloneNode(true);
+                                    ntr.appendChild(next);
+                                    const selects = child.querySelectorAll('select');
+                                    const nselects = next.querySelectorAll('select');
+                                    for (let j=0; j<selects.length; j++) {
+                                        nselects[j].value = selects[j].value;
+                                    }
+                                });
+                                tr.parentNode.insertBefore(ntr, tr.nextSibling);
```
ausi: Why not clone the `tr` element itself?

ausi (commented):
:+1:

core-bundle/src/Resources/public/sectionwizard.js
fritzmg: 🤷 the original implementation was this way. Changed in 0000907

core-bundle/src/Resources/public/sectionwizard.js
fritzmg: 👍 however, do we actually need to keep track of initialized nodes? I only included it because https://github.com/terminal42/contao-conditionalformfields/blob/develop/public/conditionalformfields.js does it this way as well 🙈

ausi: Yes, I think we have to because with the mutation observer it could be that changes in the tree trigger it again.
Unless there is another way to detect if the script did already run for a certain table?

fritzmg: We could also set an arbitrary boolean property on the node itself - that's what I usually do these days, but not sure what is best practise.

ausi: I think setting arbitrary properties on nodes is considered bad because of possible naming conflicts.

ausi (dismissed):

ausi:
> I implemented a MutationObserver for this. If we do this for all back end JavaScripts, we could switch to `async` for the combined file in the future. wdyt?

I like that very much!

fritzmg:
@leofeyer if we go with https://github.com/SortableJS/Sortable, can you create a `contao-component` for it?

aschempp:
I'm using https://github.com/kutlugsahin/smooth-dnd in one of my projects. I think to remember to look at Sortable-JS, not sure if I didn't use it because of no good VueJS wrapper or another reason like better mobile support. https://github.com/kutlugsahin/smooth-dnd didn't have commits for a while though 😕

fritzmg:
@leofeyer I'd vote to merge this as is and then replace all _Sortable_ instances separately, once we have decided what to use.

core-bundle/src/Resources/public/sectionwizard.js
```diff
@@ -0,0 +1,97 @@
...
+            const tbody = table.querySelector('tbody');
+
+            const makeSortable = (tbody) => {
+                Array.from(tbody.children).forEach((tr, i) => {
+                    tr.querySelectorAll('input, select').forEach((el) => {
+                        el.name = el.name.replace(/\[[0-9]+]/g, '[' + i + ']');
+                    });
+                });
+
+                new Sortables(tbody, {
```
leofeyer, line 24:
```suggestion
                // TODO: replace this with a vanilla JS solution
                new Sortables(tbody, {
```

core-bundle/src/Resources/contao/templates/backend/be_main.html5
```diff
@@ -40,6 +40,7 @@
       $objCombiner->add('bundles/contaocore/limit-height.min.js');
       $objCombiner->add('bundles/contaocore/tips.min.js');
       $objCombiner->add('system/themes/'.$this->theme.'/hover.min.js');
+      $objCombiner->add('bundles/contaocore/sectionwizard.min.js');
```
leofeyer, line 42: Did you add this after the `hover.min.js` script on purpose? If not, we should probably move it up.

fritzmg: No, I just added it last, since it's a new entry.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4626, opened 2022-05-07
Fix global state in tests

ausi:
Similar to #4618 but for Contao 4.13.

I also enabled the global state watcher in the CI now. Speed should be OK I think (or hope).
~~We could also combine it with the reverse order job to prevent increasing the number ob jobs too much?~~ Did that now.


leofeyer (approved):

leofeyer:
Thank you @ausi.

leofeyer:
Thank you @ausi.

PR #4628, opened 2022-05-08
Automatically reduce tl_search.url length

### LINKED ISSUE #4615: tl_search.url - specified key was too long
### Affected version(s)

4.9

### Description

In #4427 we changed `tl_search.url` to a length of `2048` to fix #4260. However, without `innodb_large_prefix` (which is currently still a recommendation and not a requirement for Contao) this will lead to the following error:

```
An exception occurred while executing 'ALTER TABLE tl_search CHANGE url url VARCHAR(2048) DEFAULT '' NOT NULL COLLATE `ascii_bin`':  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes
```

fritzmg:
Fixes #4615


core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
```diff
@@ -0,0 +1,127 @@
...
+        // Check if the field has an index
+        try {
+            $table->getIndex('url');
+        } catch (SchemaException $e) {
+            if (SchemaException::INDEX_DOESNT_EXIST === $e->getCode()) {
+                return;
+            }
+
+            throw $e;
+        }
```
ausi:
```suggestion
        if (
            !($schema = $event->getSchema())
            || !$schema->hasTable('tl_search')
            || !($table = $schema->getTable('tl_search'))
            || !$table->hasColumn('url')
            || !($column = $table->getColumn('url'))
            || !$table->hasIndex('url')
        ) {
            return;
        }
```

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
```diff
@@ -0,0 +1,127 @@
...
+            $bytesPerChar = 'utf8mb4' === $table->getOption('charset') ? 4 : 3;
+            $maxIndexSize = floor($maxIndexSize / $bytesPerChar);
+        }
+
+        if ($column->getLength() <= $maxIndexSize) {
+            return;
+        }
+
+        // Set the length
+        $column->setLength($maxIndexSize);
```
ausi, line 62: Should we trigger a warning here?

```suggestion
        $column->setLength($maxIndexSize);

        trigger_error(sprintf('The length of the tl_search.url column was reduced to %s because your database does not support longer UNIQUE indexes.'), E_USER_WARNING);
```

fritzmg: Not sure, we don't do that for the `tl_files` index either currently.

ausi: That’s a different story IMO because we only reduce the size on the (non-unique-)index there (which should not change the behavior).

fritzmg: True. But the warning makes the install tool unusable in debug mode.

fritzmg: How about a log entry instead at least?

ausi: Would be fine for me too.
A deprecation might also work?

fritzmg: > A deprecation might also work?

So increasing the minimum requirements for Contao 5.0? Would make sense to me too.

Toflar:
Just a general thought here: Doesn't every index have that issue? I mean why is it limited to the search table? I think every developer that requires an index on a varchar(whateverlength) wants the index to include as many chars as possible for the best performance, no?

fritzmg:
We could easily expand this to check all unique indexes and not just `tl_search.url`. However I would consider that a new feature, as anyone creating their own SQL definition is otherwise responsible for the appropriate varchar length themselves.

ausi:
> Just a general thought here: Doesn't every index have that issue?

Only `UNIQUE` indexes have.

> I think every developer that requires an index on a varchar(whateverlength) wants the index to include as many chars as possible for the best performance, no?

This is already the case. Contao reduces the index length automatically for too long indexes in https://github.com/contao/contao/blob/720a4e134d8446bdb528bce3a0c0fbe1d22a8d49/core-bundle/src/Doctrine/Schema/DcaSchemaProvider.php#L272

***

IMO we should not do this automatically for all fields, as it reduces the length of the field itself (not the index length) and therefore could result in data loss or other issues.

Toflar:
What I'm trying to say is that this is a general problem. It's not related to the URL and if we can fix that for everybody somehow, then we should. So how can we just reduce the `tl_search.url` field then, if that ends up in data loss? That makes no sense to me.

fritzmg:
It would be a problem (in all cases) if `innodb_large_prefix` is suddenly turned off for an already existing database that already contains data.

Toflar:
So I'm trying to summarize this: We want to store long URLs which is why we increased `tl_search.url` to `2048`. And now we have a problem because this leads to an error if `innodb_large_prefix` is not enabled, right? So what we do is we reduce the `tl_search.url` field in order not to run into this error? But then we cannot store the long URLs which we have to? How is that a solution? I don't understand this :D

fritzmg:
> So I'm trying to summarize this: We want to store long URLs which is why we increased `tl_search.url` to `2048`. And now we have a problem because this leads to an error if `innodb_large_prefix` is not enabled, right? So what we do is we reduce the `tl_search.url` field in order not to run into this error?

Correct.

> But then we cannot store the long URLs which we have to? How is that a solution? I don't understand this :D

We don't "have to". If during crawling a page with a long URL is tried to be saved to the database, an exception will occur which will be logged by the crawler to the debug log and the crawler will continue on with the next URL. So without using `innodb_large_prefix` you lose the ability to index pages with long URLs - which is the case with or without this PR.

Without this PR though, the install tool and `contao:migrate` will fail when executing the `ALTER TABLE` query to increase `tl_search.url` to `2048`.

Toflar:
Yeah but that's what I'm saying. It doesn't solve the problem. Imho what should be done:

- Adjust `tl_search.url` to `text`
- Remove the index from `tl_search.url`
- Introduce `tl_search.hash` that contains the sha256 hash of the url
- Add the unique index on the `hash` column

That would allow for very long URLs even without `innodb_large_prefix` while keeping the search index storage small and efficient.

fritzmg:
> It doesn't solve the problem.

This PR is only meant to solve the problem that `contao:migrate` is now not working, if `innodb_large_prefix` is disabled.

> Imho what should be done:

Sounds good to me, but should we do that as a new feature in 5.x or as a bugfix in 4.9?

ausi:
> Imho what should be done:

I think we should keep it as it is now. I’d prefer to deprecate `innodb_large_prefix=0` and drop support for it in Contao 5.0.

Toflar:
I'm okay with that (although I think the hash variant should be chosen anyway no matter the database requirements).

ausi:
> It doesn't solve the problem.

It still allows URLs with a length up to 767 (on systems with disalbed large prefix). 
Our previous limit of 255 was enougth for 99% of our users too, so I think this would not be a problem in the real world.

zonky2:
I could fix it with:

```php
// contao/dca/tl_search.php
$GLOBALS['TL_DCA']['tl_search']['fields']['url']['sql'] = "varchar(767) COLLATE ascii_bin NOT NULL default ''";
```

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
fritzmg: I have added the deprecation now. However, the message is a bit clunky and I am also not sure about the tests. In order to accept deprecations I had to add `@group legacy`, which seems wrong.

ausi: It is fine IMO and we have many `@group legacy` tests that are not fully legacy.

ausi (dismissed):

leofeyer (dismissed):

fritzmg:
~@leofeyer you have removed the essential deprecation message. In Contao 5 we won't support setups without `innodb_large_prefix` anymore and thus this should be mentioned somewhere.~

// edit: nvm, you only changed the expected deprecation in the tests :).

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
```diff
@@ -0,0 +1,106 @@
...
+            || !$table->hasIndex('url')
+        ) {
+            return;
+        }
+
+        // Get maximum index size for this table
+        $maxIndexSize = $this->getMaximumIndexSize($table);
+
+        // Reduce maximum length if collation is not "ascii_bin"
+        if ('ascii_bin' !== $column->getPlatformOption('collation')) {
```
leofeyer, line 50: How can this ever be true? Our column definition contains `COLLATE ascii_bin`.

fritzmg: The SQL definition of the field can be arbitrarily adjusted by extensions, application customisations or directly on the database level. The listener cannot rely on the original column definition.

leofeyer: Of course it can. Do you really think you can foresee every possible way to misconfigure the system?

fritzmg: We check what we can check... why wouldn't we? I don't really understand why you would not want to include this check.

leofeyer: Case A: You install Contao → the database server is misconfigured → it fails → you don‘t know why.
Case B: You install Contao → it works → you change the field definition → it fails → you know why.

We should add a check for case A but certainly not for case B. If you actively misconfigure the system, we cannot help you.

fritzmg: The check in question also concerns case A because it also depends on the database server configuration whether or not assuming a char length of 1 byte leads to an issue.

ausi: I think these checks are relevant, because of problems with the length, some users were forced to make modifications to this field. And since we changed this to `ascii_bin` quite recently the possibility of it being set to `utf8` is quite high I think.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4629, opened 2022-05-08
Remove search tables if default indexer is disabled

### LINKED ISSUE #3326: Dynamically add/remove the search tables
From https://github.com/contao/contao/pull/3283#issuecomment-899292222:
> Shouldn't we just not offer this option instead in the back end or automator CLI ?

Hide the "clear search index" task from the list when the indexer service is not available.

/cc @aschempp

fritzmg:
This removes the `tl_search` tables from the schema, if the default indexer is disabled.


fritzmg:
Not sure about this now, as we have a public `Search::indexPage` method which is independent from the default indexer and always expects these tables to be present.

Toflar:
We might add a check in `Search::indexPage()` to not do anything if the default indexer is disabled. Wdyt @ausi?

ausi:
I don’t see the point in deleting tl_search tables.
A `Search::indexPage()` method that does nothing would be very confusing IMO.

If we really want to go that way, I’d favor triggering a deprecation notice in Contao 4.13 if `Search::indexPage()` is called with a disabled indexer and throw an exception in 5.0.

fritzmg:
Btw. does anyone have a clue why the functional tests fail here? It seems the `contao.search.default_indexer.enable` parameter is `false` there and thus this listener removes the table which in turn makes the `Search::indexPage` calls in the tests fail.

ausi:
> It seems the `contao.search.default_indexer.enable` parameter is `false` there

I think i disabled it in https://github.com/contao/contao/commit/9e53aebaadeb9129684c758eb1ff310e9163d550#diff-a0cc02b41a487282e06723b0520b1d54e1cef64f025bfb330e53fc6dcd0db9db to speed up the functional tests.
But we could try to reenable it and see if this caused an actual speedup or not (AFAIR disabling it was a wild guess).

fritzmg:
There is indeed a difference. Stats from my system:

| | avg (sec) |
| --- | ---:|
| indexer disabled: | 13.834 |
| indexer enabled: | 15.403 |

leofeyer:
A unit tests is missing and the CI chain fails. Changes look good, though.

leofeyer:
A unit tests is missing and the CI chain fails. Changes look good, though.

leofeyer:
@fritzmg Are you still working on this one?

fritzmg:
Depends on what @contao/developers prefer. To fix the CI chain we would need to enable the indexer for the functional tests again. However, this will slow the them down as shown above. So I am not sure if this is worth it (and @ausi also has [doubts](https://github.com/contao/contao/pull/4629#issuecomment-1120800937)).

fritzmg:
As discussed in our Contao Slack Call on July 28, 2022, we do not want to put more effort into this.

Issue #4513, opened 2022-04-19
Deprecation cleanups for Contao 5


Toflar:
### Description

- [x] 1. Cleanup `$GLOBALS['TL_CRON']` and all its usages (@fritzmg) https://github.com/contao/contao/pull/4706
- [x] 2. Cleanup `$GLOBALS['TL_HOOKS']['getSearchablePages']` and all its usages (@toflar) https://github.com/contao/contao/pull/4578
- [x] 3. Cleanup `Backend::findSearchablePages()` (@toflar) https://github.com/contao/contao/pull/4578
- [x] 4. Cleanup deprecated localconfig keys (@ausi) https://github.com/contao/contao/pull/4729
- [x] 5. Cleanup `$GLOBALS['TL_CROP']` and remove legacy image resizing modes (@ausi) https://github.com/contao/contao/pull/4562
- [x] 6. Cleanup `$GLOBALS['TL_HOOKS']['getCountries']` and `countries.php` (@ausi) https://github.com/contao/contao/pull/4563
- [x] 7. Cleanup `$GLOBALS['TL_HOOKS']['getLanguages']` and `languages.php` (@ausi) https://github.com/contao/contao/pull/4564
- [x] 8. Cleanup `$GLOBALS['TL_HOOKS']['addLogEntry']` (@bytehead) https://github.com/contao/contao/pull/4701
- [x] 9. Cleanup deprecations in `AbstractPickerProvider` (@bytehead) https://github.com/contao/contao/pull/4703
- [ ] 10. Cleanup `reloadPagetree` and `fileTree` `Ajax.php` actions (@ausi)
- [x] 11. Cleanup `$GLOBALS['TL_HOOKS']['addFileMetaInformationToRequest']` and `Backend::addFileMetaInformationToRequest()` @Toflar https://github.com/contao/contao/pull/4585
- [x] 12. Cleanup deprecations in user classes related to security (@bytehead) https://github.com/contao/contao/pull/4343
- [x] 13. Cleanup `Frontend.php` (@m-vo) https://github.com/contao/contao/pull/4541
- [x] 14. Cleanup `onrestore_callback` (`Versions.php`) (@toflar) https://github.com/contao/contao/pull/4649
- [x] 15. Cleanup `timezones.php` (see task 23)
- [x] 16. Cleanup `ContentMarkdown` (@toflar) https://github.com/contao/contao/pull/4584
[...]

PR #4634, opened 2022-05-09
Remove TL_CRON usage from calendar-bundle

fritzmg:
See #4513 

Will be superseded by #4459 eventually though.


ausi (dismissed):

Toflar (dismissed):

m-vo (dismissed):

calendar-bundle/src/Resources/config/services.yml
```diff
@@ -7,6 +7,13 @@ services:
             calls:
                 - [setFramework, ['@contao.framework']]
 
+    contao_calendar.cron.generate_feeds:
```
leofeyer, line 10: Why are we not using a service tag like we do here?

https://github.com/contao/contao/blob/5b8c1d7beb54452b9a2451722b2742d1c026dca7/core-bundle/src/Cron/PurgePreviewLinksCron.php#L22

calendar-bundle/src/Resources/config/services.yml
```diff
@@ -7,6 +7,13 @@ services:
...
+        class: Contao\CalendarBundle\Cron\GenerateFeedsCron
+        arguments:
+            - '@contao.framework'
+        tags:
+            - { name: contao.cronjob, interval: daily }
```
leofeyer: Why are we not using a service tag like we do here?

https://github.com/contao/contao/blob/5b8c1d7beb54452b9a2451722b2742d1c026dca7/core-bundle/src/Cron/PurgePreviewLinksCron.php#L22

fritzmg: Currently we sometimes tag our services via annotations and sometimes we tag them via YAML. There is no directive at the moment what is preferred.

As we do not use service auto registration and thus we need to define our services via YAML anyway I thought may be we should also just use YAML for tagging anyway?

m-vo: I prefer using annotations/attributes for those things that users are also using in their applications. Because then it's easier for people to understand what they need to do in their code.

fritzmg: Not sure what you mean, in what way would you be using the cronjobs in your application for example?

m-vo: Maybe for importing things from an API, sending time based reminders, cleaning up custom tables…  All of that would be possible with a custom command and a real cronjob of course, but why not use our framework that makes it super easy to implement :slightly_smiling_face:.

Edit: This way I can also add things without having to adjust the server config. Which is nice, if you deploy to multiple instances.

fritzmg: Oh, you mean in case someone is looking for examples when it comes to cronjobs etc.? idk., I would also make the argument that it might be useful to see the intervals of all cronjobs in one file (per bundle at least). Although I suppose that should be a command (#4640 😁).

m-vo: > Oh, you mean in case someone is looking for examples when it comes to cronjobs etc.?

Yes. And opinionated of course. :upside_down_face:

fritzmg: But in that case we should expand that directive to hooks, callbacks and all event listeners, no?

leofeyer: Please add the `@CronJob` annotation in the new files, so they match the existing cron job files. Expanding the directive to other services should be done in a different pull request.

fritzmg:
Failing test is basically unrelated. I will provide a separate PR for that.

fritzmg:
The reverse order test would be fixed once https://github.com/contao/contao/pull/4656 is merged upstream.

m-vo (dismissed):

Toflar (dismissed):

ausi (dismissed):

aschempp:
Code works imho, not sure we need it with the feeds controllers though 🤷‍♂️

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4637, opened 2022-05-09
Remove TL_CRON usage from newsletter-bundle

fritzmg:


newsletter-bundle/src/Cron/PurgeSubscriptionsCron.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        /** @var NewsletterRecipientsModel $subscription */
+        foreach ($subscriptions as $subscription) {
+            $subscription->delete();
+        }
+
+        if (null !== $this->logger) {
+            $this->logger->info('Purged the unactivated newsletter subscriptions');
+        }
```
ausi:
```suggestion
        $this->logger?->info('Purged the unactivated newsletter subscriptions');
```

newsletter-bundle/src/Resources/contao/classes/Newsletter.php
```diff
@@ -932,9 +932,13 @@ public function updateAccount()
 
 	/**
 	 * Purge subscriptions that have not been activated within 24 hours
+	 *
+	 * @deprecated Deprecated since Contao 5.0, to be removed in Contao 6.0.
 	 */
 	public function purgeSubscriptions()
 	{
+		trigger_deprecation('contao/newsletter-bundle', '5.0', 'Calling "%s::%s()" has been deprecated and will no longer work in Contao 6.0.', __CLASS__, __METHOD__);
```
ausi:
```suggestion
		trigger_deprecation('contao/newsletter-bundle', '5.0', 'Calling "%s()" has been deprecated and will no longer work in Contao 6.0.', __METHOD__);
```

fritzmg: Good catch, there are actually a few places in the core where `__CLASS__` and `__METHOD__` is used in the deprecation message like this.

m-vo (dismissed):
LGTM besides the two annotations by @ausi.

newsletter-bundle/src/Resources/contao/classes/Newsletter.php
ausi: Yes, I used it like that in the past I think 🙈

fritzmg: https://github.com/contao/contao/pull/4643 🙃

ausi (dismissed):

m-vo (dismissed):

Toflar (dismissed):

ausi (dismissed):

newsletter-bundle/src/Cron/PurgeSubscriptionsCron.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    public function __construct(private ContaoFramework $framework, private LoggerInterface|null $logger)
+    {
+    }
+
+    public function __invoke(): void
+    {
+        $this->framework->initialize();
+
+        $subscriptions = $this->framework->getAdapter(NewsletterRecipientsModel::class)->findExpiredSubscriptions();
```
aschempp, line 33: I was confused by this variable name. They are not subscriptions but recipients, maybe you want to change that 🙃

fritzmg: I used that variable name because of the method name.

leofeyer (changes requested):
Same here. Changes are OK but the unit test is never run unless you set them up.

fritzmg:
> Same here. Changes are OK but the unit test is never run unless you set them up.

Not sure what you mean, the newsletter-bundle already has tests and it's already registered in the `phpunit.xml.dist`?

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

Issue #4625, opened 2022-05-07
Deprecate Automator?


fritzmg:
### Description

As the usage of `$GLOBALS['TL_CRON']` is deprecated, we need to clean up all legacy cron job of the core and register them as a properly tagged service instead (see also #4513). However, the remaining `contao/core-bundle` cron jobs are all based on the methods in the Automator. Most of them are fairly simple, the question is though how should we handle it? Because all the Automator methods are also available via the `contao:automator` command. For example:

There is a daily cronjob that purges the expired member registrations via `Automator::purgeRegistrations`. When moving the cronjob to a tagged service should we:

1. Create a service that still calls `Automator::purgeRegistrations` and thus leaves the method and thus also the `contao:automator` command intact (so Automator will not be deprecated in this case).
2. Create a service that executes the code directly and deprecate the method in `Automator`. _This would remove it from `contao:automator` in Contao 6.0 though._
3. Create a service that executes the code directly, remove deprecate the method from `Automator`, but also implement a command that executes the appropriate code.

Toflar:
I think by introducing a general command that allows to run **any** cronjob that was registered via the cron framework, we could completely deprecate the Automator stuff :) 

And then definitely migrate away from Automator to isolated cronjobs.

fritzmg:
The automator is not only used for cronjobs though. Things like `purgeXmlFiles`, `purgeSearchIndex`, `purgeImageCache`, `purgePreviewCache`, `purgePageCache` are used in various places in the framework.

Toflar:
Ideally, those should be services. So I don't think we can fix this in 5.0 but it will be a process to migrate during 5.x and then remove in 6.0.

PR #4640, opened 2022-05-09
Allow single and forced cron job execution

fritzmg:
This will allow you to:

- execute specific, single cron jobs
- force execution of one or all cron jobs, disregarding their last execution time

Examples:

```bash
vendor/bin/contao-console contao:cron "Contao\CoreBundle\Cron\PurgeExpiredDataCron::onHourly"
vendor/bin/contao-console contao:cron --force
vendor/bin/contao-console contao:cron --force "Contao\CoreBundle\Cron\PurgeExpiredDataCron::onHourly"
```

This is useful for testing and in case you want to execute certain specific cron jobs separately in your cron tab. It is also useful in case specific code only exists as a cron job - e.g. in case we move certain Automator methods only into a cron job (see #4625).

This PR could be further expanded (now or later on) with a `--list` argument that would display all available cron jobs and allow you to choose one to execute interactively.


ausi:
> This PR could be further expanded (now or later on) with a `--list` argument that would display all available cron jobs

That would be pretty important for this feature to offer better usability. IMO it should be added directly to this PR.

m-vo (dismissed):

core-bundle/src/Cron/Cron.php
```diff
@@ -87,7 +111,7 @@ public function run(string $scope): void
                 // Check if the cron should be run
                 $expression = CronExpression::factory($interval);
 
-                if (null !== $lastRunDate && $now < $expression->getNextRunDate($lastRunDate)) {
+                if (!$force && null !== $lastRunDate && $now < $expression->getNextRunDate($lastRunDate)) {
```
m-vo, line 122: Using the `--force` flag should ideally get a unit test, so that we don't mess it up in the future. :see_no_evil:

fritzmg:
> That would be pretty important for this feature to offer better usability. IMO it should be added directly to this PR.

Yes, but what I am not sure about is:

* Should `--list` also interactively ask you which cron job to execute (none, specific, all)?
* Or should there be a separate option called `--select` for example for that?

Also `--list --format=json` should probably output the list in JSON, so that the Contao Manager could in theory allow you to execute specific cron jobs.

ausi:
> Should `--list` also interactively ask you which cron job to execute (none, specific, all)?

IMO not, it should only list them.

> Or should there be a separate option called `--select` for example for that?

Don’t know if this is actually needed. I’d be happy without it.

fritzmg:
There is now also a `contao:cron:list` command:

core-bundle/src/Resources/config/commands.yml
```diff
@@ -42,6 +42,11 @@ services:
         class: Contao\CoreBundle\Command\CronCommand
         arguments:
             - '@contao.cron'
+
```
m-vo:
```suggestion

```
:upside_down_face:

core-bundle/src/Resources/config/commands.yml
```diff
@@ -42,6 +42,10 @@ services:
         class: Contao\CoreBundle\Command\CronCommand
         arguments:
             - '@contao.cron'
+    contao.command.cron_list:
```
ausi, line 46:
```suggestion

    contao.command.cron_list:
```

ausi (dismissed):

core-bundle/src/Cron/Cron.php
fritzmg: Unit test added.

core-bundle/src/Command/CronListCommand.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class CronListCommand extends Command
+{
+    protected static $defaultName = 'contao:cron:list';
+    protected static $defaultDescription = 'Lists all available cron jobs and their intervals.';
+
+    public function __construct(protected Cron $cron)
```
m-vo: Any reason why this is `protected`?

core-bundle/src/Cron/Cron.php
```diff
@@ -41,10 +41,42 @@ public function addCronJob(CronJob $cronjob): void
         $this->cronJobs[] = $cronjob;
     }
 
+    /**
+     * @return array<CronJob>
```
m-vo: Should we use `list<…>` from now on?

m-vo (dismissed):

core-bundle/src/Command/CronListCommand.php
fritzmg: Nope :D

core-bundle/src/Cron/Cron.php
fritzmg: Yeah, I wasn't sure (also since my IDE could not deal with `list<…>` 😁 )

fritzmg: As it is non-standard I am not sure whether we should use it. And as I said, my IDE at least won't recognize it and thus it would not be able to know the type of the array elements (which would of course be inconvenient for me).

@contao/developers wdyt?

ausi: I’d prefer `list<…>` and a fix for your IDE ☺️

ausi (dismissed):

Toflar (dismissed):
Fancy stuff! ❤️

aschempp:
Nice! Do we really need another command for the job list? Imho a job list choice like the automator command offered at runtime would be more useful, including a "run all" option (which would also be triggered on `--no-interaction`)

fritzmg:
> Do we really need another command for the job list?

I preferred doing it this way in the end, since we also have `contao:install:lock`, `contao:install:unlock`, `contao:user:list`, `contao:user:create` etc.

> Imho a job list choice like the automator command offered at runtime would be more useful

That was discussed (see above), but ultimately I decided it's not worth it. But if desired, I can still implement a `--select` option for `contao:cron`.

> including a "run all" option (which would also be triggered on `--no-interaction`)

That is just executing `contao:cron` without an argument in the current state of this PR (and needs to stay this way for BC).

aschempp:
>> including a "run all" option (which would also be triggered on --no-interaction)

> That is just executing contao:cron without an argument in the current state of this PR (and needs to stay this way for BC).

I meant if the `--force` option provides a command list by default, using `--no-interaction` would execute all jobs

fritzmg:
Hm, that's not really something that I would expect the `--force` option to do. `--force` is just meant to force run all cron jobs (or a single one, if given). That makes the most sense to me at least.

aschempp:
it feels the same as migrations. `contao:migrate` asks you whether to execute XY and on `--no-interaction` does run instantly 🤷

fritzmg:
> it feels the same as migrations. `contao:migrate` asks you whether to execute XY and on `--no-interaction` does run instantly 🤷

Of course, but for backwards compatibility reasons I would not change this behavior for `contao:cron` now.

core-bundle/src/Command/CronListCommand.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        $table = new Table($output);
+        $table
+            ->setHeaders(['Name', 'Interval'])
+            ->setRows($list)
+        ;
+        $table->render();
+
+        return 0;
```
m-vo:
```suggestion
        return Command::SUCCESS;
```

m-vo (dismissed):

m-vo (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4641, opened 2022-05-09
Fix missing PurgePreviewLinksCron registration

fritzmg:
The `PurgePreviewLinksCron` cron job was never actually registered as a service - and thus it was never executed so far.

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -110,6 +110,11 @@ services:
             - '@contao.framework'
             - '@database_connection'
 
+    contao.cron.purge_preview_links:
+        class: Contao\CoreBundle\Cron\PurgePreviewLinksCron
+        arguments:
+            - '@database_connection'
+
     contao.csrf.token_manager:
         class: Contao\CoreBundle\Csrf\ContaoCsrfTokenManager
         public: true


ausi (approved):

Toflar (approved):

m-vo (approved):

leofeyer:
Thank you @fritzmg.

leofeyer:
Thank you @fritzmg.

leofeyer:
Ah, I have merged this one too quickly. We are actually using a service tag to register the job:

https://github.com/contao/contao/blob/e1dcfa24e79e5307dc17cc225691d2a561f3acd5/core-bundle/src/Cron/PurgePreviewLinksCron.php#L22

fritzmg:
> Ah, I have merged this one too quickly. We are actually using a service tag to register the job:

No, the service still needs to be registered. The Service tag just _tags_ the service, it does _not_ register the service.

fritzmg:
Generally we should may be figure out a way to catch unregistered services as this is not the first time this happened. Previously we at least checked in the `ContaoCoreExtensionTest` (etc.) whether a service is registered, however we were instructed to not do that anymore. May be we can make an automatic check whether classes actually exist in any of the service definitions? At least for certain namespaces (or ignoring certain namespaces/folders).

m-vo:
IMHO the `ContaoCoreExtensionTest` tests were pretty much useless as they were more or less a direct copy of what was in the `services.yml` file. Functional tests would probably be the way to go for container related stuff, because they would also verify everything that happens during container compilation (compiler passes, validation and so on).

fritzmg:
> IMHO the `ContaoCoreExtensionTest` tests were pretty much useless as they were more or less a direct copy of what was in the `services.yml` file.

Yes, but at least it made you check whether you actually did register your service.

> Functional tests would probably be the way to go for container related stuff, because they would also verify everything that happens during container compilation (compiler passes, validation and so on).

True.. time for a functional tests for cronjobs then I guess 😬 🙈

PR #4647, opened 2022-05-10
Improve the news archive permission labels

Toflar:
Otherwise one could expect that this relates to the permissions **within** an archive (so creating and deleting news) which is not what it does.


aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #4648, opened 2022-05-10
Implement news archive permissions as a voter

Toflar:
This demonstrates the power of the new permission voter foundation in the current 5.x branch 😎 
I kept this PR as small as possible (and currently without tests) in order for us to review and maybe optimize the permissions themselves again.

~~As you can see, we can save a ton of callbacks that were solely in place to adjust the icons.
With the current state of things, however, the operations are now hidden completely (which is okay for global operations as well as buttons but it's __probably__ not for operations).~~

~~Here's an example of how it now looks like if you do not have access to deleting or creating a news archive. The delete and copy operations are just hidden:~~


m-vo:
Awesome!

>  however, the operations are now hidden completely (which is okay for global operations as well as buttons but it's probably not for operations).

I know the discussions about this, but IMHO it is perfectly fine to homogenize the behavior here and keep things as simple as possible implementation wise. And if for instance the alignment of buttons becomes an issue (i.e. grouping same icons in a single column), this should be solved independently on the UI side later on.

aschempp (changes requested):
Requesting changes since I discussed changes with @Toflar 🙃

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function __construct(private ContaoFramework $contaoFramework, private Security $security)
+    {
+    }
+
+    protected function supports(string $attribute, $subject)
+    {
+        return $subject instanceof DataContainerSubject &&
+            'tl_news_archive' === $subject->table &&
+            \in_array($attribute, self::SUPPORTED_ATTRIBUTES, true);
```
aschempp:
```suggestion
            str_starts_with($attribute, ContaoCorePermissions::DC_ACTION_PREFIX);
```

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,55 @@
...
+class NewsArchiveAccessVoter extends Voter
+{
+    private const SUPPORTED_ATTRIBUTES = [
+        ContaoCorePermissions::DC_ACTION_CREATE,
+        ContaoCorePermissions::DC_ACTION_EDIT,
+        ContaoCorePermissions::DC_ACTION_COPY,
+        ContaoCorePermissions::DC_ACTION_DELETE,
+        ContaoCorePermissions::DC_ACTION_VIEW,
+    ];
+
```
aschempp:
```suggestion
```

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,55 @@
...
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token)
+    {
+        return match ($attribute) {
+            ContaoCorePermissions::DC_ACTION_CREATE,
+            ContaoCorePermissions::DC_ACTION_COPY => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES),
+            ContaoCorePermissions::DC_ACTION_EDIT,
+            ContaoCorePermissions::DC_ACTION_VIEW => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->id),
+            ContaoCorePermissions::DC_ACTION_DELETE => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->id) &&
+                $this->security->isGranted(ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES),
+        };
```
aschempp, line 46:
```suggestion
            default: false,
        };
```

Should continue to deny access to any unknown DC_ACTION if anything has been added to DC_Table. Imho a third-party developer should not reuse that prefix for a custom `key=` function.

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,56 @@
...
+            'tl_news_archive' === $subject->table &&
+            \in_array($attribute, self::SUPPORTED_ATTRIBUTES, true);
+    }
+
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token)
+    {
+        return match ($attribute) {
+            ContaoCorePermissions::DC_ACTION_CREATE => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES),
+            ContaoCorePermissions::DC_ACTION_COPY => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES) &&
+                $this->security->isGranted(ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES),
```
aschempp:
```suggestion
                $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->id),
```

aschempp (changes requested):

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,49 @@
...
+use Contao\CoreBundle\Security\ContaoCorePermissions;
+use Contao\CoreBundle\Security\DataContainer\DataContainerSubject;
+use Contao\NewsBundle\Security\ContaoNewsPermissions;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+use Symfony\Component\Security\Core\Security;
+
+class NewsArchiveAccessVoter extends Voter
+{
+    public function __construct(private ContaoFramework $contaoFramework, private Security $security)
```
aschempp:
```suggestion
    public function __construct(private Security $security)
```

aschempp (changes requested):
last request 🎉

m-vo (dismissed):

m-vo (dismissed):

Toflar:
@aschempp and I decided, that we're going to rework the "view" part again because voters do not seem to be the right tool here. Voters can basically only say "yes, no or I don't care". However, buttons/operations etc. can have a lot more (think of our toggle eye icon which has more than just two states - it not only depends on your permission but also on whether e.g. something is published or not). @aschempp is working on a POC here.

That being said, this PR is ready to be merged. I was about to provide tests here but they wouldn't provide any added value at all. I would only test if the voter passes on the calls to `Security` which is the very same as already written in the code...

aschempp (approved):
🥳

m-vo:
> That being said, this PR is ready to be merged. I was about to provide tests here but they wouldn't provide any added value at all. I would only test if the voter passes on the calls to Security which is the very same as already written in the code...

Agreed. Though I'd love to see browser based functional tests for these things in the future. Maybe @MDevster wants to give a cypress setup a try at some point. :wink:

leofeyer:
We still need a solution to show disabled icons instead of no icon at all. Will add "up for discussion".

aschempp:
> We still need a solution to show disabled icons instead of no icon at all. 

That is not part of this PR. Removing the icons has been removed from this PR, it now only rewrites the `checkPermission` method. Operations are still checked through `button_callback` here, but will be updated in https://github.com/contao/contao/pull/4663

fritzmg:
> We still need a solution to show disabled icons instead of no icon at all. Will add "up for discussion".

Parts of that have already been merged in https://github.com/contao/contao/pull/4495

aschempp:
> Parts of that have already been merged in https://github.com/contao/contao/pull/4495

yeah but as said in https://github.com/contao/contao/pull/4648#issuecomment-1124610881 that was actually wrong and will be reverted (https://github.com/contao/contao/pull/4663 is the first step towards that)

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -0,0 +1,48 @@
...
+            str_starts_with($attribute, ContaoCorePermissions::DC_ACTION_PREFIX);
+    }
+
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token)
+    {
+        return match ($attribute) {
+            ContaoCorePermissions::DC_ACTION_CREATE => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES),
+            ContaoCorePermissions::DC_ACTION_COPY => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES) &&
+                $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->id),
+            ContaoCorePermissions::DC_ACTION_EDIT,
```
aschempp, line 41: This should also check for `USER_CAN_EDIT_FIELDS_OF_TABLE` shouldn't it?

aschempp (changes requested):

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
Toflar: No, I don't think this is correct. This should be a general voter because it's always the same. You are never allowed to edit if you cannot edit any field of a table. That's always true for all tables.
I wonder if this should be checked in the `DC_` itself even, wdyt?

aschempp: yeah, it even applies to CREATE or COPY permissions, since you would not be able to edit the record afterwards. I think having a generic voter for this would be an awesome improvement since that currently is not correctly checked. But that should be a separate PR 😊

aschempp (dismissed):

news-bundle/src/Resources/contao/dca/tl_news_archive.php
```diff
@@ -300,72 +296,6 @@ public function checkPermission()
...
 
 		$GLOBALS['TL_DCA']['tl_news_archive']['list']['sorting']['root'] = $root;
-		$security = System::getContainer()->get('security.helper');
-
-		// Check permissions to add archives
-		if (!$security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES))
-		{
-			$GLOBALS['TL_DCA']['tl_news_archive']['config']['closed'] = true;
-			$GLOBALS['TL_DCA']['tl_news_archive']['config']['notCreatable'] = true;
-			$GLOBALS['TL_DCA']['tl_news_archive']['config']['notCopyable'] = true;
```
leofeyer, line 310: We cannot remove this, can we? The voter only replaces the `switch` statement below but not this logic.

news-bundle/src/Resources/contao/dca/tl_news_archive.php
```diff
@@ -300,72 +296,6 @@ public function checkPermission()
...
-
-				if (Input::get('act') == 'deleteAll' && !$security->isGranted(ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES))
-				{
-					$session['CURRENT']['IDS'] = array();
-				}
-				else
-				{
-					$session['CURRENT']['IDS'] = array_intersect((array) $session['CURRENT']['IDS'], $root);
-				}
-				$objSession->replace($session);
```
leofeyer, line 359: This is also not handled by the voter and thus must not be removed.

leofeyer (changes requested):

news-bundle/src/Resources/contao/dca/tl_news_archive.php
Toflar: It's not needed anymore. This was a - very dangerous and delicate - logic to ensure you selected only the IDs you have access to for the edit multiple logic. Now we properly check for the permissions in the respecitve methods so it doesn't matter anymore what's in your session which is **A LOT** safer :)

news-bundle/src/Resources/contao/dca/tl_news_archive.php
Toflar: There's no reason to mark the whole table as `closed` because the buttons will not be shown anymore based on the permssions. /cc @aschempp

news-bundle/src/Resources/contao/dca/tl_news_archive.php
aschempp: Explained in a different way: before each record is edited, we ask the permission voter if it is allowed and skip if not. So even if the ID is in the session, it cannot be edited because the voter checks for it. 😎

news-bundle/src/Resources/contao/dca/tl_news_archive.php
aschempp: Being `closed`, `notCreatable` or `notCopyable` is synonymous with the respective voters (`DC_ACTION_CREATE`/`DC_ACTION_COPY`), so it is unnecessary to have both.

aschempp: additionally, voters can be overridden using our PriorityStrategy, but these flags cannot relieably

ausi: We have to check all occurences of `'closed'` in DC_Table and DataContainer then I think. 
For example the buttons “save and new” and “save and duplicate” are shown now even if the user is not allowed to create an archive.

Not sure if this is in the scope of this PR though.

aschempp: Refactoring buttons and global operations is a follow-up task after https://github.com/contao/contao/pull/4663. But this PR is the foundation to actually see the permissions being applied.

Toflar: We'll have to apply permissions on the buttons as well, we know that. However, configuring the whole DC dynamically to be `notCopyable` or `notCreatable` based on permissions has always been an artchitectural design issue. These are settings and should be immutable by definition. We can bring back the current behaviour by checking e.g. the `copy` permissions for all the IDs listed in order to restore the current behavior. It would also allow for future UX improvements like e.g. adding `data-` attributes to every single entry and using JS in order to give the users hints about e.g. why there's no copy button or which ones of them can be copied but not deleted etc.

ausi: Totally agree 👍
Removing this here is correct IMO, and permissions for “save and new”, “save and duplicate” and the like is for another PR.

ausi (dismissed):
❤️

_(The failing unit test seems to be unrelated to this PR)_

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4651, opened 2022-05-10
Remove the deprecated "show to guests only" function

leofeyer:


ausi:
Should we add a migration that enables the protected checkbox and adds “guests” to the selected groups?

leofeyer:
Good question. @contao/developers WDYT?

fritzmg:
I am in favour for a migration.

Toflar:
Yes, of course! Otherwise you cannot update.

m-vo (dismissed):
Code looks good to me. Otherwise I'm also on #teammigration. :smile:

leofeyer:
Migration added in b52404f63a7083c6f2f3b9a9f17e356a22829edd.

fritzmg (commented):
Should we create an `AbstractGuestsMigration` with a `protected static array $tables`? Through `Controller::isVisibleElement` this functionality was available for arbitrary tables.

core-bundle/src/Migration/Version500/GuestsMigration.php
```diff
@@ -0,0 +1,91 @@
...
+                LIMIT 1
+            ");
+
+            foreach ($values as $id => $value) {
+                $groups = StringUtil::deserialize($value, true);
+                $groups[] = '-1';
+
+                $data = [
+                    'protected' => '1',
+                    'groups' => serialize($groups),
```
ausi:
```suggestion
                    '`groups`' => serialize($groups),
```
This needs to be escaped AFAIK.
https://github.com/doctrine/dbal/blob/2deb141194c962c62a19f53edfdca482190eb5b0/src/Connection.php#L694

fritzmg: Shouldn't we use `$this->connection->quoteIdentifier()` here though? Although we currently have a mixture of directly using the back tick and using `$this->connection->quoteIdentifier()` in the core.

core-bundle/src/Migration/Version500/GuestsMigration.php
```diff
@@ -0,0 +1,91 @@
...
+
+use Contao\CoreBundle\Migration\AbstractMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+use Contao\StringUtil;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @internal
+ */
+class GuestsMigration extends AbstractMigration
```
Toflar: @contao/developers this is another one that looks like it should be an abstract that I can use in my extension. Maybe I did not name it `groups` but I think many extensions have the same issue, don't you think?

ausi (dismissed):

aschempp (changes requested):
Please add an abstract migration, this will certainly be needed by other extensions (e.g. Isotope eCommerce).

core-bundle/src/Migration/Version500/GuestsMigration.php
leofeyer: Abstract migration added in e6647793df121d1a855c9921c3cbad086613ed53.

Toflar (dismissed):

aschempp (dismissed):
❤️

core-bundle/src/Migration/Version500/AbstractGuestsMigration.php
```diff
@@ -0,0 +1,94 @@
...
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+
+        foreach ($this->getTables() as $table) {
+            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)['guests'])) {
```
ausi:
```suggestion
            if (
                !$schemaManager->tablesExist([$table])
                || !isset($schemaManager->listTableColumns($table)['guests'], $schemaManager->listTableColumns($table)['groups']) 
            ) {
```

core-bundle/src/Migration/Version500/AbstractGuestsMigration.php
```diff
@@ -0,0 +1,94 @@
...
+
+        return false;
+    }
+
+    public function run(): MigrationResult
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+
+        foreach ($this->getTables() as $table) {
+            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)['guests'])) {
```
ausi:
```suggestion
            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)['guests']) || !isset($schemaManager->listTableColumns($table)['groups'])) {
```

core-bundle/src/Migration/Version500/AbstractGuestsMigration.php
leofeyer: Changed in 1adcfd7706a4e03e32e645548a0286ef675ce075.

core-bundle/src/Migration/Version500/AbstractGuestsMigration.php
leofeyer: Changed in 1adcfd7706a4e03e32e645548a0286ef675ce075.

ausi (approved):

m-vo (approved):

PR #4653, opened 2022-05-10
Drop support for an empty `tl_content.ptable` column

leofeyer:
@contao/developers What do you think about adding a migration that automatically changes empty columns to `tl_article`?

```sql
UPDATE tl_content SET ptable='tl_article' WHERE ptable='';
```

I‘m not sure if that would be too destructive. On the other hand, an empty column is treated as `tl_article` in Contao 4, so the value cannot really have a different meaning, can it?

### DIFF
--- a/core-bundle/src/Migration/Version500/EmptyPtableMigration.php
+++ b/core-bundle/src/Migration/Version500/EmptyPtableMigration.php
@@ -35,11 +35,7 @@ class EmptyPtableMigration extends AbstractMigration
 
         $test = $this->connection->fetchOne("SELECT TRUE FROM tl_content WHERE ptable='' LIMIT 1");
 
-        if (false !== $test) {
-            return true;
-        }
-
-        return false;
+        return false !== $test;
     }
 
     public function run(): MigrationResult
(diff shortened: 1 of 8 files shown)


fritzmg:
I am in favour of migrating empty to `tl_article`.

Toflar:
Yes, migration!

m-vo (dismissed):
LGTM, but I'd also add a migration.

ausi (dismissed):

leofeyer:
Migration added in f9a9e21fe118768055802f07e17f55708e68afae.

m-vo (dismissed):

fritzmg (dismissed):

core-bundle/src/Migration/Version500/EmptyPtableMigration.php
```diff
@@ -0,0 +1,51 @@
...
+            return false;
+        }
+
+        $test = $this->connection->fetchOne("SELECT TRUE FROM tl_content WHERE ptable='' LIMIT 1");
+
+        if (false !== $test) {
+            return true;
+        }
+
+        return false;
```
ausi:
```suggestion
        return false !== $test;
```

ausi (dismissed):
Just a nitpick 🙈

m-vo (approved):

PR #4656, opened 2022-05-10
Unset TL_CONFIG in ContaoTestCase::tearDown()

fritzmg:
Currently the CI fails with many tests that extend from `ContaoTestCase` and use the `mockContaoFramework` method, due to the global state change watcher which detects that `$GLOBALS['TL_CONFIG']` was set.

This is because `mockContaoFramework` will automatically load the default `TL_CONFIG` configuration (and thus `$GLOBALS['TL_CONFIG']` gets populated).

This PR would fix that by unsetting the global in the `tearDown` in this base class.

### DIFF
--- a/test-case/src/ContaoTestCase.php
+++ b/test-case/src/ContaoTestCase.php
@@ -52,6 +52,14 @@ abstract class ContaoTestCase extends TestCase
         unset(self::$tempDirs[$key]);
     }
 
+    protected function tearDown(): void
+    {
+        // Unset TL_CONFIG, since we populate it in loadDefaultConfiguration
+        unset($GLOBALS['TL_CONFIG']);
+
+        parent::tearDown();
+    }
+
     /**
      * Returns the path to the temporary directory and creates it if it does not yet exist.
      */
--- a/test-case/src/ContaoTestCase.php
+++ b/test-case/src/ContaoTestCase.php
@@ -54,7 +54,7 @@ abstract class ContaoTestCase extends TestCase
 
     protected function tearDown(): void
     {
-        // Unset TL_CONFIG, since we populate it in loadDefaultConfiguration
+        // Unset TL_CONFIG, since we populate it in loadDefaultConfiguration (#4656)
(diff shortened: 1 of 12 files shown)


leofeyer:
The global state watcher has been added in Contao 4.13, therefore this pull should be rebased.

test-case/src/ContaoTestCase.php
```diff
@@ -63,6 +63,14 @@ public static function tearDownAfterClass(): void
         unset(self::$tempDirs[$key]);
     }
 
+    protected function tearDown(): void
+    {
+        // Unset TL_CONFIG, since we populate it in loadDefaultConfiguration
+        unset($GLOBALS['TL_CONFIG']);
```
leofeyer, line 58: Maybe you can also delete the `tearDown()` methods in the tests that do nothing more than this?

fritzmg:
> The global state watcher has been added in Contao 4.13, therefore this pull should be rebased.

Yes, but shouldn't we clean up the global state in 4.9 regardless?

Toflar:
I think @ausi should review this as he's the author of the feature.

ausi (dismissed):

test-case/src/ContaoTestCase.php
fritzmg: Yes, will do once this is merged and then merged upstream.

leofeyer: Please rebase this PR to the 4.13 branch and make the changes here.

m-vo (dismissed):

test-case/src/ContaoTestCase.php
fritzmg: Done in 6cf9fe27b7224529121dbb6ba0c42b3c6c6cb964

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4444, opened 2022-04-01
Modern fragments

m-vo:
Very early draft state with lots of unrelated changes due to cherry picks mainly to be able to share pieces of code.
(and to be completely honest: I just wanted to reserve this nice PR number for now :smile:)

I'll update the description as soon as I cleaned things up.

PR #4657, opened 2022-05-10
Allow adding document scoped content from within Twig templates

m-vo:
This PR introduces a new Twig tag `{% add %}`, that works like this:

```twig
{% add to body %}
  <div>I will appear at the end of the body</div>
{% endadd %}

{% add to head %}
  {# Add an additional head tag #}
  <link rel="stylesheet" href="my-fancy-style.css">
{% endadd %}
```

Currently the content inside the tag will be rendered and simply put into `$GLOBALS['TL_HEAD']` and `$GLOBALS['TL_BODY']` but the idea is to use the response context in the future as soon it supports these things. :slightly_smiling_face:.

Optionally, you can define an identifier to make sure the content gets output only once when used multiple times - this will basically use a key in the above global arrays:

```twig
{% for i in range(0, 10) %}
  {% add 'foo' to body %}
    <script type="application/javascript">
      console.log('I will only be executed once!');
    </script>
  {% endadd %}
{% endfor %}
```

This feature for instance allows to easily add small inline scripts that are then output only once - e.g. for displaying a splash screen, scrolling the viewport, and so on. In #4444 I already stumbled about multiple of these scenarios.

### DIFF
--- a/core-bundle/src/Twig/Extension/ContaoExtension.php
+++ b/core-bundle/src/Twig/Extension/ContaoExtension.php
@@ -21,6 +21,8 @@ use Contao\CoreBundle\Twig\Inheritance\TemplateHierarchyInterface;
 use Contao\CoreBundle\Twig\Interop\ContaoEscaper;
 use Contao\CoreBundle\Twig\Interop\ContaoEscaperNodeVisitor;
 use Contao\CoreBundle\Twig\Interop\PhpTemplateProxyNodeVisitor;
+use Contao\CoreBundle\Twig\ResponseContext\AddTokenParser;
+use Contao\CoreBundle\Twig\ResponseContext\DocumentLocation;
 use Contao\CoreBundle\Twig\Runtime\FigureRendererRuntime;
 use Contao\CoreBundle\Twig\Runtime\InsertTagRuntime;
 use Contao\CoreBundle\Twig\Runtime\LegacyTemplateFunctionsRuntime;
@@ -99,6 +101,8 @@ final class ContaoExtension extends AbstractExtension
             // additionally support the Contao template hierarchy
             new DynamicExtendsTokenParser($this->hierarchy),
             new DynamicIncludeTokenParser($this->hierarchy),
+            // Add a parser for the Contao specific "add" tag
+            new AddTokenParser(self::class),
         ];
     }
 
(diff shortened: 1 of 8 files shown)


Toflar (dismissed):
🥰

ausi (dismissed):

core-bundle/src/Twig/ResponseContext/AddNode.php
```diff
@@ -0,0 +1,69 @@
...
+    public function compile(Compiler $compiler): void
+    {
+        // if ($this->env->isDebug()) { ob_start(); } else { ob_start(static function () { return ''; }); }
+        // try {
+        //     <sub-compiled content>
+        //     $__contao_document_content = ob_get_contents();
+        // } finally { ob_end_clean(); }
+        // $this->extensions["Contao\\…\\ContaoExtension"]->addDocumentContent(
+        //     '<identifier>', $__contao_document_content, Contao\…\DocumentLocation::<location>
+        // );
```
leofeyer, line 48: This should be removed. 😉

m-vo: Why? We have the preview what the node outputs in every of our nodes. :slightly_smiling_face: 

https://github.com/contao/contao/blob/5.x/core-bundle/src/Twig/Interop/PhpTemplateProxyNode.php
https://github.com/contao/contao/blob/5.x/core-bundle/src/Twig/Interop/PhpTemplateParentReferenceNode.php

leofeyer: Ah, I see. It looked like you forgot to remove it. My bad.

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -240,6 +244,34 @@ protected function renderTwigSurrogateIfExists(): string|null
...
 
+    /**
+     * @see \Contao\CoreBundle\Twig\ResponseContext\AddNode
+     * @see \Contao\CoreBundle\Twig\ResponseContext\AddTokenParser
+     *
+     * @internal
+     */
+    public function addDocumentContent(string|null $identifier, string $content, DocumentLocation $location): void
+    {
+        // TODO: This should make use of the response context in the future.
```
leofeyer, line 255: Why can it not make use of it now?

leofeyer (dismissed):

core-bundle/src/Twig/Extension/ContaoExtension.php
m-vo: Because this is currently not implemented yet in the response context. But @Toflar or myself will probably do this in the foreseeable future. :slightly_smiling_face:

leofeyer:
Thank you @m-vo.

leofeyer (approved):

core-bundle/src/Twig/Extension/ContaoExtension.php
Toflar: I'll probably give it a shot even for 5.0 - depending on the permissions stuff.

PR #4662, opened 2022-05-12
Deprecate the move operation

aschempp:
I didn't even know this one existed 😂 
It generates non-JS up/down buttons and was probably the predecessor to our drag handle in parent view mode. Since I'm going to update the operations support following https://github.com/contao/contao/pull/4648 imho we don't need to support this anymore.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -993,6 +993,8 @@ abstract class DataContainer extends Backend
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


m-vo (approved):
Change looks good to me. Not sure if we use this anywhere, saw this the first time now. :smile:

leofeyer:
This is an accessibility feature. How are people supposed to move elements otherwise if they cannot use drag and drop?

aschempp:
I don't know, but the move operations do not exist on any of the current DCA 😉

ausi (approved):

PR #4661, opened 2022-05-12
Add missing try catch blocks

Toflar:
Things that we missed in the previous permission PRs.

PR #4663, opened 2022-05-12
Automatically generate DCA operations with permission checks

aschempp:
This PR changes a few things following https://github.com/contao/contao/pull/4648 and our general new permission checks. Per discussion with @Toflar, it does not make sense to vote on an operation, because the core operations always equal to a DC action permission.

However, we _cannot_ automate action permission checks based on the operation name, because we just cannot know if that's what is actually intended. So @Toflar and I came up with the fact that it is **also** annoying to re-define the operations **for every DCA file**! This PR now solves both those issues 😎 

1. it automatically adds the default operations, but lets you customise sorting and buttons if necessary. More of this will be visible in follow-up PRs or once we have more permission checks liek https://github.com/contao/contao/pull/4648
2. ~it adds a new `access_callback` that allows for flexible permission checks, which checks the default permissions on our default operation~
3. ~if access is denied on the callbacks, the button is automatically rendered as a grey icon with our default convention. `Image::getHtml` already handles the case if the icon does not exist and simply does not generate anything 💪~

You might want to read the individual commits, since the full diff can be a bit missleading.

Requires https://github.com/contao/contao/pull/4648 and https://github.com/contao/contao/pull/4661 to fully work on `tl_news_archive`. Also should have https://github.com/contao/contao/pull/4662 since that's now dropped here.


core-bundle/src/DataContainer/OperationsFactory.php
```diff
@@ -0,0 +1,150 @@
...
+        if ($hasPtable || $isTreeMode) {
+            $operations['copy'] = [
+                'href' => 'act=paste&amp;mode=copy',
+                'icon' => 'copy.svg',
+                'attributes' => 'onclick="Backend.getScrollOffset()"',
+                'access_callback' => [$this->accessCallback(ContaoCorePermissions::DC_ACTION_COPY)],
+            ];
+
+            if ($isTreeMode) {
+                // TODO: how to we check permissions for that?
```
Toflar: Guess it needs to be a Core permission because it's a core action?

Toflar (commented):
Nice!! 

I would probably use the `loadDataContainer` hook for the new `OperationsFactory` because it's basically what it is and it would allow me to register a listener in a higher or lower prio to adjust accordingly :) 
Also, we should get rid of all the view related permissions here, no? Buttons and global operations are still voters here and only operations are changed.

core-bundle/src/DataContainer/OperationsFactory.php
aschempp: Nah, I just need to figure out common grounds. `tl_page` checks if there are child records on the page, so maybe thats what the "permission check" should do. As well as the usual "copy" checks. But since I only focussed on `tl_news_archive` you provided this is not finished yet 🙃

aschempp:
Interesting idea with the hook. Regarding the other view permissions, I intentionally did not include everything to not make it so confusing 😅

aschempp:
Updated for new action permissions and `children` operation. After merging https://github.com/contao/contao/pull/4709 locally it correctly renders all default operations and correctly disables operations for news archives based on the official demo data.
 - Deleting archives is disallowed by https://github.com/contao/contao/pull/4648
 - Duplicated archive is disallowed because user is not allowed to create (https://github.com/contao/contao/pull/4648)
 - Editing the archive is disabled because no fields are allowed (https://github.com/contao/contao/pull/4709)

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
```diff
@@ -0,0 +1,222 @@
...
+use Contao\CoreBundle\Security\DataContainer\UpdateAction;
+use Contao\CoreBundle\ServiceAnnotation\Hook;
+use Contao\DataContainer;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Security\Core\Security;
+
+/**
+ * @internal
+ *
+ * @Hook("loadDataContainer", priority=200)
```
Toflar, line 28: Any reason for 200? Usually it's any power of 2 :) So 256 maybe?

core-bundle/src/DataContainer/DataContainerOperation.php
```diff
@@ -0,0 +1,99 @@
...
+use Contao\StringUtil;
+
+class DataContainerOperation implements \ArrayAccess
+{
+    private string|null $html = null;
+
+    /**
+     * @internal
+     */
+    public function __construct(private readonly string $name, private array $operation, private readonly array $record, private readonly DataContainer $dataContainer)
```
Toflar: Do we need `$record` now or can you use `$dataContainer->getCurrentRecord($id)`?

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
```diff
@@ -0,0 +1,222 @@
...
+            $field = $name;
+        }
+
+        return $field;
+    }
+
+    private function isGranted(string $actionClass, string $table, DataContainerOperation $operation): bool
+    {
+        $subject = match ($actionClass) {
+            CreateAction::class => new CreateAction($table),
```
Toflar, line 206: If the operation has a parent, we have to pass `['pid' => $pid]` here, like we do in DC_Table as well.

Toflar (changes requested):
This PR is definitely finished and can be merged into 5.0.
However, I found one bug so I'm requesting changes here.

Great work - this is the perfect addition to the new permissions and makes handling them in operations so much easier! ❤️

Toflar:
Oh, an PHPStan needs to be fixed.

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
aschempp: no reason, just a randomly "high" number 😂

core-bundle/src/DataContainer/DataContainerOperation.php
aschempp: I didn't want to make the object depend on a `DataContainer` object. But if we agree operations are always tied to a `$dc` then I could just pass that 🙃

aschempp:
can anyone tell me how to fix PHPStan?

core-bundle/src/DataContainer/DataContainerOperation.php
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\DataContainer;
+
+use Contao\DataContainer;
+use Contao\StringUtil;
+
+class DataContainerOperation implements \ArrayAccess
```
m-vo, line 21:
```suggestion
/**
 * @implements \ArrayAccess<string, mixed>
*/
class DataContainerOperation implements \ArrayAccess
```

core-bundle/src/DataContainer/DataContainerOperation.php
```diff
@@ -0,0 +1,99 @@
...
+
+        // Add the key as CSS class
+        if (str_contains($attributes, 'class="')) {
+            $attributes = str_replace('class="', 'class="'.$name.' ', $attributes);
+        } else {
+            $attributes = ' class="'.$name.'" '.$attributes;
+        }
+        $operation['attributes'] = $attributes;
+
+        $this->operation = $operation;
```
m-vo, line 55:
```suggestion
```
This should already be assigned to the implicitly declared property.

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
```diff
@@ -0,0 +1,222 @@
...
+            UpdateAction::class => new UpdateAction($table, $operation->getRecord()),
+            DeleteAction::class => new DeleteAction($table, $operation->getRecord())
```
m-vo:
```suggestion
            DeleteAction::class => new DeleteAction($table, $operation->getRecord()),
            default => throw new \InvalidArgumentException(sprinf('Invalid action class "%s".', $actionClass)),
```

m-vo:
> can anyone tell me how to fix PHPStan?

Left some suggestions. :relaxed:

core-bundle/src/DataContainer/DataContainerOperation.php
m-vo: I'm not sure about `mixed` - took that from your `offsetGet()` return type. If it should actually be `string`, we should change it at both places and maybe throw an error in `offsetSet()` if something else was set.

IMHO a `/** @param array<…> $operation /*` type should also be added to the constructor.

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
aschempp: I don't think that's necessary, we're the only ones calling that method and we know our options

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
aschempp: I just checked on that. The copy (and copychilds) operation check for this permission, but they do not know the new PID yet. We only do that in the core on the _create new after_ automatic operation in DC_Table. We need to add that check to the paste buttons, but that's a different PR since it is a bug in core atm.

core-bundle/src/DataContainer/DataContainerOperation.php
aschempp: changed in b5d188873ed4e1146dcfad08f2a98154235c3ae3

core-bundle/src/EventListener/DataContainer/DefaultOperationsListener.php
Toflar: Can you create a separate issue for this then, so we don't forget about it? Then this is resolved for now :)

aschempp: see https://github.com/contao/contao/pull/4947

Toflar (commented):
Coverage decrease due to missing tests. Rightfully so, I think we need some here :)

aschempp:
PHPStan errors seem wrong 🤷

Toflar:
Not wrong, it just cannot detect your changes on `$GLOBALS` in the listener. So this warning has to be suppressed at those very places :) Using `@phpstan-ignore-line` - wdyt, @m-vo?

core-bundle/tests/EventListener/DataContainer/DefaultOperationsListenerTest.php
```diff
@@ -0,0 +1,390 @@
...
+        $this->assertSame($operation['icon'], $icon);
+
+        if ($hasCallback) {
+            $this->assertArrayHasKey('button_callback', $operation);
+            $this->assertIsCallable($operation['button_callback']);
+
+            $ref = new \ReflectionFunction($operation['button_callback']);
+
+            $this->assertSame(1, $ref->getNumberOfParameters());
+            $type = $ref->getParameters()[0]->getType();
```
m-vo, line 384:
```suggestion

            /** @var \ReflectionNamedType $type */
            $type = $ref->getParameters()[0]->getType();
```

m-vo:
PHPStan not able to track the global state is a common problem unfortunately. 

But you can trick it by moving the assert to a helper function:
```php

/**
 * @param list<string> $expected
 * @param array<int|string, mixed> $actual
 */
private function assertSameOperationKeys(array $expected, array $actual): void
{
    $this->assertSame($expected, array_keys($actual));
}

…
    $this->assertSameOperationKeys(['edit', 'foo', 'show', 'delete'], $operations);
…
```

Toflar (dismissed):

leofeyer (approved):

PR #4664, opened 2022-05-12
Add conditional setting/unsetting to the HtmlAttributes class

m-vo:
This PR allows to conditionally set/unset boolean `HtmlAttributes` without leaving the fluid interface. This is especially helpful in Twig:

before:
```twig
{% set attributes = attrs().addClass('foo') %}
{% if condition %}
  {% set attributes = attributes.set('data-property') %}
{% endif %}
```

after:
```twig
{% set attributes = attrs().addClass('foo').setIf('data-property', condition) %}
```

Same goes for unsetting, where we now have a `unsetIf()` method.

### DIFF
--- a/core-bundle/src/String/HtmlAttributes.php
+++ b/core-bundle/src/String/HtmlAttributes.php
@@ -90,7 +90,7 @@ class HtmlAttributes implements \Stringable, \JsonSerializable, \IteratorAggrega
      */
     public function set(string $name, \Stringable|bool|int|string|null $value = true, \Stringable|bool|int|string|null $condition = true): self
     {
-        if (!$condition) {
+        if (!$condition || ($condition instanceof \Stringable && !(string) $condition)) {
             return $this;
         }
 
--- a/core-bundle/src/String/HtmlAttributes.php
+++ b/core-bundle/src/String/HtmlAttributes.php
@@ -111,6 +111,21 @@ class HtmlAttributes implements \Stringable, \JsonSerializable, \IteratorAggrega
         return $this;
     }
 
+    /**
+     * Enable the property $name if the $condition is truthy.
+     */
+    public function setIf(string $name, \Stringable|bool|int|string|null $condition): self
+    {
+        if ($condition) {
+            $this->set($name);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Set the property $name to $value if the value is truthy.
+     */
     public function setIfExists(string $name, \Stringable|bool|int|string|null $value): self
     {
         if (!empty($value)) {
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

ausi:
Should it be possible to set a value too with this method?

m-vo:
> Should it be possible to set a value too with this method?

I had that thought as well. Maybe it's worth it.

My alternative idea was to  add a third parameter `$condition = true` to the `set()`/`unset()` method:
```twig
attributes.set('data-feature', condition=feature_enabled)
attributes.set('data-feature', '', feature_enabled)
```

Would you prefer this?

m-vo:
> Would you prefer this?

Well, thanks for asking. I do. Here are the changes: 2c6f14d8f97307fa537c0d52d497895b0896650f

Toflar (dismissed):

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -85,9 +85,15 @@ public function mergeWith(self|iterable|string|null $attributes = null): self
      * Sets a property and validates the name. If the given $value is false the
      * property will be unset instead. All values will be coerced to strings,
      * whereby null and true will result in an empty string.
+     *
+     * If a falsy $condition is specified, the method is a no-op.
      */
-    public function set(string $name, \Stringable|bool|int|string|null $value = true): self
+    public function set(string $name, \Stringable|bool|int|string|null $value = true, \Stringable|bool|int|string|null $condition = true): self
     {
+        if (!$condition) {
```
ausi: Do we need to handle the Stringable objects here?

ausi:
```suggestion
        if (!$condition || ($condition instanceof \Stringable && !(string) $condition)) {
```

core-bundle/src/String/HtmlAttributes.php
```diff
@@ -120,9 +129,18 @@ public function setIfExists(string $name, \Stringable|bool|int|string|null $valu
...
-    public function unset(string $key): self
+    /**
+     * Unset the property $name.
+     *
+     * If a falsy $condition is specified, the method is a no-op.
+     */
+    public function unset(string $name, \Stringable|bool|int|string|null $condition = true): self
     {
-        unset($this->attributes[$key]);
+        if (!$condition) {
```
ausi:
```suggestion
        if (!$condition || ($condition instanceof \Stringable && !(string) $condition)) {
```

core-bundle/tests/String/HtmlAttributesTest.php
```diff
@@ -248,6 +248,39 @@ public function testSetAndUnsetProperties(): void
...
     }
 
+    public function testSetAndUnsetConditionalProperties(): void
+    {
+        $attributes = new HtmlAttributes();
+
+        $attributes->set('data-feature1', condition: null);
+        $attributes->set('data-feature2', condition: false);
+        $attributes->set('data-feature3', condition: 0);
+        $attributes->set('data-feature4', condition: '');
```
ausi:
```suggestion
        $attributes->set(
            'data-feature5',
            condition: new class implements \Stringable 
            { 
                public function __toString(): string 
                {
                    return '';
                }
            },
        );
```

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4665, opened 2022-05-12
Make the template element more flexible

doishub:
Currently, the defined values must always be traversed with a foreach or similar. For more flexibility I would rather pass the data array in the following structure. This way I would have direct access to my keys and could use a more complex dom structure in my template if needed.

We could also consider passing the flattened array in addition and keeping the old structure as well?

### DIFF
--- a/core-bundle/src/Controller/ContentElement/TemplateController.php
+++ b/core-bundle/src/Controller/ContentElement/TemplateController.php
@@ -22,7 +22,12 @@ class TemplateController extends AbstractContentElementController
 {
     protected function getResponse(Template $template, ContentModel $model, Request $request): Response
     {
-        $template->data = StringUtil::deserialize($model->data, true);
+        $arrData = StringUtil::deserialize($model->data, true);
+
+        $template->data = array_combine(
+            array_column($arrData, 'key'),
+            array_column($arrData, 'value'),
+        );
 
         return $template->getResponse();
     }
--- a/core-bundle/src/Controller/ContentElement/TemplateController.php
+++ b/core-bundle/src/Controller/ContentElement/TemplateController.php
@@ -22,11 +22,10 @@ class TemplateController extends AbstractContentElementController
 {
     protected function getResponse(Template $template, ContentModel $model, Request $request): Response
     {
-        $arrData = StringUtil::deserialize($model->data, true);
-
-        $template->data = array_combine(
-            array_column($arrData, 'key'),
-            array_column($arrData, 'value'),
+        $template->data = StringUtil::deserialize($model->data, true);
+        $template->keys = array_combine(
+            array_column($template->data, 'key'),
+            array_column($template->data, 'value'),
         );
 
         return $template->getResponse();
--- a/core-bundle/src/Controller/ContentElement/TemplateController.php
+++ b/core-bundle/src/Controller/ContentElement/TemplateController.php
@@ -22,11 +22,15 @@ class TemplateController extends AbstractContentElementController
 {
     protected function getResponse(Template $template, ContentModel $model, Request $request): Response
     {
-        $template->data = StringUtil::deserialize($model->data, true);
+        $data = StringUtil::deserialize($model->data, true);
+
         $template->keys = array_combine(
-            array_column($template->data, 'key'),
-            array_column($template->data, 'value'),
+            array_column($data, 'key'),
+            array_column($data, 'value')
         );
+
+        // Backwards compatibililty
+        $template->data = $data;
 
         return $template->getResponse();
     }
--- a/core-bundle/src/Resources/contao/templates/elements/ce_template.html5
+++ b/core-bundle/src/Resources/contao/templates/elements/ce_template.html5
@@ -2,10 +2,10 @@
 
 <?php $this->block('content'); ?>
 
-  <?php foreach ($this->data as $entry): ?>
+  <?php foreach ($this->keys as $key => $value): ?>
     <dl>
-      <dt><?= $entry['key'] ?></dt>
-      <dd><?= $entry['value'] ?></dd>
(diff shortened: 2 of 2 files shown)


doishub:
👀 https://community.contao.org/de/showthread.php?82955-Individuelles-Template-%E2%80%93-Werte-ausgeben

m-vo:
@ausi Is this intentional or should we consider this a bug? :see_no_evil: 

I would have expected the key value pairs from the back end to end up in the template like `['key1' => 'value2', 'key2' => 'value2']` but maybe I'm missing something.

core-bundle/src/Controller/ContentElement/TemplateController.php
```diff
@@ -22,7 +22,12 @@ class TemplateController extends AbstractContentElementController
 {
     protected function getResponse(Template $template, ContentModel $model, Request $request): Response
     {
-        $template->data = StringUtil::deserialize($model->data, true);
+        $arrData = StringUtil::deserialize($model->data, true);
+
+        $template->data = array_combine(
+            array_column($arrData, 'key'),
+            array_column($arrData, 'value'),
+        );
```
m-vo:
```suggestion
        $data = StringUtil::deserialize($model->data, true);

        $template->data = array_combine(
            array_column($data, 'key'),
            array_column($data, 'value'),
        );
```

fritzmg:
> I would have expected the key value pairs from the back end to end up in the template like ['key1' => 'value2', 'key2' => 'value2'] but maybe I'm missing something.

Same here. I assumed this was the case (never really looked at the implementation 🙈) until I saw the thread on the forum.

aschempp:
I would actually expect it to be `$this->key` like any other template property? 😬

doishub:
> I would actually expect it to be $this->key like any other template property? 😬

I could imagine that the idea of the key was intended more as a label. This would explain the initial foreach 😁

`Name` -> `Kevin`
`Nachname` -> `Jones`

ausi:
> Is this intentional or should we consider this a bug? 🙈

I think it was intentional, as the key/value wizard can be used with duplicate keys I thought passing the data as it is made most sense.

Because of backwards compatibility I don’t think we can change the structure of how `->data` looks like. 
But we could additionally set the values on the template data itself like so:

```php
foreach ($template->data as ['key' => $key, 'value' => $value]) {
    $template->$key = $value;
}
```

This would then work with `$this->key` as @aschempp expected it and it would be backwards compatible.
Not sure if I like that though.

m-vo:
> This would then work with $this->key as @aschempp expected it and it would be backwards compatible.
Not sure if I like that though.

I'm very :-1: on this one. In the new controllers it was a major goal to remove all possible variable collisions from the context. If someone would for instance set a key `attributes` or `headline`, there won't be a `HtmlAttributes` instance or a headline array in the context anymore and the content element base template would probably break.

I think I'd favor having a second variable set like so and would introduce this change as a bugfix in 4.13:
```php
$data = StringUtil::deserialize($model->data, true);

$template->data = $data;
$template->keys = array_combine(array_column($data, 'key'), array_column($data, 'value'));
```
This way you can access all duplicates as well as a simplified output:

```twig
{% for datum in data|filter(d => d.key == 'foo') % }
    {{ data.value }}
{% endfor %}

{{ keys.foo }}
```

m-vo:
By the way: another reason why collections must always be scoped is that you otherwise cannot iterate them. :upside_down_face:

doishub:
I have adopted @m-vo `s  suggestion for now.

fritzmg:
One question remains: should we backport this to ~~4.9~~ 4.13?

doishub:
+1 Should not cause any problems from my point of view. There is no more BC break.

m-vo:
> One question remains: should we backport this to 4.9?

IMHO yes. But 4.13, because this element doesn't exist in 4.9 yet. :slightly_smiling_face: 
@leofeyer It's not a BC break btw.

leofeyer:
We can adjust the `ce_template.html5` template if we make sure that custom versions of it still work, which is the case IIRC. Therefore I vote for applying the initial changes again:

```php
  <?php foreach ($this->keys as $key => $value): ?>
    <dl>
      <dt><?= $key ?></dt>
      <dd><?= $value ?></dd>
    </dl>
  <?php endforeach; ?>
```

m-vo:
Good idea to also adjust the default template. :+1:

doishub:
Makes sense. Done. 🙃

core-bundle/src/Controller/ContentElement/TemplateController.php
```diff
@@ -23,6 +23,10 @@ class TemplateController extends AbstractContentElementController
     protected function getResponse(Template $template, ContentModel $model, Request $request): Response
     {
         $template->data = StringUtil::deserialize($model->data, true);
+        $template->keys = array_combine(
+            array_column($template->data, 'key'),
+            array_column($template->data, 'value'),
+        );
```
leofeyer:
```suggestion
        $data = StringUtil::deserialize($model->data, true);

        $template->keys = array_combine(
            array_column($data, 'key'),
            array_column($data, 'value')
        );

        // Backwards compatibililty
        $template->data = $data;
```

This way we don‘t forget to remove the BC layer in Contao 5.

leofeyer (dismissed):

m-vo (changes requested):
This should go into 4.13.

leofeyer:
I‘ve just realized that this is a PR for Contao 5. We do not need a BC layer at all then. Or is the PR supposed to be rebased on the `4.13` branch?

**Edit:** @m-vo was faster than me. 😄

m-vo (approved):

ausi (approved):

leofeyer:
Thank you @doishub.

PR #4666, opened 2022-05-12
tl_content.ptable is mandatory now

Toflar:
Follow up to https://github.com/contao/contao/pull/4653.

I think we don't want to default to `''` anymore now, do we? It's a mandatory definition now, imho.


m-vo:
Should we default to `article` instead, so that we don't break sparse inserts?

leofeyer:
> Should we default to `article` instead, so that we don't break sparse inserts?

Yes, except the default value has to be `tl_article`.

Toflar:
Voilà :)

leofeyer:
Thank you @Toflar.

leofeyer (approved):

Issue #4278, opened 2022-03-15
Dumper::formatValueForDump() Argument #1 ($value) must be of type ?string, int given


fritzmg:
### Affected version(s)

4.13.x-dev

### Description

When using `contao:migrate (or `backup:create`) the following error happens in Contao `4.13.x-dev`:

```
[ERROR] Contao\CoreBundle\Doctrine\Backup\Dumper::formatValueForDump():      
         Argument #1 ($value) must be of type ?string, int given, called in   
         vendor\contao\contao\core-bundle\src\Doctrine\Backup\Dumper.php on line 122 
```

The column in question is `tl_article.id` and the value to be formated is an integer of value `1`. This error was introduced in https://github.com/contao/contao/commit/52753e05c0efdcb6a6cb28e2c75665d544ece208

/cc @Toflar

Toflar:
How come you have an integer there? Can you debug this? Database values are always transferred as strings.

leofeyer:
Turns out to be a new behavior of the PDO driver in PHP 8.1: https://github.com/doctrine/dbal/issues/5228

PR #4670, opened 2022-05-13
Fix type warning in event list

fritzmg:
This PR fixes the following warning:

```
ErrorException:
Warning: date() expects parameter 2 to be int, string given

  at vendor\contao\calendar-bundle\src\Resources\contao\modules\ModuleEventlist.php:213
```

`$day` will be a timestamp - but its type is `string`, since `Events::getAllEvents` will use `$objEvents->startTime` for `$intStart` for `$this->addEvent` - and database values are always returned as a string ([except in PHP 8.1+](https://github.com/contao/contao/issues/4278)).

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -210,7 +210,7 @@ class ModuleEventlist extends Events
 						continue;
 					}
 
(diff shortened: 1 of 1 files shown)


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #4671, opened 2022-05-13
Modern fragments: text content elements (part 1)

m-vo:
Here is the first PR for the new content elements that consist of fragment controllers, Twig templates and integration tests. The DCA side remains untouched, to opt into the legacy elements you can add a respective entry in the `contao/config/config.php` at every time (see `UPGRADE.md`). 

I will gradually move things out of #4444 into PRs like this one, so that we can easier discuss them. I'll still keep that PR open (draft) because it gives a better overview. I picked a few very simple controllers for this first round, because I'm sure there will be requested changes affecting the other things. After that, the others will probably follow in short sequence. :slightly_smiling_face:

#### Covered elements in this PR
We start with `headline`, `html` and `code` because they are simple but still show some general concepts: 

1) fragment controllers using the new `FragmentTemplate`
1) `content_element/_base.html.twig` as the base template for most content elements
1) reusable components in `component/…`
1) using our custom Twig features (`HtmlAttributes`, the `highlight` or `insert_tag` filter, the `add` tag…)
1) tests …

#### Test case
There is a new test-case class that facilitates testing the new elements. In there we're now actually invoking the controllers and rendering the templates, so that we can be sure all the template hierarchy setup (`extends`/`include`/`embed`, …)  works as well as our functions and filters. :tada: 

To compare the rendered HTML output there is also simple regex based function that minifies the HTML. This is merely a stackoverflow adaption for now because I'm sure @ausi will have a better idea for this. :wink:.

#### Review checklist
It would be really nice to get some input on the structure of the templates, also from non-developers. Here are some ideas to help you get going (some of them may not fully apply to this PR but to the broader picture in #4444):
* Is the markup correct?
* Is the markup accessible?
* Can you easily get your use case working by extending the template and adjusting blocks?
* Do you miss something that cannot be achieved by adjusting the template?
* Do the component semantics make sense or should we structure them differently?
* Should we have other defaults?
* Do you understand the templates or would you appreciate comments? <sup>*)</sup>

<sup>*)</sup> This could go as far as [here](https://github.com/contao/contao/blob/0f92002e1115d8776f60440a2c2f7885f83446fa/core-bundle/src/Resources/contao/templates/_new/content_element/text.html.twig) where I added type hints for all variables in a general block above the template as well as some annotations or just the latter (which is also less work to maintain). In general, comments in Twig do not hurt performance, because they will simply be stripped when compiling.

There is a lot more to say, but let's wait for your input first. :relaxed:


core-bundle/src/Resources/contao/templates/_new/component/_headline.html.twig
```diff
@@ -0,0 +1,11 @@
+<{{ headline.tagName }}>
+{%- block inner %}
+    {%- apply spaceless %}
+        {% if as_overview %}
```
Toflar: This `as_overview` is really strange. Overview of what? I found "back end wildcard" a lot better. Not because I knew what a "wildcard" was but it had a name and by learning what it is, we understood what is meant. "overview" is just a general term...

core-bundle/src/Resources/contao/templates/_new/ide-twig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "namespaces": [
```
Toflar: Tip: remove that from this PR, it will cause discussions that are unrelated to the PR again.

Toflar (commented):
Very good idea to split this PR ;) It makes things a lot easier to review :) I commented on the overview thing which I would like to have fixed first but other than that, the PR looks very promising!

core-bundle/src/Resources/contao/templates/_new/component/_headline.html.twig
m-vo: IIRC @ausi and I came up with `as_overview` as a way to tell the different "output scopes" apart. Having something like backend_preview/wildcard/overview… would to me imply that this can only work in the back end. Can we maybe find a better term that does not include back end? :slightly_smiling_face: 

`as_summary`? 

I personally like the `as_` prefix, but open for suggestions of course!

Toflar: Naming...so let me try to articulate what this does: We decide between what should be done when actually executing the template (FE) and the "raw" view for the editor in the back end (I agree back end is not good because you might want to render it in FE mode in the BE).
What do you think about `as_editor_view`? It's the view for the one that knows what insert tags etc. are, right?

m-vo: I like `as_editor_view`!

ausi: +1 for `as_editor_view` ❤️

leofeyer: I would like to throw in `is_preview`.

Toflar: It's not a preview though. It's what the user (editor) actually entered in the back end. Hence my `as_editor_view` proposal. It's the one thing that should display the HTML/source exactly as you entered it. No insert tags replacement, no code highlighting etc.

core-bundle/src/Resources/contao/templates/_new/ide-twig.json
m-vo: see d83d3e0aa7c3d2d3f8087ce64427fbc982fecff5

core-bundle/src/Resources/contao/templates/_new/component/_headline.html.twig
m-vo: I went with `as_editor_view` in 1448234be35e3959b83a530868f1aa56bcff1bfd for now,

leofeyer: But why `as_editor_view`? Wouldn‘t you rather say "if this **is** the editor view, render this, otherwise render that"? The variable should be called `is_editor_view` IMHO.

(Not to be mistaken with the PHP 8 attributes, which say "treat this class **as** content element".)

m-vo: My reasoning was output this "as editor view" i.e. a short form of `shouldBeOutputAsEditorView` instead of `isInEditorView` but ultimately I'd be fine with both.

ausi: I’m fine with both too (`as_editor_view`, `is_editor_view`).

Toflar: Same here. I don't care. My problem is the term "overview". I agree with @m-vo that you would say "render as" but also, it's not typical in a template variable :)

m-vo:
Based on feedback and more tested scenarios I updated a few things. Most important are:

* I removed the controller-built `HtmlAttributes` and only constructed them in the templates. This way we do only assume default values in the templates (where they can be changed). We may want to divert from this strategy whenever attributes are closely coupled to the implementation (e.g. think video/image attributes), but I reckon most of the time this should be fine. [f757668ab966935e923d589cf5a3f82da68bc778]

* When passing things on to included templates (in this PR for instance the `component/_headline`), we're now only passing on the respective variable (`headline`) and we're directly passing it - so the array gets the new context. This way it's easier to adjust the component templates and as a bonus you can ship variables to the included component from your controller without having collisions. [2b9f76a0fc638a173f4c92aaaa063067522c16e3]

core-bundle/src/Controller/FrontendModule/AbstractFrontendModuleController.php
```diff
@@ -91,13 +90,12 @@ protected function addDefaultDataToTemplate(FragmentTemplate $template, array $m
...
-            'as_overview' => $asOverview,
+            'as_editor_view' => $asEditorView,
             'data' => $modelData,
             'section' => $section,
             'properties' => $properties,
-            'attributes' => (new HtmlAttributes())
-                ->setIfExists('id', $attributesData[0] ?? null)
-                ->addClass($attributesData[1] ?? '', ...$classes),
+            'html_id' => $attributesData[0] ?? null,
+            'html_classes' => trim(($attributesData[1] ?? '').' '.implode(' ', $classes)),
```
Toflar: I find `html_id` and `html_classes` confusing. I understand why you went for `html_id` because you want to make clear it's https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id and not the database ID of the element itself. However, CSS classes are always "html", so `html_classes` sounds weird.
What do you think about those proposals:

* `element_html_id`
* `element_css_classes`

I know it's longer but I don't think we should try to keep variables as short as possible because it doesn't bring us any advantage. It's a lot better to use descriptive variable names because it helps users to understand what they contain without dumping endlessly.

m-vo: Fully agree and I don't mind the longer names. Thanks for the suggestions! :+1: 
Went with it in e9c749af3e075175a8bab7e35020ca4fa5340d4e.

core-bundle/src/Resources/contao/templates/_new/content_element/html.html.twig
```diff
@@ -0,0 +1,7 @@
+{% if as_editor_view %}
+    {% block overview %}
```
Toflar: What's that block? Is it needed? And if so, it should be renamed, no?

m-vo: Well spotted :smile: I renamed it to `editor_view` in ab781239abc0adc297a273e18d0dbe50fd9c58cc.

IMHO it should exist, so that you can adjust the editor view only. Especially if you extend multiple times.

Toflar (dismissed):
I love many design improvements here:

- I love the fact that all classes etc. are now fully manageable in the templates will make it so much easier to adjust to CSS frameworks
- I like the naming of the variables as they are much more self-explanatory than what we have in our current HTML5 templates
- It's great that the decision as to how to render our "back end wildcard" is now also left to the template. It fixes potential issues with encoding and it makes the process a lot more visible to the user (not mulitple templates, same process everywhere). I also love that it's not named "wildcard" anymore but "editor view" because "wildcard" is just a pretty random word which doesn't describe what it does at all.

Thanks for all the improvements ❤️

Toflar:
I've resolved the discussion regarding the `overview` because I started it and to me it's resolved. But leaving it here as a note: We can still decide between `as_editor_view` and `is_editor_view`. Neither @m-vo nor @ausi or me had any preference so at this point I'd consider this CS.

core-bundle/src/Resources/contao/templates/_new/component/_headline.html.twig
```diff
@@ -0,0 +1,11 @@
+<{{ tagName }}{% block attributes %}{{ attributes|default }}{% endblock%}>
```
ausi:
```suggestion
<{{ tagName }}{% block attributes %}{{ attributes|default }}{% endblock %}>
```

core-bundle/src/Resources/contao/templates/_new/component/_headline.html.twig
```diff
@@ -0,0 +1,11 @@
...
+{%- block inner %}
+    {%- apply spaceless %}
+        {% if as_editor_view %}
+            {% block overview %}{{ text|striptags }}{% endblock %}
```
ausi:
```suggestion
            {% block editor_view %}{{ text|striptags }}{% endblock %}
```

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
```diff
@@ -0,0 +1,34 @@
...
+            </noscript>
+        {% endadd %}
+    {% endblock %}
+{% endblock %}
+
+{% block content %}
+    <pre>
+        {%- apply spaceless %}
+            {% if as_editor_view %}
+                {% block overview %}{{ code }}{% endblock %}
```
ausi:
```suggestion
                {% block editor_view %}{{ code }}{% endblock %}
```

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
```diff
@@ -0,0 +1,26 @@
+{% if not searchable|default(true) %}<!-- indexer::stop -->{% endif %}
```
ausi: Should we assure a newline character before indexer comment? (AFAIK these comments have to be at the start of a line).
```suggestion
{% if not searchable|default(true) %}
<!-- indexer::stop -->{% endif %}
```

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
```diff
@@ -0,0 +1,26 @@
...
+                {% endif %}
+            {% endblock %}
+
+            {% block content %}{% endblock %}
+        {% endblock %}
+    </{{ block('wrapper_tag') }}>
+{% endblock %}
+{% block after %}{% endblock %}
+
+{% if not searchable|default(true) %}<!-- indexer::continue -->{% endif %}
```
ausi:
```suggestion
{% if not searchable|default(true) %}<!-- indexer::continue -->
{% endif %}
```

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
```diff
@@ -0,0 +1,34 @@
...
+            <noscript>
+                <link rel="stylesheet" href="{{ highlighter_css_file }}">
+            </noscript>
+        {% endadd %}
+    {% endblock %}
+{% endblock %}
+
+{% block content %}
+    <pre>
+        {%- apply spaceless %}
```
ausi: Doesn’t this remove whitespace inside the `<pre>` too?
Or do `apply` blocks stop on a `block` boundary?

ausi (dismissed):
Some minor comments and nitpicks ❤️

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
m-vo: see e7a3d98c85ccb0cdc14c0ca9d0656c0417d17380

Should be identical to your suggestion but has the end tags in a new line plus Twig whitespace control, so that it is not output.

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
m-vo: see e7a3d98c85ccb0cdc14c0ca9d0656c0417d17380

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
m-vo: The `spaceless` filter is supposed to remove whitespaces between HTML tags. I thought it respects `code` tags, but it turns out it's implemented quite naively: 

```php
return trim(preg_replace('/>\s+</', '><', $content ?? ''));
```

We should probably use individual whitespace control here. I'll check…

m-vo: see 48690c16cfe0877cc6c1e65e0fbca270e7acf006 (includes some more refactorings for the code template)

m-vo:
Here are the changes since your last visit :wink::

* I refactored the code template. Now the whitespace handling is correct and the response context styles won't be added for the editor view. I also refactored the stylesheet part into a component template. [[`48690c1`](https://github.com/contao/contao/pull/4671/commits/48690c16cfe0877cc6c1e65e0fbca270e7acf006)]
* Followup to:
    > When passing things on to included templates (in this PR for instance the component/_headline), we're now only passing on the respective variable (headline) and we're directly passing it - so the array gets the new context. […]
   
   This now means, that we need to merge in extra variables (like `as_editor_view`) of course. As this is completely context dependent - i.e. only the one including knows what's the meaning - it still feels right to me. I fixed the two occurrences and enabled strict variable checking, so that it won't slip anymore in the unit-tests. [[`ad06aae`](https://github.com/contao/contao/pull/4671/commits/ad06aaeb3f5a96436bb668920bdf11626963109d)]

ausi (dismissed):

Toflar (dismissed):

fritzmg (dismissed):

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
```diff
@@ -0,0 +1,30 @@
+{% if not searchable|default(true) %}
+<!-- indexer::stop -->
+{%- endif %}
+
+{% block before %}{% endblock %}
```
m-vo:
```suggestion
```

core-bundle/src/Resources/contao/templates/_new/content_element/_base.html.twig
```diff
@@ -0,0 +1,30 @@
...
+                {% if headline.text %}
+                    {% include "@Contao/component/_headline.html.twig" with headline|merge({as_editor_view}) only %}
+                {% endif %}
+            {% endblock %}
+
+            {% block content %}{% endblock %}
+        {% endblock %}
+    </{{ block('wrapper_tag') }}>
+{% endblock %}
+{% block after %}{% endblock %}
```
m-vo:
```suggestion
```

bytehead (dismissed):

bytehead:
❤️

ausi (dismissed):

core-bundle/src/Resources/config/controller.yml
```diff
@@ -55,6 +55,18 @@ services:
         tags:
             - controller.service_arguments
 
+    Contao\CoreBundle\Controller\ContentElement\CodeController:
+        tags:
+            - { name: contao.content_element, category: texts }
```
leofeyer: I think we‘ve talked about using PHP attributes for these (at least for the category), haven‘t we?

leofeyer (dismissed):

core-bundle/src/Resources/config/controller.yml
m-vo: Yes, but IMHO we decided to do this in a separate PR for all controllers (including the already existing).

m-vo: see ce70d9f0f397d1e0f0773dce065215f3dd2c4f2d + be1e53ff6ac505bd1eb18b7312d6f3bfd46ca976

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4672, opened 2022-05-13
Add a deprecation helper to detect insert tags in Twig templates

m-vo:
This PR introduces a Twig node visitor that adds `Twig\Node\DeprecatedNode` nodes when a specific usage is found during compilation.

The first implemented scenario detects this pattern `{{ '{{an_insert_tag'}}' }}` and suggests using the `insert_tag` function instead:
image

The analysis only happens at compile time and does not affect runtime performance.

/cc @ausi

### DIFF
--- a/core-bundle/src/Twig/Extension/DeprecationsNodeVisitor.php
+++ b/core-bundle/src/Twig/Extension/DeprecationsNodeVisitor.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Extension;
+
+use Twig\Environment;
+use Twig\Node\DeprecatedNode;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Node;
+use Twig\Node\PrintNode;
+use Twig\NodeVisitor\AbstractNodeVisitor;
+
+/**
+ * @internal
+ */
+class DeprecationsNodeVisitor extends AbstractNodeVisitor
+{
+    public function getPriority(): int
+    {
+        return 10;
+    }
+
+    protected function doEnterNode(Node $node, Environment $env): Node
+    {
+        return $node;
+    }
+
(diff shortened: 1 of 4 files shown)


ausi:
We could maybe even go as far as checking if there is a node that starts with `{{ '{{…` so every print node that starts with a string literal that starts with `{{`.

m-vo:
> We could maybe even go as far as checking if there is a node that starts with {{ '{{… so every print node that starts with a string literal that starts with {{.

This gets complicated quickly, because then generating suggestions (or even displaying the usage) must be done at runtime. Because there could be any sort of (nested) nodes in the print node, we would need to execute and capture the output (same as with the `add` tag). But I would avoid doing this as it could easily introduce sideeffects.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

core-bundle/src/Twig/Extension/DeprecationsNodeVisitor.php
```diff
@@ -57,19 +57,19 @@ private function handleDeprecatedInsertTagUsage(Node $node): Node
...
     }
 
-    private function addDeprecation(Node $node, string $version, string $message): Node
+    private function addDeprecation(Node $node, string $message): Node
     {
         $line = $node->getTemplateLine();
 
         $deprecatedNode = new DeprecatedNode(
-            new ConstantExpression("Since contao/core-bundle $version: $message", $line),
+            new ConstantExpression("Since contao/core-bundle 4.13: $message", $line),
```
m-vo, line 72: @leofeyer Why did you remove the version variable?

leofeyer: Because it is always `4.13` and thus we don‘t need a separate variable.

m-vo: The idea was that we could have more handlers in the future (e.g. `handleDeprecatedFooBarUsage`) and they could add deprecations similar to `trigger_deprecation`.

leofeyer: Sure. We can still add a variable then.

PR #4675, opened 2022-05-16
Undeprecate the third Input::get() parameter

ausi:
In #4510 I deprecated the third parameter of `Input::get($strKey, $blnDecodeEntities=false, $blnKeepUnused=false)`

@fritzmg raised concerns in https://github.com/contao/contao/pull/4510#issuecomment-1114279367 and the core itself still uses it in https://github.com/contao/contao/blob/67608ff61bfb118971852617c80f135c4e9aef2b/core-bundle/src/Resources/contao/classes/Frontend.php#L96

Therefore I think we should revert this deprecation.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -138,12 +138,13 @@ class Input
 	/**
 	 * Return a $_GET variable
 	 *
-	 * @param string  $strKey            The variable name
-	 * @param boolean $blnDecodeEntities If true, all entities will be decoded
+	 * @param string  $strKey                      The variable name
+	 * @param boolean $blnDecodeEntities           If true, all entities will be decoded
+	 * @param boolean $blnKeepUnusedRouteParameter If true, the route parameter will not be marked as used (see #4277)
 	 *
 	 * @return array|string|null The cleaned variable value
 	 */
-	public static function get($strKey, $blnDecodeEntities=false)
+	public static function get($strKey, $blnDecodeEntities=false, $blnKeepUnusedRouteParameter=false)
 	{
 		$varValue = static::findGet($strKey);
 
@@ -153,11 +154,7 @@ class Input
 		}
 
 		// Mark the parameter as used (see #4277)
-		if (\func_num_args() > 2 && func_get_arg(2))
-		{
-			trigger_deprecation('contao/core-bundle', '5.0', 'Using %s() with the third parameter "$blnKeepUnused" has been deprecated and will no longer work in Contao 6.0.', __METHOD__);
-		}
-		else
+		if (!$blnKeepUnusedRouteParameter)
 		{
 			unset(self::$arrUnusedRouteParameters[$strKey]);
 		}


fritzmg (approved):

leofeyer:
> I use that all the time in order to check whether an `auto_item` is present, without marking the parameter used (so that the 404 page is still generated if not used)

That does not make sense to me. Why would you check for an auto_item parameter if you don‘t intend to use it? Can you provide an example, please?

fritzmg:
> Why would you check for an auto_item parameter if you don‘t intend to use it? Can you provide an example, please?

Currently the only way of getting the currently displayed news record outside the news reader module is to do something like:

```php
$news = NewsModel::findByAlias(Input::get('auto_item'));

if (null !== $news) {
    // …
}
```

Now if that code happens to be executed for _every_ page (because either the module lives in the page layout or it is used in an event listener or hook) then certain URLs will respond with `200` instead of `404`. e.g. an non existent URL `http://example.com/lorem/ipsum` for the existent page `lorem` will return `200` instead of `404` because `ipsum` will be the `auto_item` parameter and it will be marked as used.

So instead you _always_ need to make sure to use `Input::get('auto_item', false true)` when used anywhere _outside_ the newsreader module for defensive programming purposes.

fritzmg:
For a more practical example: this updates the breadcrumb with the currently displayed news item:

```php
// src/EventListener/SetNewsInBreadcrumbListener
namespace App\EventListener;

use Contao\CoreBundle\ServiceAnnotation\Hook;
use Contao\Input;
use Contao\NewsModel;
use Contao\StringUtil;

/**
 * @Hook("generateBreadcrumb")
 */
class SetNewsInBreadcrumbListener
{
    public function __invoke(array $items): array
    {
        $news = NewsModel::findByIdOrAlias(Input::get('auto_item', false, true));

        if (null !== $news) {
            $items[count($items) - 1] = array_merge($items[count($items) - 1], [
                'link' => StringUtil::specialchars($news->headline),
                'title' => StringUtil::specialcharsAttribute($news->headline),
                'class' => $news->cssClass,
                'data' => $news->row(),
            ]);
        }

        return $items;
    }
}
```

Not using `Input::get('auto_item', false, true)` would lead to the aforementioned issue as the breadcrumb module is typically present on all pages.

_Note:_ This is of course not a perfect solutions since there could also be calendar entries with the same alias as a news alias for example. However with what is currently available there are no perfect solutions (that I know of).

leofeyer:
But in your `http://example.com/lorem/ipsum` example, the `auto_item` parameter **is** used. It just happens to be invalid, therefore you should throw a "not found" exception instead of marking the parameter as unused.

The way you are using the `blnKeepUnused` feature is not a supported use case IMHO.

fritzmg:
> But in your http://example.com/lorem/ipsum example, the auto_item parameter is used.

Not it's not, it's an invalid URL.

> It just happens to be invalid, therefore you should throw a "not found" exception instead of marking the parameter as unused.

That won't be possible as my code cannot make any assumptions about whether the `auto_item` is needed or not in another module on the same page. Only the news reader can do that.

> The way you are using the blnKeepUnused feature is not a supported use case IMHO.

How would you implement its usage then? I can tell you from my community and work experience that this is indeed used a lot. If we do not provide a good alternative (e.g. page controllers for detail pages) then we need to keep it.

leofeyer:
> That won't be possible as my code cannot make any assumptions about whether the `auto_item` is needed or not in another module on the same page. Only the news reader can do that.

That‘s right.

leofeyer:
Thank you @ausi.

PR #4677, opened 2022-05-16
Drop the "loadNavigation" Ajax action

aschempp:
This has become obsolete and is not used anymore since we use KnpMenu for navigations


Toflar (dismissed):
Not sure if we should remove this template in a bugfix version. But on the other hand, I don't think this has ever been used by anyone. I learned about it today :D

leofeyer (changes requested):
We can obviously only remove this in Contao 5.

aschempp:
I would call this an internal method that is not used by the core anymore. You think someone manually renders a partial of a back end menu section?

m-vo:
I think I would also only remove this in 5 as long as it doesn't hinder us doing certain things in 4.13. Together with the 100 other things, we already dropped there, that neither anyone was using.

leofeyer:
It does not matter what I think. It is a BC break and therefore it must not be removed in a patch release.

aschempp:
it was apparently removed in Contao 3 and not used anywhere in Contao 4 🤷 but I changed the base branch now

ausi (approved):

leofeyer (approved):

PR #4679, opened 2022-05-16
Fix several accessibility issues in the back end navigation

aschempp:
This fixes the following issues:

1. adds aria roles to the main navigation section toggles
2. adds a button to the profile toggle because a `<span>` element with click event is not available for keyboard navigation
3. adds aria roles so a screenreader/assistive device understands that the left menu is toggled by the burger menu. But only if the screen is on mobile resolution (burger is visible).
4. removes the `blur()` call when toggling navigation sections, otherwise the keyboard focus is lost and it is not possible to navigation to the now-visible sub items

I'm aware this mixes vanilla and mootools, and we'll clean that up for Contao 5. But imho the bugs should be fixed in the LTS version as well.


Toflar (dismissed):
Cannot really comment on the usage of the aria rules but at least they look logical to me ❤️ Maybe @MacKP wants to have a look at this?

MacKP:
This looks fine to me. Love the button instead of a span ;-)
Thx!

ausi (dismissed):
❤️

core-bundle/src/EventListener/Menu/BackendMenuListener.php
```diff
@@ -143,9 +147,11 @@ private function buildHeaderMenu(MenuEvent $event, BackendUser $user): void
 
         $submenu = $factory
             ->createItem('submenu')
-            ->setLabel($this->translator->trans('MSC.user', [], 'contao_default').' '.$user->username)
+            //->setLabel($this->translator->trans('MSC.user', [], 'contao_default').' '.$user->username)
```
leofeyer: This seems like a left-over?

leofeyer (changes requested):
Something is wrong with the CSS in the active state:



It should look like this:

core-bundle/src/EventListener/Menu/BackendMenuListener.php
```diff
@@ -143,9 +147,11 @@ private function buildHeaderMenu(MenuEvent $event, BackendUser $user): void
...
+            ->setLabel('<button type="button">'.$this->translator->trans('MSC.user', [], 'contao_default').' '.$user->username.'</button>')
```
leofeyer, line 150: The button element is still wrapped with a span:



Is this intended?

aschempp: you mean the class or the span? KnpMenu creates a span for all non-links, and it's the same for the hamburger menu toggle.

aschempp:
Regarding the styling issues, this should be fixed in the latest CSS, did you clear the browser cache?

core-bundle/src/EventListener/Menu/BackendMenuListener.php
aschempp: Fixed in 295d534a75015ec02a90e133cd39f3f026936813

leofeyer:
It is not a browser cache issue. See the video below:

https://user-images.githubusercontent.com/1192057/169978312-f84ac2ff-090d-43ad-b15b-00d970467d6d.mov

aschempp:
Thanks for the video, that made it easy to fix in https://github.com/contao/contao/pull/4679/commits/4bc5b1e3f340a56167f9ea208c8d3dc9a66b2519 🙃

leofeyer (approved):

PR #2946, opened 2021-04-09
Fix the page context when generating RSS feeds

fritzmg:
This is a follow up to #2848.

### Reproduction of the problem

1. Create a page article with some content.
2. Create a news archive and a published news article.
3. Put `{{insert_article::*}}` into the teaser of the news article (replace `*` with the ID of the previously generated page article).
4. Create an RSS feed and select the created news archive.

Unter PHP 8.0, the following error will occur upon saving the RSS feed or when regenerating the XML files in the maintenance section:

```
ErrorException:
Warning: Attempt to read property "datimFormat" on null

  at vendor\contao\contao\core-bundle\src\Resources\contao\modules\ModuleArticle.php:124
```

Same basic reproduction with events as well.

Since the XML feeds are always generated without a proper front end context, we also need to set the global `$objPage` variable (not just the front end scope) according to the `jumpTo` page of the news archive (or calendar respectively).

### Notes

* This PR includes a behaviour change: if the configured `jumpTo` page of a news archive or calendar is deleted, the affected news and event entries will be skipped now. In my opinion this is fine, since the news archive or calendar is then in a misconfigured state. See also the original issue of the affected lines of code: https://github.com/contao/core/issues/5781
* This PR is based against **4.11**. While the original issue does not cause an error in Contao **4.9** I do consider it a bug that the global `$objPage` object is not set properly for the feed generation. Should I rebase it against **4.9** instead?

PR #4682, opened 2022-05-17
News feeds as page controller

bezin:
This implements news feeds as page controller and superseds #4459 eventually. Also I decided to split these into a news-bundle PR and an Event-bundle PR, because I find it easier to handle. 

Still a couple things to do: 
* [x] Add feeds to layout (DCA is updated, but the correct routes need to be added to the page. Requires `ResponseContext`)
* [x] Check and improve extensibility (e.g. https://github.com/contao/core-bundle/issues/1243, overwrite image size of every image element)
* [x] Get rid of some legacy code. if possible
* [x] Update permisssion handling (see below)
* [x] Page Icon: fix offset in page tree and add required variants (not visible, not editable, ...)
* [x] Update News Feed Insert tag
* [x] Remove obsolete code
* [x] Write migration (migrate feeds to pages; update permissions, if necessary)
* [x] Add tests

There are also some permissions for news feeds, i.e. which news feed a editor has access to and if he/she is allowed to create or delete them). Since this feed is now a page controller, access in general  is handled by the page type limitation. The question is, if we want to keep this distinction? I for myself are fine with the general limitation (can access news feed page types yes/no). 

# New Features
* JSON Feeds
* Only list featured/unfeatured news articles in the feed (equivalent to the newslist module)


news-bundle/src/ContaoManager/Plugin.php
```diff
@@ -12,23 +12,28 @@
...
         return [
             BundleConfig::create(ContaoNewsBundle::class)
-                ->setLoadAfter([ContaoCoreBundle::class])
                 ->setReplace(['news']),
         ];
     }
+
+    public function registerContainerConfiguration(LoaderInterface $loader, array $managerConfig): void
+    {
+        $loader->load('@DebrilRssAtomBundle/Resources/config/services.yml');
```
aschempp: this looks like a leftover, I thought we're not using the bundle anymore?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+
+    public function __construct(private ContaoFramework $framework, private FeedIo $feedIo, private ResponseTagger $responseTagger, private ContaoContext $contaoContext, private ImageFactory $imageFactory, private RequestStack $requestStack, private InsertTagParser $insertTags, private string $projectDir)
+    {
+    }
+
+    public function __invoke(Request $request, PageModel $pageModel): Response
+    {
+        $feed = new Feed();
+        $feed->setTitle($pageModel->title);
+        $feed->setDescription($pageModel->feedDescription);
```
aschempp: any particular reason we cannot use the existing (meta) description field for that?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        $feed->setLanguage($pageModel->language);
+
+        $time = time();
+        $lastModified = null;
+
+        $archives = StringUtil::deserialize($pageModel->newsArchives);
```
aschempp: Might want to add a check that this results in a non-empty array and return a 500 or 204 otherwise?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        } else {
+            $articles = $newsModel->findPublishedByPids($archives);
+        }
+
+        // TODO: Is this still necessary?
+        $origObjPage = $GLOBALS['objPage'] ?? null;
+
+        foreach ($articles as $article) {
+            // Never add unpublished elements to the RSS feeds
+            if (!$article->published || ($article->start && $article->start > $time) || ($article->stop && $article->stop <= $time)) {
```
aschempp: I think this check was added for when the XML file was written. Now that we have a real controller, we can actually output unpublished items if front end preview is enabled, since the feed will only be for this user 😎

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+            // TODO: Throw an event here to allow for custom model-to-item transformation?
+            $item = $this->transformModelToItem($article, $pageModel->feedSource, $pageModel);
+
+            if (null === $lastModified) {
+                $lastModified = $item->getLastModified();
+            }
+
+            if ($lastModified < $item->getLastModified()) {
+                $lastModified = $item->getLastModified();
+            }
```
aschempp:
```suggestion
            $lastModified = null === $lastModified ? $item->getLastModified() : min($lastModified, $item->getLastModified());
```

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        $GLOBALS['objPage'] = $origObjPage;
+
+        $feed->setLastModified($lastModified);
+
+        $response = new Response($this->feedIo->format($feed, $pageModel->feedFormat));
+        $response->setLastModified($feed->getLastModified());
+        $response->headers->set('Content-Type', $this->contentTypes[$pageModel->feedFormat]);
+
+        foreach ($archives as $archive) {
+            $this->responseTagger->addTags(['contao.db.news_archive.'.$archive]);
```
aschempp:
```suggestion
            $this->responseTagger->addTags(['contao.db.tl_news_archive.'.$archive]);
```

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+    ];
+
+    // TODO: Fix content types, maybe make it configurable in the page settings
+    private $contentTypes = [
+        'atom' => 'application/atom+xml',
+        'json' => 'application/feed+json',
+        'rss' => 'application/rss+xml',
+    ];
+
+    public function __construct(private ContaoFramework $framework, private FeedIo $feedIo, private ResponseTagger $responseTagger, private ContaoContext $contaoContext, private ImageFactory $imageFactory, private RequestStack $requestStack, private InsertTagParser $insertTags, private string $projectDir)
```
aschempp: `$responseTagger` service must be optional since it only exists with a reverse proxy. Also need to add respective checks when calling the `->addTags()` method

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+            ->setLastModified((new \DateTime())->setTimestamp($model->date))
+            ->setLink($link)
+        ;
+
+        $item->setContent($this->getItemDescription($source, $model, $item));
+
+        /** @var UserModel $authorModel */
+        if (($authorModel = $model->getRelated('author')) instanceof UserModel) {
+            $item->setAuthor((new Author())->setName($authorModel->name));
+        }
```
aschempp: has this already been there? Not sure I like having a backend user name on news items in general, but dunno if a configuration makes sense 🤔

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        $this->responseTagger->addTags(['contao.db.tl_news.'.$model->id]);
+
+        return $item;
+    }
+
+    private function getItemDescription(string $feedSource, NewsModel $model, Item $item): string
+    {
+        // TODO: Is this still necessary?
+        // Push a new request to the request stack (#3856)
+        $request = Request::create($item->getLink());
```
aschempp: this is no longer necessary for sure

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        $format = $route->getPageModel()->feedFormat;
+        $route->setUrlSuffix($this->urlSuffixes[$format]);
+    }
+
+    public function getUrlSuffixes(): array
+    {
+        return array_unique(array_values($this->urlSuffixes));
+    }
+
+    private function transformModelToItem(NewsModel $model, string $source, PageModel $pageModel): Item
```
aschempp: maybe call it `$feedSource` instead of `$source` to be consistent with page model and other methods

news-bundle/src/EventListener/DataContainer/LayoutListener.php
```diff
@@ -0,0 +1,48 @@
...
+        $model = $this->framework->getAdapter(PageModel::class);
+
+        $feeds = $model->findByType(NewsFeedController::TYPE);
+
+        $options = [];
+
+        if (null !== $feeds) {
+            while ($feeds->next()) {
+                $label = $this->translator->trans('FEED_FORMAT.'.$feeds->feedFormat, [], 'contao_tl_page');
+                $options[$feeds->id] = $feeds->title.sprintf(' (%s)', $label);
```
aschempp:
```suggestion
                $options[$feeds->id] = sprintf('%s (%s)', $feeds->title, $label);
```

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @Callback(table="tl_page", target="fields.archives.options")
+     */
+    public function getAllowedArchives(): array
+    {
+        $model = $this->framework->getAdapter(NewsArchiveModel::class);
+        $user = $this->security->getUser();
+
+        $archive = $user->isAdmin ? $model->findAll() : $model->findMultipleByIds($user->news);
```
aschempp:
```suggestion
        if (!$user instanceof BackendUser || $this->security->isGranted('ROLE_ADMIN')) {
            $archive = $model->findAll();
        } else {
            $archive = $model->findMultipleByIds($user->news);
        }
```
(might need `use Contao\BackendUser`)

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,60 @@
...
+
+        $options = [];
+
+        if (null !== $archive) {
```
aschempp: I think this whole method would be more performant by using real DB queries to only get title and ID instead of retrieving the full model in the back end.

news-bundle/src/Resources/contao/dca/tl_layout.php
```diff
@@ -22,7 +22,6 @@
 (
 	'exclude'         => true,
 	'inputType'       => 'checkbox',
-	'foreignKey'      => 'tl_news_feed.title',
```
aschempp: could we use `tl_page.title` for this?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+
+        $response = new Response($this->feedIo->format($feed, $pageModel->feedFormat));
+        $response->setLastModified($feed->getLastModified());
+        $response->headers->set('Content-Type', $this->contentTypes[$pageModel->feedFormat]);
+
+        foreach ($archives as $archive) {
+            $this->responseTagger->addTags(['contao.db.news_archive.'.$archive]);
+        }
+
+        return $response;
```
aschempp: should this also be tagged with `tl_page.ID`? or is that done automatically for pages @Toflar ?

news-bundle/src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -0,0 +1,55 @@
...
+      <trans-unit id="tl_page.maxFeedItems.1">
+        <source>Here you can limit the number of feed items. Set to 0 to export all.</source>
+      </trans-unit>
+      <trans-unit id="tl_page.feedDescription.0">
+        <source>Feed description</source>
+      </trans-unit>
+      <trans-unit id="tl_page.feedDescription.1">
+        <source>Here you can enter a short description of the news feed.</source>
+      </trans-unit>
+      <trans-unit id="FEED_FORMAT.rss">
```
aschempp: not sure this should to in `tl_page`. Also not sure it needs translation at all 😬

news-bundle/src/Resources/contao/dca/tl_news_feed.php
```diff
@@ -11,7 +11,6 @@
 use Contao\Automator;
 use Contao\Backend;
 use Contao\BackendUser;
-use Contao\CoreBundle\EventListener\Widget\HttpUrlListener;
```
aschempp: shouldn't this whole file be dropped?

aschempp (changes requested):
Looks very promising! Added feedback and minor nitpicks, but overall its 🔥

news-bundle/src/Controller/NewsFeedController.php
Toflar: No tagging is always the task of the respective controller. So it's done automatically for a page regular but not here :) But yes it should be tagged in order to remove the cache entry when you adjust the page alias etc. Should probably also add the jumpTo page ID of the news archive because if that changes, the URLs to the entries might change.

HOWEVER 😊 You are currently not caching anything. You only set the last modified header so the reverse proxy must revalidate and thus the tags are irrelevant ;) But we have cache settings in the page settings, why not applying them here too?

news-bundle/src/Resources/contao/dca/tl_news_feed.php
bezin: Yes, will do this as last step (its one of the tasks) :-)

news-bundle/src/Resources/contao/languages/en/tl_page.xlf
bezin: Had the same two thoughts. I agree a translation is not required, but I would prefer decent labels ('RSS 2.0' rather than 'rss', JSON rather than 'json' and 'Atom' rather than 'atom') in several places. So the question is, where to put them then.

news-bundle/src/Controller/NewsFeedController.php
bezin: No, it has not, but to me this belongs here to be complete. It is also [already done for the schema.org data](https://github.com/contao/contao/blob/5.x/news-bundle/src/Resources/contao/classes/News.php#L392-L399) and it is not the user name but the regular name.

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,257 @@
...
+        $newsModel = $this->framework->getAdapter(NewsModel::class);
+
+        if ($pageModel->maxFeedItems > 0) {
+            $articles = $newsModel->findPublishedByPids($archives, null, $pageModel->maxFeedItems);
+        } else {
+            $articles = $newsModel->findPublishedByPids($archives);
+        }
+
+        // TODO: Is this still necessary?
+        $origObjPage = $GLOBALS['objPage'] ?? null;
```
fritzmg: No, it is not necessary :). Now that this is a proper page controller, an appropriate `$GLOBALS['objPage']` within the correct website root will be available.

Original issue/PR: https://github.com/contao/contao/pull/2946

bezin: Fixed in 2cdb25b

news-bundle/src/Controller/NewsFeedController.php
aschempp: true. But for schema.org data (and news templates) I can change it through the template. 🤔

news-bundle/src/Resources/contao/languages/en/tl_page.xlf
aschempp: maybe we'll just hardcode them?

aschempp:
I also noticed FeedIO currently requires Guzzle (see https://github.com/alexdebril/feed-io/issues/398) 🙈

news-bundle/src/Controller/NewsFeedController.php
bezin: Nope, not really, have updated accordingly

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * Set the cache headers according to the page settings.
+     *
+     * @param Response $response The response object
+     *
+     * @return Response The response object
+     */
+    private function setCacheHeaders(Response $response, PageModel $pageModel)
```
Toflar: Wooooh, this looks very much like something we could use in other Contao controllers too! What do you think about moving that to `AbstractController`? :)

m-vo: Or even the `EntityCacheTags` helper?

bezin: I moved this method to `AbstractController` and extended from there

bezin:
> There are also some permissions for news feeds, i.e. which news feed a editor has access to and if he/she is allowed to create or delete them). Since this feed is now a page controller, access in general is handled by the page type limitation. The question is, if we want to keep this distinction?

What is your take on this? IMO we should drop these granular permissions and handle feeds like every other page, too.

aschempp:
> > There are also some permissions for news feeds, i.e. which news feed a editor has access to and if he/she is allowed to create or delete them). Since this feed is now a page controller, access in general is handled by the page type limitation. The question is, if we want to keep this distinction?
> 
> What is your take on this? IMO we should drop these granular permissions and handle feeds like every other page, too.

I think so too. Did we consider how this can be migrated on an update? Because I guess we cannot migrate the permissions?

bezin:
> I think so too. Did we consider how this can be migrated on an update? Because I guess we cannot migrate the permissions?

As there is no direct equivalent to each option after this gets merged I think we cannot migrate these. The baseline would be, that if an editor has the permission to at least one news feed, we should allow access to the new News Feed page type in general. However, this would mean the editor may now have permission to more feeds than previously.

bezin:
As you see I have added a `TransformFeedItem` to make the transformation from a news model to feed item exensible. However, I am not exactly sure about this and have some more cases, where I fail to see how to implement this the right way:

* I use feeds (custom implementation) with CleverReach to import Contao News into their editor. They do not wait long enough when requesting images for it work with our deferred image resizing. Hence I need a way to tell Contao to resize images right away in this case.

* For the same use case, I want to be able to overwrite image sizes in Image content elements to resize them to a size that is suitable for mailings. 

From https://github.com/contao/core-bundle/issues/1243:
* Add a namespace to the Feed object, e.g. for Facebook Instant Articles. The library we use supports this, but it must be extensible by a third party. 

Any comment much appreciated. Maybe the image resizing must be solved completely different anyways.

bezin:
I experimented with a transformer service rather than throwing an event with the idea that developers may decorate this service to overwrite some behaviour. I pushed this to a separate branch: https://github.com/bezin/contao/commit/5550eb37409f68ea7e4e3e796ee910bf76fa2c4d

bezin:
I experimented with a provider service rather than an event (see https://github.com/bezin/contao/commit/19aad3be37fa3adfdfe4118b70c62ebe482479b1)

I implemented a FeedProvider that a) creates the feed instance, b) fetches the news articles depending on page settings and c) transforms articles to items. This way, you may decorate the provider and alter the feed instance and/or every item. 

Add additional namespaces to the feed instance:

```php
// src/App/NewsFeedProvider

class NewsFeedProvider implements FeedProviderInterface
{
  // ...
  public function getFeed(): FeedInterface
  {
    $feed = $this->inner->getFeed();
    $feed->addNS('content', 'http://purl.org/rss/1.0/modules/content/');
    return $feed;
  }
  // ...
}
```

Add a custom format options to `tl_page.feedSource` and alter item transformation all together in your decorator:

```php
// src/App/NewsFeedProvider

class NewsFeedProvider implements FeedProviderInterface
{
  // ...
  public function getItemFromModel(NewsModel $model): ItemInterface
  {
    if ('my_custom_source_option' !== $this->pageModel->feedSource) {
      return $this->inner->getItemFromModel($model);
    }
    
    // Handle item creation in your own way
  }
  // ...
}
```

What are your thoughts on this?

fritzmg:
Apparently you are not supposed to extend from the original service when decorating 😬 🤷 

I'd prefer an event still I think, since that is what we use for `/sitemap.xml` and `/robots.txt` as well.

m-vo:
> Apparently you are not supposed to extend from the original service when decorating

Yes, that's a bad idea in general, because you cannot guarantee that the original class will be in the correct state if you only pass on some calls. Implementing the interface, injecting the original service and using its public API is usually safer to do. :slightly_smiling_face: 

But tbh this is also true for every extended base class - so it primarily depends on if a class is designed for this use case or not (for non abstract classes: mostly not).

bezin:
> Apparently you are not supposed to extend from the original service when decorating 😬 🤷

Yes, you are right. This is actually wrong in the example – it is not required.

bezin:
I now implemented two events:
*  `FetchArticlesForFeed` (similiar to the `FetchNewsListItemHook`) to overwrite the logic to fetch the articles (filter by category, filter by author, filter by what not).
* `TransformArticleForFeed` to transform a news article to a feed item

The Contao\NewsBundle\EventListener\NewsFeedListener listens to both events. 

I still need to add tests and migrations.

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -111,6 +111,10 @@
  * @property string|boolean $hasMooTools
  * @property string         $template
  * @property string         $templateGroup
+ * @property string         $newsArchives
+ * @property string         $feedFormat
+ * @property string         $feedSource
+ * @property integer        $maxFeedItems
```
aschempp: not sure we're adding properties from extensions (`news-bundle`) in files from the core bundle @contao/developers ?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\NewsBundle\Controller;
```
aschempp: I think we moved page controllers to a subfolder

```suggestion
namespace Contao\NewsBundle\Controller\Page;
```

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+        $baseUrl = $staticUrl ?: $request->getSchemeAndHttpHost();
+
+        $feed = (new Feed())
+            ->setTitle($pageModel->title)
+            ->setDescription($pageModel->description)
+            ->setLanguage($pageModel->language)
+        ;
+
+        $event = new FetchArticlesForFeedEvent($feed, $request, $pageModel);
+        $this->container->get('event_dispatcher')->dispatch($event, FetchArticlesForFeedEvent::NAME);
```
aschempp: If the event has only one purpose, I don't think it needs a name at all, since Symfony can register it on the class name (applies to the `TransformArticleForFeedEvent` as well)

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        $formatter = $this->specification->getStandard($pageModel->feedFormat)->getFormatter();
+        $response = new Response($formatter->toString($feed));
+        $response->headers->set('Content-Type', self::$contentTypes[$pageModel->feedFormat]);
+
+        $this->setCacheHeaders($response, $pageModel);
+
+        foreach (StringUtil::deserialize($pageModel->newsArchives, true) as $archive) {
+            $this->tagResponse('contao.db.tl_news_archive.'.$archive);
```
aschempp: it feels a bit strange that this is done by the page controller, but actually fetching the feed items is not. The page controller can't really be sure if these archives are used. Shouldn't this rather be based on the articles that are fetched from the event (their PID)?

news-bundle/src/Controller/NewsFeedController.php
aschempp: @bezin seems like this still applies?

news-bundle/src/Controller/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return $response;
+    }
+
+    public function configurePageRoute(PageRoute $route): void
+    {
+        $format = $route->getPageModel()->feedFormat;
+        $route->setUrlSuffix($this->urlSuffixes[$format]);
```
aschempp: maybe handle if the `feedFormat` value in the DB is not found in the array?

news-bundle/src/Event/FetchArticlesForFeedEvent.php
```diff
@@ -0,0 +1,54 @@
...
+use Contao\Model\Collection;
+use Contao\PageModel;
+use FeedIo\FeedInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class FetchArticlesForFeedEvent
+{
+    public const NAME = 'contao_news.feed.fetch_articles';
+
+    private ?Collection $articles = null;
```
aschempp:
```suggestion
    private Collection|null $articles = null;
```

news-bundle/src/Event/FetchArticlesForFeedEvent.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return $this->request;
+    }
+
+    public function getPageModel(): PageModel
+    {
+        return $this->pageModel;
+    }
+
+    public function getArticles(): ?Collection
```
aschempp: I think we should use an array of models here instead of a `Collection`, that would be more flexible to add own articles. And maybe a `addArticle` method as well?

news-bundle/src/Event/TransformArticleForFeedEvent.php
```diff
@@ -0,0 +1,66 @@
...
+use FeedIo\Feed\ItemInterface;
+use FeedIo\FeedInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Contracts\EventDispatcher\Event;
+
+class TransformArticleForFeedEvent extends Event
+{
+    public const NAME = 'contao_news.feed.transform_article';
+
+    private ?ItemInterface $item = null;
```
aschempp:
```suggestion
    private ItemInterface|null $item = null;
```

news-bundle/src/Event/TransformArticleForFeedEvent.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return $this->request;
+    }
+
+    public function getBaseUrl(): string
+    {
+        return $this->baseUrl;
+    }
+
+    public function getItem(): ?ItemInterface
```
aschempp:
```suggestion
    public function getItem(): ItemInterface|null
```

news-bundle/src/EventListener/DataContainer/LayoutListener.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Contao\NewsBundle\EventListener\DataContainer;
+
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Contao\NewsBundle\Controller\NewsFeedController;
+use Contao\PageModel;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class LayoutListener
```
aschempp: I think this should be a `LayoutNewsfeedOptionsListener` or something like that 🙃

news-bundle/src/EventListener/DataContainer/LayoutListener.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    public function __construct(private ContaoFramework $framework, private TranslatorInterface $translator)
+    {
+    }
+
+    /**
+     * @Callback(table="tl_layout", target="fields.newsfeeds.options")
+     */
+    public function getAllFeeds(): array
```
aschempp:
```suggestion
    public function __invoke(): array
```

And the annotation should be moved to the class instead of the method

news-bundle/src/EventListener/DataContainer/LayoutListener.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Contao\NewsBundle\Controller\NewsFeedController;
+use Contao\PageModel;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class LayoutListener
+{
+    public function __construct(private ContaoFramework $framework, private TranslatorInterface $translator)
```
aschempp: `$translator` seems unused?

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $user = $this->security->getUser();
+
+        $stmt = $this->connection->createQueryBuilder()
+            ->select('id, title')
+            ->from('tl_news_archive')
+        ;
+
+        if (!($user instanceof BackendUser || $this->security->isGranted('ROLE_ADMIN'))) {
+            $stmt->where('id', $user->news);
```
aschempp: can we `->where()` on an array of IDs?

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        $results = $stmt->executeQuery();
+
+        $options = [];
+
+        if ($results->rowCount() > 0) {
```
aschempp: probably an unnecessary check, the foreach loop would just be empty?

news-bundle/src/EventListener/GeneratePageListener.php
```diff
@@ -43,21 +41,21 @@ public function __invoke(PageModel $pageModel, LayoutModel $layoutModel): void
 
         $this->framework->initialize();
 
-        $adapter = $this->framework->getAdapter(NewsFeedModel::class);
+        $adapter = $this->framework->getAdapter(PageModel::class);
 
-        if (!($feeds = $adapter->findByIds($newsfeeds)) instanceof Collection) {
+        if (!($feeds = $adapter->findMultipleByIds($newsfeeds)) instanceof Collection) {
```
aschempp: should we make sure these are feed pages?

news-bundle/src/EventListener/InsertTagsListener.php
```diff
@@ -55,13 +56,17 @@ private function replaceNewsFeedInsertTag(string $feedId): string
     {
         $this->framework->initialize();
 
-        $adapter = $this->framework->getAdapter(NewsFeedModel::class);
```
aschempp, line 58: I'm not sure we should actually keep the `news_feed` insert tag. It will not correctly work after an update, and it's the same as `link_url` anyway. Removing it (and maybe writing something to the logs) would probably be easier to spot errors.

news-bundle/src/EventListener/NewsFeedListener.php
```diff
@@ -0,0 +1,210 @@
...
+    private string $baseUrl;
+
+    public function __construct(private readonly ContaoFramework $framework, private readonly ImageFactoryInterface $imageFactory, private readonly InsertTagParser $insertTags, private readonly string $projectDir, private readonly EntityCacheTags $cacheTags)
+    {
+    }
+
+    public function onFetchArticlesForFeed(FetchArticlesForFeedEvent $event): void
+    {
+        $pageModel = $event->getPageModel();
+        $archives = StringUtil::deserialize($pageModel->newsArchives);
```
aschempp:
```suggestion
        $archives = StringUtil::deserialize($pageModel->newsArchives, true);
```

news-bundle/src/EventListener/NewsFeedListener.php
```diff
@@ -0,0 +1,210 @@
...
+        $featured = match ($pageModel->feedFeatured) {
+            'featured' => true,
+            'unfeatured' => false,
+            default => null,
+        };
+
+        $newsModel = $this->framework->getAdapter(NewsModel::class);
+
+        $articles = $newsModel->findPublishedByPids($archives, $featured, $pageModel->maxFeedItems, 0, [
+            'return' => 'Collection',
```
aschempp:
```suggestion
            'return' => 'Array',
```
as suggested above

news-bundle/src/EventListener/NewsFeedListener.php
```diff
@@ -0,0 +1,210 @@
...
+        $event->setArticles($articles);
+    }
+
+    public function onTransformArticleForFeed(TransformArticleForFeedEvent $event): void
+    {
+        $article = $event->getArticle();
+
+        $this->pageModel = $event->getPageModel();
+        $this->request = $event->getRequest();
+        $this->baseUrl = $event->getBaseUrl();
```
aschempp: this will introduce state to the listener service. I would pass on the event to the `getContent` instead

news-bundle/src/EventListener/NewsFeedListener.php
```diff
@@ -0,0 +1,210 @@
...
+        $enclosures = $this->getEnclosures($article, $item);
+
+        foreach ($enclosures as $enclosure) {
+            $item->addMedia($enclosure);
+        }
+
+        $event->setItem($item);
+    }
+
+    protected function getTitle(NewsModel $article, ItemInterface $item): string
```
aschempp: any reason these methods are all `protected` instead of `private`?

news-bundle/src/Feed/FeedProvider.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    private ?PageModel $pageModel = null;
+    private ?Request $request = null;
+    private ?string $baseUrl = null;
+
+    public function __construct(private readonly ContaoFramework $framework)
+    {
+    }
+
+    public function setContext(PageModel $pageModel, Request $request, string $baseUrl): void
```
aschempp: Without checking all the implications, doesn't the request already include the pageModel and baseUrl? Maybe passing the request to `getFeed` etc. would simplify the state problem?

news-bundle/src/Feed/FeedProvider.php
```diff
@@ -0,0 +1,81 @@
...
+
+        if ($this->pageModel->maxFeedItems > 0) {
+            return $newsModel->findPublishedByPids($archives, $featured, $this->pageModel->maxFeedItems, 0, [
+                'return' => 'Collection',
+            ]);
+        }
+
+        return $newsModel->findPublishedByPids($archives, $featured, 0, 0, [
+            'return' => 'Collection',
+        ]);
```
aschempp: Not sure what problem this is solving, but it could probably be simplified to this?

```suggestion
        return $newsModel->findPublishedByPids($archives, $featured, max((int) $this->pageModel->maxFeedItems, 0), 0, [
            'return' => 'Collection',
        ]);
```

news-bundle/src/Feed/FeedProviderInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+namespace Contao\NewsBundle\Feed;
+
+use Contao\Model\Collection;
+use Contao\PageModel;
+use FeedIo\FeedInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+interface FeedProviderInterface
```
aschempp: maybe I'm missing something, but can you explain what the `FeedProvider(Interface)` is for since we also have events?

news-bundle/src/Resources/contao/dca/tl_layout.php
```diff
@@ -22,7 +22,7 @@
...
```
aschempp: thinking about this, this should probably be a page picker instead? A checkbox list of pages does not make much sense 😆

news-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -38,6 +38,12 @@
       <trans-unit id="MSC.newsOverview">
         <source>Back to the news overview</source>
       </trans-unit>
+      <trans-unit id="PTY.news_feed.0">
+        <source>News Feed</source>
+      </trans-unit>
+      <trans-unit id="PTY.news_feed.1">
+        <source>Create a RSS/Atom/JSON Feed from your news archives</source>
```
aschempp:
```suggestion
        <source>Create a RSS/Atom/JSON feed from your news archives</source>
```

news-bundle/src/Resources/contao/languages/en/modules.xlf
```diff
@@ -8,9 +8,6 @@
       <trans-unit id="MOD.news.1">
         <source>Manage news and export them as RSS or Atom feed</source>
```
aschempp: Does the RSS/Atom hint still belong here?

aschempp (changes requested):
Love it ❤️
I've added a lot of minor nitpicks, but the overall concept totally makes sense to me and is a perfect example how page controllers can/should be used in the future 😎 

Regarding the icons, can a feed page really be visible or hidden from navigation? And what was changed on the `regular_2.svg` file?

core-bundle/src/Resources/contao/models/PageModel.php
bezin: Oh, yes you are probably right. Did not think about this. Can remove it, if required?

news-bundle/src/Controller/NewsFeedController.php
bezin: I think this was solved here: https://github.com/contao/contao/pull/4682/files#diff-e860c0628b064abb3cac479c7bafd78995fc7a8be5ef97dfc53f06580739fa69R84

news-bundle/src/Feed/FeedProvider.php
bezin: This code is obsolete since we only use events now.

news-bundle/src/Feed/FeedProviderInterface.php
bezin: You are right, this code is obsolete, since we only use events now.

news-bundle/src/Resources/contao/languages/en/modules.xlf
bezin: Fixed in [8d472c1](https://github.com/contao/contao/pull/4682/commits/8d472c1fe5c35a578df5913375efd8fe2acb36e2)

news-bundle/src/Resources/contao/dca/tl_layout.php
bezin: It is limited to only feed type pages in the corresponding listener, so it does not display too many pages.

news-bundle/src/Controller/NewsFeedController.php
bezin: Yes, you are right. That felt strange all along. I moved it up to the `foreach ($articles)` loop, where I add the tags for the article itself in 360608dca

news-bundle/src/Controller/NewsFeedController.php
bezin: Fixed in d83623cbd

news-bundle/src/Event/FetchArticlesForFeedEvent.php
bezin: Fixed  in 9455e015c

news-bundle/src/Event/FetchArticlesForFeedEvent.php
bezin: Fixed  in 9455e015c

news-bundle/src/Event/TransformArticleForFeedEvent.php
bezin: Fixed  in 728791c05

news-bundle/src/Event/TransformArticleForFeedEvent.php
bezin: Fixed in  728791c05

news-bundle/src/EventListener/DataContainer/LayoutListener.php
bezin: Fixed in  5727555af

news-bundle/src/EventListener/DataContainer/LayoutListener.php
bezin: Fixed in c65af05e4

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -75,6 +75,7 @@ public function __invoke(Request $request, PageModel $pageModel): Response
             $this->container->get('event_dispatcher')->dispatch($event, TransformArticleForFeedEvent::NAME);
             $feed->add($event->getItem());
             $this->tagResponse($article);
+            $this->tagResponse('contao.db.tl_news_archive.'.$article->pid);
```
aschempp: Not sure what happens if we tag the response with the same tag multiple times? @Toflar

bezin: [They pass through `array_unique` every time you add tags. ](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ResponseTagger.php#L129)

news-bundle/src/EventListener/DataContainer/LayoutListener.php
bezin: Fixed

news-bundle/src/EventListener/DataContainer/PageListener.php
bezin: Thanks for this comment – you are right of course. The implementation was actually completely bonkers. I fixed it and added basic unit tests in  718eab070

news-bundle/src/Feed/FeedProvider.php
bezin: The Feed Provider is obsolete and will be deleted.

news-bundle/src/EventListener/GeneratePageListener.php
bezin: Fixed in 97f8d3dff

news-bundle/src/EventListener/NewsFeedListener.php
bezin: Fixed in 395366e96

news-bundle/src/EventListener/NewsFeedListener.php
bezin: Fixed in 395366e96

news-bundle/src/EventListener/NewsFeedListener.php
bezin: Fixed in 395366e96

news-bundle/src/EventListener/NewsFeedListener.php
bezin: I experimented a bit with how to override the transformer in a listenr in your app. I ended up with extending the news bundle listener and overriding the individual `getTitle|Content|...` methods. This way one can override each field individually without having to overwrite everything else.

news-bundle/src/EventListener/InsertTagsListener.php
bezin: Yes, good point with `link_url` I'll trigger a `warning` in the logs.

bezin: See 04a5351ae

core-bundle/src/Resources/contao/models/PageModel.php
bezin: We don't add anything to the other models either. Removed these in 0bd0c6361

bezin:
I cleaned up `regular_2.svg` but I reverted the change as it is unrelated. 

> Regarding the icons, can a feed page really be visible or hidden from navigation?

Besides the icons and more about the general functionality: I think the editor should be able to choose, if the page should be shown in the navigation. I mean after all, why should you not add a link to the feed in your main navigation?

I would also alyways exclude it from being indexed by the search engine. WDYT?

news-bundle/src/Controller/NewsFeedController.php
bezin: Fixed in fe0df19e2

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+        $baseUrl = $staticUrl ?: $request->getSchemeAndHttpHost();
+
+        $feed = (new Feed())
+            ->setTitle($pageModel->title)
+            ->setDescription($pageModel->description)
+            ->setLanguage($pageModel->language)
+        ;
+
+        $event = new FetchArticlesForFeedEvent($feed, $request, $pageModel);
+        $this->container->get('event_dispatcher')->dispatch($event);
```
aschempp: we must also subscribe to the `event_dispatcher`

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if (null === ($articles = $event->getArticles())) {
+            return new Response('', Response::HTTP_NO_CONTENT);
```
aschempp: is it correct to render an empty response? Or should we render an empty feed (if feedio supports that)? Also, this should be a 200 not 204 status code.

> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204
>
> The HTTP 204 No Content success status response code indicates that a request has succeeded, but that the client doesn't need to navigate away from its current page.
> 
> This might be used, for example, when implementing "save and continue editing" functionality for a wiki site. > In this case a [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) request would be used to save the page, and the 204 No Content response would be sent to indicate that the editor should not be replaced by some other page.The HTTP 204 No Content success status response code indicates that a request has succeeded, but that the client doesn't need to navigate away from its current page.

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,81 @@
...
+    public function __construct(private Connection $connection, private Security $security)
+    {
+    }
+
+    /**
+     * @Callback(table="tl_page", target="config.onload")
+     */
+    public function onLoad(DataContainer $dc): void
+    {
+        $type = $this->connection->executeQuery('SELECT type FROM tl_page WHERE id = :id', ['id' => $dc->id])->fetchFirstColumn();
```
aschempp:
```suggestion
        $type = $dc->getCurrentRecord()['type'];
```

news-bundle/src/Migration/FeedMigration.php
```diff
@@ -0,0 +1,107 @@
...
+
+    public function shouldRun(): bool
+    {
+        if (!$this->connection->createSchemaManager()->tablesExist(['tl_news_feed'])) {
+            return false;
+        }
+
+        $result = $this->connection->executeQuery('SELECT COUNT(id) AS count FROM tl_news_feed')->fetchFirstColumn();
+
+        return !empty($result) && $result[0] > 0;
```
aschempp:
```suggestion
        return $this->connection->fetchOne('SELECT COUNT(id)  FROM tl_news_feed') > 0;
```

news-bundle/src/Migration/FeedMigration.php
```diff
@@ -0,0 +1,107 @@
...
+            'imgSize' => "varchar(255) NOT NULL default ''",
+        ];
+
+        foreach ($newFields as $field => $definition) {
+            $this->connection->executeStatement("
+                ALTER TABLE
+                    tl_page
+                ADD
+                    $field $definition
+            ");
```
aschempp: Either the `shouldRun` needs to make sure the migration is only executed if these fields don't exist, or better we should probably check for the fields here and skip existing ones.

news-bundle/src/Migration/FeedMigration.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $page = $this->connection
+            ->executeQuery("SELECT id FROM tl_page WHERE type = 'root' AND dns = :dns AND language = :language LIMIT 1", ['dns' => $feedBase, 'language' => $feed['language']])
+            ->fetchFirstColumn()
+        ;
+
+        // Find first root page, if none matches by dns and language
+        if (!$page) {
+            $page = $this->connection
+                ->executeQuery("SELECT id FROM tl_page WHERE type = 'root' ORDER BY sorting ASC LIMIT 1")
```
aschempp:
```suggestion
                ->executeQuery("SELECT id FROM tl_page WHERE type = 'root' AND fallback=1 ORDER BY sorting ASC LIMIT 1")
```

news-bundle/src/Migration/FeedMigration.php
```diff
@@ -0,0 +1,107 @@
...
+                ->fetchFirstColumn()
+            ;
+        }
+
+        return $page[0] ?? null;
```
aschempp: could also use `->connection->fetchOne()`

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: The controller extends from `\Contao\CoreBundle\Controller\AbstractController` and that already subscribes to `event_dispatcher`

news-bundle/src/EventListener/DataContainer/PageListener.php
bezin: Implemented in 9d1929cd2

news-bundle/src/Migration/FeedMigration.php
bezin: Fixed in a62be998d

news-bundle/src/Migration/FeedMigration.php
bezin: Fixed in a62be998d

news-bundle/src/Migration/FeedMigration.php
bezin: Fixed in ccc498110

news-bundle/src/Migration/FeedMigration.php
bezin: Fixed in ccc498110

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: Yes, I agree. The feed itself exists, it just does not have any articles. So returning an empty feed seems right to me. Fixed in 699713dfe

news-bundle/src/Controller/Page/NewsFeedController.php
m-vo: You still need to add it to your subscriptions. Never rely on the implementation detail of the base class.

fritzmg (commented):
I found a bug in the FeedMigration - see PR: https://github.com/bezin/contao/pull/5

fritzmg (dismissed):

aschempp (dismissed):

ausi:
> Add tests

@bezin Is this still planned? 😬

bezin:
Tests are there, the check was missing

core-bundle/src/Controller/AbstractController.php
```diff
@@ -72,4 +74,53 @@ protected function getCsrfFormOptions(): array
...
+     * Set the cache headers according to the page settings.
+     *
+     * @param Response $response The response object
+     *
+     * @return Response The response object
+     */
+    protected function setCacheHeaders(Response $response, PageModel $pageModel)
+    {
+        // Do not cache the response if caching was not configured at all or disabled explicitly
+        if ((false === $pageModel->cache || $pageModel->cache < 1) && (false === $pageModel->clientCache || $pageModel->clientCache < 1)) {
```
leofeyer:
```suggestion
        if ($pageModel->cache < 1 && $pageModel->clientCache < 1)
```

core-bundle/src/Controller/AbstractController.php
```diff
@@ -72,4 +74,53 @@ protected function getCsrfFormOptions(): array
...
+            // However, we want to be able to force the reverse proxy to load a
+            // response from cache, even if the request contains a cookie – in
+            // case the admin has configured to do so. A typical use case would
+            // be serving public pages from cache to logged in members.
+            if (!$pageModel->alwaysLoadFromCache) {
+                $response->setVary(['Cookie']);
+            }
+
+            // Tag the page (see #2137)
+            $this->tagResponse($pageModel);
```
leofeyer:
```suggestion
            $this->container->get('contao.cache.entity_tags')->tagWithModelInstance($pageModel);
```

leofeyer (changes requested):
Setting `cache` or `clientCache` to `false` is no longer supported in Contao 5.

news-bundle/config/services.yaml
```diff
@@ -30,9 +48,22 @@ services:
...
+        class: Contao\NewsBundle\EventListener\NewsFeedListener
+        arguments:
+            - '@contao.framework'
+            - '@contao.image.factory'
+            - '@contao.insert_tag.parser'
+            - '%kernel.project_dir%'
+            - '@contao.cache.entity_tags'
+        tags:
+            - { name: kernel.event_listener, event: Contao\NewsBundle\Event\FetchArticlesForFeedEvent, method: onFetchArticlesForFeed }
+            - { name: kernel.event_listener, event: Contao\NewsBundle\Event\TransformArticleForFeedEvent, method: onTransformArticleForFeed }
```
leofeyer: Please use PHP 8 attributes to tag the listener.

news-bundle/contao/classes/News.php
```diff
@@ -26,261 +21,6 @@ class News extends Frontend
...
-	 * @var array
-	 */
-	private static $arrPageCache = array();
-
-	/**
-	 * Update a particular RSS feed
-	 *
-	 * @param integer $intId
-	 */
-	public function generateFeed($intId)
```
leofeyer, line 40: @contao/developers Do we really want to remove all these methods? None of them are deprecated in Contao 4.13.

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,101 @@
...
+use Contao\NewsBundle\Event\TransformArticleForFeedEvent;
+use Contao\PageModel;
+use FeedIo\Feed;
+use FeedIo\Specification;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @Page(contentComposition=false)
+ */
```
leofeyer:
```suggestion
#[AsPage(contentComposition: false)]
```

You also have to add a `use` statement for the class.

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,101 @@
...
+            ->setLanguage($pageModel->language)
+        ;
+
+        $event = new FetchArticlesForFeedEvent($feed, $request, $pageModel);
+        $this->container->get('event_dispatcher')->dispatch($event);
+
+        if (null !== ($articles = $event->getArticles())) {
+            foreach ($articles as $article) {
+                $event = new TransformArticleForFeedEvent($article, $feed, $pageModel, $request, $baseUrl);
+                $this->container->get('event_dispatcher')->dispatch($event);
```
leofeyer: This should not be inside a loop.

news-bundle/contao/classes/News.php
Toflar: It's not something we can deprecate. Deprecating it would mean we have to write "do this instead". But here is no "do this instead" because there is no alternative. In Contao 4.13 the data was written to disk using this method. Now it's a controller. It's an entirely different concept so what you'll have to do depends on what you did with this method.

news-bundle/src/EventListener/DataContainer/LayoutFeedOptionsListener.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Contao\NewsBundle\EventListener\DataContainer;
+
+use Contao\CoreBundle\Framework\ContaoFramework;
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Contao\NewsBundle\Controller\Page\NewsFeedController;
+use Contao\PageModel;
+
+/**
+ * @Callback(table="tl_layout", target="fields.newsfeeds.options")
+ */
```
leofeyer: This should be a PHP 8 attribute as well.

news-bundle/src/EventListener/DataContainer/PageListener.php
```diff
@@ -0,0 +1,81 @@
...
+
+class PageListener
+{
+    public function __construct(private Connection $connection, private Security $security)
+    {
+    }
+
+    /**
+     * @Callback(table="tl_page", target="config.onload")
+     */
```
leofeyer: All of these should be PHP 8 attributes.

news-bundle/src/EventListener/GeneratePageListener.php
```diff
@@ -43,21 +41,24 @@ public function __invoke(PageModel $pageModel, LayoutModel $layoutModel): void
...
         foreach ($feeds as $feed) {
-            $GLOBALS['TL_HEAD'][] = $template->generateFeedTag(
-                sprintf('%sshare/%s.xml', $feed->feedBase ?: $environment->get('base'), $feed->alias),
-                $feed->format,
-                $feed->title
-            );
+            if (NewsFeedController::TYPE !== $feed->type) {
+                continue;
+            }
+            // TODO: Use ResponseContext, once it supports appending to <head>
```
leofeyer, line 54: @Toflar I figured it already would?

news-bundle/src/EventListener/NewsFeedListener.php
```diff
@@ -0,0 +1,213 @@
...
+        $enclosures = $this->getEnclosures($article, $item, $event);
+
+        foreach ($enclosures as $enclosure) {
+            $item->addMedia($enclosure);
+        }
+
+        $event->setItem($item);
+    }
+
+    protected function getTitle(NewsModel $article, ItemInterface $item, TransformArticleForFeedEvent $event): string
```
leofeyer: Why is it necessary to pass the `$item` and `$event` objects of none of them are used?

news-bundle/src/EventListener/GeneratePageListener.php
Toflar: It does not, no.

news-bundle/tests/Controller/Page/NewsFeedControllerTest.php
```diff
@@ -0,0 +1,187 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Controller\Page;
```
leofeyer: This seems wrong. It should probably be `Contao\NewsBundle\Tests\Controller\Page`.

news-bundle/tests/EventListener/DataContainer/LayoutFeedOptionsListenerTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace EventListener\DataContainer;
```
leofeyer: This seems wrong, too.

leofeyer:
Is it really necessary to duplicate the dummy images?

news-bundle/src/EventListener/NewsFeedListener.php
bezin: Because it is `protected` and you may extend this class from your own listener to overwrite the methods.

news-bundle/tests/EventListener/DataContainer/LayoutFeedOptionsListenerTest.php
bezin: Fixed in a8cda0c9c

news-bundle/tests/Controller/Page/NewsFeedControllerTest.php
bezin: Fixed in a8cda0c9c

news-bundle/src/EventListener/DataContainer/PageListener.php
bezin: Fixed in d8ddd2965

news-bundle/src/EventListener/DataContainer/LayoutFeedOptionsListener.php
bezin: Fixed in d8ddd2965

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: Fixed in 3d6ac3e04

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: Fixed in d8ddd2965

news-bundle/config/services.yaml
bezin: Fixed in d8ddd2965

core-bundle/src/Controller/AbstractController.php
bezin: Did you use $objPage by accident?

bezin:
> Is it really necessary to duplicate the dummy images?

Since I need to check if the files exist, I would say yes. I do not see how I could split up the `getEnclosure` method in `NewsFeedListener` in a meaningful way.

core-bundle/src/Controller/AbstractController.php
bezin: Changed in 0a0fdf4e4

core-bundle/src/Controller/AbstractController.php
leofeyer: Yes, sorry. I have adjusted the suggestion.

bezin: Fixed in f57ef3ab5

ausi:
> > Is it really necessary to duplicate the dummy images?
> 
> Since I need to check if the files exist, I would say yes.

Couldn’t you create those images on the fly when running the tests? like we do here: https://github.com/contao/contao/blob/5480dd09fd77c2ad5fdea497dddd932986f9bdf9/core-bundle/tests/Image/Preview/PreviewFactoryTest.php#L348-L351

news-bundle/src/EventListener/NewsFeedListener.php
ausi: Since our event listeners are not part of the BC promise I’d go with `private` methods here.

core-bundle/src/Controller/AbstractController.php
leofeyer: The change is wrong, I‘m afraid. Why did you not commit my suggestion?

news-bundle/src/EventListener/NewsFeedListener.php
leofeyer: Me too. Extending listeners is kind of an anti-pattern.

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $response = new Response($formatter->toString($feed));
+        $response->headers->set('Content-Type', self::$contentTypes[$pageModel->feedFormat]);
+
+        $this->setCacheHeaders($response, $pageModel);
+
+        return $response;
+    }
+
+    public function configurePageRoute(PageRoute $route): void
```
leofeyer, line 89: Why is this method public?

news-bundle/src/Controller/Page/NewsFeedController.php
```diff
@@ -0,0 +1,104 @@
...
+        $format = $route->getPageModel()->feedFormat;
+
+        if (!isset($this->urlSuffixes[$format])) {
+            throw new \RuntimeException(sprintf('%s is not a valid format. Must be one of: %s', $format, implode(',', array_keys($this->urlSuffixes))));
+        }
+
+        $route->setUrlSuffix($this->urlSuffixes[$format]);
+    }
+
+    public function getUrlSuffixes(): array
```
leofeyer, line 100: What's the purpose of this method?

core-bundle/src/Controller/AbstractController.php
```diff
@@ -72,4 +74,53 @@ protected function getCsrfFormOptions(): array
...
+            // However, we want to be able to force the reverse proxy to load a
+            // response from cache, even if the request contains a cookie – in
+            // case the admin has configured to do so. A typical use case would
+            // be serving public pages from cache to logged in members.
+            if (!$pageModel->alwaysLoadFromCache) {
+                $response->setVary(['Cookie']);
+            }
+
+            // Tag the page (see #2137)
+            $this->container->get('contao.cache.entity_tags')->tagWithModelInstance($tags);
```
Toflar: What's `$tags` here? :)

core-bundle/src/Controller/AbstractController.php
bezin: Fixed in 0181a84c8

core-bundle/src/Controller/AbstractController.php
bezin: Fixed in 0181a84

bezin:
> Couldn’t you create those images on the fly when running the tests? like we do here:

Did so in [931f26b](https://github.com/contao/contao/pull/4682/commits/931f26b2b454981ae5c580645ab9548cc500239e) Thanks for the hint!

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: See [DynamicRouteInterface](https://github.com/contao/contao/blob/a5c401df53881cd09edd0e5dfc9813dbb659c251/core-bundle/src/Routing/Page/DynamicRouteInterface.php)

news-bundle/src/Controller/Page/NewsFeedController.php
bezin: See [DynamicRouteInterface](https://github.com/contao/contao/blob/a5c401df53881cd09edd0e5dfc9813dbb659c251/core-bundle/src/Routing/Page/DynamicRouteInterface.php)

news-bundle/src/EventListener/NewsFeedListener.php
bezin: Agreed, however it saves you work when you only want to extend one or some methods. 
But in lack of a better idea and the fact that I just want to get this across the finish line, I made them all private in 9f9f381a7

bezin: Removed unused arguments in [8344a47](https://github.com/contao/contao/pull/4682/commits/8344a47a174d419549371e79c7556a5f3267bfb5)

leofeyer (approved):

leofeyer:
Thank you @bezin.

PR #4683, opened 2022-05-17
Drop the DBAFS file size limit

### LINKED ISSUE #4678: File synchronisation deletes big files in 4.13.4
### Affected version(s)

4.13.4

### Description

I have a few video files on my server (2,5 GB and 289MB). After uploading them via ftp, file synchronisation finds all of them and i can choose them in content elements.

When using them in a content element "video", the big file (2,5GB) gets deleted at the next synchronisation. But the file remains on the server and i can still see it in contao file manager. The content element itself is loosing the video.

**Files before sync**
dateiverwaltung

**Content Element before sync**
vorher

**Sync after file is in a content element**
dateiverwaltung-synchronisation

**Content element after sync**
nachher

**Files after sync**
dateiverwaltung

This is probably a thing due filesize. According to https://community.contao.org/de/showthread.php?83124-Gro%C3%9Fe-Dateien-(Video)-werden-aus-der-DB-gel%C3%B6scht-wenn-im-Inhaltselement-eingebunden&p=559019&viewfull=1#post559019 this behaviour doesn´t appear in 4.9.x

Contao 4.13.4

m-vo:
This removes the 2GB file size limit for the DBAFS and therefore also removes incoherent behavior between the old and new implementation.

see https://github.com/contao/contao/issues/4678#issuecomment-1128184839

@Kahmoon Can you please check, if this solves your issue?

### DIFF
--- a/core-bundle/src/Filesystem/Dbafs/Dbafs.php
+++ b/core-bundle/src/Filesystem/Dbafs/Dbafs.php
@@ -51,7 +51,6 @@ class Dbafs implements DbafsInterface, ResetInterface
 
     private string $table;
     private string $dbPathPrefix = '';
-    private int $maxFileSize = 2147483647; // 2 GiB - 1 byte (see #4208)
     private int $bulkInsertSize = 100;
     private bool $useLastModified = true;
 
@@ -90,7 +89,7 @@ class Dbafs implements DbafsInterface, ResetInterface
 
     public function setMaxFileSize(int $bytes): void
     {
-        $this->maxFileSize = $bytes;
+        trigger_deprecation('contao/core-bundle', '4.13', 'Setting a maximum file size has no effect anymore. The "%s()" method will be removed in Contao 5.', __METHOD__);
     }
 
     public function setBulkInsertSize(int $chunkSize): void
@@ -740,11 +739,6 @@ class Dbafs implements DbafsInterface, ResetInterface
 
                 // Ignore dot files
                 if (0 === strpos(basename($path), '.')) {
-                    continue;
-                }
-
-                // Ignore files that are too big
-                if ($item->getFileSize() > $this->maxFileSize) {
                     continue;
                 }
 
--- a/core-bundle/src/Filesystem/Dbafs/Dbafs.php
+++ b/core-bundle/src/Filesystem/Dbafs/Dbafs.php
@@ -87,6 +87,9 @@ class Dbafs implements DbafsInterface, ResetInterface
         $this->dbPathPrefix = Path::canonicalize($prefix);
     }
 
+    /**
+     * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0.
+     */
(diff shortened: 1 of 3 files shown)


core-bundle/src/Resources/contao/library/Contao/File.php
```diff
@@ -879,12 +879,6 @@ protected function getIcon()
 	 */
 	protected function getHash()
 	{
-		// Do not try to hash if bigger than 2 GB
-		if ($this->filesize >= 2147483648)
-		{
-			return '';
-		}
-
 		return md5_file($this->strRootDir . '/' . $this->strFile);
```
fritzmg, line 882: Wouldn't this still cause performance problems though?

Kahmoon:
The file is not removed in sync anymore. Content element also is fine before/after sync

core-bundle/src/Resources/contao/library/Contao/File.php
m-vo: @ausi performed some tests yesterday and could not find any downsides. There seemed to be an old bug on windows with files > 2GB, but I guess that is resolved in the meantime.

ausi: The performance I measured responds linear to the file size. So there is no difference between ten 1GB files and one 10GB file AFAIK.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/File.php
fritzmg: > There seemed to be an old bug on windows with files > 2GB, but I guess that is resolved in the meantime.

You guess or should I check? 😁

fritzmg: I could find no issues (other than the hash calculation taking 4500 ms for a 2.5 GiB test file in my case).

fritzmg (dismissed):

core-bundle/src/Resources/contao/library/Contao/File.php
m-vo: Thanks for testing!

core-bundle/src/Filesystem/Dbafs/Dbafs.php
```diff
@@ -90,7 +89,7 @@ public function setDatabasePathPrefix(string $prefix): void
 
     public function setMaxFileSize(int $bytes): void
```
leofeyer, line 93:
```suggestion
    /**
     * @deprecated Deprecated since Contao 4.13, to be removed in Contao 5.0.
     */
    public function setMaxFileSize(int $bytes): void
```

So the IDE will also show a deprecation notice.

bytehead (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #4686, opened 2022-05-17
Always delete indexed page for unsuccessful responses

### LINKED ISSUE #4685: Remove site from search index on 404
### Affected version(s)

4.13

### Description

If a site has been indexed and is unpublished afterwards the site can still be found using the search. Clicking the result leads to a 404 page of course but the site does not get removed from the search index then. This is due to several checks already failing

https://github.com/contao/contao/blob/50502c09eec66d75b1c1ae280ceec90a0ab4ca62/core-bundle/src/EventListener/SearchIndexListener.php#L44-L84

before we reach the actual code that would remove the site from the index

https://github.com/contao/contao/blob/50502c09eec66d75b1c1ae280ceec90a0ab4ca62/core-bundle/src/EventListener/SearchIndexListener.php#L93-L95

@fritzmg Suggested in Slack to wrap `SearchIndexListener.php#L44-L84` into a separate `needsIndex(): bool` method so we can always reach the part where the removal would take place.

fritzmg:
Fixes #4685

Our `SearchIndexListener` is supposed to delete search entries for unsuccessful responses. However, as outlined in #4685 this currently does not work, because the early outs checking for whether the page should be indexed will also prevent the URL from being _removed_ from the search index. `404` responses with the default 404 page will for instance never have any `JSON-LD` content and 404 responses with a custom 404 page will always have `X-Robots-Tag: noindex`. Thus pages will actually never be removed from the index currently.

This PR fixes that by separating the check whether the requested page should be indexed and instead always issue a document deletion in case the response is unsuccessful.

### DIFF
--- a/core-bundle/src/EventListener/SearchIndexListener.php
+++ b/core-bundle/src/EventListener/SearchIndexListener.php
@@ -57,9 +57,7 @@ class SearchIndexListener
     {
         $request = $event->getRequest();
         $response = $event->getResponse();
-
         $document = Document::createFromRequestResponse($request, $response);
-
         $needsIndex = $this->needsIndex($request, $response, $document);
 
         try {
--- a/core-bundle/src/EventListener/SearchIndexListener.php
+++ b/core-bundle/src/EventListener/SearchIndexListener.php
@@ -17,6 +17,7 @@ use Contao\CoreBundle\Search\Document;
 use Contao\CoreBundle\Search\Indexer\IndexerException;
 use Contao\CoreBundle\Search\Indexer\IndexerInterface;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\TerminateEvent;
 
 /**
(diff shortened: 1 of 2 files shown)


core-bundle/src/EventListener/SearchIndexListener.php
```diff
@@ -55,37 +56,55 @@ public function __construct(IndexerInterface $indexer, string $fragmentPath = '_
...
     {
         $request = $event->getRequest();
+        $response = $event->getResponse();
+
+        $document = Document::createFromRequestResponse($request, $response);
+
+        $needsIndex = $this->needsIndex($request, $response, $document);
+
+        try {
+            $success = $event->getResponse()->isSuccessful();
```
Toflar, line 64: Why is this inside the try block and `$needsIndex` is not?

Toflar (dismissed):
Minor nitpick

core-bundle/src/EventListener/SearchIndexListener.php
fritzmg: That's how it was before. My assumption is the try catch is only supposed to catch exceptions regarding `$event->getResponse()->isSuccessful();`?

Toflar: Can that throw anywhere? :D

fritzmg: idk, you implemented it in https://github.com/contao/contao/commit/23630c179d6a46557edef3b278ab23fadc3168ed ;). I see no reason to change it.

Toflar: kk, well just code style but Idc really :)

aschempp:
~I'm not an expert on the search indexer, but doesn't `needsIndex` return false if the document is up to date in the database?! If that's true, this change would always delete the page?~

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #4691, opened 2022-05-18
Fix failing backup on contao:migrate must abort the command

Toflar:
Unfortunately, I just had the problem that an error occured during the creation of a backup in `contao:migrate`. This failed but the migrate command still continued. So I'm now left with a broken DB and no backup :( 

Luckily, the hosting provider had one so I'm good but it should not happen again :-)).


core-bundle/src/Command/MigrateCommand.php
```diff
@@ -77,8 +77,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $this->io = new SymfonyStyle($input, $output);
 
-        if (!$input->getOption('dry-run') && !$input->getOption('no-backup')) {
-            $this->backup($input);
+        if (!$input->getOption('dry-run') && !$input->getOption('no-backup') && !$this->backup($input)) {
+            return 1;
```
ausi, line 81: Note: needs to be changed to `return Command::FAILURE;` when merged upstream, cc @leofeyer

ausi (dismissed):

aschempp (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4692, opened 2022-05-18
Harden against invalid input

leofeyer:
The changes are based on my Sentry reports. As far as I see, they have been triggered by some security research tool.

All of the requests were caught by the error handler, mostly because they triggered PHP type errors such as:

* A non-numeric value encountered
* Warning: substr() expects parameter 1 to be string, array given
* Warning: preg_match() expects parameter 2 to be string, array given

So actually, we do not have to handle these cases at all because they will trigger the error handler one way or another. But maybe a `BadRequestHttpException` is better than some unspecific `TypeError` or `ErrorException`?

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Contao\CoreBundle\Exception\PageNotFoundException;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 /**
  * Front end module "event list".
@@ -98,6 +99,11 @@ class ModuleEventlist extends Events
 		$intYear = Input::get('year');
 		$intMonth = Input::get('month');
 		$intDay = Input::get('day');
+
+		if (\is_array($intYear) || \is_array($intMonth) || \is_array($intDay))
+		{
+			throw new BadRequestHttpException('Expected string, got array');
+		}
 
 		// Handle featured events
 		$blnFeatured = null;
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -11,7 +11,6 @@
 namespace Contao;
 
 use Contao\CoreBundle\Exception\PageNotFoundException;
-use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 /**
  * Front end module "event list".
@@ -96,14 +95,9 @@ class ModuleEventlist extends Events
 
 		$blnClearInput = false;
 
-		$intYear = Input::get('year');
-		$intMonth = Input::get('month');
-		$intDay = Input::get('day');
-
-		if (\is_array($intYear) || \is_array($intMonth) || \is_array($intDay))
-		{
-			throw new BadRequestHttpException('Expected string, got array');
-		}
+		$intYear = (int) Input::get('year');
+		$intMonth = (int) Input::get('month');
+		$intDay = (int) Input::get('day');
 
 		// Handle featured events
 		$blnFeatured = null;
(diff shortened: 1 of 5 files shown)


calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
```diff
@@ -99,6 +100,11 @@ protected function compile()
 		$intMonth = Input::get('month');
 		$intDay = Input::get('day');
 
+		if (\is_array($intYear) || \is_array($intMonth) || \is_array($intDay))
```
ausi: I’d cast the values instead I think.

```php
$intYear = (int) Input::get('year');
$intMonth = (int) Input::get('month');
$intDay = (int) Input::get('day');
```

core-bundle/src/EventListener/RequestTokenListener.php
```diff
@@ -93,7 +95,13 @@ public function __invoke(RequestEvent $event): void
             }
         }
 
-        $token = new CsrfToken($this->csrfTokenName, $this->getTokenFromRequest($request));
+        $value = $this->getTokenFromRequest($request);
+
+        if (Validator::isInsecurePath($value)) {
+            throw new BadRequestHttpException('Invalid CSRF token');
+        }
```
ausi: Why is the CSRF token considered a path here?
Wouldn’t this error be checked by `isTokenValid()` from below and trigger the `InvalidRequestTokenException` anyways?

calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
leofeyer: Interesting. Casting to string gives a `Warning: Array to string conversion` but casting to int seems to be no problem. 😄

core-bundle/src/EventListener/RequestTokenListener.php
leofeyer: The request had `../1` for `REQUEST_TOKEN`, which ultimately led to a `A non-numeric value encountered` warning.

ausi: Wouldn’t `foo/1` then also lead to `A non-numeric value encountered`?
Should I investigate where the issue occurs and throw a proper exception there?

leofeyer: Yes please. 😊

fritzmg:
Generally I wonder if we really need to fix any "hacking" attempts that lead to non-harmful errors anyway. If we do, I can probably dig up some more from our customer's sentries.

m-vo:
> Generally I wonder if we really need to fix any "hacking" attempts that lead to non-harmful errors anyway. 

IMHO they should never lead to something in the 5xx regime or related, because then they probably found abnormal state which we must avoid in terms of security.

core-bundle/src/EventListener/RequestTokenListener.php
ausi: Someone else already did that last week ☺️
Will be fixed in the next version of Symfony: https://github.com/symfony/symfony/pull/46309

leofeyer: Nice. 😎

calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
leofeyer: Changed in c8995f1152e9e9b84ec3a0c9259427cfa0316c27. Hope PHP does not add a `Warning: Array to int conversion` anytime soon. 😄

core-bundle/src/EventListener/RequestTokenListener.php
leofeyer: I have reverted the changes in c8995f1152e9e9b84ec3a0c9259427cfa0316c27.

ausi (approved):

ausi:
> I can probably dig up some more from our customer's sentries.

That would be a good idea I think. If it’s not too much work. ☺️

PR #4698, opened 2022-05-19
Fix broken filemanager for unsupported images

### LINKED ISSUE #4695: Image type "heic" was not allowed to be processed
### Affected version(s)

4.13.4

### Description

When a user tried to upload a HEIC Image he got the error
`Image type "heic" was not allowed to be processed`
and could not use the file manager anymore.
I think the problem is that isImage returns true for the filetype
[File.php:364
](https://github.com/contao/contao/blob/0627361c011bb2a2295d912d682f1e4465786716/core-bundle/src/Resources/contao/library/Contao/File.php#L365) 
```
case 'isGdImage':
return \in_array($this->extension, array('gif', 'jpg', 'jpeg', 'png', 'webp', 'avif', 'heic', 'jxl'));
```
So it tries to generate the thumbnail in [DC_Folder:3807](https://github.com/contao/contao/blob/0627361c011bb2a2295d912d682f1e4465786716/core-bundle/src/Resources/contao/drivers/DC_Folder.php#L2807)
and then throws an error in [ImageFactory:92](https://github.com/contao/core-bundle/blob/b9280e5803c81c152e14c865061d134e54f68397/src/Image/ImageFactory.php#L92) as it is not in the list of default [valid_extensions](https://github.com/contao/core-bundle/blob/b9280e5803c81c152e14c865061d134e54f68397/src/DependencyInjection/Configuration.php#L345) ['jpg', 'jpeg', 'gif', 'png', 'tif', 'tiff', 'bmp', 'svg', 'svgz', 'webp']

ausi:
Fixes #4695

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2793,7 +2793,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			$thumbnail .= ')</span>';
 
 			// Generate the thumbnail
-			if ($objFile->isImage && (!$objFile->isSvgImage || $objFile->viewHeight > 0) && Config::get('thumbnails') && in_array($objFile->extension, System::getContainer()->getParameter('contao.image.valid_extensions')))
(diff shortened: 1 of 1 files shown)


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4701, opened 2022-05-19
Remove hook addLogEntry and its usages

bytehead:
See #4513

`contao/docs` needs to be adjusted as well.

### DIFF
--- a/core-bundle/tests/Monolog/ContaoTableHandlerTest.php
+++ b/core-bundle/tests/Monolog/ContaoTableHandlerTest.php
@@ -43,13 +43,8 @@ class ContaoTableHandlerTest extends TestCase
         $this->assertSame('foobar', $handler->getDbalServiceName());
     }
 
-    /**
-     * @group legacy
-     */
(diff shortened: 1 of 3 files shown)


core-bundle/tests/Monolog/ContaoTableHandlerTest.php
```diff
@@ -43,13 +42,8 @@ public function testSupportsReadingAndWritingTheDbalServiceName(): void
         $this->assertSame('foobar', $handler->getDbalServiceName());
     }
 
-    /**
```
Toflar, line 46: Imho the `tearDown()` can also be removed in this method now.

bytehead: Good catch!

bytehead: Removed in 52b5accdd.

Toflar (approved):

leofeyer:
Thank you @bytehead.

PR #4707, opened 2022-05-20
Try auto-committing CS fixes

leofeyer:


leofeyer:
I can‘t get it to work.

I have set up everything as described in [these docs](https://github.com/stefanzweifel/git-auto-commit-action#workflow-should-run-in-base-repository) and according to the discussion in [git-auto-commit-action#211](https://github.com/stefanzweifel/git-auto-commit-action/issues/211) it should work without a personal access token. However, the workflow keeps failing with a `Permission to leofeyer/contao.git denied to github-actions[bot]` error.

I‘m giving up (again) for the moment. 😄

ausi:
> If the workflow is triggered by the pull_request_target-event, the workflow will run in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. In other words, this will allow your workflow to be run in the repository where the pull request is opened to and will push changes back to the fork.

IMO this means: To actually test this, the code in the ci.yml file hast to be merged first. And if a pull request happens then afterwards it would actually execute. (We could create a test-branch for that in contao/contao)

We probably also need to check for `if: github.event_name == 'pull_request_target'`

But I’m seeing a bigger problem here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
> Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event.

But we would check out untrusted code and run it, so this would be highly insecure.
The correct way (I think) would be to reset the `tools/` directory first to the state from the pull-request-base (so that the fork cannot do anything bad to your ecs executable), and then run it.

leofeyer:
Reverting this change should partially work around the security problem:



But IIRC, the workflow can still read the secrets, so we don‘t want to use `pull_request_target` at all, I guess.

ausi:
Maybe it could work this way:

1. the CS fixer job runs normally as it does now (without any push rights) and stores the diff as an artifact (or something similar)
2. a second job runs via `pull_request_target` that just applies this stored diff and commits it to the PR.

This way everything should be secure.

PR #4708, opened 2022-05-22
Increase the tl_comments_notify.url length

fritzmg:
Similar to `tl_search.url` the following error can occur when subscribing to comments:

```
An exception occurred while executing 'INSERT INTO tl_comments_notify (`tstamp`, `source`, `parent`, `name`, `email`, `url`, `addedOn`, `active`, `tokenRemove`) VALUES (1653216741, 'tl_news', 324, 'TEST', 'foobar@example.com', 'news/article/news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long-headline-news-with-long', 1653216741, '', 'cor-86c865dfa67ad5f886f3')':

SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'url' at row 1
```

This PR increases the `tl_comments_notify.url` length, in line with `tl_content.url` and `tl_search.url`.

_Note:_ no automatic length adjustment is required here, as there is no index on this field.

### DIFF
--- a/comments-bundle/src/Resources/contao/classes/Comments.php
+++ b/comments-bundle/src/Resources/contao/classes/Comments.php
@@ -12,6 +12,7 @@ namespace Contao;
 
 use Contao\CoreBundle\Exception\PageNotFoundException;
 use Contao\CoreBundle\OptIn\OptIn;
+use Nyholm\Psr7\Uri;
 
 /**
  * Class Comments
@@ -558,6 +559,9 @@ class Comments extends Frontend
 
 		$time = time();
 
+		// Make sure the request URL does not contain non-ascii characters (see #4708)
+		$request = (string) (new Uri(Environment::get('request')));
+
 		// Prepare the record
 		$arrSet = array
 		(
@@ -566,7 +570,7 @@ class Comments extends Frontend
 			'parent'       => $objComment->parent,
 			'name'         => $objComment->name,
 			'email'        => $objComment->email,
-			'url'          => Environment::get('request'),
+			'url'          => $request,
 			'addedOn'      => $time,
 			'active'       => '',
 			'tokenRemove'  => 'cor-' . bin2hex(random_bytes(10))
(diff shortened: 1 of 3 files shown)


ausi:
Do we need to make sure that there are no unicode characters in the URL here? https://github.com/contao/contao/blob/f247bc966e59bea997b5c06683c20a95b780b729/comments-bundle/src/Resources/contao/classes/Comments.php#L569 like we did for tl_search.url here: https://github.com/contao/contao/blob/f247bc966e59bea997b5c06683c20a95b780b729/core-bundle/src/Resources/contao/library/Contao/Search.php#L61-L69

aschempp (dismissed):

fritzmg:
> Do we need to make sure that there are no unicode characters in the URL here?

Yes we do actually. Should we create a new method for this in the `StringUtil`? e.g. `StringUtil::encodeNonAsciiChars` or so. Or parse the URI with `Nyholm\Psr7\Uri`, which would do this automatically.

ausi:
> Or parse the URI with `Nyholm\Psr7\Uri`,

This is the way to go I think.

ausi (dismissed):

bytehead (dismissed):

comments-bundle/src/Resources/contao/classes/Comments.php
```diff
@@ -558,6 +559,9 @@ public static function addCommentsSubscription(CommentsModel $objComment)
 
 		$time = time();
 
+		// Make sure the request URL does not contain non-ascii characters (see #4708)
+		$request = (string) (new Uri(Environment::get('request')));
```
leofeyer, line 563: This should be added here as well, shouldn‘t it?

https://github.com/contao/contao/blob/f247bc966e59bea997b5c06683c20a95b780b729/core-bundle/src/Resources/contao/library/Contao/Search.php#L61-L69

fritzmg: Nyholm's implementation differs slightly - for BC we should probably keep it the way it is now. I personally wouldn't mind changing it though. @ausi wdyt?

ausi: I’d change the the Search.php to `Nyholm\Psr7\Uri` too, as it handles all cases, and all of them correctly.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4711, opened 2022-05-23
Correctly support legacy DC names in picker provider

aschempp:
The changes in https://github.com/contao/contao/pull/4481 lead to breaking support for DCA and picker with the legacy name.

FYI we don't need to trigger a deprecation for comparing the old name, because calling `DataContainer::getDriverForTable` already does that.

### DIFF
--- a/core-bundle/src/Picker/AbstractTablePickerProvider.php
+++ b/core-bundle/src/Picker/AbstractTablePickerProvider.php
@@ -137,9 +137,10 @@ abstract class AbstractTablePickerProvider implements PickerProviderInterface, D
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
-        return ($this->getDataContainer() === DataContainer::getDriverForTable($table)
-                || $this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
-            ) && 0 !== \count($this->getModulesForTable($table));
+        $dcName = $this->getDataContainer();
+
+        return ($dcName === DataContainer::getDriverForTable($table) || $dcName === $GLOBALS['TL_DCA'][$table]['config']['dataContainer'])
+            && 0 !== \count($this->getModulesForTable($table));
     }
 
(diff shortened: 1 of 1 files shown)


Toflar:
We should add a todo though to remove it again in the 5.x branch.

fritzmg:
Why is this necessary? According to the doc `$this->getDataContainer()` always returns the FQCN.

aschempp:
Yes, but `$this->getDataContainer()` might return the old string name

bytehead:
> Yes, but $this->getDataContainer() might return the old string name

In which cases? May we should add a test for this.

aschempp:
> In which cases? 

If an extension supports Contao versions before we added support for FQCN table names, it must use the legacy name both in the DCA file as well as the picker implementation.

core-bundle/src/Picker/AbstractTablePickerProvider.php
```diff
@@ -137,8 +137,9 @@ public function supportsContext($context): bool
         $this->framework->initialize();
         $this->framework->createInstance(DcaLoader::class, [$table])->load();
 
-        return $this->getDataContainer() === DataContainer::getDriverForTable($table)
-            && 0 !== \count($this->getModulesForTable($table));
+        return ($this->getDataContainer() === DataContainer::getDriverForTable($table)
+                || $this->getDataContainer() === $GLOBALS['TL_DCA'][$table]['config']['dataContainer']
+            ) && 0 !== \count($this->getModulesForTable($table));
```
leofeyer:
```suggestion
        $dc = $this->getDataContainer();

        return ($dc === DataContainer::getDriverForTable($table) || $dc === $GLOBALS['TL_DCA'][$table]['config']['dataContainer'])
            && 0 !== \count($this->getModulesForTable($table));
```

In case the `getDataContainer()` call is expensive, we should probably store its result in a variable.

aschempp: Changed in 9424537b68da0134f578619579cfd818f087550a

leofeyer (approved):

Issue #4699, opened 2022-05-19
Page settings inheritance and shared cache


qzminski:
### Affected version(s)

4.13.x

### Description

It seems that there is a problem with page settings inherited data when the shared page cache is enabled:

1. If you open a regular page it will get cached. Then, if you enable the maintenance mode in the website root settings and visit the same page again, there will be no maintenance mode displayed, but rather the page content exactly as it has been cached.

2. If you open a regular page that has no layout defined (= inherits layout from parent page) it will get cached. Then, if you change the layout in the parent page and visit the same page again, the old layout will be displayed.

Possible solutions:

1. Add multiple response cache tags `contao.db.tl_page.` for every page ID in the current page trail.
2. Invalidate all children if a parent record is modified in the backend.

/cc @Toflar

Toflar:
I think we should have a general response listener that adds the page tags and the trail (plus root page). I would do this in 4.13 only (even though it also affects 4.9) as there's a better way to access the page model in all circumstances. 4.9 still has quite some inconsistencies and I don't think it's worth investing. After all this is not a "bug bug". Cache invalidation when you edit is convenience, you shouldn't rely on it.

ausi:
Wouldn’t it fit the current concept better if we would fix this in the backend?

We currently invalidate all parent (ptable) and all child (ctable) items, I think we should extend that so that parent and child items from tree views are handled the same way.

fritzmg:
> 1. If you open a regular page it will get cached. Then, if you enable the maintenance mode in the website root settings and visit the same page again, there will be no maintenance mode displayed, but rather the page content exactly as it has been cached.

imho that's expected behavior. But the maintenance mode switch for the root could of course be changed to also invalidate the respective cache entries.

> Wouldn’t it fit the current concept better if we would fix this in the backend?

I agree. Also we should fix this in 4.9 then.

aschempp:
This is really easy to fix even for 4.9, see https://github.com/contao/contao/pull/4714

PR #4714, opened 2022-05-23
Tag the http cache with page trail

aschempp:
Fixes https://github.com/contao/contao/issues/4699

### DIFF
--- a/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendTemplate.php
@@ -412,8 +412,15 @@ class FrontendTemplate extends Template
 			// Tag the page (see #2137)
 			if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
 			{
+				$arrTags = array('contao.db.tl_page.' . $objPage->id);
+
+				foreach ($objPage->trail as $id)
+				{
+					$arrTags[] = 'contao.db.tl_page.' . $id;
+				}
+
 				$responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
-				$responseTagger->addTags(array('contao.db.tl_page.' . $objPage->id));
+				$responseTagger->addTags($arrTags);
 			}
 		}
 


fritzmg (approved):

ausi:
https://github.com/contao/contao/issues/4699#issuecomment-1131600298

> Wouldn’t it fit the current concept better if we would fix this in the backend?
> 
> We currently invalidate all parent (ptable) and all child (ctable) items, I think we should extend that so that parent and child items from tree views are handled the same way.

With the implementation of this pull request, changing any content element in one of the pages in the trail would clear the cache which is not desired, is it?

Toflar (changes requested):
The root page id is missing. This means the problem with e.g. the maintenance mode or any root page layout changes would not be applied.

Also, I personally do think this is correct but @ausi and @fritzmg said they'd rather invalidate the whole tree in the back end. How come @fritzmg is now approving this?

aschempp:
> With the implementation of this pull request, changing any content element in one of the pages in the trail would clear the cache which is not desired, is it?

No, that's not the case. Changing a content element would purge the `tl_content.XX` ID, which is not present on child pages. It would not purge the page cache ID. However, changing the title of a parent page would purge all child pages. But that could be desired, e.g. for a breadcrumb or because not the title but the permissions was changed.


> The root page id is missing.

I don't think that's true. `PageModel::loadDetails()` does this:

```
if ($type != 'root')
{
    $trail[] = $objParentPage->pid;
}
```

Which means the **parent ID** of all pages are added, and the parent ID of the first page within a root will be the root ID.

Toflar (approved):

Toflar:
This will be hard to merge in 4.13 as we have a entity tagger service there. I wonder if that service should somehow automatically understand if we have a tree structure and tag the parents itself. Or maybe we need some kind of `getTrail()` interface on the model, I don't know. /cc @m-vo

ausi:
> No, that's not the case. Changing a content element would purge the `tl_content.XX` ID, which is not present on child pages.

I’m unable to reproduce that.

Saving a content element purges the following tags in my test:

- `contao.db.tl_content.XX`
- `contao.db.tl_article.XX`
- `contao.db.tl_page.XX`
- `contao.db.tl_page`

ausi:
For reference, we agreed on the current concept (tag few, invalidate many) here: https://github.com/contao/contao/issues/2137#issuecomment-681944410

Toflar:
Yeah, I can see the problem. This is because we already take all the parents and children in https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/classes/DataContainer.php#L1487.
So I wonder why it doesn't work if I edit the root page? It should invalidate all children then, no? 🤔

ausi:
> So I wonder why it doesn't work if I edit the root page? It should invalidate all children then, no? 🤔

Yes, it should, but we just have not implemented this for tree structures yet, but only for ptable and ctable.

Toflar:
Ah, I agree we should do this instead then. Let me fix that.

Toflar:
Meh, a lot more questions arise so stopping here and marking `up for discussion`.

ausi (changes requested):
See https://github.com/contao/contao/pull/4714#issuecomment-1135542708

fritzmg:
> How come @fritzmg is now approving this?

It seemed like a sensible solution. But I was not aware that editing one content element would then also invalidate the whole trail.

m-vo:
>  I wonder if that service should somehow automatically understand if we have a tree structure and tag the parents itself. Or maybe we need some kind of getTrail() interface on the model, I don't know

Hmm, I tried to automate things for entities, but gave up on that idea, because everything was very use-case-dependent. I mean, we could know what the usage is and if a change affects you or not for specific fields/combinations, but not in general.

leofeyer:
As discussed in the Contao call, this should not be necessary. @Toflar will create a follow-up PR.

PR #4715, opened 2022-05-23
Support our template hierarchy in the Twig "use" tag

m-vo:
Using our template hierarchy was already possible in `{% extends … %}` and `{% include … %}` tags. This PR now also adds the last missing piece for the `{% use … %}` tag.

The `use` tag is a Twig feature, that is rarely needed. But, turns out: together with our dynamic hierarchy, this makes up for an extremely powerful tool for framework-provided reusable components. :tada: 

## Background
The use tag imports blocks from another template without displaying them - and you can use as many resources as you want from one template (works kind of like traits). Blocks can then be output with the block function and even the `parent()` function works. 

We can now utilize this behavior this to reuse components in our content element templates (instead of including/embedding them). The ultimate benefit is, that you can then adjust **every block of every used component** of the base template in your application's template :sunglasses:  - not only those directly introduced in the base template (or the base's base)<sup>1</sup>:

```twig
{# templates/content_element/gallery.html.twig #}
{% extends "@Contao/content_element/gallery.html.twig" %}

{# Add a text after the image list #}
{% block content %}
    {{ parent() }}
    <p>In case you didn't notice: This up here is a gallery!</p>
{% endblock %}

{# Add a wrapper around each image -- from the "list" component #}
{% block list_element %}
    <div class="list-wrapper">{{ parent() }}</div>
{% endblock %}

{# Do not render HTML in the caption -- from the "figure" component #}
{% block caption_inner %}
    {{- figure.metadata.caption -}}
{% endblock %}
```

FYI: In this example, under the hood, the core's (upcoming) `content_element/gallery` template is using the `component/_list` and `component/_figure` templates.  It looks somewhat like this:
```twig
{% use "@Contao/component/_list.html.twig" %}
{% use "@Contao/component/_figure.html.twig" %}

{% block content %}
    {% set items = images %}
    {{ block('list_component') }}
{% endblock %}

{% block list_item %}
    {% set figure = item %}
    {{ block('figure_component') }}
{% endblock %}
```
The components itself then contain all the blocks you're seeing above (like `list_element` or `caption_inner`).

You can also adjust components, if the changes should effect all usages:
```twig
{# templates/component/_list.html.twig #}
{% use "@Contao/component/_list.html.twig" %}

{% block list_element %}
    <div class="list-wrapper">{{ parent() }}</div>
{% endblock %}
```

<sup>1) A good naming for blocks is crucial for this, but as we control the components that we use in our templates, this is not that big of a problem. Plus: it is also possible to rename blocks on import: `{% use "my_component.html.twig" with foo as bar %}` to work around possible naming conflicts.</sup>

### DIFF
--- a/core-bundle/src/Twig/Inheritance/DynamicUseTokenParser.php
+++ b/core-bundle/src/Twig/Inheritance/DynamicUseTokenParser.php
@@ -47,7 +47,6 @@ final class DynamicUseTokenParser extends AbstractTokenParser
         if ($stream->nextIf('with')) {
             while (true) {
                 $name = $stream->expect(Token::NAME_TYPE)->getValue();
-
                 $alias = $name;
 
                 if ($stream->nextIf('as')) {
--- a/core-bundle/src/Twig/Inheritance/DynamicUseTokenParser.php
+++ b/core-bundle/src/Twig/Inheritance/DynamicUseTokenParser.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Inheritance;
+
+use Contao\CoreBundle\Twig\ContaoTwigUtil;
+use Twig\Error\SyntaxError;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Node;
+use Twig\Token;
+use Twig\TokenParser\AbstractTokenParser;
+
+/**
+ * This parser is a drop in replacement for @see \Twig\TokenParser\UseTokenParser
+ * that adds support for the Contao template hierarchy.
+ *
+ * @experimental
+ */
+final class DynamicUseTokenParser extends AbstractTokenParser
+{
+    public function __construct(private TemplateHierarchyInterface $hierarchy)
+    {
+    }
+
+    public function parse(Token $token): Node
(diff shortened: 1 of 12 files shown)


ausi (dismissed):
❤️

Toflar (dismissed):
In other words: Support for horizontal reuse is landing ❤️

core-bundle/src/Twig/Inheritance/DynamicUseTokenParser.php
```diff
@@ -0,0 +1,90 @@
...
+    public function parse(Token $token): Node
+    {
+        $template = $this->parser->getExpressionParser()->parseExpression();
+        $stream = $this->parser->getStream();
+
+        if (!$template instanceof ConstantExpression) {
+            throw new SyntaxError('The template references in a "use" statement must be a string.', $stream->getCurrent()->getLine(), $stream->getSourceContext());
+        }
+
+        $this->adjustTemplateName($stream->getSourceContext()->getPath(), $template);
```
leofeyer, line 43: The `getSourceContext()` method is marked `@internal`. Is there a way for us to work around this?

m-vo: No, unfortunately (same for the `extends` tag btw.).

It is very unlikely that this functionality will be removed, though, because Twig needs it itself. But we should probably pitch a change in the Twig repo to make it a public API once our implementation has settled.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4716, opened 2022-05-23
Add two more FilesystemItemIterator related helper functions 

m-vo:
This PR adds two things that eases using the new filesystem:

1. The `FilesystemItemIterator` now got a `limit()` function, that allows capping the total amount of yielded items. This is especially nice for large sets, because this way they are basically streamed and do not have to be processed when not needed [4079932dabd0d85f4f6340da45a54f4249f46991]:

    ```php
    foreach ($filesStorage->listContent('images')->limit(10) as $filesystemItem) {
        // …
    }
    ```
2. I added a helper function that returns a `FilesystemItemIterator` for binary-string UUIDs in arbitrary format. This makes it easy to get the resources that are in a representation as stored in `tl_content.singleSRC`, `tl_content.multiSRC` or `tl_user.homeDir`. [1e1c1aa8c6ba2c653369b3fa31256dfcf756cc46]


core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -97,6 +97,27 @@ public function all(callable $condition): bool
         return true;
     }
 
+    public function limit(int $numberOfElements): self
+    {
+        if ($numberOfElements < 0) {
```
Toflar, line 102:
```suggestion
        if ($numberOfElements <= 0) {
```

Or not? Or maybe better just throw an `\OutOfBoundsException`? In which case would passing 0 or negative ever make any sense?

core-bundle/src/Filesystem/FilesystemUtil.php
```diff
@@ -12,12 +12,36 @@
...
+     * for instance stored in "tl_content.multiSRC".
+     *
+     * The following rules apply:
+     *  - Invalid or non-existent UUIDs are skipped without producing an error.
+     *  - If the UUID points to a directory, its file contents are used (only first level).
+     *  - Duplicate files are stripped.
+     *
+     * @param string|array<string|null> $sources
+     */
+    public static function listContentsFromMultiSRC(VirtualFilesystemInterface $storage, string|array $sources): FilesystemItemIterator
```
Toflar:
```suggestion
    public static function listContentsFromSerialized(VirtualFilesystemInterface $storage, string|array $sources): FilesystemItemIterator
```

Because maybe in the future we have a json storage. `MultiSRC` is not the definition. The format is. Which is serialized in this case :)

core-bundle/src/Filesystem/FilesystemUtil.php
```diff
@@ -44,4 +70,51 @@ public static function rewindStream($resource): void
...
+            if ($new = !\in_array($path = $item->getPath(), $paths, true)) {
+                $paths[] = $path;
+            }
+
+            return $new;
+        };
+
+        foreach ($uuids as $uuid) {
+            try {
+                if (null === $uuid || null === ($item = $storage->get(Uuid::fromBinary($uuid)))) {
```
Toflar: What if `$uuid` is not binary?

core-bundle/src/Filesystem/FilesystemItemIterator.php
m-vo: IMHO providing 0 is a valid scenario. Think about yielding from multiple generators for instance. 

I'm fine with an exception for the `< 0` case - this probably makes for a better API. :+1: 
Is `\OutOfBoundsException` the right one, though, because the threshold is never dynamic? `\OutOfRangeException`?

core-bundle/src/Filesystem/FilesystemUtil.php
m-vo: Well, it is always like this in the DB fields :smile:. But we could parse from both, especially if we name the helper `…FromSerialized`.

core-bundle/src/Filesystem/FilesystemItemIterator.php
Toflar: Yeah you're right :)

core-bundle/src/Filesystem/FilesystemUtil.php
Toflar: Well I'm just asking because the API you build does not say anything about the format that has to be provided :)

core-bundle/src/Filesystem/FilesystemItemIterator.php
ausi: IMO `\InvalidArgumentException` if `< 0`

m-vo: see e92f28bd18052740e57deb92dee213710930d1ba

core-bundle/src/Filesystem/FilesystemUtil.php
m-vo: see fc7284031d3c9959c9cce4e3559a3427a7c2c30e

core-bundle/src/Filesystem/FilesystemUtil.php
m-vo: I went with your suggested name + also improved parsing to handle various formats in fc7284031d3c9959c9cce4e3559a3427a7c2c30e.

Toflar (dismissed):
❤️

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4718, opened 2022-05-24
Fix DBAFS when upload_path contains subfolders

fritzmg:
If your `contao.upload_path` contains additional subfolders, e.g. `media/files/foobar` the following error will occur when uploading a file in the back end for example:

```
InvalidArgumentException: "Invalid resource media/files/foobar" at vendor/contao/core-bundle/src/Resources/contao/library/Contao/Dbafs.php line 65
```

This is because the current DBAFS implementation simply explodes the path and then tries to add all path components to the DBAFS - which will be invalid for paths equal and above to `contao.upload_path`.

This PR fixes that by introducing additional checks whether the traversed path still lies somewhere under `contao.upload_path`.

@m-vo does this need additional handling in Contao 4.13 as well?

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
@@ -49,8 +49,8 @@ class Dbafs
 	{
 		self::validateUtf8Path($strResource);
 
-		$strUploadPath = Config::get('uploadPath') . '/';
-		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
+		$uploadPath = Path::normalize(System::getContainer()->getParameter('contao.upload_path'));
+		$projectDir = Path::normalize(System::getContainer()->getParameter('kernel.project_dir'));
 
 		// Remove trailing slashes (see #5707)
 		if (substr($strResource, -1) == '/')
@@ -62,7 +62,7 @@ class Dbafs
 		$strResource = str_replace(array('\\', '//'), '/', $strResource);
 
 		// The resource does not exist or lies outside the upload directory
-		if (!$strResource || !file_exists($projectDir . '/' . $strResource) || strncmp($strResource, $strUploadPath, \strlen($strUploadPath)) !== 0)
+		if (!$strResource || !file_exists($projectDir . '/' . $strResource) || !Path::isBasePath($uploadPath, $strResource))
 		{
 			throw new \InvalidArgumentException("Invalid resource $strResource");
 		}
@@ -96,6 +96,13 @@ class Dbafs
 		while (\count($arrChunks))
 		{
 			$strPath .= '/' . array_shift($arrChunks);
+
+			// Skip paths outside the upload_path
+			if ($strPath === $uploadPath || !Path::isBasePath($uploadPath, $strPath))
+			{
+				continue;
+			}
+
 			$arrPaths[] = $strPath;
 		}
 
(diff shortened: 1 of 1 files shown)


m-vo:
> @m-vo does this need additional handling in Contao 4.13 as well?

IMHO this should already work with the new implementation, because we treat `contao.upload_path` just like any other regular path. But probably worth checking in a real installation.

ThomasJanotta:
Hi, I can confirm , that there is STILL an error in 4.13.4.
`[2022-05-21T12:09:10.708740+02:00] request.INFO: Matched route "contao_backend". {"route":"contao_backend","route_parameters":{"_route":"contao_backend","_scope":"backend","_token_check":true,"_controller":"Contao\\CoreBundle\\Controller\\BackendController::mainAction"},"request_uri":"https://domain.de/contao?act=delete&amp%3Bamp%3Bref=h4ifRwmt&amp%3Bamp%3Brt=647a57ec9e.6XsJqirA7wjj41pPBY_FT0cKkdzasuyDL1YMRAkmxDs.2Shr6AelvELWizh8VbfzCzVn86yA9IHzXyQ8Im9LtFilN0zHQay4PteMDw&do=files&id=mediapool%2F63%2F638182%2Fdata%2Fimg%2Fsteine&ref=DAxBy5r-&rt=12b4eb3009b.SitZtDadRzbp0EdHFPmwSTedNCywMqWCovqhXCWK7ac.eng79hv4FHzcuCV0RMGGDUXwVlzqdMjy0oiROkPnncQGZxzZXfEQAN2_Eg","method":"GET"} []
[2022-05-21T12:09:10.865294+02:00] request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "Invalid resource mediapool/63/638182" at /.../vendor/contao/core-bundle/src/Resources/contao/library/Contao/Dbafs.php line 65 {"exception":"[object] (InvalidArgumentException(code: 0): Invalid resource mediapool/63/638182 at /.../vendor/contao/core-bundle/src/Resources/contao/library/Contao/Dbafs.php:65)"} []`

ThomasJanotta:
my upload_url was 
mediapool/63/638182

fritzmg:
@ThomasJanotta even with the changes in this PR? You would need to manually integrate them in your `Dbafs.php`.

ThomasJanotta:
In my case it did not help.

ThomasJanotta:
@fritzmg, I just did this manually and it did not work. can I send you some access data? (PM)

fritzmg:
Unfortunately these changes are not sufficient. The following error will occur when trying to create a folder in the upload root:

fritzmg:
@contao/developers I am wondering now if we should fix this at all? The current implementation of `Dbafs` basically assumes, that the `contao.upload_path` is a single folder within `kernel.project_dir` and does not contain sub folders. May be we should instead change the validation of the configuration tree to not allow slashes?

m-vo:
If paths with with slashes work with this PR + the adjusted regex in 4.13, I'd favor a fix.

In the future, when we remove the config option and only use mounting (probably Contao 6), you'll be able to use any path as well.

fritzmg:
The new version of this PR should now fix everything. The implementation now does not assume any more that the `upload_path` is a single folder. These changes also work in Contao 4.13.

ausi (approved):

fritzmg:
One thing I could still add here: `Dbafs::updateFolderHashes` currently has no check whether the given resource(s) are actually within `contao.upload_path`. But it can only ever work with resources within `contao.upload_path` (before and now).

ThomasJanotta:
Sorry for the work and thanks for improving Contao.
:-)
It now works for me as desired, too. The content links to the images are gone due to the back and forth, but no problem. New folder and upload and call up the file links works as desired. These links have been on the web for a very long time, e.g. on Wikipedia, so they should be preserved.
Many thanks to all of you.

leofeyer:
Thank you @fritzmg.

PR #4719, opened 2022-05-24
Fix the PHP 8 warning if a $_SERVER variable does not exist

qzminski:
Some extensions use `\Contao\Environment::get()` method to fetch custom `$_SERVER` variables: https://github.com/terminal42/contao-ajaxform/blob/main/AjaxForm.php#L43

This can lead to PHP 8 warnings as follows:

CleanShot 2022-05-24 at 10 59 42

I am not sure if it's a real bug in Contao core though, or if we should change it in the extension.


fritzmg (approved):

leofeyer:
Thank you @qzminski.

PR #4721, opened 2022-05-24
Remove support for deprecated value in tl_article.printable

Toflar:


fritzmg (approved):
~~I think you should also add~~

```
$this->Template->printable = false;
```

ausi:
> I think you should also add

This is already there, or at which position do you think it should be added?

https://github.com/contao/contao/blob/6552e7115f9edb0fabb03b69c5d6d2cf749ddc75/core-bundle/src/Resources/contao/modules/ModuleArticle.php#L179

ausi (approved):

fritzmg:
Yes, I realised while writing the comment - I accidentally submitted it 🙈

leofeyer:
Thank you @Toflar.

PR #4725, opened 2022-05-24
Stop using src/Resources

leofeyer:
This PR outlines the changes to stop using `src/Resources` in our bundles, using the calendar bundle as an example.

@contao/developers Are there any objections to me converting all bundles accordingly?


fritzmg:
No objections from my side 👍

Toflar:
The whole Contao 3 library is in `core-bundle/Resources/contao` so this would basically force every single PR to require lots of rebase work by the contributors.

leofeyer:
<del>Why? The class loading is done via Composer, so nobody should ever require `vendor/contao/core-bundle/src/Resources/contao` directly.</del>

Seems I misunderstood your comment. You are referring to open PRs, right?

Toflar:
I do yes. @aschempp and I are working heavily on DC stuff, @bytehead probably has quite a few changes because of security as well.

fritzmg:
May be do it just before the first RC or first official release?

Although I can think of one issue now: inclusion of the `initialize.php` for legacy entry points gets yet more complicated afterwards 🙃

ausi:
> inclusion of the `initialize.php` for legacy entry points gets yet more complicated afterwards 🙃

There is no initialize.php anymore in 5.0, see 9a4cefad2a462984917ca714219de2c488167666

> The whole Contao 3 library is in `core-bundle/Resources/contao` so this would basically force every single PR to require lots of rebase work by the contributors.

As the changes are quite minimal and are mostly renames without content changes I think rebasing/merging should be quite straight forward.
_(Feel free to redirect any PR rebase work to me, I’m happy to sort them out ☺️)_
Although reducing the number of open PRs before this change would still be a good idea ☺️

> Are there any objections to me converting all bundles accordingly?

No objections.

leofeyer:
Ok, then we‘ll do it right before the release of the RC1 version after the pending PRs have been merged. 👍

aschempp:
LGTM too 👍 
some of the translation files are still containing `src/Resources` but I guess you're aware of that 🙃

leofeyer:
See #4991.

PR #4726, opened 2022-05-24
Remove the deprecated request token

ausi:
Deprecate `REQUEST_TOKEN` in 4.9: #4727


core-bundle/src/Resources/contao/templates/backend/be_confirm.html5
```diff
@@ -37,7 +37,7 @@
       <form class="tl_form" method="post">
         <div class="tl_formbody_edit">
           <input type="hidden" name="FORM_SUBMIT" value="invalid_token_url">
-          <input type="hidden" name="REQUEST_TOKEN" value="<?= REQUEST_TOKEN ?>">
+          <input type="hidden" name="REQUEST_TOKEN" value="<?= Contao\System::getContainer()->get('contao.csrf.token_manager')->getDefaultTokenValue() ?>">
```
Toflar: I wonder if we should have a helper in `Template` for that. Or use an insert tag or so for that. Just think about all the users that need to change this in their templates. Looks weird to access the container. In Twig it would be/is gonna be a global function anyway, I guess? /cc @m-vo

fritzmg: 👍  for helper method in `Template`. Regarding the insert tag: we already had `{{request_token}}` but we deprecated it in Contao 4.13 and removed it in Contao 5.0.

fritzmg: Back then we argued, that the value should always be provided by the controller though, not by the template itself: https://github.com/contao/contao/pull/3620

ausi: Yes, it should be provided by the controller. Should I do that in this PR, or in a seperate one?

leofeyer: I‘d say in this one.

ausi: Done 37cc24572acc832749b3aba57a6ae1393cf7272b

Toflar (dismissed):
Not sure what characters the token can contain to make encoding necessary but @ausi will know :)

ausi:
> Not sure what characters the token can contain

Symfony changed the format of the token in patch releases so I thought it would be a sane precautionary measure ☺️

Toflar:
Well, CSRF is explicitly meant for CSRF so it sort of has to be URI safe. There's also a `UriSafeTokenGenerator` 😉 So I consider that possibility to be de-facto non-existent but I'm okay with it anyway :)

core-bundle/src/Resources/contao/templates/backend/be_wildcard.html5
```diff
@@ -8,7 +8,7 @@
     <span class="upper"><?= $this->wildcard ?></span>
     <?php if ($this->link): ?>
       <br>
-      <?= $this->link ?> (<a href="<?= $this->href.'&amp;rt='.REQUEST_TOKEN ?>" class="tl_gray">ID: <?= $this->id ?></a>)
+      <?= $this->link ?> (<a href="<?= $this->href.'&amp;rt='.Contao\System::getContainer()->get('contao.csrf.token_manager')->getDefaultTokenValue() ?>" class="tl_gray">ID: <?= $this->id ?></a>)
```
leofeyer: Not sure if I recall it wrong but haven‘t we talked about adding a `requestToken()` method to the `TemplateInheritance` class and/or Twig environment, so we do not have to pass `$template->requestToken` in every controller or call this method in every template?

ausi: I think we decided that it should be the responsibility of the controller, see https://github.com/contao/contao/pull/3620#issuecomment-961157715

leofeyer: Can we discuss this again? Because a helper method would be so much more convenient IMHO.

aschempp: I fully agree with @leofeyer

m-vo: In my view, only the controller can know which tokens to pass to the templates. I thought our goal was to move away from global usage (that must also be checked globally).

Toflar: I agree that this should be the ultimate goal. However, we'll only be able to change that in a major version because everyone is relying on the fact that every POST request is validated (and to be honest, I'm not even sure that's a bad thing). If we want to change, we can still change the implementation in the controllers then too. But as long as it's a global check, a global function does seem to make sense to me too.

ausi: As discussed we want to add a global variable or method/function in Twig and in the Template class of Contao.

ausi: Done

core-bundle/src/Controller/FrontendModule/TwoFactorController.php
```diff
@@ -71,6 +72,7 @@ public static function getSubscribedServices(): array
         $services['translator'] = TranslatorInterface::class;
         $services['contao.security.two_factor.trusted_device_manager'] = TrustedDeviceManager::class;
         $services['contao.security.two_factor.backup_code_manager'] = BackupCodeManager::class;
+        $services['contao.csrf.token_manager'] = ContaoCsrfTokenManager::class;
```
Toflar: Why that change?

ausi: Fixed in 0f3659a46e9f57a64132253f28ea2fd95fddab29

UPGRADE.md
```diff
@@ -56,6 +56,12 @@ if ('contao_backend' === $route) {
 }
 ```
 
+Use the `contao.csrf.token_manager` service instead of `REQUEST_TOKEN`:
```
m-vo: Should we also hint that you can use the global template variables?

m-vo: And maybe, that `$GLOBALS['TL_CONFIG']['disableRefererCheck']` + `$GLOBALS['TL_CONFIG']['requestTokenWhitelist']` was removed? :see_no_evil:

core-bundle/src/Twig/Extension/ContaoExtension.php
```diff
@@ -66,6 +67,20 @@ public function __construct(private Environment $environment, private TemplateHi
         // Mark classes as safe for HTML that already escape their output themselves
         $escaperExtension->addSafeClass(HtmlAttributes::class, ['html', 'contao_html']);
         $escaperExtension->addSafeClass(HighlightResult::class, ['html', 'contao_html']);
+
+        $this->environment->addGlobal(
+            'request_token',
+            new class($tokenManager) {
```
m-vo:
```suggestion
            new class($tokenManager) implements \Stringable {
```

UPGRADE.md
ausi: Added hint for global template variables.
Not sure about those config settings. 
@contao/developers WDYT?

core-bundle/src/Resources/contao/library/Contao/Template.php
```diff
@@ -157,6 +157,11 @@ public function __get($strKey)
 			return $this->arrData[$strKey];
 		}
 
+		if ($strKey === 'requestToken' && !\array_key_exists($strKey, $this->arrData))
```
m-vo, line 160: Could we alternatively do the following in the constructor and leave the magic get/isset function as is?
```php
$this->arrData['requestToken'] = static fn() => htmlspecialchars(System::getContainer()->get('contao.csrf.token_manager')->getDefaultTokenValue());
```

ausi: I thought of that, but because of `setData()`, `getData()` and access to `$this->arrData` I decided against it.

m-vo (approved):

m-vo (dismissed):

UPGRADE.md
Toflar: I would add them too. Especially the first one because that was used quite often (in 99% of all cases for the wrong reasons).

leofeyer (dismissed):
Great work! And quite a few `@runInSeparateProcess` less! 🚀

Toflar (dismissed):

leofeyer (approved):

UPGRADE.md
leofeyer: Added in 23b0d4834ab833c1b4fd626169e6841f58317f57.

leofeyer:
Thank you @ausi.

PR #4728, opened 2022-05-25
Fix DBAFS when upload_path contains subfolders

fritzmg:
This fixes #4718 for Contao 4.13. There is an additional change necessary in `ConfigureFilesystemPass` as the regex there will fail since `/` is used as a delimiter and `upload_path` can contain `/`.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Dbafs.php
@@ -47,8 +47,8 @@ class Dbafs
 	{
 		self::validateUtf8Path($strResource);
 
-		$strUploadPath = System::getContainer()->getParameter('contao.upload_path') . '/';
-		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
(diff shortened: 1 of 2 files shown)


ausi (approved):

Toflar (approved):

core-bundle/src/Resources/contao/library/Contao/Dbafs.php
```diff
@@ -84,8 +84,8 @@ public static function addResource($strResource, $blnUpdateFolders=true)
 		}
 
 		$arrPaths    = array();
-		$arrChunks   = explode('/', $strResource);
-		$strPath     = array_shift($arrChunks);
+		$arrChunks   = array_filter(explode('/', Path::makeRelative($strResource, $uploadPath)));
+		$strPath     = $uploadPath;
```
leofeyer, line 88: Are you sure that this is correct? Not only might `array_shift($arrChunks)` return something else, but now the first chunk is kept in `$arrChunks`, which will affect the while loop in line 94, won‘t it?

leofeyer:
Thank you @fritzmg.

PR #4729, opened 2022-05-25
Remove deprecated localconfig keys

ausi:
See #4513


core-bundle/src/Security/User/ContaoUserProvider.php
```diff
@@ -115,8 +114,8 @@ private function validateSessionLifetime(User $user): void
             return;
         }
 
-        $config = $this->framework->getAdapter(Config::class);
-        $timeout = (int) $config->get('sessionTimeout');
+        // TODO: where should we get this value from?
+        $timeout = (int) \ini_get('session.gc_maxlifetime');
 
         if ($timeout > 0 && time() - $this->session->getMetadataBag()->getLastUsed() < $timeout) {
             return;
```
ausi: /cc @bytehead

leofeyer: Symfony uses `ini_get('session.cookie_lifetime')` and also `framework.session.cookie_lifetime`.

core-bundle/src/DependencyInjection/Configuration.php
```diff
@@ -51,19 +50,6 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->append($this->addIntlNode())
                 ->variableNode('localconfig')
                     ->info('Allows to set TL_CONFIG variables, overriding settings stored in localconfig.php. Changes in the Contao back end will not have any effect.')
-                    ->validate()
-                        ->always(
-                            static function (array $options): array {
-                                foreach (array_keys($options) as $option) {
```
Toflar, line 57: Don't you think we'll deprecate more of them during 5.x?

core-bundle/src/Resources/contao/library/Contao/Config.php
```diff
@@ -404,16 +363,6 @@ public static function has($strKey)
 	 */
 	public static function get($strKey)
 	{
-		if ($newKey = self::getNewKey($strKey))
```
Toflar, line 405: Same here, I would leave the framework as we still have some `TL_CONFIG` keys.

core-bundle/src/Security/User/ContaoUserProvider.php
fritzmg: Do we even need our own timeout?

Toflar: That's not correct. You should take it from the configured session service. If I don't use the native implementation, the ini settings are completely irrelevant.

Toflar: @fritzmg yes we do and it's also always recommended to use a custom one on top of the server-side configured one. Just because a session is available 7 days doesn't mean your app wants it to be valid for that long.

Toflar: But ideally they should be identical. It's good practice to double check though :)

fritzmg: What I meant is: is it actually necessary to validate the session life time in our _user provider_. Isn't this the task of the storage? I always assumed we only did it there because of our legacy `sessionTimeout` config.

fritzmg: > Symfony uses ini_get('session.cookie_lifetime') and also framework.session.cookie_lifetime.

`framework.session.cookie_lifetime` then gets simply set via `ini_set('session.cookie_lifetime', …)` by `NativeSessionStorage`.

ausi: Done: 326b07d4396b27b8ed332c505342d5f84dc7d5e1

core-bundle/src/DependencyInjection/Configuration.php
ausi: Restored in 7f5a64fa3ad7674bbc56c76911f7162433799619

core-bundle/src/Resources/contao/library/Contao/Config.php
ausi: Restored in 7f5a64fa3ad7674bbc56c76911f7162433799619

core-bundle/src/Security/User/ContaoUserProvider.php
```diff
@@ -115,8 +115,7 @@ private function validateSessionLifetime(User $user): void
...
         }
 
-        $config = $this->framework->getAdapter(Config::class);
-        $timeout = (int) $config->get('sessionTimeout');
+        $timeout = (int) ($this->sessionStorageOptions['gc_maxlifetime'] ?? \ini_get('session.gc_maxlifetime'));
```
fritzmg: Again, I do not think we need that at all. Since we are not using `localconfig.sessionTimeout` anymore we can remove this method altogether.

- If you define `framework.session.gc_maxlifetime`, Symfony will execute `ini_set('session.gc_maxlifetime, …)`.
- If you define `framework.session.cookie_lifetime`, Symfony will execute `ini_set('session.cookie_lifetime, …)`.

Either way, the session will time out without our `ContaoUserProvider` checking this setting, when the `NativeSessionStorage` is in use.

And if you use the `RedisSessionHandler` for example, the session timeout will be defined by the `ttl` option in the service arguments.

fritzmg (changes requested):

core-bundle/src/Security/User/ContaoUserProvider.php
Toflar: I agree that this does not make sense. What would make sense is to have a custom session lifetime setting for BE and FE users (imho that's a feature that's been requested before). This one could reduce the lifetime of the server side setting. It cannot increase it for obvious reasons, though.

ausi: Setting `gc_maxlifetime` makes it possible for PHP to garbage collect the session, it doesn’t neccesarily invalidate it automatically after that time. So this code might still make some difference, but I’m happy to remove it ☺️

Toflar: I agree, that's the other reason for always checking yourself 🙈 GC doesn't have to happen immediately. I would keep it anyway.

leofeyer: I‘d vote for removing it, but I‘m fine either way.

ausi: Maybe the Symfony security experts should decide then? @bytehead @aschempp ?

aschempp: I think the point is as @Toflar said to have a custom session timeout. It's a feature that standard Symfony does not seem to have, but we added because Contao 3 had it. Not sure if it is really necessary, but I know there have been reports about it not working (= people at least try to use it). I'm fine either way, I think it can still be added, but I would not go for `gc_maxlifetime`.

ausi: Removed in 183c92265bc0dbc0f403e7664dbc28a8782a666f

aschempp (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -367,12 +367,7 @@ public function showAll()
 				$for = preg_quote($for);
 			}
 
-			$strPattern = "CAST(name AS CHAR) REGEXP ?";
-
-			if (substr(Config::get('dbCollation'), -3) == '_ci')
-			{
-				$strPattern = "LOWER(CAST(name AS CHAR)) REGEXP LOWER(?)";
-			}
+			$strPattern = "LOWER(CAST(name AS CHAR)) REGEXP LOWER(?)";
```
leofeyer, line 360: So now we always cast to lower even if the DB collation is case-sensitive?

ausi: Yes. But this was already the case in 4.13 IMO, because `dbCollation` from the localconfig was not used by Doctrine.

Toflar (dismissed):

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #4730, opened 2022-05-25
Modern fragments: image and list content elements

m-vo:
Followup to #4671 

### Overview 
This PR introduces additional template components. I also refactored the existing ones to make use of #4715 and added some descriptions. New are:

* `picture` &mdash; outputs a `<picture>` element from a `Figure`
* `figure` &mdash; outputs a `<figure>` element from a `Figure` using the `picture` component for the media part
* `list`  &mdash; outputs a `<ul>` / `<ol>` element with items + client side randomization/pagination capabilities

The list component is IMHO a very good example of something you might want to reuse in your bundle/application. For instance, it is as simple as setting `items_per_page` to get automatic pagination:
```twig
<div class="products">
    {% embed "@Contao/component/_list.html.twig" with {items: products, items_per_page: 5} %}
        {% block list_item%}<div class="product">{{ item }}</div>{% endblock %}
    {% endembed %}
</div>
```

The idea behind doing these things client side is, that you won't have to deal with caching issues (think randomization) and you do not have to create extra ajax routes or parse GET parameters + further complicate caching. This is IMHO suitable for the gallery scenario of this PR but can of course be accompanied by other strategies **if** the generation of content is expensive or there is an excessive amount of data that would need to be included in the first request.

This PR also adds 3 content elements that are making use of the components:

* `content_element/gallery` + `content_element/image` (both using the same controller)
*  `content_element/list`

Note, that there is also an unrelated change that fixes a context variable introduced in #4393 from camelCase to snake_case [8ab2dac605bee746fcc03d6b0d7c94f8baaa6a96].


core-bundle/src/Resources/contao/templates/_new/component/_list.html.twig
```diff
@@ -0,0 +1,152 @@
...
+
+    {% if randomize_order or add_pagination or limit_elements %}
+        {% block script %}
+            {% add "list_script" to body %}
+                <script type="application/javascript">
+                    {# Shuffle lists #}
+                    document.querySelectorAll('*[data-list-random]').forEach(list => {
+                        {# Shuffle order of children by appending them at a random position. #}
+                        for (let i = list.children.length; i >= 0; i--) {
+                            list.appendChild(list.children[Math.random() * i | 0]);
```
ausi: For better readability I’d prefer using `Math.floor()` ☺️

```suggestion
                            list.appendChild(list.children[Math.floor(Math.random() * i)]);
```

ausi (dismissed):
❤️

That `{% use … %}` thing is really awesome I think.

How much better these twig templates are than our current html5 templates can be seen for example comparing the picture templates:
old: https://github.com/contao/contao/blob/d46ab5132caa770b3507c7b147b7486782a055e1/core-bundle/src/Resources/contao/templates/picture/picture_default.html5
new: https://github.com/contao/contao/blob/65095d0cf82dbdd83cf120e57b78a63fa5340051/core-bundle/src/Resources/contao/templates/_new/component/_picture.html.twig

The only thing that might be to debate IMO is the client side `_list` things, I like them but maybe this needs more consensus.
But we should merge this as is I think as this can be easily changed later.

ausi (dismissed):

leofeyer (approved):
Great work! 🚀

leofeyer:
Thank you @m-vo.

PR #4733, opened 2022-05-25
Undeprecate reload…tree ajax post actions

ausi:
In 26eb1a34dcf0f342cce9fd5ae03cc985eff007d1 I accidentally deprecated the `reloadPagetree` and `reloadFiletree` ajax post actions. This was by mistake I think, I’ve probably mixed them up with the deprecated `FileSelector` and `PageSelector` widgets.

See also #3993
> …Be aware that the `fileTree` and `pageTree` are different! Even though some of their methods say they are deprecated, that's actually what we still use everywhere, and changing that would be a different PR.

Related: #4513

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Ajax.php
+++ b/core-bundle/src/Resources/contao/classes/Ajax.php
@@ -353,12 +353,20 @@ class Ajax extends Backend
 
 				// Set the new value
 				$varValue = Input::post('value', true);
-				$strKey = match ($this->strAction)
+
+				switch ($this->strAction)
 				{
-					'reloadPicker' => 'picker',
-					'reloadPagetree' => 'pageTree',
-					'reloadFiletree' => 'fileTree',
-				};
+					case 'reloadPicker':
+						$strKey = 'picker';
+						break;
+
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/Ajax.php
```diff
@@ -353,24 +353,12 @@ public function executePostActions(DataContainer $dc)
 
 				// Set the new value
 				$varValue = Input::post('value', true);
-
-				switch ($this->strAction)
+				$strKey = match ($this->strAction)
```
leofeyer: We cannot use `match()` in Contao 4.13, I‘m afraid.

ausi: 🙈

ausi: Fixed.

aschempp (approved):
jup, I also noticed that. Should be closed in the Contao 5 issue tracker as well 🙃

leofeyer:
Thank you @ausi.

PR #4734, opened 2022-05-25
Remove more legacy deprecations

ausi:


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -429,55 +429,8 @@ protected function row($strPalette=null)
...
-
-			// Deprecated since Contao 4.2, to be removed in Contao 5.0
-			if (Input::post($this->strInputName) === null && \in_array($this->strInputName, $paletteFields))
-			{
-				trigger_deprecation('contao/core-bundle', '4.2', 'Using $_POST[\'FORM_FIELDS\'] has been deprecated and will no longer work in Contao 5.0. Make sure to always submit at least an empty string in your widget.');
-			}
-
 			// Validate and save the field
-			if ($objWidget->submitInput() && (\in_array($this->strInputName, $paletteFields) || Input::get('act') == 'overrideAll'))
+			if ($objWidget->submitInput() && (Input::post($this->strInputName) !== null || Input::get('act') == 'overrideAll'))
```
ausi: This needs verification of someone familiar with the data container and `$_POST['FORM_FIELDS']` I think ☺️

leofeyer: I‘m not sure if we should remove all of the logic. I have created another PR to remove the deprecated `FORM_FIELDS` feature (see #4740), which keeps the "only submit fields that are in the palette" logic. Depending on the discussion there, we can decide how to proceed here.

ausi: Done.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4736, opened 2022-05-25
Fix CronTest::testDoesNotRunCronJobIfAlreadyRun

fritzmg:
Currently this test will always fail when run between the full hour and 59 seconds after. This fixes that by setting the "last" run date 2 hours into the future.


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4740, opened 2022-05-26
Remove the deprecated FORM_FIELDS feature

leofeyer:
We should probably merge this before we merge #4734.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -430,50 +430,37 @@ abstract class DataContainer extends Backend
 		if (Input::post('FORM_SUBMIT') == $this->strTable)
 		{
 			$suffix = $this->getFormFieldSuffix();
-			$key = (Input::get('act') == 'editAll') ? 'FORM_FIELDS_' . $suffix : 'FORM_FIELDS';
-
-			// Calculate the current palette
-			$postPaletteFields = implode(',', Input::post($key));
-			$postPaletteFields = array_unique(StringUtil::trimsplit('[,;]', $postPaletteFields));
 
 			// Compile the palette if there is none
 			if ($strPalette === null)
 			{
-				$newPaletteFields = StringUtil::trimsplit('[,;]', $this->getPalette());
+				$paletteFields = StringUtil::trimsplit('[,;]', $this->getPalette());
 			}
 			else
 			{
 				// Use the given palette ($strPalette is an array in editAll mode)
-				$newPaletteFields = \is_array($strPalette) ? $strPalette : StringUtil::trimsplit('[,;]', $strPalette);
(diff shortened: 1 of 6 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -430,52 +430,39 @@ protected function row($strPalette=null)
...
-			$paletteFields = array_intersect($postPaletteFields, $newPaletteFields);
-
-			// Deprecated since Contao 4.2, to be removed in Contao 5.0
-			if (Input::post($this->strInputName) === null && \in_array($this->strInputName, $paletteFields))
-			{
-				trigger_deprecation('contao/core-bundle', '4.2', 'Using $_POST[\'FORM_FIELDS\'] has been deprecated and will no longer work in Contao 5.0. Make sure to always submit at least an empty string in your widget.');
-			}
-
 			// Validate and save the field
 			if ($objWidget->submitInput() && (\in_array($this->strInputName, $paletteFields) || Input::get('act') == 'overrideAll'))
```
ausi: I tested the current state of this PR and it changes behavior and creates errors like `must be of type array|object, null given`. _(to reproduce, create content element and switch between html and text type)_

The `\in_array($this->strInputName, $paletteFields)` check doesn’t make any sense now with these changes, does it? `row()` will anyways be called only for fields that are in the current palette, so I’m pretty sure this check returns `true` every time.

The old deprecation noted: _“Make sure to always submit at least an empty string in your widget.”_ this to me implies that we wanted to switch to a null check here like so:

```suggestion
			if ($objWidget->submitInput() && Input::post($this->strInputName) !== null)
```

With this change everything works again in my test.

ausi (changes requested):

leofeyer:
@ausi I have merged your suggestion in c9a8e7192ae96c7a7d50966b355f2b553cd11c69 and removed the leftover `$paletteFields` variable in 7781e2ddd0cc90a73a92ff81283ccaf60e10c86b.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -426,101 +426,50 @@ protected function row($strPalette=null)
 		$objWidget->xlabel = $xlabel;
 		$objWidget->currentRecord = $this->intId;
 
-		// Validate the field
-		if (Input::post('FORM_SUBMIT') == $this->strTable)
+		// Validate and save the field
+		if ((Input::post('FORM_SUBMIT') == $this->strTable) && $objWidget->submitInput() && Input::post($this->strInputName) !== null)
```
```suggestion
		if (Input::post('FORM_SUBMIT') == $this->strTable && $objWidget->submitInput() && Input::post($this->strInputName) !== null)
```

ausi (dismissed):

ausi (approved):

PR #4741, opened 2022-05-26
Remove deprecated TL_MODE constant

ausi:


core-bundle/src/Framework/ContaoFramework.php
```diff
@@ -84,7 +80,7 @@ public function isInitialized(): bool
     /**
      * @throws \LogicException
      */
-    public function initialize(bool $isFrontend = false): void
+    public function initialize(): void
```
ausi, line 83: @aschempp you added this parameter in d99e339333fd56b844c1b3353809013c375b970a
I’m not 100% sure if I removed this correctly here. Can you verify if that makes sense for you?

aschempp: The argument was necessary for routing to work. We need to boot the framework to do routing, but only the complete routing can tell if we are in frontend scope. If we don't have `TL_MODE` anymore, I don't think we still need this.

aschempp:
I don't like removing this, it will really break most Contao 3 extensions (which we intentionally still support) 😕

ausi:
Keeping some of the constants for better BC like `TL_MODE` and `TL_ROOT` is open for debate for me.
But either way, I think we should remove all usages from the core itself.

Toflar:
> I don't like removing this, it will really break most Contao 3 extensions (which we intentionally still support) 😕

But you can simply replace the constant with a one-liner?

aschempp:
that is true. I can also easily copy a `contao-module` structure to be used in a `contao-bundle`, so we could drop `system/modules` support alltogether. I though we're also trying to keep important stuff that would break not-updated extensions 🤷

ausi:
~~Currently blocked by #4753~~ resolved: we want to remove TL_MODE

leofeyer (approved):

leofeyer:
Thank you @ausi.

Issue #4505, opened 2022-04-13
Do we need better extensibility for certain migrations?


ausi:
### Description

> We'll have the same problem in the future anyway. Especially with input/output encoding. But also recent PRs like e.g. https://github.com/contao/contao/pull/4396 provide migrations which contain hardcoded targets and we should somehow provide a common way. Not sure if always creating an abstract makes sense or if we should introduce some general concept?

Originally posted in https://github.com/contao/contao/pull/4453#discussion_r847020543 by @Toflar

ausi:
As discussed on Slack we want to try to make the `RelLightboxMigration` automatically run for all tinyMCE fields (including fields from extensions).

ausi:
Closed in favor of #4747

PR #4747, opened 2022-05-27
Apply the rel=lightbox migration to all rte fields

ausi:
As discussed in https://github.com/contao/contao/issues/4505#issuecomment-1118608425

### DIFF
--- a/core-bundle/src/Migration/Version413/RelLightboxMigration.php
+++ b/core-bundle/src/Migration/Version413/RelLightboxMigration.php
@@ -81,8 +81,8 @@ class RelLightboxMigration extends AbstractMigration
     private function getTargets(): array
     {
         $this->framework->initialize();
-        $schemaManager = $this->connection->createSchemaManager();
 
+        $schemaManager = $this->connection->createSchemaManager();
         $targets = [];
 
         foreach ($schemaManager->listTableNames() as $tableName) {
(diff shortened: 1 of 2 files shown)


core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -86,6 +80,25 @@ public function run(): MigrationResult
...
+
+        foreach ($schemaManager->listTableNames() as $tableName) {
+            try {
+                Controller::loadDataContainer($tableName);
+            } catch (\Throwable $e) {
+                continue;
+            }
+
+            foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
+                if ('tiny' === substr($fieldConfig['eval']['rte'] ?? '', 0, 4)) {
```
aschempp, line 96: should probably use `str_starts_with`?

aschempp (changes requested):

core-bundle/src/Migration/Version413/RelLightboxMigration.php
ausi: We’d have to require `symfony/polyfill-php80` then, don’t we?

leofeyer (approved):

core-bundle/src/Migration/Version413/RelLightboxMigration.php
leofeyer: Yes, we could do that. But let‘s try not to use these features until we can use them natively. The current solution is working and Rector will automatically change it to `str_starts_with` when merging upstream.

leofeyer:
Thank you @ausi.

leofeyer:
Thank you @ausi.

PR #4748, opened 2022-05-27
Add AVIF to the default image formats

### LINKED ISSUE #4421: Add AVIF to the default image formats
### Description

Since AVIF is supported natively in PHP since version 8.1 we should add the format to: https://github.com/contao/contao/blob/f84929174aedee9b67da999e5323e19641ead14f/core-bundle/src/DependencyInjection/Configuration.php#L329-L331

Related: #2283, #3542

ausi:
Fixes #4421


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4749, opened 2022-05-27
Add the unfiltered HTML element

### LINKED ISSUE #4405: Content Element HTML without sanitization
### Description

It would be great to have a second HTML element and HTML module that completely skips sanitization. This might be available only to administrators or should be disabled by default for regular backend users.

ausi:
As discussed in #4405

ToDo:
- [x] Add module too


core-bundle/src/Resources/contao/languages/en/tl_content.xlf
```diff
@@ -98,6 +98,12 @@
       <trans-unit id="tl_content.html.1">
         <source>You can modify the list of allowed HTML tags in the back end settings.</source>
       </trans-unit>
+      <trans-unit id="tl_content.unfilteredHtml.0">
+        <source>HTML code</source>
+      </trans-unit>
+      <trans-unit id="tl_content.unfilteredHtml.1">
+        <source>Take care when pasting things you don't understand. This could allow attackers to steal your identity or take control of the whole system.</source>
```
ausi, line 105: This text is copied from the Firefox dev tools, it appears there if you try to paste something into the console.

Toflar (dismissed):
Looks good to me! Would proceed to integrate this as module as well :)

m-vo:
The only difference between this and the html element is how the sanitizing works on save (DCA side), correct?

I wonder if we should just add a second controller annotation to the existing `HtmlController` and another template `content_element/unfiltered_html.html.twig`, that looks like this:

```twig
{% extends "@Contao/content_element/html.html.twig" %}
```

core-bundle/src/Resources/contao/templates/_new/frontend_module/unfiltered_html.html.twig
```diff
@@ -0,0 +1 @@
+{% extends "@Contao/content_element/unfiltered_html.html.twig" %}
```
ausi, line 1: @m-vo do we allow extending the content element template from the frontend module or would this be considered bad practice?

m-vo: It's probably a good thing to share templates, because then you only need to adjust a single one to affect both the element and the module. IMHO we could even use the element template in the module controller…

That said - and independent from this PR - I dislike that we have to duplicate things in order to also work as modules. Can we maybe get a general "Content Element Module" to work (using the dynamic parent property of `tl_content`)? :thinking:

m-vo (dismissed):

core-bundle/src/Resources/contao/dca/tl_user.php
```diff
@@ -609,6 +609,11 @@ public function addTemplateWarning()
 		{
 			Message::addInfo($GLOBALS['TL_LANG']['MSC']['userThemeImport']);
 		}
+
+		if ($objResult->showUnfilteredHtmlWarning > 0)
```
leofeyer:
```suggestion
		if ($objResult->showUnfilteredHtmlWarning)
```

We are not using `> 0` above, either. Not sure which variant is correct, though, as we do use `> 0` in the `tl_module.php` file. 🙈

core-bundle/src/Resources/config/controller.yml
```diff
@@ -59,6 +59,8 @@ services:
 
     Contao\CoreBundle\Controller\ContentElement\HtmlController: ~
 
+    Contao\CoreBundle\Controller\ContentElement\UnfilteredHtmlController: ~
```
leofeyer: This should be placed after `ContentElement\TemplateController`. I wonder why the service linter did not stumble across this. 🤔

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -293,6 +293,12 @@
       <trans-unit id="CTE.html.1">
         <source>Allows you to add custom HTML code.</source>
       </trans-unit>
+      <trans-unit id="CTE.unfiltered_html.0">
+        <source>Unfiltered HTML</source>
+      </trans-unit>
+      <trans-unit id="CTE.unfiltered_html.1">
+        <source>Allows you to add any custom HTML code.</source>
```
```suggestion
        <source>Allows you to add custom, unfiltered HTML code.</source>
```

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -1985,6 +1991,12 @@
       <trans-unit id="MSC.groupListingModule">
         <source>There is at least one user group granting permission to the front end module "Listing" and its "list_*" fields. Since they can contain arbitrary SQL queries, users of these groups implicitly have full control over the system!</source>
       </trans-unit>
+      <trans-unit id="MSC.groupUnfilteredHtml">
+        <source>There is at least one user group granting access to the unfiltered HTML element. Since the HTML code of that element does not get filtered, users of these groups implicitly have full control over the system!</source>
```
```suggestion
        <source>There is at least one user group that grants access to the unfiltered HTML element. Since the HTML code of this element is not filtered, the users of these groups implicitly have full control over the system!</source>
```

core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -1985,6 +1991,12 @@
...
+      </trans-unit>
+      <trans-unit id="MSC.userUnfilteredHtml">
+        <source>There is at least one regular user with access to the unfiltered HTML element. Since the HTML code of that element does not get filtered, these users implicitly have full control over the system!</source>
```
```suggestion
        <source>There is at least one regular user with access to the unfiltered HTML element. Since the HTML code of that element is not filtered, these users implicitly have full control over the system!</source>
```

core-bundle/src/Resources/contao/languages/en/modules.xlf
```diff
@@ -287,6 +287,12 @@
       <trans-unit id="FMD.html.1">
         <source>Allows you to add custom HTML code</source>
       </trans-unit>
+      <trans-unit id="FMD.unfiltered_html.0">
+        <source>Unfiltered HTML</source>
+      </trans-unit>
+      <trans-unit id="FMD.unfiltered_html.1">
+        <source>Allows you to add custom HTML code</source>
```
leofeyer:
```suggestion
        <source>Allows you to add custom, unfiltered HTML code</source>
```

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4750, opened 2022-05-27
Move assets/images/deferred to var/deferred-images

### LINKED ISSUE #3898: Move assets/images/deferred to var/deferred-images
### Description

Because the JSON files are not meant to be accessible via HTTP.

ausi:
Fixes #3898

### DIFF
--- a/core-bundle/src/Resources/config/services.yml
+++ b/core-bundle/src/Resources/config/services.yml
@@ -267,7 +267,7 @@ services:
     contao.image.deferred_image_storage:
         class: Contao\Image\DeferredImageStorageFilesystem
         arguments:
-            - '%contao.image.target_dir%'
+            - '%kernel.project_dir%/var/deferred-images'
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

aschempp:
shound't this be in the cache directory?

ausi:
> shound't this be in the cache directory?

I don’t think so, as clearing the cache would then remove the deferred images.

leofeyer:
Thank you @ausi.

aschempp:
But I could have a different image configuration per environment? Or disable deferred images in dev environment?

fritzmg:
> But I could have a different image configuration per environment?

Different image configurations yield different URLs, regardless of the environment.

> Or disable deferred images in dev environment?

If deferred images are disabled then the physical file already exists and thus no request is passed to the application - so this is also independent from the environment.

PR #4751, opened 2022-05-28
Remove the orderField

### LINKED ISSUE #4380: Remove or deprecate orderField for fileTree as well
### Description

As discussed when reviewing #4367 we should remove/deprecate the `orderField` from the gallery file tree too. This can be done by including both, directory and file hashes in the multiSRC field (in the correct order).

ausi:
Fixes #4380 

I’d like to remove the `orderField` completely in Contao 5.0 and backport the support for omitting the orderField to Contao 4.13: #4752 
@contao/developers WDYT?


core-bundle/src/Migration/Version500/OrderFieldMigration.php
```diff
@@ -0,0 +1,125 @@
...
+
+use Contao\CoreBundle\Migration\AbstractMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+use Contao\StringUtil;
+use Doctrine\DBAL\Connection;
+
+/**
+ * @internal
+ */
+class OrderFieldMigration extends AbstractMigration
```
m-vo, line 23: Should this be another abstract migration that people could reuse for their tables/fields?

ausi: Don’t think so, the first `Version410\OrderFieldMigration` is also not abstract.

leofeyer (dismissed):
Perfect 👍

core-bundle/src/Migration/Version500/OrderFieldMigration.php
Toflar: Imho both would make sense, no? :)

leofeyer (approved):

core-bundle/src/Migration/Version500/OrderFieldMigration.php
leofeyer: We should probably create a different PR that adds the abstract migration and adjusts the two existing ones.

leofeyer:
Thank you @ausi.

PR #4752, opened 2022-05-28
Deprecate orderField

ausi:
Backport of #4751


aschempp:
Has the support for ordering files changed in 4.13? Otherwise I don't think we can deprecate this in 4.13 but in 5.0 only?

ausi:
> Has the support for ordering files changed in 4.13?

What do you mean by that? It hasn’t changed yet, that’s what this PR is about.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4754, opened 2022-05-30
Replace the "edit header" icon with an "edit child elements" icon

### LINKED ISSUE #3388: Standardise the edit icons

leofeyer:
Fixes #3388

This makes so much more sense now!

It sure takes a little getting used to, but we shouldn‘t let that stop us. It is a massive improvement in terms of consistency and logic.


aschempp:
Do we also need to change the _Save & Edit_ button?

leofeyer:
Probably yes, but "save & edit child elements" is a bit long. Any ideas?

leofeyer:
Done.



A shorter button text would be preferable, though.

ausi:
I really think the icon should include the the pencil, so that users see that it is an edit action.



The edit all button is also a variation with the pencil, so this would be more consistent I think: https://github.com/contao/contao/blob/5.x/core-bundle/src/Resources/contao/themes/flexible/icons/all.svg

leofeyer:
Technically, it is not an edit view but a list view. Could we change the label from "edit content elements…" to "show content elements…" and leave the icon as is?

ausi:
I think from a UX perspective it is still an edit action. Like “Edit the content of this article”.

fritzmg:
But it views the content from this article. It does not directly edit the content of the article - apart from quick actions like publishing states and sorting.

ausi:
> apart from quick actions like publishing states and sorting.

… and adding/creating and deleting.

fritzmg:
> > apart from quick actions like publishing states and sorting.
> 
> … and adding/creating and deleting.

But then to be consistent we would have to add a pencil to the icons of the front end modules, page layout, responsive image sizes etc. as well. To me all these views are functionally the same: they primarily _list_ entries while allowing you to then edit or add entries, just like most other views that list entries.

ausi:
> But then to be consistent we would have to add a pencil to the icons of the front end modules, page layout, responsive image sizes etc. as well. To me all these views are functionally the same: they primarily _list_ entries while allowing you to then edit or add entries, just like most other views that list entries.

Then we’d need icons for these child elements instead of a generic one, don’t we?
Like for content elements, news articles, events, newsletters and so on...

fritzmg:
imho the one from this PR would simply be the default for the children view. Any extension can always add more child tables - and they will always also need to come up with an icon for those.

ausi (approved):

m-vo (approved):

leofeyer:
As discussed in the Contao call, we want to merge the PR as is. In a follow-up PR, we could try to use specific icons for news, events etc. and use them instead of the generic "show child elements" button.

PR #4755, opened 2022-05-30
Fix PHP 8.x undefined array key warning

bytehead:
Follow-up to #4742

### DIFF
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -314,7 +314,7 @@ class ModuleEventlist extends Events
 			$objTemplate->classList = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->classUpcoming = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->readMore = StringUtil::specialchars(sprintf($GLOBALS['TL_LANG']['MSC']['readMore'], $event['title']));
-			$objTemplate->more = $event['linkText'] ?? $GLOBALS['TL_LANG']['MSC']['more'];
+			$objTemplate->more = ($event['linkText'] ?? null) ?: $GLOBALS['TL_LANG']['MSC']['more'];
 			$objTemplate->locationLabel = $GLOBALS['TL_LANG']['MSC']['location'];
 
 			// Short view
--- a/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
+++ b/calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
@@ -314,7 +314,7 @@ class ModuleEventlist extends Events
 			$objTemplate->classList = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->classUpcoming = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->readMore = StringUtil::specialchars(sprintf($GLOBALS['TL_LANG']['MSC']['readMore'], $event['title']));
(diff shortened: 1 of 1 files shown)


calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
```diff
@@ -314,7 +314,7 @@ protected function compile()
 			$objTemplate->classList = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->classUpcoming = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->readMore = StringUtil::specialchars(sprintf($GLOBALS['TL_LANG']['MSC']['readMore'], $event['title']));
-			$objTemplate->more = $event['linkText'] ?: $GLOBALS['TL_LANG']['MSC']['more'];
+			$objTemplate->more = $event['linkText'] ?? $GLOBALS['TL_LANG']['MSC']['more'];
```
ausi:
```suggestion
			$objTemplate->more = ($event['linkText'] ?? null) ?: $GLOBALS['TL_LANG']['MSC']['more'];
```
As an empty `linkText` should also fallback to the default text I think?

bytehead: True, fixed in 959346256.

ausi (approved):

calendar-bundle/src/Resources/contao/modules/ModuleEventlist.php
```diff
@@ -314,7 +314,7 @@ protected function compile()
 			$objTemplate->classList = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->classUpcoming = $event['class'] . ' cal_' . $event['parent'];
 			$objTemplate->readMore = StringUtil::specialchars(sprintf($GLOBALS['TL_LANG']['MSC']['readMore'], $event['title']));
-			$objTemplate->more = $event['linkText'] ?: $GLOBALS['TL_LANG']['MSC']['more'];
+			$objTemplate->more = ($event['linkText'] ?? null) ?: $GLOBALS['TL_LANG']['MSC']['more'];
```
leofeyer, line 317: The `$event` array is created from the database row here:

https://github.com/contao/contao/blob/9f554d6e15e07b44ceb47466d43f82009009f1df/calendar-bundle/src/Resources/contao/classes/Events.php#L274-L275

So `$event['linkText']` should never be unset unless you have not updated the database yet. How do I reproduce the issue?

ausi: If the field was removed via DCA config this could also be not set I think.

To reproduce unset the `'linkText'` field via custom DCA config?

leofeyer: But following this logic, we would also have to add `?? ''` to the `$GLOBALS['TL_LANG']['MSC']['readMore']` in line 316 because it might have been unset. And to every `$event['…']` usage above, too, if we can no longer rely on the fields being present in the DB.

ausi: Yes. Not saying this should or shouldn’t be a supported use case. I don’t have an opinion here I think ☺️

leofeyer: @bytehead How exactly did you stumble upon this?

bytehead: @leofeyer maybe it's the missing database update. I'll check and otherwise close. 👍

bytehead: It was because of the missing database update, closing then.

PR #4757, opened 2022-05-31
Fix a potential PHP 8 warning in booknav frontend module

qzminski:
I do get the PHP 8 warning in one of the Contao instances when using the booknav frontend module, yet I am not sure what causes that and would be the exact steps to reproduce this. I think this could be prevented by switching the order of the statements on line 97.

CleanShot 2022-05-31 at 08 55 15


leofeyer (approved):

leofeyer:
Thank you @qzminski.

PR #4761, opened 2022-05-31
Use .yaml files everywhere

### LINKED ISSUE #758: adjust manager-bundle to current Symfony best practises?
Currently the manager-bundle loads the following configuration files from `config/`:

* `config/config.yml`
* `config/config_dev.yml`
* `config/config_prod.yml`
* `config/parameters.yml` (legacy)
* `config/routing.yml`

I was wondering if the manager-bundle should be adjusted to the current Symfony 4 practises:

1. The `.yaml` file extension is used by default rather than `.yml`. 
2. The `routing.yml` is now called `routes.yaml` by default. 
3. The configuration of bundles/packages is now split into individual `.yaml` files within the `config/packages/` folder.

I don't think 3.) makes sense within the Contao Managed Edition - but may be 1.) and 2.) should be supported for a more unified experience - and a unified documentation?

leofeyer:
Fixes #758

For BC, `.yml` files are still supported, but they trigger a deprecation notice. Open questions:

1. Will we still be using `parameters.yaml` files in Contao 5?
2. Do we want to rename the `contao-manager.yml` file or should we keep it for compatibility?


aschempp:
> 1. Will we still be using `parameters.yaml` files in Contao 5?

I don't think so, people should use the `.env` and `.env.local` file.

> 2. Do we want to rename the `contao-manager.yml` file or should we keep it for compatibility?

The file is read and written by the CLI API, not by the Contao Manager directly. So if we implement some sort of BC/migration, we should be free to rename it.

PR #4763, opened 2022-05-31
Correctly toggle checkbox groups with collapseUncheckedGroups

aschempp:
While trying to implement a vanilla-JS version of the checkbox group toggling, I noticed the `collapseUncheckedGroups` (https://github.com/contao/contao/pull/3674) is not working as expected.

If `collapseUncheckedGroups` was enabled, the session state of checkbox groups was always ignored. All empty groups are collapsed, and non-empty groups are open. This has two problems:
 1. it makes the session obsolete
 2. if Javascript is disabled, there is a fallback link to toggle the group through a GET request. But since the state is ignored, it is impossible to open (or close) a checkbox group without Javascript.

Be aware that **all** multiple checkbox fields in the core do use `collapseUncheckedGroups`, so this affects all our fields 😉 

/cc @SeverinGloeckle

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/CheckBox.php
+++ b/core-bundle/src/Resources/contao/widgets/CheckBox.php
@@ -162,32 +162,24 @@ class CheckBox extends Widget
 			$img = 'folPlus.svg';
 			$display = 'none';
 
-			if ($this->collapseUncheckedGroups || !isset($state[$id]) || !empty($state[$id]))
+			$blnIsOpen = ($state[$id] ?? null) || ($this->collapseUncheckedGroups && $blnFirst && empty($this->varValue));
+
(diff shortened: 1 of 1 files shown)


SeverinGloeckle:
Looks good to me. 

I assume the preference for the unchecked/checked groups logic over the session was intentional, to ensure groups with checked options are always visible on the initial page load. 

However, your adjusted logic fixes the no-js problem and lets the user purposely and permanently close these groups when needed which is a better solution.

> Be aware that all multiple checkbox fields in the core do use collapseUncheckedGroups, so this affects all our fields 😉

Only the checkbox groups for permissions in `tl_user` and `tl_user_group` use `collapseUncheckedGroups` by default.

leofeyer (approved):

PR #4764, opened 2022-05-31
Re-add keyboard events for section wizard

aschempp:
In https://github.com/contao/contao/pull/4624 we lost the ability to use keyboard navigation on the drag handle to move section wizard rows up and down (see https://github.com/contao/contao/pull/4624/files#diff-e6190ca787d580771d824a2150c0ddd64a7c24ce49477a8dbfe59f6df3ad323eL1896-L1918 for the old code).


fritzmg:
How do I test this feature?

fritzmg (approved):
Ah, I see now. You have to:

1. Click on the _Position_ drop down once for example.
2. Then press _Tab_  4 times.
3. Then you are able to move the entry up and down via the arrow keys.

Not sure how valuable this is - but doesn't hurt to keep having this of course 👍

aschempp:
1. go to edit a page layout
2. select one of the dropdowns, it should get focus
3. use the tab key to navigate the focus to the drag handle
4. use up/down arrow keys to move the item

PR #4770, opened 2022-06-01
Permission rework

Toflar:
This is a major concept overhaul of the newly introduced permissions after hours of discussion with @aschempp. It's a lot of work in different parts but it all plays together. I'm pretty confident that we got it right this time :) 

In the current implementation, there are some major issues:

* We noticed that the voter caching doesn't really work in our current design. It's designed to cache which voters are responsible based on the `$attribute` (`contao_dc....`) and the type of `$subject` (`get_class()`). We only had `DataContainerSubject` as `$subject` class and our attributes were based on the action (`create`, `delete`, `move` etc.). This means that basically in the back end the decision manager has to go through all voters because they all register to the 5-6 actions and it's always a `DataContainerSubject`. The new concept addresses that by including the table in the `$attribute` and making `$subject` the actual action. 
* The whole system is based on "ID voting". E.g. "can the user edit news archive ID 42" but this is way too shortsighted because permissions are based on other information as well. For example our pages have access permissions. So if you want to create a new page, the target `pid` is relevant because maybe you are not allowed to create a new page in this part of the tree. 

So we went back to look at the big picture and thought about how the permissions would work on API level. The example of creating a new page would be a `POST /pages` with something like `{"pid": 42, "type": "foobar"....}`. Voting on that is easy then because you can check the `pid` and see if the user is allowed to create a new entry there.
So we started to think about the typical CRUD stuff and noticed that our DC only supports those 4 operations: What's a "move" permission for example? Right, it's a `PUT /pages/42` with a different `pid` or `sorting` but it doesn't differentiate from editing other fields at all. On API level moving, cutting, copying and pasting always comes down to any of the 4 CRUD operations.

So here we go, this PR introduces the new concept with 4 actions:

* **C**reateAction - provides the data for the new entry. It's optional though for the case of "can this user even create anything at all" aka "do we even need to show this user any create button?"
* **R**eadAction - provides the data of the current entry.
* **U**pdateAction - provides the current data as well as the new data. Both is important because permissions might be "the user can move from A to B but not from B to C" or it might be "the user can increase the value of field X but not decrease it" - hence, we have to know the **from** and **to** states.
* **D**eleteAction - provides the data of the current entry.

Two things that imho need to be clarified on top of this:

1. ~~Our current DC still saves the fields one by one in `save()`. This makes voting on one `CreateAction` or `UpdateAction` at once impossible so it votes for every single field. This again might be a problem for dependent values. It's something we've been meaning to fix for a long time, maybe we can get it done in 5.0? (should be a separate PR though for reviewing purposes)~~ Implemented changes in https://github.com/contao/contao/pull/4792 🥳 
2. The format of the "thing that is voted on" is currently undefined. I mean it's basically the format of the database but not sure if it's the case everywhere and it will certainly be different for an API. Maybe someone comes up with a smart idea how to transform the data in a unified way before it's passed on to the voters?


core-bundle/src/Security/DataContainer/AbstractAction.php
```diff
@@ -0,0 +1,62 @@
...
+     *     $action = new ReadAction('<table>', $record);
+     *     $action->preloadHints = $allRecords;
+     *
+     *     $this->denyAccessUnlessGranted('contao_dc.<table>', $action);
+     * }
+     *
+     * This ensures, every record is checked individually for maximum security, but it also allows voters to optimize
+     * for performance (entirely optional though).
+     */
+    public array $preloadHints = [];
```
aschempp: Not sure I like this could be changed by a voter

core-bundle/src/Security/DataContainer/AbstractAction.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function __construct(
+        private string $dataSource,
```
aschempp: Did you intentionally not call this `$table` because it could somewhen be something else?

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -27,21 +30,18 @@ public function __construct(private Security $security)
 
     protected function supports(string $attribute, $subject)
     {
-        return $subject instanceof DataContainerSubject
-            && 'tl_news_archive' === $subject->table
-            && str_starts_with($attribute, ContaoCorePermissions::DC_ACTION_PREFIX);
+        return str_starts_with($attribute, ContaoCorePermissions::DC_PREFIX)
+            && ($subject instanceof CreateAction || $subject instanceof ReadAction || $subject instanceof UpdateAction || $subject instanceof DeleteAction)
+            && 'tl_news_archive' === $subject->getDataSource();
```
aschempp: This should implement the cacheable voter interface and check for the full attribute with table name.

aschempp (commented):
Looks pretty much what we discussed 🎉. Lots of traits though 🫣

core-bundle/src/Security/DataContainer/AbstractAction.php
Toflar: Yes, exactly. The concept is a general one. Who knows if a "table" is going to remain our data source forever.

core-bundle/src/Security/DataContainer/AbstractAction.php
Toflar: Made immutable in https://github.com/contao/contao/pull/4770/commits/eb2201ebc807914124d4ee730b042109f5d6c4ed

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
Toflar: Adjusted https://github.com/contao/contao/pull/4770/commits/af82b4305a69729970ac7a00ba226a9fc9aa9317

Toflar:
It's just two traits and they are used exactly in the way they were meant to be used :)

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -13,36 +13,48 @@
...
-            ContaoCorePermissions::DC_ACTION_DELETE => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->id)
-                && $this->security->isGranted(ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES),
-            default => false,
-        };
+        foreach ($attributes as $attribute) {
+            if (!$this->supportsAttribute($attribute)) {
+                continue;
+            }
+
+            return match (true) {
```
aschempp: This returns true or false, but the voter should return one of the constant values (I think they are numbers)? The boolean was only correct for the `AbstractVoter` I guess?

core-bundle/src/Security/DataContainer/UpdateAction.php
```diff
@@ -0,0 +1,35 @@
...
+        private array $current,
+        private ?array $new = null,
+    ) {
+        parent::__construct($dataSource);
+    }
+
+    protected function getSubjectInfo(): array
+    {
+        $subject = parent::getSubjectInfo();
+        $subject[] = 'ID: '.$this->getCurrentId();
```
aschempp, line 28: Should this also include the new ID? Or can the current and new ID not really be different anyway? 🤔

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
Toflar: Right, so tests are useful after all for this case :) Fixed https://github.com/contao/contao/pull/4770/commits/bd73c0f08af9f11ded363a1be23393211d256884.

core-bundle/src/Security/DataContainer/UpdateAction.php
Toflar: No idea, we can always add anything to that info array.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -854,7 +859,10 @@ public function copy($source=null, $destination=null)
 			throw new InternalServerErrorException('Attempt to copy the folder "' . $source . '" to "' . $strFolder . '" (circular reference).');
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_COPY, new DataContainerSubject($this->strTable, $source, array('destination' => $destination)));
+		$this->denyAccessUnlessGranted(
+			ContaoCorePermissions::DC_PREFIX . $this->strTable,
+			new CreateAction($this->strTable, array('id' => $source, 'pid' => $strFolder))
```
aschempp: shouldn't the ID be `$destination`? And should we also check for `ReadAction` on the source record?

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1664,7 +1672,7 @@ public function editAll()
 			{
 				try
 				{
-					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $id));
+					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new UpdateAction($this->strTable, array('id' => $this->intId)));
```
aschempp: should probably be `$id` not `$this->intId`

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1964,7 +1972,7 @@ public function source()
 			throw new InternalServerErrorException('File "' . $this->intId . '" does not exist.');
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $this->intId));
+		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new UpdateAction($this->strTable, array('id' => $this->intId)));
```
aschempp, line 1978: does `DC_Folder` have a `$this->strTable` at all? What if the file is excluded from DBAFS? Or should the action data source be the parent folder in this case (though I sure would keep the attribute with table name)?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1688,4 +1689,22 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
 
 		return $label;
 	}
+
+	protected function loadActiveRecord($id = null): void
```
aschempp: having this in `DataContainer` assumes they all have a database, not sure this makes sense? It also seems unused in `DC_Folder`, maybe it should be moved to `DC_Table` instead?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -717,6 +722,13 @@ public function cut($blnDoNotRedirect=false)
 			$this->set['ptable'] = $this->ptable;
 		}
 
+		$this->loadActiveRecord();
+
+		$this->denyAccessUnlessGranted(
+			ContaoCorePermissions::DC_PREFIX . $this->strTable,
+			new UpdateAction($this->strTable, $this->activeRecord->row(), array_merge($this->activeRecord->row(), $this->set))
```
aschempp: The new data should probably not contain the full record, since only the written data should be relevant to the permission check?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -805,22 +817,18 @@ public function copy($blnDoNotRedirect=false)
...
 		/** @var Session $objSession */
 		$objSession = System::getContainer()->get('session');
 
 		/** @var AttributeBagInterface $objSessionBag */
 		$objSessionBag = $objSession->getBag('contao_backend');
 
-		$objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=?")
-								 ->limit(1)
-								 ->execute($this->intId);
+		$this->loadActiveRecord();
```
aschempp: this should also check READ permission on the source record

m-vo:
> The format of the "thing that is voted on" is currently undefined

I have the feeling this should be the same intermediary data that you would have in a multi-step form: "all data entered so far". So it won't be an entity but at least something that is serializable, I guess. Probably a key/value store object? In any case, I would avoid tying it to the database and also avoid (Contao-)database-specific formats like `'1'` pseudo-bools, but not sure if that is feasible.

Also: should it be possible to vote on an incomplete set of data and do we need to differentiate?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1047,6 +1051,19 @@ protected function copyChilds($table, $insertID, $id, $parentId)
 
 					$copy[$v][$objCTable->id]['pid'] = $insertID;
 					$copy[$v][$objCTable->id]['tstamp'] = $time;
+
+					try
+					{
+						$this->denyAccessUnlessGranted(
+							ContaoCorePermissions::DC_PREFIX . $this->strTable,
+							new UpdateAction($this->strTable, $objCTable->row(), $copy[$v])
```
aschempp: shouldn't this be `$copy[$v][$objCTable->id]`?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1402,7 +1417,7 @@ public function delete($blnDoNotRedirect=false)
 			$this->redirect($this->getReferer());
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_DELETE, new DataContainerSubject($this->strTable, $this->intId));
+		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new DeleteAction($this->strTable, $this->intId));
```
aschempp: second argument to `DeleteAction` must be array not string/int

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1594,7 +1609,12 @@ public function deleteChilds($table, $id, &$delete)
 			{
 				foreach ($objDelete->fetchAllAssoc() as $row)
 				{
-					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_DELETE, new DataContainerSubject($v, $row['id']));
+					$this->loadActiveRecord($row['id']);
```
aschempp: we should rather fetch the full record when getting delete IDs instead of re-querying every row again

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1746,20 +1764,20 @@ public function edit($intId=null, $ajaxId=null)
...
 
 		// Redirect if there is no record with the given ID
-		if ($objRow->numRows < 1)
+		if (null === $this->activeRecord)
 		{
 			throw new AccessDeniedException('Cannot load record "' . $this->strTable . '.id=' . $this->intId . '".');
 		}
 
-		$this->objActiveRecord = $objRow;
+		// TODO: This should get $new but we can only do that once we have solved the save() issue
```
aschempp: there should be a secondary permission check with the new data, but pre-checking the original record here would be correct to me

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2682,7 +2690,14 @@ public function toggle($intId=null)
...
 
 		Input::setPost('FORM_SUBMIT', $this->strTable);
-		$this->varValue = $objRow->{$this->strField};
+		$this->varValue = $this->activeRecord->{$this->strField};
+
+		// TODO: Maybe we don't have to merge manually here anymore if we find a general solution for the save()
+		// problem.
+		$this->denyAccessUnlessGranted(
+			ContaoCorePermissions::DC_PREFIX . $this->strTable,
+			new UpdateAction($this->strTable, $this->activeRecord->row(), array_merge($this->activeRecord->row(), array($this->strField => $this->activeRecord->row())))
```
aschempp: third argument of `UpdateAction` should only contain the updated data to me (maybe that could even be made clearer in the action itself)

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2770,28 +2785,27 @@ public function overrideAll()
 			{
 				foreach ($ids as $id)
 				{
+					// Initialize the active record
+					$this->loadActiveRecord($id);
```
aschempp: this seems really unperformant to query each record separately

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4512,10 +4520,9 @@ protected function parentView()
 					if ($blnHasSorting)
 					{
 						$security = System::getContainer()->get('security.helper');
-						$subject = new DataContainerSubject($this->strTable);
 
 						// Create new button
-						if (!($GLOBALS['TL_DCA'][$this->strTable]['config']['closed'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['config']['notCreatable'] ?? null) && $security->isGranted(ContaoCorePermissions::DC_ACTION_CREATE, $subject))
+						if (!($GLOBALS['TL_DCA'][$this->strTable]['config']['closed'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['config']['notCreatable'] ?? null) && $security->isGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new CreateAction($this->strTable)))
```
aschempp: This `CreateAction` should at least get the PID of the parent record. I guess we don't have a way to also include the position (previous record ID)?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4210,13 +4214,12 @@ protected function parentView()
...
+<a href="' . preg_replace('/&(amp;)?table=[^& ]*/i', ($this->ptable ? '&amp;table=' . $this->ptable : ''), $this->addToUrl('act=edit' . (Input::get('nb') ? '&amp;nc=1' : ''))) . '" class="edit" title="' . StringUtil::specialchars(sprintf(\is_array($labelEditHeader) ? $labelEditHeader[1] : $labelEditHeader, $objParent->id)) . '">' . $imageEditHeader . '</a> ' . $this->generateHeaderButtons($objParent->row(), $this->ptable) : '') . (($blnHasSorting && !($GLOBALS['TL_DCA'][$this->strTable]['config']['closed'] ?? null) && !($GLOBALS['TL_DCA'][$this->strTable]['config']['notCreatable'] ?? null) && $security->isGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new CreateAction($this->strTable))) ? '
```
aschempp: `CreateAction` should include the ID of the parent record (new PID)

aschempp (changes requested):

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: It's protected so you can override it.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: We can still improve that at some point. It's not a concern in this PR. It will only make it bigger.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done in https://github.com/contao/contao/pull/4770/commits/33551d83282ddcc61e1f56633b7ecaedb1a37e45#diff-f31b2d0b0120c3a32429a57b669102e7c9ef86a5ca152b2553602e44bd08b96d.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Fixed in 3b6d2b4e48cbebb7c4ad5f4b8166df8348bfa209

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: It does. All the fields are coming from `tl_files`.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: No it should of course contain everything, that's why everythign is loaded.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I guess this comment is obsolete now.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Nope, it must contain it all.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Not sure how to achieve this, PR welcome

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Added in d0dfa3a1c4bd5f6659b4212a98d3d4ed073b4299.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: The active record variable itself has resided in `DataContainer` forever so it should remain here :)

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Did it anyway now in e95e7478fe02f10a0b59081914cdefb514259326.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Improved in e95e7478fe02f10a0b59081914cdefb514259326.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Fixed in 81aa45a7019e89b1cbc0e16fb6b3a8c44fc7b6cb.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Of course :) Fixed in 780b504b9e0a9749e1724c01a5171f73bbb1a7ab.

Toflar:
~~Not sure how to please PHPstan here @ausi or @m-vo ?~~
Other than that, this is ready to be reviewed and **tested** :)

core-bundle/src/Security/DataContainer/CreateAction.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Contao\CoreBundle\Security\DataContainer;
+
+class CreateAction extends AbstractAction
+{
+    use NewTrait;
+
+    public function __construct(
+        private string $dataSource,
```
ausi: this should not be a property but only passed to the parent class I guess?

```suggestion
        string $dataSource,
```

Toflar: Ah right :) Done

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -850,7 +856,14 @@ public function copy($source=null, $destination=null)
 			throw new InternalServerErrorException('Attempt to copy the folder "' . $source . '" to "' . $strFolder . '" (circular reference).');
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_COPY, new DataContainerSubject($this->strTable, $source, array('destination' => $destination)));
```
aschempp, line 853: On a side note, why does a non-existing folder give an AccessDenied instead of a 404 not found? @contao/developers 😆

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1404,12 +1419,14 @@ public function delete($blnDoNotRedirect=false)
 			throw new InternalServerErrorException('Table "' . $this->strTable . '" is not deletable.');
 		}
 
-		if (!$this->intId)
+		$this->loadActiveRecord();
+
+		if (null === $this->activeRecord)
 		{
 			$this->redirect($this->getReferer());
```
aschempp, line 1467: This should probably throw a 404 exception instead? Since it should never happen, the redirect is not "useful" for the back end.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1599,9 +1616,17 @@ public function deleteChilds($table, $id, &$delete)
 
 			if ($objDelete->numRows && !($GLOBALS['TL_DCA'][$v]['config']['doNotDeleteRecords'] ?? null) && \strlen($v))
 			{
-				foreach ($objDelete->fetchAllAssoc() as $row)
+				$rows = $objDelete->fetchAllAssoc();
+				$this->preloadActiveRecords(array_map(static function ($row) { return $row['id']; }, $rows));
```
aschempp: could use `array_column` but it still fetches the records twice and we don't really need the `activeRecord` here?

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: `$row[$i]['id']` would be the new parent ID

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2395,6 +2407,11 @@ public function editAll($intId=null, $ajaxId=null)
 						$blnAjax ? $strAjax .= $this->row($this->strPalette) : $box .= $this->row($this->strPalette);
 					}
 
+					$this->denyAccessUnlessGranted(
```
aschempp: Are we missing a `ReadAction` check when rendering the `editAll` form?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2689,31 +2707,17 @@ public function overrideAll()
...
-						{
-							$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $id));
-						}
-						catch (AccessDeniedException)
-						{
-							continue;
-						}
+						$this->loadActiveRecord($id);
 
-						$this->intId = $id;
```
aschempp, line 2755: I don't think we should remove this, might be used by callbacks or event the `submit` method.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3045,10 +3061,7 @@ protected function save($varValue)
 			$this->arrSubmit[$this->strField] = $varValue;
 			$this->varValue = StringUtil::deserialize($varValue);
 
-			if (\is_object($this->objActiveRecord))
```
aschempp: I don't think we can change this, `onsubmit_callback` might access the `activeRecord` and expect the new values to be here.

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
```diff
@@ -13,36 +13,50 @@
...
+                $subject instanceof CreateAction => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES),
+                $subject instanceof ReadAction,
+                $subject instanceof UpdateAction => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->getCurrentId()),
+                $subject instanceof DeleteAction => $this->security->isGranted(ContaoNewsPermissions::USER_CAN_EDIT_ARCHIVE, $subject->getCurrentId())
+                    && $this->security->isGranted(ContaoNewsPermissions::USER_CAN_DELETE_ARCHIVES),
+                default => false,
```
aschempp, line 55: Default case is not necessary since it can never happen

aschempp (commented):
Looks very good in general! Going with CRUD permissions certainly feels right.

I think a lot of permission checks should be move to be as early as possible, (almost) same place the `closed`, `notEditable` etc. checks are happening. We don't want to tell about circular references or other stuff if the user cannot access a record at all. We should also not modify the clipboard session for example.

I also still think we should only include the "POST"ed data in the `UpdateAction`, since we want to know what data we intend to write.

Lastly, all the changes to `activeRecord` feel "shaky" to me. First of all, I am unable to cross-check if activeRecord was already used and might be changed now. Secondly, I'm not sure we might run into memory issues when generating a large tree and keeping everything in the cache. Also, this cache might lead to incorrect data since it does (obviously) not update if the DB changes. The DB _might_ change e.g. when a record is set to `unique` which will update a column of all records...

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: `$row[$i]['id']` would be the new parent ID

Toflar:
You seem to have a considerably different view on how it should be handled than me then because in my opinion, it's the complete opposite:

As early as possible (this is implemented now), yes but **never ever** with partial data!

news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php
Toflar: Defensive programming.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: We have to, we need the old data.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: The `loadActiveRecord()` does this.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I don't think this is needed. Your permission voter is responsible for that.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: We do because we pass it on to the voters. Everything should work the same :)

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I don't want to change behaviour in this PR.

Toflar:
> Lastly, all the changes to `activeRecord` feel "shaky" to me.

The active record concept remains untouched now but is completely deprecated.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
aschempp: see https://github.com/contao/contao/pull/4826

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: see https://github.com/contao/contao/pull/4826

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Now I understand, will do.

Toflar (commented):
😅 I guess this is ready for a review. I'm quite happy with how this is turning out:

- The activeRecord remains untouched now. It's deprecated and the usages in `DC_Table` are gone. Getting rid of all other usages will be ongoing work for 5.x.
- The back end tree and list views are a lot faster now thanks to current record caching 🚀 
- Update permissions receive partial data
- The `submit()` now checks for permission at one central place


Please do not merge any other PR related to `DC_Table` before this one. I will die if I have to rebase this another time.

core-bundle/src/EventListener/DataContainer/ContentCompositionListener.php
```diff
@@ -90,14 +89,14 @@ public function generateArticleForPage(DataContainer $dc): void
         $request = $this->requestStack->getCurrentRequest();
         $user = $this->security->getUser();
 
-        // Return if there is no active record (override all)
-        if (!$dc->activeRecord || null === $request || !$user instanceof BackendUser || !$request->hasSession()) {
+        // Return if there is no current record (override all)
+        if (null === ($currentRecord = $dc->getCurrentRecord()) || null === $request || !$user instanceof BackendUser || !$request->hasSession()) {
```
aschempp: For readability that/where the current record is fetched, because it becomes relevant several times below
```suggestion
        $currentRecord = $dc->getCurrentRecord();

        // Return if there is no current record (override all)
        if (null === $currentRecord || null === $request || !$user instanceof BackendUser || !$request->hasSession()) {
```

core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
```diff
@@ -45,10 +45,11 @@ public function __invoke(DataContainer $dc): array
             return array_merge(['' => '-'], $this->controller->getTemplateGroup($this->templatePrefix));
         }
 
-        $defaultTemplate = $this->customTemplates[$dc->activeRecord->type] ?? $this->getLegacyDefaultTemplate($dc);
+        $currentRecord = $dc->getCurrentRecord();
+        $defaultTemplate = $this->customTemplates[$currentRecord['type']] ?? $this->getLegacyDefaultTemplate($dc);
```
aschempp: do we need to check if `$currentRecord['type']` exists or are we sure this key is always defined (applies a few lines below as well)?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -271,6 +277,7 @@ public function __set($strKey, $varValue)
 		switch ($strKey)
 		{
 			case 'activeRecord':
+				trigger_deprecation('contao/core-bundle', '5.0', 'The active record has been deprecated and will be removed in Contao 6. Use $dc->getCurrentRecord() instead.');
```
aschempp: You cannot `set` the current record though, so this deprecation text seems wrong? Or maybe hint for `setCurrentRecordCache`?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1706,54 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+	/**
+	 * @param array<int|string> $ids
+	 */
+	protected function preloadCurrentRecords(array $ids, string $table): void
+	{
+		if (\count($ids) === 1)
+		{
+			$rows = array($this->Database->prepare("SELECT * FROM " . $table . " WHERE id=?")
+				->limit(1)
+				->execute($ids[0])
```
aschempp: If we can' be sure the key is numeric

```suggestion
				->execute(array_shift($ids))
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1706,54 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
 
 		return $label;
 	}
+
+	/**
+	 * @param array<int|string> $ids
+	 */
+	protected function preloadCurrentRecords(array $ids, string $table): void
```
aschempp, line 1717: Why is this in the generic `DataContainer` class, which is not necessarily related to tables? Shouldn't this `abstract` here an implemented in `DC_Table`?

aschempp, line 1717:
```suggestion
	protected function loadCurrentRecords(array $ids, string $table): void
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1706,54 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+	public function getCurrentRecord(string|int $id = null, string $table = null, bool $noCache = false): ?array
+	{
+		$id = $id ?: $this->intId;
+		$table = $table ?: $this->strTable;
+
+		$key = $table . '.' . $id;
+
+		if ($noCache || !isset($this->arrCurrentRecordCache[$key]))
+		{
+			$this->preloadCurrentRecords(array($id), $table);
```
aschempp, line 1772:
```suggestion
			$this->loadCurrentRecords(array($id), $table);
```

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -714,6 +715,11 @@ public function cut($source=null)
 		}
 		else
 		{
+			$this->denyAccessUnlessGranted(
```
aschempp: this needs to come before the `file_exists`, a user should get access denied and not an overwrite-warning without access rights.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1687,7 +1700,7 @@ public function editAll()
 						$objModel = Dbafs::addResource($id);
 					}
 
-					$this->objActiveRecord = $objModel;
+					$this->objActivUpdateActioneRecord = $objModel;
```
aschempp: what's this? 😆

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1652,9 +1649,14 @@ public function deleteChilds($table, $id, &$delete)
 
 			if ($objDelete->numRows && !($GLOBALS['TL_DCA'][$v]['config']['doNotDeleteRecords'] ?? null) && \strlen($v))
 			{
-				foreach ($objDelete->fetchAllAssoc() as $row)
+				$rows = $objDelete->fetchAllAssoc();
+				$this->preloadCurrentRecords(array_column($rows, 'id'), $table);
```
aschempp: Feels like you should use `setCurrentRecordCache` since the data is already available from the database! Or maybe do that in the loop and don't "get" the current record at all?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1672,19 +1674,17 @@ public function deleteChilds($table, $id, &$delete)
 	 */
 	public function undo()
```
aschempp, line 1684: shouldn't we check permissions here if the user is allowed to read the `tl_undo` record and write the restored record?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2648,8 +2645,6 @@ public function toggle($intId=null)
 			$this->intId = $intId;
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $this->intId));
```
aschempp, line 2650: where is the new `UpdateAction` permission check for this?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2742,31 +2734,19 @@ public function overrideAll()
...
-						{
-							$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $id));
-						}
-						catch (AccessDeniedException)
-						{
-							continue;
-						}
+						$currentRecord = $this->getCurrentRecord($id);
+
+						$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new ReadAction($this->strTable, $currentRecord));
```
aschempp: I don't think `overrideAll` needs a `ReadAction` check, because no data is read or displayed to the user. Also, if we would add one, the try-catch-continue would be missing.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2742,31 +2734,19 @@ public function overrideAll()
...
 						$this->arrSubmit = array();
 						$this->blnCreateNewVersion = false;
 
-						// Get the field values
-						$objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=?")
-												 ->limit(1)
-												 ->execute($this->intId);
-
-						// Store the active record
-						$this->objActiveRecord = $objRow;
```
aschempp, line 2767: This should be restored to the current record?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3207,6 +3197,8 @@ protected function submit()
 				$arrValues['ptable'] = $this->ptable;
 			}
 
+			$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new UpdateAction($this->strTable, $currentRecord, $this->arrSubmit));
+
 			$objUpdateStmt = $this->Database
 				->prepare("UPDATE " . $this->strTable . " %s WHERE " . implode(' AND ', $this->procedure))
 				->set($arrValues)
```
aschempp, line 3223: at the end of this method (I can't add a comment there), we should clean the currentRecordCache to make sure the new data is fetched when used by someone (e.g. in the `onsubmit_callback`)!

aschempp (changes requested):
Looks veeery good now! Using currentRecord is much cleaner now. Just a lot of nitpicking and notes about some security checks 🙃

core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
Toflar: We didn't check before either so I would keep it as is.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: Because our `DataContainer` represents all sorts of code which only works for tables. Like `$procedure` and `$values`, `$activeRecord` or methods like `addPtableTags()` and `addCtableTags()`, `generateRecordLabel()`. This is all old code, it's been mixed ever since. We can discuss where things belong once we stop fiddling around with files that have 6000 LOC.

Toflar: I don't like this suggestion. It explicitly preloads them. `loading` would imply, that if I called `$dc->getCurrentRecord()` I'd get the new one which I don't.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: No the query only contains the IDs here. The method is called recursively and I didn't want to introduce yet another break here and force all child DCs to adjust the method signature. I didn't feel like the performance gain would be really relevant for a delete operation which doesn't happen 100x per minute.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Hmm, I guess we should check for `CreateAction` here. Added.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: In the `submit()` method.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: I don't think it has anything to do with `setCurrentRecordCache` at all. Adjusted.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Right, adjusted.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Jep, adjusted.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Correct, adjusted.

core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
aschempp: yeah, previously it was an object, which does not trigger the invalid array key error, does it?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3242,6 +3236,9 @@ protected function submit()
 
 			$this->invalidateCacheTags();
 		}
+
+		// Empty cached data
+		$this->arrCurrentRecordCache = array();
```
aschempp: Are we sure to delete **all** caches? And not just the one of the record we just submitted? Also, if we only clear one, we might want to do this at the top (before the `noReload` check)

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2670,17 +2667,14 @@ public function toggle($intId=null)
...
+		$currentRecord = $this->getCurrentRecord();
 
 		// Redirect if there is no record with the given ID
-		if ($objRow->numRows < 1)
+		if (null === $currentRecord)
 		{
 			throw new AccessDeniedException('Cannot load record "' . $this->strTable . '.id=' . $this->intId . '".');
 		}
 
-		$this->objActiveRecord = $objRow;
```
aschempp, line 2682: this should be restored as well

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1709,6 +1709,8 @@ public function undo()
 				// Unset fields that no longer exist in the database
 				$row = array_intersect_key($row, $arrFields[$table]);
 
+				$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new CreateAction($table, $row));
```
aschempp: the prefix should use `$table` (the table that is restored to) not `$this->strTable`. Also, this would restore partial data if one of the records cannot be written. Maybe we should use a transaction for that, and cancel it if something fails before (re)throwing the exception?

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: this should use `$v` (the current table name) not `$table`

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1652,9 +1649,14 @@ public function deleteChilds($table, $id, &$delete)
...
-				foreach ($objDelete->fetchAllAssoc() as $row)
+				$rows = $objDelete->fetchAllAssoc();
+				$this->preloadCurrentRecords(array_column($rows, 'id'), $table);
+
+				foreach ($rows as $row)
 				{
-					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_DELETE, new DataContainerSubject($v, $row['id']));
+					$currentRecord = $this->getCurrentRecord($row['id'], $table);
+
+					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new DeleteAction($this->strTable, $currentRecord));
```
aschempp: this must use `$v` not `$this->strTable`

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1652,9 +1649,14 @@ public function deleteChilds($table, $id, &$delete)
...
 			if ($objDelete->numRows && !($GLOBALS['TL_DCA'][$v]['config']['doNotDeleteRecords'] ?? null) && \strlen($v))
 			{
-				foreach ($objDelete->fetchAllAssoc() as $row)
+				$rows = $objDelete->fetchAllAssoc();
+				$this->preloadCurrentRecords(array_column($rows, 'id'), $table);
+
+				foreach ($rows as $row)
 				{
-					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_DELETE, new DataContainerSubject($v, $row['id']));
+					$currentRecord = $this->getCurrentRecord($row['id'], $table);
```
aschempp: this must use `$v` not `$table`

core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
Toflar: https://3v4l.org/uZr0s#v8.1.7

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp, line 1717: just noticed that this will always override/reload something even if it is always in the cache. Maybe we should add a `$noCache` option to prevent unnecessary reloading?

Toflar: I've considered that but I don't think it's worth the hassle. If we have an array of 50 IDs and only one is not in cache yet, it will execute the query. The overhead of getting 50 vs 1 row should be neglectable.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Good catch! Fixed in 8b2e5ca0e0ad333faf0ab84d573fae48ebb7b505.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Fixed. I think we could but it's unrelated to this PR and could even be considered a bugfix. The same happens now if one insert statement fails.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Yep, fixed.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Fixed in 7dac6bfa7495dec3cc2e13399d47a70c21766044.
Why before? We need the cache for `getCurrentRecord()`.

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: I actually tested. Not a single hit at all in any mode.

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: yes, but having a permission not granted is very much more likely than an insert statement to fail, is it not?

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -318,6 +326,8 @@ public function __get($strKey)
 				return $this->strPalette;
 
 			case 'activeRecord':
+				trigger_deprecation('contao/core-bundle', '5.0', 'The active record has been deprecated and will be removed in Contao 6. Use $dc->getCurrentRecord() instead.');
```
```suggestion
				trigger_deprecation('contao/core-bundle', '5.0', 'The active record has been deprecated and will be removed in Contao 6. Use '.__CLASS__.'::getCurrentRecord() instead.');
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -865,22 +863,18 @@ public function copy($blnDoNotRedirect=false)
...
 		/** @var Session $objSession */
 		$objSession = System::getContainer()->get('session');
 
 		/** @var AttributeBagInterface $objSessionBag */
 		$objSessionBag = $objSession->getBag('contao_backend');
 
-		$objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=?")
-								 ->limit(1)
-								 ->execute($this->intId);
+		$currentRecord = $this->getCurrentRecord();
```
aschempp, line 871: This should also check `ReadAction` permission before creating a record with the new data

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1107,6 +1094,16 @@ protected function copyChilds($table, $insertID, $id, $parentId)
 
 					$copy[$v][$objCTable->id]['pid'] = $insertID;
 					$copy[$v][$objCTable->id]['tstamp'] = $time;
+
+					try
+					{
+						$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new UpdateAction($this->strTable, $objCTable->row(), $copy[$v][$objCTable->id]));
```
aschempp: This is not updating records but creating copies of children. I guess we should check `ReadAction` at the top of the `while` loop and `CreateAction` here

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1652,9 +1649,14 @@ public function deleteChilds($table, $id, &$delete)
...
-				foreach ($objDelete->fetchAllAssoc() as $row)
+				$rows = $objDelete->fetchAllAssoc();
+				$this->preloadCurrentRecords(array_column($rows, 'id'), $v);
+
+				foreach ($rows as $row)
 				{
-					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_DELETE, new DataContainerSubject($v, $row['id']));
+					$currentRecord = $this->getCurrentRecord($row['id'], $v);
+
+					$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $v, new DeleteAction($v, $currentRecord));
```
aschempp, line 1668: this should try-catch to continue deleting other child records. I guess we can't fix the issue that this might leave orphans – except maybe by using a transaction and not deleting anything instead? 🤔

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1806,20 +1808,17 @@ public function edit($intId=null, $ajaxId=null)
 			$this->intId = $intId;
 		}
 
-		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_EDIT, new DataContainerSubject($this->strTable, $this->intId));
-
 		// Get the current record
-		$objRow = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE id=?")
-								 ->limit(1)
-								 ->execute($this->intId);
+		$currentRecord = $this->getCurrentRecord();
```
aschempp, line 1821: This should also check for `ReadAction`

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2413,9 +2413,9 @@ public function editAll($intId=null, $ajaxId=null)
 							$this->varValue = \is_array($GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['default']) ? serialize($GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['default']) : $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['default'];
 						}
 
-						if ($objRow->$v !== false)
+						if ($currentRecord[$v] !== false)
 						{
-							$this->varValue = $objRow->$v;
+							$this->varValue = $currentRecord[$v];
```
aschempp: calling `submit()` in `editAll` should probably try-catch any unallowed records? (can't add that note below on line 2452)

aschempp (commented):
I wonder if we basically need to check `ReadAction` wherever we use `getCurrentRecord` (or maybe even inside that method if the record is not cached yet?)

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Good idea, done.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Indeed, fixed in 1867143874fefd0bbdb29a3405fcefd79d201534.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Honestly, there was no issue with that in 15 years. I don't want to open pandora's box here.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Done.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Right, fixed in f64cfd7928d9ff917d7ac61f715c0f9612ce8669.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Yeah but it also does check for some affected rows and deletes it again then magic. I really don't want to touch this here.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1708,89 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
 
 		return $label;
 	}
+
+	/**
+	 * @param array<int|string> $ids
+	 */
+	protected function preloadCurrentRecords(array $ids, string $table): void
+	{
+		foreach ($ids as $id)
```
aschempp, line 1720:
```suggestion
                // Clear current cache to make sure records are gone if they cannot be loaded from the database below
		foreach ($ids as $id)
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1708,89 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+			}
+
+			$this->setCurrentRecordCache($row['id'], $table, $row);
+		}
+	}
+
+	/**
+	 * @param array|null $row Pass null to remove a given cache entry
+	 */
+	protected function setCurrentRecordCache(string|int $id, string $table, ?array $row): void
```
```suggestion
	protected function setCurrentRecordCache(string|int $id, string $table, array|null $row): void
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1697,4 +1708,89 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+
+		$key = $table . '.' . $id;
+
+		if ($noCache || !isset($this->arrCurrentRecordCache[$key]))
+		{
+			$this->preloadCurrentRecords(array($id), $table);
+		}
+
+		// In case this record was not part of the preloaded result, we don't need to apply any permission checks
+		if (!array_key_exists($key, $this->arrCurrentRecordCache))
```
aschempp: since we now store the exception instead of `null` we could go back to `isset` (I think is more performant?)
```suggestion
		if (!isset($key, $this->arrCurrentRecordCache))
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -777,6 +769,10 @@ public function cut($blnDoNotRedirect=false)
 			$this->set['ptable'] = $this->ptable;
 		}
 
+		$currentRecord = $this->getCurrentRecord();
```
aschempp: I would move this to line 746 (where the previous DC_ACTION_MOVE check was, to prevent unnecessary calculating of position, clearing session etc.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -922,6 +914,8 @@ public function copy($blnDoNotRedirect=false)
 			$this->set['ptable'] = $this->ptable;
 		}
 
+		$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX . $this->strTable, new CreateAction($this->strTable, $this->set));
```
aschempp: this should be done right before the insert statement, otherwise `$this->set` is not up to date and e.g. contains the previous record id

aschempp (changes requested):
last remaining nitpicks, then I think this is RTM (once you also adjust DC_Folder 😆) ❤️

core-bundle/src/Resources/contao/classes/DataContainer.php
Toflar: Differences are neglectable now that `array_key_exists` has its own OPCode (https://github.com/php/php-src/blob/PHP-8.1/Zend/zend_compile.c#L4416). But would need to use the global one anyway here so might as well use isset.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Done in 8d38e4342318e95a787edaba8a33c26f8d3ca1de.

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: Good find! Done a4a2bd9dd46fa2da54ecdaae3595af3eb976b594.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Done.

core-bundle/src/Resources/contao/drivers/DC_Folder.php
Toflar: Fixed.

Toflar (commented):
Done :)

aschempp (dismissed):
❤️ 🎉

core-bundle/src/EventListener/DataContainer/PageSearchListener.php
```diff
@@ -32,7 +32,7 @@ public function __construct(private ContaoFramework $framework, private Connecti
      */
     public function onSaveAlias(string $value, DataContainer $dc): string
     {
-        if ($value === $dc->activeRecord->alias) {
+        if ($value === $dc->getCurrentRecord()['alias']) {
```
leofeyer: `$dc->getCurrentRecord()` can return `null`. Shouldn‘t we handle this here?

Toflar: I didn't cover these cases because the current code didn't either. We can still do that later if there's any case but apparently there never has been because accessing `->alias` on `null` would also have caused an error.

core-bundle/src/EventListener/FilterPageTypeListener.php
```diff
@@ -27,21 +27,22 @@ public function __construct(private Connection $connection)
...
+        $currentRecord = $dc->getCurrentRecord();
 
-        if (!$dc->activeRecord) {
+        if (null === $currentRecord) {
             return;
         }
 
         // The first level can only have root pages (see #6360)
-        if (!$dc->activeRecord->pid) {
+        if (!($currentRecord['pid'] ?? null)) {
```
leofeyer, line 37: Why `($currentRecord['pid'] ?? null)` here but only `$currentRecord['pid']` in line 45?

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -101,7 +103,9 @@ public function generateAlias(string $value, DataContainer $dc): string
      */
     public function validateUrlPrefix(string $value, DataContainer $dc): string
     {
-        if ('root' !== $dc->activeRecord->type || $dc->activeRecord->urlPrefix === $value) {
+        $currentRecord = $dc->getCurrentRecord();
+
+        if (null === $currentRecord || 'root' !== $currentRecord['type'] || $currentRecord['urlPrefix'] === $value) {
```
leofeyer: You have added a `null` check here but not in line 84. Why?

Toflar: No specific reason, we might as well remove it but it doesn't matter at all.

core-bundle/src/EventListener/FilterPageTypeListener.php
Toflar: Because we need to cover both cases, `isset()` and `pid = 0`. And the second case doesn't need that check because this one here is an early return.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -28,7 +30,6 @@
  * @property string         $field
  * @property string         $inputName
  * @property string         $palette
- * @property object|null    $activeRecord
```
leofeyer, line 31: As long as the property is still there, we should keep the annotation.

Toflar: Then you don't see that it's deprecated. There's no way to deprecate an `@property` annotation or at least I couldn't find it. Neither could @ausi.

leofeyer (changes requested):
This PR looks very good. Great job @Toflar! I think the concept you have been working out with @aschempp is as good as it can be, so 👍 for you guys.

If we figure out how to consistently handle `$currentRecord`, the PR can be merged IMHO.

* Sometimes we use `($currentRecord['pid'] ?? null)`.
* Sometimes we just use `$currentRecord['pid']`.
* Sometimes we check `null !== $currentRecord`.
* Sometimes we just use `$currentRecord`.

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: I see.

Toflar:
As already pointed out in the other review comments, they are consistent in 99% of all cases.

> Sometimes we use ($currentRecord['pid'] ?? null).

Only when we need to check for both, `isset()` and `> 0`

> Sometimes we just use $currentRecord['pid'].

Yes, that's the regular case when we know it's there.

> Sometimes we check null !== $currentRecord.
> Sometimes we just use $currentRecord.

 No checks are applied when we had `$dc->activeRecord->foobar` previously. They did not have any `null` check either. Maybe some of them have now but both are valid. So I consider this CS at this point.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1699,4 +1710,89 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+			$this->setCurrentRecordCache($id, $table, null);
+		}
+
+		/** @var Connection $connection */
+		$connection = System::getContainer()->get('database_connection');
+
+		$stmt = $connection->executeQuery(
+			'SELECT * FROM ' . $table . ' WHERE id IN (?)',
+			array($ids),
+			array(is_numeric(array_shift($ids)) ? Connection::PARAM_INT_ARRAY : Connection::PARAM_STR_ARRAY)
```
ausi: Modifying `$ids` here seems confusing to me (and might have the potential for future bugs if `$ids` gets used below this line).
```suggestion
			array(is_numeric(array_values($ids)[0]) ? Connection::PARAM_INT_ARRAY : Connection::PARAM_STR_ARRAY)
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -260,6 +262,12 @@ abstract class DataContainer extends Backend
 	 */
 	protected $intCurrentPid;
 
+	/**
+	 * Active record cache
+	 * @var array<int|string, array>
+	 */
+	private $arrCurrentRecordCache = array();
```
```suggestion
	/**
	 * Current record cache
	 * @var array<int|string, array>
	 */
	private $arrCurrentRecordCache = array();
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1699,4 +1710,89 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+		{
+			unset($this->arrCurrentRecordCache[$table . '.' . $id]);
+
+			return;
+		}
+
+		$this->arrCurrentRecordCache[$table . '.' . $id] = $row;
+	}
+
+	public function getCurrentRecord(string|int $id = null, string $table = null, bool $noCache = false): ?array
```
ausi, line 1764:
```suggestion
	/**
	 * @throws AccessDeniedException if the current user has no read permission
	 */
	public function getCurrentRecord(string|int $id = null, string $table = null, bool $noCache = false): ?array
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4497,7 +4514,10 @@ protected function parentView()
 
 			for ($i=0, $c=\count($row); $i<$c; $i++)
 			{
-				$this->denyAccessUnlessGranted(ContaoCorePermissions::DC_ACTION_VIEW, new DataContainerSubject($this->strTable, $row[$i]['id']));
+				// Improve performance
+				$this->setCurrentRecordCache($row[$i]['id'], $this->strTable, $row[$i]);
```
ausi, line 4517: `$row[$i]` might contain an additional `foreignKey` column here because of line 4439, so it should not be used as the current record.

However, this can be fixed in Line 4439 by moving the subquery to the `ORDER BY` clause and keeping `$query` untouchted:

```php
$key = explode('.', $GLOBALS['TL_DCA'][$this->strTable]['fields'][$firstOrderBy]['foreignKey'], 2);
$orderBy[0] = "(SELECT " . Database::quoteIdentifier($key[1]) . " FROM " . $key[0] . " WHERE " . $this->strTable . "." . Database::quoteIdentifier($firstOrderBy) . "=" . $key[0] . ".id)";
```

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
ausi:
```suggestion
        if ('root' !== $currentRecord['type'] ?? null || $currentRecord['urlPrefix'] === $value) {
```

core-bundle/src/EventListener/DataContainer/PageUrlListener.php
```diff
@@ -142,7 +146,9 @@ public function validateUrlPrefix(string $value, DataContainer $dc): string
      */
     public function validateUrlSuffix(mixed $value, DataContainer $dc): mixed
     {
-        if ('root' !== $dc->activeRecord->type || $dc->activeRecord->urlSuffix === $value) {
+        $currentRecord = $dc->getCurrentRecord();
+
+        if (null === $currentRecord || 'root' !== $currentRecord['type'] || $currentRecord['urlSuffix'] === $value) {
```
ausi:
```suggestion
        if ('root' !== $currentRecord['type'] ?? null || $currentRecord['urlSuffix'] === $value) {
```

ausi (commented):
Wow! 🤩🚀

One relevant issue https://github.com/contao/contao/pull/4770#discussion_r906803716 , rest are nitpicks and CS ☺️

core-bundle/src/Resources/contao/drivers/DC_Table.php
Toflar: I noticed that too but ignored it :P Fixed according to your suggestion.

ausi (dismissed):

Toflar:
Had to revert some of your suggestions @ausi as they made the tests and phpstan fail :)

ausi (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -260,6 +262,12 @@ abstract class DataContainer extends Backend
 	 */
 	protected $intCurrentPid;
 
+	/**
+	 * Current record cache
+	 * @var array<int|string, array>
```
m-vo:
```suggestion
	 * @var array<int|string, array<string,mixed>|AccessDeniedException>
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1699,4 +1710,92 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+		foreach ($ids as $id)
+		{
+			$this->setCurrentRecordCache($id, $table, null);
+		}
+
+		/** @var Connection $connection */
+		$connection = System::getContainer()->get('database_connection');
+
+		$stmt = $connection->executeQuery(
+			'SELECT * FROM ' . $table . ' WHERE id IN (?)',
```
m-vo, line 1729:
```suggestion
			'SELECT * FROM ' . $connection->quoteIdentifier($table) . ' WHERE id IN (?)',
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1699,4 +1710,92 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+			{
+				continue;
+			}
+
+			$this->setCurrentRecordCache($row['id'], $table, $row);
+		}
+	}
+
+	/**
+	 * @param array|null $row Pass null to remove a given cache entry
```
m-vo:
```suggestion
	 * @param array<string, mixed>|null $row Pass null to remove a given cache entry
```

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1699,4 +1710,92 @@ public function generateRecordLabel(array $row, string $table = null, bool $prot
...
+
+			return;
+		}
+
+		$this->arrCurrentRecordCache[$table . '.' . $id] = $row;
+	}
+
+	/**
+	 * @throws AccessDeniedException if the current user has no read permission
+	 */
```
m-vo, line 1763:
```suggestion
	 * @return array<string, mixed>|null
	 */
```

m-vo (commented):
Nitpicks :relaxed:

core-bundle/src/Resources/contao/classes/DataContainer.php
aschempp: we don't do that **anywhere** in the data containers

aschempp (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
ausi: We currently don’t quote any table name in DC_Table, so I think this is not needed here.

m-vo: We do in L1469, L1495 and L1500 for instance. But I'm fine anyway.

m-vo (dismissed):

ausi (dismissed):

aschempp (dismissed):

ausi (dismissed):

ausi (approved):

m-vo (approved):

leofeyer:
Thank you @Toflar.

leofeyer (approved):

PR #4772, opened 2022-06-01
Remove the wrapper around custom layout sections

### LINKED ISSUE #2645: Remove duplicate div container with "custom" class from block_sections Template
**Affected version(s)**
Contao 4.9

**Description**
We always work with custom layouts with more than two custom layout sections. 
layout-be

Result: Divs with css class "custom" exists twice or more in the source code. This div is totally unnecessary. 
The problem is anyway that you can't address the divs differently, because all of them have the css class "custom".

The "custom" div already contains a div with the ID named in the layout section. This one contains a div with the css class "inside". So this is a 3-fold nested div. 

Whatever... 2-fold nested is enough indeed.

1. That wastes unnecessary bytes
2. Additionally it also often hinders the styling.
3. I can't see any benefits, because every custom div has the class "custom". Nobobody styles all custom classes exactly.

layout-fe

Can you take this out, please?

Solution: Please remove `<div class = "custom">` from the template block_sections.html5.

leofeyer:
Fixes #2645

### DIFF
--- a/core-bundle/src/Resources/contao/templates/block/block_sections.html5
+++ b/core-bundle/src/Resources/contao/templates/block/block_sections.html5
@@ -1,6 +1,4 @@
 
-<div class="custom">
-  <?php foreach ($this->matches as $section): ?>
-    <?php $this->insert($section['template'], ['id' => $section['id'], 'content' => $section['content']]); ?>
-  <?php endforeach; ?>
-</div>
+<?php foreach ($this->matches as $section): ?>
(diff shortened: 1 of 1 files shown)


aschempp:
I wonder if we need the `block_sections` template anymore?

leofeyer:
In case someone wants to add the wrapper again, I guess.

ausi (approved):

fritzmg (approved):

PR #4774, opened 2022-06-02
Fix a potential PHP 8 error in the Contao\Environment class

qzminski:
The `HTTP_USER_AGENT` header is optional and it may cause an unnecessary PHP 8 warning:

CleanShot 2022-06-02 at 12 01 32


Toflar (approved):

leofeyer:
Thank you @qzminski.

PR #4775, opened 2022-06-02
Modern fragments: link content elements

m-vo:
Followup to #4730

This PR adds two more content elements:

* `content_element/toplink`
* `content_element/hyperlink`

I also changed the the check  `{% if not searchable|default(true) %}` to `{% if defined searchable and not searchable %}` in the base template, because otherwise setting the value to `false` in the controller would not work (the default filter would always activate). 

~~**Note:** The PR is currently based on #4730; needs to be rebased once merged. The relevant commit is here: [5ef004369f08e97bcea50dc6c67643e6f463c4db]~~


core-bundle/src/Controller/ContentElement/HyperlinkController.php
```diff
@@ -0,0 +1,75 @@
...
+class HyperlinkController extends AbstractContentElementController
+{
+    public function __construct(private readonly Studio $studio, private readonly InsertTagParser $insertTagParser)
+    {
+    }
+
+    protected function getResponse(FragmentTemplate $template, ContentModel $model, Request $request): Response
+    {
+        // Link with attributes
+        $href = $this->insertTagParser->replace($model->url);
```
ausi:
```suggestion
        $href = $this->insertTagParser->replaceInline($model->url);
```

We don’t want to allow ESI tags here I think ☺️

core-bundle/src/Resources/contao/templates/_new/content_element/toplink.html.twig
```diff
@@ -0,0 +1,36 @@
...
+                    const toplink_anchor = document.getElementById('{% block toplink_anchor %}top{% endblock %}');
+
+                    document.querySelectorAll('*[data-toplink]').forEach(link => link.addEventListener('click', e => {
+                        e.preventDefault();
+                        toplink_anchor.scrollIntoView({
+                            block: '{% block toplink_scroll_align %}start{% endblock %}',
+                            behavior: '{% block toplink_scroll_behavior %}smooth{% endblock %}',
+                        });
+                    }));
+                </script>
```
ausi: Do we actually need this script?

IMO, if smooth scrolling is desired, the CSS code `body { scroll-behavior: smooth; }` will be used anyways and no javascript is needed. Just linking to `#` or `#top` is supported by every browser AFAIK.

m-vo: Hmm. Am I missing something? For me (Chrome and Firefox on Linux) a link `<a href="#">top</a>` loads the page `https://my-domain.wip/#` instead. Does this maybe have something to do with the base tag?

ausi: Yes, I forgot about the base tag 😓

m-vo: Is there a forward compatbile way so that we simply could remove the script once the base tag is gone and nothing else would change? I assume we should not set smooth scrolling and the like then, should we?

ausi:
```suggestion
                <script>
                    document.querySelectorAll('*[data-toplink]').forEach(link => {
                        link.href = location.href;
                        link.hash = '#top';
                    });
                </script>
```

I’d just fix the URL via the JavaScript then. WDYT?

core-bundle/src/Resources/contao/templates/_new/content_element/toplink.html.twig
```diff
@@ -0,0 +1,36 @@
+{% trans_default_domain "contao_default" %}
+{% extends "@Contao/content_element/_base.html.twig" %}
+
+{% block content %}
+    {% block link %}
+        {% set link_attributes = attrs(link_attributes|default)
+            .set('href', '#')
```
ausi:
```suggestion
            .set('href', '#top')
```

And use the correct href here for the base-tag-less future ☺️

core-bundle/src/Resources/contao/templates/_new/content_element/toplink.html.twig
m-vo: Great idea! :heart_eyes: 

Done in 006c4ab3d3f1e70c07d2c1bf91f8e11a3a0af798.

core-bundle/src/Resources/contao/templates/_new/content_element/toplink.html.twig
m-vo: see 006c4ab3d3f1e70c07d2c1bf91f8e11a3a0af798

ausi (dismissed):
Approved, but #4730 need to be merged first.

leofeyer:
#4730 has been merged. @m-vo Can you resolve the conflicts please?

m-vo:
> Can you resolve the conflicts please?

Done.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4776, opened 2022-06-02
Remove invalid duplicate CSV handling

aschempp:
we already handle CSV data a few lines below, but only if `$arrData['eval']['multiple']` is enabled. This looks like a duplicate to me, but I didn't manually test it.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2347,10 +2347,6 @@ class DC_Folder extends DataContainer implements \listable, \editable
 				{
 					$varValue = '';
 				}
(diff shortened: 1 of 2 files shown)


leofeyer (changes requested):
I really don‘t think we should change this in a bugfix release. It has been added in https://github.com/contao/core/commit/1c1d663227973b4bd9afbee64dc410f53f7fff68 in 2011 (see https://github.com/contao/core/issues/2890), so apparently the implementation has been working for the last 11 years. We should not change it now without being 100% sure that the change is correct.

aschempp:
pretty sure this does not work at all 🤷

leofeyer (approved):
I have tested the changes and they are correct indeed.

PR #4777, opened 2022-06-02
Fix the missing cache tag invalidation in tree view

Toflar:
Fixes https://github.com/contao/contao/issues/4699

Needs adjustments when merged upstream (the constant for mode `5`).


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1335,21 +1335,21 @@ public function invalidateCacheTags()
...
 
-		$ptable = $GLOBALS['TL_DCA'][$strTable]['config']['ptable'];
-
 		Controller::loadDataContainer($ptable);
 
 		$objPid = $this->Database->prepare('SELECT pid FROM ' . Database::quoteIdentifier($strTable) . ' WHERE id=?')
 								 ->execute($intId);
 
-		if (!$objPid->numRows)
+		if (!$objPid->numRows || $objPid->pid == '0')
```
```suggestion
		if (!$objPid->numRows || $objPid->pid == 0)
```

Toflar: It's a string coming from the DB so...it's not consistent in the core anyway :)

leofeyer: That is only true for integers stored in `char()` and `varchar()` fields, isn‘t it? `int()` fields should return an integer.

Toflar: Yeah it works in both cases anyway. As I said, it's not consistently used in the core either. You can find examples for both with `pid` :)

fritzmg: > int() fields should return an integer.

Only in PHP 8.1+, no? https://github.com/contao/contao/issues/4278#issuecomment-1067836255

ausi:
I think adding this generic ptable handling for tree structures is correct.
But I still think that we should handle `tl_page` differently here, because editing a single content element should not invalidate the cache of the whole tree structure, should it?

Toflar:
I think it should behave the same everywhere, no matter if it's `tl_page` or not. But I'm happy with deleting only parents of the same table if we don't find a use case for any other case. @aschempp is good at finding things like this? :D

aschempp:
Not sure I understand this correctly. It very much looks like that changing anything in a page tree will basically empty all pages, articles and content elements. Because it walks the page tree and also clears all ctables of all pages?

It feels like this should pretty much behave like `reviseTable` does, but maybe I'm wrong?

Toflar:
> Not sure I understand this correctly. It very much looks like that changing anything in a page tree will basically empty all pages, articles and content elements. Because it walks the page tree and also clears all ctables of all pages?

Yes it still does and it's still what we agreed on and discussed during the call.

Toflar:
So the current solution definitely kills to many entries. It would be more efficient to just clear the whole cache if we want to go down this road. Which doesn't make any sense either.

I think the only real solution here is this:

> Always only kill the tag of the entity in question plus - when in tree mode - the children of this entity. No parents, no foreign `ctables`.

This would:

- Fix https://github.com/contao/contao/issues/4699
- Only purge parts of trees
- Keep things simple to understand and small

In order to fix navigations (no matter if custom, standard or breadcrumb):

- They have to add the tags of all the pages they output

ausi:
> They have to add the tags of all the pages they output

Changing a content element on a page, should IMO not purge all cached pages that have a link to that page.

Maybe we need two different types (categories) of tags for this issue to resolve?
1. The tags we already have that get purged by traversing all parents and children 
   (`contao.db.tl_page`, `contao.db.tl_page.1`, `contao.db.tl_article.1`, `contao.db.tl_content.1`)
2. A new tag that is meant to only be purged if the entry is edited directly
   (`contao.db.entry.tl_content.1`, `contao.db.entry.tl_content`)

A navigation module would then only use tags of the second category for the pages it links to.
Or IMO better, the general `contao.db.entry.tl_page` tag that gets purged everytime any page gets changed (to greatly reduce the number of tags).

Toflar:
> Changing a content element on a page, should IMO not purge all cached pages that have a link to that page.

Which it would not with the concept outlined in https://github.com/contao/contao/pull/4777#issuecomment-1150048379 because purging `tl_content.42` would only purge this one entry (no parents, no children because it's no tree mode).

ausi:
> Which it would not with the concept outlined in [#4777 (comment)](https://github.com/contao/contao/pull/4777#issuecomment-1150048379)

We already had this concept implemented, but had to move away from it because of issues like creating new content elements, changing visibility settings and so on.

Toflar:
I don't recall any use case. Changing visibility settings would work. Creating new elements is the only case where parents would make sense. Maybe we just need an `invalidateCacheTags($withParents)` for that, so parents are only cleared when creating new elements?

ausi:
> Creating new elements is the only case where parents would make sense.

We discussed this in detail (extensively) see https://github.com/contao/contao/issues/2137#issuecomment-681944410

There is not only the create case, but publish, resort, remove, edit any property that affects a list filter (e.g. “featured”) and many many more.

Toflar:
Well, the current state is unusable. We can also just clear the whole cache when editing anything in the back end. It has almost the same effect as this PR.

aschempp:
> I don't recall any use case. Changing visibility settings would work. 

I don't think that would work. Publishing an element that was previously not published would not result in a purge of that page. Similar to https://github.com/contao/contao/pull/4407 we would need to add a tag for a content element if it's unpublished (which we don't want, otherwise we would need to generate unpublished elements!)

Toflar:
As discussed in our Contao Slack Call on June 30, 2022, we want to change the cache tag invalidation concept to "tag a lot, invalidate as little as possible".
The reason for it being that only the controller/page that outputs the content is able to know what content is used. The opposite (tag a little, invalidate a lot) basically means that whenever you edit anything in the back end, you might as well just clear the whole cache. 

That being said, we want to change the back end to invalidate as follows:

- We invalidate the current entity
- We invalidate only the direct parent of the current entity. This covers the case of a new entity or the update of the published state of an entity (e.g. a new news entry must invalidate the whole archive so that list views can pick up the new entry).

If the the current entity does not have any parent (e.g. news archive), we need to invalidate the general one.

Here are some examples:

- Editing news entry 42 (pid = 12) invalidates the following tags:
    - `contao.db.tl_news.42`
    - `contao.db.tl_news_archive.12`

- Editing news archive 12 (no pid) invalidates the following tags:
    - `contao.db.tl_news_archive.12`
    - `contao.db.tl_news_archive`

- Editing page ID 42 (pid = 12) invalidates the following tags:
    - `contao.db.tl_page.42`
    - `contao.db.tl_page.12`

- Editing page ID 12 (pid = 0 - type `root`) invalidates the following tags:
    - `contao.db.tl_page.12`
    - `contao.db.tl_page`


TODOs for this PR:

- [x] Adjust `DataContainer` to the new concept
- [x] Add a general `kernel.response` listener that tags the page trail for every page (`if request.attributes.pageModel`)
- [x] Adjust the navigation module to tag for all pages that are displayed (in order to invalidate when a page title changes)

Toflar:
PR adjusted as discussed.

leofeyer (approved):
Changes look good. 👍 I didn‘t test if the tagging works as expected, though.

leofeyer:
Thank you @Toflar.

PR #4778, opened 2022-06-02
Fix undefined array key while button generation

rabauss:
Fix button generation without warning if `href` is not defined in DCA.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -936,7 +936,7 @@ abstract class DataContainer extends Backend
 					}
 					else
 					{
-						$href = $this->addToUrl($v['href'] . '&amp;id=' . $arrRow['id'] . '&amp;popup=1');
+						$href = $this->addToUrl(($v['href'] ?? '') . '&amp;id=' . $arrRow['id'] . '&amp;popup=1');
 					}
 
 					$return .= '<a href="' . $href . '" title="' . StringUtil::specialchars($title) . '" onclick="Backend.openModalIframe({\'title\':\'' . StringUtil::specialchars(str_replace("'", "\\'", $label)) . '\',\'url\':this.href});return false"' . $attributes . '>' . Image::getHtml($v['icon'], $label) . '</a> ';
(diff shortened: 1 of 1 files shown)


leofeyer:
How is the button working if neither `route` nor `href` is defined?

rabauss:
It's an ajax button from haste which adds click attributes: https://github.com/codefog/contao-haste/blob/main/library/Haste/Dca/AjaxOperations.php#L168

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -949,7 +949,7 @@ protected function generateButtons($arrRow, $strTable, $arrRootIds=array(), $bln
 					}
 					else
 					{
-						$href = $this->addToUrl($v['href'] . '&amp;id=' . $arrRow['id'] . (Input::get('nb') ? '&amp;nc=1' : ''));
+						$href = $this->addToUrl(($v['href'] ?? '') . '&amp;id=' . $arrRow['id'] . (Input::get('nb') ? '&amp;nc=1' : ''));
```
leofeyer, line 952: Should we maybe change line 950 from `else` to `elseif (!empty($v['href']))` instead? (If so, the same applies to line 1099.)

rabauss: But `addToUrl` also adds the `REQUEST_TOKEN` - even with empty input.
This could be a BC break?

leofeyer: You are right. 👍

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1080,7 +1080,7 @@ protected function generateGlobalButtons()
 			if (\is_array($v['button_callback'] ?? null))
 			{
 				$this->import($v['button_callback'][0]);
-				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'], $label, $title, $v['class'], $attributes, $this->strTable, $this->root);
+				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'] ?? null, $label, $title, $v['class'], $attributes, $this->strTable, $this->root);
```
```suggestion
				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'] ?? '', $label, $title, $v['class'], $attributes, $this->strTable, $this->root);
```

ausi (approved):

leofeyer:
Thank you @rabauss.

PR #4779, opened 2022-06-02
New insert tags system

ausi:
_(Very early draft)_


Toflar (commented):
This looks like it's on a very good path! ❤️ 
Two things I noticed immediately:

- The caching information is missing on the insert tag. I think we should try to re-implement the `{{date::..}}` in the new way. This will make sure we have thought of everything :) Also see https://github.com/contao/contao/pull/4782.
- I think the `OutputType` responsibility is in the wrong place. It should not be part of the container but rather part of the resulting output. The same insert tag might return `html` or `text` based on the parameters it gets.

Just a wild thought:

```php
interface InsertTag
{
    public function getName(): string;
    public function getExpires(): \DatetimeInterface;
    public function getValue(): string;
    public function getOutputType(): OutputType;
}
```

Then I can parse the tag and ask for the caching information, the output type (and e.g. ignore the ones that won't work in my PDF or whatever) and the resolved value.

And I think we should make the caching information a required attribute so the developers **must** think about whether or not their values are cacheable and if so, for how long.
Might also think about some helper factory methods like `InsertTag::createWithExpire(\Datetime...)` or `InsertTag::createWithTtl(int $ttl....)`.

ausi:
> The same insert tag might return `html` or `text` based on the parameters it gets.

Do we have insert tags that behave that way?
IMO having to define the output when registering the tag could be a great benefit for future additions (like removing or skipping tags with certain output types).

> The caching information is missing on the insert tag.

Agree.

On that note: do we still want to support ESI for insert tags?
And should the insert tag itself decide if it can be rendered via ESI or not?

Toflar:
> Do we have insert tags that behave that way?

We could have, I think. Imho it would be great if the insert tag knew the desired context and could return acccordingly. For example if the context was a `pdf` a link could look differently than for `html`. Not saying that we need that but it sounds like the responsibilites would be correct then. I don't want the user to specify two different insert tags because for them it's the same thing: placing a link.

> IMO having to define the output when registering the tag could be a great benefit for future additions (like removing or skipping tags with certain output types).

I think that's the responsibility of the one calling. So `->insertTagParser->replace($string, $desiredOutputType)` (or better, some config object for more information).

> On that note: do we still want to support ESI for insert tags?

I would. But it should not be the default. Basically the same logic as for our fragments. That would be consistent imho.

Toflar:
Random thought: Your `OutputType` is actually just a content type. Might reuse mine types maybe? Using `symfony/mime` or so…just thinking out loud here :)

ausi:
> Might reuse mine types maybe?

Not sure. These types could be way more abstract in the future I think. Also there seems to be no agreed-on mime type for URLs.

Toflar:
I just wondered why we limit it to a few given types. Why is it important to know something is an URL rather than just text. Imho it's about the characters you used, not about the semantics of them, isn't it? As far as I understand, you're trying to ensure correct encoding here? I'm just trying to figure out where to place the information. Thinking about it, maybe we don't really need the content type in the result. If the caller outputs HTML, it has to provide this context so that the handler knows, whether it has to encode or not:

```php
echo '<p class="' . $parser->parse('{{tag::...}}', OutputType::htmlAttribute) . '">Foo</p>';
```

m-vo:
IMHO output types could be more granular than just what is needed for encoding. Think of places where HTML is fine, but only phrasing content for instance (like inside a rte `p` tag).

ausi:
This should be ready for review now.

An insert tag in the new system now looks like this:

```php
class MyInsertTag
{
    #[AsInsertTag('my_tag')]
    public function __invoke(ResolvedInsertTag $insertTag): InsertTagResult
    {
        return new InsertTagResult(
            'output from my tag, param1: '.insertTag->getParameters()->get(0),
            OutputType::text,
        );
    }
}
```

The `AsInsertTag` attribute has the following settings:

- `string $name` the name of the insert tag (in lowercase)
- `bool $resolveNestedTags = true` Whether nested insert tags should be replaced before passing it to the method or not. If `true` a `ResolvedInsertTag` object will be passed, otherwise a `ParsedInsertTag`.
- `bool $asFragment = false` Whether this tag should be rendered via `<esi>` or not (e.g. `{{date}}`)
- `int $priority = 0` makes it possible to overrule other insert tags with the same name
- `string|null $method` the method name if the attribute is set on the class.

An example for a block insert tag is the `IfLanguageInsertTag` https://github.com/contao/contao/pull/4779/files#diff-14d206fe0cbf77ad5e1e208baeb2d9e51625d11546cf6ba6532d53ad00aeb5d9

Insert tag flags are pretty similar, but they retrieve and return an `InsertTagResult`, example: https://github.com/contao/contao/pull/4779/files#diff-096b303acc4d79ec65bd5a10ba4f778cca016b4054d536b1bcb60cbe5ca8a0ba

This should be fully BC because the old hooks for insert tags and flags are still executed.

core-bundle/src/InsertTag/Resolver/LinkInsertTag.php
```diff
@@ -0,0 +1,143 @@
...
+        $strClass = '';
+
+        if (null === $insertTag->getParameters()->get(0)) {
+            throw new InvalidInsertTagException('Missing parameters for link insert tag.');
+        }
+
+        $urlParam = $insertTag->getParameters()->get(0);
+
+        // External links
+        if (str_starts_with($urlParam, 'http://') || str_starts_with($urlParam, 'https://')) {
```
Toflar, line 52:
```suggestion
        if (\is_string($urlParam) && (str_starts_with($urlParam, 'http://') || str_starts_with($urlParam, 'https://'))) {
```

This will otherwise fail with a type error when you use `{{link_url::42}}`.

core-bundle/src/DependencyInjection/Compiler/AddInsertTagsPass.php
```diff
@@ -0,0 +1,130 @@
...
+use Contao\CoreBundle\InsertTag\InsertTagResult;
+use Contao\CoreBundle\InsertTag\InsertTagSubscription;
+use Contao\CoreBundle\InsertTag\ParsedSequence;
+use Symfony\Component\Config\Definition\Exception\InvalidDefinitionException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class AddInsertTagsPass implements CompilerPassInterface
```
Toflar, line 24: This class has no coverage at all. Looks a bit too complex to have no tests, don't you think? 😇

core-bundle/src/InsertTag/Flag/StringUtilFlag.php
```diff
@@ -0,0 +1,134 @@
...
+            OutputType::html === $result->getOutputType() ? OutputType::html : OutputType::text,
+            $result->getExpiresAt(),
+            $result->getCacheTags(),
+        );
+    }
+
+    #[AsInsertTagFlag('utf8_romanize')]
+    public function utf8Romanize(InsertTagFlag $flag, InsertTagResult $result): InsertTagResult
+    {
+        return new InsertTagResult(
```
Toflar: Basically most of the flags will take over the cache tags, expire at and output type from the actuall `$result`. Don't you think it would be nice to have something like `InsertTagResult::fromResult(InsertTagResult $result, \Stringable|string $value)`?

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
```diff
@@ -0,0 +1,32 @@
...
+use Contao\CoreBundle\InsertTag\OutputType;
+use Contao\CoreBundle\InsertTag\ResolvedInsertTag;
+use Contao\Date;
+
+class DateInsertTag
+{
+    #[AsInsertTag('date', asFragment: true)]
+    public function __invoke(ResolvedInsertTag $insertTag): InsertTagResult
+    {
+        return new InsertTagResult(
```
Toflar: We could now remove our hardcoded hack in `InsertTagController` and move the logic here, right? 🥳 

```
        // Special handling for the very common {{date::Y}} (e.g. in the website footer) case until
        // we have a new way to register insert tags and add that caching information to the tag itself
        if ('{{date::Y}}' === $insertTag) {
            $response->setPublic();
            $response->setExpires(new \DateTimeImmutable(date('Y').'-12-31 23:59:59'));
            $response->headers->removeCacheControlDirective('no-store');
        }
```

Toflar: I guess we also need to tag it with the root page, in case you change the page date format? 🤔  (which would be an improvement to now as we don't have that now 😄 ).

core-bundle/src/InsertTag/Resolver/LegacyInsertTag.php
```diff
@@ -0,0 +1,722 @@
...
+    #[AsInsertTag('abbr')]
+    #[AsInsertTag('acronym')]
+    #[AsInsertTag('figure')]
+    #[AsInsertTag('image')]
+    #[AsInsertTag('picture')]
+    #[AsInsertTag('file')]
+    /**
+     * @todo Refactor into separate insert tags
+     */
+    public function __invoke(ResolvedInsertTag $insertTag): InsertTagResult
```
Toflar, line 81: Should we do this now? Or what about `picture` is legacy? 😕

fritzmg:
> * `int $priority = 0` makes it possible to overrule other insert tags with the same name

Just to clarify: if you override an Insert Tag with a higher priority this way, the original Insert Tag resolver will not be executed at all, correct? If you want to instead modify the Insert Tag's result (somehow) you can instead decorate the respective resolver service.

Toflar:
I have the same question as @fritzmg in https://github.com/contao/contao/pull/4779#issuecomment-1614704640. I think this needs to be addressed somehow. 

Use case: I would replace the `{{user::*}}` insert tag in case there's a user authenticated with e.g. an OAuth provider. Then I'd want the data to come from my logic. However, if it's a regular front end member, I would like to fall back to the default logic. I could of course decorate the `LegacyInsertTag` service but that has no interface and might move later on - feels kind of shaky...

core-bundle/src/InsertTag/Resolver/LegacyInsertTag.php
ausi: The amount of work (without existing tests) is huge.  I’d like to do this soon, but I thought it should not block this PR from getting merged.

core-bundle/src/InsertTag/Resolver/LinkInsertTag.php
ausi: I think I should remove the “magic” that converts the parameter in `{{link_url::42}}` to an integer, it only caused type problems so far without any use.

Done in 4c12f0a835492381ebd8a1eff08e98b9deea8ae7

core-bundle/src/InsertTag/Flag/StringUtilFlag.php
ausi: Done in 26d78de50ccfa33235d9e4d16a2682a3d0c46196 using “wither” methods.

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
ausi: Done in 8da05706dd8adcc6eaa01df9f75dd019e89c2e3d

ausi:
> Just to clarify: if you override an Insert Tag with a higher priority this way, the original Insert Tag resolver will not be executed at all, correct? If you want to instead modify the Insert Tag's result (somehow) you can instead decorate the respective resolver service.

Yes. Decoration would work, but you could also create a new service with higher priority and inject the core insert tag to be used when needed.



> I could of course decorate the `LegacyInsertTag` service but that has no interface and might move later on - feels kind of shaky...

Yes, `LegacyInsertTag` is the exception for that because it will likely change (I marked it as internal for now), but the rest is part of our BC promise so decration/injection should work fine IMO. I don’t think we need yet another concept for prepending/appending insert tag resolvers.

core-bundle/src/DependencyInjection/Compiler/AddInsertTagsPass.php
ausi: Will do.

core-bundle/src/Controller/InsertTagsController.php
```diff
@@ -25,13 +28,26 @@
...
+        private readonly InsertTagParser $insertTagParser,
+        private readonly ResponseTagger|null $responseTagger,
+    ) {
     }
 
     public function renderAction(Request $request, string $insertTag): Response
     {
-        $response = new Response($this->insertTagParser->replaceInline($insertTag));
+        if (!str_starts_with($insertTag, '{{') || !str_ends_with($insertTag, '}}')) {
+            throw new \InvalidArgumentException(sprintf('Invalid insert tag "%s"', $insertTag));
```
Toflar: That should be a `BadRequestHttpException` imho - we're in HTTP context here.

Toflar (commented):
I've checked out the branch and played with it and I absolutely love it! ❤️ 
I think this is very close to the finishing line - just a few more tests here and there and minor nitpicks. It's wonderful, let's get this in!

fritzmg:
I still feel like we should use interfaces, since the insert tags expect certain parameters (`AsBlockInsertTag` expects two specific parameters for example) and need to return a certain type. Though I get why you don't want to do that, so that you can use multiple `AsInsertTag` attributes on different methods instead.

ausi:
> I still feel like we should use interfaces, since the insert tags expect certain parameters (`AsBlockInsertTag` expects two specific parameters for example) and need to return a certain type.

The return type is already checked for in the compiler pass. We could also check the types of the parameters, but I’m not sure if this is needed as you would already get a pretty good error message if they are the wrong types.

fritzmg:
Yes, but for DX it's nicer to have an interface or abstract whose methods can automatically be implemented by your IDE.

Consider the following:

* `AsEventListener` does not need an interface, since the API is clearly defined via the event instance itself.
* `AsController` does not need an interface, as Symfony uses magic to automatically inject the correct parameters to the route's method.

`AsHook` cannot have an interface and implementing hooks has kind of bad DX as you will always need to look up the correct parameters and return types for each individual hook.

`AsInsertTag` wouldn't be as bad as `AsHook` of course.

What if, for DX sake, we simply add an `AbstractInsertTagResolver` and `AbstractBlockInsertTagResolver` which you can optionally use?

ausi:
> What if, for DX sake, we simply add an `AbstractInsertTagResolver` and `AbstractBlockInsertTagResolver` which you can optionally use?

I added the interfaces `InsertTagFlagInterface`, `InsertTagResolverNestedParsedInterface`, `InsertTagResolverNestedResolvedInterface`, `BlockInsertTagResolverNestedParsedInterface` and `BlockInsertTagResolverNestedResolvedInterface` in 711b54d285c1cc1eb5a5a001fadec11ced477a06.

core-bundle/src/Controller/InsertTagsController.php
ausi: Changed in e76e984d763e0e159e65fc6319b1ac8e4967ee09

core-bundle/src/InsertTag/Flag/PhpFunctionFlag.php
```diff
@@ -0,0 +1,60 @@
...
+#[AsInsertTagFlag('trim')]
+#[AsInsertTagFlag('rtrim')]
+#[AsInsertTagFlag('ltrim')]
+#[AsInsertTagFlag('urlencode')]
+#[AsInsertTagFlag('rawurlencode')]
+class PhpFunctionFlag implements InsertTagFlagInterface
+{
+    public function __invoke(InsertTagFlag $flag, InsertTagResult $result): InsertTagResult
+    {
+        if (
```
Toflar: Just FYI, we could use reflection an check the attributes here. I guess performance wise it would be pretty much irrelevant and it's one mistake less one could make in this class. But we can also leave as is :D

core-bundle/src/DependencyInjection/Compiler/AddInsertTagsPass.php
ausi: Done: e865771eaddbc5a1e5aaaff4da6e1a5e77a6ea01

core-bundle/src/InsertTag/Flag/PhpFunctionFlag.php
ausi: Done in 61a2e0965f8506e1087de27e936a7f6890f5180c

Toflar (dismissed):
Awesomeness! ❤️

core-bundle/src/InsertTag/InsertTagParser.php
```diff
@@ -12,65 +12,629 @@
...
+
+        if (!$hasLegacyTags) {
+            return $input;
+        }
+
+        $outputs = [];
+
+        foreach ($this->callLegacyClass($input->serialize(), $allowEsiTags) as [$type, $chunk]) {
+            $outputs[] = match ($type) {
+                ChunkedText::TYPE_TEXT => [...$this->doParse($chunk)],
```
leofeyer: Should this be `(array) $this->doParse($chunk)` or just `$this->doParse($chunk)`?

core-bundle/src/InsertTag/ResolvedInsertTag.php
```diff
@@ -0,0 +1,33 @@
...
+        parent::__construct($name, $parameters, $flags);
+    }
+
+    public function getParameters(): ResolvedParameters
+    {
+        if (!($parameters = parent::getParameters()) instanceof ResolvedParameters) {
+            throw new \TypeError(sprintf('%s(): Return value must be of type %s, got %s', __METHOD__, ResolvedParameters::class, $parameters::class));
+        }
+
+        return $parameters;
```
leofeyer, line 33: The same applies to the `ParsedInsertTag` class.

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
```diff
@@ -0,0 +1,41 @@
...
+use Contao\CoreBundle\InsertTag\OutputType;
+use Contao\CoreBundle\InsertTag\ResolvedInsertTag;
+use Contao\Date;
+
+#[AsInsertTag('date', asFragment: true)]
+class DateInsertTag implements InsertTagResolverNestedResolvedInterface
+{
+    public function __invoke(ResolvedInsertTag $insertTag): InsertTagResult
+    {
+        $format = $insertTag->getParameters()->get(0) ?? $GLOBALS['objPage']->dateFormat ?? Config::get('dateFormat');
```
leofeyer: We have used `$GLOBALS['TL_CONFIG']` instead of `Config::get()` in other new classes, so maybe we want to do the same here?

core-bundle/src/InsertTag/InsertTagParser.php
ausi: `[...$x]` is the short version of `iterator_to_array($x)`. I like the short version better but am fine either way.

core-bundle/src/InsertTag/ResolvedInsertTag.php
ausi: Done in 8f68b497f21ca11e19afdac4d1154d51074c2016.

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
ausi: Done in 8f68b497f21ca11e19afdac4d1154d51074c2016.

core-bundle/src/InsertTag/Resolver/LegacyInsertTag.php
```diff
@@ -0,0 +1,724 @@
...
+            case 'br':
+                $result = '<br>';
+                break;
+
+            // E-mail addresses
+            case 'email':
+            case 'email_open':
+            case 'email_url':
+                if (empty($insertTag->getParameters()->get(0))) {
+                    $result = '';
```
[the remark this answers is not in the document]
ausi: Done in 8f68b497f21ca11e19afdac4d1154d51074c2016.

core-bundle/src/InsertTag/InsertTagParser.php
ausi: Changed to `iterator_to_array()` in 16cf0c0812acc1769c497db9a583abb7eee9ab8b

leofeyer (approved):

leofeyer:
The amount of work that you have put into this is mind-blowing. 🤯 Thank you very much @ausi!

leofeyer:
Thank you @ausi.

core-bundle/src/InsertTag/InsertTagFlag.php
```diff
@@ -0,0 +1,25 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\InsertTag;
+
+final class InsertTagFlag
```
m-vo, line 15: Should this rather be in the `Contao\CoreBundle\InsertTag\Flag` namespace?

core-bundle/src/InsertTag/Resolver/BlockInsertTagResolverNestedParsedInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\InsertTag\Resolver;
+
+use Contao\CoreBundle\InsertTag\ParsedInsertTag;
+use Contao\CoreBundle\InsertTag\ParsedSequence;
+
+interface BlockInsertTagResolverNestedParsedInterface
```
m-vo, line 18: I'm not too sure about the interfaces regarding DX tbh. You still have to know how to configure the `resolved` parameter and which attribute (block/normal) to use accordingly. 

If we would utilize method type inspection to determine/configure the `resolved` parameter depending on the used type hint, we could at least cut the amount of interfaces in half:

```php
// will resolve
#[AsInsertTag('foo')]
public function __invoke(ResolvedInsertTag $insertTag): InsertTagResult
{
    // …
}

// will not resolve
#[AsInsertTag('bar')]
public function __invoke(ParsedInsertTag $insertTag): InsertTagResult
{
    // …
}
```

The interface names would also be a bit more readable IMHO. :slightly_smiling_face:

core-bundle/src/InsertTag/Flag/PhpFunctionFlag.php
```diff
@@ -0,0 +1,49 @@
...
+                $allowedNames[] = $attribute->newInstance()->name;
+            }
+        }
+
+        // Do not allow arbitrary PHP functions for security reasons
+        if (!\in_array($flag->getName(), $allowedNames, true)) {
+            throw new \LogicException(sprintf('Invalid flag "%s".', $flag->getName()));
+        }
+
+        return $result->withValue($flag->getName()($result->getValue()));
```
m-vo, line 47: IMHO most of these functions do not preserve the output type. E.g. `urlencode('body { color:red }')`  does not output valid CSS anymore.  `trim`, `ltrim`, `rtrim` are maybe the exception, for the others it's hard to tell IMHO. I think I would always set `OutputType::text` here as well.

On that note: Should we maybe make the output type a mandatory parameter when calling `InsertTagResult#withValue()`? This way it would be harder to miss setting the type. I think in nearly all cases the output type won't be preserved.

Thinking about this, this might be a small design issue: right now, we cannot really add more output types later on, because every implementer needs to know how to handle all of them. IMHO when registering an insert tag, you should probably need to specify the types your implementation *is* capable to handle. When overwriting an insert tag, you can then enhance this list. I think I would be fine if you could also tell the system that you can handle any type (as it virtually is right now) - but then you can never assume that the output type is kept (and will be kept for any type added in the future).

m-vo: (The last paragraph basically is true for any insert tag (flag) implementation. I won't comment it at the other places, though. :slightly_smiling_face:)

core-bundle/src/InsertTag/InsertTag.php
```diff
@@ -0,0 +1,55 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\InsertTag;
+
+abstract class InsertTag
```
m-vo, line 15: Should we make this `@internal`? The two implementers are already final and I don't think anyone is supposed to use this class, but this way we could change things up in the future without having to think about BC.

(Same reasoning applies for `InsertTagParameters`.)

core-bundle/src/InsertTag/InsertTagParser.php
```diff
@@ -12,65 +12,625 @@
...
 
-    public function replace(string $input): string
+    public function addSubscription(InsertTagSubscription $subscription): void
     {
-        return (string) $this->replaceChunked($input);
+        if (1 !== preg_match($this->allowedTagsRegex, $subscription->name)) {
+            return;
+        }
+
+        unset($this->blockSubscriptions[$subscription->name]);
```
m-vo, line 99: Any reason to silently move things from a block to non-block subscription (and vice versa) instead of throwing an exception, here?

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\InsertTag\Resolver;
```
m-vo, line 13: I'm not sure if I fully understand the namespace for the core insert tags. Is it because you are effectively "resolving" the specific insert tag you are subscribed to? Maybe a bit confusing as you can also get a `ResolvedInsertTag` instance as the first argument which sounds a bit like there isn't anything to do then. :slightly_smiling_face: 

But maybe it just needs some getting-used-to. :relaxed:

m-vo (commented):
Amazing work. Big thanks!! :heart:.
I'm really looking forward to the possibilities this is opening up in the future. :tada: 

Although this is already merged, I left some remarks/questions. :relaxed:

core-bundle/src/InsertTag/InsertTagFlag.php
ausi: I don’t think so as the `Contao\CoreBundle\InsertTag\Flag` namespace is meant for the flag services/resolvers/callbacks.

core-bundle/src/InsertTag/Resolver/BlockInsertTagResolverNestedParsedInterface.php
ausi: /cc @fritzmg 

> we could at least cut the amount of interfaces in half

The number of interfaces would not change as we would still need one variant for `ResolvedInsertTag` and one for `ParsedInsertTag`.

But I agree that the attribute should auto-detect the `resolveNestedTags` parameter.

core-bundle/src/InsertTag/Flag/PhpFunctionFlag.php
ausi: > right now, we cannot really add more output types later on, because every implementer needs to know how to handle all of them.

I think they should handle unknown types by default like `text` and if they don’t know the source type they should output `text`.

> Should we maybe make the output type a mandatory parameter when calling `InsertTagResult#withValue()`?

I’ll fix the output type for addslashes, urlencode and rawurlencode. Making the type mandatory would just result in everyone copy/pasting `->withValue($value, $result->getOutputType())` so the added security would be minimal while making the API design worse IMO.

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
ausi: The thoughts behind that were like this:

If you parse a string like `{{foo::{{bar}}::baz}}` you get back an instance of `InsertTag`. This can be of the subtype `ResolvedInsertTag` if all nested insert tags are already handled/resolved. Or it can be a `ParsedInsertTag` that gives you access to all nested insert tags too without executing/resolving them.

The `Parsed` and `Resolved` prefixes are used for other classes too that have similar semantics.

An insert tag resolver is a service that gets passed an `InsertTag` and returns an `InsertTagResult`. Other names could be `Handler`, `Replacer`,….

But I’m open to suggestions for a better naming.

core-bundle/src/InsertTag/Flag/PhpFunctionFlag.php
ausi: Fixed the addslashes, urlencode and rawurlencode in #6195

core-bundle/src/InsertTag/InsertTag.php
ausi: Done in #6195

core-bundle/src/InsertTag/InsertTagParser.php
ausi: Changed to exceptions in #6195

core-bundle/src/InsertTag/Resolver/BlockInsertTagResolverNestedParsedInterface.php
ausi: Auto-detection of `resolveNestedTags` parameter done in #6195

core-bundle/src/InsertTag/Resolver/DateInsertTag.php
m-vo: Thanks for the explainer! Thinking about this again, calling an implementer a `Resolver` makes the most sense to me 👍

Lets keep things as they are 🙂

PR #4782, opened 2022-06-03
Add a cache timeout for the `{{date::Y}}` insert tag

Toflar:
As 4.13 is our last 4.x version with LTS support and we won't have any new insert tag service, I think it's time to introduce special handling for `{{date::Y}}`. I see this being used over and over in hundreds of setups, mostly in the footer. And the pages would be so much faster coming from the cache if this simple case was cached the way it should be.

So I think it qualifies for a special handling.

### DIFF
--- a/core-bundle/src/Controller/InsertTagsController.php
+++ b/core-bundle/src/Controller/InsertTagsController.php
@@ -43,6 +43,13 @@ class InsertTagsController
             $response->headers->addCacheControlDirective('no-store');
         }
 
+        // Special handling for the very common {{date::Y}} (e.g. in the website footer) case until
+        // we have a new way to register insert tags and add that caching information to the tag itself
+        if ('{{date::Y}}' === $insertTag) {
+            $response->setPublic();
+            $response->setExpires(new \DateTimeImmutable(date('Y').'-12-31 23:59:59'));
+        }
+
         return $response;
     }
 }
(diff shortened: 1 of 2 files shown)


bytehead (approved):

fritzmg:
The `{{date::*}}` insert tag will still be rendered as ESI though: https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/library/Contao/InsertTags.php#L201-L227

I think we need to add an exception there, so that `{{date::Y}}` is not rendered as ESI. // although may be I am misunderstanding this PR?

Toflar:
It's correct that it's rendered as ESI. Its caching time should differ from the rest of the page. The problem is just that it is always uncacheable :)

fritzmg:
Yes, I did not realize it will still be rendered as ESI :). However, on that note: couldn't we make an exception and _not_ render the insert tag as ESI, if it is `{{date::Y}}`? And instead just add the expires header to the current main response. This would improve the response time further as the reverse proxy would not need to make a sub request for the ESI tag?

Toflar:
There is no subrequest once it's cached.

fritzmg (approved):

leofeyer:
Thank you @Toflar.

aschempp:
> There is no subrequest once it's cached.

But it still requires the httpcache to merge multiple responses, which would not be necessary if we simply render the year without ESI. I would very much prefer that …

Toflar:
That's something that should be fixed on the Symfony Proxy side. The parser can replace `<esi...>` with the resolved value if the cache time of the fragment exceeds the one of the main document. Then it would be solved in a general manner.

PR #4788, opened 2022-06-03
Remove the `imagemargin` field

### LINKED ISSUE #4575: Remove imagemargin and floating in Contao 5
### Description

Many years ago we had `margin before` and `margin after` fields for content elements (not exactly sure about that), which we removed because that was theme-related. I think we should finally remove `imagemargin` and `floating`, too, as this is theme-related as well and should be solved otherwise.

The same goes for the layout options (rows and columns) in the gallery element actually, but this might be another topic.

bezin:
Fixes #4575


core-bundle/src/Image/Studio/Figure.php
```diff
@@ -203,11 +202,10 @@ public function getOptions(): array
...
      *       template's context and directly access the specific data you need.
      *
-     * @param string|array|null $margin              Set margins that will compose the inline CSS for the "margin" key
-     * @param string|null       $floating            Set/determine values for the "float_class" and "addBefore" keys
-     * @param bool              $includeFullMetadata Make all metadata available in the first dimension of the returned data set (key-value pairs)
+     * @param string|null $floating            Set/determine values for the "float_class" and "addBefore" keys
+     * @param bool        $includeFullMetadata Make all metadata available in the first dimension of the returned data set (key-value pairs)
      */
-    public function getLegacyTemplateData(array|string|null $margin = null, string $floating = null, bool $includeFullMetadata = true): array
+    public function getLegacyTemplateData(string $floating = null, bool $includeFullMetadata = true): array
```
ausi: As these methods (getLegacyTemplateData and applyLegacyTemplateData) are getting deprecated in the long term anyways, I’d prefer to keep this parameter to keep the method signature the same. With a comment maybe like `Deprecated parameter, has no effect anymore`

@m-vo WDYT?

m-vo: Yes, I think so, too. :+1:

bezin: Fixed in 349741f

bezin:
We could also remove `Controller::generateMargin` now (after we deprecated it in 4.13), as this is the only usage of this method.

ausi (dismissed):

core-bundle/src/Resources/contao/elements/ContentImage.php
```diff
@@ -68,7 +68,7 @@ protected function compile()
 
 		if (null !== $figure)
 		{
-			$figure->applyLegacyTemplateData($this->Template, $this->imagemargin);
+			$figure->applyLegacyTemplateData($this->Template, null);
```
m-vo:
```suggestion
			$figure->applyLegacyTemplateData($this->Template);
```

m-vo:
> We could also remove Controller::generateMargin now (after we deprecated it in 4.13), as this is the only usage of this method.

Do you want to create a second PR for 4.13 for the deprecation? IMHO we could remove the method in this PR then altogether.

bezin:
Yes, that is what I thought of, too.

bezin:
Deprecated in #4794 and removed here.

ausi (dismissed):

core-bundle/src/Image/Studio/Figure.php
```diff
@@ -203,7 +202,7 @@ public function getOptions(): array
      *       when using Twig templates! Instead, add this object to your
      *       template's context and directly access the specific data you need.
      *
-     * @param string|array|null $margin              Set margins that will compose the inline CSS for the "margin" key
+     * @param string|array|null $margin              Set margins that will compose the inline CSS for the "margin" key. Deprecated, does not have any effect!
```
leofeyer: Why do we keep the parameter at all?

bezin: See https://github.com/contao/contao/pull/4788/files/9e54791c5f7a628cd7986e69730c99da139a0c97#r888876972
@ausi @m-vo

leofeyer (approved):

leofeyer:
Thank you @bezin.

Issue #812, opened 2019-10-02
Validate stop date > start date


frontendschlampe:
**Affected version(s)**

I think all versions will be affected.

**Description**

Actually the function _show from/show until_ is working just for the following cases:

1. _show from_ is activated
2. _show until_ is activated
3. _both options_ are activated, but _show until_ must be a higher date than _show from_

The following option is not working:
4. _both options_ are activated, but _show until_ is a smaller date than _show from_ 

**How to reproduce**

I have a text element (I think, the problem also exists for articles and pages), which is online, but I want to hide it for a day:
_show from_ is **2019-10-04 00:01**
_show until_ is **2019-10-03 00:01**

I expect, that the text is shown until 2019-10-03 00:01, hidden the whole day of 2019-10-03 and visible again until 2019-10-04 00:01.

But the result is: the text is never visible.

asaage:
Iinteresting use-case...
I guess if it will ever behave like that in the future we will need more icons https://github.com/contao/core-bundle/pull/940 😆

Mynyx:
See https://github.com/contao/core/issues/8818

leofeyer:
That is a very adventurous expectation indeed. 😄

frontendschlampe:
> That is a very adventurous expectation indeed. 😄

Really?

I have the following use case: I have a button on a webpage. The webpage describes an external service. The button is linked to the external service. But the service is not reachable for 3 days because of maintenance. So I want to hide the button for the 3 days and show a hint, that the service is not available. When the service is back again, I want to hide the hint and show the button. Show/hide the hint is no problem, but hide/show the button is the problem.

asaage:
I don't necessarily consider this as a bug - althought the start/stop-fields could theoretically check each other for plausibility in terms of the implemented logic to avoid adventurous expectations.

I probably would appreciate the possibility of start>stop as a feature.

@frontendschlampe An intermediate solution would be to duplicate the content (button) and show one *until* and the other one *from*.

frontendschlampe:
I don't know, if we can call it bug or feature! ;-)

It's not described, that _show from/show until_ is working only from _show from_ to _show until_. I think it's not relevant if it's working from _show from_ to _show until_ or _show until_ to _show from_. There're just two fields:

1. _show from_
2. _show until_

I think, there's no technical limitation which field should be filled or which date should be higher.

frontendschlampe:
[...]

PR #4792, opened 2022-06-03
Save and submit database records

aschempp:
In any current Contao version, the DataContainer renders each "row"/widget/database column separately, and if data was submitted, it also saves each "row" separately. This has always caused issues, foremost because if one field fails to be saved, all previous fields are already written to the database.

This PR introduces a new save&submit behaviour, meaning that only full records are saved to the database if everything went smoothly. The existing `DataContainer::save()` method now stores the value inside a class property `$arrSubmit` instead of writing directly to the database. Only after all rows are rendered, the new `submit()` method is called to submit. It only submits to the database if no error occurred.

### New `onbeforesubmit_callback`

The new `onbeforesubmit_callback` receives an array of all field values before they are saved in the database. Not only allows this to do a final manipulation of the data, it also allows to **validate across multiple fields**. This made it easy to implement https://github.com/contao/contao/issues/812 with just a few lines of code.

Since all rows (and therefore all widgets) are already rendered at this stage, we cannot add new errors to the existing fields. That's why we use the flashbag/`Contao\Message` class to add global errors. That is the reason there's quite some markup change, but it is basically only moving around where the HTML is generated so `Message::generate()` can be added after the data is submitted.

Because `editAll` and `overrideAll` can affect multiple records, each record must be saved separately, and generate separate error messages. However, if anything on the current page fails, nothing should be written to the database, we're now using transactions for that. But "local" error messages should help to know what should be fixed. When editing multiple content elements, this is what the validation will look like. Nothing was saved to the database in this case, the first record neither!

Bildschirmfoto 2022-06-03 um 16 35 38

PS: since all of the behaviour is actually happening in methods of `DC_Table`, this PR cannot provide a generic implementation for all DCs. We might add this to DC_Folder later on.

PPS: be aware that this does not change the behaviour that we are currently saving autosubmit data to the database (e.g. when checking a checkbox with a subpalette or toggling a dropdown). We _might_ be able to improve on that later but that's really out of the scope of this PR.

PPPS: this does not yet implement new permissions checks as intended in https://github.com/contao/contao/pull/4770, but it will make it very easy to do so (by checking `isGranted` in the `submit()` method).


Toflar:
Imho we need to add `abstract protected function submit()` to `DataContainer` (can be left empty in DC_Folder for the time being) but it should be clear that it's an API change and `save()` is not supposed to submit the data anymore but rather just store the values of one field (phpdocs must be adjusted).

m-vo:
Love this ❤️

Two things that come to my mind are:

> Since all rows (and therefore all widgets) are already rendered at this stage, we cannot add new errors to the existing fields.

Sf allows using property paths for validation errors. It would be very nice, if we could support sth like this as well. Maybe with placeholders in the rendered fields and later by refactoring the HTML in a (top down) template? Just thinking out loud.

And: should we maybe call the "submit" method `commit()`? Because this is what it is to me - committing a prepared record.

aschempp:
> Imho we need to add `abstract protected function submit()` to `DataContainer` (can be left empty in DC_Folder for the time being) but it should be clear that it's an API change and `save()` is not supposed to submit the data anymore but rather just store the values of one field (phpdocs must be adjusted).

I thought about this as well, but since the `submit` method is never used by `DataContainer`, it does not make sense to enforce one. Handling of save & submit is entirely up to the actual DC class. If another DC driver works differently (or already implements something like this), a new abstract method does not make sense.

> should we maybe call the "submit" method commit()? Because this is what it is to me - committing a prepared record.

I actually named it that (as you can see in the first commit). And had a `oncommit_callback`. But while refactoring some more I noticed the `onsubmit_callback` was implemented three times identically and belonged to that method, and since we already call stuff "submit" I think it's better to be consistent.

Regarding the template stuff, I think this is a minor problem we don't need to fix now.

Toflar:
`save()` is abstract too and its description is not correct anymore. That's why we need another abstract to ensure people adjust their own DCs in a way where `save()` does not submit anymore.

aschempp:
> `save()` is abstract too and its description is not correct anymore. That's why we need another abstract to ensure people adjust their own DCs in a way where `save()` does not submit anymore.

No, `save()` is abstract because it is called by `DataContainer::row()` to make sure data is saved. `row()` is called by `DC_Table`, which implements `save()` and needs to call `submit()` after all `row()`s. Its entirely up to the DC to know how data is stored.

Since this only affects DataContainers extending from DC_Table and only overriding _some_ methods, I have added an exception that is triggered if data is not correctly submitted. If someone implemented their own `save()` method, that still works as-is. If someone implemented `edit()` and does not `submit()`, the exception will be thrown now.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2274,15 +2246,27 @@ public function editAll($intId=null, $ajaxId=null)
 		{
 			$class = 'tl_tbox';
 
+			if (Input::post('FORM_SUBMIT') == $this->strTable)
+			{
+				$this->Database->beginTransaction();
+			}
```
ausi: I think we should wrap the code until line 2414 https://github.com/contao/contao/blob/16c02503f7e143257f7c1f380471b039b4256852/core-bundle/src/Resources/contao/drivers/DC_Table.php#L2413-L2415 with a try catch block, call `rollbackTransaction()` in the `catch(\Throwable)` and rethrow the exception.

ausi: I think we should wrap the code until line 2414 https://github.com/contao/contao/blob/16c02503f7e143257f7c1f380471b039b4256852/core-bundle/src/Resources/contao/drivers/DC_Table.php#L2413-L2415 with a try catch block, call `rollbackTransaction()` in the `catch(\Throwable)` and rethrow the exception.

aschempp: added in aa6513c5f7c17a3c3ff0c82f0be41e8ec7f60b80

leofeyer:
How about `persist()` and `onpersist_callback`?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -224,6 +230,14 @@ public function __construct($strTable, $arrModule=array())
 		}
 	}
 
+	public function __destruct()
+	{
+		if (!empty($this->arrSubmit))
+		{
+			throw new \RuntimeException(__CLASS__ . ' must call the submit() method to store data in Contao 5.');
+		}
+	}
```
ausi: I’m not a big fan of exceptions in destructors, because when this acutally gets executed depends on who holds a reference to `$dc` 😕

How about using something like this 
```php
try {
    return $dc->$act();
} finally {
    $dc->validateSubmit();
}
```
here https://github.com/contao/contao/blob/7b8f4ddcdc957db3f2efcf9c797fd0cbca4200c2/core-bundle/src/Resources/contao/classes/Backend.php#L572

Or maybe there is an even better approach to check this inside `$act()` itself?

ausi:
There seems to be an issue with input validation still:

1. Create a new page
2. Activate checkbox „Protect page“
3. Do **NOT** select any group
4. Save

This results in an exception: `Contao\DC_Table must call the submit() method to store data in Contao 5`

Toflar:
> How about `persist()` and `onpersist_callback`?

I agree that `commit()` and `persist()` would be better terms if we were to design something from scratch. However, I think for these DC adjustments we should stick with the existing wording and callbacks for maximum compatibility. All our developers already know the `onsubmit` callback and there's nothing new for them to learn.

aschempp:
> This results in an exception: `Contao\DC_Table must call the submit() method to store data in Contao 5`

Does this mean the destruct exception works as intended? 😎

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: > Or maybe there is an even better approach to check this inside $act() itself?

That won't work. The whole point is that this method could have been replaced, otherwise it would already work as intended. Adding this to `Backend` would work, but only for the default backend module case, not for anyone calling the DC methods.

aschempp:
> There seems to be an issue with input validation still:

Fixed in 2b8ae4a69a1fc670739ea9430ff58de827a8069f

ausi:
> Does this mean the destruct exception works as intended? 😎

No.

leofeyer:
> All our developers already know the `onsubmit` callback and there's nothing new for them to learn.

But now there is an `onbeforesubmit_callback`, so this statement is not correct, is it?

Toflar:
It is? It would be even more unintuitive if we had an `onsubmit_callback` plus a new `onpersist_callback`. `onbeforesubmit` is way more speaking (and super common in the whole wild PHP world...).

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3144,54 +3034,166 @@ protected function save($varValue)
...
+			return;
+		}
+
+		if ($this->noReload)
+		{
+			// Data should not be submitted due to validation errors
+			$this->arrSubmit = array();
+
+			return;
+		}
```
ausi, line 3068: Shouldn’t this be checked on the calling side instead?

leofeyer:
My point was that there **is** in fact something new for them to learn, so we should not use this as an excuse for bad naming decisions. But as long as both the `onbeforesubmit` and the `onsubmit` callbacks are triggered in the `submit()` method, I guess it is fine.

core-bundle/src/Resources/contao/drivers/DC_Table.php
m-vo: We cannot throw exceptions in destructors - this interferes with handling of already thrown exceptions during execution + makes exceptions appear at places they were not intended. Including tests for this logic for instance. If you want to enforce a call pattern for an API, this must be done in the methods/accessors. Not calling something and then having no interaction anymore afterwards can't be an error. 😊

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: Similar to calling the `onsubmit_callback`, the idea is to not duplicate the code in every place that needs to `submit` the saved data.

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: Well _cannot_ is a hard statement. https://github.com/symfony/http-client/blob/6.1/Response/AsyncResponse.php#L179 can as well, as @Toflar pointed out 😂 
anyway, I don't see another solution that works in all cases, does anyone else?

ausi: I will try to resolve this in another PR to unblock this PR for now.

Toflar (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3145,54 +3035,166 @@ protected function save($varValue)
...
 
+		// Make sure unique fields are unique
+		if ((\is_array($varValue) || (string) $varValue !== '') && ($arrData['eval']['unique'] ?? null) && !$this->Database->isUniqueValue($this->strTable, $this->strField, $varValue, $this->objActiveRecord->id))
+		{
+			throw new \Exception(sprintf($GLOBALS['TL_LANG']['ERR']['unique'], $arrData['label'][0] ?: $this->strField));
+		}
+
 		// Save the value if there was no error
-		if ((\is_array($varValue) || (string) $varValue !== '' || !($arrData['eval']['doNotSaveEmpty'] ?? null)) && ($this->varValue !== $varValue || ($arrData['eval']['alwaysSave'] ?? null)))
+		if ((\is_array($varValue) || (string) $varValue !== '' || !($arrData['eval']['doNotSaveEmpty'] ?? null)) && ($this->varValue != $varValue || ($arrData['eval']['alwaysSave'] ?? null)))
```
leofeyer: This seems to be a behavior change. What are the implications?

aschempp: yes that is intentional. It should probably be fixed in 4.9 as well. In my test case, one of the values was `'0'` and the other was `0` (int vs. string), and the comparison failed to not save the value.

fritzmg: No fix is needed in Contao 4.9 as far as I can see. It uses `$this->varValue != $varValue` already.

leofeyer: We have changed this in https://github.com/contao/core/issues/5471 and reverted it again for `DC_File` in https://github.com/contao/core/issues/7016, so maybe it is correct to revert it here as well (should also be reverted in `DC_Folder` then). Did you test if `0` and empty strings are still handled correctly with PHP 8 this way?

@ausi Any objections?

leofeyer: @fritzmg No, it doesn‘t?

https://github.com/contao/contao/blob/c9fc13d208a481ea3997cb10314a393567223292/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3138-L3140

fritzmg: Hm, yes, that is strange. The [blame](https://github.com/contao/contao/commit/d332d56b4b67eccbfdc448c327a80b60269d606a) does not show the change 🤔

ausi: > Any objections?

I don’t see how this is related to this PR.

aschempp: That is true. I noticed it because it always updated an empty select column.

ausi: I think this need more investigation. 
For example array changes like `['some string'] != [true]` would then not be saved anymore. Although I’m not sure if such changes are actually possible here.
Also, didn’t we have issues in the past with storing `'0'` in fields that also accept the empty string? This could bring back this issue I think (`0 != ''`)

I’d rather convert the type to match the type in the database (as we do in #4800) and use a strict comparison then.
But please let’s move this into its own PR.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3145,54 +3035,166 @@ protected function save($varValue)
 			}
 		}
 
+		// Make sure unique fields are unique
+		if ((\is_array($varValue) || (string) $varValue !== '') && ($arrData['eval']['unique'] ?? null) && !$this->Database->isUniqueValue($this->strTable, $this->strField, $varValue, $this->objActiveRecord->id))
+		{
+			throw new \Exception(sprintf($GLOBALS['TL_LANG']['ERR']['unique'], $arrData['label'][0] ?: $this->strField));
+		}
+
```
leofeyer: Moving this is in fact also a behavior change that should be discussed in advance and – if we agree on it – be changed in all DC_ classes instead of only here.

ausi (dismissed):

Toflar (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Table.php
leofeyer: > In my test case, one of the values was '0' and the other was 0 (int vs. string), and the comparison failed to not save the value.

As discussed, this is likely caused by PDO returning integers instead of strings in PHP 8.1. We should definitely debug this (also with regards to the `'' == 0` changes in PHP 8) and adjust the comparison accordingly.

leofeyer (approved):

leofeyer:
The unique check fix PR is here: #4813

PR #4797, opened 2022-06-06
Change boolean columns from char(1) to tinyint(1)

### LINKED ISSUE #4184: Use boolean fields for checkboxes
Since Contao 4.9.11 we fully support "boolean" database fields in the data container. In regards to https://github.com/contao/contao/pull/4179 it would thus be more convenient to use

```php
['type' => 'boolean', …]
```

instead of

```php
"char(1) COLLATE ascii_bin NOT NULL default ''"
```

in order to reduce the size of the field from 5 bytes to 1. `char` and `varchar` fields with `utf8mb4` collation currently require 5 bytes in total under InnoDB (4 + 1 for saving the length of the content).

I propose we change this in Contao 5.0. Note that this will need a migration as a database field cannot be changed from `char` (with empty content) to `tinyint` directly (since an empty string is not valid content for `tinyint` and thus will lead to an error in strict mode).

fritzmg:
Implements #4184

_Note:_ the migration automatically converts the empty string value of _all_ `char(1)` fields to `0`, if the DCA SQL definition is of type `boolean`. So this would also automatically be done for any extensions, if applicable.


core-bundle/src/Controller/FrontendModule/TwoFactorController.php
```diff
@@ -150,7 +150,7 @@ private function enableTwoFactor(Template $template, Request $request, FrontendU
         if ('tl_two_factor' === $request->request->get('FORM_SUBMIT')) {
             if ($authenticator->validateCode($user, $request->request->get('verify'))) {
                 // Enable 2FA
-                $user->useTwoFactor = '1';
+                $user->useTwoFactor = true;
```
ausi, line 153: If we read the user from the database would the value not still be fetched as `'1'` (or `'0'` if false)?

fritzmg:
PHP stan complains about

```
Property Contao\User::$useTwoFactor (bool|string) does not accept int.
```

How should we handle this?

1. Change doc to `string|integer` and always use `0`/`1` in assignments.
2. Change all assignments from `0`/`1` to `false`/`true`.
3. Change doc to `string|integer|boolean`.
4. Change doc to `integer|boolean`.

Thoughts:

- When coming from the database, the value will either be a `string` or an `integer` under PDO, depending on the PHP version. However, since Contao 5 already requires PHP 8.1, we can assume that these values are always loaded as integers from the database, correct?
- When using the model, users will likely use `false`/`true` (in the future) when assigning these variables.

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
```diff
@@ -0,0 +1,96 @@
...
+            $tableName = $table->getName();
+
+            try {
+                Controller::loadDataContainer($tableName);
+            } catch (\Throwable) {
+                continue;
+            }
+
+            foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
+                if (!\is_array($fieldConfig['sql'] ?? null) || 'boolean' !== $fieldConfig['sql']['type'] ?? null) {
```
ausi:
```suggestion
                if ('boolean' !== ($fieldConfig['sql']['type'] ?? null)) {
```
is_array check is not needed as this is handled by `??` too.

core-bundle/src/Controller/FrontendModule/TwoFactorController.php
fritzmg: No, it will be `0` or `1` (with PDO under PHP 8.1).

ausi:
> ~~Change doc to `string|integer` and always uses `0`/`1` in assignments.~~

~~I’d favor this one for now I think.~~

Changed my mind to:

> Change doc to `string|integer|boolean`.

ausi:
> we can assume that these values are always loaded as integers from the database, correct?

Probably not because of mysqli :(

fritzmg:
> > Change doc to `string|integer` and always uses `0`/`1` in assignments.
> 
> I’d favor this one for now I think.

It's not very intuitive though to have to write `$model->someBoolean = 1` instead of `true`.

fritzmg:
I think we would need to go with option no. 3 then in that case? i.e. change the doc to `string|integer|boolean` because:

* when loading from the database the value can either be an integer (PDO) or a string (MySQLi)
* when setting values, users will likely want to use `true`/`false`

ausi:
I think we should merge this one with `string|integer|boolean` as this can be changed to `boolean` in #4800 then.

ausi (approved):
❤️

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
```diff
@@ -0,0 +1,96 @@
...
+        $targets = [];
+
+        foreach ($schemaManager->listTables() as $table) {
+            $tableName = $table->getName();
+
+            try {
+                Controller::loadDataContainer($tableName);
+            } catch (\Throwable) {
+                continue;
+            }
```
ausi: This needs to be changed, see #4803

ausi (changes requested):

core-bundle/tests/EventListener/DataContainer/PreviewLinkListenerTest.php
```diff
@@ -141,7 +141,7 @@ public function testSetsTheDefaultValueForDcaFields(string $url, bool $showUnpub
 
         $this->assertTrue($GLOBALS['TL_DCA']['tl_preview_link']['config']['notCreatable']);
         $this->assertSame($url, $GLOBALS['TL_DCA']['tl_preview_link']['fields']['url']['default']);
-        $this->assertSame($showUnpublished ? '1' : '', $GLOBALS['TL_DCA']['tl_preview_link']['fields']['showUnpublished']['default']);
+        $this->assertSame($showUnpublished ? true : false, $GLOBALS['TL_DCA']['tl_preview_link']['fields']['showUnpublished']['default']);
```
leofeyer: `$showUnpublished` is a boolean value already, isn‘t it?

core-bundle/tests/Routing/ResponseContext/CoreResponseContextFactoryTest.php
```diff
@@ -124,7 +124,7 @@ public function testContaoWebpageResponseContext(): void
         $pageModel->title = 'My title';
         $pageModel->description = 'My description';
         $pageModel->robots = 'noindex,nofollow';
-        $pageModel->enableCanonical = '1';
+        $pageModel->enableCanonical = 1;
```
leofeyer: Shouldn‘t this be `true` instead?

fritzmg: My thinking was that we should simulate in the tests how the value would be in a real application. Under PHP 8.1 with PDO, the value will be `1`, not `true`. Except when #4800 gets merged, then it would be `true`.

leofeyer:
As discussed, we want to wait for #4800 before we merge this PR.

fritzmg:
Unfortunately I have hit a roadblock here. Doctrine seems to convert parameter values that are `false` to an empty string, when using

```php
$objDatabase->prepare("INSERT INTO tl_search %s")
            ->set($arrSet)
            ->execute();
```

With this PR the following error will occur when trying to log into the back end for example:

```
An exception occurred while executing a query: SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect integer value: '' for column `c50`.`tl_user`.`disable` at row 1

  at vendor\doctrine\dbal\src\Driver\API\MySQL\ExceptionConverter.php:119
  at Doctrine\DBAL\Driver\API\MySQL\ExceptionConverter->convert(object(Exception), object(Query))
     (vendor\doctrine\dbal\src\Connection.php:1814)
  at Doctrine\DBAL\Connection->handleDriverException(object(Exception), object(Query))
     (vendor\doctrine\dbal\src\Connection.php:1749)
  at Doctrine\DBAL\Connection->convertExceptionDuringQuery(object(Exception), '…', array(…), array())
     (vendor\doctrine\dbal\src\Connection.php:1055)
  at Doctrine\DBAL\Connection->executeQuery('…', array(…), array())
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Database\Statement.php:269)
  at Contao\Database\Statement->query('', array(…))
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Database\Statement.php:228)
  at Contao\Database\Statement->execute(1)
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\User.php:297)
```

The functional tests will also fail [here](https://github.com/contao/contao/blob/eb37857c26de5fd3a230d583681e78ced0178ce4/core-bundle/src/Resources/contao/library/Contao/Search.php#L223-L225) for the same reason.

fritzmg:
As briefly discussed with @ausi I am now casting any booleans to integer in `Statement::query`, so that Doctrine does not convert `false` to an empty string. However, if you use the Doctrine database connection directly, you will need to make of that sure yourself.

core-bundle/src/Command/UserPasswordCommand.php
```diff
@@ -99,7 +99,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 'password' => $hash,
                 'locked' => 0,
                 'loginAttempts' => 0,
-                'pwChange' => $input->getOption('require-change') ? '1' : '',
+                'pwChange' => $input->getOption('require-change') ? 1 : 0,
             ],
             ['username' => $input->getArgument('username')]
```
ausi:
```suggestion
                'pwChange' => $input->getOption('require-change'),
            ],
            ['username' => $input->getArgument('username')],
            [ParameterType::STRING, ParameterType::BOOLEAN, ParameterType::INTEGER, ParameterType::BOOLEAN]
```

something like that?

fritzmg:
Btw.: https://bugs.php.net/bug.php?id=57157 😑 

> From the documentation of PDOStatement::execute()[1]:
> 
> > An array of values with as many elements as there are bound
> > parameters in the SQL statement being executed. All values are
> > treated as PDO::PARAM_STR.
> 
> Coercing TRUE to string yields "1"; coercing FALSE to string
> yields "". Thus, the reported behavior is expected.  Explicitly
> bind as PDO::PARAM_INT instead.
> 
> [1] <https://www.php.net/manual/en/pdostatement.execute.php>

ausi (dismissed):

ausi (dismissed):

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
ausi:
```suggestion
            Controller::loadDataContainer($tableName);
```

I think we should just remove the try catch block here, and fix this together with #4803 in another PR.

fritzmg:
I still need to check if there are more places where we need to cast `bool` to `int` (or pass the parameter type).

fritzmg:
_Note:_ I only checked a few more places. I'd vote to merge this as is and then fix any issue later on as discovered.

comments-bundle/src/Resources/contao/models/CommentsNotifyModel.php
```diff
@@ -15,16 +15,16 @@
...
- * @property string         $tokenRemove
+ * @property integer $id
+ * @property integer $tstamp
+ * @property string  $source
+ * @property integer $parent
+ * @property string  $name
+ * @property string  $email
+ * @property string  $url
+ * @property integer $addedOn
+ * @property boolean $active
```
aschempp, line 26: are these really booleans now? Not integers?

core-bundle/src/Resources/contao/dca/tl_user.php
```diff
@@ -890,7 +890,7 @@ public function checkAdminStatus($varValue, DataContainer $dc)
 	{
 		if (!$varValue && $this->User->id == $dc->id)
 		{
-			$varValue = '1';
+			$varValue = true;
```
aschempp, line 893: since this is a `save_callback`, it should probably set an integer value?

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -253,7 +253,13 @@ public function query($strQuery='', array $arrParams = array(), array $arrTypes
 		$arrParams = array_map(
 			static function ($varParam)
 			{
-				if (\is_string($varParam) || \is_bool($varParam) || \is_float($varParam) || \is_int($varParam) || $varParam === null)
+				// Cast boolean to integer, otherwise PDO will convert "false" to an empty string (see https://bugs.php.net/bug.php?id=57157)
```
aschempp: not sure I like this. We should include the type information of columns, then Doctrine would automatically convert the values according to the database abstraction (not all platforms use integer as boolean). Not sure if we can do that in all places though, maybe we only cast if the column is not given in `$arrTypes`?

comments-bundle/src/Resources/contao/models/CommentsNotifyModel.php
fritzmg: Yes, see https://github.com/contao/contao/pull/4800

core-bundle/src/Resources/contao/dca/tl_user.php
fritzmg: It will automatically be converted to `int` by Contao anyway.

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
fritzmg: > Not sure if we can do that in all places though

We can in fact not do it with `Contao\Database\Statement::set()` currently. Also it would be cumbersome to do that in any case.

However yes, we could at least check for `empty($arrTypes)`.

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
aschempp: now we can 😎 https://github.com/contao/contao/pull/4839
Yes it might be cumbersome, but that's just how Doctrine works. If you use boolean fields in DBAL, you must include type information for it to work. Maybe our model abstraction could automatically include it. Also, I'm fine with still casting if there is no type 🙃. Just be aware that params and types are NOT keyed, they are by order.

core-bundle/src/Command/UserListCommand.php
```diff
@@ -89,7 +89,7 @@ private function getUsers(bool $onlyAdmins = false): Collection|null
         $userModel = $this->framework->getAdapter(UserModel::class);
 
         if ($onlyAdmins) {
-            return $userModel->findBy('admin', '1');
+            return $userModel->findBy('admin', 1);
```
leofeyer: Shouldn‘t this be `true` instead? I know that the DB column expects `0` or `1` but this seems bad DX to me:

```
$user = $userModel->findBy('admin', 1);

dump($user->admin); // true
```

And if we want to change the value:

```
$user->admin = false;
$user->save();
```

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -697,7 +697,7 @@ public static function addPageIcon($row, $label, DataContainer $dc=null, $imageA
 		}
 
 		$image = Controller::getPageStatusIcon((object) $row);
-		$imageAttribute = trim($imageAttribute . ' data-icon="' . Image::getPath(Controller::getPageStatusIcon((object) array_merge($row, array('published'=>'1')))) . '" data-icon-disabled="' . Image::getPath(Controller::getPageStatusIcon((object) array_merge($row, array('published'=>'')))) . '"');
+		$imageAttribute = trim($imageAttribute . ' data-icon="' . Image::getPath(Controller::getPageStatusIcon((object) array_merge($row, array('published'=>1)))) . '" data-icon-disabled="' . Image::getPath(Controller::getPageStatusIcon((object) array_merge($row, array('published'=>0)))) . '"');
```
leofeyer, line 700: Shouldn‘t this be `true` here as well?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2691,7 +2691,7 @@ public function toggle($intId=null)
 		Input::setPost('FORM_SUBMIT', $this->strTable);
 		$this->varValue = $objRow->{$this->strField};
 
-		$this->save($this->varValue ? '' : '1');
+		$this->save($this->varValue ? false : true);
```
leofeyer:
```suggestion
		$this->save(!$this->varValue);
```

core-bundle/tests/Contao/ControllerTest.php
```diff
@@ -142,181 +142,181 @@ public function pageStatusIconProvider(): \Generator
         yield 'Published' => [
             $this->mockClassWithProperties(PageModel::class, [
                 'type' => 'regular',
-                'hide' => '',
-                'protected' => '',
+                'hide' => 0,
```
leofeyer: Why is this `0` here but `false` on line 157?

core-bundle/tests/Controller/SitemapControllerTest.php
```diff
@@ -103,7 +103,7 @@ public function testGeneratesSitemapForRegularPages(): void
             'pid' => 42,
             'type' => 'regular',
             'groups' => [],
-            'published' => '1',
+            'published' => 1,
```
leofeyer: This should be `true`, shouldn‘t it? Because we are mocking a model here. (Seems like you have been using boolean values as of line 435 already?)

core-bundle/tests/EventListener/Security/TwoFactorFrontendListenerTest.php
```diff
@@ -261,15 +261,15 @@ public function testReturnsIfTheUserIsAlreadyAuthenticated(): void
...
 
         $pageModel = $this->mockClassWithProperties(PageModel::class);
         $pageModel->id = 1;
-        $pageModel->enforceTwoFactor = '';
+        $pageModel->enforceTwoFactor = false;
         $pageModel->twoFactorJumpTo = 1;
 
         $page401 = $this->mockClassWithProperties(PageModel::class);
-        $page401->autoforward = '';
+        $page401->autoforward = 1;
```
leofeyer: Why an integer here but booleans on line 264 and 268?

leofeyer (changes requested):
This PR still seems very inconsistent to me. If I understand correctly, we have to use integers when working with DB queries and result objects, and booleans when working with models. Any exception to this rule, e.g. in a unit test, should be explicitly noted.

fritzmg:
Yes, the inconsistency comes from the change in #4800 and I found it to not be as easy anymore to find all `0` and `1` that need to be `false` and `true` now.

core-bundle/src/Resources/contao/classes/Backend.php
fritzmg: The `$row` is a `\Contao\Database\Result` row and not the row of a Model.

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
```diff
@@ -0,0 +1,92 @@
...
+        $schemaManager = $this->connection->createSchemaManager();
+        $targets = [];
+
+        foreach ($schemaManager->listTables() as $table) {
+            $tableName = $table->getName();
+
+            Controller::loadDataContainer($tableName);
+
+            foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
+                if ('boolean' !== ($fieldConfig['sql']['type'] ?? null)) {
```
rabauss:
```suggestion
                if (\Doctrine\DBAL\Types\Types::BOOLEAN !== ($fieldConfig['sql']['type'] ?? null)) {
```
why not using the constant? could also be used in the dca, couldn't it?

fritzmg: For 5.x this is ok. But if you support older Doctrine versions you can't use them as they had been moved once (just as a side note).

fritzmg: I would leave the DCAs as they are for now and may be later switch to the constant in a separate PR.

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -253,8 +253,9 @@ public function query($strQuery='', array $arrParams = array(), array $arrTypes
 		$arrParams = array_map(
 			static function ($varParam)
 			{
-				// Cast boolean to integer, otherwise PDO will convert "false" to an empty string (see https://bugs.php.net/bug.php?id=57157)
-				if (\is_bool($varParam))
+				// Automatically cast boolean to integer when no types are defined, otherwise
+				// PDO will convert "false" to an empty string (see https://bugs.php.net/bug.php?id=57157)
+				if (empty($arrTypes) && \is_bool($varParam))
```
ausi, line 258: Missing `use ($arrTypes)` I think?

ausi (dismissed):

comments-bundle/src/Resources/contao/models/CommentsModel.php
```diff
@@ -119,7 +119,7 @@ public static function findPublishedBySourceAndParent($strSource, $intParent, $b
 
 		if (!static::isPreviewMode($arrOptions))
 		{
-			$arrColumns[] = "$t.published='1'";
+			$arrColumns[] = "$t.published=1";
```
m-vo, line 122: Should we also go for `=TRUE` and `=FALSE` in comparisons?
These are synonyms for `1` and `0` in MySQL: https://dev.mysql.com/doc/refman/8.0/en/boolean-literals.html

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
```diff
@@ -0,0 +1,97 @@
...
+    public function run(): MigrationResult
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+
+        foreach ($this->getTargets() as [$table, $column]) {
+            if (!$schemaManager->tablesExist([$table]) || !isset($schemaManager->listTableColumns($table)[$column])) {
+                continue;
+            }
+
+            $this->connection->update($table, [$column => 0], [$column => '']);
```
m-vo:
```suggestion
            $this->connection->update($table, [$column => '0'], [$column => '']);
```

core-bundle/tests/Security/User/UserCheckerTest.php
```diff
@@ -107,9 +107,9 @@ public function testThrowsAnExceptionIfTheAccountIsNotActiveYet(): void
         $user = $this->mockClassWithProperties(FrontendUser::class);
         $user->username = 'foo';
         $user->locked = 0;
-        $user->disable = '';
-        $user->login = '1';
-        $user->start = $time;
+        $user->disable = false;
+        $user->login = true;
+        $user->start = (string) $time;
```
m-vo, line 112: The `$user->$start` property is supposed to be `string|integer`, but why needs this test to be changed?

m-vo (dismissed):
LGTM :heart: 

I just left some minor questions/nitpicks.

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
fritzmg: Why a string?

fritzmg:
Note: the changes from https://github.com/contao/contao/pull/4881 also need to be integrated into the migration of this PR.

core-bundle/tests/Security/User/UserCheckerTest.php
fritzmg: `$user->start` is a `varchar` in the database - and it will stay a string when loaded from the database. As this test simulates a model returned from the database, the value should be a string, not an integer.

comments-bundle/src/Resources/contao/models/CommentsModel.php
fritzmg: I personally prefer `0`/`1` as it's shorter, but I am fine either way. /cc @contao/developers

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
m-vo: Isn't the idea of the migration to prepare the database before the column types get changed? They will still be `CHAR(1)` at this point.

fritzmg: I suppose so, not that it matters much in the end :)

m-vo: Yes, it was more of a mental help to help you understand when this is executed. But feel free to resolve. :slightly_smiling_face:

comments-bundle/src/Resources/contao/models/CommentsModel.php
leofeyer: I also prefer the integers. 👍

ausi (dismissed):

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -544,24 +544,24 @@ static function ($value)
 		(
 			'exclude'                 => true,
 			'inputType'               => 'select',
-			'options'                 => array(''=>'http://', '1'=>'https://'),
+			'options'                 => array(0=>'http://', 1=>'https://'),
```
leofeyer:
```suggestion
			'options'                 => array('http://', 'https://'),
```

Adding the array keys is redundant, isn‘t it?

fritzmg: I'd still prefer it in this case, but I'm fine either way.

fritzmg: Ah - actually neither will work. Now that the values are normal integers, the widget will treat the labels as values. I think we'll need a load and save callback here. Or use `reference`.

m-vo: Would `'isAssociative' => true` solve this?

fritzmg: @m-vo I'll try that. Otherwise I fixed it in bfe3b49aca5f42bd38e6246dfb9fe6c1860efb06

fritzmg: Yes, that works too 👍

fritzmg:
~~There is still an issue with the `checkbox` widget. It currently gets `checked` even though the value is `0`.~~ // false alarm

m-vo (dismissed):

core-bundle/src/Resources/contao/dca/tl_page.php
```diff
@@ -544,8 +544,8 @@ static function ($value)
 		(
 			'exclude'                 => true,
 			'inputType'               => 'select',
-			'options'                 => array(0=>'http://', 1=>'https://'),
-			'eval'                    => array('tl_class'=>'w50'),
+			'options'                 => array('http://', 'https://'),
+			'eval'                    => array('tl_class'=>'w50', 'isAssociative' => true),
```
leofeyer, line 548: Why the flag?

leofeyer (changes requested):
There are still plenty of wrong values in the code base, I‘m afraid. A first quick search revealed the following:

- CalendarEventsModel.php line 269
- Dbafs.php line 516
- FormFieldModel.php line 201
- FormModel.php line 121
- ImageSizeItemModel.php line 95
- MemberGroupModel.php line 83 and 101
- MemberModel.php line 197
- Module.php line 476
- NewsModel.php line 238

We have to make sure to search for `='1'` and `=''` as well as `= '1'` and `= ''` to find all occurrences.

core-bundle/src/Resources/contao/dca/tl_page.php
fritzmg: See https://github.com/contao/contao/pull/4797#discussion_r909607112 - otherwise the select won't work.

fritzmg:
> We have to make sure to search for `='1'` and `=''` as well as `= '1'` and `= ''` to find all occurrences.

Also `=> ''` and `=> '1'`.

I made another sweep.

core-bundle/src/Migration/Version500/BooleanFieldsMigration.php
```diff
@@ -0,0 +1,101 @@
...
+
+                $field = $columns[$fieldName];
+
+                if ($field->getType() instanceof StringType) {
+                    $targets[] = [$tableName, $fieldName];
+                }
+            }
+        }
+
+        return $targets;
```
rabauss, line 99: Are fields of extensions included too, right?

fritzmg: Yes, see my initial comment. However in the new version, any fields of _dynamcially_ added DCAs (like Catalog Manager or MetaModels) would not be included.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4592, opened 2022-05-02
Support both `collation:` and `collate:`

leofeyer:
Fixes #4560

PR #4799, opened 2022-06-06
Always set both collate and collation to the same value

### LINKED ISSUE #4798: 'utf8mb4_unicode_ci' is not valid for CHARACTER SET 'utf8'
### Affected version(s)

4.13, 5.x

### Description

In #4592 we added `collation: utf8mb4_unicode_ci` to the config skeleton. However, many users have something like this in their `config.yaml`

```yaml
doctrine:
    dbal:
        connections:
            default:
                default_table_options:
                    charset: utf8
                    collate: utf8_unicode_ci
```

due to https://docs.contao.org/manual/en/installation/system-requirements/#mysql-minimum-requirements

The addition in the config skeleton in the manager-bundle will then create a mismatch between the `charset` and the `collation`, as the `charset` is set to `utf8`, but the `collation` is set to `utf8mb4_unicode_ci` - which will result in an

```
'utf8mb4_unicode_ci' is not valid for CHARACTER SET 'utf8'
```

error when updating the database in the Install Tool or via `contao:migrate`.

Can anyone think of a solution for this, other than updating the documentation and changing `collate` to `collation`? (I am not sure if _both_ are actually needed as the original related Doctrine issues are hard to follow.)

fritzmg:
Fixes #4798, follow up to #4592

This PR adds an additional extension config via the Contao Manager Plugin so that both the `collate` _and_ the `collation` options are always set to the same value - in this case whichever value came last in the collection of extension configs.

### DIFF
--- a/manager-bundle/src/ContaoManager/Plugin.php
+++ b/manager-bundle/src/ContaoManager/Plugin.php
@@ -247,8 +247,10 @@ class Plugin implements BundlePluginInterface, ConfigPluginInterface, RoutingPlu
 
                 $extensionConfigs = $this->addDefaultPdoDriverOptions($extensionConfigs, $container);
                 $extensionConfigs = $this->addDefaultDoctrineMapping($extensionConfigs, $container);
+                $extensionConfigs = $this->enableStrictMode($extensionConfigs, $container);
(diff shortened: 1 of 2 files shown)


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4800, opened 2022-06-06
Cast model values to the correct PHP type

ausi:
~~Proof of concept.~~

With this we would get the correct types (bool, int or float) depending on the database type for the values in our models.
Together with #4797 we could then cleanup the model property types quite good, see `PageModel` as an example in this PR.

Setting the wrong type is still allowed for now, but triggers a deprecation notice.

Todo:

- [x] ~~Merge #4797 first~~ No: merge this one first!
- [x] Do the same for the `User` class
- [x] Adapt all tests to use the correct types

@contao/developers WDYT?


Toflar:
Love this! Not sure about the implementation, though. Did you look into the types of Doctrine itself? They all have a `convertToDatabaseValue()` and `convertToPhpValue()` ;)

ausi:
> Did you look into the types of Doctrine itself?

Yes, I tried `convertToPhpValue()` first but it caused more problems than it solved. Because of how our models work, I think applying this only for the simple scalar types is the better way for now.

fritzmg (commented):
General concept looks good to me 👍 

Not sure about the cache file location, but I have no better suggestion. Also may be the functionality itself should be its own service, rather than building it into the Model class?

ausi:
> Also may be the functionality itself should be its own service, rather than building it into the Model class?

I thought that performance wise a static array in the `Model` class would be better than having to call `System::getContainer()` all the time 😕

core-bundle/src/Cache/ContaoCacheWarmer.php
```diff
@@ -199,6 +201,14 @@ private function generateTemplateMapper(string $cacheDir): void
         );
     }
 
+    private function generateColumnCastTypes(string $cacheDir): void
+    {
+        $this->filesystem->dumpFile(
+            Path::join($cacheDir, 'contao/config/column_cast_types.php'),
```
leofeyer: Just a nitpick, I would like `column-types.php` better. 🙈

ausi: Renamed in 336ec3db7663371f3c6ad74aa6c67d32bd280d5c

fritzmg (commented):
Just to confirm: in my PR I will then be able to change all `string|boolean` type hints to `boolean`, correct?

ausi:
> in my PR I will then be able to change all `string|boolean` type hints to `boolean`, correct?

yes

core-bundle/src/Resources/contao/models/UserModel.php
```diff
@@ -45,18 +45,18 @@
  * @property string|array|null $formp
  * @property string|array|null $amg
  * @property string|boolean    $disable
- * @property string|integer    $start
- * @property string|integer    $stop
+ * @property integer           $start
+ * @property integer           $stop
```
fritzmg: These are `varchar` in the database and thus they would still be returned as a string from there - and not converted to integer automatically, will they?

ausi: Fixed in abdd3530a5f4bd2cf3240e7f7634e876ce33b014

fritzmg: On a side note: why _do_ we save those timestamps as a string in the first place? I guess it's just an historic remnant from Contao 2 and we could may be change the field type to `int(10) default NULL` and then alter the queries to

```diff
-start='' OR start<='$time'
+start IS NULL OR start<=$time
```

❓

calendar-bundle/src/Resources/contao/models/CalendarEventsModel.php
```diff
@@ -16,50 +16,50 @@
...
+ * @property string|null       $description
+ * @property string            $location
+ * @property string            $address
+ * @property string|null       $teaser
+ * @property string|boolean    $addImage
+ * @property string|boolean    $overwriteMeta
+ * @property string|null       $singleSRC
+ * @property string            $alt
+ * @property string            $imageTitle
+ * @property integer           $size
```
fritzmg: `tl_calendar_events.size` and the other image size fields are a `varchar` and contains either a string consisting of a serialized array or a database ID - so this needs to stay `string|integer` as well.

core-bundle/src/Resources/contao/models/FilesModel.php
```diff
@@ -21,21 +21,21 @@
...
- * @property string              $hash
- * @property string|boolean      $found
- * @property string              $name
- * @property string|float        $importantPartX
- * @property string|float        $importantPartY
- * @property string|float        $importantPartWidth
- * @property string|float        $importantPartHeight
- * @property string|array|null   $meta
+ * @property integer           $id
+ * @property integer|null      $pid
```
fritzmg: This seems to have been a previous mistake. `tl_files.pid` is always an UUID or NULL. So this should always have been `string|null` (just like `tl_files.uuid`).

core-bundle/src/Resources/contao/models/LayoutModel.php
```diff
@@ -32,7 +32,7 @@
  * @property string|array|null $modules
  * @property string            $template
  * @property string|boolean    $minifyMarkup
- * @property string|integer    $lightboxSize
+ * @property integer           $lightboxSize
```
fritzmg: Should be `string|integer` (see above).

fritzmg (changes requested):

core-bundle/src/Resources/contao/models/UserModel.php
ausi: Agree, we should change that.
In a separate PR?

fritzmg: Yes, definitely a separate PR as this would then be BC breaking change.

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4804, opened 2022-06-08
Fix SearchIndexSubscriber not respecting noindex hint

### LINKED ISSUE #4795: Contao search engine: News article with "noindex" is still indexed
### Affected version(s)

4.13.5

### Description

When you select "Robots tag: noindex,follow" in a single news article setting, this article is still indexed by the Contao search engine and visible in the list of search results.

**How to reproduce:**

- Contao installation with newslist and newsreader (both pages with default "robots: index" settings)
- some news articles
- a page with the search engine module

Now go to the news articles and change the robots settings of one single news article to "noindex".

Then under "Maintenance" do "Purge the search index" and after that "Update the search index" > "Start crawling".

Then to the frontend: If you now search for a term that is only mentioned in the "noindex" news article, this article is still shown in the search results (but shouldn't because of the "noindex" setting).

Toflar:
Fixes https://github.com/contao/contao/issues/4795

### DIFF
--- a/core-bundle/src/Crawl/Escargot/Subscriber/SearchIndexSubscriber.php
+++ b/core-bundle/src/Crawl/Escargot/Subscriber/SearchIndexSubscriber.php
@@ -29,6 +29,7 @@ use Terminal42\Escargot\EscargotAwareInterface;
 use Terminal42\Escargot\EscargotAwareTrait;
 use Terminal42\Escargot\Subscriber\ExceptionSubscriberInterface;
 use Terminal42\Escargot\Subscriber\HtmlCrawlerSubscriber;
+use Terminal42\Escargot\Subscriber\RobotsSubscriber;
 use Terminal42\Escargot\Subscriber\SubscriberInterface;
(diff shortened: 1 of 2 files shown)


ausi:
Just to double check: Even if `"noSearch": false,` is specified in the JSON-LD, meta robots noindex should overrule that?

fritzmg:
I think so. We never really specified/documented this, but at the moment (meaning even without this PR via the `kernel.terminate` indexing) it's like this:

* You can allow Google to index a page, while disallowing it for Contao's search index independently.
* You can set a page to `noindex`, which will then not be indexed by either Google or Contao.
* Disallowing Google while still allowing Contao is not possible.

Toflar:
Needs work -> draft.

Toflar:
Needs work -> draft.

Toflar:
Now we're good here :)

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #4808, opened 2022-06-09
Various PHP 8 fixes

aschempp:


core-bundle/src/Translation/MessageCatalogue.php
```diff
@@ -178,6 +178,10 @@ private function getFromGlobals(string $id): ?string
             $item = &$item[$part];
         }
 
-        return $item;
+        if (\is_array($item)) {
+            return $id;
```
ausi:
```suggestion
            return null;
```

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4761,7 +4761,7 @@ protected function parentView()
 	protected function listView()
 	{
 		$table = ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) == self::MODE_TREE_EXTENDED ? $this->ptable : $this->strTable;
-		$orderBy = $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['fields'] ?? array();
+		$orderBy = $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['fields'] ?? array('id');
 		$firstOrderBy = preg_replace('/\s+.*$/', '', $orderBy[0]);
```
ausi, line 4765: Shouldn’t we fix this on the second line instead?

```suggestion
		$orderBy = $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['fields'] ?? array();
		$firstOrderBy = preg_replace('/\s+.*$/', '', $orderBy[0] ?? '');
```

aschempp: I thought about this as well, but ordering by an empty string is imho worse than ordering by the record id (maybe sorting by empty won't even work).

ausi: I’m not unserstanding all the rest of the code, is it guaranteed that there is always a column named `'id'`?
If so, using `array('id')` might be better I agree.

aschempp: well it's a DCA, and our DCA does not work without an ID column imho 🤷

ausi (dismissed):

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -211,7 +211,7 @@ public static function getTemplateGroup($strPrefix, array $arrAdditionalMapper=a
 					}
 				}
 
-				$arrTemplates[$strTemplate][] = $GLOBALS['TL_LANG']['MSC']['global'];
+				$arrTemplates[$strTemplate][] = $GLOBALS['TL_LANG']['MSC']['global'] ?? 'Global';
```
leofeyer:
```suggestion
				$arrTemplates[$strTemplate][] = $GLOBALS['TL_LANG']['MSC']['global'] ?? 'global';
```

leofeyer:
What was the issue in the `MessageCatalogue` class?

aschempp:
> What was the issue in the `MessageCatalogue` class?

```
$GLOBALS['TL_LANG']['MSC']['foo']['bar'] = 'baz';

$this->trans('MSC.foo')
```

will lead to an array to string casting error.

core-bundle/src/Resources/contao/library/Contao/Controller.php
aschempp: why? It's the title of the global template `optgroup`?

leofeyer: The default string for `$GLOBALS['TL_LANG']['MSC']['global']` is also `global`, therefore I figured it should be consistent. What do you mean by _global template optgroup_?

leofeyer:
The same error would be thrown if you did `echo $GLOBALS['TL_LANG']['MSC']['foo']`. I don‘t think we should ignore it, so the developer realizes that something is wrong.

leofeyer (approved):

PR #4693, opened 2022-05-18
Reduce the row sizes of tl_module and tl_content

ausi:
Backport of https://github.com/contao/contao/pull/4179#issuecomment-1081716123

PR #1731, opened 2020-05-12
Prevent arbitrary column renaming in the schema update

leofeyer:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #646 #1716
| Docs PR or issue | -

Solution provided by @dmolineus. Thank you.

PR #4809, opened 2022-06-09
Add a migration for bad playerColor data

ausi:
We shortened `playerColor` in #4693 to 6 characters.

The community noticed that there are many installations where a erroneous column rename from `com_template` to `playerColor` because of #1731 which now results in an SQL error `1265 Data truncated for column 'playerColor'` because the column is filled with `'com_default'` values.

This migration should fix the update for this case.

### DIFF
--- a/core-bundle/src/Migration/Version409/PlayerColorMigration.php
+++ b/core-bundle/src/Migration/Version409/PlayerColorMigration.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Migration\Version409;
+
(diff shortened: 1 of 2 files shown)


core-bundle/src/Migration/Version409/PlayerColorMigration.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        return (bool) $this->connection
+            ->executeQuery("
+                SELECT EXISTS(
+                    SELECT playerColor
+                    FROM tl_content
+                    WHERE
+                        CHAR_LENGTH(playerColor) > 6
+                        AND playerColor LIKE 'com_%'
```
ausi, line 55: I wanted to make sure we only handle the known case of `com_*` values here, so that we don’t accidentially delete data that does not come from that column rename issue.

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4810, opened 2022-06-09
Modern fragments: text content elements (part 2)

m-vo:
Followup to #4775 

This PR adds the missing text content elements:
* `content_element/text`
* `content_element/table`

It also improves the usage of the `HtmlAttributes` class by allowing conditions everywhere.

Todo:

- [x] add tests
- [x] adjust/inline table sorting script

~~**Note:** This PR is currently based on #4775 and #4730; needs to be rebased once merged.~~


core-bundle/src/String/HtmlAttributes.php
```diff
@@ -145,11 +147,20 @@ public function unset(string $name, \Stringable|bool|int|string|null $condition
         return $this;
     }
 
-    public function addClass(string ...$classes): self
+    /**
+     * Add a single class ("foo") or multiple from a class string ("foo bar baz").
+     *
+     * If a falsy $condition is specified, the method is a no-op.
+     */
+    public function addClass(string $classes, mixed $condition = true): self
```
ausi: Just a thought: how about allowing an array of strings too?

core-bundle/src/Controller/ContentElement/TableController.php
```diff
@@ -0,0 +1,51 @@
...
+ * @ContentElement(category="texts")
+ */
+class TableController extends AbstractContentElementController
+{
+    protected function getResponse(FragmentTemplate $template, ContentModel $model, Request $request): Response
+    {
+        // Table fields
+        $rows = StringUtil::deserialize($model->tableitems, true);
+        $header = $model->thead && !empty($rows) ? array_shift($rows) : [];
+        $footer = $model->tfoot && !empty($rows) ? array_pop($rows) : [];
```
ausi:
```suggestion
        $header = $model->thead && $rows ? array_shift($rows) : [];
        $footer = $model->tfoot && $rows ? array_pop($rows) : [];
```
Nitpick ☺️

core-bundle/src/Controller/ContentElement/TableController.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $template->set('header', $header);
+        $template->set('footer', $footer);
+        $template->set('rows', $rows);
+        $template->set('use_row_headers', (bool) $model->tleft);
+
+        $template->set('caption', $model->summary ?: null);
+
+        // Client side sorting
+        $sorting = empty($header) || !$model->sortable ? null : [
```
ausi:
```suggestion
        $sorting = !$header || !$model->sortable ? null : [
```

core-bundle/src/Resources/contao/templates/_new/content_element/table.html.twig
```diff
@@ -0,0 +1,79 @@
...
+
+{# todo: remove once the table sorting script was adjusted or inlined #}
+{% set attributes = attrs().addClass('ce_table') %}
+
+{% block content %}
+    {% set table_attributes = attrs(table_attributes|default)
+        .set('id', 'table_' ~ data.id, sorting)
+        .addClass('sortable', sorting)
+        .set('data-sort-default', sorting.column|default ~ '|' ~ sorting.order|default, sorting)
+    %}
```
ausi:
```suggestion
    {% set table_attributes = attrs(table_attributes|default) %}
    {% if sorting %}
        {% table_attributes
            .set('id', 'table_' ~ data.id)
            .addClass('sortable')
            .set('data-sort-default', sorting.column|default ~ '|' ~ sorting.order|default)
        %}
    {% endif %}
```
Would something like this be more readable?

core-bundle/src/Resources/contao/templates/_new/content_element/table.html.twig
```diff
@@ -0,0 +1,79 @@
...
+            {% endblock %}
+        {% endif %}
+
+        {# Header #}
+        {% if header %}
+            {% block header %}
+                <thead>
+                {% block header_inner %}
+                    <tr>
+                        {% for col in header %}
```
ausi: Maybe rename `col` to `cell` everywhere in this template?

core-bundle/src/String/HtmlAttributes.php
m-vo: Ah, you mean `string|array $classes`? I like :+1:

core-bundle/src/Resources/contao/templates/_new/content_element/table.html.twig
m-vo: Had this before, but then changed it to the inline version. :smile: 

But already convinced. I used the `{% do %}` tag in 69828a6251c752e1d073c458ec3944c7d0fbb7b2.

core-bundle/src/Resources/contao/templates/_new/content_element/table.html.twig
m-vo: see ea7409be00c212879e48208b6dd86f63f8b89ee2

core-bundle/src/String/HtmlAttributes.php
m-vo: see fa0b4690c3778a48e23d1498f0897959940d189b

ausi (dismissed):
Approved, but #4775 and #4730 need to be merged first.

core-bundle/src/Controller/ContentElement/TableController.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $template->set('header', $header);
+        $template->set('footer', $footer);
+        $template->set('rows', $rows);
+        $template->set('use_row_headers', (bool) $model->tleft);
+
+        $template->set('caption', $model->summary ?: null);
+
+        // Client side sorting
+        $sorting = !$header || !$model->sortable ? null : [
```
leofeyer, line 41:
```suggestion
        $sorting = (!$header || !$model->sortable) ? null : [
```

Not sure if the operator precedence is correct otherwise.

m-vo: `||` has a higher precedence than `? :`, so it should work, But I agree, that the parantheses make it more clear, so let's go with it. :+1:

m-vo: Or maybe not. The CS fixer (`UselessParenthesesSniff`) removed it again :smile:

m-vo:
This is rebased and IMHO ready to go now as well. :+1:

ausi (dismissed):

m-vo:
I updated the PR to include an accessible vanilla table sorting script (https://www.dhs.state.il.us/page.aspx?item=55196).

I also refactored the table creation into a neat component. Now you could even create tables on the fly:
```twig
{% with {rows: [['foo', 'bar'], ['baz','foobar']], sorting: true} %}
  {{ block('table_component') }}
{% endwith %}
```

/cc @ausi

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
```diff
@@ -0,0 +1,196 @@
...
+                descendingText: 'MSC.descending'|trans,
+            }|merge(table_sorting_settings|default({})) %}
+
+            {% add "table_sorting_script" to head %}
+                <script>
+                    {
+                        const instances = new WeakMap();
+                        const initTree = (root) => {
+                            const elements = root.tagName === 'TABLE' && root.hasAttribute('data-sortable-table') ?
+                                [root] : root.querySelectorAll('table[data-sortable-table]');
```
ausi:
```suggestion
                            const elements = root.matches('table[data-sortable-table]') ?
                                [root] : root.querySelectorAll('table[data-sortable-table]');
```

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
```diff
@@ -0,0 +1,196 @@
...
+                                    } catch (e) {
+                                        // ignore
+                                    }
+                                }
+                            });
+                        }
+
+                        const script = document.createElement('script');
+                        script.setAttribute('src', 'bundles/contaocore/sortabletable.js');
+                        script.setAttribute('async', true);
```
ausi:
```suggestion
                        script.src = 'bundles/contaocore/sortabletable.js';
                        script.async = true;
```

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
```diff
@@ -0,0 +1,196 @@
...
+                            new MutationObserver(function (mutations) {
+                                mutations.forEach(function (mutation) {
+                                    initTree(mutation.target)
+                                });
+                            }).observe(document.documentElement, {
+                                attributes: true,
+                                attributeFilter: ['data-sortable-table'],
+                                childList: true,
+                                subtree: true
+                            });
```
ausi:
```suggestion
                            }).observe(document.documentElement, {
                                childList: true,
                                subtree: true
                            });
```

Observing attributes is overkill here I think and not worth the performance cost IMO.

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
```diff
@@ -0,0 +1,196 @@
...
+                        })
+
+                        document.head.appendChild(script);
+                        document.addEventListener('domready', () => initTree(document));
```
ausi: I think we should remove the `domready` event here and call `initTree(document)` in the script onload listener instead.

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
```diff
@@ -0,0 +1,196 @@
+{#
+    This component outputs a table with client side sorting capabilities.
+
+        <table data-sortable-table="1|asc">
```
ausi: This format seems odd, should we use a space seperated string, which is more common in HTML attributes?

```suggestion
        <table data-sortable-table="1 asc">
```

Or two attributes?

```suggestion
        <table data-sortable-table="1" data-sortable-table-direction="asc">
```

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
m-vo: But then the script might be loaded before the DOM is ready and we might miss the initial elements. Or can't this be the case?

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
m-vo: Agreed, I just used this because we had this in the old element. Changed in c6e5938283edc05dcbbd9adf8aa74aca04ecfbbf

core-bundle/src/Resources/contao/templates/_new/component/_table.html.twig
m-vo: Convinced. :relaxed: 
See ad12295de57854252bdf9817a0f47cdc6dc03690.

ausi (dismissed):

leofeyer (changes requested):
The changes are perfect, however, we cannot ship the `sortable.js` script this way. On the one hand, it should be a reusable component, and on the other hand, it should also be used e.g. here:

https://github.com/contao/contao/blob/47697073752ae4d6fa6b63b28e0b54c50a8143f3/core-bundle/src/Resources/public/modulewizard.js#L23-L31

ausi (dismissed):
Only open question is IMO which script to use for `sortabletable.js` and how to distribute that script (e.g. via contao-components).

I’d be happy with the what’s currently in sortabletable.js too.

m-vo:
I switched to [tristen/tablesort](https://github.com/tristen/tablesort) in b06d17381f519f5989493bb50c71db7434fb4342. 

For the purpose of testing, the script is still in the resources folder - we would need to remove it from there and adjust the path to a contao-component once released as one.

ausi (dismissed):
Approved apart from the script needing to be moved to contao-components

m-vo:
With fffb3e943c1a85ecc7e4d4f79c756bd7384ba701 we're now using https://github.com/contao-components/tristen-tablesort instead of manually distributing the files.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4812, opened 2022-06-09
Do not use abbreviations for section templates

### LINKED ISSUE #3386: rename com_default to comment_default (and others)
It's a small issue, but nevertheless something that comes up frequently on the community forum for example. When someone wants to create their own template for the comments output, they often just search for or `comment` in the template section of the back end and thus do not find the actual template (just the `ce_comments`, `mod_comment_form` or `mod_comments` templates).

In general there are some `*_default` templates that have no abbreviation, like

* gallery_default
* mail_default
* member_default
* picture_default
* search_default
* info_default
* list_default

and some that do, like

* cal_default
* com_default
* nav_default
* rss_default (that's fine of course)
* nl_default

There does not seem to be a reason why `comment_default` is shortened to `com_default`. The word `comment` has 7 letters for example, but so does `gallery` or `picture`, which are not abbreviated. 

Of course this change can only be done in Contao 5.0.0, but I wanted to mention it nevertheless.

leofeyer:
Fixes #3386


ausi:
~~Is it worth changing that now that we are switching to Twig?~~ 
~~it is worth it because there are no twig templates yet for these.~~
switched opinion again 🙈

leofeyer:
IDK. #3386 was assigned to the "accepted for Contao 5" project, therefore I have made the changes. 😄

m-vo:
I think we assigned these things before the strategy for the new content elements had settled. There are more related things as well that we already refactored (like the odd/even information, renaming content element classes + templates, …). Maybe we should rethink this, though.

aschempp:
I think we should still change that, but we should include a migration to update the respective modules/places 🙃

leofeyer:
As discussed in the Contao call, we do not want to change this anymore, because in the course of Contao 5 all templates will be replaced by Twig templates anyway.

PR #4819, opened 2022-06-10
Deprecate CURRENT_ID

### LINKED ISSUE #4258: Ids stored in session will be changed in checkPermission in dca picker for non-admins
### Affected version(s)

4.9 and 4.13

### Description

In short the problem has to do with the approach that the currently used `id`s for some action are stored in `$session['CURRENT']['IDS']` and the `checkPermission` callback also works with and changes this data if we are in the dca picker for non admin users to check permissions for an other table.

I produce the problem in our contao installation as it would be very tedious to set the requirements for this bug without manually altering the `id`s of some entries in the database.

A simple example, even though I think this is a bigger problem with all dcas that have a callback that in some way changes  `$session['CURRENT']['IDS']` and can be accessed via the dca picker:

We assume we have 2 forms set up with `id`s `101` and `102`. Just to be clear they both have no `jumpTo` set. **And all** page `id`s are below 100. Could also be incresed to `100001` and `100002` and page `id`s below `100000`.

- So we, as a non admin user, go to `Forms` -> `Edit multiple` -> select these two forms and chose `Override` .
- Now internally contao stores those `id`s in `$session['CURRENT']['IDS']` 
- As Field we pick `Redirect page [jumpTo]`.
- So far so good. Now if we open the dca picker with `Change selection` and see the page tree. Now the damage has already been done. As for non admin users this has called the `checkPermission` callback.
https://github.com/contao/contao/blob/3194ba1cc082aa3c7145a85c3f353fbcdcccfa7f/core-bundle/src/Resources/contao/dca/tl_page.php#L802
In this callback the code is working with `$session['CURRENT']['IDS']` thinking it has to check if these `id`s are allowed for this user according to page permissions. And as these pages in our example do not exist they will be filtered out
https://github.com/contao/contao/blob/3194ba1cc082aa3c7145a85c3f353fbcdcccfa7f/core-bundle/src/Resources/contao/dca/tl_page.php#L838-L845
and overriden 
https://github.com/contao/contao/blob/3194ba1cc082aa3c7145a85c3f353fbcdcccfa7f/core-bundle/src/Resources/contao/dca/tl_page.php#L861
Just to state this absolutely clear: `$session['CURRENT']['IDS']` was filled with **`id`s from `tl_form`** and will be **checked against** permissions for the table **`tl_page`**
- We choose page with `id` `1` as `jumpTo` and `Apply`.
- Click `Save and close` to exit the `override multiple` action
- If we now check the `jumpTo` in the forms nothing has changed as `$session['CURRENT']['IDS']` was empty while performing the final override.

I've only tested it for the `tl_page` in dca picker and the actions override and edit, but I think this will also cause problems with other tables that are rendered in a dca picker.

ausi:
This should also fix several `CURRENT_ID` related issues like working with multiple browser tabs, opening a picker and so on.

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -2,9 +2,9 @@
 
 ## CURRENT_ID
 
-The constant `CURRENT_ID` as well as the session variable `CURRENT_ID` has been
-deprecated and will be removed in Contao 5.0. Use `DataContainer::$currentPid`
-instead to determine the ID of the current parent record.
+The `CURRENT_ID` constant and session variable have been deprecated and will be
+removed in Contao 5.0. Use `DataContainer::$currentPid` instead to determine
+the ID of the current parent record.
 
 ```php
 $intCurrentParentRecordId = $dc->currentPid;
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -3,8 +3,12 @@
 ## CURRENT_ID
 
 The constant `CURRENT_ID` as well as the session variable `CURRENT_ID` has been
-deprecated and will be removed in Contao 5.0. Use the current request to
-determine the ID of the current or the parent record.
+deprecated and will be removed in Contao 5.0. Use `DataContainer::$currentPid`
+instead to determine the ID of the current parent record.
+
+```php
+$intCurrentParentRecordId = $dc->currentPid;
+```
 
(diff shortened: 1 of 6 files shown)


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4820, opened 2022-06-10
Remove CURRENT_ID constant and session

ausi:
Followup to #4819

This removes `CURRENT_ID` completely and replaces it with `$dc->currentPid`

I named it `currentPid` because that’s IMO a way better description of what this value holds: The ID of the parent record if there is one.

The ID of the current entry itself can always be taken from `$dc->id`.


Toflar:
Isn't that just `$dc->activeRecord->pid`?

ausi:
No, in most places `currentPid` is used, there is no active record.

Toflar:
I guess my question was more "Why do we need yet another variable here?" :)

ausi:
> Why do we need yet another variable here?

It is not another variable IMO, but the constant `CURRENT_ID` moved to the correct place (a property on the data container).

You can see where this value is needed by checking the 36 occurences of `->intCurrentPid` and `‌->currentPid`.
Many of them can possibly be refactored to not needing this variable at all, but that is very hard IMO and would take me tremendously long to achieve.

Toflar:
Sure, I was just wondering what it exactly stands for. Because if I read "currentPid" then this means "the PID of the current record" to me. And that's what `$dc->activeRecord->pid` is. So I don't know why we need two different variables.

ausi:
> And that's what `$dc->activeRecord->pid` is. So I don't know why we need two different variables.

In most places where `currentPid` is used, there is no active record.

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -231,6 +232,58 @@ public function __construct($strTable, $arrModule=array())
...
+			// Mode “paste after”
+			$id = $pid;
+		}
+
+		if (!$id || !$this->Database->fieldExists('pid', $this->strTable))
+		{
+			return null;
+		}
+
+		$objPid = $this->Database->prepare("SELECT pid FROM `$this->strTable` WHERE id=?")
```
aschempp, line 280: This should probably be updated to `$this->getCurrentRecord($id)` once https://github.com/contao/contao/pull/4770 is merged, so the record is already cached if used later 🙃

leofeyer: See https://github.com/contao/contao/pull/4892.

PR #4824, opened 2022-06-13
Add getBackendWildcard to AbstractContentElementController

bezin:
I would love to see this in the Core bundle. For e.g. my list / detail controllers I do not need the reusability provided by modules. In this situation I rather dislike the way they are configured in a different place. So I prefer content elements over modules more and more. Yet for list and detail controllers I just want to show a placeholder in the backend which IMHO could just look like the module placeholder. This is what this PR implements.

### DIFF
--- a/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/core-bundle/src/Controller/ContentElement/AbstractContentElementController.php
@@ -97,5 +97,18 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
         ]);
     }
 
+    protected function getBackendWildcard(ContentModel $element): Response
+    {
+        $context = [
+            'id' => (int) $element->id,
+            'name' => $element->name,
+            'type' => $element->type,
+            'title' => StringUtil::deserialize($element->headline, true)['value'] ?? null,
+            'request_token' => $this->container->get('contao.csrf.token_manager')->getDefaultTokenValue(),
+        ];
+
+        return $this->render('@Contao/backend/content_wildcard.html.twig', $context);
+    }
+
     abstract protected function getResponse(FragmentTemplate $template, ContentModel $model, Request $request): Response;
 }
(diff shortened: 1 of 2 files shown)


bezin:
I just read in https://github.com/contao/contao/pull/4671 that it is handled differently anyways in the future. Don't know if we still want to add this as the legacy content elements are still part of the core?

m-vo:
> Don't know if we still want to add this as the legacy content elements are still part of the core?

Not sure what you mean - the legacy content elements never show a wildcard in the back end. But maybe I'm misunderstanding your idea.

For new elements, you could include the existing wildcard template (the new one) when in editor view in the template. We could make it easier to reuse if needed. But overall this should IMHO be a decision of the template, not the controller.

fritzmg:
> Not sure what you mean - the legacy content elements never show a wildcard in the back end. But maybe I'm misunderstanding your idea.

I think @bezin means modern content elements with legacy templates.

> For new elements, you could include the existing wildcard template (the new one) when in editor view in the template. We could make it easier to reuse if needed. But overall this should IMHO be a decision of the template, not the controller.

That's what this PR is about - it would make it easer to use the wildcard template, otherwise you'd always have to do it yourself, i.e. always inject the scope matcher, check for the scope etc.

m-vo:
> I think @bezin means modern content elements with legacy templates.

That is not possible. The new content elements are Twig-only. :upside_down_face: 

>  otherwise you'd always have to do it yourself, i.e. always inject the scope matcher, check for the scope etc.

You can already use `as_editor_view` and the context already contains the `id`, `name`, `type` and `title`. The `request_token` is now also always available with #4726. So there would never be a need to call `getBackendWildcard()`. If we want this,  we should IMHO rather make the wildcard a component and make it easily usable from your templates. For this to work without the need for `{% with %}` blocks around, we could adjust how it's currently done for modules.

fritzmg:
> That is not possible. The new content elements are Twig-only. 🙃

You mean the core content elements. But (I think) we are talking about custom content elements here - i.e.: https://docs.contao.org/dev/framework/content-elements/#example

m-vo:
Ah got you. Since this is for 5.x I would not add more features for these tbh.

bezin:
@fritzmg got it right, I was talking about Custom Content Elements.

But I get your point @m-vo that it should be handled in the template and when as_editor_view is automatically provided in the template than I am fine with that solution. The only thing I dislike about this is that I know have to wrap my whole template in an `if (asEditorView)-else-endif` statement. 

This PR probably would have had to be targetted against 4.x.

PR #4826, opened 2022-06-13
Throw correct exceptions on error in DC

aschempp:
Stuff noticed in https://github.com/contao/contao/pull/4770

Be aware that none of these exception should ever happen if the UI is used correctly, only when manipulating URLs.


Toflar (dismissed):
Not sure why we need our own exceptions here but it's good practice anyway 👍

aschempp:
Yeah just because good practice 🙃

core-bundle/src/Exception/BadRequestException.php
```diff
@@ -0,0 +1,19 @@
...
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Exception;
+
+use Symfony\Component\HttpFoundation\Exception\BadRequestException as BaseBadRequestExceptionAlias;
+
+class BadRequestException extends BaseBadRequestExceptionAlias
```
ausi:
```suggestion
use Symfony\Component\HttpFoundation\Exception\BadRequestException as BaseBadRequestException;

class BadRequestException extends BaseBadRequestException
```

I don’t think we use an `Alias` suffix anywhere else for such `use` statements.
Just a nitpick though. ☺️

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -658,13 +658,13 @@ public function create()
...
 	 * @throws AccessDeniedException
-	 * @throws InternalServerErrorException
+	 * @throws BadRequestException
 	 */
 	public function cut($source=null)
 	{
 		if ($GLOBALS['TL_DCA'][$this->strTable]['config']['notSortable'] ?? null)
 		{
-			throw new InternalServerErrorException('Table "' . $this->strTable . '" is not sortable.');
+			throw new AccessDeniedException('Table "' . $this->strTable . '" is not sortable.');
```
ausi: Wouldn’t a `BadRequestException` be a better fit here?
There is no way to gain access, even an admin user would get this exception so “access denied” seems wrong to me.

leofeyer: Probably also 422? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

aschempp: To me, the `notEditable` and `notSortable` are permission settings. We often set them in the `checkPermission` callbacks. And we might migrate them to voters later on.

ausi: I’d prefer to check then if `notSortable` was set as a permission, or as a configuration (that also affects admins).

But this can also be reworked afterwards IMO, so from my point of view my comment can be marked as resolved here.

aschempp:
making this a draft until https://github.com/contao/contao/pull/4770 is merged

leofeyer (changes requested):
I like the PR, except that we are misusing the `BadRequestException`, which is supposed to be used when the request syntax is incorrect, the request message is invalid, or the request routing is deceptive.

However, in our cases, e.g. when there is a sort request for an non-sortable table, it is rather a 422 Unprocessable Entity error, isn‘t it?

aschempp:
imho it's a bad request because you requested to sort a table that cannot be sorted. The request syntax is incorrect because you requested an invalid table / invalid operation on that table.

Toflar:
As discussed in our Contao Slack Call on July 28, 2022, we decided that 422 is the correct HTTP status code.

Toflar (approved):

ausi (approved):

leofeyer (approved):

PR #4834, opened 2022-06-15
Validate start/stop date

### LINKED ISSUE #812: Validate stop date > start date

aschempp:
Fixes https://github.com/contao/contao/issues/812


Toflar (dismissed):
LGTM. Any reason this is a draft?

aschempp:
Nope, just the stupid GitHub defaults 🤦‍♂️

fritzmg (dismissed):

Toflar (dismissed):

leofeyer (changes requested):
The changes are fine, however, there is a usability problem. After entering a stop date that is before the start date, the back end scrolled down to the input fields again, making it look like everything went fine:



Only if I scroll up, I notice that there has actually been an error:



To improve this, we could either auto-focus the error message on top of the page or we could make the widgets available in the `onbeforesubmit_callback`, so I can call `->addError()` on the "stop date" widget. (We then would have to move the `$this->noReload` check after the `onbeforesubmit_callback`.)

What do you think?

aschempp:
The widgets are rendered in `row`, making them available would not render an error message. I tried that first 😂

However, we can easily fix the scrolling, it already checks if an error exists. Should probably be a separate PR though?

leofeyer:
Since the invisible error message is a consequence of the changes in this pull request, I think the fix should also be in this pull request.

leofeyer:
#4846 has to be merged first.

aschempp:
> #4846 has to be merged first.

Why? #4846 improves the widget error "possibilities", but everything here still works as is? This check method could be improved in #4846 to output the error to the widget, then we'd also have a test case for #4846

leofeyer (approved):

Issue #4502, opened 2022-04-12
Infinite loop while loading of languages


rabauss:
### Affected version(s)

4.13.2

### Description

Unfortunately, my fix from https://github.com/contao/contao/pull/4254 does not fully solve the infinite loop while loading of languages. And I'm a little bit afraid to release our big update to Contao 4.13 in one of our customer projects 😨 

Today I had a scenario, that something was broken in the cache in local enviroment and so I got stuck in a loop again.
For now I cannot reproduce my broken cache but it was endless switching between the `de` and `de_DE` language. Although I only wanted to load the `de_DE`!

Can we somehow prevent the recursion while calling `System::getContainer()->get('contao.intl.locales')->getLocales()` here? https://github.com/contao/contao/blob/b986083e001cfe4ce79f4ff4b2e5fc9858824a91/core-bundle/src/Resources/contao/library/Contao/System.php#L500

ausi:
@rabauss Is there a `$GLOBALS['TL_HOOKS']['getLanguages']` hook used in your project?

ausi:
Maybe using something like the following code could prevent such endless loops?

```php
$fromLocalesClass = false;

foreach (debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS) as $trace) {
	if (($trace['class'] ?? null) === Locales::class) {
		$fromLocalesClass = true;
		break;
	}
}

if (!$fromLocalesClass) {
	foreach (self::getContainer()->get('contao.intl.locales')->getLocales($strCacheKey) as $strLocale => $strLabel)
	{
		$GLOBALS['TL_LANG']['LNG'][$strLocale] = $strLabel;
	}
}
```

rabauss:
@ausi: Yes, the hook `getLanguages` is used for custom sorting + filtering of the languages!

ausi:
Can you check if the error disappears when disabling the hook?

rabauss:
I'm not sure how to test it - because as I said, I can't reproduce my broken cache.
However, the loop can always occur if there was no previous successful return from `loadLanguageFile` - here: https://github.com/contao/contao/blob/b986083e001cfe4ce79f4ff4b2e5fc9858824a91/core-bundle/src/Resources/contao/library/Contao/System.php#L460

PR #4836, opened 2022-06-15
Fix infinite loop while loading of countries

rabauss:
Relates to #4254 

While loading of countries we get a infinite loop like described in #4254
The problem occurs if you use a non 2-letter language in the frontend.

I'm still not really happy with the solution because there could be again a loop if there is something broken with the cache - see: https://github.com/contao/contao/issues/4502 But maybe the broken cache was also related to this problem while loading of the countries.


ausi (approved):

leofeyer:
Thank you @rabauss.

PR #4839, opened 2022-06-16
Enable typed statements

aschempp:
This PR makes it easier to add type information to our database statements. See https://github.com/contao/contao/pull/4797/files/9df8cf2d71eaa253e1845fda426c10e171a25c9e#r897680666 for why this would an improvement.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
@@ -256,7 +256,8 @@ class Statement
 			throw new \Exception('Empty query string');
 		}
 
-		$arrParams = array_merge($this->arrSetParams, $arrParams);
+		$arrTypes = array_map(static fn ($key) => $arrTypes[$key] ?? null, array_keys($arrParams));
+		$arrParams = array_merge($this->arrSetParams, array_values($arrParams));
 		$arrTypes = array_merge($this->arrSetTypes, $arrTypes);
 
 		$arrParams = array_map(
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
@@ -60,6 +60,11 @@ class Statement
 	/**
 	 * @var array
 	 */
+	private $arrSetTypes = array();
+
+	/**
+	 * @var array
+	 */
 	private $arrLastUsedParams = array();
 
 	/**
@@ -138,7 +143,7 @@ class Statement
 	 *
 	 * @return Statement The statement object
 	 */
-	public function set($arrParams)
+	public function set($arrParams, $arrTypes = array())
 	{
 		if (substr_count((string) $this->strQuery, '%s') !== 1 || !\in_array(strtoupper(substr($this->strQuery, 0, 6)), array('INSERT', 'UPDATE'), true))
 		{
@@ -146,6 +151,7 @@ class Statement
 		}
 
 		$this->arrSetParams = array_values($arrParams);
+		$this->arrSetTypes = array_map(static fn ($key) => $arrTypes[$key] ?? null, array_keys($arrParams));
 
 		$arrParamNames = array_map(
 			static function ($strName)
@@ -225,7 +231,7 @@ class Statement
 	 */
 	public function execute()
 	{
-		return $this->query('', array_merge($this->arrSetParams, \func_get_args()));
+		return $this->query('', \func_get_args());
 	}
 
 	/**
@@ -249,6 +255,9 @@ class Statement
 		{
 			throw new \Exception('Empty query string');
 		}
+
+		$arrParams = array_merge($this->arrSetParams, $arrParams);
+		$arrTypes = array_merge($this->arrSetTypes, $arrTypes);
 
 		$arrParams = array_map(
 			static function ($varParam)


fritzmg:
An alternative would be to use named parameters instead. See https://github.com/contao/contao/pull/4797/commits/e2620c8e8da3943199686202359f406bcdde7ed9 for example (which was once part of my #4797 PR).

aschempp:
how is that related?

fritzmg:
In your PR you are filling the types array with null values for each parameter that did not have a type set. In my approach I used named parameters instead. Currently `Statement::set` only saves the parameter values:

```php
$this->arrSetParams = array_values($arrParams);
```

In my approach that would change to

```php
$this->arrSetParams = $arrParams;
```

and then later on the query would use named parameters:

```diff
-$strQuery = 'SET ' . implode('=?, ', $arrParamNames) . '=?';
+$strQuery = 'SET ' . implode(', ', array_map(function($key) { return $key.'=:'.$key; }, array_keys($arrParams)));
```

Thus if you would pass

```php
['foobar' => ParameterTypes::BOOLEAN]
```

for `$arrTypes` it would automatically work.

Toflar:
Why do we need new features in a library that's going to be replaced by Doctrine completely? Can't you just use Doctrine already in your case?

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -225,7 +231,7 @@ public function limit($intRows, $intOffset=0)
 	 */
 	public function execute()
 	{
-		return $this->query('', array_merge($this->arrSetParams, \func_get_args()));
+		return $this->query('', \func_get_args());
```
ausi, line 234: This cannot be removed here, as it would break how `query()` works currently.

Instead the types have to be passed here too.

```suggestion
 		return $this->query('', array_merge($this->arrSetParams, \func_get_args()), $this->arrSetTypes);
```

core-bundle/src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -250,6 +256,10 @@ public function query($strQuery='', array $arrParams = array(), array $arrTypes
 			throw new \Exception('Empty query string');
 		}
 
+		$arrTypes = array_map(static fn ($key) => $arrTypes[$key] ?? null, array_keys($arrParams));
+		$arrParams = array_merge($this->arrSetParams, array_values($arrParams));
+		$arrTypes = array_merge($this->arrSetTypes, $arrTypes);
+
```
ausi, line 262:
```suggestion
```

see https://github.com/contao/contao/pull/4839#discussion_r899051211

ausi (changes requested):

aschempp:
> In your PR you are filling the types array with null values for each parameter that did not have a type set. In my approach I used named parameters instead. 

You cannot mix named and unnamed parameters in Doctrine, therefore I don't think we can use named queries at all in our database class.

> Why do we need new features in a library that's going to be replaced by Doctrine completely? Can't you just use Doctrine already in your case?

Because Doctrine does not offer a `->set()` method, and because our models use our database class, and our models now include type information.

> This cannot be removed here, as it would break how `query()` works currently.

The only change is that `->query()` now includes `->set()` parameters, which was not the case before. But `->query()` already uses the build-in query (if the parameter is empty). I think this _behaviour change_ is reasonable and possible in a new major version.

fritzmg:
> You cannot mix named and unnamed parameters in Doctrine, therefore I don't think we can use named queries at all in our database class.

We can in the `set` method.

> Because Doctrine does not offer a ->set() method, and because our models use our database class, and our models now include type information.

But you could use `->update` and `->insert`.

ausi:
> > This cannot be removed here, as it would break how `query()` works currently.
> 
> The only change is that `->query()` now includes `->set()` parameters, which was not the case before. But `->query()` already uses the build-in query (if the parameter is empty). I think this _behaviour change_ is reasonable and possible in a new major version.

The following code wouldn’t work anymore I think:

```php
$statement = $database->prepare("UPDATE foo %s WHERE id=1")->set($data)->execute();
$statement->query('SELECT 1');
```

I don’t see why we should change the API here, what is the benefit?
Why should I expect that `->query('SELECT 1')` now magically tries to use some internal state that it never has before.

aschempp:
> The following code wouldn’t work anymore I think:
> 
> ```
> $statement = $database->prepare("UPDATE foo %s WHERE id=1")->set($data)->execute();
> $statement->query('SELECT 1');
> ```

Yeah, but why would it be a supported use case to re-run a statement with a different query?

> I don’t see why we should change the API here, what is the benefit? Why should I expect that `->query('SELECT 1')` now magically tries to use some internal state that it never has before.

Because the statement has internal state? If you want to run a new query, create a new statement? The API change is necessary because only `query()` (currently) allows to include type information.

ausi:
> Yeah, but why would it be a supported use case to re-run a statement with a different query?

Because that is the API of that class.

> The API change is necessary because only `query()` (currently) allows to include type information.

You should create a new method then `executeTyped(array $params, array $types)` or something in that direction.

aschempp:
> You should create a new method then `executeTyped(array $params, array $types)` or something in that direction.

That would work as well. And would be better than calling `query()` in probably all cases 👍 😆

PR #4840, opened 2022-06-16
Add class to back end login

netzarbeiter:
I have added two classes to the login templates.

If I for example add the following lines in the config.yml:

```yml
contao:
  backend:
    custom_css:
      - files/backend/backend.min.css
```

and want to change CSS on the `#container` it also affects the backend.

With the new classes you can add `.be_login #container` to only change the behaviour on the login screen.


Toflar (dismissed):
Looks okay to me. The ending line feeds in the templates must not be adjusted but that can be fixed as coding style by @leofeyer :)

leofeyer:
I think the PR needs to be rebased because `contao.backend.custom_css` is not available in Contao 4.9, is it?

Toflar:
How does this matter? One could have added their custom CSS with `$GLOBALS['TL_CSS']` ever since :)

leofeyer (changes requested):
The pull request needs to be rebased onto the correct branch.

ausi:
> The pull request needs to be rebased onto the correct branch.

Couldn’t this also be seen as a bugfix? Because it is currently not possible to distinguish between backend and login?

Toflar:
Yeah, imho that's a bugfix too.

aschempp (dismissed):
I think we should merge this into 4.9. I always had a hard time to apply CSS to the login screen and the back end, even in 4.9. I had to load a CSS based on the current route, which sure can be improved with this.

leofeyer (changes requested):
As briefly discussed in Slack, this can be seen as a bugfix. In order to merge the PR, however, the linefeed changes have to be reverted and @fritzmg‘s suggestion needs to be addressed:

> Der Vollständigkeit halber bräuchte aber dann auch das `be_main` Template die `be_main` Klasse (oder überall `$this->getName()`).

ausi (approved):
❤️

leofeyer:
Thank you @netzarbeiter.

leofeyer (approved):

PR #4846, opened 2022-06-18
Display error message near the affected widget

ausi:
This builds on #4834 and displays the error message near the affected widget.
This is just a first draft to see how something like this can work.

I think this is an important usability feature.

### DIFF
--- a/calendar-bundle/src/Resources/config/services.yaml
+++ b/calendar-bundle/src/Resources/config/services.yaml
@@ -12,6 +12,13 @@ services:
         arguments:
             - '@contao.framework'
 
+    contao_calendar.listener.data_container.start_stop_validation:
+        class: Contao\CoreBundle\EventListener\DataContainer\StartStopValidationListener
+        arguments:
+            - '@translator'
+        tags:
+            - { name: contao.callback, table: tl_calendar_events, target: config.onbeforesubmit }
+
     contao_calendar.listener.generate_page:
         class: Contao\CalendarBundle\EventListener\GeneratePageListener
         arguments:
--- a/core-bundle/src/EventListener/DataContainer/StartStopValidationListener.php
+++ b/core-bundle/src/EventListener/DataContainer/StartStopValidationListener.php
@@ -23,8 +23,8 @@ class StartStopValidationListener
 
     public function __invoke(array $values, DataContainer $dc): array
     {
-        $start = (string) (array_key_exists('start', $values) ? $values['start'] : $dc->activeRecord->start);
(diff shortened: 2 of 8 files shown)


ausi:
Now the widgets get rendered after the submit, so errors can now be properly set on the widgets themself.

Bildschirmfoto 2022-06-19 um 15 04 22

m-vo:
I did not have time to tinker with this PR, just throwing in a random thought:

Can we maybe reuse how Symfony handles violations? There you typically have a builder that collects violations in a `ViolationList` and has a nice API (see [ConstraintViolationBuilderInterface](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Validator/Violation/ConstraintViolationBuilderInterface.php) for instance). But not sure if this applies to our use case.  :relaxed:

aschempp (commented):
I like the idea of the lazy rendering, feels clean an simple. However, this will only work in edit mode, it will not work in editAll or overrideAll, because the field names will be different. Also, `propertyPath` seems a complex name for something that is actually meant to be "field name"?

On a side note regarding editAll, maybe we could dynamically map the field names from the exception to the current ones in the loop? Because one can't really know the _current_ field name in the editAll loop

ausi:
> On a side note regarding editAll, maybe we could dynamically map the field names from the exception to the current ones in the loop?

👍

> Also, `propertyPath` seems a complex name for something that is actually meant to be "field name"?

This is borrowed from symfony, not sure if “field name” is better, I’d like to make this usable for fieldsets too, to make it possible to add an error message to the top of a fieldset.

In the data container, the items of a palette seem to be called “boxes”, so how about `boxNames`?

aschempp:
> This is borrowed from symfony, not sure if “field name” is better, I’d like to make this usable for fieldsets too, to make it possible to add an error message to the top of a fieldset.

What fieldsets do you see in the Contao back end? A checkbox wizard using a fieldset for its options still has one "field name" for the overall widget. Or do you mean palettes?

> In the data container, the items of a palette seem to be called “boxes”, so how about `boxNames`?

I think boxes is group of fields in a legend inside a palette. Do you intend to add an error to a legend (title)?

ausi:
> > In the data container, the items of a palette seem to be called “boxes”, so how about `boxNames`?
> 
> I think boxes is group of fields in a legend inside a palette. Do you intend to add an error to a legend (title)?

Yes, “legend” is what I meant with fieldset ☺️

So that I can pin an error either to `{text_legend}` or to `text`.

But I think @m-vo is right in https://github.com/contao/contao/pull/4846#issuecomment-1159807076 we also need a way to add multiple errors, with the current solution I only ever get the first error. Seeing all validation errors would be way better UX I think.

PR #4847, opened 2022-06-20
Add a new feed reader implementation

### LINKED ISSUE #4689: Drop simplepie/simplepie in favor of debril/feedio
### Description

While working on #4682 I stumbled upon the RssReaderModule. It uses the simplepie/simplepie library to read and parse feeds. As the new Feed page controllers will use debril/feedio (because it supports reading and writing out of the box), we could refactor the RssReaderModule and drop SimplePie.

bezin:
This implements a new Feed Reader as front end module fragment to replace the legacy RSS Reader module eventually. 

It uses `debril/feed-io` rather than `simplepie` to fix #4689 and I thought why not add it as a new fragment controller while I am at it. 

This Feed Reader now also parses JSON feeds as well as RSS and atom feeds. 

* [x] Unit tests


core-bundle/composer.json
```diff
@@ -65,6 +65,7 @@
         "doctrine/orm": "^2.10",
         "doctrine/persistence": "^2.0",
         "dragonmantank/cron-expression": "^2.3",
+        "debril/feed-io": "^6.0",
```
m-vo, line 69: You also need to include this change in the monorepo `composer.json`. You can use the `tools/monorepo/vendor/bin/monorepo-tools composer-json` command to automatically make the changes for you.

bezin: Merci ❤️

bezin:
I added basic unit tests (fetch feeds from remote server and fetch from cache), but struggle to correctly test the pagination logic. Anyways happy to hear your comments so far.

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
```diff
@@ -0,0 +1,68 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Migration\Version500;
+
+use Contao\CoreBundle\Migration\AbstractMigration;
+use Contao\CoreBundle\Migration\MigrationResult;
+use Doctrine\DBAL\Connection;
+
+class RssToFeedReaderMigration extends AbstractMigration
```
aschempp: Feels like this whole migration could be simplified by renaming the columns if the old ones exist and the new ones do not?

core-bundle/src/Resources/config/controller.yaml
```diff
@@ -99,6 +99,12 @@ services:
 
     Contao\CoreBundle\Controller\FrontendModule\RootPageDependentModulesController: ~
 
+    Contao\CoreBundle\Controller\FrontendModule\FeedReaderController:
+        arguments:
+            - '@contao.feed.feedio'
+            - '@contao.feed.null_logger'
```
aschempp: any reason not to use our real logger?

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,106 @@
...
+
+                try {
+                    $this->feedIo->read($url, $feed);
+
+                    if ($model->feed_cache > 0) {
+                        $item->expiresAfter($model->feed_cache);
+                    }
+                    $item->set($feed);
+                } catch (\Exception $exception) {
+                    $this->logger->error(sprintf('Could not read feed %s: %s', $url, $exception->getMessage()));
```
aschempp: should this somehow cancel the feed (`return`)? Looks like `$feed` might not exist?

core-bundle/src/Resources/contao/languages/en/modules.xlf
```diff
@@ -305,6 +305,12 @@
       <trans-unit id="FMD.rssReader.1">
         <source>Adds an RSS feed to the page</source>
       </trans-unit>
+      <trans-unit id="FMD.feed_reader.0">
```
aschempp: should remove `rssReader` a few lines above?

core-bundle/src/Resources/contao/languages/en/tl_module.xlf
```diff
@@ -260,22 +260,22 @@
...
-        <source>Please enter the URL of one or more RSS feeds.</source>
+      <trans-unit id="tl_module.feed_urls.1">
+        <source>Please enter the URL of one or more feed (RSS, Atom or JSON).</source>
       </trans-unit>
-      <trans-unit id="tl_module.rss_template.0">
+      <trans-unit id="tl_module.feed_template.0">
         <source>Feed template</source>
       </trans-unit>
-      <trans-unit id="tl_module.rss_template.1">
+      <trans-unit id="tl_module.feed_template.1">
```
aschempp: there's no `feed_template` anymore

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,106 @@
...
+
+            $feeds[] = $feed;
+        }
+
+        $allItems = [];
+
+        foreach ($feeds as $feed) {
+            $feedItems = \array_slice([...$feed], $model->skipFirst, $model->numberOfItems ?: null);
+            $allItems = [...$feedItems, ...$allItems];
+        }
```
aschempp: I'm not sure this is the same behaviour as previously. This would take the first X items of each feed, whereas it previously took the first items from the SimplePie reader. I'm unsure how SimplePie behaves, but I would e.g. expect it to sort the feed items by date or something like that before limiting the overall items to X?

core-bundle/src/Resources/contao/templates/rss/rss_items_only.html5
```diff
@@ -3,7 +3,7 @@
 <?php $this->block('content'); ?>
 
   <?php foreach ($this->items as $item): ?>
-    <div class="rss_items_only<?= $item['class'] ?>">
+    <div class="rss_items_only<?= $item['class'] ?? '' ?>">
```
aschempp: the `rss_...` templates are not used anymore, right?

aschempp (commented):
Nice! Not sure the changes to the pagination belong to this PR (or what they are necessary for, or if we need them or should use Pagerfanta or thelike for Twig).

core-bundle/src/Resources/contao/languages/en/modules.xlf
bezin: Fixed in 243491c5f

core-bundle/src/Resources/contao/languages/en/tl_module.xlf
bezin: Fixed in 243491c5f

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: Yes, you are right. There seems to be no real way to cancel from within the callback, so I moved the try/catch outside of it and handle exceptions there (see a1821ec7b)

core-bundle/src/Resources/contao/templates/rss/rss_items_only.html5
bezin: Yes, these are obsolete!

bezin: Fixed in 0a29f8a8d

core-bundle/src/Resources/config/controller.yaml
bezin: I wanted to silence the internals and only log exceptions catched at controller level. But on the other hand, if I log the internals, I could leave out logging the error in the controller.

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
bezin: Fixed

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: The first x items from the SimpleReader [where internally handled per feed](https://github.com/simplepie/simplepie/blob/master/src/SimplePie.php#L3266-L3276). So this is the same as here. But you are right I missed the sorting by date.

bezin: Added sorting in 7c0f8bd

bezin:
> Nice! Not sure the changes to the pagination belong to this PR (or what they are necessary for, or if we need them or should use Pagerfanta or thelike for Twig).

The changes where necessary to render the pagination in a separate Twig component. The `generate` method would always return a string. I added another method, that compiles everything and than returns it as an array. But I also thought about Pagefanta – I wonder, if moving this to the template is the right apporach though?

aschempp:
>  I wonder, if moving this to the template is the right apporach though?

If possible and for the sake of simplicity of this PR, I would keep it as-is and keep the existing pagination. If we decide to go with pagerfanta or anything thelike later, there's a lot of other places to refactor as well anyway 🙃

core-bundle/src/Resources/config/controller.yaml
aschempp: I don't know how the feed logger works or what level it logs to. If it generates a lot of noise for every page view and is only meant to be set for debugging purposes, maybe we should keep the `null_logger`.

bezin:
I added unit tests for the pagination and rebased on recent `5.x` locally, but ran into same issues. Almost all is fixed, but not all (Need to refactor some static calls to adapter calls in `Contao\Pagination` in order to properly test the controller). Will finish on Monday.

Regarding CI: PHPStan is wrong here, don't know why it comes up with this?

bezin:
All done for now from my end. As said, PHPStan is wrong here.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,132 @@
...
+
+        uasort($allItems, [$this, 'sortItems']);
+
+        $offset = 0;
+        $limit = \count($allItems);
+
+        if ($model->perPage > 0) {
+            $param = 'page_r'.$model->id;
+            $page = (int) $request->query->get($param, 1);
```
aschempp:
```suggestion
            $page = $request->query->getInt($param, 1);
```

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,132 @@
...
+
+            $pagination = new Pagination(\count($allItems), $model->perPage, $config->get('maxPaginationLinks'), $param);
+            $template->set('pagination', $pagination->getContextForTwigComponent($request));
+        }
+
+        $items = [];
+
+        for ($i = $offset, $c = \count($allItems); $i < $limit && $i < $c; ++$i) {
+            $items[] = $allItems[$i];
+        }
```
aschempp: Shouldn't this be the same?
```suggestion
        $items = array_slice($allItems, $offset, $limit);
```

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,132 @@
...
+        return $template->getResponse();
+    }
+
+    private function sortItems(Item $a, Item $b): int
+    {
+        $aDate = $a->getLastModified();
+        $bDate = $b->getLastModified();
+
+        if ($aDate && $bDate) {
+            return $aDate > $bDate ? -1 : 1;
```
aschempp:
```suggestion
            return $aDate <=> $bDate;
```

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+        $columns = $schemaManager->listTableColumns('tl_module');
```
aschempp: you should probably check if the table exists at all (e.g. on a fresh installation). Also make sure `feed_urls` and `feed_cache` do not exist.

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
```diff
@@ -0,0 +1,57 @@
...
+        );
+
+        $this->connection->update('tl_module', ['type' => 'feed_reader'], ['type' => 'rssReader']);
+
+        return $this->createResult(true);
+    }
+
+    private function renameField(string $fromQuoted, string $toQuoted, string $type): void
+    {
+        $tableQuoted = $this->connection->quoteIdentifier('tl_module');
```
aschempp: looks like you could quote the fields here as well instead of every time you call the method?

core-bundle/src/Resources/contao/languages/en/modules.xlf
```diff
@@ -299,11 +299,11 @@
...
       </trans-unit>
-      <trans-unit id="FMD.rssReader.0">
-        <source>RSS reader</source>
+      <trans-unit id="FMD.feed_reader.0">
+        <source>Feed reader</source>
       </trans-unit>
-      <trans-unit id="FMD.rssReader.1">
-        <source>Adds an RSS feed to the page</source>
+      <trans-unit id="FMD.feed_reader.1">
+        <source>Adds an RSS, Aom or JSON feed to the page</source>
```
aschempp:
```suggestion
        <source>Adds an RSS, Atom or JSON feed to the page</source>
```

aschempp (changes requested):
Very good! Just a few remaining changes. Requesting review from @m-vo for the Twig stuff 🙃

core-bundle/src/Resources/contao/library/Contao/Pagination.php
```diff
@@ -386,6 +422,11 @@ public function getItemsAsArray()
 		return $arrLinks;
 	}
 
+	public function getContextForTwigComponent(Request $request): ?array
```
Toflar: What's this? I mean, why does it have a parameter that's not used? Why can't it just call `->compile()` directly?

bezin: Yep, you are right, this is a left-over. Removed in [8653532](https://github.com/contao/contao/pull/4847/commits/8653532e703255ee582a929c9123a49b83caf521)

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: Fixed in [97503bb](https://github.com/contao/contao/pull/4847/commits/97503bb7c765f681a44912e05d5dc4d37fa99974)

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: Fixed in [97503bb](https://github.com/contao/contao/pull/4847/commits/97503bb7c765f681a44912e05d5dc4d37fa99974)

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: Fixed in [97503bb](https://github.com/contao/contao/pull/4847/commits/97503bb7c765f681a44912e05d5dc4d37fa99974)

core-bundle/src/Resources/contao/languages/en/modules.xlf
bezin: Fixed in [97503bb](https://github.com/contao/contao/pull/4847/commits/97503bb7c765f681a44912e05d5dc4d37fa99974)

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
bezin: Fixed in [03843d0](https://github.com/contao/contao/pull/4847/commits/03843d08211a783cc7297133702e5e55ae0ffdf3)

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
bezin: Fixed in [03843d0](https://github.com/contao/contao/pull/4847/commits/03843d08211a783cc7297133702e5e55ae0ffdf3)

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,128 @@
...
+        if ($aDate) {
+            return 1;
+        }
+
+        if ($bDate) {
+            return -1;
+        }
+
+        return 0;
+    }
```
m-vo:
```suggestion
```

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        $allItems = [];
+
+        foreach ($feeds as $feed) {
+            $feedItems = \array_slice([...$feed], $model->skipFirst, $model->numberOfItems ?: null);
+            $allItems = [...$feedItems, ...$allItems];
+        }
+
+        uasort($allItems, [$this, 'sortItems']);
```
m-vo:
```suggestion
        usort($allItems, static fn(Item $a, Item $b): int => $a->getLastModified() <=> $b->getLastModified());
```

Values of `null` are already sorted to the top with the spaceship operator, see https://3v4l.org/bm30V#v8.1.7.

Edit: Sorry for the confusion. Forgot the function in my previous suggestion.

m-vo (dismissed):

m-vo:
Nice!

I have some remarks about the Twig part. But I would instead go ahead and create a PR against your PR with some suggestions - would this be OK for you?

One thing we might want to discuss first is the pagination, though. I don't think we should introduce breaking changes in the `\Contao\Pagination` class and add a Twig component for this. If we add the latter, we should IMHO also bring in a new pagination library altogether and tidy things up in general (nothing for this PR, but maybe for 5.1). Any objections if I try to rework this bit? Or maybe you like to explain your thoughts behind these changes?

bezin:
Happy to see your suggestions!

I think the changes to the `Pagination` class are not breaking, see `PaginationTest`. But I had second thoughts about the changes, too. Maybe it is sufficient to use the old implementation and just pass the generated string to the twig template and `|raw` it there.

As you said, when we introduce a twig component we should rewrite the implementation, too.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,128 @@
...
+                $feeds[] = $feed;
+            }
+        }
+
+        $allItems = [];
+
+        foreach ($feeds as $feed) {
+            $feedItems = \array_slice([...$feed], $model->skipFirst, $model->numberOfItems ?: null);
+            $allItems = [...$feedItems, ...$allItems];
+        }
```
m-vo:
```suggestion
        $allItems = array_merge(
            ...array_map(
                static fn(array $feed) => \array_slice([...$feed], $model->skipFirst, $model->numberOfItems ?: null),
                $feeds
            )
        );
```

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,128 @@
...
+
+    protected function getResponse(FragmentTemplate $template, ModuleModel $model, Request $request): Response
+    {
+        $this->initializeContaoFramework();
+
+        $feeds = [];
+
+        foreach (StringUtil::trimsplit('[\n\t ]', trim($model->feed_urls)) as $url) {
+            try {
+                $feed = $this->cache->get('feed_reader_'.$model->id.'_'.md5($url), function (ItemInterface $item) use ($url, $model) {
```
m-vo: Why do we access the system cache here? Isn't it enough to cache the whole response with the HTTP Cache if needed?

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
aschempp: imho this change is unnecessary. It's the same logic written in a different personal style (and less readable to me personally). It does not improve anything 🤷

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
aschempp: that's how the RSS feed module works. It has a configuration for how long to cache the feeds, which is independent from how long the page is cached. I would refrain from discussing ESI options or thelike at the current stage of development (and Contao 5 release time).

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: I am still confused ;-) IMO this does not yield the same result than the previous implementation, see https://3v4l.org/Z41og#v8.1.7

m-vo: You want to sort items to the top if `getLastModified()` is null (was missing in my example :see_no_evil:). 

See https://3v4l.org/Kd5P2#v8.1.7. I also updated the suggestion.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
m-vo: It's one merge vs n merges - this is faster. It's also not less readable for me.
(see https://3v4l.org/9EZUZ#v8.1.7 for instance)

bezin: Implemented in 3c708f08c

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
bezin: Implemented in eb68c985f

aschempp (dismissed):
❤️

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
```diff
@@ -0,0 +1,55 @@
...
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function shouldRun(): bool
+    {
+        $schemaManager = $this->connection->createSchemaManager();
+        $columns = $schemaManager->listTableColumns('tl_module');
+
+        if (!$schemaManager->tablesExist(['tl_news_feed'])) {
```
leofeyer: Why this check for `tl_news_feed`? And why no check for `tl_module` above?

core-bundle/src/Resources/config/controller.yaml
leofeyer: This seems strange to me. If the logger is optional, it should be provided as last argument and allow `null`. Then, when you don‘t want to log, you simply don‘t pass a logger instead of having to pass a null logger.

core-bundle/src/Resources/contao/dca/tl_module.php
```diff
@@ -125,7 +125,7 @@
 		'html'                        => '{title_legend},name,type;{html_legend},html;{template_legend:hide},customTpl;{protected_legend:hide},protected',
 		'unfiltered_html'             => '{title_legend},name,type;{html_legend},unfilteredHtml;{template_legend:hide},customTpl;{protected_legend:hide},protected',
 		'template'                    => '{title_legend},name,headline,type;{template_legend},data,customTpl;{protected_legend:hide},protected;{expert_legend:hide},cssID',
-		'rssReader'                   => '{title_legend},name,headline,type;{config_legend},rss_feed,numberOfItems,perPage,skipFirst,rss_cache;{template_legend:hide},rss_template;{protected_legend:hide},protected;{expert_legend:hide},cssID',
+		'feed_reader'                 => '{title_legend},name,headline,type;{config_legend},feed_urls,numberOfItems,perPage,skipFirst,feed_cache;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},cssID',
```
leofeyer: The fields should be called `feedUrls` and `feedCache`. `rss_` was the old prefix, therefore it had an underscore.

core-bundle/src/Resources/contao/library/Contao/Pagination.php
```diff
@@ -217,11 +219,13 @@ public function generate($strSeparator=' ')
 			return '';
 		}
 
+		$environment = System::getContainer()->get('contao.framework')->getAdapter(Environment::class);
```
leofeyer: We should use the Symfony request here instead of the old `Environment` class.

leofeyer:
I feel like we should discuss the logging concept again.

core-bundle/src/Resources/contao/library/Contao/Pagination.php
bezin: I tried to touch this class as little as possible, actually only to make it testable in my context.

core-bundle/src/Migration/Version500/RssToFeedReaderMigration.php
bezin: Good catch, this is a mix-up between the two feed related PRs 👍

core-bundle/src/Resources/contao/dca/tl_module.php
bezin: Fixed in772aaabb5

core-bundle/src/Resources/config/controller.yaml
bezin: Ok, now I saw what was wrong here: I only wanted to silence the logging from the library. But the comented code snippet is regarding the controller 🙈 This is oviously wrong, I wanted the real logger here oc.

Fixed in 5706e65d7

bezin:
@m-vo provided some [thoughts and twig-related improvements](https://github.com/bezin/contao/pull/4). 

I transfer the open discussions to this PR: 

* Regarding HTML semantics: We went with `<article>`, `<header>` and `<section>` for the markup, but would appreciate some feedback from experts on that field. Our thinking was: We have headlines and blocks in there that have nothing in common with the rest of the document, so it needs to be a separate group. But it could probably also be treated as a list…
* Currently no sanitation of the incoming HTML is taking place (SimplePie handled it by it self before).  We'll probably need a |sanitize('html') or |sanitize_html filter for that. Pinging @ausi here to check if this is something we could still do for 5.0 or if that needs more thinking.
* We are now passing FeedIo objects to the template. On the one hand this is a very good thing, because you have a clear API with auto-completion and all that, on the other we are tying ourselves to this specific library. This basically means we cannot exchange it under the hood in the future without breaking existing templates. But I also think it's not really worth abstracting.
* We should ideally also add a test similar to the ones in tests/Controller/ContentElement that actually renders the module and verifies its output.

ausi:
> Currently no sanitation of the incoming HTML is taking place (SimplePie handled it by it self before). We'll probably need a |sanitize('html') or |sanitize_html filter for that. Pinging @ausi here to check if this is something we could still do for 5.0 or if that needs more thinking.

Would `Contao\Input::stripTags(…, $allowedTags, $allowedAttributes)` be enough for this use case?

ausi:
As discussed in the call today we should add a ~~`|sanitize('html')` or~~ `|sanitize_html` filter to this PR that calls `Contao\Input::stripTags()` with a fixed set of allowed tags and attributes.

In a later PR we could extend the sanitize filter with multiple configs that can be customized. We do not want to use the configuration for the backend users here because the source of the HTML is not a backend user.

ausi:
@bezin is this pull request completed apart from the `sanitize_html` filter?
~~Should I create a pull request on your pull request that adds the `sanitize_html` filter?~~ https://github.com/bezin/contao/pull/7

core-bundle/composer.json
```diff
@@ -94,13 +95,13 @@
         "scheb/2fa-trusted-device": "^6.0",
         "scrivo/highlight.php": "^9.18",
         "scssphp/scssphp": "^1.5",
-        "simplepie/simplepie": "^1.3",
```
ausi: If we remove this dependency, we have to remove the *contao/modules/ModuleRssReader.php* file too I think because it is still used there.
But for better BC we probably want to keep `ModuleRssReader` and therefore should keep `simplepie` too?
Alternatively we could add a class_exists check to `ModuleRssReader` and throw a meaningful exception if `simplepie` is missing.

ausi (changes requested):
*core-bundle/src/Resources/contao/templates/_new/frontend_module/…* needs to be moved to 
*core-bundle/contao/templates/_new/frontend_module/…*

core-bundle/composer.json
bezin: Yeah, we probably have to keep the module and its dependencies around for a while now. As much as I would like to remove both I'd rather leave it as is until 6.0.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
Toflar: Also, I keep saying this in all kinds of PRs and extensions: Just because we have an HTTP cache does not mean you shouldn't use any other cache anymore. The HTTP cache is just another layer on top and it can be disabled if your logged in (as FE or BE user) or working in DEV mode or preview mode etc. All these cases would cause live requests which are completely unnessecary. So this is definitely correct here.

core-bundle/composer.json
bezin: I gave this some more thought: I would have to reverse much more if we want to keep ModuleRssReader, esp. column names. Currently all uses of the legacy RSS reader will be migrated to the new FeedReader so there kinda is a smooth migration path available, IF you use the core legacy module and template 🤔 Still a break though.

Toflar: Honestly, the `ModuleRssReader` remains there now only for BC. And I'd argue that the use cases to extend from that were rather little. So I don't think there will be many cases out there. Mostly, people will just use that module and not even notice that the implementation in the background has changed :)
So I'd prefer to get rid of this dependency and add a `class_exists()` test in the `ModuleRssReader`. 
Also `ModuleRssReader` should  have a `trigger_deprecation()` call.

Toflar (commented):
Almost there!

core-bundle/composer.json
Toflar: Templates are not covered by the BC promise. And for the BC of the column names in `ModuleRssReader`: Maybe we can map them?

bezin: See 
8e60be8

core-bundle/contao/modules/ModuleRssReader.php
```diff
@@ -36,6 +36,13 @@ class ModuleRssReader extends Module
 	 */
 	public function generate()
 	{
+		trigger_deprecation('contao/core-bundle', '5.1', sprintf('The "%s" has been deprecated and will be removed in Contao 6. Use the Feed Reader Module instead.', __CLASS__));
+
+		if (!class_exists('\SimplePie'))
```
ausi:
```suggestion
		if (!class_exists('SimplePie'))
```

core-bundle/contao/dca/tl_module.php
```diff
@@ -406,15 +407,15 @@
 			'explanation'             => 'insertTags',
 			'sql'                     => "mediumtext NULL"
 		),
-		'rss_cache' => array
+		'feedCache' => array
```
ausi: Would be keeping the old column names `rss_cache` and `feedUrls` an option?
Would be better for BC I think.

core-bundle/contao/templates/_new/frontend_module/feed_reader.html.twig
```diff
@@ -0,0 +1,63 @@
+{% extends "@Contao/frontend_module/template.html.twig" %}
```
ausi:
```suggestion
{% extends "@Contao/content_element/_base.html.twig" %}
```

ausi (dismissed):
One thought: we mabe just could overwrite the `rssReader` instead of creating a new type `feed_reader`, if one wants to use the old module they can register the old class manually (as we do with the new content element controllers).

Apart from that this is IMO RTM ✅

core-bundle/contao/modules/ModuleRssReader.php
```diff
@@ -36,6 +36,13 @@ class ModuleRssReader extends Module
 	 */
 	public function generate()
 	{
+		trigger_deprecation('contao/core-bundle', '5.1', sprintf('The "%s" has been deprecated and will be removed in Contao 6. Use the Feed Reader Module instead.', __CLASS__));
```
Toflar: This should be outside of the class, no? @ausi 
Otherwise you don't get the deprecation if you `extend ModuleRssReader` and override `generate()`.

fritzmg (changes requested):
I agree that the new module should replace the old one, so that you can switch back just as with the content elements. So the type of the new module should stay at `rssReader`. With an `onload_callback` the palette could then be adjusted so that the same options are still available in case the old module is registered.

There is also no need to change `tl_module.rss_cache` to `tl_module.feedCache` etc. These fields can simply stay the same.

core-bundle/contao/modules/ModuleRssReader.php
ausi: I think I agree. But it seems we have no deprecations for the old content element classes yet, but that is probably its own issue.

m-vo:
> I agree that the new module should replace the old one, so that you can switch back just as with the content elements. So the type of the new module should stay at `rssReader`. 

IIRC we discussed how to migrate modules to new ones at the last dev meeting and came up with this:

1) Old modules are left untouched (until Contao 6 and then removed).
1) New modules get a different type, for example `foo_v2`, so that they can coexist with the old ones.

It was a different story with the new content elements as we had a major version change there and could simply adjust the defaults (= force people to migrate). With the above strategy the transition is made easy.

bezin:
I refactored everything to be compatible to only a minor version update (this PR was originally targeted against 5.0)
I do not rename the columns and do not migrate ModuleRssReader usage to FeedReaderController.

core-bundle/contao/modules/ModuleRssReader.php
bezin: I trigger the deprecation outside of the class now.

core-bundle/contao/dca/tl_module.php
bezin: I kept the original column names, even though they are not very accurately named.

core-bundle/contao/languages/en/modules.xlf
```diff
@@ -311,6 +311,12 @@
       <trans-unit id="FMD.rssReader.1">
         <source>Adds an RSS feed to the page</source>
       </trans-unit>
+      <trans-unit id="FMD.feed_reader.0">
```
Toflar, line 314: So basically this is "just" a new FE module now, replacing the old one, right?
We should adjust the title of the existing one to `RSS reader (deprecated)` :)

core-bundle/contao/templates/_new/frontend_module/feed_reader.html.twig
```diff
@@ -0,0 +1,63 @@
+{% extends "@Contao/content_element/_base.html.twig" %}
```
fritzmg: We cannot extend from the content element base template as this would add the wrong CSS classes to the wrapper. In #5617 I added a new `_base.html.twig` for front end modules: https://github.com/contao/contao/pull/5617/files#diff-a6575ccf1263c730ea2b7a16a25de8c2b5c21ded553ccbef9ea24e8a1fed4bf6

May be we need a more general `@Contao/_base.html.twig`?

fritzmg (changes requested):

core-bundle/contao/templates/_new/frontend_module/feed_reader.html.twig
ausi: +1 for the `frontend_module/_base.html.twig`

bezin: I copied the base template from your pull request and use that one here, too.

core-bundle/contao/languages/en/modules.xlf
bezin: I labeled the RSS reader module as deprecated.

bezin:
What is the right way to update translation labels depending on the actual module type? Since we now reuse the original columns, the labels for `rss_cache` and `rss_feed` do not match the features of the new feed reader controller: `Here you can define how long the RSS feed is being cached.` and `Please enter the URL of one or more RSS feeds.` but it could also be Atom or JSON.

fritzmg:
For the latter I would just change the translation to 

> Please enter the URL of one or more feeds.

For the former: what's different in the new implementation?

core-bundle/contao/library/Contao/Pagination.php
```diff
@@ -147,9 +147,11 @@ public function __construct($intRows, $intPerPage, $intNumberOfLinks=7, $strPara
 		$this->lblLast = $GLOBALS['TL_LANG']['MSC']['last'];
 		$this->lblTotal = $GLOBALS['TL_LANG']['MSC']['totalPages'];
 
-		if (Input::get($strParameter) > 0)
+		$input = System::getContainer()->get('contao.framework')->getAdapter(Input::class);
+
+		if ($input->get($strParameter) > 0)
 		{
-			$this->intPage = (int) Input::get($strParameter);
+			$this->intPage = (int) $input->get($strParameter);
```
fritzmg, line 154: These changes seem unrelated to the PR. And besides, you do not need to fetch an adapter here anyway, since there are no tests.

core-bundle/contao/library/Contao/Pagination.php
```diff
@@ -217,11 +219,13 @@ public function generate($strSeparator=' ')
...
 
+		$environment = System::getContainer()->get('contao.framework')->getAdapter(Environment::class);
+
 		$blnQuery = false;
-		list($this->strUrl) = explode('?', Environment::get('requestUri'), 2);
+		list($this->strUrl) = explode('?', $environment->get('requestUri'), 2);
 
 		// Prepare the URL
-		foreach (preg_split('/&(amp;)?/', Environment::get('queryString'), -1, PREG_SPLIT_NO_EMPTY) as $fragment)
+		foreach (preg_split('/&(amp;)?/', $environment->get('queryString'), -1, PREG_SPLIT_NO_EMPTY) as $fragment)
```
fritzmg, line 228: Same as above, changes are not related to the PR (and not necessary).

core-bundle/contao/library/Contao/Pagination.php
bezin: [There is a test](https://github.com/contao/contao/blob/5.x/core-bundle/tests/Contao/PaginationTest.php) and the changes are required. We have been over this in this PR twice already.

core-bundle/contao/library/Contao/Pagination.php
bezin: Same as above.

core-bundle/contao/library/Contao/Pagination.php
fritzmg: Oh 🙈

fritzmg:
I tested using `https://rss.orf.at/news.xml` as the feed URL. However, I get the following error in the front end:

```
TypeError:
Contao\CoreBundle\Twig\Runtime\SanitizerRuntime::sanitizeHtml(): Argument #1 ($html) must be of type string, null given, called in var\cache\dev\twig\46\469973027822ed3c8b6960f040bb3270.php on line 396

  at vendor\contao\contao\core-bundle\src\Twig\Runtime\SanitizerRuntime.php:71
  at Contao\CoreBundle\Twig\Runtime\SanitizerRuntime->sanitizeHtml(null)
     (var\cache\dev\twig\46\469973027822ed3c8b6960f040bb3270.php:396)
```
The line in question:
```php
echo $this->env->getRuntime('Contao\CoreBundle\Twig\Runtime\SanitizerRuntime')->sanitizeHtml(twig_get_attribute($this->env, $this->source, (isset($context["item"]) || array_key_exists("item", $context) ? $context["item"] : (function () { throw new RuntimeError('Variable "item" does not exist.', 52, $this->source); })()), "content", [], "any", false, false, false, 52));
```

It looks like `item.content` can be `null` for certain feeds.

core-bundle/contao/templates/_new/frontend_module/feed_reader.html.twig
```diff
@@ -0,0 +1,63 @@
...
+                <section{{ feed_item_attributes|default }}>
+                    {% block feed_item_headline %}
+                        <h2>
+                            {%- with {href: item.link, text: item.title} %}{{ block('external_link') }}{% endwith -%}
+                        </h2>
+                    {% endblock %}
+                    {% block feed_item_content %}
+                        <div class="content">
+                            {{ item.content|sanitize_html }}
+                        </div>
```
fritzmg:
```suggestion
                        {% if item.content|default %}
                            <div class="content">
                                {{ item.content|sanitize_html }}
                            </div>
                        {% endif %}
```
fixes https://github.com/contao/contao/pull/4847#issuecomment-1377534038

fritzmg (changes requested):

bezin:
> For the latter I would just change the translation to

I simply removed RSS from both labels.

fritzmg (dismissed):

core-bundle/contao/templates/_new/frontend_module/template.html.twig
```diff
@@ -1 +0,0 @@
-{% extends "@Contao/content_element/template.html.twig" %}
```
ausi: Was this file removed on purpose?

ausi (dismissed):
Apart from the deleted *frontend_module/template.html.twig* ✅

core-bundle/contao/templates/_new/frontend_module/template.html.twig
bezin: Yes, I did not see, why it would be relevant anymore?

fritzmg: `frontend_module/template.html.twig` is the template for the _Template_ module.

bezin: Oh, I thought that was one I added during the process 😂 Will restore

ausi (dismissed):

fritzmg (dismissed):

core-bundle/contao/library/Contao/Input.php
```diff
@@ -545,9 +545,9 @@ private static function cleanKeyInternal($varValue)
 	/**
 	 * Strip HTML and PHP tags preserving HTML comments
 	 *
-	 * @param mixed  $varValue             A string or array
-	 * @param string $strAllowedTags       A string of tags to preserve
-	 * @param string $strAllowedAttributes A serialized string of attributes to preserve
+	 * @param mixed                                       $varValue             A string or array
+	 * @param string                                      $strAllowedTags       A string of tags to preserve
+	 * @param string|list<array{key:string,value:string}> $strAllowedAttributes A serialized string of attributes to preserve
```
leofeyer: This change seems unrelated. What is it for?

bezin: @ausi

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
```diff
@@ -0,0 +1,114 @@
...
+                $feeds
+            )
+        );
+
+        usort($elements, static fn (array $a, array $b): int => $a['item']->getLastModified() <=> $b['item']->getLastModified());
+
+        if ($model->perPage > 0) {
+            $param = 'page_r'.$model->id;
+            $page = $request->query->getInt($param, 1);
+            $config = $this->container->get('contao.framework')->getAdapter(Config::class);
```
leofeyer: This is wrong. The framework should be injected instead.

leofeyer (changes requested):

core-bundle/contao/library/Contao/Input.php
ausi: PHPStan fix because of https://github.com/contao/contao/blob/9cfdff0907612cba333bb82d1a8c1d20a2e00458/core-bundle/src/Twig/Runtime/SanitizerRuntime.php#L75

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
ausi: Even though we explicitly require that service in the parent class? https://github.com/contao/contao/blob/0f81487b38bc93af15abc341ea05ea5400a297f7/core-bundle/src/Controller/AbstractController.php#L33

core-bundle/contao/library/Contao/Input.php
leofeyer: We will encapsulate the change in #5682.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
leofeyer: We also require the logger in the abstract controller and still the service is injected in line 36 instead of fetched via `$this->container->get('logger')`.

https://github.com/contao/contao/blob/0f81487b38bc93af15abc341ea05ea5400a297f7/core-bundle/src/Controller/AbstractController.php#L35

Not sure what the best practice is here. 🤔

leofeyer: I just found out that we have a shortcut method to get Contao adapters in controllers. 😄 So I guess a34fc052bab1bfec9718112722ab3b170be201a8 is the way to go.

leofeyer (approved):

leofeyer:
Thank you @bezin.

core-bundle/src/Controller/FrontendModule/FeedReaderController.php
m-vo: > Even though we explicitly require that service in the parent class

In general: yes, that's the best practice. This way you can alter things in the base class (e.g. refactor stuff and thereby remove an injected service) without breaking stuff.

leofeyer:
Thank you @bezin.

PR #4849, opened 2022-06-20
Show array keys instead of nothing in show column mode

Tastaturberuf:
Instead of null, show the array keys as header columns if there is no translation.

image


aschempp (approved):

leofeyer:
Thank you @Tastaturberuf.

PR #4850, opened 2022-06-20
Show array keys als fallback for sorting dropdowns

Tastaturberuf:
Show array key as fallback when no translation is available and prevent php warning >= 8.0: Warning: Undefined array key "title"

image


leofeyer (approved):

leofeyer:
Thank you @Tastaturberuf.

PR #4851, opened 2022-06-21
Get rid of the base tag

leofeyer:
This will finally allow us to remove the `<base>` tag.

### Problem 1: Assets in Twig

@m-vo There is a problem with the new `code.html.twig` file because it does not prepend the base path. Case we use the `asset()` function to fix this?

https://github.com/contao/contao/blob/47697073752ae4d6fa6b63b28e0b54c50a8143f3/core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig#L4

### Problem 2: FilesModel

I have added a new `getFullPath()` method, so we don‘t have to prepend the base path to every `$filesModel->path` occurrence. Any objections?

### Problem 3: TinyMCE

I still have to check if everything works there. That‘s why this is a draft PR.


ausi:
Should we maybe still keep the base tag in the template for better backwards compatibilty?
Or (probably better) add a note to the DEPRECATED.md of Contao 4.13 that relying on `<base>` is deprecated?

ausi:
> ~I still have to check if everything works there. That‘s why this is a draft PR.~ Everything seems to work. 👍

Did we already implement https://github.com/contao/contao/pull/4326#issuecomment-1118672471 ?

> As discussed in the Contao call, we want to add the `{{env::path}}` insert tag automatically in TinyMCE for relative URLs.

leofeyer:
It was not necessary according to my tests. TinyMCE did resolve the base path correctly.

ausi:
> TinyMCE did resolve the base path correctly.

Yes, but it stores base path in the URL too instead of `{{env::path}}`, right?
This means if you develop locally URLs like `/subdir/index.php/foo.html` would end up in the HTML, which will then break once going online with that project.

fritzmg:
> It was not necessary according to my tests. TinyMCE did resolve the base path correctly.

That's not the issue. The issue comes when you want to switch your application from being in a subfolder to not being in a subfolder or vice versa. The `<base>` tag gives us the advantage of storing relative URLs in the database, no matter how you drive Contao. Removing it forces us to adapt a few things.

In general this PR conflicts with my own. If you want to continue the work yourself - imho you should use my PR as base may be?

leofeyer:
> Yes, but it stores base path in the URL too instead of `{{env::path}}`, right? This means if you develop locally URLs like `/subdir/index.php/foo.html` would end up in the HTML, which will then break once going online with that project.

No, it won‘t. I expected the described behavior, however, the actual behavior was different.

First URL: `https://a.wip/contao?do=article…`
Second URL: `https://b.wip/managed-edition/public/contao?do=article…`
Relative link: `en/content-elements.html`

In both cases, the relative link stayed the same after saving it in TinyMCE.

fritzmg:
> In both cases, the relative link stayed the same after saving it in TinyMCE.

If the relative link stays `en/content-elements.html` then it won't work when removing the `<base>`.

> Should we maybe still keep the base tag in the template for better backwards compatibilty?

I think that's what we decided on for 5.x yes. And remove it in 6.x.

core-bundle/src/Resources/contao/library/Contao/File.php
```diff
@@ -806,6 +806,19 @@ public function isUnprotected()
 		return (new Folder(\dirname($this->strFile)))->isUnprotected();
 	}
 
+	/**
+	 * Return the full path
+	 */
+	public function getFullPath()
```
ausi: should we maybe add this as `fullPath` to the `__get` method for a better consistency with the current API of the File class?

Or did you use a method to have the same API as the FilesModel has?

leofeyer: I have removed the method again in 19a2f96639ce2eb9bcf5bf091cebee0abbe84728.

leofeyer:
I have added the base tag in the front end again in 590b4e890ac539cd742c1bfc82bb8869182fdea2.

Toflar:
I wonder what the upgrade path here is though. Shouldn't we have an option to remove the base tag without having to edit the `fe_page` template? I mean, it's the recommended new way, why do I have to adjust the template if I want to get it right and not the other way around?

m-vo:
> @m-vo There is a problem with the new code.html.twig file because it does not prepend the base path. Case we use the asset() function to fix this?

Yes, that should work I think. 

If the base tag is gone, we can also remove the script and `data-toplink` attribute from the `content_element/toplink.html.twig` template. But that's something for a later step I guess.

ausi:
> Shouldn't we have an option to remove the base tag without having to edit the `fe_page` template?

Like an app configuration to opt out of the base tag?

Toflar:
The other way around (opt in) but yes, exactly. I think you should explicitly say you want it if you still need it. Then it's also easier to remove it in 6 (because you had `n` months/years to update to the non-base-tag version).

ausi:
> The other way around (opt in)

Not sure about that, that would break many extensions by default.

Toflar:
Whis is why we have Contao 5, no? I mean, what's the point then? If the base tag is there by default, nobody will adjust a thing and we have the same situation in version 6.

fritzmg:
> I mean, what's the point then?

We still want keep things as compatible as possible, to ease the upgrade. And the existence of the `<base>` tag does not hurt, even if you don't need it.

Toflar:
I'm just saying that we'll never get rid of it then. Because the situation will be **EXACTLY** the same for version 6.

leofeyer:
Link handling in TinyMCE has been added in 44f724b12f66a14c4e2dbea63b8ff35d7220c385. I still have to check the other fields that allow to enter URLs.

ausi:
> I'm just saying that we'll never get rid of it then. Because the situation will be **EXACTLY** the same for version 6.

I’m fine either way in this case.

But it will not be the same for version 6. Our BC breaks usually go with:
1. change the core and deprecate in version X
2. remove BC layer in version X+1

So if we change `<base>` tag handling in version 5, the default strategy is to remove it in version 6. 
_(Although I’m fine either way in this case)_

Toflar:
Yeah, but this is exactly what I'm saying?

Toflar:
Nothing is deprecated in this PR now which is the issue.

ausi:
> Nothing is deprecated in this PR now which is the issue.

https://github.com/contao/contao/pull/4851/files#diff-a82dba25f56e719cef404e54d5447f38b3c5571749ef3e0d0898da452886c83fR3-R7

Toflar:
Yeah, nobody's going to read that and I have to adjust the templates for every single project if I want to go the right way. Sounds really wrong to me.

ausi:
> Nothing is deprecated in this PR now which is the issue.

> Yeah, nobody's going to read that

So the issue you have with this PR is that “Nothing is deprecated” but “nobody's going to read that”??
I don’t understand this reasoning. 

Either deprecate `<base>` in 4 and remove in 5, or deprecate in 5 and remove in 6.
As we do every single day for every other change.

Toflar:
The issue is that I need to get active in order to go the new way (adjust the FE page). Rather than to actively say that I still want to use the base tag even though I shouldn't be doing this anymore. I think it should be a layout checkbox that says "Use \<base> tag (deprecated)" - we can enable it by default for v5.

leofeyer:
I have added the `base_path_prefixer` service in 1630fe3b0ce8111858373ec0bdc5b998359141ee.

leofeyer:
All issues have been addressed. This one is ready to be reviewed now. 🥳

core-bundle/src/Resources/contao/elements/ContentPlayer.php
```diff
@@ -144,6 +144,7 @@ protected function compile()
 
 			$objFile = new File($objFileModel->path);
 			$objFile->title = StringUtil::specialchars($strTitle ?: $objFile->name);
+			$objFile->path = Environment::get('path') . '/' . $objFileModel->path;
```
ausi: Causes a `Creation of dynamic property Contao\File::$path is deprecated` notice. but same with the line one above. Because this whole class gets deprecated with Contao 5 we can probably ignore it?
_(just click resolve if you agree ☺️)_

core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1031,7 +1031,7 @@ public static function addToUrl($strRequest, $blnAddRef=true, $arrUnset=array())
 			$uri = '?' . http_build_query($pairs, '', '&amp;', PHP_QUERY_RFC3986);
 		}
 
-		return substr($request->getBaseUrl() . $request->getPathInfo(), \strlen($request->getBasePath() . '/')) . $uri;
+		return $request->getBasePath() . $request->getPathInfo() . $uri;
```
ausi: This change causes the script filename to get removed I think, is this intentional?
Should it be `return $request->getBaseUrl() . $request->getPathInfo() . $uri;` instead?

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -314,7 +314,7 @@ protected static function path(Request $request)
 	 */
 	protected static function script()
 	{
-		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', static::get('scriptName'));
+		return static::get('scriptName');
```
ausi:
```suggestion
		trigger_deprecation('contao/core-bundle', '5.0', 'Using "%s()" has been deprecated and will no longer work in Contao 6.0. Use "%s::%s" instead.', __METHOD__, __CLASS__, 'scriptName');

		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', static::get('scriptName'));
```
I think we should not change these methods, but instead deprecate them and replace all usages of `Environment::get('script')` with `Environment::get('scriptName')`

core-bundle/src/Resources/contao/library/Contao/Environment.php
```diff
@@ -324,7 +324,7 @@ protected static function script()
 	 */
 	protected static function request()
 	{
-		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', static::get('requestUri'));
+		return static::get('requestUri');
```
ausi:
```suggestion
		trigger_deprecation('contao/core-bundle', '5.0', 'Using "%s()" has been deprecated and will no longer work in Contao 6.0. Use "%s::%s" instead.', __METHOD__, __CLASS__, 'requestUri');

		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', static::get('requestUri'));
```
I think we should not change these methods, but instead deprecate them and replace all usages of `Environment::get('request')` with `Environment::get('requestUri')`

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -730,7 +730,7 @@ static function ($allowedTag)
 							break;
 
 						case 'path':
-							$arrCache[$strTag] = Idna::decode(Environment::get('base'));
+							$arrCache[$strTag] = Idna::decode(Environment::get('path')) . '/';
```
ausi: This needs to be reverted IMO.
`{{env::path}}` always referred the base href, and is documented this way too.

Which means all my comments about using `{{env::path}}` were wrong 🙈

Instead we should add a new one `{{env::base_path}}` that uses Symfonys `getBasePath()` and prefix everything with `{{env::base_path}}/`.
(The naming is then consistent with the existing `base_url` and with Symfony Request)

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -635,6 +635,67 @@ public static function insertTagToSrc($data)
...
+		for ($i=0, $c=\count($paths); $i<$c; $i+=4)
+		{
+			$return .= $paths[$i];
+
+			if (!isset($paths[$i+1]))
+			{
+				continue;
+			}
+
+			if (preg_match('(^(https?://|#|/|\{\{))i', $paths[$i+3]))
```
ausi:
```suggestion
			if (preg_match('(^([0-9a-z+.-]+:|#|/|\{\{))i', $paths[$i+3]))
```

We have to keep any links with a protocol untouched I think (e.g. `tel:…`).

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -635,6 +635,67 @@ public static function insertTagToSrc($data)
...
+			{
+				$return .= $paths[$i+2] . '="' . $paths[$i+3] . '"';
+			}
+			else
+			{
+				$return .= $paths[$i+2] . '="{{env::path}}' . $paths[$i+3] . '"';
```
ausi:
```suggestion
				$return .= $paths[$i+2] . '="{{env::base_path}}/' . $paths[$i+3] . '"';
```
The prefixes need to be changed then everywhere to this 😬

core-bundle/src/Routing/BasePathPrefixer.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        if (preg_match('(^(https?://|/|#|\{\{))i', $url)) {
+            return $url;
+        }
+
+        if (!$request = $this->requestStack->getCurrentRequest()) {
+            throw new \RuntimeException('The request stack did not contain a request');
+        }
+
+        return $request->getBasePath().'/'.$url;
```
ausi:
```suggestion
        return ($this->requestStack->getCurrentRequest()?->getBasePath() ?? '').'/'.$url;
```

I think an exception is not necessary here, just using `/` if there is no request should be what a developer wants in most cases.

ausi (changes requested):
Looks great! ❤️

Some final touches and it will be 🚀

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: <del>This is intentionally, otherwise the script would be added to the assets, too. We are using `$request->getBasePath()` in `Environment::get('path')` as well.</del>

I was wrong. We have to use `getBaseUrl()` here. Why do I keep questioning your suggestions anyway? 😄

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
leofeyer: I would prefer to document the change and fix the `{{env::path}}` insert tag in Contao 5.

core-bundle/src/Resources/contao/library/Contao/Environment.php
leofeyer: Usages have been replaced in fc5b8d1081f62fbfb51f3e4ded9516ccac49dbd3.

core-bundle/src/Resources/contao/library/Contao/Controller.php
leofeyer: Fixed in 5dc3c9efe1a7526b3b3ec5ae2e2552d53285d526.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: I dug a bit in the code history, `{{env::path}}` was referring to `get('base')` ever since: https://github.com/contao/core/blob/4a2809d34ce08193c26b1c5815ecad2c1cdb909b/system/libraries/Controller.php#L1303-L1305 (Initial Commit)

I think we should do the following:
1. Deprecate `{{env::path}}` but keep the functionality as it is currently
2. Add `{{env::base}}` as the replacement
3. Add `{{env::base_path}}` (as a counterpart to `{{env::base_url}}`) and use the prefix `{{env::base_path}}/` everywhere.

This would streamline the naming to be straight forwad IMO:
1. `{{env::base}}` refers to `get('base')`
2. `{{env::base_path}}` refers to `$request->getBasePath()`
3. `{{env::base_url}}` refers to `$request->getBaseUrl()` (as it already does)

core-bundle/src/Resources/contao/elements/ContentPlayer.php
leofeyer: I guess we can ignore it, however, I have still fixed it in 6d2a1c6fcec7f8562213b9442b9d9108df34c88e.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
leofeyer: Changed in 56aa6a53b57d6db22ab782ffac95d8a2e5d2effe.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
leofeyer: Changed in 56aa6a53b57d6db22ab782ffac95d8a2e5d2effe.

core-bundle/src/Routing/BasePathPrefixer.php
```diff
@@ -0,0 +1,31 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Routing;
```
aschempp: This service is unrelated to routing, same as the Symfony request is not related to routing (but to `HttpFoundation`)

ausi (dismissed):
Approved, but not yet (fully) tested locally.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
```diff
@@ -143,9 +143,12 @@ protected function compile()
 			}
 
 			$objFile = new File($objFileModel->path);
-			$objFile->title = StringUtil::specialchars($strTitle ?: $objFile->name);
 
-			$arrFiles[$objFile->extension] = $objFile;
+			$arrFiles[$objFile->extension] = array
```
aschempp: this was previously an object and now is an array, that would break existing `ce_player` templates. Any reason to not just cast this? Also, this should most likely use the asset helper to generate the path of a file (which would include the asset URL if configured in the root page), not manually append a path.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -730,6 +730,10 @@ static function ($allowedTag)
 							break;
 
 						case 'path':
+							trigger_deprecation('contao/core-bundle', '5.0', 'Using the "env::path" insert has been deprecated and will no longer work in Contao 6.0. Use the "env::base" insert tag instead.');
```
aschempp: What's the point of renaming an insert tag?

core-bundle/src/Resources/contao/templates/elements/ce_player.html5
```diff
@@ -6,13 +6,13 @@
     <?php if ($this->isVideo): ?>
       <video<?= $this->size ?><?php if ($this->poster): ?> poster="<?= $this->poster ?>"<?php endif; ?><?php if ($this->preload): ?> preload="<?= $this->preload ?>"<?php endif; ?> <?= implode(' ', $this->attributes) ?>>
         <?php foreach ($this->files as $file): ?>
-          <source type="<?= $file->mime ?>" src="<?= $file->path.$this->range ?>">
+          <source type="<?= $file['mime'] ?>" src="<?= $file['path'].$this->range ?>">
```
aschempp: Any reason to break existing templates? Why not just convert the (new) array to an object?

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends "@Contao/content_element/_base.html.twig" %}
 {% use "@Contao/component/_stylesheet.html.twig" %}
 
-{% set highlighter_css_file = 'vendor/scrivo/highlight_php/styles/foundation.css' %}
+{% set highlighter_css_file = asset('vendor/scrivo/highlight_php/styles/foundation.css') %}
```
aschempp: I don't think this is correct. Without arguments, the `asset` helper will use the configured asset strategy of the application, which might be configured to be somewhere else. In our case for example, assets are our layout files and are always located in `/web/layout/`. We should probably register an asset package for this extension (similar to `AddAssetsPackagesPass` but maybe "hardcoded" since we know which one to register).

core-bundle/tests/Contao/TemplateTest.php
```diff
@@ -241,7 +241,7 @@ public function testStripsLeadingSlashFromAssetUrl(): void
         $template = new FrontendTemplate();
         $url = $template->asset('/path/to/asset', 'package_name');
 
-        $this->assertSame('path/to/asset', $url);
+        $this->assertSame('/path/to/asset', $url);
```
aschempp: The test method name does not make sense anymore. Not sure we need that test at all now?

core-bundle/tests/Contao/TemplateTest.php
```diff
@@ -275,7 +275,7 @@ public function testStripsTheBasePathFromAssetUrl(): void
         $template = new FrontendTemplate();
         $url = $template->asset('/path/to/asset', 'package_name');
 
-        $this->assertSame('path/to/asset', $url);
+        $this->assertSame('/foo/path/to/asset', $url);
```
aschempp: The test method name does not make sense anymore. Not sure we need that test at all now?

news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
```diff
@@ -106,7 +106,7 @@ protected function compile()
 			case 'external':
 				if ($objArticle->url)
 				{
-					throw new RedirectResponseException($objArticle->url, 301);
+					throw new RedirectResponseException(System::getContainer()->get('contao.routing.base_path_prefixer')->prefix($objArticle->url), 301);
```
aschempp: not sure why we would do that. Either the URL is an insert tag which should (now) generate an absolute URL, or we should not touch the URL at all in my opinion. Not sure why we never replaced insert tags in this URL?

aschempp (changes requested):
All the `request => requestUri` changes look good to me (which is like 80% here 😆).

What I dislike is that there is a lot of string manipulation. Paths should either be generated by routing, or by assets, but not by appending strings to relative file or route paths. I know this is not always possible, but I feel like this should be where we aim for. 

As an example, if `{{file::...}}` uses the asset service to generate a file path, the path would already be correct. We don't need the `StringUtil` helper methods then.

core-bundle/src/Routing/BasePathPrefixer.php
leofeyer: @ausi /cc

core-bundle/src/Resources/contao/elements/ContentPlayer.php
leofeyer: We are talking Contao 5, so breaks are allowed. It does not matter anyway because the `ContentPlayer` element will be replaced with a modern content element fragment before the release of Contao 5.0.

core-bundle/src/Resources/contao/templates/elements/ce_player.html5
leofeyer: See https://github.com/contao/contao/pull/4851#discussion_r903546350

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
leofeyer: It should use our `assets.packages` service, which already handles everything correctly:

https://github.com/contao/contao/blob/47697073752ae4d6fa6b63b28e0b54c50a8143f3/core-bundle/src/Resources/contao/library/Contao/Template.php#L452

@m-vo How do we add this to Twig?

core-bundle/src/Routing/BasePathPrefixer.php
ausi: One one hand you write _“Paths should either be generated by routing”_, on the other you say generating this path is not related to routing. I’m confused ☺️

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: See https://github.com/contao/contao/pull/4851#discussion_r903451622

news-bundle/src/Resources/contao/modules/ModuleNewsReader.php
leofeyer: You are right. Reverted in 200677111be272d769ddd356cee0bbd12e2805b6.

core-bundle/tests/Contao/TemplateTest.php
leofeyer: Removed in e0dc286384a8712dde1fba1632bb73ea8fbd609f.

core-bundle/tests/Contao/TemplateTest.php
leofeyer: Removed in e0dc286384a8712dde1fba1632bb73ea8fbd609f.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
leofeyer: The `env::path` insert tag outputs the base URL, therefore it should be `env::base`. But we can discuss keeping the "wrong" insert tag instead.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
fritzmg: > It does not matter anyway because the ContentPlayer element will be replaced with a modern content element fragment before the release of Contao 5.0.

But the old one will still be available and we still need to support that.

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
aschempp: no, as I said, the **default** assets packages can be configured in your application. Like we do in https://github.com/contao-association/website/blob/main/config/config.yml#L27-L29 for example. This might influence the path to the public folder. See https://members.contao.org/layout/manifest.json for the files we use on that website, not sure what would happen if the path does not exist in the manifest.json

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
aschempp: Thanks @ausi. I'm fine with changing/adding the new insert tag, I would then just not deprecate the `{{env::path}}` and just keep it as a synonym for now, there's no point in asking people to rename it if there is no issue except for "nicer" naming imho.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
aschempp: yeah, which even more so makes it a bad experience to change it for no reason imho

core-bundle/src/Routing/BasePathPrefixer.php
aschempp: prefixing a **path** is not related to routing. Routing always generates absolute paths in Symfony, so there can never be a _path prefixer_ necessary for routing. The only reason we use this service (which I don't like, see review comment) is the _fake generate_ absolute paths, mostly for files and not for routes.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
ausi: The name causes confusion, so it should not be used anymore, so we have to deprecate it IMO.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
leofeyer: It is not changed "for no reason". See https://github.com/contao/contao/pull/4851#discussion_r903085380.

aschempp: I did mean the change is "for no reason" because we can simply cast the array to an object for the template

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
leofeyer: Ok, so let‘s not make this more complicated than necessary. We do need a way to prefix the base path here and we sure don‘t want to use `env::base_path`. @contao/developers Suggestions?

core-bundle/src/Routing/BasePathPrefixer.php
leofeyer: @aschempp Which namespace do you suggest instead?

core-bundle/src/Resources/contao/elements/ContentPlayer.php
ausi:
```suggestion
			$arrFiles[$objFile->extension] = (object) array
```
This would be enough to make @aschempp and @fritzmg happy, correct?
If so, I’m in favor ☺️

ausi (dismissed):

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
ausi: I think `app.request.base_path` should be available in twig?

```suggestion
{% set highlighter_css_file = app.request.base_path|default ~ '/vendor/scrivo/highlight_php/styles/foundation.css' %}
```

But I’m not sure if `app` always exists in all cases in our environment?

core-bundle/src/Routing/BasePathPrefixer.php
ausi: > prefixing a *path* is not related to routing. 

https://github.com/contao/contao/blob/47697073752ae4d6fa6b63b28e0b54c50a8143f3/core-bundle/src/Routing/ImagesLoader.php#L36

core-bundle/src/Resources/contao/elements/ContentPlayer.php
leofeyer: But I‘m against this. There is no point in using a `stdClass` object instead of an array. It is Contao 5 – we are allowed to make backwards incompatible changes. Our BC promise does not even cover templates, so we could even do this in Contao 4.13!

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
leofeyer: It does and this works. 👍 Changed in ca951813c4d0bb25f63adba7aab8c54d9623674d.

leofeyer:
We could use the `@assets.context` service to get the base path. It contains a `RequestStackContext` object that has a `getBasePath()` method. @aschempp Since the base path is the same for all assets, we should be able to rely on it, shouldn‘t we?

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
m-vo: Using `app.request.base_path` is probably fine. :+1: I'd go with this. If we find a better solution later on, we can still change it.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
aschempp: yes we are allowed to break. But I dislike to break stuff for no apparent reason just because you touched it 🤷

core-bundle/src/Routing/BasePathPrefixer.php
aschempp: > @aschempp Which namespace do you suggest instead?

Honestly, I think we should not have this service at all. There are two cases we use it as far as I can tell:
1. to "prefix" URLs of a string input field (e.g. hyperlink URL or event URL). An insert tag in this field should already generate the correct URL, and manually entering a relative URL **to the current system** would be a user error imho.
2. Prefixing a file path, which should be done through the assets component instead (to also prefix the root-page configured asset URL).

fritzmg: > and manually entering a relative URL to the current system would be a user error imho.

If you let users enter absolute URLs, then the URLs will be wrong if the base path changes.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
ausi: > I dislike to break stuff for no apparent reason

Except in your own PRs 🤣 https://github.com/contao/contao/pull/4839#issuecomment-1158586258

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
aschempp: The correct solution would be `asset('foundation.css', 'scrivo/highlight.php')` after merging https://github.com/contao/contao/pull/4866
This will
 1. remove dependency on hardcoded path in templates
 2. automatically add a version strategy, so the link will be `foundation.css?v={version of composer package}`
 3. automatically prepend the root-page configured asset URL

aschempp:
> We could use the `@assets.context` service to get the base path. It contains a `RequestStackContext` object that has a `getBasePath()` method. @aschempp Since the base path is the same for all assets, we should be able to rely on it, shouldn‘t we?

yes, the asset.context contains the base path. But we most likely have a package (see https://github.com/contao/contao/pull/4866), so I'm unsure what place you are referring to?

leofeyer:
Pretty much everything else that is not a package. 😄 Files for example and relative links.

core-bundle/src/Routing/BasePathPrefixer.php
leofeyer: The service has been removed in 5c8b7883def1166e40ad8d5e7b69df0396d3fa88.

core-bundle/src/Resources/contao/templates/_new/content_element/code.html.twig
leofeyer: Changed in 8c069aaf92c6927a8cf59ec317b4f66f43f7b999.

leofeyer:
Prefixing for `imageUrl` and `PageRedirect` has been added in 4e82628174e9002923c8eab1b6ca766407c465b9.

core-bundle/src/Resources/contao/forms/FormSubmit.php
```diff
@@ -100,7 +100,7 @@ public function parse($arrAttributes=null)
 
 			if ($objModel !== null && is_file(System::getContainer()->getParameter('kernel.project_dir') . '/' . $objModel->path))
 			{
-				$this->src = $objModel->path;
+				$this->src = Environment::get('path') . '/' . $objModel->path;
```
aschempp:
```suggestion
				$this->src = System::getContainer()->get('contao.assets.files_context')->getStaticUrl() . $objModel->path;
```

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -1009,7 +1017,7 @@ static function ($allowedTag)
 
 						if ($objFile !== null)
 						{
-							$arrCache[$strTag] = System::urlEncode($objFile->path);
+							$arrCache[$strTag] = System::urlEncode(System::getContainer()->get('assets.context')->getBasePath() . '/' . $objFile->path);
```
aschempp:
```suggestion
							$arrCache[$strTag] = System::urlEncode(System::getContainer()->get('contao.assets.files_context')->getStaticUrl() . $objFile->path);
```

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -635,6 +635,67 @@ public static function insertTagToSrc($data)
...
+	 * Adds {{env::base_path}} to relative links
+	 *
+	 * @param string $data The markup string
+	 *
+	 * @return string
+	 */
+	public static function addBasePath($data)
+	{
+		$return = '';
+		$paths = preg_split('/((src|href)="([^"]+)")/i', $data, -1, PREG_SPLIT_DELIM_CAPTURE);
```
aschempp, line 648: what about single quote HTML tags or inline CSS urls etc? Also, should the method name somehow reflect that this is meant for HTML content search-replace? It currently sounds like it would just prefix something.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -635,6 +635,67 @@ public static function insertTagToSrc($data)
...
 	}
 
+	/**
+	 * Adds {{env::base_path}} to relative links
+	 *
+	 * @param string $data The markup string
+	 *
+	 * @return string
+	 */
+	public static function addBasePath($data)
```
aschempp, line 645: we should probably have unit tests for these new methods?

aschempp (commented):
I guess there are more places to use `contao.assets.files_context` 🤔

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
leofeyer: We already do.

core-bundle/src/Resources/contao/library/Contao/StringUtil.php
leofeyer: The methods are built the same way as the `insertTagToSrc()` and `srcToInsertTag()` methods, which we already use for TinyMCE. If you think we should support single quotes or inline CSS URLs, feel free to create a pull request that adds this feature to the existing methods, too.

core-bundle/src/Resources/contao/forms/FormSubmit.php
leofeyer: Changed in 5a2b802fa3571b4bdd9bb5fcb38dc70950fe7a78.

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
leofeyer: Changed in 5a2b802fa3571b4bdd9bb5fcb38dc70950fe7a78.

leofeyer:
We are now consistently using `Environment::get('path')` or `$request->getBasePath().'/'` for links and `System::getContainer()->get('contao.assets.files_context')->getStaticUrl()` for files. 👍

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -1009,7 +1017,7 @@ static function ($allowedTag)
 
 						if ($objFile !== null)
 						{
-							$arrCache[$strTag] = System::urlEncode($objFile->path);
+							$arrCache[$strTag] = System::urlEncode(System::getContainer()->get('contao.assets.files_context')->getStaticUrl() . $objFile->path);
```
ausi:
```suggestion
							$arrCache[$strTag] = System::getContainer()->get('contao.assets.files_context')->getStaticUrl() . System::urlEncode($objFile->path);
```
Otherwise this would double encode the static url part.

leofeyer: Good catch! 🦊

ausi (dismissed):
❤️❤️❤️

core-bundle/src/Resources/contao/library/Contao/InsertTags.php
leofeyer: The deprecation has been removed again in a94dc2fe6e2d12c9068485d5c532cd322626b4bb.

core-bundle/src/Resources/contao/elements/ContentPlayer.php
leofeyer: BC break reverted in 3832d2462c84cf13325ac3eade58acd8ae8cf453.

ausi (approved):

PR #4853, opened 2022-06-21
Determine the InnoDB index length only once per schema

ausi:
This should reduce the amount of SQL queries during schema creation a lot. 🚀

/cc @Toflar


fritzmg (approved):

bytehead (approved):

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4857, opened 2022-06-22
Correctly add the preview script

leofeyer:
If Contao runs in a subfolder (which we allow locally), the `/preview.php` fragment will be added incorrectly.

URL: `https://domain.wip/managed-edition/public/contao/preview`
Before: `https://domain.wip/preview.php/managed-edition/public/contao/preview`
After: `https://domain.wip/managed-edition/public/preview.php/contao/preview`

### DIFF
--- a/core-bundle/src/Controller/BackendPreviewController.php
+++ b/core-bundle/src/Controller/BackendPreviewController.php
@@ -51,8 +51,14 @@ class BackendPreviewController
     {
         // Skip the redirect if there is no preview script, otherwise we will
         // end up in an endless loop (see #1511)
-        if ($this->previewScript && $request->getScriptName() !== $this->previewScript) {
-            return new RedirectResponse($this->previewScript.$request->getRequestUri());
+        if ($this->previewScript) {
+            $script = '/'.basename($request->getScriptName());
+
+            if ($script !== $this->previewScript) {
+                $path = dirname($request->getScriptName());
+
+                return new RedirectResponse($path.$this->previewScript.str_replace($path, '', $request->getRequestUri()));
+            }
         }
 
         if (!$this->authorizationChecker->isGranted('ROLE_USER')) {
--- a/core-bundle/src/Controller/BackendPreviewController.php
+++ b/core-bundle/src/Controller/BackendPreviewController.php
@@ -51,14 +51,8 @@ class BackendPreviewController
     {
         // Skip the redirect if there is no preview script, otherwise we will
         // end up in an endless loop (see #1511)
-        if ($this->previewScript) {
-            $script = '/'.basename($request->getScriptName());
-
-            if ($script !== $this->previewScript) {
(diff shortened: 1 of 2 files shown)


core-bundle/src/Controller/BackendPreviewController.php
```diff
@@ -51,8 +51,14 @@ public function __invoke(Request $request): Response
...
-        if ($this->previewScript && $request->getScriptName() !== $this->previewScript) {
-            return new RedirectResponse($this->previewScript.$request->getRequestUri());
+        if ($this->previewScript) {
+            $script = '/'.basename($request->getScriptName());
+
+            if ($script !== $this->previewScript) {
+                $path = \dirname($request->getScriptName());
+
+                return new RedirectResponse($path.$this->previewScript.str_replace($path, '', $request->getRequestUri()));
+            }
```
ausi: Just stripping the base path length should be enough here I think?
https://github.com/leofeyer/contao/pull/5/files#diff-f9889dbdd986f7bd5770e80b782c7a15ed4c8998d51ea7d8fefd1f1e2f4d0dbb

aschempp (commented):
what about all the other places, like `PageModell::getPreviewUrl()`?

also, should it ever be possible to set a preview script that includes a subfolder/path? Or maybe we should add a validation for that?

also, this seems to assume the default entry point is never called with a script, or am I reading this wrong?

core-bundle/src/Controller/BackendPreviewController.php
leofeyer: Does not seem to work:

leofeyer: Changed in 1983f21ec8d6188f202a3ee17d498e4cb3e9d52f.

leofeyer:
> what about all the other places, like `PageModell::getPreviewUrl()`?

They were not affected in my tests, probably because we are using the UrlGenerator with a custom context there instead of the request object.

aschempp (approved):

ausi (approved):

qzminski:
It look like that after this commit, the frontend preview URLs like `https://domain.tld/contao/preview?page=6` no longer work. What happens is that the user is redirected to the `https://domain.tld/preview.php/contao/preview` (notice missing query parameters) and the frontend preview always shows the homepage.

I think the fix would be:

```diff
-return new RedirectResponse($request->getBasePath().$this->previewScript.$request->getPathInfo());
+return new RedirectResponse($request->getBasePath().$this->previewScript.$request->getPathInfo().($request->getQueryString() ? ('?' . $request->getQueryString()) : ''));
```

But I am not completely sure.

Toflar:
Looks correct to me. Would you be so kind to create a PR? I think to show the problem it should be sufficient to add a query to any of the tests ;)

qzminski:
See https://github.com/contao/contao/pull/4891

PR #4860, opened 2022-06-22
Cache index default length per table engine in DcaSchemaProvider

bezin:
When in `dev` mode (without warming the cache first), I noticed that when I open the back end dashboard, the database is queried 224 times. 172 queries of these are `SHOW VARIABLES LIKE 'innodb_large_prefix'` and `SELECT @@version` (88 times each) making up for about 83 % of the total time spent on database queries. With this PR the index default length is cached per engine in `DcaSchemaProvider`, so these queries are only run once per engine.


bytehead:
Duplicate of #4853 🙈

PR #4861, opened 2022-06-22
Fix a potential version comparison if the field definitions are missing

qzminski:
I have a setup where I use Doctrine entities and have no SQL definitions in the DCA table. This results in a PHP 8 warning when comparing the record versions due to missing field definitions like `tstamp`.

The field definitions are not available, because the `Versions` class uses `DcaExtractor` behind the scenes, which sees only the fields that have a SQL definition. That is of course not the case for me, because those are defined in the Doctrine entities.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Versions.php
+++ b/core-bundle/src/Resources/contao/classes/Versions.php
@@ -459,14 +459,19 @@ class Versions extends Controller
 							continue;
 						}
 
-						if (\is_array($arrFields[$k]))
+						$blnIsBinary = false;
+
+						if (isset($arrFields[$k]))
 						{
-							// Detect binary fields using Doctrine's built-in types or Contao's BinaryStringType (see #3665)
-							$blnIsBinary = \in_array($arrFields[$k]['type'] ?? null, array(BinaryType::class, BlobType::class, Types::BINARY, Types::BLOB, BinaryStringType::NAME), true);
-						}
-						else
-						{
-							$blnIsBinary = strncmp($arrFields[$k], 'binary(', 7) === 0 || strncmp($arrFields[$k], 'blob ', 5) === 0;
+							if (\is_array($arrFields[$k]))
+							{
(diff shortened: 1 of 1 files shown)


aschempp:
Some changes look unrelated, can you check the base and target branch?

ausi:
> Some changes look unrelated, can you check the base and target branch?
Bildschirmfoto 2022-06-22 um 20 36 13

These are the changes I see. Something wrong with your GitHub?

ausi (approved):

core-bundle/src/Resources/contao/classes/Versions.php
```diff
@@ -459,14 +459,19 @@ public function compare($blnReturnBuffer=false)
...
 						{
-							$blnIsBinary = strncmp($arrFields[$k], 'binary(', 7) === 0 || strncmp($arrFields[$k], 'blob ', 5) === 0;
+							if (\is_array($arrFields[$k]))
+							{
+								// Detect binary fields using Doctrine's built-in types or Contao's BinaryStringType (see #3665)
+								$blnIsBinary = \in_array($arrFields[$k]['type'] ?? null, array(BinaryType::class, BlobType::class, Types::BINARY, Types::BLOB, BinaryStringType::NAME), true);
+							}
+							else
+							{
+								$blnIsBinary = strncmp($arrFields[$k], 'binary(', 7) === 0 || strncmp($arrFields[$k], 'blob ', 5) === 0;
```
leofeyer, line 473: What about `tinyblob`, `mediumblob` and `longblob`? Do we have to handle there here as well?

ausi: Isn’t this unrelated to this PR?
This pull request only wraps the existing code in an `if (isset($arrFields[$k]))`

leofeyer: Ah, "hide whitespace" makes it more clear. Of course then. 👍

leofeyer:
Thank you @qzminski.

PR #4823, opened 2022-06-11
 Modern fragments: video content elements

m-vo:
Followup to #4810

This PR adds the  video content elements…

* `content_element/vimeo`
* `content_element/youtube`

… together with a generic `_splash_screen.html.twig` component.

~~**Note:** This PR is currently based on #4810; needs to be rebased once merged.~~

PR #4862, opened 2022-06-22
Modern fragments: download content elements

m-vo:
Followup to #4823 

This PR adds the files content elements…

* `content_element/download`
* `content_element/downloads`

… together with a  `_download.html.twig`. The changes in aa21eee36735c2e496de0e9302e2aeb1a756a36a are basically fixing bugs with the already merged components that I noticed now when writing the download component. They are otherwise unrelated to this PR.

One of the biggest features of the new download content element is, that it uses a real controller when streaming files. I added a little helper class to ease creating signed download URLs and handling requests on these URLs.

This one is a bit more complex than the other controllers, so I'd be happy to get some feedback before writing tests. :relaxed: 

**Note:** This PR is currently based on #4832 and predecessors; needs to be rebased once merged.


m-vo:
Seems the tests got lost during rebase. I'll try to republish them later this day.

m-vo:
> Seems the tests got lost during rebase.

There they are: 0d34372e586af137f159318b36465b763ae8787c

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -0,0 +1,245 @@
...
+    #[Route('/_download', name: self::DOWNLOAD_ROUTE_NAME)]
+    public function downloadAction(Request $request): Response
+    {
+        $this->initializeContaoFramework();
+
+        return $this->fileDownloadHelper->handle(
+            $request,
+            $this->filesStorage,
+            function (FilesystemItem $item, array $context): Response|null {
+                // TODO: What's needed here in terms of permissions?
```
leofeyer: This needs to be addressed before we can merge.

m-vo: /cc @Toflar 🙏

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -0,0 +1,243 @@
...
+     * If the content should be displayed inline or if the resource does not
+     * have a public URI, a URL pointing to this controller's download action
+     * will be generated, otherwise the direct download URL will be returned.
+     */
+    private function generateDownloadUrl(FilesystemItem $filesystemItem, ContentModel $model): string
+    {
+        $path = $filesystemItem->getPath();
+        $inline = $model->inline;
+
+        if (!$inline && null !== ($publicUri = $this->filesStorage->generatePublicUri($path))) {
```
ausi: I think we should not use the public URI only if `!$inline` as the webserver might still deliver the response with content disposition inline...

Instead I would add a `contentDisposition` option to the `generatePublicUri` method (this is also supported by services like S3 (`ResponseContentDisposition`)).

core-bundle/src/Controller/ContentElement/DownloadsController.php
ausi: I guess we’d need to check, start/stop, invisible/published as well as group permissions (`Controller::isVisibleElement()`) for:

1. The download element
2. The parent article
3. The parent page

But this would not work in all cases, like if your download element is placed on an invisible page and injected elsewhere via an insert/alias element.

So I’m not sure if this is solvable right now as we do not have a defined security model (yet) for things like this.

core-bundle/src/Controller/ContentElement/DownloadsController.php
m-vo: see 32015fe9e52aae1319df5fe0b39cf260ce9f9bf6

core-bundle/src/Controller/ContentElement/DownloadsController.php
m-vo: See 22e4829f9f15ff25451833c04df942f25b4ec9eb for a revert to the old strategy for now.

I tried to stick as closely as possible to the original implementation, so that it will be easy to migrate once we found a solution for the permissions. It was mostly about moving the bit out of the route action into a `handleDownload()` method that gets executed in the beginning of the controller and that - if the current URL contains the relevant parameters and is correctly signed - will issue a `RedirectException` with the `StreamedResponse` instead.

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -0,0 +1,248 @@
...
+    {
+        $response = $this->fileDownloadHelper->handle(
+            $request,
+            $this->filesStorage,
+            function (FilesystemItem $item, array $context): Response|null {
+                if (
+                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($context['id'] ?? null)) ||
+                    !$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())
+                ) {
+                    return new Response('The resource can not be accessed.', Response::HTTP_FORBIDDEN);
```
ausi: Should this be a 404 instead?

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -0,0 +1,248 @@
...
+        $inline = (bool) $model->inline;
+
+        if (null !== ($publicUri = $this->filesStorage->generatePublicUri($path, new ContentDispositionOption($inline)))) {
+            return (string) $publicUri;
+        }
+
+        // Todo: Use an exclusive route once we have a strategy how to handle
+        //       permissions for it. Right now we use the current route and
+        //       then throw a ResponseException to initiate the download.
+        $currentRoute = $request->attributes->get('_route');
```
ausi: Does this work on routes that require parameters like a news detail page?
Or should we use the request URL as the base here?

core-bundle/src/Filesystem/FileDownloadHelper.php
```diff
@@ -0,0 +1,170 @@
...
+ *     generateDownloadUrl() with the controller that should handle the
+ *     download.
+ *
+ *  2) In the controller's action, call handle(), which will generate a
+ *     StreamedResponse of the requested file's content. The URL signature as
+ *     well as the file's existence in the storage are already verified for
+ *     you - if you want to add additional checks, there is a closure you can
+ *     hook into.
+ */
+class FileDownloadHelper
```
ausi, line 38: I’d mark this as `@internal` until we implement the desired permission model for downloads in the frontend.

core-bundle/src/Controller/ContentElement/DownloadsController.php
m-vo: Thinking about it, it could probably be even 410 (gone). Because you would only get to this point if the signed URI was already verified. :thinking:

core-bundle/src/Controller/ContentElement/DownloadsController.php
m-vo: Good point, changed in f6185aa5cbab9daf98d0bd561a5acc9964412be2.

core-bundle/src/Filesystem/FileDownloadHelper.php
m-vo: See 78f437aec18be42f905b127e44ccfa333730bc10.

ausi (dismissed):

core-bundle/src/Controller/ContentElement/DownloadsController.php
m-vo: See  696d5776e5b8032eac3fe331691f7ad9fdc7a7e8.

ausi (dismissed):

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4866, opened 2022-06-23
Use asset packages for vendor packages

aschempp:
see https://github.com/contao/contao/pull/4851#discussion_r903651735

PS: not sure if this should be Contao 5 only or should also be backported @contao/developers ?

### DIFF
--- a/core-bundle/src/Command/SymlinksCommand.php
+++ b/core-bundle/src/Command/SymlinksCommand.php
@@ -35,7 +35,7 @@ use Symfony\Component\Finder\SplFileInfo;
  */
 class SymlinksCommand extends Command
 {
-    public const PACKAGES = [
+    final public const PACKAGES = [
         'scrivo/highlight.php' => 'vendor/scrivo/highlight.php/styles',
     ];
 
--- a/core-bundle/src/Command/SymlinksCommand.php
+++ b/core-bundle/src/Command/SymlinksCommand.php
@@ -35,10 +35,6 @@ use Symfony\Component\Finder\SplFileInfo;
  */
 class SymlinksCommand extends Command
 {
-    final public const PACKAGES = [
-        'scrivo/highlight.php' => 'vendor/scrivo/highlight.php/styles',
-    ];
-
     protected static $defaultName = 'contao:symlinks';
     protected static $defaultDescription = 'Symlinks the public resources into the public directory.';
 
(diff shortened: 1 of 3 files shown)


core-bundle/src/Command/SymlinksCommand.php
```diff
@@ -35,6 +35,10 @@
  */
 class SymlinksCommand extends Command
 {
+    public const PACKAGES = [
+        'scrivo/highlight.php' => 'vendor/scrivo/highlight.php/styles',
```
ausi: Was the rename from `highlight_php` to `highlight.php` on purpose?

m-vo:
I like this!

Just a question: In plain Symfony I'm only adjusting the asset component if I need to use a CDN, in Contao we also have the static URL page properties. How do these two things interact? Or are they completely independent?

core-bundle/src/Command/SymlinksCommand.php
leofeyer: That is the original package name: https://packagist.org/packages/scrivo/highlight.php

m-vo: But why did we have an underscore in the name then? Did the dot cause any problems?

ausi: Yes, but we currently symlink it to `public/vendor/scrivo/highlight_php/styles` and with this change we would symlink it to `public/vendor/scrivo/highlight.php/styles` instead.

leofeyer: Good catch. I have to investigate.

leofeyer: Ah, I remember. The web server would treat `highlight.php` as PHP file, therefore I have renamed the symlink. I will restore the original behavior.

leofeyer: I have fixed the symlink in 4087dc79a8e6a534a3aaed19a68436f9cbc03558. Also, I have removed the `SymlinksCommand::PACKAGES` constant again for now because finding a way to symlink vendor assets is a different topic that might or might not be solved with a class constant. We should discuss it separately once there is more than one package.

leofeyer (approved):

aschempp:
> In plain Symfony I'm only adjusting the asset component if I need to use a CDN, in Contao we also have the static URL page properties. How do these two things interact? Or are they completely independent?

Not sure what you mean. The asset component supports an unlimited amount of "packages". Each package has its location and possibly a version strategy. 
In Contao, we automatically register all contao-components and all bundles as asset packages, to automate paths and cache invalidation on package update. 
The app is complete independent of that. We use Symfony Encore, which provides a manifest.json for asset packages that maps file names to hashfile names. Not related to a CDN at all. But I guess you could register a CDN as well **for the app**. In Contao, "our CDN" is configured in the root page, which is something the asset component does not support out of the box.

PR #4867, opened 2022-06-23
Do not update the search index on redirects

Toflar:
Every redirect currently causes a database connection because it tries to delete the entry from the search index.
I think it's safe to assume that we don't need this.


fritzmg:
Pages that were previously a regular page and then turned into a redirect should be deleted from the search index though, shouldn't they?

Toflar:
That's the 0.1% case which I want to ignore in order to speed up the process for 99.9%.

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

Issue #4868, opened 2022-06-23
Backend readonly mode does not work for all widgets


netzarbeiter:
### Affected version(s)

4.9 & 4.13

### Description

```
contao:
    localconfig:
        timeZone: 'Europe/Berlin'
```

`timeZone` is not displayed as readonly field.

![bug](https://user-images.githubusercontent.com/1942550/175317027-e3548258-8082-48ab-84df-0c308504ed6d.png)

`defaultChmod` is not displayes ad readonly.

![bug](https://user-images.githubusercontent.com/1942550/175322767-da24e4b5-fece-4429-8b43-5bef0ba128ef.png)

fritzmg:
This also applies to widgets in 4.9, not just 4.13.

aschempp:
see https://github.com/contao/contao/pull/4870

fritzmg:
The following widgets do not (fully) support `readonly => true`:

* `checkbox`
* `checkboxWizard`
* `chmod`
* `fileTree`
* `imageSize`
* `inputUnit` (the select is not disabled)
* `metaWizard` (removing still works, but the add button is disabled)
* `moduleWizard`
* `optionWizard`
* `pageTree`
* `picker`
* `radio`
* `sectionWizard`
* `select` with `chosen => true`
* `tableWizard` (you can still add/delete rows and columns, but you cannot edit cells)
* `timePeriod` (the select is not disabled)
* `trbl` (the select is not disabled)

ausi:
> The following widgets do not (fully) support `readonly => true`:

I just tried to fix this but found out that there is no `readonly` for many form inputs, MDN writes:

> Only text controls can be made read-only, since for other controls (such as checkboxes and buttons) there is no useful distinction between being read-only and being disabled, so the readonly attribute does not apply.

We could implement the functionality ourselfs pretty easily with `$this->readonly ? ' onclick="return false"' : ''` but I’m not sure if this is needed.

leofeyer:
If the input field does not support readonly, we do not have to, either, in my opinion.

ausi:
Currently we half-support it for checkboxes, any change gets lost when saving, but the user can not see that directly.

I think we should either fully support it for all fields, or not support it for non-text fields (including the save mechanism).

aschempp:
shouldn't we simply convert `readonly` to `disabled` for checkboxes?

ausi:
> shouldn't we simply convert `readonly` to `disabled` for checkboxes?

[...]

PR #4870, opened 2022-06-24
Disable widgets for configured settings

aschempp:
Fixes https://github.com/contao/contao/issues/4868

It looks like Chosen does not support disabling a select menu. I decided to fix this only in Contao 4.13 because we only disable settings there. Patching Chosen seems out of scope to me for now.


Toflar (dismissed):

bytehead (dismissed):

fritzmg:
> I decided to fix this only in Contao 4.13 because we only disable settings there.

This should be fixed in Contao 4.9 as well, as _any_ widget can be set to `readonly => true`. Also `select` with `chosen` are not the only widget where `readonly` does not work and if we want to fix them down the line as well it should also be done for 4.9 imho.

leofeyer (dismissed):

leofeyer:
The unit tests are still failing though.

fritzmg (changes requested):
> I decided to fix this only in Contao 4.13 because we only disable settings there.

This should be fixed in Contao 4.9 as well, as _any_ widget can be set to `readonly => true` in 4.9 as well, not just in 4.13. Also `select` with `chosen` are not the only widget where `readonly` does not work and if we want to fix them down the line as well it should also be done for 4.9.

aschempp:
Yes, its a general issue on multiple widgets. This PR fixes it for the system settings. Feel free to fix any other widgets you notice in 4.9, I just didn't find the time to test every widget. Also, this is about `disabled` not `readonly`, but I guess that does not really matter 😆

leofeyer (approved):

PR #4874, opened 2022-06-24
Replace the contao_figure function

### LINKED ISSUE #4831: Improve the contao_figure Twig function
### Description

The `contao_figure` function currently allows to control a `FigureBuilder` on the fly and output an image. While the first part is nice, the second should have rather output a `Figure` object that can then be used in an image template but also directly queried for attributes. Currently it's a string.

Maybe we can improve this by returning sth like a `RenderableFigure extends Figure implements \Stringable` instead that calls the `FigureRenderer` when output as a `string` but otherwise is just a `Figure`. Besides keeping BC this would have the benefit, that, if we allow `RenderableFigure` as a safe HTML object, outputting it would be possible without the need for the `|raw` filter.

m-vo:
Fixes #4831.

The `contao_figure` function currently outputs a pre-rendered string… which was a bad idea to be honest. It should have really only returned a `Figure` object. Then you could either access its properties or output it manually. 

This PR fixes this by replacing the function with a new `figure` function. This way, we also get rid of the `contao_` prefix that turned out to be unnecessary as well. I decided against removing the old function and only deprecated it for now, so that there is an easier upgrade path.

Usage:
```twig
{# before #}
{{ contao_figure('image.jpg', [800, 600]) }}

{# after #}
{% include "@Contao/component/_figure.html.twig" with {
    figure: figure('image.jpg', [800, 600])
} %}
```

And now also supported:
```twig
{% set figure = figure('image.jpg', [800, 600]) %}

{# query all the properties #}
{{ figure.image.imageSrc }} 
{{ figure.image.metadata.title }} 
…
```


Toflar (dismissed):

leofeyer:
Am I the only one to think that the "before" variant is much nicer than the "after" variant? Maybe we should keep both? @ausi WDYT?

Toflar:
The first variant is more like an insert tag which can (or maybe does already?) still exist. But for template work, the new way looks better I think.

m-vo:
> Am I the only one to think that the "before" variant is much nicer than the "after" variant?

The "before" variant is an anti-pattern, though. Pre-rendered HTML is kind of working against having templates and the various options to integrate them (`include`, `embed`, `use`, …). 

Maybe it gets a bit more obvious if you want to adjust things - because then you need to define a custom template instead of just using it inline:

```twig
{{ contao_figure('image.jpg', [800, 600], {}, "my_picture_template.html.twig"}) }}
```

```twig
{# my_picture_template.html.twig #}
{% extends "@Contao/component/_figure.html.twig" %}

{% block caption_inner %}
    {{ parent() }}
    <div>Foo</div>
{% endblock %}
```

vs:

```twig
{% embed "@Contao/component/_figure.html.twig" with {figure: figure('image.jpg', [800, 600])} %}
  {% block caption_inner %}
      {{ parent() }}
      <div>Foo</div>
  {% endblock %}
{% endembed %}
```

The variant below is also identical to how you would render a `Figure` if it was already in the context. You could even use the function to replace an existing one.

ausi:
> The variant below is also identical to how you would render a `Figure` if it was already in the context.

This is very good IMO and shows that this concept is the right one.
Creating my own figure inside a template should behave the same as one created by a controller and passed to the template.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4875, opened 2022-06-26
Improve the `debug:contao-twig` command

m-vo:
This PR improves the output of the `debug:contao-twig` command. Now, all available blocks in a template as well as a preview of the content are displayed, which is very helpful when working with a lot of templates.

Example usage: `contao-console debug:contao-twig content_element/gallery`

image

Under the hood there is a new `Twig/Inspector/Inspector` service that provides the information. For now it's a thin wrapper around information that is already available in standard Twig, but my intention is to add more stuff later that allows for a better back end experience when creating/modifying templates.


core-bundle/src/Twig/Inspector/TemplateInformation.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Inspector;
+
+use Twig\Source;
+
+/**
+ * @experimental
```
Toflar, line 18: Probably `@internal` as well?

Toflar (commented):
Nice!

core-bundle/src/Twig/Inspector/TemplateInformation.php
m-vo: IMHO this class should be accessible for users at some point. I.e it can only be constructed via the `Inspector` service, but you should be able to query it.

Toflar (dismissed):
Twig support getting better by the day! ❤️

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4876, opened 2022-06-26
Allow reading from VFS storages in the FigureBuilder

m-vo:
Currently, our image processing like the resizer cannot handle resources from streams. But we need this in order to be usable with the VFS - there, by design, only storage-local paths and streams exist, but never absolute paths.

@ausi found out, that you can retrieve a stream's metadata which besides others contains the stream type and resource URI. This makes it possible to get the local file path for streams of type `STDIO`.

This PR implements a `fromStorage()` function in the `FigureBuilder` that uses this trick to be forward compatible. Now, we can already have proper usages (e.g. in the new controllers) and, once stream support has landed, we can just switch out the implementation. :tada:


core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -265,6 +269,33 @@ public function from(FilesModel|ImageInterface|int|string|null $identifier): sel
...
+
+            return $this;
+        }
+
+        // TODO: Our image libraries cannot deal with streams, yet. However,
+        //       in case of STDIO streams, we can already be forward compatible
+        //       by reading the local file path from stream's meta data.
+        $metadata = stream_get_meta_data($stream);
+
+        if ('STDIO' !== ($type = $metadata['stream_type'])) {
```
ausi: Should we check for `wrapper_type` being `plainfile` here too?

ausi: And probably also check `Path::isAbsolute($metadata['uri'])`?

m-vo: Sounds like a sensible idea :+1: 
Done in e72484d910b2afcc3bdb362e7a3b412c3dc6e0a1.

m-vo: > And probably also check Path::isAbsolute($metadata['uri'])?

~~Are you sure relative paths would even be possible?~~ found it out myself - yes they can

ausi: The docs state `URI/filename` and I saw some comments about this not always being a path. Not sure if this can happen here, but I’d prefer to be extra cautious ☺️

m-vo: Yeah, it can indeed be relative, e.g. when you have something like `$stream = fopen('my-file', 'rb');`

Adjusted in  185f449c3daa62a7ff530cb1ccd29c415267d759.

ausi (dismissed):

core-bundle/src/Image/Studio/FigureBuilder.php
```diff
@@ -265,6 +269,34 @@ public function from(FilesModel|ImageInterface|int|string|null $identifier): sel
...
+    {
+        try {
+            $stream = $storage->readStream($location);
+        } catch (VirtualFilesystemException|UnableToResolveUuidException $e) {
+            $this->lastException = new InvalidResourceException(sprintf('Could not read resource from storage: %s', $e->getMessage()), previous: $e);
+
+            return $this;
+        }
+
+        // TODO: Our image libraries cannot deal with streams, yet. However,
```
leofeyer: What exactly is the TODO here?

core-bundle/tests/Image/Studio/FigureBuilderStub.php
```diff
@@ -67,6 +69,17 @@ public function fromImage(ImageInterface $image): FigureBuilder
         throw new \RuntimeException('not implemented');
     }
 
+    public function fromStorage(VirtualFilesystemInterface $storage, Uuid|string $location): FigureBuilder
+    {
+        if (!\is_string($location)) {
+            throw new \RuntimeException('not implemented');
```
leofeyer, line 75: Why do we allow to pass a Uuid object if we handle only strings?

core-bundle/src/Image/Studio/FigureBuilder.php
m-vo: After stream support is added to `contao/image`, we should remove this workaround and type restriction and directly pass on the stream to the resizer.

core-bundle/tests/Image/Studio/FigureBuilderStub.php
m-vo: This is only the stub file that helps mocking the image studio for integration tests like those in the `Controller` namespace. I'd only add UUID logic if this is needed. The actual `FigureBuilder` class does support UUIDs.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4878, opened 2022-06-27
Make sure `eval.rte` is a string

fritzmg:
If some DCA erroneously defined something like

```php
'eval' => ['rte' => true]  
```

the `RelLightboxMigration` will currently fail with 

```
TypeError:
substr() expects parameter 1 to be string, bool given

  at vendor/contao/core-bundle/src/Migration/Version413/RelLightboxMigration.php:96
```

This PR makes the migration more robust by checking whether the value is actually a string.

### DIFF
--- a/core-bundle/src/Migration/Version413/RelLightboxMigration.php
+++ b/core-bundle/src/Migration/Version413/RelLightboxMigration.php
@@ -93,7 +93,9 @@ class RelLightboxMigration extends AbstractMigration
             }
 
             foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
-                if (0 === strpos($fieldConfig['eval']['rte'] ?? '', 'tiny')) {
+                $rte = $fieldConfig['eval']['rte'] ?? '';
+
+                if (\is_string($rte) && 0 === strpos($rte, 'tiny')) {
                     $targets[] = [$tableName, strtolower($fieldName)];
                 }
             }
--- a/core-bundle/src/Migration/Version413/RelLightboxMigration.php
+++ b/core-bundle/src/Migration/Version413/RelLightboxMigration.php
@@ -93,7 +93,7 @@ class RelLightboxMigration extends AbstractMigration
             }
 
             foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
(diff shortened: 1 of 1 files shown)


ausi (dismissed):

leofeyer:
I don‘t understand why we have to make up for configuration errors? If you misconfigure the app, it will trigger an error. And that is a good thing, otherwise you wouldn‘t know that you did something wrong. 🤷‍♂️

m-vo:
I also think we should not fix this. The [docs](https://docs.contao.org/dev/reference/dca/fields/#evaluation) state it should be a `string` - why should this ever be a boolean?

fritzmg:
> I don‘t understand why we have to make up for configuration errors?

We want migrations to be as robust as possible.

> And that is a good thing, otherwise you wouldn‘t know that you did something wrong. 🤷‍♂️

In this particular instance the Contao user was not able to figure out the source of the problem, as the error message won't help you find the the erroneous DCA definition.

> why should this ever be a boolean?

As mentioned in the description of this PR, this can happen if you (or an extension) _erroneously_ set `eval.rte` to something other than a string, regardless of whether this was made consciously or by mistake.

ausi:
> > I don‘t understand why we have to make up for configuration errors?
> 
> We want migrations to be as robust as possible.

Agree. Migrations should be as defensive as possible IMO. They might run in very odd environments and should not fail for things like a wrong type somewhere in a DCA config.

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -93,7 +93,9 @@ private function getTargets(): array
             }
 
             foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
-                if (0 === strpos($fieldConfig['eval']['rte'] ?? '', 'tiny')) {
+                $rte = $fieldConfig['eval']['rte'] ?? '';
```
leofeyer:
```suggestion
                $rte = $fieldConfig['eval']['rte'] ?? null;
```

This way the `is_string()` check will already return false and `strpos()` will never be called for non-string values.

leofeyer (changes requested):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4881, opened 2022-06-27
Use the resource finder in the lightbox migration

### LINKED ISSUE #4803: Contao ignores dca files if any throwable occurs
### Affected version(s)

4.13.5

### Description

Since 4.13.5 the `RelLightboxMigration` loads dca containers and silences any `Throwable`. If there occurs an error (e.g. undefined array index warning) the error is ignored and Contao wants to delete existing and required database fields.

https://github.com/contao/contao/blob/e95c3202701dfbf8e240f9755d91bc47a399c863/core-bundle/src/Migration/Version413/RelLightboxMigration.php#L89-L93

leofeyer:
Fixes #4803

### DIFF
--- a/core-bundle/src/Migration/Version413/RelLightboxMigration.php
+++ b/core-bundle/src/Migration/Version413/RelLightboxMigration.php
@@ -93,13 +93,14 @@ class RelLightboxMigration extends AbstractMigration
         $files = $this->resourceFinder->findIn('dca')->depth(0)->files()->name('*.php');
 
         foreach ($files as $file) {
-            if (\in_array($file->getBasename(), $processed, true)) {
+            $tableName = $file->getBasename('.php');
+
+            if (\in_array($tableName, $processed, true)) {
                 continue;
             }
 
-            $processed[] = $file->getBasename();
+            $processed[] = $tableName;
 
-            $tableName = $file->getBasename('.php');
             Controller::loadDataContainer($tableName);
 
             foreach ($GLOBALS['TL_DCA'][$tableName]['fields'] ?? [] as $fieldName => $fieldConfig) {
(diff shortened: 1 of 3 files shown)


fritzmg:
Can we really rely on DCA files alone? In theory a full DCA could have been added via an event listener.

core-bundle/src/Migration/Version413/RelLightboxMigration.php
```diff
@@ -82,16 +86,22 @@ private function getTargets(): array
...
+        foreach ($files as $file) {
+            if (\in_array($file->getBasename(), $processed, true)) {
                 continue;
             }
 
+            $processed[] = $file->getBasename();
+
+            $tableName = $file->getBasename('.php');
+            Controller::loadDataContainer($tableName);
+
```
ausi: I think we should use the same basename here.
```suggestion
        foreach ($files as $file) {
            $tableName = $file->getBasename('.php');

            if (\in_array($tableName, $processed, true)) {
                continue;
            }

            $processed[] = $tableName;

            Controller::loadDataContainer($tableName);

```

ausi:
> In theory a full DCA could have been added via an event listener.

I don’t think that’s possible because the Installer/DcaSchema rely on the same mechanism:
https://github.com/contao/contao/blob/0e6512a5f010f6fb4d1359a2798c29eb74e3304c/core-bundle/src/Resources/contao/library/Contao/Database/Installer.php#L177

ausi:
I think we should also add something like 

```php
try {
    $this->loadDcaFiles($blnNoCache); 
} catch (\Throwable $e) {
    unset(static::$arrLoaded['dcaFiles'][$this->strTable]);
    throw $e;
}
```

to here: https://github.com/contao/contao/blob/9554021bb21e73c3a85a38c4a1d2a906b9fe2cb6/core-bundle/src/Resources/contao/library/Contao/DcaLoader.php#L68

WDYT?

leofeyer:
Added in 1851753b612108460d8083bd72ba2ef71a27f230.

ausi (approved):

PR #4886, opened 2022-06-27
Always unlock search tables to prevent deadlocks

ausi:
Currently, if there is an exception thrown in the search class between `LOCK TABLES …` and `UNLOCK TABLES` the tables stay locked until the database connection gets closed.

This can cause problems especially in long running processes like `contao:crawl`.

Thanks to @Toflar for debugging this issue.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -194,130 +194,144 @@ class Search
 		// Prevent deadlocks
 		$objDatabase->query("LOCK TABLES tl_search WRITE, tl_search_index WRITE, tl_search_term WRITE");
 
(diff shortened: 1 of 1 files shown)


Toflar:
We also moved the `UNLOCK` query above updating the vector length here which is something I figured would be faster (and keeping the lock for updating the vector length is not needed) and I can confirm this after running some tests with this PR. `contao:crawl` now runs way faster! Previously, if you had a concurrency of 20 and 19 requests that did not need to update the search index (because already up to date) they still had to wait for the 20st request to unlock the tables.

We're talking orders of magnitude here 🚀 

Best viewed without whitespace: https://github.com/contao/contao/pull/4886/files?diff=unified&w=1

Toflar (approved):

leofeyer:
Thank you @ausi.

PR #4887, opened 2022-06-27
Update the Google Analytics tracking snippet

### LINKED ISSUE #4884: Update default template for Google Analytics
### Description

Google Analytics 4 is now default, Universal Analytics is deprecated and will stop processing data in the next year (https://support.google.com/analytics/answer/11583528).

The tracking snippet in `analytics_google.html5` should probably be updated, it was last updated 8 years ago.

ameotoko:
Fixes #4884


core-bundle/src/Resources/contao/templates/analytics/analytics_google.html5
```diff
@@ -1,16 +1,16 @@
...
 
 <script>
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-  ga('create', '<?= $GoogleAnalyticsId ?>', 'auto');
-  ga('set', 'anonymizeIp', true);
-  ga('send', 'pageview');
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+  gtag('config', '<?= $GoogleAnalyticsId ?>');
```
fritzmg: How can this new script work? This does not include any external JavaScript.

ameotoko: missed a line🤦🏻‍♂️

fritzmg (dismissed):

core-bundle/src/Resources/contao/templates/analytics/analytics_google.html5
```diff
@@ -1,16 +1,17 @@
...
+<script async src="https://www.googletagmanager.com/gtag/js?id=<?= $GoogleAnalyticsId ?>"></script>
 <script>
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-  ga('create', '<?= $GoogleAnalyticsId ?>', 'auto');
-  ga('set', 'anonymizeIp', true);
-  ga('send', 'pageview');
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+  gtag('config', '<?= $GoogleAnalyticsId ?>');
```
leofeyer:
```suggestion
  gtag('config', '<?= $GoogleAnalyticsId ?>', { 'anonymize_ip': true });
```

ameotoko: This is a default now:

https://support.google.com/analytics/answer/2763052?hl=en

> In Google Analytics 4, IP anonymization is not necessary since IP addresses are not logged or stored.

https://support.google.com/analytics/answer/9310895?hl=en&ref_topic=11192706#gtagjs-configuration

> IP Anonymization is standardized to "true" and not configurable in a Google Analytics 4 property. As such, the IP address is automatically masked with the standard event to Google Analytics 4 property ("G-XXXXXXXX").

leofeyer: That only applies to GA4 properties. If someone still uses `UA-XXXXXXXX`, they will have to request IP anonymization by adding `{ 'anonymize_ip': true }`, won‘t they?

ameotoko: If someone already uses UA, they probably already have the customized template and will not be affected by this update? For new sites, GA4 is default choice, and this template provides a default configuration for that.

That being said, there is no harm in keeping a deprecated setting either. Should I still add it explicitly?

NOTE: GA4 will be the only choice for new properties after July 1st, 2023.

leofeyer: @contao/developers WDYT?

Toflar: I wonder if Contao 5 should ship this template at all. Why does it provide defaults for providers such as Google but not privacy focussed ones like plausible.io? We provide a super easy way to include whatever scripts you want to - imho that's all we need to do.

Toflar: I didn't realize this is a PR against 4.9, sorry. That being said, I don't have any preference at all. Again, I don't see why we would ship those templates and I haven't used any Google Analytics services for years.

frontendschlampe: I agree with @Toflar for Contao 5 not to ship any of the templates for analytics. Every tracking software provides the tracking code and with a simple copy & paste you can create your own tracking snippet and you already have to improve the existing templates.

For Contao 4.x I would update to GA4 because as @ameotoko said, there's no possibility creating a new GA3 property. We have to make a comment in the template that it's GA4.

ameotoko:
Sorry, deleted wrong branch :(

ameotoko:
@leofeyer can you reopen this? Or should I start the new PR?

Toflar:
There you go :)

ameotoko:
After another check, the tracking snippet is actually identical for both GA4 and UA. So I just set ip anonymization explicitly, which will work in both cases.

core-bundle/src/Resources/contao/templates/analytics/analytics_google.html5
```diff
@@ -1,16 +1,18 @@
...
+if ('XXXXX' != $GoogleAnalyticsId && !BE_USER_LOGGED_IN && !$this->hasAuthenticatedBackendUser()): ?>
 
+<script async src="https://www.googletagmanager.com/gtag/js?id=<?= $GoogleAnalyticsId ?>"></script>
 <script>
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-  ga('create', '<?= $GoogleAnalyticsId ?>', 'auto');
-  ga('set', 'anonymizeIp', true);
-  ga('send', 'pageview');
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
```
leofeyer, line 13:
```suggestion
  function gtag(){dataLayer.push(arguments)};
```

According to https://developers.google.com/tag-platform/tag-manager/web/datalayer#installation.

leofeyer (changes requested):

core-bundle/src/Resources/contao/templates/analytics/analytics_google.html5
ameotoko: They suggest my variant when you look for instructions directly while setting up a property 😁 But I suppose it doesn't matter.

image

leofeyer: I see. 😄

leofeyer (approved):

leofeyer:
Thank you @ameotoko.

PR #4888, opened 2022-06-27
Move the installer into the core

m-vo:
This PR moves the `Installer` into the core bundle and removes the ability to run migrations from the install tool.

With this change, also both the `sqlCompileCommands` hook as well as the parsing and grouping of SQL commands is removed. The order and command grouping now matches the output of doctrine's schema diff (which finally solves https://github.com/contao/contao/pull/2792#issuecomment-798210483).

The first commit (1406bc390f10a40f10885d5c33d75dfe96cce78a) is only about moving things around without changing the implementation - to get a better idea about the actual code changes, you might want to base your diff view on this commit.

/cc @ausi


core-bundle/src/Migration/CommandCompiler.php
```diff
@@ -0,0 +1,200 @@
...
+     * Execute a SQL command identified by its hash.
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function execCommand(string $hash): void
+    {
+        $command = $this->commands[$hash] ?? throw new \InvalidArgumentException(sprintf('Invalid hash: %s', $hash));
+
+        $this->connection->executeStatement($command);
+    }
```
m-vo: We could IMHO get rid of this function and also get rid of the caching of commands. IMHO this would be an even cleaner API. Wdyt?

core-bundle/src/Migration/CommandCompiler.php
```diff
@@ -13,193 +13,88 @@
...
-                if (false === $indexB) {
-                    $indexB = \count($this->commandOrder);
-                }
-
-                return $indexA - $indexB;
-            }
-        );
-
-        return $commandsByHash;
+        return $this->commands ?? ($this->commands = $this->compileCommands());
```
ausi:
```suggestion
        return $this->commands ??= $this->compileCommands();
```

core-bundle/src/Migration/CommandCompiler.php
ausi: Yes :+1: for removing this, removing the caching and removing the md5 hashes.

m-vo: see 4d2f0ee64c56a44b63deb6976be012af8913bc22

core-bundle/src/Migration/CommandCompiler.php
m-vo: see 4d2f0ee64c56a44b63deb6976be012af8913bc22

core-bundle/src/Migration/CommandCompiler.php
m-vo: Now that the md5 hashes are an implementation detail, we could likely also refactor the migrate command. But maybe in another PR?

UPGRADE.md
```diff
@@ -390,3 +390,10 @@ The public folder is now called `public` by default. It can be renamed in the `c
 The `Contao\CoreBundle\Image\Studio\Figure::getLinkAttributes()` method will now return an
 `Contao\CoreBundle\String\HtmlAttributes` object instead of an array. Use `iterator_to_array()` to transform it
 back to an array representation. If you are just using array access, nothing needs to be changed.
+
+### Install Tool
+
+The ability to run execute migrations has been removed from the Install Tool, use the `contao:migrate` command or the
+Contao Manager instead.
+
+The `sqlCompileCommands` hook hasd been removed, use a custom schema provider instead.
```
Toflar:
```suggestion
The `sqlCompileCommands` hook has been removed, use the Doctrine DBAL `postGenerateSchema` event instead.
```

core-bundle/src/Migration/CommandCompiler.php
```diff
@@ -13,193 +13,54 @@
...
-            foreach ($GLOBALS['TL_HOOKS']['sqlCompileCommands'] as $callback) {
-                $return = System::importStatic($callback[0])->{$callback[1]}($return);
-            }
-        }
+        // Get a list of SQL commands that adjust the engine and collation options
+        $engineAndCollationCommands = $this->compileEngineAndCollationCommands($fromSchema, $toSchema);
 
-        $this->commands = $return;
-        $this->commandOrder = $order;
+        return array_unique([...$diffCommands, ...$engineAndCollationCommands]);
```
ausi, line 70: Do we need this unique check?
Could there be (in the future) the same query twice like a scenario like `do a, do b, revert a, do c, do a`?

core-bundle/src/Migration/CommandCompiler.php
ausi: I’d do it in this PR, should be straight forward I think?
A quick and dirty way would be to just replace `$commands[md5($command)] = $command;` with `$commands[$command] = $command;` 😅

core-bundle/src/Migration/CommandCompiler.php
m-vo: In the old code, we manually checked for duplicates (e.g. so that add index statements are only executed once), so that's why I added the unique check in there. 

I think I would only refactor this once doctrine starts to do something like you outlined (which is quite unlikely tbh). Wdyt?

core-bundle/src/Migration/CommandCompiler.php
m-vo: Ok, I'll look into it. This will change how the `sha1` will look, but they weren't designed to be persisted, correct?

leofeyer: I think we have used md5 hashes so the array keys don‘t become huge.

aschempp:
Two questions:
 1. any particular reason we're not just moving the existing classes to the core-bundle and keep everything as is, since I assume we want to get rid of all this stuff anyway?
 2. any reason to remove stuff from the installation bundle? I mean shouldn't we just remove the installation-bundle from contao/contao in one PR?

core-bundle/src/Migration/CommandCompiler.php
ausi: and probably so that the `Input` class does’t mangle the values ☺️

ausi: > but they weren't designed to be persisted, correct?

Correct. Their main puprose is for the Contao Manager to make sure that nothing changed between reading and executing the migrations.

ausi:
> since I assume we want to get rid of all this stuff anyway?

Don’t think so, this stuff is all needed, isn’t it?

core-bundle/src/Migration/CommandCompiler.php
m-vo: I refactored the logic in 8c7b7844e4aa4012ad24b6829276f613cd8555a9 - IMHO it's much cleaner now and does not rely on identifying array keys anymore.

m-vo:
>any reason to remove stuff from the installation bundle? 

The install bundle will not work anymore without the removed command grouping logic.

UPGRADE.md
```diff
@@ -390,3 +390,10 @@ The public folder is now called `public` by default. It can be renamed in the `c
 The `Contao\CoreBundle\Image\Studio\Figure::getLinkAttributes()` method will now return an
 `Contao\CoreBundle\String\HtmlAttributes` object instead of an array. Use `iterator_to_array()` to transform it
 back to an array representation. If you are just using array access, nothing needs to be changed.
+
+### Install Tool
+
+The ability to run execute migrations has been removed from the Install Tool, use the `contao:migrate` command or the
```
fritzmg:
```suggestion
The ability to execute migrations has been removed from the Install Tool, use the `contao:migrate` command or the
```

m-vo:
I readded the ability to distinguish between all schema updates + those that remove columns in b12d4ab52fe1578801e19db6fd6788b5218688c4. We're now effectively copying the columns that should be remove to the `$toSchema` which makes them disappear in the diff.

core-bundle/src/Migration/CommandCompiler.php
```diff
@@ -28,13 +28,45 @@ public function __construct(private readonly Connection $connection, private rea
     /**
      * @return list<string>
      */
-    public function compileCommands(): array
+    public function compileCommands(bool $doNotDropColumns = false): array
```
ausi:
```suggestion
    public function compileCommands(bool $skipDropStatements = false): array
```
We should also not drop tables I guess?

core-bundle/src/Migration/CommandCompiler.php
```diff
@@ -28,13 +28,45 @@ public function __construct(private readonly Connection $connection, private rea
...
         // Get a list of SQL commands from the schema diff
         $schemaManager = $this->connection->createSchemaManager();
         $fromSchema = $schemaManager->createSchema();
         $toSchema = $this->schemaProvider->createSchema();
 
+        // If columns should not get dropped, we copy missing definitions
+        // over to the $toSchema, so that they won't appear in the diff
+        if ($doNotDropColumns) {
+            foreach ($fromSchema->getTables() as $table) {
+                $toSchemaTable = $toSchema->getTable($table->getName());
```
ausi, line 50: Do we need to check hasTable here?

core-bundle/src/Migration/CommandCompiler.php
m-vo: True that :smile:

m-vo: See 66652ccc9f931151f18ed1d0e0d6a66d23982f6d + d22dbbfb2b3883b9c41d9fe39e17633dfc8ca38e.

I also went with reflection here, because this way copying definitions is much saner. Unfortunately.

core-bundle/src/Migration/CommandCompiler.php
m-vo: see https://github.com/contao/contao/pull/4888#discussion_r908320116

core-bundle/src/Migration/CommandCompiler.php
ausi: I looked into this and reflection seems to be the best way to do this 😕

ausi (dismissed):
❤️

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4891, opened 2022-06-28
Fix the query string being lost during the preview script redirect

qzminski:
This is a follow-up on #4857. Since that pull request got merged, the frontend preview has always shown the homepage due to missing query string parameters.


Toflar (approved):

fritzmg (approved):

leofeyer:
Thank you @qzminski.

PR #4893, opened 2022-06-28
Modern fragments: article teaser + player content elements

m-vo:
Followup to #4823 

This PR adds the last missing content elements that aren't wrappers or plain includes…

* `content_element/teaser`
* `content_element/player`

… together with a  `_read_more_.html.twig` component and a formatter filter for byte values `format_bytes`.

Todo:
- [x] rebase
- [x] add tests for the player element
- [x] add tests for the `FormatterRuntime` + adjust `ContaoExtension` tests
- [x] add test for `FilesystemItemIterator#first()`

**Note:** This PR is currently based on #4862 and predecessors; needs to be rebased once merged.


ausi (dismissed):

ausi (dismissed):

core-bundle/contao/templates/_new/component/_figure.html.twig
```diff
@@ -54,7 +54,7 @@
 
         {# Caption #}
         {% block caption %}
-            {% if figure.hasMetadata|default(false) and figure.metadata.caption %}
+            {% if figure.metadata and figure.metadata.caption %}
```
leofeyer, line 57:
```suggestion
            {% if figure.metadata.caption %}
```

Would this work as well?

core-bundle/contao/templates/_new/component/_stylesheet.html.twig
```diff
@@ -19,7 +19,7 @@
 {% block stylesheet_component %}
     {% set stylesheet = stylesheet|default(_context) %}
 
-    {% if stylesheet.lazy|default(true) %}
+    {% if stylesheet.lazy is defined ? stylesheet.lazy : true %}
```
leofeyer, line 22: I liked the shorter version more. But either way, why is the default `true`? If the style sheet is lazy, it is loaded lazily. If the style sheet does not define anything, it is loaded lazily, too?

leofeyer (dismissed):

core-bundle/contao/templates/_new/component/_figure.html.twig
ausi: I think this would throw an exception because it can boil down to `null.caption`

core-bundle/contao/templates/_new/component/_stylesheet.html.twig
ausi: Yes, it defaults to true but you can pass `lazy: false` if you want to load it synchronously

ausi: > I liked the shorter version more.

The short version doesn’t work if `false` is not the default because `false|default(true)` is still true.

leofeyer: > Yes, it defaults to true but you can pass `lazy: false` if you want to load it synchronously

I would then prefer to call the option `synchronously: true` instead of `lazy: false`. WDYT?

ausi: `synchronously` would be way hard to type for non-native speakers I guess 🙈😅

Maybe `sync: true`? But I’d still prefer `lazy: false` because the term “lazy loading” is pretty well understood I think.

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #4894, opened 2022-06-29
Display an error message if the version cannot be restored

qzminski:
This change will display an error message if the version cannot be restored instead of throwing an exception. It's rather an edge case, but it can happen that the column SQL definition has changed, and the previous value of the column is no longer valid.

Example:

1. Have a `my_column` field that is nullable.
2. The record is saved, and the version data contains the `my_column=null`.
3. The `my_column` SQL definition changes; it is no longer nullable.
4. A user wants to restore the previous version, but it ends up in an error because it wants to set the `my_column=null` value on the column that can no longer be null. It's probably not the only case, and it could probably also happen if the new column length is shorter than the previous record value resulting in the "data truncated for column" error.
5. An exception is thrown.

This pull request changes point 5. to display a nice error message explaining what happened instead of throwing the exception.

CleanShot 2022-06-29 at 10 00 01

### DIFF
--- a/core-bundle/src/Resources/contao/languages/en/default.xlf
+++ b/core-bundle/src/Resources/contao/languages/en/default.xlf
@@ -243,7 +243,7 @@
         <source>Unknown username: %s</source>
       </trans-unit>
       <trans-unit id="ERR.versionNotRestored">
-        <source>Version %d could not be restored! Please consider restoring the data manually.</source>
+        <source>Version %d could not be restored. Previous data is not compatible with the current database definition anymore. Please consider restoring the data manually.</source>
       </trans-unit>
       <trans-unit id="SEC.question1">
         <source>Please add %d and %d.</source>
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/languages/en/default.xlf
```diff
@@ -242,6 +242,9 @@
       <trans-unit id="ERR.previewSwitchInvalidUsername">
         <source>Unknown username: %s</source>
       </trans-unit>
+      <trans-unit id="ERR.versionNotRestored">
+        <source>Version %d could not be restored! Please consider restoring the data manually.</source>
```
Toflar:
```suggestion
        <source>Version %d could not be restored. Previous data is not compatible with the current database definition anymore. Please consider restoring the data manually.</source>
```

Toflar (commented):
Love it! Maybe let's provide the user with a bit more information as to why it could not be restored? This would make more sense to me as a user.

Toflar (dismissed):

aschempp:
nice! Wouldn't it be useful to see the error message? At least in the log files or backend log?

Toflar:
True, we could add a system log entry on top. `Could not restore version x of y due to the following exception: %s`. Or something along those lines 👍

qzminski:
Just added logging the error to system logs 👍🏻

Toflar (approved):

fritzmg (approved):

PR #4895, opened 2022-06-29
Fix an error in page forward if the $_GET parameter is an array

qzminski:
If the requested URL contains the $_GET parameter that is an array, we'd get the `Warning: Array to string conversion` error.

Example:

`https://domain.tld/de/my-page-forward.html?categories[]=category_1`

CleanShot 2022-06-29 at 10 29 28


leofeyer:
I assume you are having the same issues as I did in #4692, aren‘t you? The invalid input comes from security tools running checks against live websites to find potential vulnerabilities.

I think we should throw an exception in this case:

qzminski:
I am not entirely sure if we should throw an exception because, in fact, it's a valid URL. The redirect logic, however, should be able to handle the arrays `key[]=param1`.

If I access the `https://domain.tld/de/my-page-forward.html?categories[]=category_1` page, I'd expect to land on `https://domain.tld/de/my-target.html?categories[]=category_1`.

fritzmg:
The same error likely also occurs in Contao 4.9? If yes then this PR should be targeted towards Contao 4.9.

qzminski:
Rebased on 4.9 :+1:

fritzmg:
I am not sure the solution is correct though. If the problem is caused by URLs like

```
https://domain.tld/de/my-page-forward.html?categories[]=category_1
```

then this should not occur in the first place, since the `categories` parameter will be ignored here:

https://github.com/contao/contao/blob/26381dc96693b8f1b01bc27040c2a149eca4a0a7/core-bundle/src/Resources/contao/pages/PageForward.php#L104-L108

And with the current PR, an URL like 

```
/lorem/lorem[]=ipsum
```

would be created, which would also not be correct, unless I am missing something?

qzminski:
> then this should not occur in the first place, since the categories parameter will be ignored here:

I have debugged it and it will NOT be ignored, because the `$key` is `categories`, whereas `$arrQuery` contains `categories[]`.

… and that is actually the real problem, I think. I have updated my pull request, please have a look.

fritzmg (changes requested):
I don't think this change is correct. With your changes, `$strGet` will then look like `…./categories[]/1`, which is not supported by Contao.

Instead I think you need to make sure that

```php
// Ignore the query string parameters (see #5867)
if (\in_array($key, $arrQuery))
```

is not true for any query parameters, including array GET parameters.

qzminski:
What if we just ignore arrays completely?

fritzmg:
I rather think line 90 needs to be changed, so that the actual name of the parameter gets added to the array, instead of `categories[]`.

qzminski:
Would you check the param name for `[]`, or the actual `$_GET` value there? 🤔

aschempp (dismissed):
I think this is correct. An array value can never have be part of the URL parameters. And the point of this is to "keep" parameters that have been in the URL when forwarding to another page.

What I find confusing is that the parameters that **are** kept in the URL are not removed from query parameters. But since that has always been the case I would just ignore it (at least in the PR).

fritzmg (dismissed):

ausi (approved):

leofeyer:
Thank you @qzminski.

PR #4897, opened 2022-06-29
Remove InitializeApplicationListener from manager bundle

### LINKED ISSUE #4896: InitializeApplicationListener wants to copy non-existing file
### Affected version(s)

5.x

### Description

When developing for Contao 5.x I got stuck with the following error:
`Failed to copy "contao/manager-bundle/src/EventListener/../Resources/skeleton/system/initialize.php" because file does not exist.`

Obviously the file does not exist anymore here:
https://github.com/contao/contao/tree/5.x/manager-bundle/src/Resources/skeleton

But it is referenced here:
https://github.com/contao/contao/blob/5.x/manager-bundle/src/EventListener/InitializeApplicationListener.php#L40

I'm not sure what's the correct solution for this problem 🤔

rabauss:
Fixes #4896


Toflar (approved):

leofeyer:
Thank you @rabauss.

PR #4898, opened 2022-06-29
Add error handling to the form data processing

rabauss:
Currently, it's not possible to pass errors to the form while executing the hook `processFormData`, e.g. in the notification center (see: https://github.com/terminal42/contao-notification_center/issues/258). That's why I added some error handling to the `Form` class.

In the notification center there needs to be added something like this:
```php
//classes/tl_form.php
public function sendFormNotification($arrData, $arrForm, $arrFiles, $arrLabels, $form)
{
...
    $form->addError('A mail error occurred!');
...
}
```

I'm not sure how to automatically test this new feature or is it not really possible? 🤔

### DIFF
--- a/core-bundle/contao/forms/Form.php
+++ b/core-bundle/contao/forms/Form.php
@@ -12,6 +12,8 @@ namespace Contao;
 
 use Contao\CoreBundle\Routing\ResponseContext\HtmlHeadBag\HtmlHeadBag;
 use Contao\CoreBundle\Session\Attribute\AutoExpiringAttribute;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 
 /**
  * Provide methods to handle front end forms.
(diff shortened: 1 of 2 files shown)


rabauss:
Any news here? Or should we wait for the notification center v2?

Toflar:
How is that related to core?

rabauss:
It's not really core related, but maybe there will be new experiences with notification center v2 🤔 
Although, the missing error handling in the core will only be solved with the idea of this PR 🚀

Toflar (commented):
I like the PR for sure but do we really have to depend on a doctrine collection here? I don't really see a benefit in this very case? 🤔

rabauss:
Well, we could also use a simple array. I just wanted to be a little bit modern :-D
What would you prefer?

Toflar (commented):
I understand but I don't really see an advantage the way you use it right now, do you?
I think a simple array would be sufficient in this old piece of code. I'd love to see this in 5.1 :)

rabauss:
Okay I refactored the `ArrayCollection` to a simple `array` 🚀

Toflar (dismissed):
Love it, the CI on `5.x` is currently broken so I guess unrelated.
Approved - thank you! <3

Toflar (approved):

leofeyer:
Thank you @rabauss.

PR #4899, opened 2022-06-29
Use `disable=0` instead of `disable!=1`

leofeyer:
Now that this column is a boolean, we can safely use `disable=0` instead of `disable!=1`.

### DIFF
--- a/core-bundle/src/Controller/BackendPreviewSwitchController.php
+++ b/core-bundle/src/Controller/BackendPreviewSwitchController.php
@@ -169,7 +169,7 @@ class BackendPreviewSwitchController
             WHERE
                 username LIKE ? $andWhereGroups
                 AND login=1
-                AND disable!=1
+                AND disable=0
                 AND (start='' OR start<='$time')
                 AND (stop='' OR stop>'$time')
             ORDER BY
--- a/core-bundle/src/Resources/contao/classes/BackendUser.php
+++ b/core-bundle/src/Resources/contao/classes/BackendUser.php
@@ -250,7 +250,7 @@ class BackendUser extends User
 
 		foreach ((array) $this->groups as $id)
 		{
-			$objGroup = $this->Database->prepare("SELECT * FROM tl_user_group WHERE id=? AND disable!=1 AND (start='' OR start<='$time') AND (stop='' OR stop>'$time')")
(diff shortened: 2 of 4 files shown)


fritzmg:
I think we should leave all queries at `!=1`, just in case someone did override the SQL definition of a boolean field and thus still might have a string type. Unless there is a performance difference?

core-bundle/src/Controller/BackendPreviewSwitchController.php
```diff
@@ -169,7 +169,7 @@ private function getMembersDataList(BackendUser $user, Request $request): array
             WHERE
                 username LIKE ? $andWhereGroups
                 AND login=1
-                AND disable!=1
+                AND disable=0
```
ausi, line 172:
```suggestion
                AND !disable
```
How about using boolean expressions?

leofeyer: Does this work in all MySQL versions?

ausi: Yes (tested with 5.5 and up)
https://www.db-fiddle.com/f/66jye1UoJzVkL52qdrySne/0

leofeyer: Like this? 60c0f0f1ba7bfa87a54c4b7324f0087b5ca7764c

Would this also work for `login AND !disable AND published` instead of `login=1 AND !disable AND published=1`?

ausi: Yes. To me this would be the nicest syntax, not sure if it is the fastest though.

ausi: It seems for performance reasons we should use `=0` and `<>0` 😕
https://stackoverflow.com/a/52073119/1031606
(But we should verify this if it is acutally the case in mariadb or mysql 8)

ausi: Verified: https://www.db-fiddle.com/f/sskh7B7ZRiRCSeM92NZUep/1
We should use `=1` and `=0` for best performance.

So my suggestion needs to be reverted (sorry 🙈).

leofeyer: No biggie. I have reverted the changes.

ausi (approved):

PR #4903, opened 2022-06-29
Show until/from values for content elements

### LINKED ISSUE #3826: Show start/stop publishing status for content elements
### Description

Since we are going down this road: #3784, I would love to see `start`/`stop` values here as well😁

image

Should I make a PR?

ameotoko:
Fixes #3826 

This is a basic implementation, as discussed here: https://github.com/contao/contao/issues/3826#issuecomment-1064219220

image

image


aschempp:
I wonder if it ever makes sense to show both values? If an element starts in the past, I don‘t care about that because it will currently be visible. 

I would probably show this
- Start date only (no stop date) if it is invisible and starts in the future
- Stop date if it is visible and will be stopped
- Stop date if it is invisible because of the stop date

What do you think?

ameotoko:
Good idea. That would not solve the issue though.

aschempp:
> That would not solve the issue though.

What issues would it not solve?

ameotoko:
See here: https://github.com/contao/contao/issues/3826#issuecomment-995566286

> The idea is to eliminate extra clicks just to see if there are any start/stop values assigned to an element. This is actually mentioned in the original [feature request](https://community.contao.org/de/showthread.php?65548-Anzeigen-von-und-Anzeigen-bis-Element-unsichtbar-sichtbar-setzen-m%C3%B6glich):

> > Sonst muss man in jedes Element rein gehen -und dort nachsehen - was bei 60-70 Elementen recht zeitaufwändig ist.

I thought this was already discussed in the Contao call?

fritzmg (dismissed):
> * Start date only (no stop date) if it is invisible and starts in the future

imho it should show both start and stop in that case. Which essentially would mean that the stop date is always shown (if set). And for consistency I think we should then show both start and stop date if set (regardless of whether the start date has been reached).

leofeyer (approved):

leofeyer:
Thank you @ameotoko.

PR #4904, opened 2022-06-30
Fix the "click2edit" function

### LINKED ISSUE #4902: "Click to edit" is not working properly in 5.x
### Affected version(s)

5.x

### Description

Following up the #4754, it breaks the `click2edit` feature, that has been in Contao [since 3.1](https://contao.org/en/news/contao-3_1_RC1.html) (which is undocumented BTW) in a following way:

- `[Ctrl] + click`: ~~edit the element~~ now edits the element settings
- `[Ctrl] + [Shift] + click`: ~~edit the element settings~~ does not work

It can be fixed by adding missing class in [tl_article.php](https://github.com/contao/contao/blob/94c2febf9e9da4aca45cc39077ee12e229014c9e/core-bundle/src/Resources/contao/dca/tl_article.php):

```diff
'operations' => array
	(
		'edit' => array
		(
			'href'                => 'act=edit',
			'icon'                => 'edit.svg',
			'button_callback'     => array('tl_article', 'editHeader')
		),
		'children' => array
		(
			'href'                => 'table=tl_content',
			'icon'                => 'children.svg',
+++			'attributes'          => 'class="editheader"',
			'button_callback'     => array('tl_article', 'editArticle')
		),
```

but then the functionality will be reversed, so a decision is needed here – either accept and document the reversed functionality, or maybe interchange it in `hover.js` here:

https://github.com/contao/contao/blob/94c2febf9e9da4aca45cc39077ee12e229014c9e/core-bundle/src/Resources/contao/themes/flexible/hover.js#L72-L84

ameotoko:
This fixes #4902 in a way, that reverses the existing functionality.

Although I personally don't like changing my 10-years habits, I must admit that this is probably the right way to fix the issue, because then the behaviour would always be consistent – you can now always expect to get an `edit` form, when you `ctrl-click` on a row in any data container. If the row has children, like an article with content elements – you can get there with `ctrl-shift-click`.


aschempp (approved):
I agree changing the behavior - and working with Contao 4 and 5 at the same time for different clients - will be a pita. But I think the click change is consistent with UI change

leofeyer:
Thank you @ameotoko.

PR #4905, opened 2022-06-30
Remove redundant type casts

leofeyer:
Now that our model class converts integers and booleans, we no longer need these type casts.


ausi (approved):
😎

PR #4907, opened 2022-06-30
Update the tools and run the code inspector

leofeyer:


core-bundle/src/Command/DebugContaoTwigCommand.php
```diff
@@ -154,7 +154,7 @@ private function formatMultiline(string $attribute, string $multilineValue, Tabl
         $lines = explode("\n", $multilineValue);
         $rows = [];
 
-        for ($i = 0; $i < \count($lines); ++$i) {
+        for ($i = 0, $c = \count($lines); $i < $c; ++$i) {
```
ausi: Not a big fan of this optimization ☺️

ausi (dismissed):

core-bundle/src/Command/DebugContaoTwigCommand.php
leofeyer: Why not?

ausi: Makes the code less readable to me, lets me check if $lines will get modified in the loop, and the performance benefit is negligible IMO.

If this gets merged, I’d vote for renaming `$c` to `$count` ☺️

core-bundle/src/Command/DebugContaoTwigCommand.php
```diff
@@ -154,7 +154,7 @@ private function formatMultiline(string $attribute, string $multilineValue, Tabl
...
             $rows[] = [0 === $i ? $attribute : '', new TableCell($lines[$i], ['style' => $style])];
         }
```
ausi:
```suggestion
        foreach ($lines as $index => $line) {
            $rows[] = [0 === $index ? $attribute : '', new TableCell($line, ['style' => $style])];
        }
```

But not really relevant to this PR, so feel free to ignore.

core-bundle/src/Command/DebugContaoTwigCommand.php
leofeyer: If you search the codebase for `for ($i = 0, $c = \count(` and `for ($i = 0, $c = count(`, you will see that this is quite a common pattern.

core-bundle/src/Command/DebugContaoTwigCommand.php
leofeyer: Changed in 8b476f40ab555c8592b59f9e77f2ca84af590f72.

core-bundle/src/Command/DebugContaoTwigCommand.php
m-vo: Count is an extremely fast operation on standard iterables. The only time I would consider this, is if an array object would be used with an expensive custom count operation.

I second that this is less readable and probably nor faster at all (see https://3v4l.org/A1BWH/perf vs https://3v4l.org/HgJgm/perf).

core-bundle/src/Command/DebugContaoTwigCommand.php
leofeyer: And I have also changed the `for` loop in the `SimpleTokenParser` class in 11abda33018da2913f5f02e59fcb23887750f6f0. 😉

core-bundle/src/String/SimpleTokenParser.php
```diff
@@ -168,8 +168,6 @@ private function getVariables(string $expression): array
 
             // Skip functions
             if (isset($tokens[$i + 1]) && '(' === $tokens[$i + 1]->value) {
-                ++$i;
-
```
m-vo: This change is not functionally equivalent anymore, because now the next iteration *will* be executed, while it would be skipped before.

leofeyer: However, it doesn't seem to make any difference. Or maybe the unit tests do not cover this completely?

m-vo: Yeah, you cannot really write a unit test for this. Here, we already test the next item - a `(` - at this point and we know the context (i.e. its belonging to a function), it doesn't make sense to reevaluate the character again in the next iteration. :slightly_smiling_face: 

Also, I'm pretty convinced that a `for` loop is the right thing to do here. The loop branches have a linear dependency, while a `foreach` loop should in theory be parallelizable.

leofeyer: I have reverted the change.

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -216,7 +216,7 @@ abstract class DataContainer extends Backend
 
 	/**
 	 * Active record
-	 * @var \stdClass|null
+	 * @var Model|FilesModel|object|null
```
m-vo: Just a question: why `Model` and `FilesModel` if every `FilesModel` is a `Model`?

m-vo (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: Fixed in 418f447fa067631cc2ba98d09c4d227d45debb05.

m-vo (approved):

PR #4908, opened 2022-06-30
Pass the request object to the Environment::phpSelf() method

leofeyer:
This fixes the `Required parameter $request missing` error.


ausi:
This was a mistake from #4510 then, right?
Should we verify that every `Environment` method is at least called once in the unit tests? And add the missing tests?

leofeyer:
Not sure if that is worth the effort. I don‘t expect any more changes to the class except when we eventually remove it in one of the next major versions.

ausi (approved):

aschempp (approved):

PR #4910, opened 2022-06-30
Fix the click2edit behavior

leofeyer:
Fixes #4904

I have merged #4904 too quickly, I‘m afraid. There really is no reason to break the click2edit behavior in Contao 5; it already takes some time to get used to the new icons! This PR restores the previous behavior: `ctrl-click` takes you to the child records view and `ctrl-shift-click` takes you to the edit view.


fritzmg:
I never knew this functionality existed 🙈. Thus intuitively I would expect `ctrl-click` to edit the element and `ctrl-shift-click` to show the child elements though.

ameotoko:
@leofeyer this breaks the `ctrl-click` for any DCA list view, other than `MODE_PARENT`, though.

aschempp:
Please also test this behaviour with any old DCA where `editheader` is still a valid operation.

PR #4912, opened 2022-06-30
Remove the "move" operation from DC_Table

aschempp:
The move operation swaps two adjacent records in sorting. The operation has been deprecated in 4.13 (https://github.com/contao/contao/pull/4662) and has not been used for years (I didn't even know it existed).

I noticed this because the `DC_Table::move()` operation did not have any permission checks 😉

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Backend.php
+++ b/core-bundle/src/Resources/contao/classes/Backend.php
@@ -408,7 +408,7 @@ abstract class Backend extends Controller
 				case 'cutAll':
 				case 'copy':
 				case 'copyAll':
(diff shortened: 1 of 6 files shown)


aschempp:
Hmmm… the `move()` operation seems to be part of the interface... Would you agree to remove it from there?

Toflar:
We can, we can also leave it there and have it do nothing. Frankly, this interface never served any purpose at all because it's not a contract at all. It says there is a `move()` method but it has no arguments, doesn't define what moves what etc. It's completely pointless.

leofeyer (approved):

PR #4915, opened 2022-06-30
Modern fragments: template for the markdown element

m-vo:
This PR adds a modern template for the existing markdown element/fragment controller.

Note, that you cannot use the `ce_markdown.html5` template anymore without manually including it in the `content_element/markdown.html.twig` template.


ausi (approved):

leofeyer:
Thank you @m-vo.

PR #4919, opened 2022-06-30
Do not auto-save DC record

### LINKED ISSUE #1328: Backend: "Zurück"-Button im Bearbeitungsmodus sollte nie eine Speicherung von Änderungen auslösen
**Affected version(s)**
getestet: 4.4, 4.8, 4.9-RC2

**Description**
Wird im Backend z.B. eine Seite oder ein Artikel bearbeitet und eine Änderung durchgeführt, der Bearbeitungsmodus dann jedoch ohne zu speichern (Buttons unten) mit "Zurück" (rechts oben) verlassen, werden die Änderungen (ungewollt) gespeichert.
Die "Zurück"-Schaltfläche sollte wirklich nur "zurück" gehen, ohne Änderungen zu speichern.

**How to reproduce**
Beispiel:
1. Layout -> Seitenstruktur -> eine Seite bearbeiten
2. Unter "Zugriffsrechte" die Schaltfläche "Zugriffsrechte zuweisen" anhaken
3. Bearbeitungsmodus mit der "Zurück"-Schaltfläche rechts oben verlassen
4. Kontrolle durch erneutes Bearbeiten der Seite -> Häkchen bei "Zugriffsrechte zuweisen" ist da, obwohl die Änderung nicht explizit mit den "Speichern"-Buttons bestätigt wurde

### LINKED ISSUE #4900: Broken subpalette toggling in 5.x
### Affected version(s)

5.x

### Description

Toggling subpalettes currently seems to be broken in `5.x`:

image

(same problem in `tl_content` when editing content elements, for example)

I didn't get to the bottom of this yet, but the hint might be, that the whole palette is returned in the response, instead of requested subpalette, see screenshots below:

`5.x`:

image

`4.13`:

image

The problem appears on a clean fresh install of `5.x-dev`

aschempp:
While trying to fix https://github.com/contao/contao/issues/4900 I noticed that we can use the new `getCurrentRecord` cache as well as the `submit()` method to **not store the data on `AUTO_SUBMIT`! This will be a major behaviour change/improvement, since a record is only saved if you really hit the save button.

Also fixes two errors in DC_Table (line 2309) and DataContainer (line 1786).

Fixes https://github.com/contao/contao/issues/1328
Fixes https://github.com/contao/contao/issues/4900
Improved https://github.com/contao/contao/issues/3547 by not creating a new versions


core-bundle/src/Resources/contao/classes/Ajax.php
```diff
@@ -327,41 +327,25 @@ public function executePostActions(DataContainer $dc)
...
 					else
 					{
-						$objVersions = new Versions($dc->table, $dc->id);
-						$objVersions->initialize();
-
-						$this->Database->prepare("UPDATE " . $dc->table . " SET " . Input::post('field') . "='" . ((Input::post('state') == 1) ? 1 : 0) . "' WHERE id=?")->execute($dc->id);
-
-						$objVersions->create();
+						$row = $dc->getCurrentRecord();
+						$row[Input::post('field')] = (Input::post('state') == 1) ? 1 : 0;
```
Toflar, line 342: I'm not sure this is correct. You are not supposed to fiddle around with the current record because that is supposed to represent the record as it currently is stored in the database. I think what you're looking for is setting this value to `$arrSubmit` but not call `submit()` yet. That way, I will still get the correct values when my listener calls `$dc->getCurrentRecord()`. See what I mean?

Toflar: What you're doing here is kind of introducing the active record pattern again which we certainly should avoid :)

aschempp: `$dc->getCurrentRecord()` never includes the submitted data. The palette rendering checks the current record data, and this is the easiest way to manipulate that. I agree its not beautiful, but then again it's `DC_Table` 😂

Toflar: But what if the palette rendering would do `$fieldValue = $this->arrSubmit[$field] ?? $this->getCurrentRecord()[$field]`?

Toflar: Or in other words: We want to put the DC in a state and render the palette for that state. We don't want to submit it, nor fiddle around with the current record. It's basically just `->save()` or `->row()` without submit 🤔

Issue #4409, opened 2022-03-29
Consider keeping unknown insert tags as plain strings


m-vo:
### Description

Currently, insert tags that do not exist are replaced with an empty string. We should IMHO consider changing this in Contao 5 and rather treat unknown insert tags as plain strings. Example: `<{{foobar}}div>` will then remain `<{{foobar}}div>` instead of resulting in a valid div tag. 

Yes, it has always been like this, but it will make sanitization simpler and lower the attack surface, which, to me, is more important.

cliffparnitzky:
Generally I'm am on both sides. On the one hand it is great, that a non existing insert tag doesn't destroy the output in the frontend. On the other hand it is very hard to find broken insert tags. Maybe it is an idea to implement something like a switch, e.g. in production mode inserttags are replaced like it has always been and in debug mode they are not replaced with an empty string.

m-vo:
> On the other hand it is very hard to find broken insert tags.

Unknown insert tags are already logged.

> Maybe it is an idea to implement something like a switch, e.g. in production mode inserttags are replaced like it has always been and in debug mode they are not replaced with an empty string.

If necessary, I'd rather have a config switch than a different output depending on the environment. Again for sanity. :slightly_smiling_face:

fritzmg:
> > On the other hand it is very hard to find broken insert tags.
> 
> Unknown insert tags are already logged.

@m-vo the logging does not really help in finding them.

m-vo:
I don't understand the problem, if an insert tag is not replaced with an empty string, but kept, the front end won't be "destroyed".  You can simply hit ctrl+f and search for double curly braces to find them?

aschempp:
> but it will make sanitization simpler and lower the attack surface

Can you explain that?

In my (current) opinion, removing them is correct. We are building a CMS that is supposed to generate useful webpage content **for visitors**, not for developers or sanitizers. If it is hard to debug, we need to improve debugging options (like log the current page, a PR has already been drafted). We should jot make the journey for visitors worse.

m-vo:
Is this evil? Try checking for this: 
`ev{{non-existing}}il`

I do neither get why it's harder to debug something that you can see instead of something that you can't nor why this worsens the experience for visitors. Can you give an example where this would be the case?

leofeyer:
I‘m for keeping unknown insert tags visible, too.

aschempp:
> Can you give an example where this would be the case?

[...]

PR #4920, opened 2022-06-30
Keep unknown insert tags as plain strings

ausi:
See #4409


leofeyer (approved):

leofeyer:
Thank you @ausi.

Toflar:
Imho that should also be documented in the `UPGRADE.md`. It's kind of important.

leofeyer:
Not sure. The `UPGRADE.md` is meant to document breaking changes IMHO and the PR does not break anything.

m-vo:
Yeah, but people are using `Controller::replaceInsertTags('…')` as an argument for functions and this might now break. Not saying that you should do this at all. :smile:

PR #4928, opened 2022-07-01
Add a template helper method to prefix relative URLs

### LINKED ISSUE #4871: Issue with relative URLs
### Affected version(s)

5.x

### Description

In #4851 we enabled the possibility to remove the `<base>` tag by outputting only absolute URLs. However, this does not work for any custom fields. For example:

```php
// contao/dca/tl_content.php

$GLOBALS['TL_DCA']['tl_content']['fields']['someLink'] = [
    'label' => ['Some link', 'Some link relating to the text.'],
    'inputType' => 'text',
    'eval' => ['rgxp' => 'url'],
    'sql' => ['type' => 'string', 'length' => 2048, 'default' => ''],
];

PaletteManipulator::create()
    ->addField('someLink', 'text_legend', PaletteManipulator::POSITION_APPEND)
    ->applyToPalette('text', 'tl_content')
;
```

```php
<!-- templates/ce_text.html5 -->
<?php $this->extend('ce_text'); ?>

  <?php $this->parent(); ?>

  <?php if ($this->someLink): ?>
    <p class="someLink"><a href="<?= $this->someLink ?>">Some Link</a></p>
  <?php endif; ?>
<?php $this->endblock(); ?>
```

This will still output a relative URL and thus would not work if the `<base>` tag was removed.

In my original PR we discussed, that for any back end input field that can hold a URL a `save_callback` needs to be implemented so that the URL is always saved with `{{env::base}}` (or whichever insert tag is the correct one now) prefixed, if it's not an absolute URL (or insert tag) yet.

leofeyer:
Fixes #4871

I am open for a different method name.

@m-vo I am not sure if adding another dependency to the `ContaoExtension` class was the correct thing to do. If not, please point me in the right direction and I will adjust the PR accordingly.

### DIFF
--- a/core-bundle/src/Twig/Runtime/UrlRuntime.php
+++ b/core-bundle/src/Twig/Runtime/UrlRuntime.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Twig\Runtime;
+
+use Contao\Validator;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Twig\Extension\RuntimeExtensionInterface;
+
+final class UrlRuntime implements RuntimeExtensionInterface
+{
+    /**
+     * @internal
+     */
+    public function __construct(private RequestStack $requestStack)
+    {
+    }
+
+    public function prefixUrl(string $url): string
+    {
+        if (!Validator::isRelativeUrl($url)) {
+            return $url;
+        }
+
+        return $this->requestStack->getMainRequest()?->getBasePath().'/'.$url;
+    }
+}
--- a/core-bundle/src/Resources/config/services.yaml
+++ b/core-bundle/src/Resources/config/services.yaml
@@ -950,6 +950,7 @@ services:
         arguments:
             - '@twig'
             - '@contao.twig.filesystem_loader'
+            - '@request_stack'
             - '@contao.csrf.token_manager'
 
     contao.twig.fail_tolerant_filesystem_loader:
--- a/core-bundle/src/Resources/config/services.yaml
+++ b/core-bundle/src/Resources/config/services.yaml
@@ -950,7 +950,6 @@ services:
         arguments:
             - '@twig'
             - '@contao.twig.filesystem_loader'
-            - '@request_stack'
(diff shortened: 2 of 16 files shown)


m-vo:
> @m-vo I am not sure if adding another dependency to the ContaoExtension class was the correct thing to do. If not, please point me in the right direction and I will adjust the PR accordingly.

I think we should create another runtime for this - then it will only get instantiated if needed and can have arbitrary dependencies:

```php
class UrlRuntime implements RuntimeExtensionInterface { // or whatever name fits
   public function __construct(private readonly RequestStack $requestStack) {}

    public function prefixUrl(string $url): string {
        // …
    }
}
```

And then reference it like the other runtimes.

leofeyer:
Changed in b2b075581874a2f2ef26e5e8e7c2ba51f4963d40.

ausi (dismissed):

core-bundle/src/Twig/Runtime/UrlRuntime.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+    }
+
+    public function prefixUrl(string $url): string
+    {
+        if (!Validator::isRelativeUrl($url)) {
+            return $url;
+        }
+
+        return $this->requestStack->getMainRequest()?->getBasePath().'/'.$url;
```
m-vo, line 34: Is the leading slash correct in case there is no main request?

ausi: IMO Yes :)

m-vo (commented):
Looks good to me!

Two small things:
1) What about adding a small test for the runtime? :grimacing: 
2) Are the CS changes to the tests intentional or a leftover of the initial version?

core-bundle/src/Twig/Runtime/UrlRuntime.php
leofeyer: Yes.

leofeyer:
> 1. What about adding a small test for the runtime?

I don‘t think that makes sense. Both the validator and the `Request` class are tested, so we can safely assume that the method does what we expect it to do.

**Edit**: I have added tests anyway in 242457344362400c21df0421064ce3df6097dd7c. 😄 

> 2. Are the CS changes to the tests intentional or a leftover of the initial version?

The are leftover from the initial version. Do you want me to revert?

ausi (approved):

m-vo:
:+1:

PR #4930, opened 2022-07-01
Do not set HTTP header in Versions::addToTemplate

fritzmg:
`Versions::addToTemplate`, which is used by the Contao Dashboard, sets a `HTTP/1.1 404 Not Found` header if the given page is outside the scope of available pages of versions.

However, this does not actually work. The main response will still actually be `200`.

As we want to move away from using `header()` (and thus deprecate and remove the `MergeHttpHeadersListener`) anyway - and throwing a not found exception there is not particularly useful imho, I propose to clamp the value instead. This also mirrors the behaviour in `Contao\Pagination`.


Toflar (approved):

leofeyer (changes requested):
This change is wrong. We want to render the 404 page if the URL is invalid and not silently fall back to `vp=3` if someone calls `vp=4`.

fritzmg:
> We want to render the 404 page if the URL is invalid

There is no 404 page in the back end. Only the pretty error page would be shown.

> and not silently fall back to vp=3 if someone calls vp=4

Currently the dash board simply shows no version records then.

fritzmg:
Also consider the following example:

1. You log into the Contao back end and visit the dashboard.
2. Meanwhile the last view pages of the versions history get purged.
3. Now you click on the "last" link in the pagination of the version list in the dashboard.

With your proposed change you would see an error screen, even though you did nothing wrong. With the change of this PR the last page would be shown.

leofeyer:
> I propose to clamp the value instead. This also mirrors the behaviour in `Contao\Pagination`.

I don‘t know what you mean by this, but if the pagination value is invalid, Contao renders the 404 page in the front end (e.g. see https://contao.org/de/news.html?page_a3=71 or https://contao.org/de/news.html?page_a3=0) and the back end is supposed to behave the same. Maybe use the existing `PageNotFoundException` or something alike.

fritzmg:
> I don‘t know what you mean by this

`Contao\Pagination` clamps the value: 

https://github.com/contao/contao/blob/69419e07426a4d898708658862bc95b4bb8cc475/core-bundle/src/Resources/contao/library/Contao/Pagination.php#L241-L244

> and the back end is supposed to behave the same

I disagree there. So far we have not done that in the back end in this particular instance and as explained above, it can be detrimental to the user experience while also serving no purpose.

In the front end it's a different story, as the amount of pages typically does not suddenly get reduced on a regular basis during a session. And in the front end you want to tell the visitor (or a crawler) that a certain URL is invalid.

fritzmg:
@leofeyer just to give a further example: if you have pagination in a list in the back end and you manipulate the pagination URL to a page that does not exist, e.g.:

```
http://example.com/contao?do=news&table=tl_news&id=1&lp=9999999
```

no error will be shown in the back end. Instead the closest valid page will be shown, just like in this PR.

ausi (approved):

leofeyer:
> Instead the closest valid page will be shown, just like in this PR.

No, the page number will not be altered at all in this case:

https://github.com/contao/contao/blob/dbd44e945bbc8b791daf81c58fbafa0559c46255/core-bundle/src/Resources/contao/drivers/DC_Table.php#L6021-L6025

Maybe we should first define which behavior we want and then adjust all pagination menus accordingly.

leofeyer:
As discussed in the Contao call, we want to implement the feature as suggested.

ausi (approved):

leofeyer:
Thank you @fritzmg.

leofeyer (approved):

PR #4931, opened 2022-07-01
Remove uncached insert tag flag

ausi:
Insert tag flags are used for post processing.

With #4779 it will no longer be possible to implement flags that do something completley different like the `|uncached` flag does. Therefore we have to replace this flag with a proper insert tag now in order to not break BC once #4779 is finished.

Deprecated the flag in Contao 4.13: #4932

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,7 +4,7 @@
 
 ### Insert tag flag uncached
 
-The `|uncached` insert tag flag was removed. Use the `{{uncached::*}}` insert tag instead.
+The `|uncached` insert tag flag was removed. Use the `{{fragment::*}}` insert tag instead.
 
 ### Insert tag hooks
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,10 @@
 
 ## Version 4.* to 5.0
 
+### Insert tag flag uncached
+
+The `|uncached` insert tag flag was removed. Use the `{{uncached::*}}` insert tag instead.
+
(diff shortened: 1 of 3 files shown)


Toflar (dismissed):

m-vo:
I wonder if there is a more meaningful name than `uncached`. 🤔 I never really got what this meant.

Maybe `esi` or `isolate` or something?

ausi:
The current behavior tries to render it as `esi` if enabled otherwise it falls back to inline rendering.

As it uses the `fragment.handler` service to do that, maybe `{{fragment::*}}` or `{{esi_fragment::*}}`

If we use `esi_fragment` there might be an opportunity to add `hinclude_fragment` later on too.

ausi:
Named it `{{fragment::*}}` now. 

Additional renderes might be available via parameter in the future like `{{fragment::renderer=hinclude::*}}` 
(Idea by @Toflar)

UPGRADE.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 4.* to 5.0
 
+### Insert tag flag uncached
+
+The `|uncached` insert tag flag was removed. Use the `{{uncached::*}}` insert tag instead.
```
leofeyer:
```suggestion
The `|uncached` insert tag flag was removed. Use the `{{fragment::*}}` insert tag instead.
```

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #4933, opened 2022-07-01
Do not set header in Ajax::executePostActions

fritzmg:
This gets rid of the last remaining `header()` call (after #4930 is merged upstream), apart from the maintenance mode in the entry points and in `NativeHeaderStorage`. This should allow us to deprecate and disable the `MergeHttpHeadersListener` (we can probably only remove it in Contao 6.0).

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Ajax.php
+++ b/core-bundle/src/Resources/contao/classes/Ajax.php
@@ -177,8 +177,6 @@ class Ajax extends Backend
 	 */
 	public function executePostActions(DataContainer $dc)
 	{
-		header('Content-Type: text/html; charset=' . System::getContainer()->getParameter('kernel.charset'));
-
 		// Bypass any core logic for non-core drivers (see #5957)
 		if (!$dc instanceof DC_File && !$dc instanceof DC_Folder && !$dc instanceof DC_Table)
 		{


fritzmg:
On second thought: should these changes may be go into 4.13 instead?

core-bundle/src/Resources/contao/classes/Ajax.php
```diff
@@ -158,7 +158,9 @@ public function executePreActions()
 	 */
 	public function executePostActions(DataContainer $dc)
 	{
-		header('Content-Type: text/html; charset=' . System::getContainer()->getParameter('kernel.charset'));
+		$response = (new Response('', 200, array('content-type' => 'text/html')))
+			->setCharset(System::getContainer()->getParameter('kernel.charset'))
+		;
```
aschempp: this is unnecessary because it is already the default for all responses. I think we can just drop this line and not change anything else...

aschempp (changes requested):

fritzmg:
Changed to just removing the `header` (see https://github.com/contao/contao/pull/4933#discussion_r913699214) and rebased to 4.13. I did not rebase to 4.9 as we fetch the character set from `Config::get('characterSet')` there and that could in theory really be different from the default response character set.

ausi (approved):

aschempp (approved):

Toflar (approved):

leofeyer:
Thank you @fritzmg.

PR #4936, opened 2022-07-02
Support the native date input type for text fields in the form generator

fritzmg:
This enables support of the native [`date`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) HTML5 input type.


core-bundle/src/Resources/contao/forms/FormText.php
```diff
@@ -189,6 +191,9 @@ public function __get($strKey)
 					case 'url':
 					case HttpUrlListener::RGXP_NAME:
 						return 'url';
+
```
m-vo: The comment about https://github.com/contao/core/issues/5918 in L 171 should probably be adjusted/addressed as well.

aschempp (commented):
How does this behave for browsers that don't support a date picker?

fritzmg:
> How does this behave for browsers that don't support a date picker?

Internet Explorer will display it as a regular text field and you will need to input the date in the configured date format manually (as before).

aschempp (dismissed):
I probably wouldn't use a constant that is only used in one if-condition and `protected` available to child classes, but it sure will work this way 😊

fritzmg:
> and `protected` available to child classes

It should be protected, as you are able overwrite this function in a child class and you might want to further alter the implementation.

aschempp:
> It should be protected, as you are able overwrite this function in a child class and you might want to further alter the implementation.

Not sure I get that. The implementation uses `self::CONSTANT` so you couldn't change it in a child class? Also, why would you change the HTML standard? Thinking of it, maybe the constant should be part of the date class instead?

fritzmg:
> Not sure I get that. The implementation uses self::CONSTANT so you couldn't change it in a child class?

I mean in a Widget you typically overwrite the parent methods, specifically `Widget::validator` for example. And you might want to access that constant in your own method, thus it needs to be public or protected, not private.

Toflar (dismissed):

core-bundle/contao/forms/FormText.php
```diff
@@ -245,6 +251,18 @@ protected function validator($varInput)
 		{
 			$varInput = Idna::encodeEmail($varInput);
 		}
+		elseif ($this->rgxp == 'date')
+		{
+			$targetFormat = Date::getNumericDateFormat();
+
+			// Check if date format matches the HTML5 standard
+			if (self::HTML5_DATE_FORMAT !== $targetFormat && preg_match('~^' . Date::getRegexp(self::HTML5_DATE_FORMAT) . '$~i', $varInput ?? ''))
```
leofeyer, line 258: Why the `preg_match()` check?

fritzmg: `\DateTimeImmutable::createFromFormat` will fail if the input format does not match the expected format. Hence the check so that the date format is only transformed, if the input is actually valid.

leofeyer: What happens if the input is invalid?

fritzmg: Then `Widget::validator` will add an error (as usual).

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4935, opened 2022-07-01
Remove the install tool

m-vo:
This PR moves the last bits of the install tool into the `contao:migrate` command [951ddb4593e04d68295a6827de77b2db9e1af74f] and completely removes the `installation-bundle` [2c80e6397387ea41cc426ea093ed1dbc36903f06].

The migrate command now checks for configuration errors on every run unless you specify `--no-check`. It also outputs configuration warnings (right now for the strict mode) when migrations are run. We need to make sure these errors and warnings will be output in a suitable format when using the `--format=ndjson` option.

Todo:
* [x] errors/warnings for `ndjson`
* [x] check if the monorepo split needs to be adjusted
* [x] maybe add some tests for the configuration errors

PR #4938, opened 2022-07-02
Always report the installation as being complete

ausi:
The existence of the *localconfig.php* is no longer a requirement for an installation to be “complete”.


fritzmg (dismissed):

ausi:
Needs to be merged after #4935 
(the failing tests should then resolve themself)

m-vo (approved):

leofeyer:
The tests are failing because we are still using the method here:

https://github.com/contao/contao/blob/b1233f34a9990937ace27c5d5329473b4cb3824e/core-bundle/src/Framework/ContaoFramework.php#L223-L225

I think we can remove these lines as well, can‘t we?

fritzmg:
@leofeyer see https://github.com/contao/contao/pull/4938#issuecomment-1172925642

#4935 needs to be merged first. Then that would be gone anyway (see https://github.com/contao/contao/pull/4935/files#diff-11c0c0e6a08183ce10e460bd6381f63a6b7a773abf1692e4e1b85f85dec0558b).

leofeyer:
Thank you @ausi.

PR #4939, opened 2022-07-02
Fix sectionwizard.js

### LINKED ISSUE #4844: sectionwizard.js is broken
### Affected version(s)

5.x

### Description

In the current 5.x, the copy/delete/enable buttons in the `sectionwizard` are not working for me.

I get this error in the JS console - for instance when editing a layout:
image

Here is the stacktrace in the unminified source:
```
sectionwizard.min.js?v=f3c63d9b:17 Uncaught TypeError: Cannot read properties of null (reading 'children')
    at makeSortable (sectionwizard.min.js?v=f3c63d9b:17:34)
    at sectionwizard.min.js?v=f3c63d9b:102:13
    at NodeList.forEach (<anonymous>)
    at init (sectionwizard.min.js?v=f3c63d9b:7:52)
    at sectionwizard.min.js?v=f3c63d9b:115:25
    at NodeList.forEach (<anonymous>)
    at MutationObserver.observe.attributes (sectionwizard.min.js?v=f3c63d9b:113:37)
```

The same is true for the `modulewizard`, though I'm not sure if this is just a follow-up error.

fritzmg:
Tries to fix #4844

This also restructures the implementation a bit, so that it can in theory be used with `async`.

This in theory does not fix it completely though as it is possible that the `tbody` element is there - but its children (`tr` and `button`) are still missing. However, I could not figure out a way to check whether a specific child node has all its children fully loaded and parsed when using a MutationObserver. @ausi @m-vo any ideas?


ausi:
> However, I could not figure out a way to check whether a specific child node has all its children fully loaded and parsed

I think this is not really possible because of the way how HTML works. Instead we should write the code in a way that would support adding more child elements after initialization, so that it doesn’t matter if javascript adds these nodes or the HTML parser.

The mutation observer part could look like this then:

```js
if (element.matches && element.matches('.tl_sectionwizard, .tl_sectionwizard *')) {
	init(element.closest('.tl_sectionwizard'));
}
```

And the `init()` method should re-initialize the table instead of skipping it. (Not sure if this can be done without too much effort in this case).

fritzmg:
> The mutation observer part could look like this then:

Doesn't the event fire for the parent anyway, if its children are changed?

ausi:
> > The mutation observer part could look like this then:
> 
> Doesn't the event fire for the parent anyway, if its children are changed?

No, `mutation.addedNodes` contains the added nodes, this could be a `<td>` or `<tr>` or anything else that gets added to the DOM via the HTML parser or the DOM APIs.

aschempp:
I've had similar issues recently in my extension and the solution was actually to wait for the first `DOMContentLoaded` event, see https://github.com/terminal42/contao-conditionalformfields/commit/714e7e86a34a3f40659d6b0597dce40ce6bf7ea7. Maybe this would be a correct/better fix here?

fritzmg:
I have redone the implementation. Use the split view and "hide whitespace" to compare the changes. Basically I am now initializing per table row and count the number of necessary elements (all the inputs, selects and buttons). Only if all elements are present the row is initialized. The rows are also immediately initialized, meaning any rows that were rendered before this JavaScript is executed get captured - and all rows rendered after that get captured by the MutationObserver.

One caveat of the new implementation is that `makeSortable(tbody)` gets called multiple times initially (once for each pre-existing row) - which is fine functionality wise, as `makeSortable(tbody)` will always be called again when rows are duplicated or deleted for example.

core-bundle/public/sectionwizard.js
```diff
@@ -1,118 +1,121 @@
...
-            }
+    const init = (row) => {
+        // Check if this row has already been initialized
+        if (initializedRows.has(row)) {
+            return;
+        }
 
-            initialized.set(table, true);
+        // Check if this row has all necessary elements
+        if (7 !== row.querySelectorAll('input, select, button').length) {
```
m-vo, line 13:
```suggestion
        // Skip header row
        if (row.querySelector('th')) {
```

In my tests, the `row` was either the header (clarified: the `tr` containing the following)…

```
<th>Title</th>
<th>ID</th>
<th>Template</th>
<th>Position</th>
<th></th>
```

…or a valid row with buttons. Could this have been the problem?

fritzmg: The `row` variable should always be a `tr` element, since that is what I pass to `init` [here](https://github.com/contao/contao/pull/4939/files#diff-4511fe5a886b411f394444cab142d2d24b5b1726b4a8a1538c7f2b864a3ae4f7R111) and [here](https://github.com/contao/contao/pull/4939/files#diff-4511fe5a886b411f394444cab142d2d24b5b1726b4a8a1538c7f2b864a3ae4f7R118) 🤔

fritzmg: Also the check is there in order to disregard any MutationObserver events where not all elements (per `<tr>`) have been parsed yet that are required for the functionality of this JavaScript.

m-vo (approved):

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4940, opened 2022-07-03
Fix default callback order

### LINKED ISSUE #1795: Default priority for callbacks uncertain
**Affected version(s)**

Contao 4.9?

**Description**

I registered a callback with annotations (`@Callback(table="tl_content", target="config.onload")`) but without priority.

Deviating from the documentation the callback is not executed after all other callbacks - it's the first one. Although my Bundle is registered after ContaoCoreBundle!

See the docs here: https://github.com/contao/docs/blame/master/docs/dev/framework/dca/_index.md#L96

@aschempp noticed that it was like the docs in the first draft - see:
https://github.com/contao/contao/pull/39/commits/6577835f22a5811b8387aa70804fd7668c0870b3#diff-26ec4b3e6c5b2ff5898a7c6caa86c1c6R128

 But now it isn't anymore the last! Is it intended, that the callback is registered as first instead of last with `priority=0` (default) or is it a failure in the docs? Only with `priority=-1` the Callback gets correctly loaded after all other callbacks!

fritzmg:
Fixes #1795

With this change tagged callbacks with a priority of `0` will be inserted _after_ pre-existing callbacks of the same priority.


m-vo (dismissed):

ausi (dismissed):

leofeyer (changes requested):
Changes look good, however, you forgot to document the BC breaks in the `UPGRADE.md` file.

aschempp (dismissed):

leofeyer (dismissed):

leofeyer:
Thank you @fritzmg.

leofeyer (approved):

PR #4945, opened 2022-07-04
Hard-code timezone to avoid time deviation

bezin:
When I run the test suite locally, the InsertTagsControllerTest always fails due to the expected and actual `expires` time being off by one hour. My local timezone is 'Europe/Berlin`.

### DIFF
--- a/core-bundle/tests/Controller/InsertTagsControllerTest.php
+++ b/core-bundle/tests/Controller/InsertTagsControllerTest.php
@@ -50,6 +50,7 @@ class InsertTagsControllerTest extends TestCase
 
     public function testSpecialDateInsertTagHandling(): void
     {
+        ini_set('date.timezone', 'UTC');
         $year = date('Y');
 
         $insertTagParser = $this->createMock(InsertTagParser::class);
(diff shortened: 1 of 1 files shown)


fritzmg:
Does this problem also occur if you only run this _one_ test?

bezin:
> Does this problem also occur if you only run this one test?

Yes, it does, but I also thought about setting the time zone in the `setUp` method to set it for every test?

fritzmg:
Well in theory it should not matter. The test and the controller use the same timezone, so I wonder why it's different.

fritzmg:
I see the issue now. `Response::setExpires` (obviously) saves the time in UTC. `Response::getExpires` then returns a `DateTimeInterface` instance where the timezone is also UTC.

So rather than forcing a specific timezone, the test needs to be changed accordingly in order to be agnostic to that. e.g.:

```diff
-   $this->assertSame($year.'-12-31 23:59:59', $response->getExpires()->format('Y-m-d H:i:s'));
+   $this->assertSame((new \DateTimeImmutable($year.'-12-31 23:59:59'))->getTimestamp(), $response->getExpires()->getTimestamp());
```

bezin:
I updated the PR accordingly, thank you @fritzmg

fritzmg:
@bezin this is ready for review now, no?

ausi (approved):

leofeyer:
Thank you @bezin.

PR #4948, opened 2022-07-05
Fix potential PHP 8 warnings when resizing an uploaded image

qzminski:
In case the image size of an uploaded file cannot be determined, we may get a few PHP 8 warnings (see Sentry reports below). This pull request fixes the problem, although I am not entirely sure if it's the best solution.

CleanShot 2022-07-05 at 09 33 53
CleanShot 2022-07-05 at 09 34 05
CleanShot 2022-07-05 at 09 34 12


leofeyer:
I think the solution is good, but let‘s wait for @ausi‘s review.

ausi (approved):

leofeyer:
Thank you @qzminski.

PR #4952, opened 2022-07-05
Allow iterating FilesystemItemIterator multiple times

### LINKED ISSUE #4913: Allow iterating multiple times in FilesystemItemIterator
### Affected version(s)

4.13

### Description

The current `FilesystemItemIterator` implementation has a subtle problem with already closed iterators.

I'd basically like to make the following access pattern possible:

```php
$iterator = $filesystemItemIterator;

// Call any function that iterates without returning the list
$iterator->any(…);

foreach($iterator as $item) {
  // this should work, but the iterator is already closed
}
```

This is especially important to support functions that allow peeking like `getFirst()`.

m-vo:
Fixes #4913 

/cc @ausi


ausi (dismissed):

Toflar (commented):
Are we sure we want to always cache this? It's going to potentially use lots of RAM. I mean, I'm fine with the default but maybe it should be deactivatable? Just thinking about some CLI command doing lots of work...

m-vo:
> Are we sure we want to always cache this? It's going to potentially use lots of RAM. 

IMHO the objects are quite small, so the used memory might only be noticeable when iterating through thousands of files. I'm fine with adding an option to disable caching but I would postpone this until needed. Wdyt?

Could probably be implemented like this then:
```php
class FilesystemItemIterator  implements \IteratorAggregate
{
    public function withoutCache(): self 
    {
        return $this->listing instanceof CachingTraversable ? 
            new self($this->listing->getInnerIterator()) : $this;
    }

    // …
}

class CachingTraversable implements \IteratorAggregate, \OuterIterator
{
    public function getInnerIterator(): ?Iterator { /*…*/ }

    // …
}
```

The primary goal of this PR is to remove the burdens from the current API. E.g. in the download element I need to peek into the first element to check if it's a video or audio type and then decide what to do. With the behavior before this PR, you would not be able to iterate the list afterwards.

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

m-vo:
… and big thanks to @ausi as well. :relaxed:

PR #4957, opened 2022-07-06
Fixed str_replace errors when passing null

aschempp:


bytehead (approved):

core-bundle/src/Resources/contao/modules/Module.php
```diff
@@ -458,7 +458,7 @@ protected function compileNavigationRow(PageModel $objPage, PageModel $objSubpag
 		$row['rel'] = '';
 		$row['nofollow'] = false; // backwards compatibility
 		$row['target'] = '';
-		$row['description'] = str_replace(array("\n", "\r"), array(' ', ''), $objSubpage->description);
+		$row['description'] = str_replace(array("\n", "\r"), array(' ', ''), (string) $objSubpage->description);
```
leofeyer, line 461: This will trigger an "unnecessary type cast" warning, because the property is declared as string already. You can either use `$objSubpage->description ?? ''` instead or even better:

```php
$row['description'] = '';

if ($objSubpage->description)
{
    $row['description'] = str_replace(array("\n", "\r"), array(' ', ''), $objSubpage->description);
}
```

aschempp: well since the field is nullable, shouldn't the property be declared as `string|null` then? To correctly raise an error in this method that the value could be null?

leofeyer: Yes, in this case, the `@property` annotation needs to be fixed.

aschempp: it already appears to be 🤷 
https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/models/PageModel.php#L39

PR #4961, opened 2022-07-07
Always set both collate and collation to the same value

fritzmg:
Follow up to #4799. This was meant to go into 4.13 but it seems I accidentally based it against 5.x only.


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4964, opened 2022-07-07
Fix PHP 8 warning in ModuleRegistration

ameotoko:
Registration module triggers a warning under PHP 8 while sending admin notification.

```
Warning: Trying to access array offset on value of type null
```

This happens because `id` column does not have translated label, but this field is, in fact, not really needed in this list altogether, because it is already present in the first line of the default message:

```
A new member (ID 10) has registered at your website.

Username: test9
Allow login: 1
Member groups: 
Deactivate: 1

If you did not allow e-mail activation, you have to enable the account manually in the back end.
```

So I fixed it by adding `id` to the list of skipped fields. For any other similar cases (e.g. if a developer adds another field to the DCA but forgets to add translation) the proposed fix is to use the column name instead.

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModuleRegistration.php
+++ b/core-bundle/src/Resources/contao/modules/ModuleRegistration.php
@@ -647,7 +647,7 @@ class ModuleRegistration extends Module
 		// Add user details
 		foreach ($arrData as $k=>$v)
 		{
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

aschempp (approved):

aschempp:
the failing PHP 8.1 tests seem to be unrelated to me

leofeyer:
Thank you @ameotoko.

PR #4965, opened 2022-07-07
Use cache_suffix for TinyMCE

### LINKED ISSUE #2258: Use version strategy for contao-components assets
When updating to Contao 4.10 a lot of users and editors stumble over the problem that they still have old assets cached in their browser and thus they need to clear their browser cache in order to get the TinyMCE working again for example.

Since the `be_tinyMCE` template for example uses `$this->asset` anyway, wouldn't it make sense to define a version strategy for these assets, so that the cache can be automatically busted if necessary?

fritzmg:
When updating a Contao instance to a new minor version sometimes the TinyMCE version changes along with it. But this sometimes leads to a situation where you will see errors like this in the back end:

image (5)

So far the only way to solve this is to tell all back end users to clear their browser cache, which is not ideal.

However, this can actually be easily fixed by using the [`cache_suffix`](https://www.tiny.cloud/docs/configure/integration-and-setup/#cache_suffix) setting (also available for [TinyMCE 4](https://www.tiny.cloud/docs-4x/configure/integration-and-setup/#cache_suffix)). This PR sets the `cache_suffix` to `?v=5.10.1` for example (in 4.13).

_Note:_ not sure if this really needs to be its own template block - but may be you want to be able to override the `cache_suffix` 🤷 

I'll create a separate PR for Contao 4.9.


bytehead (dismissed):

rabauss (dismissed):

ausi (dismissed):

core-bundle/src/Resources/contao/templates/backend/be_tinyMCE.html5
```diff
@@ -64,6 +64,10 @@ window.tinymce && tinymce.init({
     contextmenu: false,
   <?php $this->endblock(); ?>
 
+  <?php $this->block('cache_suffix'); ?>
+    cache_suffix: '?v=<?= System::getContainer()->get('assets.packages')->getVersion('js/tinymce.min.js', 'contao-components/tinymce4') ?>',
```
leofeyer: I don‘t like having to use `System::getContainer()` in templates. @m-vo How would we handle this in Twig?

fritzmg: @leofeyer Symfony provides an [`asset_version`](https://github.com/symfony/symfony/blob/212de381d981b7cd7152f1ba69e12185f13da777/src/Symfony/Bridge/Twig/Extension/AssetExtension.php#L43) Twig method in its Twig Extension. We could also add a `Template::assetVersion` helper method, alongside our existing `Template::asset` method.

leofeyer: Yes, please add an `assetVersion()` method. 👍

fritzmg:
`Template::assetVersion` method added in https://github.com/contao/contao/pull/4966. This PR now depends on an upstream merge from 4.9 once https://github.com/contao/contao/pull/4966 is merged.

bytehead (approved):

leofeyer:
Thank you @fritzmg.

PR #4966, opened 2022-07-07
Use cache_suffix for TinyMCE

### LINKED ISSUE #2258: Use version strategy for contao-components assets

fritzmg:
This implements #4965 for Contao 4.9.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Template.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Template.php
@@ -413,6 +413,19 @@ abstract class Template extends Controller
 	}
 
 	/**
+	 * Returns an asset version
+	 *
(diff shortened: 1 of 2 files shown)


bytehead (dismissed):

ausi (dismissed):

bytehead (approved):

leofeyer:
Thank you @fritzmg.

PR #4968, opened 2022-07-08
Set login constants in request listener

### LINKED ISSUE #2576: Autologin / Rememberme for FrontendUser only works on second page load
**Affected version(s)**

Contao 4.9

**Description**
When using the autologin function (Cookie REMEMBERME) in the frontend login module, the actual autologin never works on first, but only on second page load.

**How to reproduce**

1. Create a simple page
2. Create MemberGroup Testgroup
3. Create Member Testmember, allow login and assign Testgroup
4. Insert frontend user login module in pages article
5. Insert text element "Loggedin" and protect it for MemberGroup Testgroup
6. Browse to page with a clean browser session, login as Testmember and tick "remember me / autologin"
7. Text "Loggedin" will show.
8. Let session expire or manually delete PHPSESSION Cookie using your browsers development tools
9. Reload page.

**Expected Result**
Text "Loggedin" will show, login form will be hidden and button Logout shows.

**Actual Result**
Text Loggedin does **not** not show, but login form is hidden and button Logout shows. 
When refreshing the page now, the Text Loggedin will show.

In addition to this, if you also you make the page itself procteded for MemberGroup Testgroup this exception will be logged / shown in dev mode:

`ErrorException:
Warning: implode(): Invalid arguments passed

  at vendor/contao/core-bundle/src/Resources/contao/controllers/FrontendIndex.php:281
     (web/index.php:31)
  at require('.../web/index.php')
     (web/app.php:4)      `

fritzmg:
Fixes #2576 and other issues.

In #2628 we removed the usage of the deprecated `FE_USER_LOGGED_IN` constant within the core and switched to using the token checker instead, in order to fix #2576.

However, the core issue still remains: for any Contao extension that still uses the `FE_USER_LOGGED_IN` constant, the constant will still have the wrong value for any authenticated request without a previous authenticated session. For example: if you implement a custom guard authenticator that authenticates a Contao user e.g. via an `Authorization` header, or some other custom solution, the constant will always be `false` (if no session cookie of a previous authenticated request is passed along with the request).

The cause of this is the setting of the constants within `ContaoFramework::initialize`. `ContaoFramework::initialize` will be executed _very early_ - earlier than the Firewall listener, which is a regular request listener. So since `$this->tokenChecker->hasFrontendUser()` will be executed before Symfony's Firewall listener, it will always be false, because no security token has been set by the Symfony Firewall yet.

To fix this we must not query the `TokenChecker` before the Symfony Firewall listener. i.e. we need to set this constant in a regular request listener that has a lower priority than the Firewall request listener of Symfony. This is implemented in this PR.

### DIFF
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -28,7 +28,7 @@ class LegacyLoginConstantsListener
      */
     private $tokenChecker;
 
-    /** 
+    /**
      * @var ScopeMatcher
      */
     private $scopeMatcher;
@@ -46,7 +46,7 @@ class LegacyLoginConstantsListener
             \define('FE_USER_LOGGED_IN', $this->tokenChecker->hasFrontendUser());
         } else {
             \define('BE_USER_LOGGED_IN', false);
-            \define('FE_USER_LOGGED_IN', false);   
+            \define('FE_USER_LOGGED_IN', false);
         }
     }
 }
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -13,8 +13,6 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\EventListener;
 
 use Contao\CoreBundle\Framework\ContaoFramework;
-use Contao\CoreBundle\Routing\ScopeMatcher;
-use Contao\CoreBundle\Security\Authentication\Token\TokenChecker;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 
 /**
@@ -40,6 +38,7 @@ class LegacyLoginConstantsListener
         // Otherwise allow the framework to set them itself during initialize.
         if (!$this->framework->isInitialized()) {
             $this->framework->setLoginConstantsOnInit(true);
+
             return;
         }
 
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -12,6 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\EventListener;
 
+use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\CoreBundle\Routing\ScopeMatcher;
 use Contao\CoreBundle\Security\Authentication\Token\TokenChecker;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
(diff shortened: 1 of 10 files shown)


aschempp:
I think there are two things to consider/fix here:
1. The listener should not set the constants if the framework was not initialized
2. if the framework is initialized _after_ the firewall (e.g. if a Symfony route matches), we must still set the constants when booting the framework.

Otherwise 👍  for using a listener after firewall!

core-bundle/src/EventListener/LegacyLoginConstantsListener.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(TokenChecker $tokenChecker, ScopeMatcher $scopeMatcher)
+    {
+        $this->tokenChecker = $tokenChecker;
+        $this->scopeMatcher = $scopeMatcher;
+    }
+
+    public function __invoke(RequestEvent $event): void
+    {
+        if ($this->scopeMatcher->isFrontendRequest($event->getRequest())) {
+            \define('BE_USER_LOGGED_IN', $this->tokenChecker->hasBackendUser() && $this->tokenChecker->isPreviewMode());
```
aschempp: be aware that in upstream-merge we must remove the `hasBackendUser` because in 4.13 a user can be in preview mode without being backend-authenticated (through preview links)

fritzmg: This has now been implemented. Now the framework service always sets the constants itself again. The listener checks whether the framework was initialized and if yes, it orders the framework service to set the constants. If not, it allows the framework service to set the constants during initialize later on (if anything initializes the framework).

aschempp (commented):
I would change two things here
1. always use `getCurrentRequest` for the constants to keep it consistent
2. I don't think we need two methods. Calling `setLoginConstants()` when the framework is not initialised should set the internal flag. Calling the method (again) if the framework is initialised (internally) should set the constants automatically.

fritzmg:
> always use getCurrentRequest for the constants to keep it consistent

Are you sure? Should we not pass the request of the request event?

aschempp (commented):
Not sure about that request, but it probably does not matter. But why did you remove the mode-check and the session initialization?

fritzmg:
> But why did you remove the mode-check

~It does not really make sense to check whether this is a front end request or not via `$this->getMode()`?~ Ah I see, because of the `$isFrontend` variable.

> and the session initialization?

I don't understand the purpose of `$session->start()` here and I do not think it is needed.

Also the previous condition had `$this->request->hasPreviousSession()` - which also does not make sense. This was likely a workaround, as the Contao framework is initialized early before the firewall listener and thus this bit of code never worked for the initial authentication request.

aschempp:
> Also the previous condition had $this->request->hasPreviousSession() - which also does not make sense. This was likely a workaround, as the Contao framework is initialized early before the firewall listener and thus this bit of code never worked for the initial authentication request.

It made sense because without a previous session you could never be logged in – since a security token would have been stored in the DB. But I agree if the constant is only set **after** the firewall this should never be necessary.

aschempp (approved):

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #4971, opened 2022-07-11
Fix the getCurrentRecord() method

leofeyer:
Not sure if it should be `if (!isset(self::$arrCurrentRecordCache[$key]))` instead as in line 1758?


aschempp:
both works imho, I would go with `isset()` for consistency

leofeyer:
Changed in 5462d19a00adb084624b45a7ce7dc2dd83b49d55.

fritzmg (approved):

ausi (approved):

PR #4976, opened 2022-07-13
Improve currentRecord checks and permissions

aschempp:
Fixes multiple issues:
1. always check if `getCurrentRecord` returned a valid value
2. Skip records if they cannot be loaded in loops (deleteChilds, editAll, reviseTable etc)
3. Check `UpdateAction` permissions **before** rendering the `edit` and `editAll` masks


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1749,7 +1749,7 @@ protected static function setCurrentRecordCache(string|int $id, string $table, a
 	 * @throws AccessDeniedException     if the current user has no read permission
 	 * @return array<string, mixed>|null
 	 */
-	public function getCurrentRecord(string|int $id = null, string $table = null): ?array
+	public function getCurrentRecord(string|int $id = null, string $table = null, bool $throw = true): ?array
```
leofeyer: I don‘t like the API. IMHO the method should always throw the exception and we should use a try-catch block if we want to handle the case. So instead of

```php
$currentRecord = $this->getCurrentRecord($row['id'], $v, false);

if ($currentRecord === null)
{
	continue;
}
```

it should be

```php
try
{
	$currentRecord = $this->getCurrentRecord($row['id'], $v);
}
catch (AccessDeniedException $e)
{
	continue;
}
```

Toflar (changes requested):
Agree with @leofeyer. That's a bad API.

aschempp:
Just something I've seen in a lot of places 🤷‍♂️

https://github.com/symfony/symfony/blob/123b1651c4a7e219ba59074441badfac65525efe/src/Symfony/Contracts/HttpClient/ResponseInterface.php

https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/DependencyInjection/ContainerInterface.php#L41

https://github.com/symfony/symfony/blob/870eeb975feb1abb4b8a1722e1fd57beeab2b230/src/Symfony/Component/PropertyAccess/PropertyAccessor.php

I first added a flag to not check permissions, but thats also difficult if the same record has thrown before.

Toflar:
Yep. Bad APIs everywhere.

aschempp:
Really? So you want https://github.com/contao/contao/pull/4976/files#diff-e4bf1535e70e3fbc2b9692ba8add4c45237305ff815df14c7ba9c9ce53dc9204L1669 to looks like this??

```php
try 
{
    $currentRecord = $this->getCurrentRecord($row['id'], $v, false);
}
catch (AccessDeniedException)
{
    continue;
}

if ($currentRecord === null)
{
    continue;
}

try
{
    $this->denyAccessUnlessGranted(ContaoCorePermissions::DC_PREFIX.$v, new DeleteAction($v, $currentRecord));
}
catch (AccessDeniedException)
{
    continue;
}
```

Toflar:
Yes, can be wrapped into the same block.

aschempp:
updated in 09123de21fac6c907ca687a42e404c1e4c24f220

Toflar (approved):

PR #4979, opened 2022-07-13
Fix legacy routing matcher not matching the route if page has no alias

qzminski:
If we have a legacy routing mode enabled and the page has no alias, its route cannot be matched by the page ID (e.g. `/79.html`). This pull request fixes this problem.

CleanShot 2022-07-13 at 11 02 22

/cc @aschempp


aschempp (approved):
the problem only arises if there's a `getPageIdFromUrl` hook, but that does not matter its still a bug 🙃

leofeyer:
Thank you @qzminski.

PR #4981, opened 2022-07-13
Merge branch '4.13' into 5.x

leofeyer:


core-bundle/src/Filesystem/FilesystemItemIterator.php
```diff
@@ -22,6 +24,7 @@ class FilesystemItemIterator implements \IteratorAggregate
      */
     public function __construct(private iterable $listing)
     {
+        $this->listing = \is_array($listing) ? $listing : new CachingTraversable($listing);
```
leofeyer: @m-vo Is this correct in PHP 8.1 or should it be the following?

```php
public function __construct(private iterable $listing)
{
    if (!\is_array($listing)) {
        $this->listing = new CachingTraversable($listing);
    }
}
```

m-vo: Both are fine IMHO. :+1: 

To me, the suggested version with the `!\is_array` makes it more clear, in that the `CachingTraversable` is only used on non-arrays.

PR #4984, opened 2022-07-13
Always retrieve the session from the request instead of the container

leofeyer:
```
  1x: Since symfony/framework-bundle 5.3: The "session" service and "SessionInterface" alias are deprecated, use "$requestStack->getSession()" instead. It is being referenced by the "contao.security.frontend_preview_authenticator" service.
    1x in DbafsTest::testAlterFilesystemAndSync from Contao\CoreBundle\Tests\Functional

  1x: Since symfony/framework-bundle 5.3: The "session" service and "SessionInterface" alias are deprecated, use "$requestStack->getSession()" instead. It is being referenced by the "contao.security.token_checker" service.
    1x in DbafsTest::testAlterFilesystemAndSync from Contao\CoreBundle\Tests\Functional
```

### DIFF
--- a/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
+++ b/core-bundle/src/DependencyInjection/ContaoCoreExtension.php
@@ -382,10 +382,11 @@ class ContaoCoreExtension extends Extension implements PrependExtensionInterface
         }
 
         $tokenChecker = $container->getDefinition('contao.security.token_checker');
-        $tokenChecker->replaceArgument(5, new Reference('security.access.simple_role_voter'));
 
         if ($container->hasParameter('security.role_hierarchy.roles') && \count($container->getParameter('security.role_hierarchy.roles')) > 0) {
-            $tokenChecker->replaceArgument(5, new Reference('security.access.role_hierarchy_voter'));
+            $tokenChecker->replaceArgument(4, new Reference('security.access.role_hierarchy_voter'));
+        } else {
+            $tokenChecker->replaceArgument(4, new Reference('security.access.simple_role_voter'));
         }
     }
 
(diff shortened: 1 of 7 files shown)


core-bundle/src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -382,10 +382,11 @@ private function handleTokenCheckerConfig(ContainerBuilder $container): void
         }
 
         $tokenChecker = $container->getDefinition('contao.security.token_checker');
-        $tokenChecker->replaceArgument(5, new Reference('security.access.simple_role_voter'));
```
bytehead, line 385: Should that be part of that  PR?

bytehead (dismissed):
LGTM

core-bundle/src/DependencyInjection/ContaoCoreExtension.php
leofeyer: I have moved it to the `else` part trying to fix https://github.com/contao/contao/runs/7323906436?check_suite_focus=true. However, it did not work.

There is no point in replacing the argument twice, though.

bytehead (approved):

aschempp:
I wonder why this always uses `$request->getSession` although the error message clearly says **use "$requestStack->getSession()" instead**?

fritzmg:
I would assume that the message is deliberately phrased in a more general way as the request won't be directly available in every service. The UPGRADE.md does mention to use `$request->getSession()` though.

leofeyer:
> I wonder why this always uses `$request->getSession` although the error message clearly says **use "$requestStack->getSession()" instead**?

This is a very good question. `RequestStack::getSession()` will use the **last** request in the stack whereas `RequestStack::getMainRequest()` will use the **first** request in the stack. So far, we have always been using the main request, therefore I did not use `RequestStack::getSession()`. Maybe it would be better, though?

Issue #4987, opened 2022-07-14
Preview of 401, 403, 404 pages executed in wrong TL_MODE


fritzmg:
### Affected version(s)

4.13

### Description

When showing the preview of the `40x` pages in Contao 4.13, an error like this might occur:

```
Exception:
Could not find template "js_foobar"

  at vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\TemplateLoader.php:156
  at Contao\TemplateLoader::getDefaultPath('js_foobar', 'html5')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\TemplateLoader.php:112)
  at Contao\TemplateLoader::getPath('js_foobar', 'html5')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Controller.php:97)
  at Contao\Controller::getTemplate('js_foobar')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\TemplateInheritance.php:351)
  at Contao\Template->getTemplatePath('js_foobar', 'html5')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\TemplateInheritance.php:95)
  at Contao\Template->inherit()
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Template.php:295)
  at Contao\Template->parse()
     (vendor\contao\contao\core-bundle\src\Resources\contao\classes\FrontendTemplate.php:43)
  at Contao\FrontendTemplate->parse()
     (vendor\contao\contao\core-bundle\src\Resources\contao\pages\PageRegular.php:692)
  at Contao\PageRegular->createFooterScripts(object(LayoutModel), object(PageModel))
     (vendor\contao\contao\core-bundle\src\Resources\contao\pages\PageRegular.php:256)
  at Contao\PageRegular->prepare(object(PageModel))
     (vendor\contao\contao\core-bundle\src\Resources\contao\pages\PageRegular.php:60)
  at Contao\PageRegular->getResponse(object(PageModel))
     (vendor\contao\contao\core-bundle\src\Resources\contao\pages\PageError401.php:89)
  at Contao\PageError401->getResponse(object(PageModel), true)
     (vendor\contao\contao\core-bundle\src\Resources\contao\controllers\FrontendIndex.php:320)
  at Contao\FrontendIndex->renderPage(object(PageModel))
     (vendor\contao\contao\core-bundle\src\Controller\Page\ErrorPageController.php:46)
[...]

PR #4988, opened 2022-07-14
Redirect to fragment URL on preview URL error

aschempp:
Fixes https://github.com/contao/contao/issues/4987

as @fritzmg pointed out, forwarding the request will lead to issues with any constant defined in the application, because we cannot break scope. This PR changes the preview URL to forward to the fragment controller. Since the URL is signed, it can be safely shared in the front end and not abused by anyone.

PS: there is still a chance that generating this endpoint will fail, and the resulting error URL will look _strange_ because it includes the fragment URL. But the same already happes for an internal forward as far as I know (since we're essentially faking the URL in both cases).


fritzmg:
> PS: there is still a chance that generating this endpoint will fail, and the resulting error URL will look _strange_ because it includes the fragment URL. But the same already happes for an internal forward as far as I know (since we're essentially faking the URL in both cases).

Hm, I see, yes. The URL for the preview of a `40x` page will now look like this for example:

```
https://c413.local/preview.php/_fragment?_hash=59FZd5WA%2BbGEsodQKr1g1PweegK%2FvRKYg7pKWmiLUSY%3D&_path=_token_check%3D1%26_controller%3DContao%255CCoreBundle%255CController%255CPage%255CErrorPageController%26_scope%3Dfrontend%26_locale%3Den%26_format%3Dhtml%26_canonical_route%3Dtl_page.23%26pageModel%3D23
```

fritzmg (dismissed):

fritzmg (dismissed):

leofeyer (approved):

PR #4989, opened 2022-07-14
Make the Input::post() method compatible with Symfony 6

leofeyer:
Symfony 6 wants `$request->request->get($key)` to always return a string and `$request->request->all($key)` to always return an array. Since we don‘t know what type to expect, I have added a workaround. Although I‘m not sure if that is really what Symfony wants people to do.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Input.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Input.php
@@ -1119,7 +1119,9 @@ class Input
 				return $arrPost[$strKey];
 			}
 
-			return $request->request->get($strKey);
+			// Since we don't know whether the value will be an array or a string,
+			// we can neither use ->get($strKey) nor ->all($strKey) in Symfony 6.
+			return $request->request->all()[$strKey] ?? null;
 		}
 
 		trigger_deprecation('contao/core-bundle', '5.0', 'Getting data from $_POST with the "%s" class has been deprecated and will no longer work in Contao 6.0. Make sure the request_stack has a request instead.', __CLASS__);


fritzmg:
> Symfony 6 wants `$request->request->get($key)` to always return a string

Why do you think so? The return type is defined as `string|int|float|bool|null`:  https://github.com/symfony/symfony/blob/v6.1.2/src/Symfony/Component/HttpFoundation/InputBag.php#L28-L41

bytehead (approved):
Fine for me.

bytehead:
@fritzmg https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpFoundation/InputBag.php#L39

fritzmg:
That deprecation seems to be a mistake, as Symfony 6 only requires a scalar value to be returned and the return type can be `string|int|float|bool|null`. See [here](https://github.com/symfony/symfony/blob/v6.1.2/src/Symfony/Component/HttpFoundation/InputBag.php#L28-L41).

leofeyer:
<del>Interesting. So this will actually not be a problem in Symfony 6? Then I would close this PR.</del>

**Edit**: It will be a problem in Symfony 6.

bytehead:
I don't get the change here: https://github.com/symfony/symfony/commit/8312f90ff78195d66e1f9667841f14bb5df4dedf#diff-94e19ba3b933d919da3a23761ecef19e001863c18676880231d420d60fa6a142

fritzmg:
> Interesting. So this will actually not be a problem in Symfony 6? Then I would close this PR.

Yes, it seems to be a bug in 5.4. https://github.com/symfony/symfony/pull/46936

leofeyer:
That only fixes the deprecation message.

This PR still needs to be merged to be compatible with Symfony 6, because an array is not a scalar value. `$request->request->get($key)` does not allow arrays and `$request->request->all($key)` does not allow non-array values.

fritzmg:
> That only fixes the deprecation message.

No, it also changes the check itself, so that `bool`, `null` and `int` do not trigger the deprecation.

> This PR still needs to be merged to be compatible with Symfony 6, because an array is not a scalar value. 

True. Though I do wonder: why did Symfony deprecate arrays for InputBag? I mean what happens now if you have an URL like

```
https://example.com/foobar?lorem[]=1
```

and then execute `$request->query->get('lorem')`? Seems like a pretty normal use case to me.

leofeyer:
They want you to execute `$request->request->all($key)` instead, which of course only works if you know what type the value will be. 🤷‍♂️

PR #4992, opened 2022-07-14
Remove the old app.php entry point

leofeyer:
I think we don‘t need this anymore, do we?


fritzmg:
Yes, anyone needing it can still easily create it for themselves. But it should be mentioned in the `UPGRADE.md`. Also: should we add a deprecation to that entrypoint in 4.9?

aschempp (dismissed):

contaoacademy:
I think it should be mentioned that the old app.php lines have to be changed to index.php in .htaccess

PR #4993, opened 2022-07-15
Use service annotations to tag DCA callbacks

leofeyer:


aschempp:
shouldn't we use PHP8 attributes?

leofeyer:
Yes, that will be the next PR. 😄

PR #5001, opened 2022-07-18
Use Controller::getContentElement() in the alias element

leofeyer:
This PR makes the `ContentAlias` and `ContentModule` classes use `Controller::getContentElement()` and `Controller::getFrontendModule()` as discussed in #2678.

### DIFF
--- a/core-bundle/contao/elements/ContentAlias.php
+++ b/core-bundle/contao/elements/ContentAlias.php
@@ -27,7 +27,10 @@ class ContentAlias extends ContentElement
 			return '';
 		}
 
-		$objElement = ContentModel::findByPk($this->cteAlias);
+		if (!$objElement = ContentModel::findByPk($this->cteAlias))
+		{
+			return '';
+		}
 
 		// Clone the model, so we do not modify the shared model in the registry
 		$objModel = $objElement->cloneOriginal();
--- a/core-bundle/contao/elements/ContentAlias.php
+++ b/core-bundle/contao/elements/ContentAlias.php
@@ -22,45 +22,19 @@ class ContentAlias extends ContentElement
 	 */
 	public function generate()
 	{
+		if ($this->isHidden())
+		{
+			return '';
+		}
+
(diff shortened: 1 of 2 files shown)


core-bundle/contao/elements/ContentAlias.php
```diff
@@ -22,46 +22,24 @@ class ContentAlias extends ContentElement
...
-		$strClass = static::findClass($objElement->type);
+		$objElement = ContentModel::findByPk($this->cteAlias);
 
-		if (!class_exists($strClass))
+		if ($objElement === null)
 		{
 			return '';
 		}
 
-		if (is_a($strClass, ContentProxy::class, true))
```
leofeyer, line 39: @aschempp Do we still need this now that we are using `Controller::getContentElement()`?

aschempp: no I don't think we still need that. But did you manually test keeping CSS class and ID still works, even when using fragments?

leofeyer: No. How do I test this? Why don‘t we have a unit tests for this?

aschempp: > No. How do I test this? 

create an alias content element, select another element, add CSS classes to both and see what you get in the front end is what I would do.

> Why don‘t we have a unit tests for this?

I guess because we don't have tests for any old modules or elements?

leofeyer: > create an alias content element, select another element, add CSS classes to both and see what you get in the front end is what I would do.

This is what I did. But it does not seem to fulfill the `is_a($strClass, ContentProxy::class, true)` condition, does it?

ausi (approved):

core-bundle/contao/elements/ContentAlias.php
leofeyer: @ausi and I have tested it and the CSS classes are assigned correctly to the alias element.

PR #5005, opened 2022-07-18
Fix database type error in contao:user:create command

fritzmg:
Fixes the following issue:

```
Invalid datetime format: 1366 Incorrect integer value: '' for column `tl_user`.`admin` at row 1
```


bytehead (approved):

leofeyer:
Thank you @fritzmg.

PR #5009, opened 2022-07-19
Keep the ResponseContextAccessor available for autowiring

aschempp:
Any fragment that needs to change the page title etc. needs the `ResponseContextAccessor`, so this service should continue to be available for autowiring.

FYI we might need to re-add this in Contao 5 since the deprecations have already been removed.


bytehead (commented):
I don't get this change - you're basically just removing the deprecation? 🤔

aschempp:
yes thats correct

fritzmg (approved):

m-vo (approved):

PR #5010, opened 2022-07-19
Fix adjustTime in tl_calendar_events

### LINKED ISSUE #4999: `endTime` can be string under certain conditions in `tl_calendar_events`
### Affected version(s)

4.13

### Description

Follow up to #3454.

> `endTime` can be zero for certain times and thus ?: would not work here. The `endTime` is not set if it is `null`.

If `addTime` is checked, but `startDate` is not set yet (it is empty by default), and user changes any other field that has `submitOnChange => true` more than once, then by the time of this check `$dc->activeRecord->endTime` is an empty string:

https://github.com/contao/contao/blob/95636f5853da18fba2e2bf74bf510ab5db91e44f/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php#L996-L1001

This results in a warning under PHP 7.4:

```
ErrorException:
Warning: date() expects parameter 2 to be int, string given

  at vendor/contao/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php:1000
  at tl_calendar_events->adjustTime(object(DC_Table))
     (vendor/contao/core-bundle/src/Resources/contao/drivers/DC_Table.php:2082)
  at Contao\DC_Table->edit()
     (vendor/contao/core-bundle/src/Resources/contao/classes/Backend.php:667)
  at Contao\Backend->getBackendModule('calendar', null)
     (vendor/contao/core-bundle/src/Resources/contao/controllers/BackendMain.php:168)
  at Contao\BackendMain->run()
     (vendor/contao/core-bundle/src/Controller/BackendController.php:49)
  at Contao\CoreBundle\Controller\BackendController->mainAction()
     (vendor/symfony/http-kernel/HttpKernel.php:152)
     (vendor/symfony/http-kernel/HttpKernel.php:74)
     (vendor/symfony/http-kernel/Kernel.php:202)
     (public/index.php:44)
```

I found out because my editors have two custom `submitOnChange` fields, that come in the first legend, i.e. before date/time settings, and I changed `addTime` to be checked by default. But it is reproducible on a fresh install:

1. set `APP_ENV=dev`
2. create an event
3. check `Add time`
4. change `Redirect target` to any other option
5. change `Redirect target` to any other option again

fritzmg:
Fixes #4999

There are two issues during a `submitOnChange` event:

1. `startDate` - although mandatory - will not be set during a `submitOnChange` event and thus `$arrSet['endTime']` can be empty.
2. `$dc->activeRecord->endTime` can be an empty string when not set during a `submitOnChange` event.

### DIFF
--- a/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
+++ b/calendar-bundle/src/Resources/contao/dca/tl_calendar_events.php
@@ -932,8 +932,8 @@ class tl_calendar_events extends Contao\Backend
 	 */
 	public function adjustTime(Contao\DataContainer $dc)
 	{
-		// Return if there is no active record (override all)
-		if (!$dc->activeRecord)
+		// Return if there is no active record (override all) or no start date has been set yet
+		if (!$dc->activeRecord || !$dc->activeRecord->startDate)
 		{
 			return;
 		}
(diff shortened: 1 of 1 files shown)


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5011, opened 2022-07-20
Stop using utf8_decode() and utf8_encode()

fritzmg:
Backport of https://github.com/contao/contao/pull/4829 to 4.9 to fix the CI. See also https://github.com/contao/contao/pull/4841

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
+++ b/core-bundle/src/Resources/contao/library/Contao/StringUtil.php
@@ -945,16 +945,6 @@ class StringUtil
 			return $str;
 		}
 
-		if ($from == 'UTF-8' && $to == 'ISO-8859-1')
-		{
-			return utf8_decode($str);
-		}
-
-		if ($from == 'ISO-8859-1' && $to == 'UTF-8')
-		{
-			return utf8_encode($str);
-		}
-
 		return mb_convert_encoding($str, $to, $from);
 	}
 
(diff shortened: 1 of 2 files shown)


fritzmg:
I am at a loss here - why is the result of `StringUtil::convertEncoding` in 4.9 different than in 4.13 - even though they are exactly the same (now)?

m-vo:
> I am at a loss here - why is the result of `StringUtil::convertEncoding` in 4.9 different than in 4.13 - even though they are exactly the same (now)?

The issue happens because the "substitute character" gets set to `none` before running this test - it is `0x63` in the upstream version.

This happens in `Patchwork\Utf8\Bootup::initMbstring()` where for some reason this default is hardcoded:

```php
if (extension_loaded('mbstring')) {
    // […]

    if ('none' !== strtolower(mb_substitute_character())) {
        mb_substitute_character('none');
    }
}
```

To fix the issue we could add a `mb_substitute_character(0x3F);` before the test to make sure the output is the same everywhere. (Then there will be another test case failing, but this just needs to be adjusted to match the upstream version as well).

fritzmg:
@m-vo thank you for the analysis, PR updated accordingly :)

m-vo (dismissed):

ausi (dismissed):

leofeyer (changes requested):
Why is it necessary to add a class property that is set and reset for every test? The `mb_substitute_character()` function is only called in one single method.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #4366, opened 2022-03-22
Revise drafts only when navigating back

ausi:
Fixes #3597
Fixes #4127

Todo:

- [x] Remove `DC_Table::$intPreserveRecord`
- [x] Improve styling of drafts

PR #5012, opened 2022-07-20
Correctly handle drafts in DC_Table column view

aschempp:
Fixes an array to string error because `$label` could be an array.

Edit: also correctly position the draft label because the parent element is not always `position: relative`.


ausi (dismissed):

ausi:
Does this not need to be fixed in Contao 4.9 too?

aschempp:
I don't think 4.9 has drafts?

ausi:
> I don't think 4.9 has drafts?

Why not? #4366 was merged into 4.9.

aschempp:
Rebased onto 4.9 🙃 

Edit: dunno why tests are not running 😢

ausi (dismissed):

ausi:
> Edit: dunno why tests are not running 😢

Probably because you force-pushed before switching the base?
Maybe try pushing an empty commit? (or force-pushing an amendet commit)?

ausi (approved):

leofeyer (changes requested):
The fix should be upstream compatible:

https://github.com/contao/contao/blob/0245320dfae5fa1a9938eccec6d5b451b161d86b/core-bundle/contao/drivers/DC_Table.php#L5119-L5122

aschempp:
not sure what you mean, but I don't think that works. `$label` can be an array and cannot be joined with a string!

leofeyer:
That's why there is the `!\is_array($label)` check.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -5004,7 +5004,7 @@ protected function listView()
     ';
 
 				$colspan = 1;
-				$label = ($blnDraft ? '<p class="draft-label">' . $GLOBALS['TL_LANG']['MSC']['draft'] . '</p> ' : '') . $label;
+				$label = $blnDraft ? '<p class="draft-label">' . $GLOBALS['TL_LANG']['MSC']['draft'] . '</p> ' : $label;
```
leofeyer, line 5007: This fix is unnecessary, because `$label` is never an array here. There is a `preg_replace()` on line 4975 that would already fail if `$label` was an array.

leofeyer (changes requested):

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: `preg_replace` allows for an array argument!

aschempp:
apparently the issue does not exist in 4.9, but it sure does on 4.13: https://github.com/contao/contao/pull/5119

PR #5013, opened 2022-07-20
Do not insert fields that do not exist in the DB

aschempp:
I have a field in my DCA that has a default value but does not exist in the database. DC_Table tries to add that field value when creating a new record.


fritzmg:
Typically you have to enable `doNotSaveEmpty` for that in your eval and return `null` in the `save_callback` on that field.

aschempp:
this is not about _saving_ a record, but creating the initial one, which uses the `default` DCA option

m-vo:
With "does not exist in the database" you mean the column does not exist, i.e. the migration did not run, yet?

aschempp:
the column does not exist, because thats a virtual field and not used in any back end palette. In my case it is only there for the front end form that is based upon the DCA fields.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -615,11 +615,13 @@ public function create($set=array())
...
 		}
 
+		$databaseFields = array_column($this->Database->listFields($this->strTable), 'name');
+
 		// Get all default values for the new entry
 		foreach ($GLOBALS['TL_DCA'][$this->strTable]['fields'] as $k=>$v)
 		{
 			// Use array_key_exists here (see #5252)
-			if (\array_key_exists('default', $v))
+			if (\array_key_exists('default', $v) && \in_array($k, $databaseFields, true))
```
ausi, line 624: Do we need to compare the columns in lowercase?

fritzmg: Yes, `$this->Database->listFields($this->strTable)` will return the column names in lower case.

fritzmg (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: I don't think so. Doctrine does, but our method does not.

ausi: Would the case match if you use entities as well?

I’d prefer to compare them in lowercase to be sure.

fritzmg: Yeah, you are right - our method does not return it in lower case, my bad.

leofeyer: Fact is that most DCAs use lowercase keys, so for this to work, the column names must also be lowercase, right?

ausi: We would need to lowercase them both, `$databaseFields` as well as `$k`, to be sure that we compare correctly IMO.

aschempp: I have no idea what you're talking about. The DCA uses the exact same column names as the database, we never lowercase that anywhere. Or should we lowercase the field names when writing to the database as well???

ausi (approved):

PR #5016, opened 2022-07-20
Fix several argument warnings on PHP methods

aschempp:
I got this error message in Sentry.io

> Deprecated: class_exists(): Passing null to parameter #1 ($class) of type string is deprecated

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -409,7 +409,7 @@ abstract class DataContainer extends Backend
 				}
 			}
 			// Use strlen() here (see #3277)
-			elseif (!\strlen($this->varValue))
+			elseif ('' !== (string) $this->varValue)
 			{
 				$arrData['eval']['required'] = true;
 			}
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -408,7 +408,6 @@ abstract class DataContainer extends Backend
 					$arrData['eval']['required'] = true;
 				}
 			}
-			// Use strlen() here (see #3277)
 			elseif ('' !== (string) $this->varValue)
 			{
 				$arrData['eval']['required'] = true;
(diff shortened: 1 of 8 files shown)


fritzmg:
Is it a warning or an error under PHP 8? If it is an error, it needs to be fixed in 4.9.

aschempp:
it's a deprecation (warning) 😉

fritzmg (dismissed):

aschempp:
found and fixed some more method argument issues (mostly passing NULL to a method that expects a string)

core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -409,7 +409,7 @@ protected function row($strPalette=null)
 				}
 			}
 			// Use strlen() here (see #3277)
-			elseif (!\strlen($this->varValue))
+			elseif ('' !== (string) $this->varValue)
```
m-vo: Comment above needs to be adjusted. Or maybe use `?: '' `?

aschempp: I would prefer to remove the comment since its pretty useless,  is meant for contao/core and itself references several other issues. Ok with you @leofeyer ?

aschempp: removed in 9c6f417e2fab70fa4b6b64a2835396125e73d5b9

m-vo (dismissed):

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: Not sure about this. The comment explicitly says to use `strlen()`…

ausi: `!\strlen(…)` would be the same same as `'' === (string) …`. This change would reverse the logic because it uses `!==` instead of `===`.

To get rid of the warning we could also use:

```suggestion
			elseif (!\strlen((string) $this->varValue)) 
```

leofeyer: I‘m fine with replacing `strlen()` if we are 100% sure that the behavior does not change. 👍 Obviously, we need to fix the `!==` then.

ausi: I’m 100% sure that `!\strlen((string) …)` is exactly the same same as `'' === (string) …`

aschempp: so why do we need to change anything on the PR if the current is exactly the same as the suggested change? The comment intended to say we must compare a string, not we have to use a specific function …

leofeyer: It is not the same? You have reversed the logic in your changes! See @ausi‘s comment above.

aschempp: Aah, right, thanks 😊 
fixed in bcfe39a1265ae7c65b87fa41ab181866056faceb then

leofeyer (approved):

PR #5020, opened 2022-07-21
Fix wrong UUID being applied when moving resources

### LINKED ISSUE #4962: Dbafs: Incorrect pid when moving file outside Contao
### Affected version(s)

4.13.*, ~~maybe even 4.9~~

### Description

When moving a file outside Contao (directly on the filesystem) the Dbafs sync will correct the path of the file but it'll not change the pid. This leads to a wrong behaviour when having "Downloads"-Elements which are pointed to an entire directory instead of specific files.

**The Setup:**

1. Create the following folder structure

```
files
├── folder-a
├── folder-b
│   └── file.zip
```

2. Sync the DB via backend or `contao-console contao:filesync`
3. Move `file.zip` from `folder-b` to `folder-a` directly on the filesystem (not in Contao backend)
4. Sync again
5. The record for `file.zip` will now have a `pid` which points to `folder-b` instead of `folder-a`

**Screenshot for clarification of the effect:**

image

m-vo:
Fixes #4962

### DIFF
--- a/core-bundle/src/Filesystem/Dbafs/Dbafs.php
+++ b/core-bundle/src/Filesystem/Dbafs/Dbafs.php
@@ -617,7 +617,7 @@ class Dbafs implements DbafsInterface, ResetInterface
 
             if (null !== ($newPath = $changedValues[ChangeSet::ATTR_PATH] ?? null)) {
                 $dataToUpdate['path'] = $this->convertToDatabasePath($newPath);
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5022, opened 2022-07-21
Lazy-load the RememberMeRepository

aschempp:
Similar to the problems fixed by https://github.com/contao/contao/pull/4265, having a service that requires a doctrine repository is problematic because it could trigger a DB connection.

In current dev, easily reproducible by running `contao-console list --format=json`

### DIFF
--- a/core-bundle/config/services.yaml
+++ b/core-bundle/config/services.yaml
@@ -811,7 +811,7 @@ services:
     contao.security.remember_me_token_provider:
         class: Contao\CoreBundle\Security\Authentication\RememberMe\RememberMeTokenProvider
         arguments:
-            - '@contao.repository.remember_me'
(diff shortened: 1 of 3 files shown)


bytehead:
CS 😉

bytehead:
What about adding `lazy: true` to the service definition?

m-vo:
> What about adding lazy: true to the service definition?

Then we would need to rely on the proxy manager...

Toflar:
FYI: Not anymore soon ;) https://github.com/symfony/symfony/pull/46752

aschempp:
> What about adding `lazy: true` to the service definition?

see https://github.com/contao/contao/pull/4265

Toflar (approved):

bytehead (approved):

PR #5031, opened 2022-07-22
Add a dark mode for the back end

leofeyer:
Also fixes #4144


Toflar:
I personally like this, thanks a lot! But I don't think we can ship this without an option to opt-out as I think it might really cause trouble for some projects depending on the combination of extensions etc.
What if we add a user setting. A dropdown with the values: "auto", "light" and "dark" -> default can be "auto" so people automatically see that there's a new dark theme :)

Not sure how to best implement that as CSS then because I'm not the expert here but I'd probably output it as a data attribute on `<html>`.

```html
<html data-backend-theme="light">
<head>
</head>
<style>
    [data-backend-theme="dark"] body {
        background: black;
    }
    [data-backend-theme="light"] body {
        background: white;
    }
</style>
```

leofeyer:
Having a user setting is actually pretty complicated and should be done in a follow-up PR. There are a lot of questions to discuss in advance such as "how do we know which mode to use before a user has logged in and we can read their preferences".

Aybee:
I would like to change the orange headline. It burns in my eyes and I hardly can see the white stuff in it.

Aybee:
Thank you 👍

PR #5032, opened 2022-07-24
Allow null for PaletteManipulator parent

fritzmg:
There are palettes that do not contain any legends, e.g. `tl_files`. If you want to add a field to that legend via the `PaletteManipulator` you currently have to do either of these methods:

```php
// The parent is null but PaletteManipulator::applyFallback will add the field to the end
PaletteManipulator::create()
    ->addField('groups', null)
    ->applyToPalette('default', 'tl_files')
;
```

```php
// The palette "foobar" does not exist but PaletteManipulator::applyFallback will add the field to the end
PaletteManipulator::create()
    ->addField('groups', 'foobar')
    ->applyToPalette('default', 'tl_files')
;
```

Since it is possible (and necessary) to define the `$parent` as `null`, this PR fixes the type hint in the doc. And since it is not necessary to define a parent the argument can be omitted altogether.

It's important to merge this upstream to `5.0` as there you _must_ define a not null `$parent`, otherwise you get the following error:

```
TypeError:
Contao\CoreBundle\DataContainer\PaletteManipulator::addField(): Argument #2 ($parent) must be of type array|string, null given, called in vendor\fritzmg\contao-file-access\src\DataContainer\FilesCallbacks.php on line 34

  at vendor\contao\contao\core-bundle\src\DataContainer\PaletteManipulator.php:57
  at Contao\CoreBundle\DataContainer\PaletteManipulator->addField('groups', null)
```

To circumvent this you would have to apply option no. 2 from above - but having to pass a legend that does not exist only to force the fallback (since there is no valid legend to begin with) makes no sense.

The same applies to when you want to add a new legend to the end of any palette.

### DIFF
--- a/core-bundle/src/DataContainer/PaletteManipulator.php
+++ b/core-bundle/src/DataContainer/PaletteManipulator.php
@@ -44,10 +44,10 @@ class PaletteManipulator
     /**
      * If the legend already exists, nothing will be changed.
      *
-     * @param string|array $parent
-     * @param bool         $hide
+     * @param string|array|null $parent
+     * @param bool              $hide
      */
-    public function addLegend(string $name, $parent, string $position = self::POSITION_AFTER, $hide = false): self
+    public function addLegend(string $name, $parent = null, string $position = self::POSITION_AFTER, $hide = false): self
     {
         $this->validatePosition($position);
 
@@ -65,13 +65,13 @@ class PaletteManipulator
      * If $position is PREPEND or APPEND, pass a legend as parent; otherwise pass a field name.
      *
      * @param string|array               $name
-     * @param string|array               $parent
+     * @param string|array|null          $parent
      * @param string|array|\Closure|null $fallback
      * @param string                     $fallbackPosition
      *
      * @throws PalettePositionException
      */
-    public function addField($name, $parent, string $position = self::POSITION_AFTER, $fallback = null, $fallbackPosition = self::POSITION_APPEND): self
+    public function addField($name, $parent = null, string $position = self::POSITION_AFTER, $fallback = null, $fallbackPosition = self::POSITION_APPEND): self
     {
         $this->validatePosition($position);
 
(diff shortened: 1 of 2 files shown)


bytehead (dismissed):

m-vo (dismissed):

aschempp (changes requested):
I don't think we should have a default value of `null` since the would lead people to not add the value. Also I'm unsure this is even necessary, since you probably should use a field as parent to place the item, instead of a palette?

fritzmg:
> I don't think we should have a default value of null since the would lead people to not add the value.

If I don't care about the position of a legend or field or I always want to add the legend or field at the end of the palette then I do not need to provide a value.

> Also I'm unsure this is even necessary, since you probably should use a field as parent to place the item, instead of a palette?

You mean in case of `tl_files` I should choose `meta` for example? But that would put the field not at the end of the palette anymore if another palette manipulation added another field after `meta` as well.

aschempp:
> If I don't care about the position of a legend or field or I always want to add the legend or field at the end of the palette then I do not need to provide a value.

Yes, but if you want to append then you need to provide the third argument anyway, because the default is "after". We could allow null value (instead of anything or empty string, since the value is ignored in append mode), but it sure must not be a default value.


> You mean in case of `tl_files` I should choose `meta` for example? But that would put the field not at the end of the palette anymore if another palette manipulation added another field after `meta` as well.

That is true. But it would not be at the end either if another field is added to the end. So you can't really control that anyway 😉

fritzmg:
> Yes, but if you want to append then you need to provide the third argument anyway, because the default is "after".

No, the third parameter is (currently at least) irrelevant, if the second parameter is `null` or a non-existent field or legend. The output will always be the same (it's appended to the last legend or at the end).

> But it would not be at the end either if another field is added to the end. So you can't really control that anyway 😉 

But I can make sure that it is at the end, no matter what any of the previously loaded bundles (defined via the Contao Manager Plugin) are doing in their regular DCAs.

ausi (dismissed):

leofeyer:
> It's important to merge this upstream to `5.0` as there you _must_ define a not null `$parent`, otherwise you get the following error:

Please create a separate PR for Contao 5 then.

fritzmg:
> Please create a separate PR for Contao 5 then.

Done: https://github.com/contao/contao/pull/5111

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5036, opened 2022-07-25
 Don’t cast null values in Model::convertToPhpValue 

### LINKED ISSUE #5034: Model::convertToPhpValue casts null values to type as well
### Affected version(s)

5.0

### Description

In 5.0 we introduced that the model's values are automatically cast from string (or also integer under PHP 8.1+ PDO) to their actual PHP type according to the column type.

However, this also applies to `null` values in the database. For example if you have a field with a definition like `smallint unsigned NULL` and a record's value is `null` in the database, then the value of the field will actually be `0` when loaded via the Model. Nullable database fields should still be able to hold `null` values in the Model (as is noted in the doc block anyway).

This applies to any non-string value in the database, specifically integers, floats, and booleans. String values will stay `null` (if nullable) as `Model::convertToPhpValue` does not convert those at all.

@ausi I think we probably also need to store whether a column field is nullable or not? And then don't cast those columns, if the value is null.

ausi:
Fixes #5034

@fritzmg while implementing this I concluded that storing if a column is nullable should not be necessary because the database would never return `null` in that case anyway. And `null` never needs to be casted because it is already retrieved correctly from the mysql connection.

### DIFF
--- a/core-bundle/contao/library/Contao/Model.php
+++ b/core-bundle/contao/library/Contao/Model.php
@@ -426,6 +426,11 @@ abstract class Model
 	 */
 	public static function convertToPhpValue(string $strKey, mixed $varValue): mixed
 	{
+		if (null === $varValue)
+		{
+			return null;
+		}
+
 		if (!self::$arrColumnCastTypes)
 		{
 			$path = Path::join(System::getContainer()->getParameter('kernel.cache_dir'), 'contao/config/column-types.php');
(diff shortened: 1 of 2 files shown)


fritzmg (approved):

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #5057, opened 2022-07-27
Fix empty form uploads causing an error

### LINKED ISSUE #5053: forms -> empty "file upload"-field leads to an error
### Affected version(s)

5.0.0-RC1

### Description

Steps to reproduce:

1. create a new form in CONTENT -> Forms
    -> select "Send form data via email"
    -> Data format: raw data
    -> Submission method: POST
2. create 2 fields in this new form: one with field type "file upload" and one with field type "Submit field"
3. publish form as content element in frontend
4. don't select a file for upload and send the form

The Contao error screen appears ("Es ist ein Fehler aufgetreten"...).

Details in /var/logs:

```
[2022-07-26T22:10:43.924022+02:00] request.INFO: Matched route "tl_page.2.root". {"route":"tl_page.2.root","route_parameters":{"_token_check":true,"_controller":"Contao\\FrontendIndex::renderPage","_scope":"frontend","_locale":"de","_format":"html","_canonical_route":"tl_page.2","parameters":"","pageModel":{"Contao\\PageModel":[]},"_route":"tl_page.2.root","_route_object":{"Symfony\\Component\\Routing\\Route":[]}},"request_uri":"https://contao5.xyz.tld/","method":"POST"} []
[2022-07-26T22:10:44.028500+02:00] request.CRITICAL: Uncaught PHP Exception TypeError: "Cannot access offset of type string on string" at /var/www/clients/web/contao5/vendor/contao/core-bundle/contao/forms/Form.php line 458 {"exception":"[object] (TypeError(code: 0): Cannot access offset of type string on string at /var/www/clients/web/contao5/vendor/contao/core-bundle/contao/forms/Form.php:458)"} []
```

Expected:
Form should be sent regardless of whether the "file upload"-field is empty or not.

fritzmg:
Fixes #5053 

Previously we only checked `!empty($_SESSION['FILES'])`. In Contao 5.0 we check for `!empty($arrFiles)` and we build `$arrFiles` via the upload widgets' values. This leads to an error since `$arrFiles` will be filled with an empty value if a non-mandatory upload field was submitted without a file.

### DIFF
--- a/core-bundle/contao/forms/Form.php
+++ b/core-bundle/contao/forms/Form.php
@@ -218,7 +218,11 @@ class Form extends Hybrid
 
 				if ($objWidget instanceof UploadableWidgetInterface)
 				{
-					$arrFiles[$objField->name] = $objWidget->value;
+					if ($objWidget->value)
+					{
+						$arrFiles[$objField->name] = $objWidget->value;
+					}
+
 					$hasUpload = true;
 				}
 


core-bundle/contao/forms/Form.php
```diff
@@ -218,7 +218,11 @@ protected function compile()
 
 				if ($objWidget instanceof UploadableWidgetInterface)
```
Toflar, line 219:
```suggestion
				if ($objWidget instanceof UploadableWidgetInterface && $objWidget->value)
```

Toflar (changes requested):
Looks correct to me but we can merge the if statement, no?

core-bundle/contao/forms/Form.php
fritzmg: No, we cannot, because then `$hasUpload = true;` will not be executed and consequently the wrong `enctype` will be output in the front end for example, breaking uploads altogether.

Toflar: Ah I see!

Toflar (approved):

m-vo (approved):

aschempp (approved):

leofeyer:
Thank you @fritzmg.

Issue #5041, opened 2022-07-25
"SYSTEM" -> "Vorschau-Links" broken


jochi44:
### Affected version(s)

5.0.0-RC1

### Description

After successfully creating and saving a preview-link in SYSTEM -> Vorschau-Links, accessing the menu item SYSTEM -> Vorschau-Links afterwards leads to an "internal server error".

Details in /var/logs:

```
[2022-07-25T20:12:50.176378+02:00] request.INFO: Matched route "contao_backend". {"route":"contao_backend","route_parameters":{"_route":"contao_backend","_scope":"backend","_token_check":true,"_controller":"Contao\\CoreBundle\\Controller\\BackendController::mainAction"},"request_uri":"https://contao5.xyz.tld/contao?do=preview_link&ref=CxfBwhUv","method":"GET"} []
[2022-07-25T20:12:50.193804+02:00] request.CRITICAL: Uncaught PHP Exception Error: "Cannot create references to/from string offsets" at /var/www/clients/web/contao5/vendor/contao/core-bundle/src/Translation/MessageCatalogue.php line 171 {"exception":"[object] (Error(code: 0): Cannot create references to/from string offsets at /var/www/clients/web/contao5/vendor/contao/core-bundle/src/Translation/MessageCatalogue.php:171)"} []

```

fritzmg:
I can confirm the problem. 

The issue is, that for some reason, `$GLOBALS['TL_LANG']['tl_preview_link']['share']` will, at that point, not be an array containing two elements anymore, but a reference to a string containing the value of `$GLOBALS['TL_LANG']['tl_preview_link']['share'][0]`.

The issue can be fixed by removing the assignment by reference here

https://github.com/contao/contao/blob/eb669536225740719a99927f1518605e4bfe2c48/core-bundle/src/Translation/MessageCatalogue.php#L164

and here

https://github.com/contao/contao/blob/eb669536225740719a99927f1518605e4bfe2c48/core-bundle/src/Translation/MessageCatalogue.php#L171

but I assume the problem lies somewhere else.

ausi:
The `Cannot create references to/from string offsets` error should get fixed by #5058 and #5059

> but I assume the problem lies somewhere else.

I think so too, we should probably investigate where `$GLOBALS['TL_LANG']['tl_preview_link']['share']` get’s modified.

ausi:
> > but I assume the problem lies somewhere else.
> 
[...]

PR #5058, opened 2022-07-27
Fix illegal string offsets in the translator

ausi:
Fixes the `Cannot create references to/from string offsets` bug from #5041


m-vo (approved):

aschempp (approved):

leofeyer:
Thank you @ausi.

PR #5060, opened 2022-07-27
Do not modify the global TL_LANG array

ausi:
Fixes the actual cause of #5041


aschempp (approved):

m-vo (approved):

bytehead (approved):

ausi:
Should we add a helper method to `ArrayUtil` that removes references?

```php
public static function resolveReferences(array $array): array 
{
    return array_map(fn ($value) => \is_array($value) ? self::resolveReferences($value) : $value, $array);
}
```

See https://3v4l.org/niF5u#v8.1.8

leofeyer:
Thank you @ausi.

PR #5065, opened 2022-07-29
Rename the database error type

aschempp:
It is currently impossible to distinguish between an exception and a database error. By renaming the type the Contao Manager will at least be able to know what _could_ be shown.

FYI we cannot rename the exception because the same type is already in use in Contao 4.13

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -71,7 +71,7 @@ class MigrateCommand extends Command
             if ($errors = $this->compileConfigurationErrors()) {
                 if ($asJson) {
                     foreach ($errors as $message) {
-                        $this->writeNdjson('error', ['message' => $message]);
+                        $this->writeNdjson('problem', ['message' => $message]);
                     }
                 } else {
                     foreach ($errors as $error) {
@@ -419,6 +419,7 @@ class MigrateCommand extends Command
     {
         $this->io->writeln(
             json_encode(
+                // make sure $type is the first in array but always wins
                 array_merge(['type' => $type], $data, ['type' => $type]),
                 JSON_INVALID_UTF8_SUBSTITUTE
             )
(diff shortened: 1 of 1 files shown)


m-vo (dismissed):

Toflar (dismissed):

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -419,6 +419,7 @@ private function writeNdjson(string $type, array $data): void
     {
         $this->io->writeln(
             json_encode(
+                // make sure $type is the first in array but always wins
                 array_merge(['type' => $type], $data, ['type' => $type]),
```
ausi:
```suggestion
                // make sure $type is the first in array but always wins
                ['type' => $type] + $data,
```

ausi (dismissed):

Toflar (approved):

ausi (approved):

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -71,7 +71,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             if ($errors = $this->compileConfigurationErrors()) {
                 if ($asJson) {
                     foreach ($errors as $message) {
-                        $this->writeNdjson('error', ['message' => $message]);
+                        $this->writeNdjson('problem', ['message' => $message]);
```
leofeyer, line 74: I would prefer `database-error` instead of `problem` because I don‘t want to end up with `error`, `problem`, `fault`, `issue`, `failure` and other synonyms in case we have to add more error categories in the future. 😉

aschempp: we've extensively discussed these options in the public call and agreed that `problem` would be sufficient 🤷

ausi: But wouldn’t `database-error` be sufficient as well?

aschempp: anything would be sufficient, `foobar` would work as well as long as the Manager knows about it 🤷 
To me
 - there was an **error** executing the script
 - there is a **problem** in your database (there is no error in your DB, but a problem in the config!)
 - there is a **warning** about your database (or something else, but the app works)

PR #5066, opened 2022-07-29
Auto-generate and dump the APP_SECRET during contao-setup

m-vo:


manager-bundle/src/Command/ContaoSetupCommand.php
```diff
@@ -89,7 +102,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $this->executeCommand(array_merge($php, [$this->consolePath], $command, $commandFlags), $output);
         }
 
-        $output->writeln('<info>Done! Please run the contao:migrate command to make sure the database is up-to-date.</info>');
+        $io->success('Done! Please run the contao:migrate command to make sure the database is up-to-date.');
```
aschempp: I guess that should be `$io->info`

aschempp (dismissed):

manager-bundle/src/Command/ContaoSetupCommand.php
m-vo: See 5c3802e840a2c229a81cf54202ef6a56b474ef8f.

fritzmg (changes requested):
As pointed out in Slack, `.env` should never be populated with any credentials, only with default values for the application. See https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local

> This file should be committed to your repository and (due to that fact) should only contain "default" values that are good for local development. This file should not contain production values.

> The `.env` and `.env.<environment>` files should be committed to the repository because they are the same for all developers and machines. However, the env files ending in `.local` (`.env.local` and `.env.<environment>.local`) should not be committed because only you will use them. In fact, the .gitignore file that comes with Symfony prevents them from being committed.

Anyone adhering to this standard would inadvertently commit this secret to their repository if they do not pay attention.

m-vo:
> Anyone adhering to this standard would inadvertently commit this secret to their repository if they do not pay attention.

Not generating a secret is worse IMHO. If you check in the `parameters.yaml` you've got the same problem. 

We could check if a `env.local` is there and write into that file instead, would you prefer this?

fritzmg:
> If you check in the `parameters.yaml` you've got the same problem.

You are not supposed to commit the `parameters.yml`. You _are_ however _specifically_ supposed to commit the `.env` file (see above). Yet you are dumping this secret into the `.env` file.

> We could check if a env.local is there and write into that file instead, would you prefer this?

No, it just never makes any sense under any circumstance to dump this into the `.env`. And it would not help the issue in any way because you cannot create a `.env.local` before you run `composer create-project` - and once you run `composer create-project` Contao will already have created the `APP_SECRET` in `.env` in the current state of this PR.

m-vo:
So what would you do? Always dump into the `.env.local` file instead? I would be fine with that as well.

Edit: see ba5b6d0918274f702d08e7e0db3310eab7ea8c11

fritzmg:
Yes, always dump into `.env.local`. Thus you will also need something like

```php
if (!file_exists($envPath = Path::join($this->projectDir, '.env'))) {
    $fs->touch($envPath);
}
```

otherwise the `.env.local` will not be loaded by the system.

ausi (approved):

fritzmg (approved):

manager-bundle/src/Command/ContaoSetupCommand.php
```diff
@@ -56,6 +59,25 @@ protected function configure(): void
...
+        if (empty($this->kernelSecret) || 'ThisTokenIsNotSoSecretChangeIt' === $this->kernelSecret) {
+            $filesystem = new Filesystem();
+
+            $dotenv = new DotenvDumper(Path::join($this->projectDir, '.env.local'), $filesystem);
+            $dotenv->setParameter('APP_SECRET', bin2hex(random_bytes(32)));
+            $dotenv->dump();
+
+            $io->info('An APP_SECRET was generated and written to your .env.local file.');
+
+            if (!$filesystem->exists($envPath = Path::join($this->projectDir, '.env'))) {
```
leofeyer, line 74: Why this? Is it technically necessary to have both an `.env` and `.env.local` file? Because if not, we should not generate the `.env` file.

fritzmg: The `.env.local` will only be loaded by Symfony's Dotenv component, if an `.env` file is present.

leofeyer: Why not use a `.env.local.php` file? It does not require an `.env` file to be present and it is the fastest way of using .env files.

fritzmg: The `.env.local.php` is created by running `composer dump-env prod`/`vendor/bin/contao-console dotenv:dump`. You should not generate it yourself. For the `APP_SECRET` to be defined in the `.env.local.php` it needs to be defined in the `.env.local` first. See also https://symfony.com/doc/current/deployment.html#b-configure-your-environment-variables

leofeyer:
Thank you @m-vo.

PR #5070, opened 2022-08-02
Fix DcaLoader exceptions

ausi:
Followup to https://github.com/contao/contao/pull/4881#issuecomment-1167421657

Loading DCAs inside a try/catch block was still buggy because a second `->load()` possibly caused a `Cannot declare class …, because the name is already in use` error.

Also removed the `$blnNoCache` parameter as this has never worked (would’ve caused the same redeclare class issue).

### DIFF
--- a/core-bundle/tests/Contao/DcaLoaderTest.php
+++ b/core-bundle/tests/Contao/DcaLoaderTest.php
@@ -35,8 +35,6 @@ class DcaLoaderTest extends TestCase
         $this->resetStaticProperties([System::class, Config::class, DcaLoader::class]);
 
         parent::tearDown();
-
-        (new Filesystem())->remove($this->getTempDir());
     }
 
     public function testThrowsTheSameExceptionWhenLoadingTwice(): void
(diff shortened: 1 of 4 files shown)


aschempp (dismissed):

Toflar (dismissed):

core-bundle/tests/Contao/DcaLoaderTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+    protected function tearDown(): void
+    {
+        unset($GLOBALS['TL_DCA'], $GLOBALS['TL_MIME']);
+
+        $this->resetStaticProperties([System::class, Config::class, DcaLoader::class]);
+
+        parent::tearDown();
+
+        (new Filesystem())->remove($this->getTempDir());
```
leofeyer: This should not be necessary:

https://github.com/contao/contao/blob/78abdd8cfd0520b960d12a1a6d9591e4709d56d8/test-case/src/ContaoTestCase.php#L47-L49

ausi:
```suggestion
```

ausi: Removed in bf799faaeba49ea39d3e9ad7e2f2d4c0e769ec70

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5073, opened 2022-08-02
Fix broken insert tag flags

ausi:
With #4512 we removed many global functions, but some of them were still used by the insert tag flags. This PR fixes those flags.


aschempp (approved):

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #5075, opened 2022-08-03
Do not include "index" in folderUrl

aschempp:
see https://github.com/contao/contao/pull/4581

### DIFF
--- a/core-bundle/src/Resources/contao/models/PageModel.php
+++ b/core-bundle/src/Resources/contao/models/PageModel.php
@@ -934,7 +934,7 @@ class PageModel extends Model
 					{
 						// If $folderUrl is not yet set, use the alias of the first
 						// parent page if it is not a root page (see #2129)
-						if (!$folderUrl && $objParentPage->alias)
+						if (!$folderUrl && $objParentPage->alias && $objParentPage->alias !== 'index' && $objParentPage->alias !== '/')
 						{
 							$folderUrl = $objParentPage->alias . '/';
 						}
--- a/core-bundle/tests/Contao/PageModelTest.php
+++ b/core-bundle/tests/Contao/PageModelTest.php
@@ -161,6 +161,83 @@ class PageModelTest extends ContaoTestCase
         $this->assertSame($databaseResultData[2]['subpages'], $pages->offsetGet(2)->subpages);
     }
 
+    /**
+     * @group legacy
+     * @runInSeparateProcess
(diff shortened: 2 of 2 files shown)


core-bundle/tests/Contao/PageModelTest.php
```diff
@@ -161,6 +161,83 @@ public function testFindPublishedSubpagesWithoutGuestsByPid(): void
         $this->assertSame($databaseResultData[2]['subpages'], $pages->offsetGet(2)->subpages);
     }
 
+    /**
+     * @group legacy
+     * @runInSeparateProcess
+     * @dataProvider folderUrlProvider
+     */
+    public function testFolderurlInheritsTheParentAlias(array $databaseResultData, string $expectedFolderUrl): void
```
Toflar:
```suggestion
    public function testFolderUrlInheritsTheParentAlias(array $databaseResultData, string $expectedFolderUrl): void
```

Toflar (dismissed):

core-bundle/src/Resources/contao/models/PageModel.php
```diff
@@ -934,7 +934,7 @@ public function loadDetails()
 					{
 						// If $folderUrl is not yet set, use the alias of the first
 						// parent page if it is not a root page (see #2129)
-						if (!$folderUrl && $objParentPage->alias)
+						if (!$folderUrl && $objParentPage->alias && $objParentPage->alias !== 'index' && $objParentPage->alias !== '/')
```
fritzmg, line 937: When would the parent page alias ever be `/`? You cannot fill `alias` with just `/`.

aschempp: it's a supported use case 🤷. I was surprised by this as well when implementing the new routing, but apparently that did work before as well (even though you maybe only add that in the DB 😂)
https://github.com/contao/contao/blob/5.x/core-bundle/src/Routing/RouteProvider.php#L174

fritzmg: > it's a supported use case 🤷

How so? You cannot create a page in the back end with just `/` as an alias.

aschempp: 🤷 I remember having to explicitly adding support because @leofeyer said that was previously supported and has to remain so. It shouldn't really affect this PR anyway 😉

leofeyer (approved):

PR #5077, opened 2022-08-03
Fix indeterministic order for non-unique filters in DC_Table

Toflar:
Currently, the `DC_Table` pagination might be broken, if you apply a filter on a non-unique field as the databases are not forced to return a deterministic order within all the results that share the same result.
See example database queries:

That’s because `course_day` is not a unique field. All that was done is configuring the DCA to sort by `course_day` in the back end and now the same course is displayed on both, pages 2 and 3.

Many courses happen on the very same day so if you `ORDER BY course_day` only, you get a correct order overall but you are never guaranteed to always get the same order of rows within all the entries that share the same date. That’s because `DC_Table` did not ask for it. Databases are explicitly allowed to return random orders within all of the entries that have the **exact same** `course_day` value here. That allows them to apply additional file sort optimizations. While this was always true, I think MySQL did not have any such optimization in place until version 5.7 (and I suppose in MariaDB it landed as well at some point) and also, those optimizations only seem to kick in with a certain amount of database entries. So it might very well be the case that it always worked for most of our users even though the queries were always indeterministic.

You can read more on the subject here: https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html

The solution is actually rather simple. We just always have to add `id` to the `ORDER BY` clause as well. That way, it's ordered by the value we want and within that set, it's always deterministc as it's ordered by ID there.
I also applied the fix to all the `ORDER BY sorting` queries because that is not unique either.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -211,7 +211,7 @@ class DC_Table extends DataContainer implements \listable, \editable
 			// Unless there are any root records specified, use all records with parent ID 0
 			if (!isset($GLOBALS['TL_DCA'][$table]['list']['sorting']['root']) || $GLOBALS['TL_DCA'][$table]['list']['sorting']['root'] === false)
 			{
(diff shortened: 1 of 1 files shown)


fritzmg (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5092, opened 2022-08-06
Check database version in migrate command

ausi:
As discussed in https://github.com/contao/contao/pull/4158#issuecomment-1169685620 (and on Slack)

This only fails if the configured server version causes the wrong doctrine platform to be used. 
So if the actual version is `8.0.29` but `8.0.28` is configured no error occurs.
But if `5.7.0` is configured it would show an error message like: 
`Wrong database version, please set it to "8.0.29". Expected "Doctrine\DBAL\Platforms\MySQL80Platform" was "Doctrine\DBAL\Platforms\MySQLPlatform".`


fritzmg (dismissed):
Tested and looks good to me. (However, I don't find the _"Expected ..."_ part of the error message particularly useful.)

fritzmg:
Btw. shouldn't this go into `4.9`? It would be just as helpful there (as for performance you can also define the database version in your `parameters.yaml`, `DATABASE_URL` environment variable or `config.yaml` there).

ausi:
> Btw. shouldn't this go into `4.9`?

Because of several merge conflicts that would arise I created a separate PR for 4.9 now: #5101

aschempp (changes requested):
If the output format is `ndjson`, this should not throw an exception (result in a `error` type message) but generate a `problem` (see https://github.com/contao/contao/pull/5065). Otherwise the Contao Manager cannot see if there is a connection problem/runtime exception or a database-level error.

ausi:
> If the output format is `ndjson`, this should not throw an exception

Done in 6340b8642a842f653645fd7fc166be1d9fe6400e

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -537,4 +546,40 @@ private function compileTableWarnings(Table $table): array
...
+        }
+
+        $message = sprintf('Wrong database version configured, please set it to "%s"', $version);
+
+        if ($currentVersion = $this->connection->getParams()['serverVersion'] ?? null) {
+            $message .= sprintf(', currently set to "%s"', $currentVersion);
+        }
+
+        if ($asJson) {
+            $this->writeNdjson('database-error', ['message' => $message]);
```
ausi: The correct name needs to be resolved in https://github.com/contao/contao/pull/5065#discussion_r940079950 first

ausi:
```suggestion
            $this->writeNdjson('problem', ['message' => $message]);
```

core-bundle/tests/Command/MigrateCommandTest.php
```diff
@@ -429,6 +434,65 @@ public function testDoesAbortOnFatalError(string $format): void
...
+
+        $code = $tester->execute(['--format' => $format, '--no-backup' => true], ['interactive' => 'ndjson' !== $format]);
+        $display = $tester->getDisplay();
+
+        $this->assertSame(1, $code);
+
+        if ('ndjson' === $format) {
+            $json = $this->jsonArrayFromNdjson($display)[0];
+
+            $this->assertSame('database-error', $json['type']);
```
ausi:
```suggestion
            $this->assertSame('problem', $json['type']);
```

aschempp (dismissed):
not sure why we need this for upstream merge, but code looks good 🙃

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -537,4 +546,40 @@ private function compileTableWarnings(Table $table): array
 
         return $warnings;
     }
+
+    private function validateDatabaseVersion(bool $asJson): bool
+    {
+        // TODO: Find a replacement for getWrappedConnection() once doctrine/dbal 4.0 is released
+        $driverConnection = $this->connection->getWrappedConnection();
+        $currentPlatform = \get_class($this->connection->getDatabasePlatform());
```
leofeyer: Is this going to be a problem?

ausi: I don’t think so, see https://github.com/contao/contao/pull/5133

leofeyer: See https://github.com/ausi/contao-contao/pull/4 😄

ausi: Merged.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5093, opened 2022-08-06
Fix undefined array key breadcrumb

ausi:


core-bundle/contao/dca/tl_templates.php
```diff
@@ -218,6 +218,7 @@ public function addBreadcrumb()
 		$GLOBALS['TL_DCA']['tl_templates']['list']['sorting']['root'] = array($strNode);
 
 		// Insert breadcrumb menu
+		$GLOBALS['TL_DCA']['tl_templates']['list']['sorting']['breadcrumb'] ??= '';
 		$GLOBALS['TL_DCA']['tl_templates']['list']['sorting']['breadcrumb'] .= '
```
ausi: Is this the correct fix, or should the next line just be `=` instead of `.=`?

ausi:
Failing CI unrelated, see #5094

core-bundle/contao/dca/tl_templates.php
leofeyer: The latter, I think. Theoretically, there could be a callback that has already added something to `$GLOBALS['TL_DCA']['tl_templates']['list']['sorting']['breadcrumb']` but I don‘t think that is a supported use-case, is it?

ausi:
```suggestion
		$GLOBALS['TL_DCA']['tl_templates']['list']['sorting']['breadcrumb'] = '
```

ausi: Done in ad05006a10846545d9eb2ca3d8c1ba3584c133ce

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5094, opened 2022-08-06
Fix compatibility with doctrine/dbal 3.3.8

ausi:
See https://github.com/doctrine/dbal/blob/f873a820227bc352d023791775a01f078a30dfe1/src/Platforms/MySQL/Comparator.php#L25

```
     * @internal The comparator can be only instantiated by a schema manager.
```


bytehead (approved):

Toflar (approved):

leofeyer:
Thank you @ausi.

ausi:
This needs to be fixed in Contao 4.13 (in a different file) too: https://github.com/contao/contao/blob/cb2e53f3a228cb27da0645ffd68a46d8bf94c9f1/installation-bundle/tests/Database/InstallerTest.php#L520
PR: #5098

Issue #5049, opened 2022-07-26
PurgePreviewLinksCron deletes not expired preview links


OMOSde:
### Affected version(s)

4.13

### Description

Preview links was removed also if they are not expired.

It seems there is a logic error in https://github.com/contao/contao/blob/4.13/core-bundle/src/Cron/PurgePreviewLinksCron.php#L38

use `DATE_SUB` instead of `DATE_ADD`?

aschempp:
it intentionally deletes preview links that are older than 31 days, since anything older than 30 days is not possible. What's wrong with this in your opinion?

fritzmg:
@aschempp the SQL query currently deletes every preview link that was _created_ at least 31 days in the _future_.

I think the intention was to write it this way:

```diff
-           'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_ADD(NOW(), INTERVAL 31 DAY))'
+           'DELETE FROM tl_preview_link WHERE expiresAt<=UNIX_TIMESTAMP(DATE_ADD(NOW(), INTERVAL 31 DAY))'
```

fritzmg:
### Reproduction:

1. Open the preview for a page.
2. Click on _Share URL_.
3. Save the created preview link.
4. Execute `TRUNCATE tl_cron_job`.
5. Execute `vendor/bin/contao-console cron:run`.

The preview link will be deleted, even though it has not expired yet.

ausi:
> use `DATE_SUB` instead of `DATE_ADD`?

I think this would be the correct fix, yes.

```diff
-           'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_ADD(NOW(), INTERVAL 31 DAY))'
+           'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 31 DAY))'
```

> I think the intention was to write it this way:

I don’t think so as the code comment explicitly states to not use the expiration on purpose: https://github.com/contao/contao/blob/4.13/core-bundle/src/Cron/PurgePreviewLinksCron.php#L20

fritzmg:
> I don’t think so as the code comment explicitly states to not use the expiration on purpose: https://github.com/contao/contao/blob/4.13/core-bundle/src/Cron/PurgePreviewLinksCron.php#L20

[...]

PR #5099, opened 2022-08-08
Fix the preview link purge job

aschempp:
Fixes https://github.com/contao/contao/issues/5049


Toflar (dismissed):

ausi:
Should we maybe check for both, `createdAt` **and** `expiresAt` to make sure we don’t break customized usages like the one outlined here: https://github.com/contao/contao/issues/5089#issuecomment-1206149076

aschempp:
it is intentionally not possible, but you can always bypass the cronjob if you wanna break the system 🤷

ausi:
I don’t think that people who adjust the DCA would know that they have to adapt the cron job too. What would be the downside of checking `expiresAt` too (and only deleting when both are out of time)

core-bundle/src/Cron/PurgePreviewLinksCron.php
```diff
@@ -35,7 +35,7 @@ public function __construct(Connection $connection)
     public function __invoke(): void
     {
         $this->connection->executeStatement(
-            'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_ADD(NOW(), INTERVAL 31 DAY))'
+            'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 31 DAY))'
```
ausi, line 38:
```suggestion
            'DELETE FROM tl_preview_link WHERE createdAt<=UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 31 DAY)) AND expiresAt<=UNIX_TIMESTAMP(NOW())'
```
_(untested)_

aschempp: This change would make it easy to create a preview link that lives indefinitely, and I don't think we should ever allow for that for security reasons.

ausi (dismissed):

core-bundle/src/Cron/PurgePreviewLinksCron.php
leofeyer: I have created a separate ticket hereto (see #5184).

leofeyer (approved):

PR #5104, opened 2022-08-09
Skip the database backup in contao:migrate command if there is no work to do

qzminski:
We bumped into a deployment performance problem. Our ecosystem has around 20 systems, and we run the `contao:migrate` command on every deployment. This takes quite some time, although it's not always necessary.

This pull request will change the command to not perform a database backup if there are no pending migrations, runonce files to be processed, or install tool commands to be executed.

/cc @Toflar


core-bundle/src/Command/MigrateCommand.php
```diff
@@ -186,6 +196,30 @@ private function executeCommand(InputInterface $input): int
         return 0;
     }
 
+    private function hasWorkToDo(): bool
+    {
+        // There are some pending migrations
+        if ($this->migrations->hasPending()) {
```
aschempp, line 212: can't you use `iterator_count($this->migrations->getPending()) > 0`?

aschempp (commented):
LGTM and would certainly improve my setups as well!

core-bundle/src/Command/MigrateCommand.php
qzminski: Wouldn't it run through all migrations? I introduced `hasPending` to return early on the first pending one.

ausi: You can use a `foreach` with an early return, this would also stop on the first migration that returns `true` for `shouldRun()`.

```suggestion
        foreach ($this->migrations->getPending() as $migration) {
```

ausi:
I’m not 100% happy with the implementation, migrations and schema are now checked twice 😬

A better approach might be to call the `backup()` method as soon as we find pending changes in the other methods instead of calling it at the very beginning. In the `backup()` method we could then make sure that it runs only once.

qzminski:
That was my initial idea to terminate early, but when I saw the complexity of `executecommand()` method, I went the easier way 😅

ausi:
Maybe we should merge this PR as it is and overhaul the whole class in Contao 5.1 later.

core-bundle/src/Command/MigrateCommand.php
Toflar: That's what `hasPending()` is doing, no?

ausi: Yes, but with the foreach we don’t need an extra method for that.

ausi (dismissed):

ausi (dismissed):

qzminski:
How do we fix the unsued variable in ECS now? 🤔

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -186,6 +196,30 @@ private function executeCommand(InputInterface $input): int
         return 0;
     }
 
+    private function hasWorkToDo(): bool
+    {
+        // There are some pending migrations
+        foreach ($this->migrations->getPending() as $migration) {
```
ausi:
```suggestion
        foreach ($this->migrations->getPending() as $migration) { // phpcs:ignore
```

Does this work?

qzminski: Unfortunately, it does not. I also tried other comments, but it seems that all inline annotations are ignored.

Shall I add `core-bundle/src/Command/MigrateCommand.php` to the `tools/ecs/config/default.php` to ignore the whole file there?

ausi: Yes, but only for the specific rule I guess: https://github.com/contao/contao/blob/12c7f37523f4f4bfedaaccb1b322ade711dab52c/tools/ecs/config/default.php#L24

qzminski: Done.

ausi (dismissed):

core-bundle/src/Command/MigrateCommand.php
leofeyer: Why not `if ($this->migrations->getPending()) { return true; }`? This would save us another iteration:

With foreach: https://3v4l.org/fIM4m
With if: https://3v4l.org/qDPLi

qzminski: Improved in [df8a88e](https://github.com/contao/contao/pull/5104/commits/df8a88e80bf389313a04333dd4d1c67d9e240da5).

ausi: But `if ($this->migrations->getPending())` would always be true, even if no migrations are pending.
(The unit tests probalby don’t fail because they don’t return a real generator for `getPending()`)

ausi: https://3v4l.org/m4KO5

leofeyer: I see. How about using `$this->migrations->getPending()->current()` then?

https://3v4l.org/2nmVd

ausi: PHPStan would probably complain because `iterable` can also be an array or an `IteratorAggregate` which have no `current()` method...

leofeyer: After discussing this with @ausi, we concluded that adding a `hasPending()` method is probably the easiest fix. I have therefore restored @qzminski's initial version in 267b1cb076cf832d74629eee06f2a9035854c71b. Sorry for the back and forth. 🙈

leofeyer (approved):

leofeyer:
Thank you @qzminski.

PR #5108, opened 2022-08-10
Do not override manually defined ptable configuration

dmolineus:
Since https://github.com/contao/contao/pull/1446 the ptable configuration is always overriden. The default algorithm assumes that only one table defines a ptable in the same backend module. But if I have following table structure `tl_root > tl_parent > tl_child` and both `tl_root` and `tl_child` refers to `tl_example` this does not work. 

Therefore checking if the ptable if configured manually is required.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/DcaLoader.php
+++ b/core-bundle/src/Resources/contao/library/Contao/DcaLoader.php
@@ -203,7 +203,7 @@ class DcaLoader extends Controller
 	 */
 	private function setDynamicPTable(): void
 	{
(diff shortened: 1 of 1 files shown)


fritzmg:
My original PR already included the check. However it was removed in https://github.com/contao/contao/pull/1881 afterwards.

leofeyer (approved):

leofeyer:
Thank you @dmolineus.

leofeyer:
Unfortunately, this change causes issues: #5157

PR #5112, opened 2022-08-10
Allow subpalettes based on value 0 in case of selects

dennisbohn:
I have the problem, that I want to show additional fields in a datacontainer, when the user selects a blank option generated with the eval 'includeBlankOption'.

In this case I have a list of locations from another table.

image

When no location is selected, the user should have the option to enter a custom location.

image

The object for the field looks like this.

```php
'home' => [
    'inputType'              => 'select',
    'foreignKey'             => 'tl_gs_home.name',
    'eval'                   => [
        'submitOnChange'     => true,
        'includeBlankOption' => true,
        'tl_class'           => 'clr w50'
    ],
    'sql'                    => "int(10) unsigned NOT NULL default '0'",
]
```

The option 'includeBlankOption' generates an option without a value which results in a int(0) when selected. So the subpalette would look like this.

```php
'subpalettes' => [
    'home_0'                    => 'street,housenumber,zipcode,city'
]
```

The problem is, that I cannot define a subpalette for this case because the complete logic is wrapped in an if statement checking for the value ($trigger) which is false, when the value is 0.

https://github.com/contao/contao/blob/12c7f37523f4f4bfedaaccb1b322ade711dab52c/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3319-L3342

It would be great, if 0 would be allowed in case of selects by using the if statement only in case of a checkbox.

It seems, there is no workaround at the moment. I tried to use the callback "fields.home.options" to generate a select with "none" instead of "0", used callback "fields.home.save" to convert "none" to "0" when writing to database and used callback "fields.home.load" to convert "0" back to "none" when reading from database.

Unfortunately the load callback is completely ignored because the value comes directly from the database.

https://github.com/contao/contao/blob/12c7f37523f4f4bfedaaccb1b322ade711dab52c/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3297-L3299

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3316,9 +3316,9 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 						}
 					}
 
-					if ($trigger)
+					if (($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['inputType'] ?? null) == 'checkbox' && !($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['eval']['multiple'] ?? null))
 					{
-						if (($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['inputType'] ?? null) == 'checkbox' && !($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['eval']['multiple'] ?? null))
+						if ($trigger)
 						{
 							$sValues[] = $name;
 
@@ -3328,16 +3328,16 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 								$subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$name];
 							}
 						}
-						else
-						{
-							$sValues[] = $trigger;
-							$key = $name . '_' . $trigger;
+					}
+					else
+					{
+						$sValues[] = $trigger;
+						$key = $name . '_' . $trigger;
 
-							// Look for a subpalette
-							if (isset($GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key]))
-							{
-								$subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key];
-							}
+						// Look for a subpalette
+						if (isset($GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key]))
+						{
+							$subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key];
 						}
 					}
 				}


dennisbohn:
Is everything ok with the pull request?

ausi (approved):

ausi:
Does this issue appear in Contao 4.9 too? If so we should probably backport this, or rebase the pull request on the 4.9 branch.

dennisbohn:
Tested it with 4.9 today with the same behavior, so a new pull request for 4.9 would make sense.

Maybe it would also make sense to allow subpalettes for disabled checkboxes.

```php
if (!$GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['eval']['multiple'])
{
  if ($trigger)
  {
    if ($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['inputType'] == 'checkbox')
    {
      $sValues[] = $name;
      $key = $name;
    }
    else
    {
      $sValues[] = $trigger;
      $key = $name . '_' . $trigger;
    }
  }
  else
  {
    $key = $name . '_';
  }

  // Look for a subpalette
  if (isset($GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key]))
  {
    $subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key];
  }
}
```

This would result in the following subpalettes:

Checkbox, activated
- active

Checkbox, not activated
- active_

Select, value "a" selected
- active_a

Select, no value selected
- active_

dennisbohn:
Or we say, that the subpalette always just ends with a underscore, when trigger is 0 or "".

```php
if (!$GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['eval']['multiple'])
{
  if ($GLOBALS['TL_DCA'][$this->strTable]['fields'][$name]['inputType'] == 'checkbox')
  {
    if ($trigger)
    {
      $sValues[] = $name;
      $key = $name;
    }
    else
    {
      $key = $name . '_';
    }

    // Look for a subpalette
    if (isset($GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key]))
    {
      $subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key];
    }
  }
  else
  {
    if ($trigger)
    {
      $sValues[] = $trigger;
      $key = $name . '_' . $trigger;
    }
    else
    {
      $key = $name . '_';
    }

    // Look for a subpalette
    if (isset($GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key]))
    {
      $subpalettes[$name] = $GLOBALS['TL_DCA'][$this->strTable]['subpalettes'][$key];
    }
  }
}
```

This would result in the following subpalettes:

Checkbox, activated
- active

Checkbox, not activated [char(1) NOT NULL default ''] and [tinyint(1)]
- active_

Select, value "a" selected
- active_a

Select, no value selected [varchar(255) NOT NULL default ''] and [int(10) unsigned NOT NULL default '0']
- active_

dennisbohn:
The last code snippets conflicts with a javascript for the ajax requests and toggling the checkboxes.

dennisbohn:
It seems, to allow a subpalette for a disabled checkbox, there are much more changes needed, because the ajax request just loads the subpalette for the activated checkbox and toggles it after that.

Maybe its better to make a feature request and find a workaround for a fast soluition.

ausi:
> Maybe its better to make a feature request and find a workaround for a fast soluition.

Agree. Probably better to focus in pull request just on just allowing subpalettes with falsey values (`0`, `false`, `''`, `'0'`, `null`).

leofeyer:
Thank you @dennisbohn.

ausi:
> Tested it with 4.9 today with the same behavior, so a new pull request for 4.9 would make sense.

We should not forget to create a pull request for 4.9 too, @dennisbohn can you take care of that?

dennisbohn:
> > Tested it with 4.9 today with the same behavior, so a new pull request for 4.9 would make sense.
> 
> We should not forget to create a pull request for 4.9 too, @dennisbohn can you take care of that?

done :)

PR #5116, opened 2022-08-10
Add the MemberActivationMailEvent

fritzmg:
Within `ModuleRegistration` of the `contao/core-bundle` we still have implementations specific to `contao/newsletter-bundle`. This is because we want to provide an additional `##channels##` simple token for the activation email text when newsletter subscriptions are enabled in the registration module. And currently there is no way to manipulate the email message or the provided simple tokens for the email message.

This PR introduces a `MemberActivationMailEvent` with which you can add additional simple tokens for the activation email. You can also change the subject of the email or the whole text - or even prevent sending the activation mail (through `OptInToken::send`) altogether. 

The latter would be useful for `terminal42/notification_center` for example which allows you to send the activation link separately - and currently the Notification Center disables `reg_enable` of the module as a work around, so that the activation email is not sent. With this new event the Notification Center can extract the already provided tokens from the event and provide them in their own notification while also preventing the default email from being sent.


aschempp (commented):
shouldn't we rather not send the email if the event (propagation) is stopped (instead of sendOptInToken)?

fritzmg:
> shouldn't we rather not send the email if the event (propagation) is stopped (instead of sendOptInToken)?

Hmm, you might still want to be able to collect all simple tokens from all registered event listeners. Imagine if the Notification Center stops propagation - but is executed before the Newsletter's listener. Then the `channels` simple token would be missing. Of course the Notification Center will set a very low priority on its own listener so it is more likely to be the last decider. However I still feel like the decision to send or not should be independent from the event's propagation.

Toflar:
Will be 4.13 compatible so this isn't of great use for me just yet 😓

fritzmg:
Yes, but I meant in principle. It's not just for the NC anyway - see also the issue with the newsletter-bundle.

core-bundle/src/Event/MemberActivationMailEvent.php
```diff
@@ -0,0 +1,91 @@
...
+
+        return $this;
+    }
+
+    public function getSimpleTokens(): array
+    {
+        return $this->simpleTokens;
+    }
+
+    public function setSendOptInToken(bool $sendOptInToken): self
```
Toflar: Can we not use setters and getters here but be more speaking?

* disableSendingTheOptInToken(): self
* enableSendingTheOptInToken(): self
* shouldSendOptInToken(): bool

core-bundle/src/Event/MemberActivationMailEvent.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public function setSimpleTokens(array $simpleTokens): self
+    {
+        $this->simpleTokens = $simpleTokens;
+
+        return $this;
+    }
+
+    public function setSimpleToken(string $token, string $value): self
```
Toflar: Can we have an `addSimpleToken(string $key, string $value)`?

newsletter-bundle/src/EventListener/MemberActivationMailListener.php
```diff
@@ -0,0 +1,43 @@
...
+            return;
+        }
+
+        $channels = $this->connection->fetchFirstColumn('SELECT title FROM tl_newsletter_channel WHERE id IN (?)', [$newsletter], [Types::SIMPLE_ARRAY]);
+
+        if (empty($channels)) {
+            return;
+        }
+
+        $event->setSimpleToken('channels', implode("\n", $channels));
```
Toflar: This place is where you'd use `addSimpleToken()`.

Toflar (changes requested):
I think the event can be improved :)

core-bundle/src/Event/MemberActivationMailEvent.php
Toflar: And a `removeSimpleToken(string $key)`?

Toflar (dismissed):
❤️

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5117, opened 2022-08-11
Change default CSS class names for content elements

### LINKED ISSUE #5082: Use of slash in new css class names for content elements
### Affected version(s)

5.0.0-RC1

### Description

The divs for "content elements" get new class names in Contao 5 including a slash as separator (eg. content_element/gallery).
Although a slash in css class names is allowed according to the wc3 css specification, I'm not shure if another character would be a better choice.

Reason:
When defining css rules for elements including slashes in a class name, the slash must be escaped.
I'm not shure, how many people even know this.

```
.content_element\/gallery ul {
    list-style: none;
}
```

What do you think?

ausi:
Fixes #5082


bytehead (approved):

leofeyer:
Thank you @ausi.

PR #5125, opened 2022-08-12
Fix warning in DcaExtractor

### LINKED ISSUE #5122: DcaExtractor triggers warning if DCA field has no sql definition
### Affected version(s)

4.13, maybe 4.9 too

### Description

Hello, 

I upgraded a project to 4.13 and PHP 8 and run into a Warning triggered in DcaExtractor

[PR 2762 ](https://github.com/contao/contao/pull/2762) is related and claims to fix this, however it does not under all conditions: In my case I have a field without SQL definition which is a MCW field on which I update another table manually (via load and save callback). Now in the extractor [$blnLoadFromFile will be set to true](https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L409), because of the missing sql definition. It will then proceed to [actually load `database.sql`](https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L447-L457) file from an installed extension. 

Then it will trigger [a warning here](https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php#L459): Warning: Undefined array key "tl_my_custom_table"

To reproduce:
Have a DCA field with inputType, but without sql definition. 
Have a database.sql
Use DcaExtractor

bezin:
Fixes #5122

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
+++ b/core-bundle/src/Resources/contao/library/Contao/DcaExtractor.php
@@ -456,11 +456,11 @@ class DcaExtractor extends Controller
 				static::$arrSql = $arrSql;
 			}
 
-			$arrTable = static::$arrSql[$this->strTable];
+			$arrTable = static::$arrSql[$this->strTable] ?? null;
 			$engine = null;
 			$charset = null;
 
-			if (isset($arrTable['TABLE_OPTIONS']))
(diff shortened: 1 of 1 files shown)


bezin:
(4.9 is probably affected, too, but the fix is different so I'd rather do these separately)

leofeyer:
Closing this in favor of #5129.

PR #5130, opened 2022-08-12
Only set login constants in master request

fritzmg:
This adds a check for the master request in the `LegacyLoginConstantsListener` - otherwise the constants might get set twice in one request due to sub requests, leading to a warning.

### DIFF
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -34,6 +34,10 @@ class LegacyLoginConstantsListener
 
     public function __invoke(RequestEvent $event): void
     {
+        if (!$event->isMasterRequest()) {
+            return;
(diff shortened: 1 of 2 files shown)


aschempp:
Shouldn't we rather/also not set the constants if the method is called again?

fritzmg:
I thought about that too. However, in `ContaoFramework::setConstants` we only check `\defined` for constants where we know you are supposed to be able to set them before the framework is initialized, e.g. `TL_MODE` and `TL_SCRIPT`. The same does not apply to the login constants and thus I think we should not check for `\defined` there either.

fritzmg:
Added defined check (see https://github.com/contao/contao/pull/5137#issuecomment-1213556296). #5137 should be merged first.

aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #5132, opened 2022-08-12
Check for column options in AdjustSearchUrlLengthListener

fritzmg:
Fixes an

```
Undefined array key "row_format"
```

error.

### DIFF
--- a/core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
+++ b/core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
@@ -34,15 +34,17 @@ class AdjustSearchUrlLengthListener
     {
         $schema = $event->getSchema();
 
-        if (
-            !$schema->hasTable('tl_search')
-            || !($table = $schema->getTable('tl_search'))
-            || !$table->hasColumn('url')
-            || !($column = $table->getColumn('url'))
-            || !$table->hasIndex('url')
-        ) {
+        if (!$schema->hasTable('tl_search')) {
             return;
         }
+
+        $table = $schema->getTable('tl_search');
+
+        if (!$table->hasColumn('url') || !$table->hasIndex('url')) {
+            return;
+        }
+
+        $column = $table->getColumn('url');
 
         // Get maximum index size for this table
         $maxIndexSize = $this->getMaximumIndexSize($table);
--- a/core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
+++ b/core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
@@ -32,9 +32,10 @@ class AdjustSearchUrlLengthListener
 
     public function __invoke(GenerateSchemaEventArgs $event): void
     {
+        $schema = $event->getSchema();
+
         if (
-            !($schema = $event->getSchema())
-            || !$schema->hasTable('tl_search')
+            !$schema->hasTable('tl_search')
             || !($table = $schema->getTable('tl_search'))
             || !$table->hasColumn('url')
             || !($column = $table->getColumn('url'))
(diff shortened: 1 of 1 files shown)


core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
```diff
@@ -48,7 +49,8 @@ public function __invoke(GenerateSchemaEventArgs $event): void
 
         // Reduce maximum length if collation is not "ascii_bin"
         if ('ascii_bin' !== $column->getPlatformOption('collation')) {
-            $bytesPerChar = 'utf8mb4' === $table->getOption('charset') ? 4 : 3;
+            $charset = $table->hasOption('charset') ? $table->getOption('charset') : 'utf8mb4';
```
leofeyer, line 54: Can we safely assume `utf8mb4` as fallback value?

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
```diff
@@ -64,14 +66,16 @@ public function __invoke(GenerateSchemaEventArgs $event): void
 
     private function getMaximumIndexSize(Table $table): int
     {
-        $engine = $table->getOption('engine');
+        $engine = $table->hasOption('engine') ? $table->getOption('engine') : 'InnoDB';
```
leofeyer, line 71: Can we safely assume `InnoDB` as fallback value?

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
fritzmg: It's the same in our existing `MysqlInnodbRowSizeCalculator`. /cc @ausi

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
fritzmg: It's the same in our existing `MysqlInnodbRowSizeCalculator`. /cc @ausi

core-bundle/src/EventListener/DoctrineSchema/AdjustSearchUrlLengthListener.php
ausi: I think so?
https://github.com/contao/contao/blob/e5995ffa484b0dddb8acdc20eb751f0e976ddcd9/manager-bundle/skeleton/config/config.yaml#L55

Or can we ask doctrine for the actually configured default instead?

leofeyer: I guess we can rely on what we have configured in the skeleton config.yml.

fritzmg: The default table options can be changed - and this is done quite frequently, especially on environments with older MySQL versions (see also https://docs.contao.org/manual/de/installation/systemvoraussetzungen/#mysql-mindestanforderungen).

> Or can we ask doctrine for the actually configured default instead?

It would be correct to do so, yes. However, usually the table option for the `charset` for example is unlikely to be missing. It is only missing in the functional tests as the database is set up differently there.

ausi: So we keep the hard coded default `utf8mb4` for now as is until we see issues with it?

leofeyer: I guess so.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5127, opened 2022-08-12
Merge branch '4.9' into 4.13

leofeyer:

PR #5133, opened 2022-08-12
Fix PHPStorm warning

ausi:
See https://github.com/contao/contao/pull/5127#discussion_r944257781

But not sure if this should be merged as I think PHPStorm is wrong here. 
`Connection::getDatabasePlatform()` can never return NULL I think.


ausi:
> `Connection::getDatabasePlatform()` can never return NULL I think.

PHPStan agrees and fails now: https://github.com/contao/contao/runs/7805465964?check_suite_focus=true

leofeyer:
See https://github.com/ausi/contao-contao/pull/4

ausi:
Closed in favor of #5092

PR #5137, opened 2022-08-12
Set login constants in request listener

fritzmg:
#4968 + #5130 for Contao 4.13

### DIFF
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -22,10 +22,7 @@ use Symfony\Component\HttpKernel\Event\RequestEvent;
  */
 class LegacyLoginConstantsListener
 {
-    /**
-     * @var ContaoFramework
-     */
-    private $framework;
+    private ContaoFramework $framework;
 
     public function __construct(ContaoFramework $framework)
     {
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -34,6 +34,10 @@ class LegacyLoginConstantsListener
 
     public function __invoke(RequestEvent $event): void
     {
+        if ($event->isMainRequest()) {
+            return;
+        }
+
         $this->framework->setLoginConstants();
     }
 }
--- a/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
+++ b/core-bundle/src/EventListener/LegacyLoginConstantsListener.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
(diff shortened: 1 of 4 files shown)


fritzmg:
The reason why the functional tests fail is the following:

1. When the functional tests are set up, our `DcaSchemaProvider` will initialize the `ContaoFramework` (see [1]) when loading fixtures.
2. However, this is done before there is a request in the request stack. Thus in our current logic the `ContaoFramework` will also already set the constants to `false`.
3. Later on in the functional test a request is pushed to the request stack and processed. Now our `LegacyLoginConstantsListener` wants to set the constants _again_, leading to an error.

[1]

```
contao\core-bundle\src\Framework\ContaoFramework.php:220
contao\core-bundle\src\Framework\ContaoFramework.php:121
contao\core-bundle\src\Doctrine\Schema\DcaSchemaProvider.php:317
contao\core-bundle\src\Doctrine\Schema\DcaSchemaProvider.php:58
contao\core-bundle\src\EventListener\DoctrineSchemaListener.php:35
contao\vendor\symfony\doctrine-bridge\ContainerAwareEventManager.php:68
contao\vendor\doctrine\orm\lib\Doctrine\ORM\Tools\SchemaTool.php:408
contao\vendor\doctrine\orm\lib\Doctrine\ORM\Tools\SchemaTool.php:115
contao\vendor\doctrine\orm\lib\Doctrine\ORM\Tools\SchemaTool.php:93
contao\test-case\src\FunctionalTestCase.php:138
contao\test-case\src\FunctionalTestCase.php:35
contao\core-bundle\tests\Functional\PageControllerTest.php:102
contao\core-bundle\tests\Functional\PageControllerTest.php:57
```

Now the question is how do we handle this. Checking whether the constants have already been set within `ContaoFramework::setLoginConstants` would fix the tests. However, this would not technically be correct as our functional tests would now potentially have the wrong value defined for these constants when the request is processed. On the other hand our functional tests do not really need these constants and in the real world a setup like this would never happen anyway. Besides, it would just be a known limitation of these legacy constants.

Thus for simplicity I would opt for checking whether the constants have already been defined.

aschempp (dismissed):
LGTM. Yes we should check the constants anyway because one could reset the framework (or create a second instance like in tests).

leofeyer:
> and in the real world a setup like this would never happen anyway.

Can we be sure about that?

> However, this is done before there is a request in the request stack.

Can we not define the constants if there is no request?

fritzmg:
> Can we not define the constants if there is no request?

Then they would never be defined if there never is a request (e.g. on the command line).

leofeyer:
But as you said, _this would not technically be correct as our functional tests would now potentially have the wrong value defined for these constants when the request is processed_. Do we really want to merge this? @contao/developers /cc

ausi:
I think if we have functional tests that rely on the values of those constants they would need to run in a seperate process anyway. I think this should be merged.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5139, opened 2022-08-13
Add static URLs to data-icon attributes

### LINKED ISSUE #5030: Image getPath used for backend icons should return absolute urls
### Affected version(s)

5.0

### Description

Problem is the following markup for the toggle icon:
```php
<img 
    src="/system/themes/flexible/icons/visible.svg" 
    width="16" 
    height="16" 
    alt="Element ID 1 veröffentlichen/unveröffentlichen" 
    data-icon="system/themes/flexible/icons/visible.svg" 
    data-icon-disabled="system/themes/flexible/icons/invisible.svg" 
    data-state="1"
>
```
Specifically the `data-icon` and `data-icon-disabled` attributes are not absolute so after one toggles the icon it changes to 
```php
<img 
    src="system/themes/flexible/icons/visible.svg"
    ...
>
```
which leads to a problem now that the `base`-tag is missing. 
If we use a custom backend route or custom backend controller generating an url like `://domain/contao/test` for the backend.

Generally we should now always use absolute urls.

So the cause generating this path is here:
https://github.com/contao/contao/blob/87b3a7cf676461608c08146678c046bd5debf4d8/core-bundle/contao/library/Contao/Image.php#L24

ausi:
Fixes #5030

### DIFF
--- a/core-bundle/contao/library/Contao/Image.php
+++ b/core-bundle/contao/library/Contao/Image.php
@@ -74,6 +74,27 @@ class Image
 	}
 
 	/**
+	 * Get the URL to an image
+	 *
+	 * @param string $src The image name or path
+	 *
+	 * @return string The path-absolute or absolute URL depending on the assets context
+	 */
+	public static function getUrl($src)
+	{
+		$src = System::urlEncode(self::getPath($src));
+
+		if (!$src)
+		{
+			return '';
+		}
+
+		$context = str_starts_with($src, 'assets/') ? 'assets_context' : 'files_context';
+
+		return Controller::addStaticUrlTo($src, System::getContainer()->get('contao.assets.' . $context));
+	}
+
+	/**
 	 * Generate an image tag and return it as string
 	 *
 	 * @param string $src        The image path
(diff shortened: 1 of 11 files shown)


aschempp (commented):
`Controller::addStaticUrlTo` does not check whether the path is currently relative. Should we also `ltrim('/')` in the Controller method?

ausi:
> Should we also `ltrim('/')` in the Controller method?

I don’t think so, as the meaning would change. If we want to support path-absolute URLs in the `addStaticUrlTo()` method we’d need to only apply the host part (without the path part) of the assets context. But this is unreleated to this PR IMO.

leofeyer:
Why are we not adjusting the `Image::getPath()` method?

ausi:
> Why are we not adjusting the `Image::getPath()` method?

This would be a BC break that would affect way to many usages I think.

If we want something shorter I’d add an `Image::getUrl()` method that does something similar to  `Controller::addStaticUrlTo(Image::getPath(…))`

ausi:
> I’d add an `Image::getUrl()` method

Done in 09a0b94a67809065b66e723cd30905c8c8217c6c

core-bundle/contao/library/Contao/Image.php
```diff
@@ -73,6 +73,27 @@ public static function getPath($src)
...
+	public static function getUrl($src)
+	{
+		$src = System::urlEncode(self::getPath($src));
+
+		if (!$src)
+		{
+			return '';
+		}
+
+		$context = str_starts_with($src, 'assets/') ? 'assets_context' : 'files_context';
```
aschempp, line 92: This should be inverted (to check for files) because we use the asset context e.g. for system/themes as well.

aschempp (commented):
Good idea with the `Image::getUrl`!

core-bundle/contao/library/Contao/Image.php
ausi: But this check was always made this way: https://github.com/contao/contao/blob/2f5d8fffe300fa930942a3bff69c1ab57c473d29/core-bundle/contao/library/Contao/Image.php#L128

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5141, opened 2022-08-13
Don’t use sprintf() for alias URLs

### LINKED ISSUE #5007: PageModel::getAbsoluteUrl parameters are automatically URL encoded
### Affected version(s)

5.0

### Description

In our news and events `SitemapListener` we use [`$objPage->getAbsoluteUrl('/%s')`](https://github.com/contao/contao/blob/5.0.0-RC1/news-bundle/src/EventListener/SitemapListener.php#L88) - i.e. we use `%s` as a placeholder for the news alias to be later inserted by `sprintf()`. However, in Contao 5.0, the resulting URL will have an URL encoded representation of the `%` character, resulting in an URL like

```
https://example.local/news/entry/%25s
```
instead of
```
https://example.local/news/entry/%s
```

which means the URL will not properly be generated by the listener and thus you end up with multiple `https://example.local/news/entry/%25s` URLs in your sitemap instead of the real news URL.

For some reason the `router` automatically URL encodes the parameters in Contao 5.0 - but does not do so in Contao 4.13 (where we use the same listener and same principle).

ausi:
Fixes #5007


calendar-bundle/contao/classes/Calendar.php
```diff
@@ -410,7 +395,14 @@ protected function addEvent($objEvent, $intStart, $intEnd, $strUrl, $strBase='',
 				break;
 
 			default:
-				$link = sprintf(preg_replace('/%(?!s)/', '%%', $strUrl), ($objEvent->alias ?: $objEvent->id));
+				$link = $objParent->getAbsoluteUrl('/' . ($objEvent->alias ?: $objEvent->id));
+
+				// Backwards compatibility (see #8329)
+				if ($strBase && !preg_match('#^https?://#', $link))
```
aschempp: This should never be necessary if we generate the URL here ourself.

faq-bundle/src/EventListener/SitemapListener.php
```diff
@@ -57,35 +56,27 @@ public function __invoke(SitemapEvent $event): void
                 continue;
             }
 
-            // Get the URL of the jumpTo page
-            if (!isset($arrProcessed[$objFaq->jumpTo])) {
-                $objParent = $this->framework->getAdapter(PageModel::class)->findWithDetails($objFaq->jumpTo);
+            $objParent = $this->framework->getAdapter(PageModel::class)->findWithDetails($objFaq->jumpTo);
```
aschempp, line 59: Afair `PageModel::findWithDetaild` removed it from the registry so this would be quite the performance penalty. Maybe we should still cache the jumpTo object instead?

aschempp (changes requested):

faq-bundle/src/EventListener/SitemapListener.php
ausi: Unable to reproduce. The numer of DB queries before and after this change didn’t change in my test.

> Afair `PageModel::findWithDetaild` removed it from the registry

What do you mean by that?

calendar-bundle/contao/classes/Calendar.php
ausi: Unrelated to this PR.
This was added in https://github.com/contao/core/pull/8329 @fritzmg do you know if this can be removed?

faq-bundle/src/EventListener/SitemapListener.php
aschempp: PageModel::loadDetails()` calls `preventSaving` which removes it from the registry. https://github.com/contao/contao/blob/5.x/core-bundle/contao/models/PageModel.php#L1067
not sure if I‘m missing something?

calendar-bundle/contao/classes/Calendar.php
aschempp: How is that not related? Previously the URL was passed in to the method and could therefore be relative. Now it uses `PageModel::getAbsoluteUrl` so the http-Regex can never be true?

ausi: > How is that not related?

This code was just moved, I didn’t change that.

> Previously the URL was passed in to the method and could therefore be relative. Now it uses PageModel::getAbsoluteUrl so the http-Regex can never be true?

Same as before, see https://github.com/contao/contao/blob/e5995ffa484b0dddb8acdc20eb751f0e976ddcd9/calendar-bundle/contao/classes/Calendar.php#L176-L180

faq-bundle/src/EventListener/SitemapListener.php
ausi: > PageModel::loadDetails() calls preventSaving which removes it from the registry.

No, it doesn’t.

calendar-bundle/contao/classes/Calendar.php
fritzmg: That check was actually added in https://github.com/contao/contao/commit/d822ca8da20bdf30656df20088ba53045a5b0548 (not in my original PR).

But in any case this is not related to this PR.

faq-bundle/src/EventListener/SitemapListener.php
fritzmg: @aschempp `$model->preventSaving()` and `$model->preventSaving(true)` keeps a clone in the registry. Only if you call `$model->preventSaving(false)` will the model be removed from the registry.

calendar-bundle/contao/classes/Calendar.php
aschempp: I don't understand why this is so complicated. Yes obviously the code was just moved, but it was to sanitize a method argument that _anyone_ could have called. Now noone can call it/we know it's already same, why would we still want to sanitize?

faq-bundle/src/EventListener/SitemapListener.php
aschempp: true, I didn't notice that. But it does not keep the details data! So for every page model fetched, it will re-iterate the page tree (without DB queries, because all parents are cached as well) and call the `loadDetails` method. Feels pretty wrong to me?

calendar-bundle/contao/classes/Calendar.php
```diff
@@ -410,7 +395,14 @@ protected function addEvent($objEvent, $intStart, $intEnd, $strUrl, $strBase='',
...
+				{
+					$link = $strBase . $link;
+				}
```
leofeyer, line 404: I think we can remove this, because `getAbsoluteUrl()` will always return an absolute URL now.

calendar-bundle/contao/classes/Calendar.php
leofeyer: I agree with @aschempp. Lines 399 to 405 can be deleted as the URL will never be relative.

ausi: As this is not related to this pull request and also affects Cotnao 4.9, I will create seperate pull requests for that issue.

ausi: See https://github.com/contao/contao/pull/5152

faq-bundle/src/EventListener/SitemapListener.php
ausi: See https://github.com/contao/contao/issues/5153

calendar-bundle/contao/classes/Calendar.php
aschempp: why? This PR makes the base obsolete, why would you remove the argument in a separate PR and not in this one directly?

ausi: > This PR makes the base obsolete

How does it do that? I don’t understand what you mean by that.

fritzmg: The check was likely there for a use-case where you extend from `Contao\Calendar` and then call

```
$this->addEvent(…)
```

with a custom URL. In this case the method wants to check whether the URL has scheme and host or not and prepend the base for the latter case - because it was previously ok to pass a non-absolute URL.

In this PR you are changing the `addEvent` method so that you are not able to pass a custom URL anymore and thus the `if ($strBase && !preg_match('#^https?://#', $strUrl))` can be omitted altogether there.

The question then is:

1. Do we want to keep the possibility to pass a custom URL to `addEvent`?
2. If not, should we may be make the method private and deprecate it in 4.13?

ausi: I integrated this now in this PR: 54e13992a0ae89e51ddae7b3734e440a88ce7a7f

ausi: > should we may be make the method private and deprecate it in 4.13?

I think so, yes. See https://github.com/contao/contao/pull/5155

faq-bundle/src/EventListener/SitemapListener.php
ausi: For the record: `findWithDetails()` only gets called once per FAQ category, not once per FAQ item.

ausi: Added a page cache for `ModuleFaqList` and `ModuleNewsletterList` in 96008592a54300a60a7e68dde72c51cc22ae4f38

faq-bundle/contao/modules/ModuleFaqList.php
```diff
@@ -140,18 +140,28 @@ protected function generateFaqLink($objFaq)
...
-			}
+	/**
+	 * Return the page object with loaded details for the given page ID
+	 *
+	 * @param  integer        $intPageId
+	 * @return PageModel|null
+	 */
+	private function getPageWithDetails($intPageId)
+	{
+		if (!isset(self::$arrPageCache[$intPageId]))
```
aschempp: this would return false if the page model was not found in the registry. Maybe we should use `array_key_exists` instead (which would yield true in case of `null` value afaik)?

aschempp (changes requested):
Thanks for the update! I would personally not use a static property (because of memory usage) but just a var outside the while-loop to store page cache, but both is fine for me. A respective cache seems to be missing for the Calendar class still as far as I can see?

ausi:
> A respective cache seems to be missing for the Calendar class still as far as I can see?

The Calendar class is where I copied the cache logic from, as by your request _“Da sollten wir denselben Cache einbauen wie in Calendar.php”_
https://github.com/contao/contao/blob/9f6e33c7c9041c868f10d3785b9f22cf2191add2/calendar-bundle/contao/classes/Calendar.php#L539-L547

faq-bundle/contao/modules/ModuleFaqList.php
ausi: Done in fed7a4d874d34e2f4bde601d04432c49e4b61775

aschempp (dismissed):
stupid me 😂

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5142, opened 2022-08-14
Support pid foreign keys in DC_Table

richardhj:
This change would make the tree view in the DC_Table compatible with Doctrine entities or foreign keys in general :-)

```php
#[ORM\Entity]
class ProductCategory
{
    #[ORM\OneToOne(targetEntity: ProductCategory::class)]
    #[ORM\JoinColumn(name: 'pid', referencedColumnName: 'id')]
    private ?ProductCategory $parent;
}
```

I based on the 4.13 branch because merging up from the 4.9 branch would become more tricky.


ausi (approved):

leofeyer:
Thank you @richardhj.

PR #5143, opened 2022-08-14
Do not run the DC constructor inside the picker widget

### LINKED ISSUE #3011: Permission-Check in tl_news uses wrong ID for picker when no admin user
**Affected version(s)**

Contao 4.9

**Description**

When a non-admin user uses the picker for tl_news, the tl_news::checkPermission()-Method use the wrong ID for checking permission. This problem only occurs if the source entity (where the picker field is locatet, for example tl_article) has an id, that don't exist in tl_news (for example there is an article with id 4 but no news with id 4).

Steps to reproduce (use tl_article as example):
1. Create a picker input in the dca
2. Login as a backend user without admin rights
3. Edit an article with id 2 when there is no news with id 2
4. Choose a news entry with the picker - it stay empty

The behavior is triggered by this line https://github.com/contao/contao/blob/204fb5a978e1cd49bc2819bdc34df421a1cd1296/news-bundle/src/Resources/contao/dca/tl_news.php#L548 (where the id is read from get parameter, but the get parameter is the id of the source entity, tl_article in my example) and afterwards it throws the exeption her: https://github.com/contao/contao/blob/204fb5a978e1cd49bc2819bdc34df421a1cd1296/news-bundle/src/Resources/contao/dca/tl_news.php#L604

Thread on slack: https://contao.slack.com/archives/CK4J0KNDB/p1621256386378300

ausi:
Fixes #3011 (especially the second reproduction from https://github.com/contao/contao/issues/3011#issuecomment-888945467 )

In the picker widget we instantiate a data container just to be able to pass it to callbacks in `renderLabel()`. But this executes load callbacks and permission checks that are then based on the wrong `Input` parameters. Therefore we should not execute the constructor here.

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/Picker.php
+++ b/core-bundle/src/Resources/contao/widgets/Picker.php
@@ -247,8 +247,8 @@ class Picker extends Widget
 
 			if ($objRows->numRows)
 			{
-				$dataContainer = DataContainer::getDriverForTable($strRelatedTable);
-				$dc = new $dataContainer($strRelatedTable);
+				$dc = (new \ReflectionClass(DataContainer::getDriverForTable($strRelatedTable)))->newInstanceWithoutConstructor();
+				$dc->table = $strRelatedTable;
 
 				while ($objRows->next())
 				{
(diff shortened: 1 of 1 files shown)


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5152, opened 2022-08-15
Remove base url handling from Calendar

ausi:
As discussed in https://github.com/contao/contao/pull/5141#discussion_r945230137

Not sure if this can be back ported to Contao 4.9, as the `protected` method is part of the API IMO.


fritzmg:
> Not sure if this can be back ported to Contao 4.9, as the `protected` method is part of the API IMO.

We cannot backport this  to 4.x - see https://github.com/contao/contao/pull/5141#discussion_r945846928

In 4.x we need to deprecate the respective usage, depending on what we decide on.

ausi:
Closed in favor of #5141

PR #5155, opened 2022-08-15
Deprecate Calendar::addEvent()

ausi:
See https://github.com/contao/contao/pull/5141#discussion_r945846928 and https://github.com/contao/contao/pull/5152#issuecomment-1215214386


aschempp:
Isn't this method supposed to be used by third party modules? Pretty sure a lot of calendar extensions use that 🤔

fritzmg:
Extensions likely use the `getAllEvents` hook - unless they completely override the XML generation for example.

aschempp (approved):
I now noticed this is only a helper class, I though it is part of a front end module. Then we should be fine to change that 👍

leofeyer:
Thank you @ausi.

PR #5158, opened 2022-08-16
Do not define ptable for tl_content

### LINKED ISSUE #5157: Editing News Item bring content from tl_article instead of tl_content
### Affected version(s)

4.13.7

### Description

After updaing to Contao 4.13.7, when try to edit news item content (or add new news item), it show content from "tl_article" with same pid of news item if exists.

fritzmg:
Fixes #5157

We must not define a `ptable` for `tl_content` here since we no longer define the `ptable` manually. Otherwise automatically determining the `ptable` will not work due to #5108.


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5163, opened 2022-08-16
Fix check for existing label in DcaLoader

### LINKED ISSUE #5029: Custom labels for operations in DCA not respected anymore
### Affected version(s)

\>= 4.13

### Description

It seems that since 4.13 there is a bug regarding the labels for the `global_operations` and `operations` in DCA's. 

We have in various extensions, historically, custom labels for `edit`, `copy`, `delete` etc. but since 4.13 these labels are no longer used but only the "standard labels" from the core. For example, `edit` no longer says "Edit dealer ID 1" but "Edit element ID 1" and `global_operations`, which then have a translation from `$GLOBALS['TL_LANG']` in `attributes`, don't load them anymore.

**Small example**

https://github.com/numero2/contao-storelocator/blob/69d95a2e83e4492fa759807d9ce114ab7ecef978/src/Resources/contao/dca/tl_storelocator_stores.php#L52 

The translation for this operation is missing completely while the default translation gets used in line 63.

[The disussion in Slack](https://contao.slack.com/archives/CK4J0KNDB/p1658493931807159)

ausi:
Fixes #5029

### DIFF
--- a/core-bundle/contao/library/Contao/DcaLoader.php
+++ b/core-bundle/contao/library/Contao/DcaLoader.php
@@ -159,7 +159,7 @@ class DcaLoader extends Controller
 
 			foreach ($GLOBALS['TL_DCA'][$this->strTable]['list'][$key] as $k=>&$v)
 			{
-				if (array_key_exists('label', $v))
+				if (\array_key_exists('label', $v))
 				{
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


core-bundle/contao/library/Contao/DcaLoader.php
```diff
@@ -159,7 +159,7 @@ private function addDefaultLabels()
 
 			foreach ($GLOBALS['TL_DCA'][$this->strTable]['list'][$key] as $k=>&$v)
 			{
-				if (isset($v['label']))
+				if (array_key_exists('label', $v))
```
ausi: At this point the reference to the `TL_LANG` array already exists. But if the language file was not loaded yet, the value of the reference is still `&NULL`.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #2267, opened 2020-09-09
Keep the fileTree sorting order when changing selection

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #2266

PR #5167, opened 2022-08-16
Allow usages of picker without active record

bezin:
Before #2267 it was possible to use the picker outside the context of editing a single database row. E.g. we used it in the filter panel. After that change this is no longer possible as the Ajax class now always expects an active record to be present. This change checks for the existence before trying to access it. 

I know that this might not be a supported use case, however the fix is unobtrusive.


ausi (approved):

leofeyer:
Thank you @bezin.

PR #5169, opened 2022-08-16
Fix server check in migrate command and improve message

aschempp:
`getNativeConnection` does not return a connection with server information, so this needs to be reverted to the state of Contao 4.13.

I have adjusted the error message to match the other warnings and problems. The Contao Manager will use the first line of the message as title and the remaining message as description, therefore it would be good to have them similar 🙃 

Bildschirmfoto 2022-08-16 um 15 39 11


ausi (dismissed):

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -684,10 +685,18 @@ private function validateDatabaseVersion(bool $asJson): bool
...
+            $message =
+                <<<EOF
+                    Wrong database version configured!
+                    You have version $version but the database connection is configured to $currentVersion.
+                    EOF;
+        } else {
+            $message =
+                <<<EOF
+                    Wrong database version configured!
+                    You have version $version but the database connection is configured to a different version.
```
leofeyer:
```suggestion
                    Cannot detect database version!
                    Your database version could not be detected automatically. Please check you database configuration.
```

It does not have to be the exact wording, just something that describes the situation more accurately. 😄

aschempp: huh? that message has a completely different meaning?

aschempp:
no idea why the tests still fail, they work for me locally 🤔

core-bundle/src/Command/MigrateCommand.php
ausi: I think this else branch can never be reached anyways because if `'serverVersion'` is not set, doctrine would run the same auto-detection that we run and would detect the same version anyway.

So it would probably be better to default to `""` or so for this rare edge-case.
Remove the `if (…)` completely and use something like this?

```php
$currentVersion = $this->connection->getParams()['serverVersion'] ?? '""';
```

core-bundle/tests/Command/MigrateCommandTest.php
```diff
@@ -452,7 +452,7 @@ public function testAbortsOnWrongServerVersion(string $format): void
             $this->assertSame('problem', $json['type']);
             $this->assertSame($errorMessage, $json['message']);
         } else {
-            $this->assertSame('[ERROR] '.$errorMessage, trim(preg_replace('/\s*\n\s*/', ' ', $display)));
+            $this->assertSame('[ERROR] '.$errorMessage, trim(preg_replace('/\s*\n\s*/', "\n", $display)));
```
ausi: This needs to be reverted to fix the test.
_(the command inserts line breaks depending on the with of the terminal window)_

core-bundle/src/Command/MigrateCommand.php
leofeyer: > huh? that message has a completely different meaning?

Exactly, because the original message is wrong. If `getParams()['serverVersion']` is `null`, the database has not been configured at all, therefore it is not correct to say _is configured to a different version_.

core-bundle/tests/Command/MigrateCommandTest.php
```diff
@@ -439,7 +439,7 @@ public function testAbortsOnWrongServerVersion(string $format): void
 
         $command = $this->getCommand([], [], null, null, $connection);
         $tester = new CommandTester($command);
-        $errorMessage = 'Wrong database version configured, please set it to "8.0.29", currently set to "5.7.39"';
+        $errorMessage = "Wrong database version configured!\nYou have version 8.0.29 but the database connection is configured to 5.7.39.";
```
ausi:
```suggestion
        $errorMessage = "Wrong database version configured! You have version 8.0.29 but the database connection is configured to 5.7.39.";
```
Line breaks need to be normalized in the test.

core-bundle/tests/Command/MigrateCommandTest.php
aschempp: the JSONLD tests will fail then, because they contain a line break in the message 😕

core-bundle/tests/Command/MigrateCommandTest.php
```diff
@@ -452,7 +452,7 @@ public function testAbortsOnWrongServerVersion(string $format): void
             $this->assertSame('problem', $json['type']);
             $this->assertSame($errorMessage, $json['message']);
```
ausi:
```suggestion
            $this->assertSame($errorMessage, preg_replace('/\s*\n\s*/', "\n", $json['message']));
```

core-bundle/tests/Command/MigrateCommandTest.php
ausi: See https://github.com/contao/contao/pull/5169#discussion_r946929127

core-bundle/src/Command/MigrateCommand.php
aschempp: I tried to keep the meaning of the existing message. But if the serverVersion is not configured, shouldn't we handle this as an error? Because Doctrine would retrieve the version on every page load?

ausi: I don’t think we should force users to set it. This check should only validate that there is not a wrong version configured.

aschempp: Then we should change the whole method in Contao 4.9 and upstream?

ausi: Yes, after you finished this PR and it was merged you should create a backport for 4.9.

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -685,11 +685,13 @@ private function validateDatabaseVersion(bool $asJson): bool
             return true;
         }
 
-        $message = sprintf('Wrong database version configured, please set it to "%s"', $version);
-
-        if ($currentVersion = $this->connection->getParams()['serverVersion'] ?? null) {
-            $message .= sprintf(', currently set to "%s"', $currentVersion);
-        }
+        // If serverVersion is not configured we would actually never end up here
+        $currentVersion = $this->connection->getParams()['serverVersion'] ?? '';
```
ausi, line 689:
```suggestion
        $currentVersion = $this->connection->getParams()['serverVersion'] ?? 'NULL';
```
To make the error message more readable? But totally optional, just a nitpick ☺️

ausi (dismissed):

core-bundle/src/Command/MigrateCommand.php
aschempp: since we never end up here, I don't think that makes sense

ausi (dismissed):

leofeyer (approved):

PR #5173, opened 2022-08-16
Move media above rich text

ausi:
I think the image has to be above the text to enable the classic layouts where the image floats right or left beside the text.

The `media--below` use-case should be easily achievable via CSS by using `order: 1` with flexbox or grid.
But for `float: right` or `float: left` this is not possible the other way around.


ausi:
Failing CI unrelated, see #5174

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5174, opened 2022-08-16
Fix compatibility with Symfony 6

ausi:


composer.json
```diff
@@ -154,7 +154,7 @@
         "symfony/yaml": "^5.4 || ^6.0",
         "terminal42/escargot": "^1.4.1",
         "terminal42/service-annotation-bundle": "^1.1",
-        "toflar/psr6-symfony-http-cache-store": "^3.0",
+        "toflar/psr6-symfony-http-cache-store": "^3.0 || ^4.0",
```
ausi, line 157: @Toflar is there something I need to check with this update?

Toflar: It's just because compatibility with both is not possible, see https://github.com/Toflar/psr6-symfony-http-cache-store#supported-branches

core-bundle/tests/Picker/PickerTest.php
```diff
@@ -98,7 +98,7 @@ public function testReturnsNullAsCurrentUrlIfThereIsNoCurrentMenuItem(): void
         $config = new PickerConfig('page');
         $picker = new Picker($factory, [$provider], $config);
 
-        $this->assertNull($picker->getCurrentUrl());
+        $this->assertSame('', (string) $picker->getCurrentUrl());
     }
```
ausi, line 102: I don’t understand the code behind that test. @aschempp is my fix correct here?

The change from `null` to `''` is because the router cannot return `null` anymore in Symfony 6 I think. https://github.com/symfony/symfony/blob/7eaefd2245ae9f0ac4e47cc42ce0e287c489cc0e/src/Symfony/Component/Routing/Generator/UrlGeneratorInterface.php#L79

bytehead (dismissed):
❤️

core-bundle/src/Routing/Page/PageRoute.php
```diff
@@ -130,7 +130,7 @@ public function setContent(mixed $content): self
         return $this;
     }
 
-    public function getContent()
+    public function getContent(): ?object
```
leofeyer:
```suggestion
    public function getContent(): object|null
```

Does this work, too?

leofeyer (approved):

core-bundle/tests/Picker/PickerTest.php
leofeyer: The change is correct IMHO.

leofeyer:
Thank you @ausi.

PR #5176, opened 2022-08-16
Add CSS class for number of columns to gallery template

ausi:
The gallery template was simplified on purpose AFAIR, but without the information about the number of columns (`perRow`) in the template it can’t be styled correctly I think.


ausi:
Failing CI unrelated, see https://github.com/contao/contao/pull/5174

core-bundle/contao/templates/_new/content_element/gallery.html.twig
```diff
@@ -2,6 +2,8 @@
 {% use "@Contao/component/_list.html.twig" %}
 {% use "@Contao/component/_figure.html.twig" %}
 
+{% set attributes = attrs(attributes|default).addClass("cols-#{items_per_row|default(4)}") %}
```
leofeyer: I‘m not sure about the class name because `cols-*` is also used by Bootstrap (and probably others). Should we pick a more unique name?

ausi: `per-row-4` would match the `perRow` DB column name?
Or `content-gallery--cols-4` to be based on the class name of the element itself?

leofeyer: I‘m fine with both, so unless one of the @contao/developers has a preference, you can choose either one.

ausi: I’d go with `content-gallery--cols-4` then.

```suggestion
{% set attributes = attrs(attributes|default).addClass("content-#{type}--cols-#{items_per_row|default(4)}") %}
```

fritzmg: I'm in favor of 💩, I mean BEM, although I have no strong opinion here (I am just used to using it now). However - may be we should go over all the classes in our current templates again to be mroe consistent with the usage of BEM? For example in the `text` template we just use the class `rte` for the wrapper containing the text. If we follow BEM it would be `content-text__rte`, etc.

ausi: > For example in the `text` template we just use the class `rte` for the wrapper containing the text. If we follow BEM it would be `content-text__rte`, etc.

I’m not sure, the `rte` element can be used in other elements too so I would see it as it’s own block.

core-bundle/tests/Controller/ContentElement/ImagesControllerTest.php
```diff
@@ -72,7 +72,7 @@ public function testOutputsGallery(): void
         );
 
         $expectedOutput = <<<'HTML'
-            <div class="content-gallery">
+            <div class="cols-2 content-gallery">
```
ausi:
```suggestion
            <div class="content-gallery--cols-2 content-gallery">
```

core-bundle/contao/templates/_new/content_element/gallery.html.twig
fritzmg: I am assuming the `rte` class will also be used in the news templates at some point though. Currently they use `ce_text` for the teaser.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5179, opened 2022-08-17
Fix download element back end layout

ausi:
The checkbox needs an `m12` because it’s next to the image size widget.


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5180, opened 2022-08-17
Fix configuration of TemplateOptionsListener

ausi:
This reverts #4993

There are two different services, one needs to be tagged for `tl_module` and one for `tl_content`.
Currently the `contao.listener.element_template_options` is tagged with both callbacks and therefore `ce_` instead of `mod_` templates are shown for the custom templates in `tl_module`.


fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #5189, opened 2022-08-17
Fix compatibility with Symfony 6 session

### LINKED ISSUE #5187: Contao 5.0 does not work with Symfony 6
### Affected version(s)

5.0.0

### Description

The `session` service is no longer available in Symfony 6, therefore the following is broken:

1. The `AddSessionBagsPass` which does not register the Contao sessions anymore.
2. Every `System::getContainer()->get('session')`call.

No. 2 can be solved by replacing `System::getContainer()->get('session')` with `System::getContainer()->get('request_stack')->getSession()`, however this will only work if the Contao sessions are registered beforehand:

ausi:
Fixes #5187

Based on #5188 (so that has to be merged first)

I replaced the `AddSessionBagsPass` with a decorated `SessionFactory` in da0c38456b74c3b099f989363651b59c9375edae


Toflar (approved):

bytehead (approved):

leofeyer:
Thank you @ausi.

PR #5190, opened 2022-08-18
Session cleanup

leofeyer:


core-bundle/contao/forms/Form.php
```diff
@@ -325,10 +325,7 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields, $arrFi
 
 		// Store submitted data (possibly modified by hook or data added) in the session for 10 seconds,
 		// so it can be used on any forward page using the {{form_session_data::<form-field-name>}} insert tag
-		if ($request = System::getContainer()->get('request_stack')->getCurrentRequest())
-		{
-			$request->getSession()->set(self::SESSION_KEY, new AutoExpiringAttribute(10, $arrSubmitted));
-		}
+		System::getContainer()->get('request_stack')->getSession()->set(self::SESSION_KEY, new AutoExpiringAttribute(10, $arrSubmitted));
```
ausi, line 328: Would this now fail if there is no request in the stack?

leofeyer: Yes, but we use the same code in multiple other places as well. Sometimes we check if the request stack contained a request and sometimes we don‘t. Pretty inconsistent. 🙈 

In the `Form` class, there will always be a request, therefore we don‘t have to check. But I can revert the change if you feel better then.

fritzmg: > In the Form class, there will always be a request, therefore we don‘t have to check.

Not if the Form is inserted as a content element within an event and you have a RSS feed for events with detailed content. But that's the case for any content element or module, not just for `Form`.

// edit, but I think we already fixed that by pushing a faux request and session, so nvm ;)

leofeyer: At this very place in the code, there will always be a request because `processFormData()` will only be called if there is POST data.

ausi (approved):

core-bundle/contao/forms/Form.php
ausi: Agree 👍
Was just curious because of the check for the request in the current version.

Issue #5006, opened 2022-07-19
Inconsistent request handling


leofeyer:
### Description

Currently we mix `$requestStack->getCurrentRequest()` and `$requestStack->getMainRequest()` without a clear concept of when to use which. The difference between the two is that `getMainRequest()` returns the **first** request of the stack and `getCurrentRequest()` returns the **last** request.

As already mentioned in https://github.com/contao/contao/pull/4984#issuecomment-1186981655, `$requestStack->getSession()` also uses the **last** request, but we have used the session of the main request so far.

aschempp:
every subrequest (in almost any case) inherits the session from the main request, so I think this is not relevant. But I never really understood when Symfony uses which request. Main request holds the actual browser request, so I would always think that's the "better" one for "global" stuff 🤷

ausi:
As discussed in the Contao Call today this would be our propsal:

1. Prefer to use use `getCurrentRequest()`
2. If this does not work because of technical reasons (like firewall related stuff) use `getMainRequest()` and add a code comment that describes why this is the case.

_(leaving this as up for discussion to rediscuss with all core developers in the next call)_

leofeyer:
I have fixed the obviously wrong calls in #5192. This leaves us with the following classes still using `getMainRequest()`:

* `ContaoContext`
* `ContaoCsrfTokenManager`
* `ContaoFramework`
* `CoreResponseContextFactory`
* `AccessDecisionManager`
* `FrontendPreviewAuthenticator`
* `TokenChecker`

leofeyer:
As discussed in the Contao call, we want to adjust #5192 and replace all `getMainRequest()` occurrences with `getCurrentRequest()`, except in the `AccessDecisionManager` and the `FrontendPreviewAuthenticator` because they are firewall-related.

BTW, the `AccessDecisionManager` should be rewritten to check the firewall instead of the scope matcher.

leofeyer:
The PR has been updated.

PR #5192, opened 2022-08-18
Use getCurrentRequest() instead of getMainRequest()

leofeyer:
This PR fixes the obviously wrong `getMainRequest()` calls as discussed in #5006. It does not touch the other calls, which are subject to further discussion.

### DIFF
--- a/core-bundle/src/Controller/ContentElement/HyperlinkController.php
+++ b/core-bundle/src/Controller/ContentElement/HyperlinkController.php
@@ -39,7 +39,7 @@ class HyperlinkController extends AbstractContentElementController
         $href = $this->insertTagParser->replaceInline($model->url);
 
         if (Validator::isRelativeUrl($href)) {
-            $href = $this->requestStack->getMainRequest()?->getBasePath().'/'.$href;
+            $href = $this->requestStack->getCurrentRequest()?->getBasePath().'/'.$href;
         }
 
         $linkAttributes = (new HtmlAttributes())
--- a/core-bundle/src/Controller/ContentElement/HyperlinkController.php
+++ b/core-bundle/src/Controller/ContentElement/HyperlinkController.php
@@ -20,17 +20,13 @@ use Contao\CoreBundle\String\HtmlAttributes;
 use Contao\CoreBundle\Twig\FragmentTemplate;
 use Contao\Validator;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 
 #[AsContentElement(category: 'links')]
 class HyperlinkController extends AbstractContentElementController
 {
-    public function __construct(
-        private readonly Studio $studio,
-        private readonly InsertTagParser $insertTagParser,
-        private readonly RequestStack $requestStack,
-    ) {
+    public function __construct(private readonly Studio $studio, private readonly InsertTagParser $insertTagParser)
+    {
     }
 
     protected function getResponse(FragmentTemplate $template, ContentModel $model, Request $request): Response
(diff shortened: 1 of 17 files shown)


core-bundle/src/EventListener/DbafsMetadataSubscriber.php
```diff
@@ -119,7 +119,7 @@ private function getDefaultLocales(): array
             return $this->defaultLocales;
         }
 
-        $page = $this->requestStack->getMainRequest()?->attributes->get('pageModel');
+        $page = $this->requestStack->getCurrentRequest()?->attributes->get('pageModel');
```
aschempp, line 122: This might return an ID in a subrequest, is that intended?

core-bundle/src/Controller/ContentElement/HyperlinkController.php
```diff
@@ -39,7 +39,7 @@ protected function getResponse(FragmentTemplate $template, ContentModel $model,
         $href = $this->insertTagParser->replaceInline($model->url);
 
         if (Validator::isRelativeUrl($href)) {
-            $href = $this->requestStack->getMainRequest()?->getBasePath().'/'.$href;
+            $href = $this->requestStack->getCurrentRequest()?->getBasePath().'/'.$href;
```
ausi: Couldn’t we use `$request` here as this would be the current request?

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
ausi: Wouldn’t this be converted by the `PageAccessListener` so that it is always a model and not an ID?

leofeyer: So only the main request hast the model? Then we should probably use `getMainRequest()` here as well.

core-bundle/src/Controller/ContentElement/HyperlinkController.php
leofeyer: Of course. How could I miss this. 😄

leofeyer: Changed in b7aa4467b64dc59776da5c0949b5bc8039fe70ee.

core-bundle/src/EventListener/DbafsMetadataSubscriber.php
leofeyer: However, we are using `->attributes->get('pageModel')` on the current request a lot, too:

https://github.com/contao/contao/blob/c10cecec52ed9a1be6c4cf45e8dec6830514f93a/core-bundle/src/Controller/AbstractFragmentController.php#L60-L64

Are you sure it has to be the main request?

aschempp: @ausi is right, we have a listener that converts it (in Contao 5) 😎

bytehead (dismissed):

aschempp (dismissed):

leofeyer:
PR updated to what has been discussed in the Contao call (see https://github.com/contao/contao/issues/5006#issuecomment-1227265346).

core-bundle/src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -47,11 +48,12 @@ public function authenticateFrontendUser(string $username, bool $showUnpublished
 
         $token = new UsernamePasswordToken($user, 'contao_frontend');
 
-        if ((!$request = $this->requestStack->getMainRequest()) || !$request->hasSession()) {
+        try {
+            $session = $this->requestStack->getSession();
```
bytehead, line 52: This will return the session from the `currentRequest`: https://github.com/symfony/symfony/blob/95004c2b1ff6a5a84c49be6eae1131a43694e520/src/Symfony/Component/HttpFoundation/RequestStack.php#L122

leofeyer: Yes, as discussed, Symfony passes the session to every sub-request, therefore it does not matter.

bytehead: Ah, ok 👍

bytehead (approved):

ausi (approved):

PR #5193, opened 2022-08-18
Clear the DEPRECATED.md file and remove some leftovers

leofeyer:


core-bundle/contao/library/Contao/Controller.php
```diff
@@ -1439,20 +1439,6 @@ public static function addEnclosuresToTemplate($objTemplate, $arrItem, $strKey='
 		$objTemplate->enclosure = $arrEnclosures;
 	}
 
-	/**
-	 * Set the static URL constants
-	 */
-	public static function setStaticUrls()
```
leofeyer, line 1445: Removing the constants is not a problem because they have been deprecated every since Contao 4.0:

https://github.com/contao/contao/blob/f28fa5e71d37959b027820ad266a832fa28e6c19/DEPRECATED.md?plain=1#L63-L78

However, removing the static method is a BC break as the method is not deprecated. @contao/developers Would you prefer to keep the method as a no-op and flag it for removal in Contao 6?

bytehead (dismissed):

core-bundle/contao/library/Contao/Controller.php
bytehead: We're allowed to break, no? Or deprecate it in 4.13 now?

Toflar: +1, add a deprecation for the next 4.13 release and then remove 👍

DEPRECATED.md
```diff
@@ -4,198 +4,3 @@
...
-```php
-// Old syntax
-echo TL_ASSETS_URL;
-echo TL_FILES_URL;
-
-// New syntax
-$container = System::getContainer();
-echo $container->get('contao.assets.assets_context')->getStaticUrl();
-echo $container->get('contao.assets.files_context')->getStaticUrl();
-```
```
ausi, line 31: This needs to be moved to the *UPGRADE.md* file then I think.

core-bundle/contao/library/Contao/Controller.php
ausi: Agree, remove in 5.0 and add a deprecation in 4.13.

DEPRECATED.md
leofeyer: Done in 348b908366de0479df75bb888e10893566bf831d.

core-bundle/contao/library/Contao/Controller.php
leofeyer: Deprecation added in #5194.

bytehead (approved):

ausi (approved):

PR #5195, opened 2022-08-18
Fix back end preview redirect

ausi:
Related #4326

### DIFF
--- a/core-bundle/contao/models/PageModel.php
+++ b/core-bundle/contao/models/PageModel.php
@@ -1165,7 +1165,7 @@ class PageModel extends Model
 		$baseUrl = $context->getBaseUrl();
 
 		// Add the preview script
-		$context->setBaseUrl(\dirname($baseUrl) . $previewScript);
+		$context->setBaseUrl(rtrim(\dirname($baseUrl), '/') . $previewScript);
 
 		$objRouter = System::getContainer()->get('router');
 


core-bundle/contao/models/PageModel.php
```diff
@@ -1165,7 +1165,7 @@ public function getPreviewUrl($strParams=null)
 		$baseUrl = $context->getBaseUrl();
 
 		// Add the preview script
-		$context->setBaseUrl(\dirname($baseUrl) . $previewScript);
+		$context->setBaseUrl(rtrim(\dirname($baseUrl), '/') . $previewScript);
```
ausi, line 1168: This prevents URLs like `//preview.php`

leofeyer: <del>This can never happen, can it? https://3v4l.org/VI2Sh</del>

NVM, it can happen if `$baseUrl` is `/`. https://3v4l.org/JS7LG

leofeyer (approved):

core-bundle/contao/models/PageModel.php
ausi: Yes it is for the case of $baseUrl being `/preview.php`.
This causes the issue that the pages preview URLs all redirect to `//preview.php` currently.

leofeyer:
Thank you @ausi.

PR #5198, opened 2022-08-18
Fix skipping backups with schema diff

ausi:
Currently the backups are skipped even though pending schema changes.

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -114,7 +114,7 @@ class MigrateCommand extends Command
         $asJson = 'ndjson' === $input->getOption('format');
 
         // Return early if there is no work to be done
-        if (!$this->migrations->hasPending()) {
+        if (!$this->hasWorkToDo()) {
             if (!$asJson) {
                 $this->io->info('Database dump skipped because there are no migrations to execute.');
             }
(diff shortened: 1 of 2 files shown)


Toflar:
Shouldn't that go to 4.13?

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -114,7 +114,7 @@ private function backup(InputInterface $input): bool
         $asJson = 'ndjson' === $input->getOption('format');
 
         // Return early if there is no work to be done
-        if (!$this->migrations->hasPending()) {
+        if (!$this->hasWorkToDo()) {
             if (!$asJson) {
                 $this->io->info('Database dump skipped because there are no migrations to execute.');
```
bytehead, line 119:
```suggestion
                $this->io->info('Database dump skipped because there are no migrations or SQL queries to execute.');
```

leofeyer (approved):

leofeyer:
> Shouldn't that go to 4.13?

No, Contao 4.13 is fine.

leofeyer:
Thank you @ausi.

fritzmg:
> > Shouldn't that go to 4.13?
> 
> No, Contao 4.13 is fine.

This is a little confusing 😁. And I agree - doesn't this need to be changed in 4.13 as well?

ausi:
Only 5.0 is affected because the error was introduced when merging upstream.

PR #5199, opened 2022-08-18
Fix Input compatibility with Symfony 6

ausi:
Fixes `Input value "crawl_subscriber_names" contains a non-scalar value.` when trying to rebuild the search index.


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5200, opened 2022-08-18
Fix the limit-height.js script

leofeyer:
Always resize all elements but only add a toggle to those elements where the content height exceeds the element height.


core-bundle/public/limit-height.js
```diff
@@ -18,21 +18,18 @@ window.addEventListener('DOMContentLoaded', function () {
         // Return if there is no height value
         if (!hgt) return;
 
-        const toggler = document.createElement('div');
-        toggler.classList.add('limit_toggler');
+        const height = div.offsetHeight;
+
+        // Resize the element
+        div.style.height = hgt+'px';
```
aschempp, line 24: should this be skipped as well if the element is high enough?

leofeyer: No, if an element has `.h40`, it shall be 40px high. And only if its content is higher, a toggle shall be added additionally.

leofeyer: ### Before



### After

aschempp: Strange, but I see that was already the case currently.

aschempp (approved):

PR #5205, opened 2022-08-19
Drop the deprecated article-to-PDF functionality

aschempp:
Looks like we forgot to drop this since `contao/tcpdf-bundle` is not compatible with Contao 5.


Toflar (approved):
Should go to the UPGRADE.md too, no?

ausi:
> Should go to the UPGRADE.md too, no?

Yes, and the docs need to be updatet too https://docs.contao.org/dev/reference/hooks/printArticleAsPdf

ausi (approved):

PR #5208, opened 2022-08-22
Fix custom paste modes not working properly

qzminski:
In one of our projects, we allow backend users to have some custom paste logic. What I mean by that are custom global operation buttons that trigger the "select new record position" backend view but have a custom `mode` parameter, which then is used to apply custom logic (e.g., paste several custom content elements instead of one):

`act=paste&mode=multiple_content_elements`

Since #4819 got merged, this is no longer possible, as records are not rendered due to missing `CURRENT_ID`:

image

This pull request should fix that. I am aware that this is an extreme edge case, but this used to work, and I think we should not limit this flexibility. Let me know what you think about it.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Backend.php
+++ b/core-bundle/src/Resources/contao/classes/Backend.php
@@ -667,7 +667,7 @@ abstract class Backend extends Controller
 		$mode = Input::get('mode');
 
 		// For these actions the id parameter refers to the parent record
-		if (($act === 'paste' && $mode === 'create') || \in_array($act, array(null, 'select', 'editAll', 'overrideAll', 'deleteAll'), true))
+		if (($act === 'paste' && $mode !== 'copy' && $mode !== 'cut') || \in_array($act, array(null, 'select', 'editAll', 'overrideAll', 'deleteAll'), true))
 		{
 			return $id;
 		}


ausi:
I think we should be very careful with these changes.
Are there no cases where `act=paste` is used together with `pid=`?

core-bundle/src/Resources/contao/classes/Backend.php
```diff
@@ -692,7 +692,7 @@ protected function findCurrentId(?string $table): ?string
 		$mode = Input::get('mode');
 
 		// For these actions the id parameter refers to the parent record
-		if (($act === 'paste' && $mode === 'create') || \in_array($act, array(null, 'select', 'editAll', 'overrideAll', 'deleteAll'), true))
+		if ($act === 'paste' || \in_array($act, array(null, 'select', 'editAll', 'overrideAll', 'deleteAll'), true))
```
ausi:
```suggestion
		if (\in_array($act, array(null, 'paste', 'select', 'editAll', 'overrideAll', 'deleteAll'), true))
```

If we decide to merge this PR, this should go into the in_array check I think.

qzminski:
I checked the `DC_Table` and it seems that `pid` is never used with `act=paste`, but I am only 99% sure. `pid` seems to be used with `DC_Table` methods (`create()`, `copy()`, etc.) but never together with `paste`.

ausi:
Two examples: 

- `?do=article&id=86&table=tl_content&act=paste&mode=create` in this case, the `id` parameter refers to the parent record from `tl_article` this is the ID we want to get back from `findCurrentId` 
- `?do=article&id=70&table=tl_content&act=paste&mode=copy` in this case, the `id` parameter refers to the row from `tl_content` so the method has to fetch `tl_content.pid` for `id=70` in this case

The second case would break now with your pull request.

There seem to be four modes for `act=paste`:

1. `create`, id refers to the parent
2. `copy`, id refers to the element itself
3. `cut`, id refers to the element itself
4. `move`, no id is used there AFAIK

So we could probably check for `($act === 'paste' && $mode !== 'copy' && $mode !== 'cut')` to support your case. WDYT?

qzminski:
Hmm, good point. I incorrectly assumed that we are looking for a combination of `act=paste` together with `pid` URL parameters, but in fact the `id` parameter is important here. I think your proposal is totally correct, and I pushed the changes.

ausi:
This needs to be rebased on Contao 4.9 as the method exists there too I think.

qzminski:
Rebased on 4.9 :+1:

ausi (approved):

leofeyer:
Thank you @qzminski.

PR #5212, opened 2022-08-22
Write dotenv parameters to .env.local

### LINKED ISSUE #5209: DATABASE_URL in .env.local instead of .env
In https://github.com/contao/contao/pull/5066#pullrequestreview-1055496311 @fritzmg wrote:

> As pointed out in Slack, `.env` should never be populated with any credentials, only with default values for the application. See [symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local](https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local)

But the manager currently store the `DATABASE_URL` in .env instead of .env.local should we change this here too? https://github.com/contao/contao/blob/e7ee456282da3f0b821001ac2c571fa793dd32b7/manager-bundle/src/ContaoManager/ApiCommand/SetDotEnvCommand.php#L49

ausi:
Fixes #5209


ausi:
As discussed in https://github.com/contao/contao/issues/5209#issuecomment-1223645818 we should probably add a `--local` option to the env commands to be backwards compatible.

manager-bundle/src/ContaoManager/ApiCommand/GetDotEnvCommand.php
```diff
@@ -50,12 +50,20 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $path = $this->projectDir.'/.env';
+        $pathLocal = $path.'.local';
 
-        if (!file_exists($path)) {
+        if (!file_exists($path) && !file_exists($pathLocal)) {
```
leofeyer: I think this change can be reverted because Symfony does not load the `.env.local` file if there is no `.env` file and so should we, too, don‘t you think?

ausi:
```suggestion
        if (!file_exists($path)) {
```

fritzmg (commented):
The `--local` option has not been implemented yet - or has something different been discussed in the last call? Also shouldn't the `SetDotEnvCommand` also optionally be able to touch the `.env` if not existent? Otherwise the Contao Manager would have to do that.

ausi:
As briefly discussed in the last call we may not need the `--local` parameter if we agree that we do not need backwards compatibility for the api commands.

This only needs to work with the Contao Manager in a backwards compatible way I think.

ausi:
As discussed in the call today, we want the command to write in the `.env.local` directly and we do not want the `--local` parameter.

ausi:
This is RTM I think ☺️

manager-bundle/src/ContaoManager/ApiCommand/RemoveDotEnvCommand.php
```diff
@@ -49,14 +50,18 @@ protected function configure(): void
...
     {
+        $key = $input->getArgument('key');
         $file = $this->projectDir.'/.env';
+        $fileLocal = $file.'.local';
 
-        if (!file_exists($file)) {
-            return 0;
+        $dotenv = new DotenvDumper($fileLocal);
+
+        if (file_exists($file) && isset((new Dotenv(false))->parse(file_get_contents($file))[$key])) {
```
leofeyer: Shouldn‘t this be `$fileLocal` here?

ausi: No, if the `.env` file exists and the `$key` is set in that file, we have to set the `$key` to the empty value in `.env.local` in order to “remove” that value.

If `$key` does not exist in `.env` we can simply remove it from `.env.local`.

ausi: See `RemoveDotEnvCommandTest::testSetsToEmptyStringIfNonLocalFileExists()`

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5218, opened 2022-08-24
Order listView by joined table values in label fields

dennisbohn:
At the moment label fields in listView are only sorted by id, when they reference to another table.

```php
$GLOBALS['TL_DCA']['tl_gs_document'] = [
    'list' => [
        'sorting' => [
            'mode' => DataContainer::MODE_SORTABLE,
            'flag' => DataContainer::SORT_ASC,
            'panelLayout' => 'filter;sort,search,limit',
            'fields' => ['type']
        ],
        'label' => [
            'fields' => ["name", "type:tl_gs_documenttype.name", "date"],
            'showColumns' => true
        ]
    ]
]
```

image

The additional code will join the referenced table in ORDER BY.

before

```
SELECT * FROM tl_gs_document ORDER BY type, id
```

after

```
SELECT * FROM tl_gs_document ORDER BY (SELECT name FROM tl_gs_documenttype WHERE tl_gs_documenttype.id = tl_gs_document.type), id
```

It also works with CONCAT.

```php
'label' => [
    'fields' => ["name", "type:tl_gs_documenttype.CONCAT(id, ' ', name)", "date"],
    'showColumns' => true
]
```

```
SELECT * FROM tl_gs_document ORDER BY (SELECT CONCAT(id, ' ', name) FROM tl_gs_documenttype WHERE tl_gs_documenttype.id = tl_gs_document.type), id
```

And also with a descending order.

```
SELECT * FROM tl_gs_document ORDER BY (SELECT CONCAT(id, ' ', name) FROM tl_gs_documenttype WHERE tl_gs_documenttype.id = tl_gs_document.type) DESC, id
```


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -4827,6 +4827,23 @@ protected function listView()
...
+
+				// If the label field contains a reference to another table, join the table for correct sorting
+				foreach ($GLOBALS['TL_DCA'][$this->strTable]['list']['label']['fields'] as $field)
+				{
+					$arrField = explode(':', $field, 2);
+
+					if ($arrField[0] === $key && \count($arrField) === 2)
+					{
+						$fieldRef = explode('.', $arrField[1], 2);
+						$orderBy[$k] = sprintf('(SELECT %s FROM %s WHERE %s.id = %s.%s)', $fieldRef[1], $fieldRef[0], $fieldRef[0], $this->strTable, $key);
```
```suggestion
						$orderBy[$k] = sprintf('(SELECT %s FROM %s WHERE %s.id = %s.%s)', Database::quoteIdentifier($fieldRef[1]), $fieldRef[0], $fieldRef[0], $this->strTable, $key);
```

As we do this in the DataContainer that uses these fields too: https://github.com/contao/contao/blob/b1455051078bee153bcfdd7af1b26651c42e6e78/core-bundle/src/Resources/contao/classes/DataContainer.php#L1679-L1683

We might also want change the coding style to be similar to the one from the DataContainer.

ausi (changes requested):

dennisbohn:
Closed for a new pull request for a more elegang way of  foreignKey based sorting.

Opened a new pull request where the sorting isnt based on a foreignKey in the list section. Instead it uses the foreignKey in the field.

https://github.com/contao/contao/pull/5917

PR #5222, opened 2022-08-25
Use the firewall name instead of the scope to determine the access strategy

aschempp:
As discussed regarding https://github.com/contao/contao/pull/5192 we should base the firewall access decision strategy based on the firewall and not the request context.

### DIFF
--- a/core-bundle/src/Security/Authentication/AccessDecisionManager.php
+++ b/core-bundle/src/Security/Authentication/AccessDecisionManager.php
@@ -12,8 +12,7 @@ declare(strict_types=1);
 
 namespace Contao\CoreBundle\Security\Authentication;
 
-use Contao\CoreBundle\Routing\ScopeMatcher;
-use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Bundle\SecurityBundle\Security\FirewallContext;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authorization\AccessDecisionManagerInterface;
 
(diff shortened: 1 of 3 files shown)


core-bundle/tests/Security/Authentication/AccessDecisionManagerTest.php
```diff
@@ -94,4 +69,16 @@ private function mockAccessDecisionManager(bool $shouldBeCalled): AccessDecision
 
         return $manager;
     }
+
+    private function mockFirewallContext(string $name)
```
ausi:
```suggestion
    /**
     * @return FirewallContext&MockObject
     */
    private function mockFirewallContext(string $name): FirewallContext
```

core-bundle/src/Security/Authentication/AccessDecisionManager.php
```diff
@@ -12,34 +12,32 @@
...
 class AccessDecisionManager implements AccessDecisionManagerInterface
 {
     private AccessDecisionManagerInterface $inner;
     private AccessDecisionManagerInterface $contaoAccessDecisionManager;
-    private ScopeMatcher $scopeMatcher;
-    private RequestStack $requestStack;
+    private FirewallContext $firewallContext;
 
-    public function __construct(AccessDecisionManagerInterface $inner, AccessDecisionManagerInterface $contaoAccessDecisionManager, ScopeMatcher $scopeMatcher, RequestStack $requestStack)
+    public function __construct(AccessDecisionManagerInterface $inner, AccessDecisionManagerInterface $contaoAccessDecisionManager, FirewallContext $firewallContext)
```
ausi: Because the constructor parameters changed, the services.yaml needs to be updated too: https://github.com/contao/contao/blob/001cea17fe99c1d2a44206c57333f49dfb594bae/core-bundle/src/Resources/config/services.yml#L710-L717

bytehead: Isn't this a BC break?

ausi: ~~Why? The constructor parameters of our services are not part of our BC promise AFAIK.~~

ausi: You are right, according to our BC promise the constructor needs to be marked as `@internal` to allow that change: https://docs.contao.org/dev/internals/bc-promise/

leofeyer: Added in 211066b26bfc4a0fb919feb754952f6d93768abc.

leofeyer:
@aschempp Please fix the functional tests.

ausi (dismissed):

aschempp:
Code updated, it now uses the same code as the token checker (https://github.com/contao/contao/blob/5.x/core-bundle/src/Security/Authentication/Token/TokenChecker.php#L149-L164)

bytehead (dismissed):

leofeyer (approved):

PR #5231, opened 2022-08-30
Do not use POST values directly for Reply-To and Cc

fritzmg:
Forms from the form generator can automatically set a `Reply-To` address, if the form contains an `email` field. It will also automatically pull values from a `name` or `firstname` and `lastname` field for this `Reply-To` address.

However the current implementation allows this even when _no_ `email` field is present in the form, as it directly checks `Input::post()`  rather than the actually submitted values.

This PR fixes that by using `$arrSubmitted` instead.

_Note:_ this PR also contains another change/bugfix - it adds a `!empty($arrSubmitted['email'])` check for the automatic `Cc`. Within the community it often happens that users will add a `cc` field to their form, without having an `email` field - or the `email` field not being mandatory. This then leads to an error if no `email` is provided.

@leofeyer should I create separate PRs for 5.0 and 4.13 in this case?


core-bundle/src/Resources/contao/forms/Form.php
```diff
@@ -405,18 +405,18 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields)
 			$email->fromName = $GLOBALS['TL_ADMIN_NAME'];
 
 			// Get the "reply to" address
-			if (!empty(Input::post('email', true)))
+			if (!empty($arrSubmitted['email']))
 			{
-				$replyTo = Input::post('email', true);
+				$replyTo = $arrSubmitted['email'];
```
ausi, line 410: Do we need to decode entities here? As the second parameter of `Input::post()` enabled entity decoding.

fritzmg: True. We could still use `Input::post('email', true)` etc. here, as long as we use `$arrSubmitted` for the `empty()` checks I suppose.

fritzmg: I opted for `StringUtil::decodeEntities` instead.

ausi (dismissed):

core-bundle/src/Resources/contao/forms/Form.php
```diff
@@ -405,18 +405,18 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields)
 			$email->fromName = $GLOBALS['TL_ADMIN_NAME'];
 
 			// Get the "reply to" address
-			if (!empty(Input::post('email', true)))
+			if (!empty($arrSubmitted['email']))
 			{
-				$replyTo = Input::post('email', true);
+				$replyTo = StringUtil::decodeEntities($arrSubmitted['email']);
```
leofeyer: This should not be necessary because `decodeEntities` is enabled for all widgets by default:

https://github.com/contao/contao/blob/3281624a9095b467a85fe8a012d601a9df24d009/core-bundle/src/Resources/contao/forms/Form.php#L166

fritzmg: Indeed (tested it as well). Last commit reverted.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5234, opened 2022-08-31
Correctly list the templates in "override all" mode

### LINKED ISSUE #5131: CustomTpl field in tl_form_field acts strange in act=overrideAll
### Affected version(s)

4.13

### Description

Steps to reproduce in Contao demo

- go to `Forms` -> `Contact`
- `edit multiple` and select the fields `name` and `email`
- click `override`
- select the field `customTpl` and `continue`

Then you can select the template `form_upload` but the both fields have the type `text`. So this is weird. But on `act=edit` or `act=editAll` we can select the template `form_textfield (Default)`.

Even stranger if you have a custom template like `form_textfield_custom`. The selects have following options:
- on `act=edit` and `act=editAll`
  -  `form_textfield (Default)`
  -  `form_textfield_custom`
- on `act=overrideAll`
  -  `form_textfield_custom`

So the default option is missing and we cannot reset to the default while on `act=overrideAll`

leofeyer:
Fixes #5131

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Controller.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Controller.php
@@ -175,6 +175,14 @@ abstract class Controller extends System
 			$strGlobPrefix = substr($strGlobPrefix, 0, -1) . '[_-]';
 		}
 
+		// Add the matching bundle templates, e.g. in "override all" mode (see #5131)
+		$arrBundleMatches = array_unique(preg_grep('/' . $strGlobPrefix . '/', $arrBundleTemplates));
+
+		foreach ($arrBundleMatches as $strTemplate)
+		{
+			$arrTemplates[$strTemplate][] = 'root';
+		}
+
 		$projectDir = System::getContainer()->getParameter('kernel.project_dir');
 		$arrCustomized = self::braceGlob($projectDir . '/templates/' . $strGlobPrefix . '*.html5');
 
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/library/Contao/Controller.php
```diff
@@ -175,6 +175,14 @@ public static function getTemplateGroup($strPrefix, array $arrAdditionalMapper=a
 			$strGlobPrefix = substr($strGlobPrefix, 0, -1) . '[_-]';
 		}
 
+		// Add the matching bundle templates, e.g. in "override all" mode (see #5131)
+		$arrBundleMatches = array_unique(preg_grep('/' . $strGlobPrefix . '/', $arrBundleTemplates));
```
leofeyer: @ausi Not sure if there is a better way to match the template names against the glob pattern? (Or to convert the glob pattern into a regular expression?)

ausi:
Wouldn’t the correct solution be to have one blank option (maybe named something like “Default Template”) and the two `*_custom` options?

leofeyer:
Probably yes. In "edit all" mode, the first option is `form_textfield (Default)`, so providing a `Default` option with an empty value in "override all" mode seems logical.

leofeyer:
Changed in 4b539d6bcf5cb0019a9cf2536091d90b5a927d62. The selection now behaves exactly like the `TemplateOptionsListener`, which handles custom templates for content elements and modules.



In Contao 5, we can use the `TemplateOptionsListener` for form fields, too, because we no longer have the BC layer:

https://github.com/contao/contao/blob/1051a3608b8fe221869395153f591782e317517a/core-bundle/src/Resources/contao/dca/tl_form_field.php#L775-L779

I will provide a separate PR for that.

ausi (approved):

Issue #4818, opened 2022-06-10
Inconsistent deprecations


leofeyer:
While working on #4816, I have noticed that the deprecations are inconsistent:

**We have deprecated**

* `BackendUser::isAllowed()`
* `BackendUser::canEditFieldsOf()`
* `FrontendUser::isMemberOf()`

**But we have not deprecated**

* `BackendUser::hasAccess()`
* `User::isMemberOf()`

@contao/developers Why is that?

_Originally posted by @leofeyer in https://github.com/contao/contao/issues/4816#issuecomment-1152109114_

aschempp:
As far as I remember that's because `hasAccess` and `isMemberOf` are still used by the `BackendAccessVoter`, but the other methods have been re-implemented.

leofeyer:
So should we move the two methods to the `BackendAccessVoter` class then?

Toflar:
If we move them, we have to also deprecate them in 4.13. Just as a reminder :)

aschempp:
- `FrontendUser::isMemberOf` is deprecated since 4.12 because we have a voter for frontend groups. It is still used directly by `Backend::findSearchablePages` which is deprecated though. Look like we can deprecate it in 4.13 and remove in 5.0
- `BackendUser::hasAccess` contains some special handling for `filemounts`, and I remember there was a discussion about `pagemounts` as well. I think we should improve that in Contao 5.x and remove the method then (simply because it has no priority for 5.0)

leofeyer:
As discussed in the Contao all, we can deprecate the `FrontendUser::isMemberOf()` method for removal in Contao 6 because we do not use it anymore in Contao 5.

The `BackendUser::hasAccess()` method has to be reworked before we can deprecate it as described by @aschempp above.

aschempp:
see https://github.com/contao/contao/pull/6008

> The `BackendUser::hasAccess()` method has to be reworked before we can deprecate it as described by @aschempp above.

this has been done in https://github.com/contao/contao/pull/5893

PR #5236, opened 2022-09-01
Deprecate the User::isMemberOf() method

leofeyer:
Fixes the first part of #4818.

### DIFF
--- a/core-bundle/contao/library/Contao/User.php
+++ b/core-bundle/contao/library/Contao/User.php
@@ -296,9 +296,13 @@ abstract class User extends System implements UserInterface, EquatableInterface,
 	 * @param mixed $ids A single group ID or an array of group IDs
 	 *
 	 * @return boolean True if the user is a member of the group
+	 *
+	 * @deprecated Deprecated since Contao 5.0, to be removed in Contao 6.
 	 */
 	public function isMemberOf($ids)
 	{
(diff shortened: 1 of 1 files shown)


bytehead (approved):

aschempp:
Are you aware the permission voter only works for the front end, but this method works for the back end user as well? Not sure if we ever use (or should use) it for backend though.

leofeyer:
Good point. Should we revert the changes or add a voter that votes on back end groups?

aschempp:
there is never a reason to vote on the back end groups. Imho we should just keep the method (remove the deprecation). _If_ we want to add a deprecation it should be done on the `FrontendUser` class (add a method with deprecation that runs the parent method).

leofeyer:
> there is never a reason to vote on the back end groups.

Why not?

aschempp:
Because for the back end, groups "generate" a set of permissions for modules, elements, alexf etc. There is never a permission on a group.

For the front end thats the opposite, because pages, articles, elements etc. are given permission to groups.

leofeyer:
> There is never a permission on a group.

Maybe not in your installations, but we cannot know how the feature is used in the wild.

ausi:
> For the front end thats the opposite, because pages, articles, elements etc. are given permission to groups.

Don’t we have the same for pages and articles in the backend in `tl_page.chmod`?

PR #5237, opened 2022-09-01
Pass the contao.web_dir in the Automator::generateSymlinks() method

### LINKED ISSUE #5215: Public images not accessible
### Affected version(s)

5.0.1

### Description

I installed Contao 5.0.1 (fresh install, not updated) and created folders and uploaded images via the file browser. The "public" checkmark for the folders is checked. The uploaded images are correctly shown in the file browser.

But neither on the website, nor in the article preview of the backend, the pictures are accessible. I only get a 404 for the images and the debug mode for the frontend shows me the following message: 

`
Router Symfony\Bundle\FrameworkBundle\Routing\Router was not able to match, message "No routes found for "/files/test/logo.png/"."
`

Never had a problem this way. Am I missing something?

leofeyer:
Fixes #5215

However, I wonder if we should remove the `Automator::generateSymlinks()` completely as we already have the `contao:symlinks` command? We could deprecate the usage in Contao 4.13 and remove it in Contao 5.

### DIFF
--- a/core-bundle/contao/library/Contao/Automator.php
+++ b/core-bundle/contao/library/Contao/Automator.php
@@ -365,7 +365,7 @@ class Automator extends System
 	{
 		$container = System::getContainer();
 
-		$webDir = ltrim(str_replace($container->getParameter('kernel.project_dir'), '', $container->getParameter('contao.web_dir')), '/');
+		$webDir = Path::makeRelative($container->getParameter('contao.web_dir'), $container->getParameter('kernel.project_dir'));
 		$command = $container->get('contao.command.symlinks');
 		$status = $command->run(new ArgvInput(array('', $webDir)), new NullOutput());
 
--- a/core-bundle/contao/library/Contao/Automator.php
+++ b/core-bundle/contao/library/Contao/Automator.php
@@ -14,6 +14,7 @@ use FOS\HttpCache\CacheInvalidator;
 use FOS\HttpCacheBundle\CacheManager;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Filesystem\Path;
 
 /**
  * Provide methods to run automated jobs.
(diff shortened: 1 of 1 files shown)


core-bundle/contao/library/Contao/Automator.php
```diff
@@ -365,8 +365,9 @@ public function generateSymlinks()
 	{
 		$container = System::getContainer();
 
+		$webDir = ltrim(str_replace($container->getParameter('kernel.project_dir'), '', $container->getParameter('contao.web_dir')), '/');
```
ausi:
```suggestion
		$webDir = Path::makeRelative($container->getParameter('contao.web_dir'), $container->getParameter('kernel.project_dir'));
```

ausi:
> We could deprecate the usage in Contao 4.13 and remove it in Contao 5.

I’d vote for deprecating in 5 and removing in 6.

core-bundle/contao/library/Contao/Automator.php
leofeyer: Of course. 🤦‍♂️

ausi:
Do we need to replace its usages then?

https://github.com/contao/contao/blob/596ea88104c4543adcdebe494064f77f4bd5f5df/core-bundle/contao/dca/tl_files.php#L896
https://github.com/contao/contao/blob/596ea88104c4543adcdebe494064f77f4bd5f5df/core-bundle/contao/dca/tl_files.php#L907
https://github.com/contao/contao/blob/596ea88104c4543adcdebe494064f77f4bd5f5df/core-bundle/contao/classes/Theme.php#L719
https://github.com/contao/contao/blob/596ea88104c4543adcdebe494064f77f4bd5f5df/core-bundle/contao/config/config.php#L419

leofeyer:
Ah, good point. That‘s why we have kept the method. I have reverted the deprecation.

ausi (approved):

Issue #5069, opened 2022-08-02
flysystem symlink alternative


superMi88:
### Affected version(s)

4.13

### Description

Symlinks dont work in flysystems virtuell filesystem. When I press the synchronize button it will tell me that anything is deleted and removes them from the database. I can still see files in my folder but it will break if I want to use files in my contao application.

Is there any way to handle this, so I can still use smylinks, or is there any other way to add just one folder the the virtuel file system "files"? And access them in the contaos backend in contaos file managment system.

![image](https://user-images.githubusercontent.com/57102323/182361299-a598ca29-ef60-4aaf-9a59-aa374072de6a.png)
![image](https://user-images.githubusercontent.com/57102323/182361577-84da3f86-103a-44e7-9e19-32b237796e98.png)

ausi:
~~I think this was fixed in version 4.13.5.~~ no that was something else I think ☺️
Which version are you using?

ausi:
This issue should already be solved by #4141 @superMi88 can you check why that doesn’t work in your installation?

superMi88:
> ~I think this was fixed in version 4.13.5.~ no that was something else I think ☺️ Which version are you using?

I am using contao 4.13.6

ausi:
@superMi88 did you rebuild the container (run `vendor/bin/contao-setup`) after creating the symlink?

superMi88:
@ausi thank u that works! :) 
but I can only run this command in my shell, and I need this in my code. Is there a quick way to run this in my code without shell_exec? I cant find it in the docu

ausi:
Building the Symfony container should not be done inside your code I think.

superMi88:
I create this symlink in my code and it didnt really worked without running this command. Do I have any other options at this point to solve this?

fritzmg:
How do you create the symlinks? Are you not using the [event](https://docs.contao.org/dev/reference/events/#contao-generate-symlinks)?

superMi88:
I am currently using SymlinkUtil::symlink form contao for creating my symlinks, I guess thats old and I need to switch to this new Event?

fritzmg:
The question is rather _where_/_when_ do you execute your code. But the safest way would be to register your own symlinks via the proper event, yes.

chiron501:
Hello, I'm afraid I have to take up the theme with the symlinks again, as we are still running into problems here.

As recommended, we create the symlink via the event contao.generate_symlinks.
[...]

PR #5240, opened 2022-09-01
Improve the order of operations during the Contao setup

m-vo:
See https://github.com/contao/contao/issues/5069#issuecomment-1226903463.

This PR changes that clear/warmup operations are executed last when running `contao:setup`.

This makes sure the compiler passes that act depending on the filesystem structure will see the structure they need (directories are created, symlinks were made, …). The `bin/contao-setup.php` entrypoint makes sure the cache directory gets nuked beforehand, so the commands will still run on a freshly built container.

@chiron501 Can you test if this solves your issue from #5059?
/cc @ausi

### DIFF
--- a/manager-bundle/src/Command/ContaoSetupCommand.php
+++ b/manager-bundle/src/Command/ContaoSetupCommand.php
@@ -77,12 +77,12 @@ class ContaoSetupCommand extends Command
 
         $commands = [
             ['contao:install-web-dir', $this->webDir, '--env=prod'],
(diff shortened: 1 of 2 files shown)


ausi (approved):

leofeyer:
Thank you @m-vo.

chiron501:
Hi, thanks for your feedback. Unfortunately I'm on vacation at the moment and can't test it right away.
I have forwarded it to a colleague who may be able to do the tests.

@superMi88

superMi88:
Looks like everything is working fine now :)

PR #5241, opened 2022-09-02
Fix several things when submitting data in DC_Table

m-vo:
This PR changes a few things in the `DC_Table#submit()` method.

1) The `UpdateAction` will be constructed with the actual modified values as 3rd parameter. Before this was set to `$this->arrSubmit` which was always empty at that point.

2) The permission check now gets executed every time (unless `Input::post('FORM_SUBMIT') != $this->strTable`, then the method returns early) and before any database operation. I'm not 100% sure this is correct. Before it was executed after the fallback field logic was already applied to the DB (which seems wrong) but also only if the changeset was not empty (which might be OK).

3) The method now does not return early anymore if `$this->arrSubmit` is empty or if there are validation errors, which makes sure the `onforesubmit_callback` always gets executed like it did in earlier Contao versions.

It's hard to get the subtleties of this code right, very happy for some feedback.


aschempp (commented):
The changes mostly make sense to me. The permission callback certainly had the wrong values. Thoughts:

1. was on `onsubmit_callback` previously executed when the widgets did not validate? Because it seems it is now (`$this->noReload` has no early exit anymore)
2. I'm not sure the `tstamp` and `ptable` fields should be in the `onbeforesubmit_callback`, since that is not user data to be modified.
3. Also feels like the array return exception should always have been outside the try-catch block to fail the app if a developer does something wrong, not just show a message in the back end (and all data being lost).

m-vo:
> Also feels like the array return exception should always have been outside the try-catch block to fail the app if a developer does something wrong, not just show a message in the back end (and all data being lost).

Agree :+1:  adjusted in bb40bcaecaa0bb5800384ee535635e79a8684b9a.

m-vo:
> I'm not sure the tstamp and ptable fields should be in the onbeforesubmit_callback, since that is not user data to be modified.

I'm not sure either but I think I would have expected to get the full set of data that will be persisted, same as with the voter. Or would you move the code down to where the actual DB update occurs (i.e. and also hide it for the voters)?

m-vo:
> was on onsubmit_callback previously executed when the widgets did not validate? Because it seems it is now ($this->noReload has no early exit anymore)

Tested it and it was indeed skipped in earlier versions. I readded the check in 935ee6ed46bc69fb5733073d748a1b8bba9128f9.

aschempp (commented):
not sure this make sense to me

1. why not early-return if `noReload` is set? There's really nothing to do then, is there?
2. I also don't see why we would check permissions if there are no values, since the record will never be updated. The only action would be the `onsubmit_callback` right?

m-vo:
> why not early-return if noReload is set? There's really nothing to do then, is there?

I think versions can still get created. In 4.13, the following comment appears multiple times: "Always create a new version if something has changed, even if the form has errors".

> I also don't see why we would check permissions if there are no values, since the record will never be updated. The only action would be the onsubmit_callback right?

I don't get why we would do this optimization tbh. There might be things outside the scope of `DC_Table`, that you could still vote on if you knew the current record ID - and even if there are no database changes made by `DC_Table`. Otherwise you would have to manually check permissions in the `onbeforesubmit` callback yourself which defeats the purpose of having voters to me. But maybe I'm not seeing the full picture, no hard feelings. :relaxed:

core-bundle/contao/drivers/DC_Table.php
```diff
@@ -3205,109 +3205,109 @@ protected function save($varValue)
...
+					$blnVersionize = true;
+				}
 
-		if (!empty($arrValues))
-		{
-			$arrValues['tstamp'] = time();
+				// Update the active record and current field/value (backwards compatibility)
+				$varValue = StringUtil::deserialize($varValue);
+				$this->varValue = $varValue;
+				$this->strField = $strField;
```
ausi: Why setting `$this->varValue` and `$this->strField` here inside the foreach? There is no external code called inside the loop so these should not be necessary.
I think only the `$this->objActiveRecord->{$strField} = StringUtil::deserialize($varValue);` below is needed for BC.

ausi (dismissed):
One thing I noticed, rest LGTM 👍

core-bundle/contao/drivers/DC_Table.php
m-vo: Would you rather set them after the loop like so or completely remove the lines?
```php
$this->strField = array_key_last($arrValues);
$this->varValue = $arrValues[$this->strField];
```

ausi: You mean like if someone relies on the fact that these are set to the last item when executing the `onsubmit_callback`?

But even then, this would already have occured on line 2007, so I’d completely remove them here.

m-vo: Agree. Adjusted in 13c5f0f2cd3ddc4e1a408074a2141a05651d7170.

ausi (approved):

core-bundle/contao/drivers/DC_Table.php
```diff
@@ -3205,109 +3205,105 @@ protected function save($varValue)
...
-			$arrTypes[] = $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['sql']['type'] ?? null;
-
-			if (!isset($arrData['eval']['versionize']) || $arrData['eval']['versionize'] !== false)
-			{
-				$blnVersionize = true;
-			}
+					$varEmpty = Widget::getEmptyValueByFieldType($GLOBALS['TL_DCA'][$this->strTable]['fields'][$strField]['sql'] ?? array());
+					$arrType = array_filter(array($GLOBALS['TL_DCA'][$this->strTable]['fields'][$strField]['sql']['type'] ?? null));
 
-			$this->varValue = StringUtil::deserialize($varValue);
```
leofeyer, line 3292: It seems like this line is missing in the new code. Did you remove it on purpose?

core-bundle/contao/drivers/DC_Table.php
```diff
@@ -3205,109 +3205,105 @@ protected function save($varValue)
...
+					catch (\Exception $e)
 					{
-						$arrValues = $callback($arrValues, $this);
+						$this->noReload = true;
+						Message::addError($e->getMessage());
+
+						break;
 					}
 
 					if (!\is_array($arrValues))
```
leofeyer, line 3250: Shouldn‘t this `if` condition be outside the `foreach` loop?

core-bundle/contao/drivers/DC_Table.php
m-vo: Yes, see the discussion in https://github.com/contao/contao/pull/5241#discussion_r969487342.

core-bundle/contao/drivers/DC_Table.php
m-vo: As `$arrValues` is used as an argument, in case of an error, the next `onbeforesubmit` callback would then potentially fail. Then you might not see the exception.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5242, opened 2022-09-02
Add implicit Twig namespace roots

### LINKED ISSUE #1021: Override templates located in sub folders not working
**Affected version(s)**
4.4

**Description**  
I have the following setup:

A multi-domain project with multiple themes (tl_theme). Each theme has its own templates-folder (let's say `project1` and `project2`).

Now I have two templates named `fe_page.html5` in the following folders:

`<root>/templates/project1/fe_page.html5`
`<root>/templates/project2/fe_page.html5`

Now when I visit both pages everything works as expected.

Then I wanted to bring more structure to my templates by using sub folders (as done in the core):

`<root>/templates/project1/frontend/fe_page.html5`
`<root>/templates/project2/frontend/fe_page.html5`

Now it doesn't work anymore. Both sites use the second template now (in project2).

Maybe this is because somewhere in the code contao is only checking for

`<root>/templates/<template group name>/fe_page.html5`

If not found, it seems to search recursively (*) and the last one wins. Hence `<root>/templates/project2/frontend/fe_page.html5` is used.

Is this a bug? And if so, my solution would be that the step (*) that searches recursively in `templates` should only search recursively in `templates/<current template group`. If something is found, takt it, if not search in `templates` like before.

Bye Defcon0

m-vo:
With this PR the global templates directory (`/templates`) as well as any theme directory selected in the back end (e.g. `/templates/foo/my/theme`) are automatically treated as namespace roots. Same as if they would contain a `.twig-root` file. We can safely do this as nested directories were not allowed in these locations before.

This should IMHO greatly simplify things for people starting to work with the new elements.

Idea by @ausi (https://github.com/contao/contao/issues/1021#issuecomment-1233902891)


ausi (dismissed):

aschempp:
are theme files not nested directory in the `/templates` directory? Also, it was/is always possible to nest theme files to subdirectory, you could have `/foo/bar` and `foo/baz` as separate theme folders. Not sure if that would be an issue here?

m-vo:
> are theme files not nested directory in the /templates directory?

Yes.


> Also, it was/is always possible to nest theme files to subdirectory, you could have /foo/bar and foo/baz as separate theme folders. Not sure if that would be an issue here?

This is still possible. If you've got a `templates/foo` theme directory, listing templates from `templates` won't show your `foo/bar.html.twig` theme template in this case, but will show your `ba/baz.html.twig` from the `ba` directory.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #4121, opened 2022-02-11
Show custom Twig templates in the back end dropdowns

m-vo:
Fixes #3274 

This is a rather simple solution to also get the custom Twig templates (`ce_text_foobar.html.twig`) to show up in the back end dropdowns. We need a more sophisticated one, that replaces the `TemplateLoader` completely, later on. But this will only be doable in Contao 5.

With this PR we're now merging the results from the new template hierarchy with the ones from the `TemplateLoader`. In theory the former would be sufficient because it also contains all PHP templates, but I think implementing it this way is safer, because it will still include files that were manually added.

PR #5243, opened 2022-09-02
Make modern custom templates selectable in the back end

m-vo:
If you want to overwrite the template for the `text` content element and add two variants (`highlight` and `info`), you would setup the following structure in your template directory:

```
📂 templates
 ├─ 📄.twig-root (might be optional, see https://github.com/contao/contao/pull/5242)
 └─ 📂 content_element
     ├─ 📝 text.html.twig
     └─ 📂 text
         ├─ 📝 highlight.html.twig
         └─ 📝 info.html.twig
```

This PR makes the variant templates (here: `content_element/text/highlight` and `content_element/text/info`) selectable in the back end dropdowns for the respective element. Seems I forgot to implement this… :see_no_evil:

The implementation still follows the same ideas as in #4121. I'd like to deprecate `Controller::getTemplateGroups()` at some point and then properly implement this as a service, but we need to get rid of some BC layers first.

/cc @netzarbeiter

### DIFF
--- a/core-bundle/contao/library/Contao/Controller.php
+++ b/core-bundle/contao/library/Contao/Controller.php
@@ -118,7 +118,9 @@ abstract class Controller extends System
 
 		/** @var TemplateHierarchyInterface $templateHierarchy */
 		$templateHierarchy = System::getContainer()->get('contao.twig.filesystem_loader');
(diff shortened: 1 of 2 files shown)


ausi (dismissed):

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5245, opened 2022-09-03
Add a tree view for the debug:contao-twig command

m-vo:
With custom templates like described in #5243 I figured it could be very helpful if the `debug:contao-twig` command was able to display the hierarchy in tree form. This makes it easier to debug your template structure.

This PR does exactly that: If you now add `--tree` as an option, you get an output of existing templates sorted and displayed in a prefix-tree (identifier parts split by `/`). 

Here is an example how this looks if you filter for text content elements for instance: 

image


ausi (dismissed):

ausi (dismissed):

bytehead (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

Issue #5223, opened 2022-08-25
Allow creating modern templates from within the back end


netzarbeiter:
### Affected version(s)

5.0.1

### Description

At the moment it is not possible to create a template in the back end that can be picked in the content element afterwards.  I have to create the Twig templates with the appropriate structure by hand.

![Bildschirmfoto 2022-08-25 um 16 10 42](https://user-images.githubusercontent.com/1942550/186688052-1d3152f9-54aa-49a4-840b-c747f314b7b7.png)

m-vo:
I think we need a more sophisticated solution here that we should discuss first. In my point of view, the template contents should not be copied like before but extended or used from the original template. This is different from the current workflow and IMHO also makes the functionality a bit less important than before.

Nevertheless, some things to consider:
* If this is a convenience function, should you also be able to create variant templates (`content_element/text/foo`)? 
* You should probably only be able to select directories as a target that are namespace roots (`.twig-root`).
* It should maybe be possible to select if you want to extend/use vs overwrite the content. We could also think about generating a comment with some instructions inside the file - e.g. telling you which blocks are available and so on. Probably based on a `backend/template_skeleton.html.twig` or something. Inception :smile:.
* We could add a second global action (e.g. `Derive template`) for this and rename the existing one `Add template (legacy)` instead of creating one complicated screen that does it all. Or can someone think of a simple enough UI for this?

Based on the complexity I'd consider this a feature for 5.1 to be honest.

ausi:
I think a simple solution as a bug fix for 5.0 would be great. 
Creating the file with just the `{% extends … %}` line would be enough IMO.

All the other great features could come with 5.1 then I think.

m-vo:
As dicussed in the Contao call on 22th of September, we might want to conditionally show other options like a field for custom variants with JS (depending on if its a modern fragment template) and handle them as well.

m-vo:
Comment by @mlwebworker in #5635:
> Ich wollte ein Template headline.html.twig im Backend als Varianten-Template sofort im Zielverzeichnis headline [templates/content_element/headline] ablegen. Dabei entsteht dann leider folgende Struktur:
templates/content_element/headline/content_element/headline.html.twig
Für das Anlegen im Ordner Templates finde ich das automatische hinzusetzen des notwendigen Ordners content_element sehr sinnvoll, im Unterordner sollte das aber wenn möglich nicht erfolgen.

m-vo:
Another point made by @mlwebworker in #5636:
[...]

PR #5247, opened 2022-09-04
Allow creating modern templates from within the back end

m-vo:
Addresses https://github.com/contao/contao/issues/5223#issuecomment-1236307138

This PR adds a simple way to create modern templates from within the back end. 

image

There now is a `@Contao/backend/template_skeleton.html.twig` that gets used to create the content of the derived templates. As an example, for the default text content element template this produces an output like this:

```twig
{% extends "@Contao/content_element/text.html.twig" %}

{#
  ** Add changes to the base template here. **

  Hint: Try adjusting blocks and attributes instead of
  overwriting the whole template. This way your version
  can remain compatible with future changes to the base
  template as well as adjustments made by extensions.

  Currently available blocks:
    "picture_component", "image", "sources", "source",
    "schema_org", "figure_component", "media",
    "media_link", "caption", "caption_inner",
    "content", "text_media", "text",
    "text_attributes", "headline_component",
    "headline_attributes", "headline_inner",
    "wrapper", "wrapper_tag", "attributes", "inner"

  Example:
    {% block picture_component %}
       {{ parent() }}
       My additional content for 'picture_component'…
    {% endblock %}
#}

```


ausi (dismissed):

leofeyer (changes requested):
The PR is technically flawless. There is just one thing that should be optimized IMHO:



As you see, the **new** templates are all grouped nicely. However, the **legacy** templates are not:



Why are they not grouped under the `forms/` namespace?

m-vo:
> Why are they not grouped under the forms/ namespace

I did not touch the grouping of the old templates but I reckon changing this could be problematic: In your example `forms` is just a directory and not part of the template and there could be multiple `forms` directories. :thinking: 

I'm also not fully getting the idea behind the existing grouping - care to explain? :smile:

leofeyer:
> I did not touch the grouping of the old templates

You must have. Templates are grouped by bundle in Contao 4.13:



There is no **forms/form_upload.html5** group header in Contao 4.13.

m-vo:
Ah interesting! I'll have a look.

m-vo:
I reverted the unrelated change to the old template listing. I had that in there because this produced an exception in my setup where `contao/contao` is a symlinked composer resource: 

```
Path "/path/to/symlinked/core-bundle/contao/templates/analytics/analytics_matomo.html5" is not inside the Contao root dir "/path/to/app"
```

But we should fix this independently of this PR (and also in 4.9 / 4.13 I guess). Edit: see #5270

aschempp:
Is there an option whether I want to copy the whole template or "only" get the "extends" option? Because I can't easily inspect the current template can I?

m-vo:
> Is there an option whether I want to copy the whole template or "only" get the "extends" option? Because I can't easily inspect the current template can I?

There is no option, we can think about one in #5223. But honestly, I'm not sure what this should achieve. Would you want to have all the sources of all inherited templates including all used templates in your template? 

If you want get fancy, use an IDE like PhpStorm with the Symfony plugin, then you can follow extends like with PHP classes. The `debug:contao-twig <optional prefix>` command also outputs a preview of the first lines to identify a template.

m-vo:
With the changes from #5270 this is how it looks now (filtered for "headline" in this example):

image

ausi (approved):

aschempp:
> Would you want to have all the sources of all inherited templates including all used templates in your template?

No, I would expect exactly the same as it is now: get a copy of the current template with all its content and a new name. Otherwise I have no chance to get to what I want without extensively looking at the original template (which is currently not possible through the back end).

Also be aware that the diff options makes little sense now with this feature 🤷

m-vo:
> No, I would expect exactly the same as it is now: get a copy of the current template with all its content and a new name. 

It's a hierarchy - the current template is just the last one and it likely extends from the same managed name as well. And even if there is one template, it will most definitely extend from yet another template (e.g. the CE base template) and include components. Maybe at some point the template inspector can return the original/last contents (source) of the found blocks and we could improve the `template_skeleton` but that's definitely out of scope for this quick fix.

Copying the *source* template instead of extending it will just destroy every adjustment made by extensions and never makes sense to me.

> Also be aware that the diff options makes little sense now with this feature :shrug:

That's a different issue.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5249, opened 2022-09-05
Fix cache tagging for aliased content elements

### LINKED ISSUE #5248: Cache is not refreshed when a change is made to a related content item.
### Affected version(s)

4.13.8

### Description

A content element is maintained centrally in a page tree and used in further page trees. After changes, the cache on the related pages is not discarded.

fritzmg:
Fixes #5248

`ContentElement::generate` automatically tags the response with the ID of the content element. However, `ContentAlias` overwrites the ID of the model _with its own ID_ before that happens and thus only the ID of the alias element itself gets tagged - but not the ID of the _included_ element. Thus, if you make any changes to the _included_ content element, the cache will not be invalidated.

This PR fixes that by letting `ContentAlias` always also tag the response with the model of the _included_ element.

Furthermore I noticed that if you are using fragments, the ID of the _alias_ content element does not get tagged at all. This PR also fixes that. Now always both IDs (the original ID and the current ID) get tagged, regardless of whether it's a legacy content element or a fragment.

_Note:_ this issue also happens in Contao 4.9, I will provide a separate PR for that. This PR should be merged first.

### DIFF
--- a/core-bundle/src/Resources/contao/elements/ContentAlias.php
+++ b/core-bundle/src/Resources/contao/elements/ContentAlias.php
@@ -51,8 +51,17 @@ class ContentAlias extends ContentElement
 				$proxy->cssID = ' id="' . $this->cssID[0] . '"';
 			}
 
+			// Tag the alias element
+			if ($this->objModel !== null)
+			{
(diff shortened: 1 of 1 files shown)


aschempp (approved):

leofeyer:
Thank you @fritzmg.

PR #5254, opened 2022-09-07
Fix template options when using legacy elements

fritzmg:
Currently the template options in the back end only allow you to select the new fragment templates, even if you switch back to the legacy content element, e.g. via

```php
// contao/config/config.php
$GLOBALS['TL_CTE']['texts']['text'] = \Contao\ContentText::class;
```

Thus you won't be able to select a `templates/ce_text_custom.html5` template for example.

This is because the actual fragment and its template is still registered and currently in the `TemplateOptionsListener` any custom template takes precedence over the legacy logic.

This PR fixes that by always using the legacy logic when the element/module type is not registered via the proxy class.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
+++ b/core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
@@ -46,7 +46,10 @@ class TemplateOptionsListener
         }
 
         $type = $dc->getCurrentRecord()['type'] ?? null;
-        $defaultTemplate = $this->customTemplates[$type] ?? $this->getLegacyDefaultTemplate($dc);
+
+        if (null !== $type) {
+            $defaultTemplate = $this->getLegacyDefaultTemplate($type) ?? $this->customTemplates[$type] ?? null;
+        }
 
         if (empty($defaultTemplate)) {
             $defaultTemplate = $this->templatePrefix.$type;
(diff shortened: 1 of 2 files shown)


m-vo (approved):
:heart:

core-bundle/src/EventListener/DataContainer/TemplateOptionsListener.php
```diff
@@ -46,7 +46,10 @@ public function __invoke(DataContainer $dc): array
         }
 
         $type = $dc->getCurrentRecord()['type'] ?? null;
-        $defaultTemplate = $this->customTemplates[$type] ?? $this->getLegacyDefaultTemplate($dc);
+
+        if (null !== $type) {
+            $defaultTemplate = $this->getLegacyDefaultTemplate($type) ?? $this->customTemplates[$type] ?? null;
```
leofeyer, line 51: How can this be correct? Now the legacy template takes precedence over the custom template.

fritzmg: Custom templates only exist for non-legacy elements and modules thus `$this->getLegacyDefaultTemplate()` will always return `null` for fragments.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5255, opened 2022-09-07
Respect showHidden option in last breadcrumb as well

AlexanderWillner:
We've identified this issue in Contao 4.13.8. The option not to show hidden elements in the breadcrumb is not respected in case we're on the last item in the list.

### DIFF
--- a/core-bundle/contao/modules/ModuleBreadcrumb.php
+++ b/core-bundle/contao/modules/ModuleBreadcrumb.php
@@ -190,15 +190,18 @@ class ModuleBreadcrumb extends Module
 		// Active page
 		else
 		{
-			$items[] = array
-			(
-				'isRoot'   => false,
-				'isActive' => true,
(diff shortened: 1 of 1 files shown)


aschempp:
I think this is intentional, because the breadcrumb should always show the page you are currently on. Can you give an example why that should be changed?

AlexanderWillner:
> I think this is intentional, because the breadcrumb should always show the page you are currently on. Can you give an example why that should be changed?

I guessed that this would be the answer ;) Well, besides that the documentation should probably then state that the showHidden option does not apply to the last entry and that it was a customer request, the specific use case involved Isotope:

- Go to list of orders
- Select an order (the link includes further GET parameters)
- The last bread crumb does not include the GET parameters (fixing this might be another alternative / another issue)

aschempp:
> The last bread crumb does not include the GET parameters (fixing this might be another alternative / another issue)

Then the link is a bug to me, but not the `showHidden`

PR #5261, opened 2022-09-09
Make sure language is always a string

aschempp:
I have no idea why, but I got a request logged with header `Accept-Language: 1`. This did lead to an error of `Contao\CoreBundle\Util\LocaleUtil::getFallbacks(): Argument #1 ($locale) must be of type string, int given` which should not happen anymore with this change.

### DIFF
--- a/core-bundle/src/Routing/AbstractPageRouteProvider.php
+++ b/core-bundle/src/Routing/AbstractPageRouteProvider.php
@@ -217,6 +217,8 @@ abstract class AbstractPageRouteProvider implements RouteProviderInterface
         $result = [];
 
         foreach ($languages as $language) {
+            $language = (string) $language;
+
             if (!$locales = LocaleUtil::getFallbacks($language)) {
                 continue;
             }


fritzmg:
That's odd. Isn't this a bug in Symfony and shouldn't an error already be thrown there?

https://github.com/symfony/symfony/blob/5fd94ec1106a36a25f6426fb4865b6c8f06e6ad1/src/Symfony/Component/HttpFoundation/Request.php#L1693

Because at this point `$lang` is already an `int` and [`str_contains`](https://www.php.net/manual/en/function.str-contains.php) (both the polyfill as well as the native function) require the first parameter to be a `string`.

I can confirm the issue though, via

```
curl --header "Accept-Language: 1" https://c413.local/
```

leofeyer:
Yes, it is a Symfony error. We should fix it there instead of here.

ausi:
> We should fix it there instead of here.

https://github.com/symfony/symfony/pull/47530

leofeyer:
https://github.com/symfony/symfony/pull/47530 has been merged, so I‘m closing this PR now.

PR #5265, opened 2022-09-12
Fix filter options with options callback

aschempp:
This line in DC_Table has two problems:
 1. sorting the options by callback makes no sense, because the list it later sorted by label (I guess that was added later)
 2. this line causes options to be removed if they are not returned by the options callback. For example if you have some "leftover" values you can never filter for them even though they exist in the database. 

FYI what this does not yet fix is multidimensional array from the options_callback, but I think that's a separate issue.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -5868,9 +5868,6 @@ class DC_Table extends DataContainer implements \listable, \editable
 					{
 						$options_callback = $GLOBALS['TL_DCA'][$this->strTable]['fields'][$field]['options_callback']($this);
 					}
-
-					// Sort options according to the keys of the callback array
-					$options = array_intersect(array_keys($options_callback), $options);
 				}
 
 				$options_sorter = array();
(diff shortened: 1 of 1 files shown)


leofeyer:
How can we reproduce the issue before the fix?

aschempp:
have a table with select options filter, and an options_callback that returns nothing (or not all options from the DB)

leofeyer (changes requested):
I‘m sorry, but just as in #4776, I don‘t think we should remove things that have been in the code for more than 11 years just because "you guess that something has been added later" or you are "pretty sure that something does not work".

If you want to remove things like that, please provide a detailed reproduction case and explain how your change fixed the issue and how you made sure that your change does not have unintended side-effects. Otherwise I cannot merge PRs like this one. 🤷‍♂️

aschempp:
I don't understand your problem. The code has a description that says "sort by options_callback". However, a few lines below, the options are sorted **again** for the label. I'm pretty sure this label sorting has been added later, otherwise the now removed line never made sense. Also, you can clearly see the line will remove any option that does not exist in the options_callback, which obviously is a problem?

Btw I manually tested this multiple times in my application where the problem occurs 😉

aschempp:
@leofeyer just as https://github.com/contao/contao/pull/4776 could you please look into this again? 😜 (or mark it up-for-discussion)

ausi (dismissed):

m-vo (dismissed):

ausi (approved):

fritzmg (approved):

leofeyer (approved):

Issue #5067, opened 2022-08-01
TinyMCE prevents `accesskey` shortcuts


ameotoko:
### Affected version(s)

4.x

### Description

Since long time, ["accesskey" shortcuts](https://docs.contao.org/manual/en/administration-area/back-end-keyboard-shortcuts/#keyboard-shortcuts-in-edit-mode) do not work when invoked while TinyMCE instance is in focus, probably because TinyMCE does not let them bubble.

It's quite easy to fix by adding custom shortcuts to TinyMCE, but maybe it should be considered a bug in TinyMCE? I posted an issue there (see https://github.com/tinymce/tinymce/issues/8026), but also posting here so it's not forgotten.

I can also provide a PR if you decide to fix it in the meantime.

aschempp:
That's something that always annoyed me but I saw as a known limitation. Fixing it would be great imho. Do you think we can fix the bubbling or just catch our known access keys?

leofeyer:
See #5266 for a fix in Contao 5.0. Needs to be backported once merged.

PR #5266, opened 2022-09-12
Fix several autofocus issues in the back end

leofeyer:
The PR fixes two autofocus issues:

* `input.autocomplete` is an empty string if not set and never `undefined`, therefore the first matching field was never focused.
* Our access keys did not work in TinyMCE (see #5067).


core-bundle/contao/templates/backend/be_tinyMCE.html5
```diff
@@ -16,26 +16,29 @@ window.tinymce && tinymce.init({
...
-  setup: function(editor) {
+  setup: editor => {
     editor.getElement().removeAttribute('required');
+    document.querySelectorAll('[accesskey]').forEach(el => {
+      editor.addShortcut('access+' + el.accessKey, el.id, () => el.click());
+    });
   },
-  init_instance_callback: function(editor) {
+  init_instance_callback: editor => {
+    console.log(document.activeElement.id);
```
aschempp: that's probably a leftover?

core-bundle/contao/templates/backend/be_tinyMCE.html5
```diff
@@ -16,26 +16,29 @@ window.tinymce && tinymce.init({
...
-    file_picker_callback: function (callback, value, meta) {
+    file_picker_callback: (callback, value, meta) => {
       Backend.openModalSelector({
         'id': 'tl_listing',
         'title': document.getElement('.tox-dialog__title').get('text'),
         'url': Contao.routes.backend_picker + '?context=' + (meta.filetype == 'file' ? 'link' : 'file') + '&amp;extras[fieldType]=radio&amp;extras[filesOnly]=true&amp;extras[source]=<?= $this->source ?>&amp;value=' + value + '&amp;popup=1',
-        'callback': function(table, val) {
-          callback(val.join(','));
-        }
+        'callback': (table, val) => callback(val.join(','))
```
aschempp: using array functions will drop support for IE completely, is that intentional?

core-bundle/contao/templates/backend/be_tinyMCE.html5
leofeyer: Yes. Removed in 9819d6cfbba5771b096a646f6199792d92a7d0a8.

core-bundle/contao/templates/backend/be_tinyMCE.html5
leofeyer: <del>No, not really. Reverted in 9819d6cfbba5771b096a646f6199792d92a7d0a8.</del>

We are already using arrow functions, so IE is no longer supported anyway:

https://github.com/contao/contao/blob/4af387e032da7259112bbc5ea8f235a9f3b26286/core-bundle/src/Resources/contao/drivers/DC_Table.php#L2095-L2098

core-bundle/contao/templates/backend/be_tinyMCE.html5
```diff
@@ -16,26 +16,28 @@ window.tinymce && tinymce.init({
...
+    document.querySelectorAll('[accesskey]').forEach(function(el) {
+      editor.addShortcut('access+' + el.accessKey, el.id, () => el.click());
+    });
   },
   init_instance_callback: function(editor) {
     if (document.activeElement && document.activeElement.id && document.activeElement.id == editor.id) {
       editor.editorManager.get(editor.id).focus();
     }
-    editor.on('focus', function() { Backend.getScrollOffset(); });
+    editor.on('focus', () => Backend.getScrollOffset());
```
aschempp, line 31: this is still an arrow function

m-vo: That's what the commit message says, no?
image

ausi (approved):

core-bundle/contao/templates/backend/be_tinyMCE.html5
aschempp: but its not the case for all other functions 🤷

leofeyer: See https://github.com/contao/contao/pull/5266#discussion_r968344179

PR #5267, opened 2022-09-12
Move DC_Table HTML into templates

m-vo:
This PR introduces templates for `DC_Table` and removes the inline HTML wherever possible. This already made it possible to reuse several parts (see for instance the `_buttons` component template or the `override_all` template extending from `edit_all`). 

**Main goals here:** 
  - output encoding by default
  - refactorable HTML (no more disjoined opening and closing parts) 
  - working in the file got a bit less stressful :smile:, i.e. I can see a lot of refactoring opportunities now

What wasn't my goal, was to make the back end adjustable by overwriting templates. While it would technically be possible now, I intentionally did not add blocks and other template abstractions for now. This can be done at a later point when things are settled.

**Review/testing:**

The PR is structured in two parts: the first commits (until c84d8a9466cd9e24e1439c7fa7adf4650df05808) only *add* the new templates and build the context in parallel to the existing implementation. The commits after this then remove the inline HTML and clean things up a bit.

While working on the PR, I wrote myself a little helper function that compared the normalized HTML of both the old and new versions and generated a diff. That's why things like the order of HTML attributes are still the same or there is an occasional `&nbsp;` (even if not consistent across templates) in order to get a result that - besides whitespaces - is identical to the old one. I'm happy to clean these things on the template side once we're moving on. 

**Bugs that were accidentally removed:**
- The parent view created unclosed `<ul>` elements before.
- Nested palettes in an AJAX request when using `editAll` were cut after the first `EOF` and not wrapped - not 100% sure if this problem actually can happen but still a weird behavior.


leofeyer:
Please be aware that this PR will not only be difficult to review, but it is also impossible not to break something. Therefore I‘d vote for not spending time on this and leaving the inline HTML where it is.

m-vo:
> Please be aware that this PR will not only be difficult to review, but it is also impossible not to break something. Therefore I‘d vote for not spending time on this and leaving the inline HTML where it is.

I wholeheartedly disagree. :relaxed:  I took a lot of care to refactor these things and diffed the output to not break stuff (see approach in initial description). The only thing, that is fundamentally different, is the encoding. And that can easily be fixed in the templates (e.g. if a `|raw` is missing at a place where HTML should be output). The business logic is still nearly identical.

To me, things are already so much more approachable now, as you can understand the markup and the individual components, future XSS vulnerabilities will become less likely and so on… Sure, we can also switch out the whole DC concept and start from scratch, but honestly I'm not seeing this coming in the near future (it's technical debt…). Till then every bit of fixing and adjusting stuff in this file is a major pain. With the refactorings from this PR, I can already see a lot of subsequent easy wins that could make things even better, so why make our life harder as it needs to be? :slightly_smiling_face:

aschempp:
I'm sorry but I fully agree with @leofeyer here. Before we start refactoring DC_Table, I would prefer to have actual functional tests (afaik @MDevster has started something on that), and I would prefer to split the components away from DC_Table if a template is added (e.g. generating the buttons should become a separate class).

m-vo:
Did you even look at the code - especially the commit I pointed at in the initial description? 

> Before we start refactoring DC_Table, I would prefer to have actual functional tests

Sure, I would love to have functional tests, too. But I'm not seeing this coming, either. 

> I would prefer to split the components away from DC_Table if a template is added (e.g. generating the buttons should become a separate class).

For BC the individual methods in `DC_Table` must still return a string, so to avoid composition in code (which is the goal), the methods must render the *full* template anyways. The templates reuse things of course, that's why the context for the buttons for instance is all put under `$context['buttons']` and there is a respective include/component template for this. Constructing the context for this part can still be done in a separate class later on and would sure be nice. 

The approach the other way round - creating idealized classes and then trying to make use of them - does not work for me here. Without refactoring the other bits, you would not even be able to identify the closing `</div>` for your component container, because it gets output at three different places.

Toflar:
We should definitely discuss this at the developers meeting :)

MDevster:
I wrote some tests and created a repo at https://github.com/pdir/contao-cypress-testing. It's just not so easy to know where to start and what to test. Also nothing is optimized yet and I still have to define a lot (how do I address the elements, order of tests, etc). I would include suitable tests for this pull request and make them available tomorrow. Would it make sense for the Cypress tests to have their own isssue to discuss them?

m-vo:
> I wrote some tests and created a repo at [pdir/contao-cypress-testing](https://github.com/pdir/contao-cypress-testing).

Amazing stuff! Manually set it up and ran the tests against both 5.0 and this PR branch. Works like a charm. :rocket: 
Let's create a new issue for this and discuss how we can proceed and then maybe add more specific tests for this PR, wdyt?

m-vo:
I added the `up for discussion` tag because I think we should discuss this before the PR gets stale / too many changes are piling up.

core-bundle/contao/templates/_new/backend/crud/DC_Table/tree_view.html.twig
```diff
@@ -0,0 +1,80 @@
...
+                </div>
+            {% endif %}
+
+            {# Tree #}
+            {% set tree_attributes = attrs()
+                .addClass('tl_listing_container tree_view')
+                .set('id', 'tl_listing')
+                .mergeWith(picker.value_attribute)
+            %}
+            <div{{ tree_attributes }}>
```
ausi, line 30:
```suggestion
            <div{{ tree_attributes }}>
                {{ breadcrumb_rendered|raw }}
```

core-bundle/contao/templates/_new/backend/crud/DC_Table/_picker.html.twig
```diff
@@ -0,0 +1,25 @@
+{% trans_default_domain "contao_default" %}
+
+{% block picker_select_all %}
+    {% if select_mode or picker.field_type == 'checkbox' %}
+        {% if wrapped is not defined or not wrapped %}
```
ausi:
```suggestion
        {% if wrapped is defined and not wrapped %}
```

The default for `wrapped` is meant to be true, right?

m-vo:
Superseded by #8975 and followup PRs.

PR #5268, opened 2022-09-12
Fix page mounts not being applied correctly

Toflar:
Fixes page mount permissions not being applied correctly.

Steps to reproduce:

1. Log into demo.contao.org
2. Create a second root page and name it `Test root`
3. Add a page `Foobar` to your new root page `Test root`
4. Edit e.g. h.lewis and configure the page mounts to only `Contao Official Demo` and your new `Foobar Page`

Bildschirmfoto 2022-09-12 um 16 11 02

5. Switch to `h.lewis`
6. Access the page structure. It will look like this:

Bildschirmfoto 2022-09-12 um 16 12 50

The entire `Contao Official Demo` page tree is missing.

That's because the configured page mount itself is already a root page and does not have any parent records. It needs to be added itself to the visible root trails.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -6246,7 +6246,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 			{
 				$parentRecords = $this->Database->getParentRecords($id, $table, true);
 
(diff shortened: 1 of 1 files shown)


leofeyer (dismissed):

leofeyer:
Thank you @Toflar.

leofeyer (approved):

PR #5269, opened 2022-09-12
Don't return empty theme directories in the TemplateLocator

m-vo:
This small change makes sure that empty entries in the `tl_theme.templates` column aren't considered as valid theme directories. As these directories are meant to be relative, the `is_dir` check in L69 will otherwise always return `true`, because the effective directory is then the project root. :smiling_face_with_tear:


ausi (approved):

leofeyer:
Thank you @m-vo.

PR #5270, opened 2022-09-12
Allow symlinked bundle paths in the template files listing

m-vo:
See https://github.com/contao/contao/pull/5247#issuecomment-1243987509

`StringUtil::stripRootDir` throws an exception if the given path is outside the project dir. Exactly this is the case, when listing templates from a symlinked source. In my development setup this happens if my app requires `contao/contao` and there is a respective `repositories` entry in the `composer.json`, so that the bundle files will get symlinked. The resource paths will now be outside the project dir.

This PR allows arbitrary relative paths and adjusts the pattern matching.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -246,15 +246,15 @@ class tl_templates extends Contao\Backend
 			$strRelpath = Path::makeRelative($file->getPathname(), $projectDir);
 
 			$modulePatterns = array(
-				"vendor/(?'module'[^/]+/[^/]+)",
-				"\\.\\..*(?'module'[^/]+/[^/]+)/(?:src/Resources/contao/templates|contao/templates)",
-				"system/modules/(?'module'[^/]+)"
+				"vendor/([^/]+/[^/]+)",
+				"\\.\\..*?([^/]+/[^/]+)/(?:src/Resources/contao/templates|contao/templates)",
+				"system/modules/([^/]+)"
 			);
-			preg_match('@^(?:' . implode('|', $modulePatterns) . ')/.*$@UJ', $strRelpath, $matches);
+			preg_match('@^(?|' . implode('|', $modulePatterns) . ')/.*$@', $strRelpath, $matches);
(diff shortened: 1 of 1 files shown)


fritzmg:
Ah, this should finally fix https://github.com/contao/core-bundle/issues/1114 for good 🎉

core-bundle/src/Resources/contao/dca/tl_templates.php
```diff
@@ -235,11 +237,25 @@ public function addNewTemplate()
...
+			$modulePatterns = array(
+				"vendor/(?'module'[^/]+/[^/]+)",
+				"\\.\\..*(?'module'[^/]+/[^/]+)/(?:src/Resources/contao/templates|contao/templates)",
+				"system/modules/(?'module'[^/]+)"
+			);
+			preg_match('@^(?:' . implode('|', $modulePatterns) . ')/.*$@UJ', $strRelpath, $matches);
+
+			// Use the matched "module" group and fall back to the full
+			// directory path (e.g. "contao/templates" in the app).
+			$strModule = $matches['module'] ?? dirname($strRelpath);
```
ausi:
```suggestion
			$modulePatterns = array(
				"vendor/([^/]+/[^/]+)",
				"\\.\\..*?([^/]+/[^/]+)/(?:src/Resources/contao/templates|contao/templates)",
				"system/modules/([^/]+)"
			);
			preg_match('@^(?|' . implode('|', $modulePatterns) . ')/.*$@', $strRelpath, $matches);

			// Use the matched "module" group and fall back to the full
			// directory path (e.g. "contao/templates" in the app).
			$strModule = $matches[1] ?? dirname($strRelpath);
```

I’d write the regex without the `UJ` modifiers, but that’s just a nitpick, I’m fine with this version too. ☺️
Duplicate group numbers: <https://www.pcre.org/current/doc/html/pcre2pattern.html#SEC15>

ausi (dismissed):

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5154, opened 2022-08-15
Allow empty key for subpalette suffix based on select

dennisbohn:
This pull request for contao 4.9 is the equivalent to a pull request for contao 4.13. https://github.com/contao/contao/pull/5112

It allows to set a subpalette for a select field, when the value is empty or 0.

PR #5271, opened 2022-09-13
Fix the order of the palette combiner

ausi:
Currently having two selectors (`palettes.__selector__`) where one has the value `foo` and the other `bar` the palette `foobar` gets used as expected.

But if another selector appears before them in the `__selector__` array that has a value of `null`, `false` or `''` now the palett `foo` gets selected which is unexpected IMO.

This issue currently exists for example the `tl_module.rocksolid_sliderrsts_default` palette when installing `madeyourday/contao-rocksolid-slider` together with `alnv/catalog-manager`.

This PR solves this by skipping empty values when generating the palette names.


core-bundle/tests/Contao/DataContainerTest.php
```diff
@@ -0,0 +1,64 @@
...
+                ['foo'],
+                ['foo'],
+            ],
+            [
+                ['foo', 'bar'],
+                ['bar', 'foobar', 'foo'],
+            ],
+            [
+                ['foo', 'bar', 'baz'],
+                ['baz', 'barbaz', 'bar', 'foobar', 'foobarbaz', 'foobaz', 'foo'],
```
ausi: Also, this order seems totally weird to me and I’d like to change it to:

```php
['foobarbaz', 'barbaz', 'foobaz', 'foobar', 'baz', 'bar', 'foo']
```

But I’m not sure if this could cause BC breaks.

leofeyer: I‘m not sure, either. We should probably only change it in Contao 5 (or even6)?

ausi: I will leave this PR as it is then, and create a separate issue for changing this order.

ausi:
Looks like this issue was introduced by #5154 cc @dennisbohn

So an alternative fix would be to wrap the `$sValues[] = $trigger;` inside an `if ($trigger)`

leofeyer:
👍 for the alternative fix.

dennisbohn:
Thank you for fixing this.

leofeyer:
The failing test seems to be related to the changes:

leofeyer (dismissed):

core-bundle/tests/Contao/DcTableTest.php
```diff
@@ -0,0 +1,87 @@
...
+            ->method('prepare')
+            ->willReturn($statement)
+        ;
+
+        $reflection = new \ReflectionClass(DC_Table::class);
+        $dataContainer = $reflection->newInstanceWithoutConstructor();
+
+        $method = $reflection->getMethod('import');
+        $method->setAccessible(true);
+        $method->invoke($dataContainer, $database, 'Database');
```
leofeyer: This did fix the test, however, if you run `composer run unit-tests -- --filter DcTableTest`, there is a new issue:



Maybe we should revert the fix and add `#[AllowDynamicProperties]` to the `DC_Table` class? Or we could adjust the `getPalette()` method so it uses `System::getContainer()->get('database_connection')` instead of `$this->Database`?

ausi: How about b30c8500eabd407fa5f7eb6e4e8eb661e1323451 ?

m-vo: > How about https://github.com/contao/contao/commit/b30c8500eabd407fa5f7eb6e4e8eb661e1323451 ?

IMHO that's the right fix to get test isolation. :+1: If this is still not enough, we could also create a `DC_TableStub` (or an inline class) that removes the constructor and overrides `getCurrentRecord()`.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5277, opened 2022-09-15
Fix the order of the palette combiner

ausi:
Same as #5271 for Contao 4.13 (different tearDown method)


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5279, opened 2022-09-15
Add id check to dc table

koertho:
This PR adds additional checks for id to DC_Table.

Problem: DC_Table accepts values other then int when loading the id (any value coming from Input::get() is accepted and not further checked). Since the the database queries are prepared statements, this is direct issue as queries like `/contao?act=edit&do=nc_notifications&id=1%20or%20true&table=tl_nc_message` lead to database queries like `SELECT * FROM tl_nc_message WHERE id='1 or true' LIMIT 0,1`, which still works. But since the DataContainer object is passed to many events/callbacks, some code may expect an integer value (which the name and annotation suggest) and therefore produce problems. In my case DcaWizard produced wrong queries leading to this issue: https://github.com/terminal42/contao-dcawizard/issues/32

My idea is to check if the value is numeric, and, if not, don't use it. Otherwise cast it to integer.

This problem and possible solution were also discussed her: https://contao.slack.com/archives/CK4J0KNDB/p1663238445510959

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -117,6 +117,12 @@ class DC_Table extends DataContainer implements \listable, \editable
 
 		$this->intId = Input::get('id');
 
+		if (!is_numeric($this->intId)) {
+			$this->intId = null;
+		} else {
+			$this->intId = (int) $this->intId;
+		}
+
 		// Clear the clipboard
 		if (isset($_GET['clipboard']))
 		{
@@ -1788,6 +1794,10 @@ class DC_Table extends DataContainer implements \listable, \editable
 		if ($intId)
 		{
 			$this->intId = $intId;
+		}
+
+		if (!$this->intId || !is_numeric($this->intId)) {
+			throw new AccessDeniedException('Invalid ' . $this->strTable . ' ID.');
 		}
 
 		// Get the current record


leofeyer:
Please do not post links to non-public resources here. Explain the problem and how your PR solves it.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -117,6 +117,12 @@ public function __construct($strTable, $arrModule=array())
 
 		$this->intId = Input::get('id');
 
+		if (!is_numeric($this->intId)) {
+			$this->intId = null;
```
m-vo, line 121: The type annotation for `DataContainer#$intId` currently says `string|int`, sure this should be `null`?

Can't we just always cast to `int` in `DC_Table`?

fritzmg: It can be string within `DC_Folder` for example.

koertho: Input::get() returns null, if the parameter is not set, so I thought null would be reasonable.

ausi:
Casting `intId` to integer (in DC_Table, not DC_Folder) would be great I think. 
But are we sure that DC_Table cannot be used with string primary keys?

koertho:
I just updated the description!

core-bundle/src/Resources/contao/drivers/DC_Table.php
koertho: > It can be string within `DC_Folder` for example.

This change would be only for `DC_Table`

koertho:
Something about this topic? Or should I close this PR?

aschempp:
As @ausi pointed out, there are valid cases of `$dc->id` not being an integer. If I understand correctly, DC_Table (always) uses prepared statements and does not have an injection issue, but dcawizard does?

koertho:
The annotation points out it should be an integer, but it is never checked or casted. This can lead to issues in extensions which rely on the value. In my case this is dca wizard.

aschempp:
I would suggest to close this as known limitation since the original issue is fixed in https://github.com/terminal42/contao-dcawizard/issues/32

PR #5284, opened 2022-09-16
Include mounted root pages in the topMostRootIds

### LINKED ISSUE #5283: Anlegen von Seiten bei frischen Installationen nicht möglich
### Affected version(s)

4.13.9

### Description

Mit einer frischen Installation von Contao 4.13.9 konnte ich lediglich "Startpunkt einer Website" hinzufügen. 

image

Das Anlegen neuer Seiten ist nicht möglich. Ich kann zwar beliebig viele Startpunkte hinzufügen, diese dann aber auch nicht mehr bearbeiten oder löschen:

image

Workaround: Downgrade auf 4.13.8, um eine Seite anzulegen, danach wieder Update auf 4.13.9 und ich kann dann zumindest unter der neu angelegten Seite weitere Elemente anlegen. Aber das Bearbeiten von Startpunkten oder Einfügen in Startpunkten ist nicht möglich.

ausi:
Fixes #5283

Reverts #5268 and applies a different fix

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3702,7 +3702,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 		if (!empty($this->visibleRootTrails))
 		{
 			// Make sure we use the topmost root IDs only from all the visible root trail ids and also ensure correct sorting
-			$topMostRootIds = $this->Database->prepare("SELECT id FROM $table WHERE (pid=0 OR pid IS NULL) AND id IN (" . implode(',', $this->visibleRootTrails) . ")" . ($this->Database->fieldExists('sorting', $table) ? ' ORDER BY sorting, id' : ''))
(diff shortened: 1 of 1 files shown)


ausi:
As far as I understand it, all IDs that are in `$this->visibleRootTrails` are considered readonly and get no action buttons. So we should not add a mounted page to this array.

But when loading the `$topMostRootIds` we should include all mounted pages too so that we get all root pages (directly mounted as well as readonly ones)

Toflar (approved):
Should we add `array_unique()`?

ausi:
> Should we add `array_unique()`?

Not needed I think as the query would handle that, and also `$this->root` should never overlap with `$this->visibleRootTrails`.

Toflar (changes requested):
Blocking this because I'm not entirely sure this is correct. `$topMostRootIds` should only include the IDs from `$this->root` which are root themselves (`pid=0`). Otherwise the tree will start rendering subpages from the top, no?

Toflar (approved):
Ah, it should work because `pid=0` is part of the query. It's fine then :)

leofeyer:
Thank you @ausi.

PR #5287, opened 2022-09-18
Use main request in Input class

bytehead:
Fixes #5252

PR #5288, opened 2022-09-18
Store Input::setGet in the request attributes

### LINKED ISSUE #5252: Input::get('auto_item') is always empty in contao fragment controllers
### Affected version(s)

Contao 5.0.1, PHP 8.1

### Description

This issue refers to https://community.contao.org/de/showthread.php?83670-auto_item-mit-AbstractFrontendModuleController&p=562745&viewfull=1#post562745

I'm not sure if this is a bug. But somehow I'm not able to grab the auto item parameter without setting it like this:

```php
        // Set the item from the auto_item parameter and remove auto_item from unused route parameters
        if (isset($_GET['auto_item']) && '' !== $_GET['auto_item']) {
            // Bug? Why $this->input->get('auto_item') is empty here?
            // Bug? Without Input::setGet() the code execution ends in a \Contao\UnusedArgumentsException (unused argument auto_item)
            $this->input->setGet('auto_item', $_GET['auto_item']);
        }
```

See my whole controller:

```php
<?php

declare(strict_types=1);

namespace App\Controller\FrontendModule;

use Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController;
use Contao\CoreBundle\DependencyInjection\Attribute\AsFrontendModule;
use Contao\CoreBundle\Framework\Adapter;
use Contao\CoreBundle\Framework\ContaoFramework;
use Contao\Input;
use Contao\ModuleModel;
use Contao\Template;

#[AsFrontendModule(category: 'miscellaneous')]
class TestController extends AbstractFrontendModuleController
{
    protected ContaoFramework $framework;
    protected ?string $item = null;

    // Adapters
    protected Adapter $input;

    public function __construct(ContaoFramework $framework)
    {
        $this->framework = $framework;

        // Adapters
        $this->input = $this->framework->getAdapter(Input::class);
    }

    public function __invoke(Request $request, ModuleModel $model, string $section, array $classes = null): Response
    {

        // Set the item from the auto_item parameter and remove auto_item from unused route parameters
        if (isset($_GET['auto_item']) && '' !== $_GET['auto_item']) {
            // Bug? Why $this->input->get('auto_item') is empty here?
            // Bug? Without Input::setGet() the code execution ends in a \Contao\UnusedArgumentsException (unused argument auto_item)
            $this->input->setGet('auto_item', $_GET['auto_item']);
        }

        // It's important to call Input::get('auto_item') at least once, if we have an auto item set.
        // Otherwise, Contao will throw a \Contao\UnusedArgumentsException (unused argument auto_item).
        $this->item = $this->input->get('auto_item') ?? null;

        return parent::__invoke($request, $model, $section, $classes);
    }

    protected function getResponse(Template $template, ModuleModel $model, Request $request): Response
    {

        $template->result = $this->item ? 'Has auto-item' : 'Has no auto-item';

    }
}

```

ausi:
Fixes #5252
Alternative to #5287

I think using the current request (instead of the main request) is correct, but we should store the data that is added via `setGet()` like the `auto_item` value in the request attributes directly so that they are kept when duplicating the request.

This also has the benefit that we could now access the `auto_item` value without using the `Input` class.

### DIFF
--- a/core-bundle/contao/library/Contao/Input.php
+++ b/core-bundle/contao/library/Contao/Input.php
@@ -152,6 +152,11 @@ class Input
 		{
 			$keys = $request->query->keys();
 
+			if ($request->attributes->has('auto_item'))
+			{
+				$keys[] = 'auto_item';
+			}
+
 			foreach ($request->attributes->get('_contao_input')['setGet'] ?? array() as $key => $value)
 			{
 				if ($value === null)
@@ -306,9 +311,23 @@ class Input
 
 		if ($request = static::getRequest())
 		{
-			$inputAttributes = $request->attributes->get('_contao_input', array());
-			$inputAttributes['setGet'][$strKey] = $varValue;
-			$request->attributes->set('_contao_input', $inputAttributes);
+			if ('auto_item' === $strKey)
+			{
+				if (null !== $varValue)
+				{
+					$request->attributes->set('auto_item', $varValue);
+				}
+				else
+				{
+					$request->attributes->remove('auto_item');
+				}
+			}
+			else
+			{
+				$inputAttributes = $request->attributes->get('_contao_input', array());
+				$inputAttributes['setGet'][$strKey] = $varValue;
+				$request->attributes->set('_contao_input', $inputAttributes);
+			}
 		}
 
 		if ($varValue === null)
@@ -1058,6 +1077,11 @@ class Input
 	{
 		if ($request = static::getRequest())
 		{
+			if ('auto_item' === $strKey && $request->attributes->has('auto_item'))
+			{
+				return $request->attributes->get('auto_item');
+			}
+
 			$arrGet = $request->attributes->get('_contao_input')['setGet'] ?? array();
 
 			if (\array_key_exists($strKey, $arrGet))
--- a/core-bundle/contao/library/Contao/Input.php
+++ b/core-bundle/contao/library/Contao/Input.php
@@ -37,27 +37,6 @@ class Input
 	private static array $arrUnusedRouteParameters = array();
 
 	/**
-	 * Parameters set via setGet() are stored by request
-	 *
-	 * @var \WeakMap<Request,array<string,array|string>>
-	 */
-	private static \WeakMap $setGet;
-
-	/**
-	 * Parameters set via setPost() are stored by request
-	 *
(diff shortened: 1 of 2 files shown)


leofeyer:
I like this a lot. 👍

bytehead (dismissed):

core-bundle/contao/library/Contao/Input.php
```diff
@@ -330,10 +306,9 @@ public static function setGet($strKey, $varValue)
 
 		if ($request = static::getRequest())
 		{
-			static::$setGet ??= new \WeakMap();
-			$arrGet = static::$setGet->offsetExists($request) ? static::$setGet->offsetGet($request) : array();
-			$arrGet[$strKey] = $varValue;
-			static::$setGet->offsetSet($request, $arrGet);
+			$inputAttributes = $request->attributes->get('_contao_input', array());
```
ausi: Should we discuss how exactly we want to store this in the request attributes?

Currently an array ist stored in the `_contao_input` attribute, that has the keys `setGet`, `setPost` and `setCookie`.

fritzmg:
On another note: for simplicity I would prefer if we could retrieve the `auto_item` parameter more directly, e.g. via `$request->attributes->get('auto_item')` (or `$request->attributes->get('_contao_auto_item')`).

core-bundle/contao/library/Contao/Input.php
leofeyer: You mean if we want to have `_contao_input_get`, `_contao_input_post` and `_contao_input_cookies` instead?

ausi:
> for simplicity I would prefer if we could retrieve the `auto_item` parameter more directly, e.g. via `$request->attributes->get('auto_item')`

Done in b4515a4f6809ee4a5e40c4cc5d63893074ab214d

core-bundle/contao/library/Contao/Input.php
ausi: Yes or something completely different. But I think this is fine now as we have `$request->attributes->get('auto_item')` now 👍

Toflar (approved):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5296, opened 2022-09-20
Do not use exponential notation for float values

ausi:
Fixes #5294

PR #5301, opened 2022-09-21
Do not use exponential notation for float values

ausi:
Same as #5296 for Contao 5.0


core-bundle/contao/library/Contao/StringUtil.php
```diff
@@ -1097,4 +1097,36 @@ public static function revertInputEncoding(string $strValue): string
 
 		return $strValue;
 	}
+
+	public static function numberToString(float|int $number, int|null $precision = null): string
```
ausi, line 1101: Changed parameter type to `float|int` here and removed the extra type check.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5303, opened 2022-09-22
Remove the old .gif theme images

leofeyer:
Seems we have forgotten to remove them. 🙈


ausi:
~~Should we remove them in Contao 5.1? Otherwise every usage of these images in extensions would now suddenly fail.~~

fritzmg:
We should also mention that in the `DEPRECATED.md` of Contao 4.

leofeyer:
> Should we remove them in Contao 5.1? Otherwise every usage of these images in extensions would now suddenly fail.

I‘d prefer to remove them in Contao 5.0. There is no BC promise for themes, otherwise we would have to wait for Contao 6.

leofeyer:
I have added an `UPGRADE.md` entry.

Toflar (approved):

PR #5306, opened 2022-09-22
Use real placeholders in password fields

### LINKED ISSUE #5126: Do not show the password toggle if the value is *****
### Affected version(s)

4.13

### Description

As discussed [here ](https://community.contao.org/de/showthread.php?83546-BE-gt-Passworteinstellungen-gt-Passwort-best%C3%A4tigen-Eingabefeld-fehlt) the password show/hide toggler in the backend only shows the placeholder instead of the entered passwordphrase.

leofeyer:
Fixes #5126

### DIFF
--- a/core-bundle/src/Resources/contao/widgets/Password.php
+++ b/core-bundle/src/Resources/contao/widgets/Password.php
@@ -90,6 +90,24 @@ class Password extends Widget
 	}
 
 	/**
+	 * Return a parameter
+	 *
+	 * @param string $strKey The parameter key
+	 *
+	 * @return mixed The parameter value
+	 */
+	public function __get($strKey)
+	{
+		// If there is a password, the field is no longer mandatory
+		if ($strKey == 'mandatory')
+		{
+			return !$this->varValue;
+		}
+
+		return parent::__get($strKey);
+	}
+
+	/**
 	 * Validate input and set value
 	 *
 	 * @param mixed $varInput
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/widgets/Password.php
```diff
@@ -140,12 +158,17 @@ protected function validator($varInput)
...
+			unset($this->arrAttributes['required']);
+		}
+
 		return sprintf(
-			'<input type="password" name="%s" id="ctrl_%s" class="tl_text tl_password%s" value="%s" autocomplete="new-password"%s onfocus="Backend.getScrollOffset()">%s%s',
+			'<input type="password" name="%s" id="ctrl_%s" class="tl_text tl_password%s" value="" autocomplete="new-password"%s onfocus="Backend.getScrollOffset()">%s%s',
 			$this->strName,
 			$this->strId,
 			($this->strClass ? ' ' . $this->strClass : ''),
-			($this->varValue ? '*****' : ''),
```
ausi: Should we set a `placeholder="•••••"` if there is already a password set?
This would keep the feature that the user can see if the field is empty or not.

leofeyer: Added in cddbc12a750e8f7a3ac0927ecce6641f39d69caa. I left the five stars, though, because `•••••` does not quite generate the expected output.



Maybe we should use a real placeholder that says "Set a new password" or something alike?

leofeyer: I guess the stars are fine, aren‘t they?

ausi (approved):

Toflar (approved):

PR #5307, opened 2022-09-23
Add a confirmation message to forms and provide Ajax out of the box

qzminski:
I have added an option to define a confirmation message in the form settings. There is no longer a need to create a dedicated confirmation page – the confirmation will appear on the same page after a reload.

The redirect feature is still in place, however. Moreover, I have added the `{{form_message}}` insert tag, which can be used on the target page to display the confirmation message defined in the form settings.

Another significant change is that the form now works with ajax out of the box. This concept was taken from the well-known extension https://github.com/terminal42/contao-ajaxform

Problems to discuss:

1. The JavaScript file is included without any URL fingerprint, which can cause cache issues. Perhaps we should switch to Webpack Encore to generate the assets?
2. I am not entirely sure if the insert tags should be replaced using the `replace()` or `replaceInline()` method. @ausi can you shed some light on this?
3. Shall we use Twig templates here already?

I'm looking forward to your feedback!


Toflar:
Concept by yours truly 😇 I think it would be a great addition to the core, thanks for the PR @qzminski!

core-bundle/contao/library/Contao/InsertTags.php
```diff
@@ -181,7 +181,7 @@ static function ($allowedTag)
 			// Skip certain elements if the output will be cached
 			if ($blnCache)
 			{
-				if ($elements[0] == 'date' || $elements[0] == 'form_session_data' || $elements[0] == 'fragment' || ($elements[1] ?? null) == 'referer' || strncmp($elements[0], 'cache_', 6) === 0)
+				if ($elements[0] == 'date' || $elements[0] == 'form_session_data' || $elements[0] === 'form_confirmation_message' || $elements[0] == 'fragment' || ($elements[1] ?? null) == 'referer' || strncmp($elements[0], 'cache_', 6) === 0)
```
Toflar: @ausi is this okay until https://github.com/contao/contao/pull/4779 is ready?

core-bundle/contao/templates/forms/form_wrapper.html5
```diff
@@ -6,19 +6,15 @@
...
-    </div>
-  </form>
+  <?php if ($this->confirmation): ?>
+    <?php $this->insert('form_confirmation', $this->arrData) ?>
+  <?php else: ?>
+    <?php $this->insert('form_inline', $this->arrData) ?>
+  <?php endif; ?>
+
+  <?php if (!$this->action): ?>
+    <script src="bundles/contaocore/ajax-form.min.js"></script>
```
Toflar: This would need `$this->asset()` but yes, we don't have a `manifest.json` in the core-bundle yet.

core-bundle/contao/templates/forms/form_wrapper.html5
```diff
@@ -6,19 +6,15 @@
...
-        <?php endif; ?>
-      <?php endif; ?>
-      <?= $this->hidden ?>
-      <?= $this->fields ?>
-    </div>
-  </form>
+  <?php if ($this->confirmation): ?>
+    <?php $this->insert('form_confirmation', $this->arrData) ?>
+  <?php else: ?>
+    <?php $this->insert('form_inline', $this->arrData) ?>
```
Toflar, line 12: @m-vo we're adding a new template here so it would need to be Twig only. But how can we mix them here? 🤔

core-bundle/contao/library/Contao/InsertTags.php
```diff
@@ -704,6 +704,18 @@ static function ($allowedTag)
 					$arrCache[$strTag] = $attribute?->getValue()[$elements[1]] ?? null;
 					break;
 
+				// Form confirmation message
+				case 'form_confirmation_message':
+					if ($request?->getSession()->getFlashBag()->has(Form::SESSION_CONFIRMATION_KEY))
```
ausi, line 709: We should add a `$flags[] = 'attr';` before this line to ensure that the message gets HTML encoded correctly.

core-bundle/contao/forms/Form.php
```diff
@@ -560,10 +587,41 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields, $arrFi
...
-			$this->jumpToOrReload($objJumpTo->row());
+			$targetPageData = $objJumpTo->row();
+		}
+
+		// Set the confirmation message, if any
+		if ($this->objModel->confirmationMessage)
+		{
+			$confirmationMessage = $this->objModel->confirmationMessage;
+			$confirmationMessage = System::getContainer()->get('contao.string.simple_token_parser')->parse($confirmationMessage, $arrSubmitted);
+			$confirmationMessage = System::getContainer()->get('contao.insert_tag.parser')->replaceInline($confirmationMessage);
```
ausi: Should we replace this in reverse order to ensure that no insert-tag gets hacked into the message with the simple tokens?

core-bundle/contao/library/Contao/InsertTags.php
ausi: I think so, yes.

core-bundle/contao/forms/Form.php
qzminski: Not sure. On the other hand, the form data may be used in the insert tags like `{{email::##email##}}`.

core-bundle/contao/forms/Form.php
```diff
@@ -560,10 +587,41 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields, $arrFi
...
+
+			$requestStack = System::getContainer()->get('request_stack');
+			$request = $requestStack->getCurrentRequest();
+
+			// Throw the response exception if it's an AJAX request
+			if ($request && $request->isXmlHttpRequest() && $request->headers->get('X-Contao-Ajax-Form') === $this->getFormId() && $targetPageData === null)
+			{
+				$confirmationTemplate = new FrontendTemplate('form_confirmation');
+				$confirmationTemplate->setData($this->Template->getData());
+				$confirmationTemplate->confirmation = $confirmationMessage;
```
ausi:
```suggestion
				$confirmationTemplate->confirmation = StringUtil::specialchars($confirmationMessage);
```

core-bundle/contao/library/Contao/InsertTags.php
qzminski: I did it before, but since the confirmation message is an HTML (tinyMCE), it would output as follows:

```<p>Thank you for the message</p>```

instead of just

```Thank you for the message```

Due to HTML tags being encoded.

File: core-bundle/contao/forms/Form.php (hunk 6 above)
Comment (qzminski) [reply core-bundle/contao/forms/Form.php]: The confirmation message field contains HTML (tinyMCE).

File: core-bundle/contao/library/Contao/InsertTags.php (hunk 4 above)
Comment (ausi) [reply core-bundle/contao/library/Contao/InsertTags.php:709]: I thought it’s just text. In this case this should be fine as is, but we need to make sure that the HTML itself is safe.

File: core-bundle/contao/forms/Form.php (hunk 5 above)
Comment (ausi) [reply core-bundle/contao/forms/Form.php]: `$arrSubmitted` seems to be input-encoded so it should probably be fine. But for this uses case I think it would make sense to do something like `array_map(StringUtil::specialchars(...), $arrSubmitted)` to make this is safe even if a form widget is configured incorrectly.

File: core-bundle/contao/library/Contao/InsertTags.php (hunk 4 above)
Comment (qzminski) [reply core-bundle/contao/library/Contao/InsertTags.php:709]: How can we do that? I think it's filtered in the backend on save, isn't it?

File: core-bundle/contao/forms/Form.php (hunk 5 above)
Comment (qzminski) [reply core-bundle/contao/forms/Form.php]: Implemented in 185f2893.

File: core-bundle/contao/forms/Form.php [hunk 7]
```diff
@@ -560,10 +587,41 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields, $arrFi
...
 		{
-			$this->jumpToOrReload($objJumpTo->row());
+			$targetPageData = $objJumpTo->row();
+		}
+
+		// Set the confirmation message, if any
+		if ($this->objModel->confirmationMessage)
+		{
+			$confirmationMessage = $this->objModel->confirmationMessage;
+			$confirmationMessage = System::getContainer()->get('contao.string.simple_token_parser')->parse($confirmationMessage, array_map(array(StringUtil::class, 'specialchars'), $arrSubmitted));
```
ausi:
```suggestion
			$confirmationMessage = System::getContainer()->get('contao.string.simple_token_parser')->parse($confirmationMessage, array_map(StringUtil::specialchars(...), $arrSubmitted));
```

As Contao 5 requires PHP 8.1 we can use [first class callables](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) here ☺️

core-bundle/contao/library/Contao/InsertTags.php
ausi: The HTML from the backend should be save (or from a trusted user), I was more concerned about the input from the frontend, but this should be fine now with 185f28934d6e5f2bfa219b647a0f12631a57af0e I think.

ausi:
> I am not entirely sure if the insert tags should be replaced using the `replace()` or `replaceInline()` method. @ausi can you shed some light on this?

Using `replace()` could result in having `<esi>` tags in the output, `replaceInline()` ensures that everything is replaced completely without `<esi>` tags. So `replaceInline()` is correct here I think.

core-bundle/contao/templates/forms/form_wrapper.html5
m-vo: The only way we could render Twig templates here, would be by something like this: 
```suggestion
    <?php echo System::getContainer()->get('twig')->render('@Contao/form/inline.html.twig', $this->arrData)); ?>
```

And - depending on if `arrData` contains closures, it might be necessary to also adjust the context: 

```suggestion
    <?php echo System::getContainer()->get('twig')->render('@Contao/form/inline.html.twig', System::getContainer()->get('contao.twig.interop.context_factory')->fromData($this->arrData))); ?>
```

I don't really like this. :neutral_face:

Toflar: Agree, so for this use case, we'd introduce new html5 templates, agree?

ausi (dismissed):

core-bundle/contao/templates/forms/form_wrapper.html5
Toflar: @contao/developers we need to decide how to proceed with this one. Imho, this is the only blocker in this PR because we want to have cache busting hashes here.
Currently, we do not yet have any option to use `$this->asset('....', 'contao_core')`. This is, because currently, basically all of our FE scripts are `contao-components` which are registered as asset package here: https://github.com/contao/contao/blob/5.x/core-bundle/src/DependencyInjection/Compiler/AddAssetsPackagesPass.php#L88
And I agree this makes sense for stuff like `tinymce` because there are probably different versions of that component and you want to decide on which one to use by requiring the component yourself in your `composer.json`.
But here, we don't have multiple versions. This feature is part of the core and I don't see why this would be a replacable component? 
Imho, we should introduce a build process for FE assets in the core too and make sure, we have a `manifest.json` in `public` as well. Only if we have this, we'll be able to provide assets too, see https://github.com/contao/contao/blob/5.x/core-bundle/src/DependencyInjection/Compiler/AddAssetsPackagesPass.php#L67.

For that, we'd need probably need to introduce Webpack/Symfony Encore alongside our Gulp process. Wdyt?

fritzmg: You can get cache busting via

```php
<?= Contao\Template::generateScriptTag('bundles/contaocore/ajax-form.min.js', false, null) ?>
```

Toflar: I know, that's not the point though.

fritzmg: Yes, just pointing out the possibilities for _this_ PR. I would love for Contao to support a build process for FE assets out of the box.

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
...
+
+                form.ariaBusy = 'false';
+                form.dataset.ajaxForm = '';
+
+                const event = new CustomEvent('ajax-form-onload', {
+                    detail: { body, form, xhr },
+                });
+
+                form.dispatchEvent(event);
+                window.dispatchEvent(event);
```
ausi:
```suggestion
                    detail: { body, form, xhr },
                    bubbles: true,
                });

                form.dispatchEvent(event);
```

If we let the event bubble up the DOM tree we wouldn’t need to dispatch it twice I think.

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
...
+        function initForm(form) {
+            form.addEventListener('submit', e => {
+                e.preventDefault();
+
+                const formData = new FormData(form);
+
+                // Send the triggered button data as well
+                if (triggeredButton) {
+                    formData.append(triggeredButton.name, triggeredButton.value);
+                }
```
ausi:
```suggestion
                if (e.submitter) {
                    formData.append(e.submitter.name, e.submitter.value);
                }
```
(untested)

we could remove the whole handling of `triggeredButton` then.

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
...
+
+                request('POST', form.action, formData, xhr => {
```
ausi:
```suggestion
                request('POST', form, formData, xhr => {
```

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
+window.addEventListener('DOMContentLoaded', () => {
+    document.querySelectorAll('[data-ajax-form]').forEach(form => {
+        let triggeredButton = null;
+
+        function request(method, uri, body, callback) {
```
ausi:
```suggestion
        function request(method, form, body, callback) {
```

we need to pass the form here otherwise the reference is broken after the call to `form.replaceWith(newForm)`

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
...
+            const xhr = new XMLHttpRequest();
+
+            xhr.open(method, uri, true);
```
ausi:
```suggestion
            xhr.open(method, form.action, true);
```

core-bundle/public/ajax-form.js
```diff
@@ -0,0 +1,69 @@
+window.addEventListener('DOMContentLoaded', () => {
+    document.querySelectorAll('[data-ajax-form]').forEach(form => {
```
ausi: I’d prefer to rename the `form` parameter here to prefent confusion with the form parameter of nested functions.

qzminski:
@ausi I have added your suggestions in one commit: 82d802c2facd5eb8f002e8e50cbe57060f1408b0

core-bundle/contao/templates/forms/form_wrapper.html5
ausi: I’d also use the suggestion from https://github.com/contao/contao/pull/5307#discussion_r980017468 for this PR and create a new PR for the FE build process assets topic.

ausi (commented):
https://github.com/contao/contao/pull/5307#discussion_r978393568 needs to be addressed, rest approved ✅

core-bundle/contao/templates/forms/form_wrapper.html5
qzminski: Done in df13bae6.

Toflar (dismissed):

ausi (dismissed):

leofeyer:
So now forms are always submitted via Ajax unless there is a target URL? What if someone does not want this?

Toflar (commented):
If JS is disabled, it will just reload :)

qzminski:
Right now the only way to disable Ajax is to remove the JavaScript file from the `form_wrapper` template.

ausi:
> Right now the only way to disable Ajax is to remove the JavaScript file from the `form_wrapper` template.

Or by removing the `data-ajax-form` attribute from the `<form>`, right?

qzminski:
@ausi or that, yes, but then including JavaScript is pointless and just adds unnecessary kilobytes to download.

core-bundle/contao/templates/forms/form_wrapper.html5
```diff
@@ -6,19 +6,15 @@
...
+  <?php endif; ?>
+
+  <?php if (!$this->action): ?>
```
leofeyer: I would prefer this to be a checkbox in the form settings. It can be enabled by default, so that new forms are submitted via Ajax and existing ones are not.

qzminski: I have added the checkbox in [d19de43](https://github.com/contao/contao/pull/5307/commits/d19de432f73101bd3992710a116e041e357b73e0).

core-bundle/contao/dca/tl_form.php
```diff
@@ -261,6 +261,13 @@
 			'inputType'               => 'checkbox',
 			'eval'                    => array('tl_class'=>'w50'),
 			'sql'                     => array('type' => 'boolean', 'default' => false)
+		),
+		'ajax' => array
+		(
+			'filter'                  => true,
+			'inputType'               => 'checkbox',
+			'eval'                    => array('tl_class'=>'w50'),
+			'sql'                     => array('type' => 'boolean', 'default' => true)
```
ausi: I think we should add a migration that adds this column with a default values of `false` for all existing forms, so that only newly created forms get this activated by default.

fritzmg: But it only has an effect if you provided a message, does it not? Also imho it should simply not be `true` by default.

Toflar: I don't really mind. Can you guys agree on a way forward then, please? :)

m-vo: Wouldn't an SQL default value of `false` and a DCA default value of  `true` work to have BC and still the right value when creating new records? Or does this not work with booleans?

I'm fine with any version btw. 😌

ausi: @m-vo that would technically work, but the intentions would not be clear from reading the code. That is why we use migrations in these cases AFAIK.

> Also imho it should simply not be `true` by default.

I think I’d agree. And in this case we also don’t need a migration.

```suggestion
			'sql'                     => array('type' => 'boolean', 'default' => false)
```

ausi (dismissed):

leofeyer (changes requested):
The changes do not work for me. As soon as I have stored a confirmation message, the form is never shown again. Only the confirmation message.



Even if I reload the page (GET request), the form keeps showing the confirmation message.

core-bundle/contao/library/Contao/InsertTags.php
leofeyer: Can we maybe shorten the parameter name and insert tags? For example `form_message` and `$formMessage`?

qzminski: I have changed the insert tag to `{{form_message}}` in [2e6e817](https://github.com/contao/contao/pull/5307/commits/2e6e81781af40b4111db68285bf09cd12f9d89fc). I am not sure what parameter you mean for `$formMessage`, though.

qzminski:
> The changes do not work for me. As soon as I have stored a confirmation message, the form is never shown again. Only the confirmation message.

Is the `Redirect page` form setting empty in that case? For me it works as expected (I reload the page at the end):

https://user-images.githubusercontent.com/193483/196371268-11f56e78-35a8-42d9-9435-6becb2d6e3fa.mp4

core-bundle/contao/library/Contao/InsertTags.php
leofeyer: 😉

qzminski: Frankly, I think the `confirmationMessage` column name is more self-explanatory than just a `message`, because that – theoretically – could be of any type 🤔 Maybe then we could also change insert tag from `{{form_message}}` to `{{form_confirmation}}` …? What do you think?

leofeyer: I like `{{form_confirmation}}`. Can we rename the variable to `$this->objModel->confirmation` then as well or would this be too unspecific again?

qzminski: Changed in https://github.com/contao/contao/pull/5307/commits/70fb99778529cbb7bfb4711ce9469b625c887191 :+1:

core-bundle/contao/templates/forms/form_wrapper.html5
```diff
@@ -6,19 +6,15 @@
...
-    </div>
-  </form>
+  <?php if ($this->confirmation): ?>
+    <?php $this->insert('form_confirmation', $this->arrData) ?>
+  <?php else: ?>
+    <?php $this->insert('form_inline', $this->arrData) ?>
+  <?php endif; ?>
+
+  <?php if ($this->ajax): ?>
+    <script src="<?= $this->asset('ajax-form.min.js', 'contao_core') ?>"></script>
```
leofeyer: This is not ideal because if there are several forms on the page, the script will be loaded and executed multiple times. It should be added to `$GLOBALS['TL_BODY']` instead.

qzminski: I have added it to `$GLOBALS['TL_JAVASCRIPT']`.

leofeyer (changes requested):
Now that #5406 has been merged, this PR needs to be adjusted accordingly, doesn‘t it? @aschempp /cc

qzminski:
I have just pushed the commit that uses Webpack to generate the JS assets. Please review and confirm.

Toflar (dismissed):

ausi (dismissed):

leofeyer (changes requested):
The changes look all good, but the problem remains that as soon as there is a confirmation message, the form is never rendered in the front end (see https://github.com/contao/contao/pull/5307#pullrequestreview-1144858617).

fritzmg:
I can confirm the issue. The problem is that the DCA field for the confirmation message and the template variable have the same name. The template will automatically be filled with all the `tl_form` data here: 

https://github.com/contao/contao/blob/f21945658b216b0ce76c3682a20d9927070bda67/core-bundle/contao/classes/Hybrid.php#L228

And thus

https://github.com/contao/contao/blob/324cab3943239d84d2cbe687f5285218964c0edb/core-bundle/contao/templates/forms/form_wrapper.html5#L9

will always be true once a confirmation message has been set.

leofeyer:
I have fixed the issue in 176c393f08d3c6427c215e7c51343ff92dc75157.

leofeyer (approved):

leofeyer:
Thank you @qzminski.

PR #5310, opened 2022-09-25
Fix and adjust the template content element/module

m-vo:
As discussed in the last Contao call, this PR adds a Twig template for the `template` element/module. I also made sure the two controllers were in line again and added a note regarding the existing fragment controllers to the `UPGRADE.md`.


fritzmg (dismissed):

core-bundle/src/Controller/ContentElement/TemplateController.php
```diff
@@ -14,25 +14,28 @@
...
+            array_combine(
+                array_column($data, 'key'),
+                array_column($data, 'value')
+            )
         );
 
-        // Backwards compatibililty
-        $template->data = $data;
+        // Backwards compatibility
+        $template->set('data', $data);
```
ausi: Do we still need this backwards compatibility?

ausi (dismissed):

core-bundle/src/Controller/ContentElement/TemplateController.php
m-vo: I wasn't sure about this tbh. IIRC we intentionally left it in for 5.0 but then again we did not think about making the templates Twig only back then. Shall we remove it?

ausi: I’d remove it as the templates are not backwards compatible anyways.

core-bundle/src/Controller/ContentElement/TemplateController.php
```diff
@@ -14,25 +14,28 @@
...
 
         return $template->getResponse();
```
m-vo:
```suggestion
        return $template->getResponse();
```

core-bundle/src/Controller/FrontendModule/TemplateController.php
```diff
@@ -13,20 +13,29 @@
...
+        $template->set(
+            'keys',
+            array_combine(
+                array_column($data, 'key'),
+                array_column($data, 'value')
+            )
+        );
+
+        // Backwards compatibility
+        $template->set('data', $data);
```
m-vo:
```suggestion

```

core-bundle/src/Controller/ContentElement/TemplateController.php
m-vo: Done in b67226d

core-bundle/src/Controller/FrontendModule/TemplateController.php
```diff
@@ -13,21 +13,26 @@
...
+        $data = StringUtil::deserialize($model->data, true);
 
+        $template->set(
+            'keys',
+            array_combine(
+                array_column($data, 'key'),
+                array_column($data, 'value')
+            )
+        );
         return $template->getResponse();
```
m-vo, line 37:
```suggestion

        return $template->getResponse();
```

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #5309, opened 2022-09-25
Backport support for nested template paths

m-vo:
This PR backports #3973 and #4620 (nested template paths via `.twig-root`) for Contao 4.13. 

This makes it possible for extensions to embrace the new directory structures while being compatible with bot 4.13 and 5.

/cc @Toflar

PR #5315, opened 2022-09-27
Add a template finder service for Twig and fix custom template selection

m-vo:
This PR introduces a new template finder service and makes use of it in the `TemplateOptionsListener`:

```php
$options = $this->finderFactory
    ->create()
    ->identifier($identifier)
    ->withVariants()
    ->asTemplateOptions()
;
```

The service intentionally only considers Twig templates and follows the new naming scheme for variant templates (`foo/bar.html.twig` / `foo/bar/variant.html.twig`). This way there is no BC layer in the new class. I also removed the logic added in #5243 again as this case is now correctly handled by the new service.

The `TemplateOptionsListener` now differentiates and either uses the template finder or calls `Controller::getTemplateGroup()`. Once we migrated everything to the new structure we can deprecate the latter and finally remove the special handling in the listener in Contao 6.

The template finder should be easy to backport to 4.13. Together with #5309 extensions should then be able to work with the new structure while being forward compatible with Contao 5.

Todo:
* [x] manual testing
* [x] extension independent selections
    * [x] analyze
    * [x] update the filesystem loader/locator to support arbitrary extensions
* [x] check implications of #5254 / maybe revert parts of it

/cc @Toflar


Toflar (commented):
Looks awesome! I can certainly approve the concept. As already stated in the issue, I think the template finder should be able to be restricted to certain file endings.

Maybe something like this?

```
private string $fileEndingRegex = '.*`;

/**
 * Filters templates based on a file ending regex, e.g. "html\.twig" or "svg\.twig".
 */
public function fileEndiing(string $regex): self
{
    $this->fileEndingRegex = trim($regex, '$');
}
```

Or maybe even easier and just `str_ends_with()`? A custom filter can be implemented manually or we just add a `withFilter(Closure...` that overrides all the other settings?

m-vo:
>  I think the template finder should be able to be restricted to certain file endings

Agree - I had that planned already. :slightly_smiling_face:  Currently the thing identifying templates in Contao (e.g. `ce_foo` or `content_element/text`) does not include a file extension, so `foo.json.twig` and `foo.html.twig` both have the identifier `foo`. This is something I'd like to move away from in Twig but it's not easy with BC in mind.

What should be possible, though, and what I prepared in 2834c6c, is to allow arbitrary extensions in the hierarchy, but still grouped under extensionless identifiers. So a template `@Contao/notification_center/token_transformer.json.twig` will also be identified with `notification_center/token_transformer` and only one unique extension per identifier is allowed (so either `.html.twig` or `json.twig` for example). I have the feeling this limitation isn't really one, because I don't see why you would want to mix types. 

As a consumer you know which extension you need, so you would filter by `->extension('json.twig')` for the dropdown and add it again when rendering `->render("@Contao/$identifier.json.twig")`.

What still needs to be adjusted is our filesystem loader, so that the hierarchy will actually be populated with anything else than `html.twig` files.

Toflar:
I agree, I would go as far as throwing an exception if the template loader detects identical templates with different extensions. I don't think this should be a supported use case. The reason for it being exactly that in the dropdowns we never show extensions.

m-vo:
This should be ready for review now.

Besides the things mentioned in the initial description, these changes now also made it possible to drop the arbitrary HTML file type restriction (`.html.twig`). I also added/enhanced some exceptions for a better DX.

core-bundle/src/Controller/AbstractFragmentController.php
```diff
@@ -94,7 +94,17 @@ protected function createTemplate(Model $model, string|null $fallbackTemplateNam
...
                 $legacyTemplate->setData($template->getData());
 
-                $response = $legacyTemplate->getResponse();
+                try {
+                    $response = $legacyTemplate->getResponse();
+                } catch (\Exception $e) {
+                    // Enhance exception if a modern template name is defined,
+                    // but we still delegate to the legacy framework
+                    if (null !== ($definedTemplateName = $this->options['template'] ?? null) && preg_match('/^Could not find template "\S+"$/', $e->getMessage())) {
+                        throw new \LogicException(sprintf('Could neither find template "%s" nor the legacy fallback template "%s".', $definedTemplateName, $templateName), 0, $e);
```
Toflar: What I like about Symfony's exception messages is the "Did you mean ...?" part. Can we maybe provide guidance here too?

Toflar (dismissed):
Absolutely love the new service! If we get this backported to Contao 4.13, I think we now have everything in place to work nicely with new templates in both, Contao 4.13 LTS as well as 5.x.

On top of code review, I checked out the PR and tested the implementation on the current development branch of Notification Center 2.0. I can confirm that the new finder service works like a charm 👍

core-bundle/src/Controller/AbstractFragmentController.php
m-vo: What about adding something like `Did you forget to create a default template or manually define the "template" property of the controller's service tag/attribute?`

m-vo: See 17f6447538694e7260be434cbb41aa72ac7dcf0c

Toflar (dismissed):

fritzmg (changes requested):
This PR is missing the changes from https://github.com/contao/contao/pull/5254, thus legacy tepmlates of content elements that were changed back to its legacy versions will not be found (since it will always find the still registered fragment element's template).

fritzmg (changes requested):
Unfortunately this does not find PHP template variants of fragments. Consider the following example:

```php
// src/Controller/ContentElement/ExampleController.php

use Contao\ContentModel;
use Contao\CoreBundle\DependencyInjection\Attribute\AsContentElement;
use Contao\Template;

#[AsContentElement(category:'texts', template: 'ce_foobar')]
class ExampleController extends AbstractContentElementController
{
    protected function getResponse(Template $template, ContentModel $model, Request $request): Response
    {
    }
}
```
```php
// contao/dca/tl_content.php
$GLOBALS['TL_DCA']['tl_content']['palettes']['example'] = '{type_legend},type;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID;{invisible_legend:hide},invisible,start,stop';
```

* `contao/templates/ce_foobar.html5`
* `contao/templates/ce_foobar_test1.html5`
* `templates/ce_foobar_test2.html5`

The template options will only show `ce_foobar`, the other two templates will be missing.

m-vo:
Thanks for checking. I added those cases to the unit tests and fixed them. Hopefully we're good now. :grin:

fritzmg:
This unfortunately introduces a BC break. Previously you would have been able to do the following:

```php
// src/Controller/ContentElement/ExampleController.php

use Contao\ContentModel;
use Contao\CoreBundle\DependencyInjection\Attribute\AsContentElement;
use Contao\Template;

#[AsContentElement(category: 'texts')]
class ExampleController extends AbstractContentElementController
{
    protected function getResponse(Template $template, ContentModel $model, Request $request): Response
    {
    }
}
``` 
```html
<!-- contao/templates/ce_example.html5 -->
<p>Hello World!</p>
```
This works in Contao `^4.9` as well as `5.0.0` to `5.0.4`. However with this PR the following error would then appear:
```
Tried to list template options for the modern fragment type "content_element/example" but could not find any template. In case you wanted to use the legacy type "ce_example", define it explicitly in the "template" property of your controller's service tag/attribute.
```
So as of Contao `5.0.5` you would need to add `template: 'ce_example'` to your service tag. I personally would have expected that fragments continue to work as they have done before (since that is also our currently documented way).

However, the original intention was always for Contao 5 to switch to the new twig template structure and using legacy templates should be opt-in (by way of service tag properties). Thus it is actually a bug that the legacy templates still worked for fragments (without explicitly saying so).

Note that you would still be able to be compatible with both Contao ^4.9 and ^5.0 by explicitly defining the `template` in your service tag.

fritzmg (dismissed):

leofeyer:
As discussed in the Contao call, we want to merge this PR despite of the BC break.

Toflar (dismissed):

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5321, opened 2022-09-28
Fixed unknown opt-in token source

Toflar:
We have an `OptInInterface` as well as an `OptInTokenInterface` which allows me to decorate the opt in logic. This is a great helper for me while working on Notification Center 2.0.
However, I currently have no way to find out where that token is coming from. It could've been any module ID. I need to know which one in order to be able to only apply my logic in case of the correct module ID.

Basically, the whole decoration logic is useless if one cannot distinguish where the token is coming from so I consider this a bug.

Can anybody else test those changes too, please?

### DIFF
--- a/core-bundle/src/Resources/contao/modules/ModulePassword.php
+++ b/core-bundle/src/Resources/contao/modules/ModulePassword.php
@@ -185,14 +185,13 @@ class ModulePassword extends Module
 
 		// Find an unconfirmed token with only one related tl_member record
 		if (
-		    (!$optInToken = $optIn->find(Input::get('token')))
+			(!$optInToken = $optIn->find(Input::get('token')))
 			|| !$optInToken->isValid()
 			|| !isset($optInToken->getRelatedRecords()['tl_member'])
 			|| \count($arrRelated = $optInToken->getRelatedRecords()['tl_member']) != 1
 			|| \count($arrIds = current($arrRelated)) != 1
 			|| (!$objMember = MemberModel::findByPk($arrIds[0]))
-		)
-		{
+		) {
 			$this->strTemplate = 'mod_message';
 
 			$this->Template = new FrontendTemplate($this->strTemplate);
--- a/core-bundle/src/Resources/contao/modules/ModulePassword.php
+++ b/core-bundle/src/Resources/contao/modules/ModulePassword.php
@@ -183,8 +183,15 @@ class ModulePassword extends Module
 	{
 		$optIn = System::getContainer()->get('contao.opt_in');
 
-		// Find an unconfirmed token with only one related record
-		if ((!$optInToken = $optIn->find(Input::get('token'))) || !$optInToken->isValid() || \count($arrRelated = $optInToken->getRelatedRecords()) != 1 || key($arrRelated) != 'tl_member' || \count($arrIds = current($arrRelated)) != 1 || (!$objMember = MemberModel::findByPk($arrIds[0])))
+		// Find an unconfirmed token with only one related tl_member record
+		if (
+		    (!$optInToken = $optIn->find(Input::get('token')))
+			|| !$optInToken->isValid()
+			|| !isset($optInToken->getRelatedRecords()['tl_member'])
+			|| \count($arrRelated = $optInToken->getRelatedRecords()['tl_member']) != 1
+			|| \count($arrIds = current($arrRelated)) != 1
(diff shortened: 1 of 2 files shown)


core-bundle/src/Resources/contao/modules/ModulePassword.php
```diff
@@ -183,9 +183,15 @@ protected function setNewPassword()
...
 
-		// Find an unconfirmed token with only one related record
-		if ((!$optInToken = $optIn->find(Input::get('token'))) || !$optInToken->isValid() || \count($arrRelated = $optInToken->getRelatedRecords()) != 1 || key($arrRelated) != 'tl_member' || \count($arrIds = current($arrRelated)) != 1 || (!$objMember = MemberModel::findByPk($arrIds[0])))
-		{
+		// Find an unconfirmed token with only one related tl_member record
+		if (
+			(!$optInToken = $optIn->find(Input::get('token')))
+			|| !$optInToken->isValid()
+			|| !isset($optInToken->getRelatedRecords()['tl_member'])
+			|| \count($arrRelated = $optInToken->getRelatedRecords()['tl_member']) != 1
```
leofeyer, line 191:
```suggestion
			|| \count($arrRelated = $optInToken->getRelatedRecords()) != 1
```

Otherwise this would be a behavior change.

Previously `count($optInToken->getRelatedRecords())` had to be `1` but with your change there could be several related records as long as there is only one `tl_member` property.

core-bundle/src/Resources/contao/modules/ModuleRegistration.php
```diff
@@ -524,9 +524,15 @@ protected function activateAcount()
...
 
-		// Find an unconfirmed token with only one related record
-		if ((!$optInToken = $optIn->find(Input::get('token'))) || !$optInToken->isValid() || \count($arrRelated = $optInToken->getRelatedRecords()) != 1 || key($arrRelated) != 'tl_member' || \count($arrIds = current($arrRelated)) != 1 || (!$objMember = MemberModel::findByPk($arrIds[0])))
-		{
+		// Find an unconfirmed token with only one related tl_member record
+		if (
+			(!$optInToken = $optIn->find(Input::get('token')))
+			|| !$optInToken->isValid()
+			|| !isset($optInToken->getRelatedRecords()['tl_member'])
+			|| \count($arrRelated = $optInToken->getRelatedRecords()['tl_member']) != 1
```
leofeyer, line 532:
```suggestion
			|| \count($arrRelated = $optInToken->getRelatedRecords()) != 1
			|| !isset($arrRelated['tl_member'])
```

core-bundle/src/Resources/contao/modules/ModulePassword.php
Toflar: Yes, that's the point of this PR. It doesn't matter how many related records there are. It matters that there's only one `tl_member` :)

leofeyer (changes requested):
I don‘t your changes are correct. The related records are the ones that need to be published once the token is confirmed and that does not apply to `tl_module`. We will have to find another way to provide additional information in my opinion.

core-bundle/src/Resources/contao/modules/ModulePassword.php
leofeyer: No, I don‘t think so. We are looking for a token with one single related record that holds the member ID and nothing else here.

Toflar:
I see. You're the author of this service, what were the design decisions then?
My goal is to adjust the `$token->send()` process but only, if it was enabled in a given FE module. That's basically the thing I'm trying to do here. Any ideas?

leofeyer:
What about adding an `array $context = []` argument to the `OptIn::create()` method?

Toflar:
We have an interface, we can't add that.

Toflar:
Also, the context would need to be added to `find()` as well as `send()`. E.g. for `send()` one does not get the raw data (tokens) but only the text and subject which are already parsed :/

leofeyer:
> We have an interface, we can't add that.

We can add new optional arguments to the `OptIn` class. The interface can only be changed in Contao 6, of course.

Toflar:
The `OptIn` class also has an interface. This one's even tighter coupled to the implementation. It forces a subject and a text which is also tied to an e-mail. So I'll probably have to find a way to not use the opt in service at all. Which is a pity because it was probably meant to be decorated 🙈

PR #5325, opened 2022-09-29
Fix a potential PHP 8 warning in the backend help wizard

qzminski:
It can happen when a backend user opens the help wizard inside a Rocksolid custom content element:

CleanShot 2022-09-29 at 12 13 01

Although this is likely a misconfiguration in my RSCE, I think the warning should be prevented.

CleanShot 2022-09-29 at 12 11 44


bytehead (approved):

leofeyer:
Thank you @qzminski.

PR #5330, opened 2022-10-01
Add an exception for a missing administrator email address

fritzmg:
Having an administrator email address defined in the system settings is pretty important - otherwise you will not be able to send emails from forms or with the registration module for example:

```
Swift_RfcComplianceException:
Address in mailbox given [] does not comply with RFC 2822, 3.6.2.

  at vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\Headers\MailboxHeader.php:355
  at Swift_Mime_Headers_MailboxHeader->assertValidAddress('')
     (vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\Headers\MailboxHeader.php:272)
  at Swift_Mime_Headers_MailboxHeader->normalizeMailboxes(array(''))
     (vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\Headers\MailboxHeader.php:117)
  at Swift_Mime_Headers_MailboxHeader->setNameAddresses(array(''))
     (vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\Headers\MailboxHeader.php:74)
  at Swift_Mime_Headers_MailboxHeader->setFieldBodyModel(array(''))
     (vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\SimpleMimeEntity.php:600)
  at Swift_Mime_SimpleMimeEntity->setHeaderFieldModel('From', array(''))
     (vendor\swiftmailer\swiftmailer\lib\classes\Swift\Mime\SimpleMessage.php:213)
  at Swift_Mime_SimpleMessage->setFrom('')
     (vendor\contao\contao\core-bundle\src\Resources\contao\library\Contao\Email.php:458)
```

However, during the installation process of Contao it is not actually required to enter an administrator email address - so this might be missed completely and the error will not really tell you that the problem is a missing administrator email address.

This PR adds an exception to the `Email` class in case the administrator email address is not defined - and a `getSystemMessage` hook in order to show that the Contao installation is still in a misconfigured state.

This PR also adds a small CSS change so that links in error messages are actually identifiable as such:

image

⬇️ 

image


ausi (dismissed):

core-bundle/src/EventListener/AdministratorEmailListener.php
```diff
@@ -0,0 +1,62 @@
...
+    public function __construct(TranslatorInterface $translator, RouterInterface $router, RequestStack $requestStack)
+    {
+        $this->translator = $translator;
+        $this->router = $router;
+        $this->requestStack = $requestStack;
+    }
+
+    public function __invoke(): ?string
+    {
+        if (Config::get('adminEmail')) {
```
m-vo: Does this need an initialized framework here?

m-vo (dismissed):

core-bundle/src/EventListener/AdministratorEmailListener.php
fritzmg: I'd argue that for hooks you never really need that (as the framework already needs to be running for the hook to be executed).

fritzmg: All the other hook listeners in Contao that _do_ inject the `ContaoFramework` also only use it to retrieve an adapter - they do not initialize the framework themselves.

m-vo: Yeah, makes sense. 👍

leofeyer (changes requested):
This is a new feature, so please rebase it onto the `5.x` branch and change the milestone accordingly.

fritzmg:
> This is a new feature, so please rebase it onto the `5.x` branch and change the milestone accordingly.

I consider the fact that

* you cannot send emails when not providing an administrator email address
    (which as pointed out above is never set during the installation process)
* the error in the front end does not tell you what is wrong

a bug and this PR intends to fix that by telling you what is wrong.

fritzmg:
@leofeyer or do you only view the new `getSystemMessage` hook (plus its accompanying CSS change and additional translation label) as a feature? We should at the very least add the exception.

leofeyer:
Adding just the exception as a bug fix would be fine with me.

ausi (approved):

leofeyer:
Thank you @fritzmg.

leofeyer (approved):

aschempp:
we should consider adding the back end hint in Contao 5.0/5.x since the Contao Manager will not ask for the admin email (contrary to the install tool).

fritzmg:
> (contrary to the install tool).

The install too also will not ask for the administrator email address.

Other than that I wanted to create a PR for 5.x, yes.

ausi:
Just a quick Idea: can we fallback to `$_SERVER['SERVER_ADMIN']`?
Or is this something only apache has set but nginx hasn’t?

fritzmg:
On Hostingwerk for example `$_SERVER['SERVER_ADMIN']` is `[no address given]`, so I don't think we can/should use that.

fritzmg:
We could fall back to the email address of the first non-disabled admin back end user (order by ID ascending) - but that might be too arbitrary.

aschempp:
> The install too also will not ask for the administrator email address.

it adds the admin email from the first registered user, which is no longer the case with the Contao Manager.

fritzmg:
Huh, never noticed that before 😁. So then we could indeed implement [this fallback](https://github.com/contao/contao/pull/5330#issuecomment-1290580444) in Contao 5 at least (may be even 4.9+).

PR #5331, opened 2022-10-01
Do not show errors twice in console

ausi:
Fixes part of https://github.com/contao/contao-manager/issues/726

I think this needs to be fixed here directly because every usage of `--format=json` should never end up with invalid output due to printed PHP warnings. (And they still show up in STDERR)

But I’m not sure if we should instead just set `ini_set('display_errors', '0');` regardles of the `--format` option?

### DIFF
--- a/manager-bundle/bin/contao-console
+++ b/manager-bundle/bin/contao-console
@@ -17,13 +17,7 @@ use Symfony\Component\Console\Input\ArgvInput;
 
 set_time_limit(0);
 @ini_set('zlib.output_compression', '0');
-
-foreach ($_SERVER['argv'] as $arg) {
-    if (0 === strncmp($arg, '--format=', 9) && 0 !== strncmp($arg, '--format=txt', 12)) {
-        ini_set('display_errors', '0');
-        break;
-    }
-}
+@ini_set('display_errors', '0');
 
 if (file_exists(__DIR__.'/../autoload.php')) {
     $projectDir = \dirname(__DIR__, 2);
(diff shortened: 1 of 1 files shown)


m-vo:
Thoughts:

* What about using the `ArgvInput` class to check for the format option? It already parses the input; warnings before the autoloader is initialized are probably not worth hiding.

* Could we register a custom error handler instead of suppressing errors? So that they could maybe even returned as `"warnings": []` in the JSON output? (might be something for a custom command class, though)

aschempp:
does Symfony handle any of this when running `--format=json` commands? Not sure I like that the app silences this, it should probably be silenced by the caller …

m-vo:
> Not sure I like that the app silences this, it should probably be silenced by the caller …

I think I would also expect valid JSON output in case of calling it from a bash script for instance. Otherwise the shebang doesn't do anything anymore...

ausi:
The more I think about it the more I’m convinced that we should just set `ini_set('display_errors', '0');` always, regardless of the `--format` option.

With normal usage of `contao-console` in a shell the warnings show up anyway because `STDERR` gets printed by default. So currently I get every warning twice like for example:

```
$ vendor/bin/contao-console contao:migrate
PHP Warning:  Undefined array key "does-not-exist" in …/contao/config/config.php on line 3

Warning: Undefined array key "does-not-exist" in …/contao/config/config.php on line 3


 [OK] All migrations completed.
```

Updated the PR accordingly

ausi:
> does Symfony handle any of this when running `--format=json` commands?

I don’t think so, `vendor/bin/contao-console list --format=json` also outputs invalid JSON without this PR.

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #5333, opened 2022-10-03
Correctly handle invalid path in DC_Folder

aschempp:
I got this error in Sentry

> Deprecated: strtr(): Passing null to parameter #1 ($string) of type string is deprecated

Because calling `Validator::isInsecurePath` with `null`-value leads to invalid PHP method calls. I decided to fix the origin issue instead of changing `Validator::isInsecurePath` so we will also catch other invalid handling.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2974,7 +2974,7 @@ class DC_Folder extends DataContainer implements ListableDataContainerInterface,
 			throw new AccessDeniedException('Invalid file name "' . $strFile . '" (hacking attempt).');
 		}
 
-		if (Validator::isInsecurePath($strFolder))
+		if (null === $strFolder || Validator::isInsecurePath($strFolder))
 		{
 			throw new AccessDeniedException('Invalid folder name "' . $strFolder . '" (hacking attempt).');
 		}
--- a/core-bundle/src/Resources/contao/drivers/DC_Folder.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Folder.php
@@ -2974,7 +2974,7 @@ class DC_Folder extends DataContainer implements ListableDataContainerInterface,
 			throw new AccessDeniedException('Invalid file name "' . $strFile . '" (hacking attempt).');
 		}
 
-		if (null === $strFolder || Validator::isInsecurePath($strFolder))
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2974,7 +2974,7 @@ protected function isValid($strFile)
 			throw new AccessDeniedException('Invalid file name "' . $strFile . '" (hacking attempt).');
 		}
 
-		if (Validator::isInsecurePath($strFolder))
+		if (null === $strFolder || Validator::isInsecurePath($strFolder))
```
ausi: Doesn’t this change the behaviour? Because `Validator::isInsecurePath(null)` would be `false` I think.

leofeyer: We should adjust the `Validator::isInsecurePath()` method instead and return early if `$strPath` is not a string.

ausi: But in this case we should be careful what we consider to be “secure”. 
IMO every type other than `string|null` should be considered insecure by this method.

aschempp: The method is called `isValid` and should throw an exception if the file is not valid. I would argue that `null` is not a valid file (or folder)?

leofeyer: That‘s true…

ausi (changes requested):
With this pull request editing files or using the source code editor now always fails with an `Invalid folder name "" (hacking attempt).` error so it is completely broken now.

leofeyer:
Thanks for checking. It seems that the changes were not tested, were they?

core-bundle/src/Resources/contao/drivers/DC_Folder.php
ausi:
```suggestion
		if (null !== $strFolder && Validator::isInsecurePath($strFolder))
```

I think this would be the correct fix to get rid of the PHP warning.

leofeyer: @aschempp Are you still working on this PR?

ausi (approved):

PR #5337, opened 2022-10-05
Backport the localconfig file check to Contao 4.13

### LINKED ISSUE #5210: Misleading message in "SYSTEM -> Settings" in fresh Contao 5 installation
### Affected version(s)

5.0.1

### Description

After installing Contao 5.0.1 via the new Contao Manager v1.6.0, entering SYSTEM -> Settings shows this warning on top of the screen:

`The file "system/config/localconfig.php" is not writeable and can therefore not be updated!`

This is because the directory system/config does not exist at this time.

If this is the intended behaviour in Contao 5 the message should either be removed or say something different.

As a second note the "E-mail address of the system administrator*" is empty (which is not the case in Contao 4.13).

fritzmg:
Fixes #5210 for Contao 4.13.

Even in Contao 4 it is not strictly necessary to run the Install Tool once, especially if you use something like this in your application:

```yaml
# config/config.yaml
contao:
    localconfig:
        licenseAccepted: true
        installPassword: '%env(INSTALL_PASSWORD)%'
```

Hence even in Contao 4 the `system/config/localconfig.php` file might not exist yet when opening the back end system settings for the first time and thus the aforementioned error occurs.

This PR backports #5238 to Contao 4.13.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_File.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_File.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Symfony\Component\Filesystem\Path;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5344, opened 2022-10-06
Fix the LONG_TERM_SUPPORT constant

### LINKED ISSUE #5328: constants.php - LONG_TERM_SUPPORT
### Affected version(s)

4.13

### Description

Shouldn't it be "true" here?

https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/config/constants.php#L15

leofeyer:
Fixes #5328


bytehead:
better late than never 😬

bytehead (approved):

PR #5346, opened 2022-10-06
Ignore invalid jumpTo pages in the FAQ back end module

### LINKED ISSUE #4990: Creating/editing FAQ in category with deleted jumpTo page leads to error
### Affected version(s)

4.13.6

### Description

When a FAQ catagory has a `jumpTo` page defined, that is not longer valid (page deleted), creating a new or editing an existing entry in the category leads to an error.

This is an exception when creating the serp preview: [tl_faq.php line 596](https://github.com/contao/contao/blob/4.13/faq-bundle/src/Resources/contao/dca/tl_faq.php#L601)

### How to reproduce:
- create a page
- define the new page as jumpTo for a FAQ category
- delete the new page
- enter the FAQ category and try to create a new entry

leofeyer:
Fixes #4990

Now the FAQ module behaves the same as the calendar and new modules.

### DIFF
--- a/faq-bundle/src/Resources/contao/dca/tl_faq.php
+++ b/faq-bundle/src/Resources/contao/dca/tl_faq.php
@@ -16,6 +16,7 @@ use Contao\CoreBundle\Exception\AccessDeniedException;
 use Contao\DataContainer;
 use Contao\Date;
 use Contao\DC_Table;
+use Contao\Environment;
 use Contao\FaqCategoryModel;
 use Contao\FaqModel;
 use Contao\Input;
@@ -598,7 +599,7 @@ class tl_faq extends Backend
 
 		if (!$objTarget = PageModel::findByPk($jumpTo))
 		{
-			throw new Exception('Invalid jumpTo page: ' . $jumpTo);
+			return StringUtil::ampersand(Environment::get('request'));
 		}
 
 		$strSuffix = StringUtil::ampersand($objTarget->getAbsoluteUrl(Config::get('useAutoItem') ? '/%s' : '/items/%s'));


faq-bundle/src/Resources/contao/dca/tl_faq.php
```diff
@@ -598,7 +599,7 @@ public function getSerpUrl(FaqModel $objFaq)
 
 		if (!$objTarget = PageModel::findByPk($jumpTo))
 		{
-			throw new Exception('Invalid jumpTo page: ' . $jumpTo);
+			return StringUtil::ampersand(Environment::get('request'));
```
ausi, line 602: Is `Environment::get('request')` correct here? I think this is deprecated in Contao 5.0.

ausi: Oh, the events seem to do the same currently so this is consistent.

ausi (approved):

faq-bundle/src/Resources/contao/dca/tl_faq.php
leofeyer: Yes, but we are still using it a lot in Contao 5 and especially in the two other methods that generate URLs:

https://github.com/contao/contao/blob/4586b2f5b58075e5cfd076e1913cbf86dfd4337e/calendar-bundle/src/Resources/contao/classes/Events.php#L465-L468

https://github.com/contao/contao/blob/4586b2f5b58075e5cfd076e1913cbf86dfd4337e/news-bundle/src/Resources/contao/classes/News.php#L446-L449

PR #5349, opened 2022-10-07
Prevent loading of DCA column types in User class

ausi:
This is an attempt to fix https://github.com/contao/contao/issues/5029#issuecomment-1216531922

But I’m not sure if I’m happy with this solution 😐

### DIFF
--- a/core-bundle/contao/library/Contao/User.php
+++ b/core-bundle/contao/library/Contao/User.php
@@ -182,7 +182,27 @@ abstract class User extends System implements UserInterface, EquatableInterface,
 	 */
 	public function __get($strKey)
 	{
-		return $this->arrData[$strKey] ?? parent::__get($strKey);
+		if (!isset($this->arrData[$strKey]))
+		{
+			return parent::__get($strKey);
+		}
+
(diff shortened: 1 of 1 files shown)


m-vo:
This feels very hacky to honest. But I did not investigate too much, so I'm not sure there is a better solution :relaxed:.

IMHO the root of the problem is the `DcaExtractor` running without loading the language files first. Can't we fix this? Maybe we are overoptimizing by trying not to load unnecessary language files?

ausi:
The user object gets loaded very early (by the firewall), we don’t know the language at this point I think.

bytehead:
Isn't there a default language on the `Request` object in case there is no language set on the user?

ausi:
> Isn't there a default language on the `Request` object in case there is no language set on the user?

Yes, but this would mean that when using the debug mode all language files of all DCAs would get loaded in the wrong language, for every single request.

Maybe a better solution would be to fall back to the live schema from the database if the type cache is missing. This way we wouldn’t have to load the DCAs at all.

ausi:
> Maybe a better solution would be to fall back to the live schema from the database if the type cache is missing. This way we wouldn’t have to load the DCAs at all.

See #5428

PR #5350, opened 2022-10-08
Fix an "unknown system variable" error

ausi:
Running the migrate command with a MySQL 5.5 server currently results in `ER_UNKNOWN_SYSTEM_VARIABLE: Unknown system variable 'innodb_page_size'`

/cc @fritzmg

### DIFF
--- a/core-bundle/src/Doctrine/Schema/MysqlInnodbRowSizeCalculator.php
+++ b/core-bundle/src/Doctrine/Schema/MysqlInnodbRowSizeCalculator.php
@@ -85,7 +85,10 @@ class MysqlInnodbRowSizeCalculator
     {
         static $size = null;
 
-        return $size ??= (int) $this->connection->executeQuery('SELECT @@innodb_page_size / 2 - 66')->fetchOne();
+        return $size ??= (int) (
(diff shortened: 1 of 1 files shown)


fritzmg (dismissed):

core-bundle/src/Doctrine/Schema/MysqlInnodbRowSizeCalculator.php
```diff
@@ -85,7 +85,10 @@ public function getInnodbRowSizeLimit(): int
     {
         static $size = null;
 
-        return $size ??= (int) $this->connection->executeQuery('SELECT @@innodb_page_size / 2 - 66')->fetchOne();
+        return $size ??= (int) (
+            $this->connection->executeQuery("SHOW STATUS LIKE 'innodb_page_size'")->fetchAssociative()['Value']
+            / 2 - 66
+        );
```
leofeyer:
```suggestion
        if (null === $size) {
            $res = $this->connection->executeQuery("SHOW STATUS LIKE 'innodb_page_size'");
            $size = (int) ($res->fetchAssociative()['Value'] / 2 - 66);
        }

        return $size;
```

Just looking for a nicer way to format this. 🙈

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5351, opened 2022-10-09
Added some hints for pull requests

c4y:
If you do a composer update, since composer 2.1 no sources will be installed for dev versions. So you nee to add the option --prefer-install=auto

I would like a little more explanation of how a pull request works. So i thing "so you can submit your pull request directly
from within your application" is not the full truth :-) The way i described it here is how it works for me. I don't know if it is the right way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,14 +58,14 @@ Again, use `5.x-dev` if you want to add a new feature or `<lts-version>.x-dev` i
 Next, install the dependencies:
 
 ```bash
-composer update
+composer update --prefer-install=auto
 ```
 
 Composer will automatically clone the Git repo into the `vendor/contao/contao` folder. You can finish the setup by
-running `contao:setup` on the command line.
+running `vendor/bin/contao-setup` on the command line.
 
 All the changes you make in `vendor/contao/contao` are tracked via Git, so you can submit your pull request directly
-from within your application.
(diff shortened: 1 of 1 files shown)


aschempp (dismissed):

aschempp (approved):

README.md
```diff
@@ -58,14 +58,17 @@ Again, use `5.x-dev` if you want to add a new feature or `<lts-version>.x-dev` i
 Next, install the dependencies:
 
 ```bash
-composer update
+composer update --prefer-install=auto
```
leofeyer, line 61: I disagree with this change, because it would use `source` for every package with a dev version. The only dev package that we want is `contao/contao` itself, which already uses `source` due to the branch name (e.g. `4.13.x-dev` or `5.x-dev`).

README.md
```diff
@@ -58,14 +58,17 @@ Again, use `5.x-dev` if you want to add a new feature or `<lts-version>.x-dev` i
...
 Composer will automatically clone the Git repo into the `vendor/contao/contao` folder. You can finish the setup by
-running `contao:setup` on the command line.
+running `vendor/bin/contao-setup` on the command line.
 
 All the changes you make in `vendor/contao/contao` are tracked via Git, so you can submit your pull request directly
-from within your application.
+from within your application. To create a pull request, you first have to create a fork on Github. In your
+`vendor/contao/contao` folder you add a new git remote (`git add remote NAME FORK-URL`) to your fork. Now, you create a
+new branch (`git checkout -b fix/bug`) to fix the bug. After fixing the bug, push it to your fork (`git push NAME`) and
+create a pull request at Github.
```
leofeyer, line 71: I also disagree with this change, because it is not Contao‘s job to explain how to fork a repo and create a pull request on GitHub. A developer either knows already or they should learn it from docs.github.com or a similar, comprehensive documentation.

leofeyer (changes requested):

PR #5352, opened 2022-10-10
Fix `DC_Folder` ID not being determined correctly due to special characters

fritzmg:
Within `DC_Folder` the `id` parameter is not a numeric ID, but the path of the file. Entries from `tl_files` are retrieved via `tl_files.path`, rather than `tl_files.uuid` or `tl_files.id`. However, this will fail if the path contains special characters. So for a file like

```
files/foobar/moo(bar).jpg
```

The query in `Ajax::executePostActions` would thus be

```
SELECT * FROM tl_files WHERE path = 'files/foobar/moo&#41;bar&#92;.jpg';
```

resulting in a Bad Request.


ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5355, opened 2022-10-11
Backport the database configuration error checks

### LINKED ISSUE #4970: Database configuration errors missing from contao:migrate
### Affected version(s)

4.9, 4.13

### Description

The Contao Install Tool currently executes a few checks for the current database configuration and shows an error and what you will need to do, if applicable. These checks are missing from `contao:migrate` currently.

For example, if you configure something like:

```yaml
# config/config.yaml
doctrine:
    dbal:
        connections:
            default:
                default_table_options:
                    charset: utf8
                    collation: utf8_unicode_ci
```

but your database server does not support the `utf8` designation anymore (which is deprecated - it is now called `utf8mb3`), the install tool would usually tell you:

image

However, if you execute `contao:migrate --no-interaction` in such a case, it will run indefinitely, as it tries to convert `utf8mb3` to `utf8` over and over again.

I think we should add some of these checks to `contao:migrate` as well, at least the ones that can lead to a situation as mentioned instead of an exception.

m-vo:
Fixes #4970 

This PR backports the database configuration error checks in the `MigrateCommand` to Contao 4.13. They were originally introduced in #4935 (5.0) when removing the Install tool. I did only adjust the existing tests but did not port the tests for the error checks to minimize the amount of work needed when merging upstream.

### DIFF
--- a/core-bundle/src/Command/MigrateCommand.php
+++ b/core-bundle/src/Command/MigrateCommand.php
@@ -83,17 +83,35 @@ class MigrateCommand extends Command
     {
         $this->io = new SymfonyStyle($input, $output);
 
-        if (!$input->getOption('dry-run') && !$input->getOption('no-backup') && !$this->backup($input)) {
-            return 1;
-        }
-
-        if ('ndjson' !== $input->getOption('format')) {
-            return $this->executeCommand($input);
-        }
+        $asJson = 'ndjson' === $input->getOption('format');
 
         try {
+            if ($errors = $this->compileConfigurationErrors()) {
+                if ($asJson) {
+                    foreach ($errors as $message) {
+                        $this->writeNdjson('problem', ['message' => $message]);
+                    }
+                } else {
+                    foreach ($errors as $error) {
+                        $this->io->block($error, '!', 'fg=yellow', ' ', true);
+                    }
+
+                    $this->io->error('The database server is not configured properly. Please resolve the above issue(s) and rerun the command.');
+                }
+
+                return 1;
+            }
+
+            if (!$input->getOption('dry-run') && !$input->getOption('no-backup') && !$this->backup($input)) {
+                return 1;
+            }
+
             return $this->executeCommand($input);
         } catch (\Throwable $exception) {
+            if (!$asJson) {
+                throw $exception;
+            }
+
             $this->writeNdjson('error', [
                 'message' => $exception->getMessage(),
                 'code' => $exception->getCode(),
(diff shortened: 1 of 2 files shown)


ausi (dismissed):

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -83,17 +83,35 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                    $this->io->error('The database server is not configured properly. Please resolve the above issue(s) and rerun the command.');
+                }
+
+                return 1;
+            }
+
+            if (!$input->getOption('dry-run') && !$input->getOption('no-backup') && !$this->backup($input)) {
+                return 1;
+            }
+
```
leofeyer, line 108: In Contao 5, this whole block is outside of the try/catch block. Why is it different here?

core-bundle/src/Command/MigrateCommand.php
```diff
@@ -83,17 +83,35 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
             return $this->executeCommand($input);
         } catch (\Throwable $exception) {
+            if (!$asJson) {
+                throw $exception;
+            }
+
```
leofeyer, line 114: This condition does not exist in Contao 5. Does it need to be added when merging upstream?

core-bundle/tests/Command/MigrateCommandTest.php
```diff
@@ -597,11 +590,42 @@ private function getCommand(array $pendingMigrations = [], array $migrationResul
...
+        $connection = $this->createMock(Connection::class);
+        $connection
+            ->method('fetchOne')
+            ->willReturnCallback(
+                static function (string $query) use ($sqlMode) {
+                    switch ($query) {
+                        case 'SELECT @@sql_mode':
+                            return $sqlMode;
+                        case 'SELECT @@version':
+                            return '8.0.0';
```
leofeyer, line 613:
```suggestion

                        case 'SELECT @@version':
                            return '8.0.0';
```

core-bundle/src/Command/MigrateCommand.php
m-vo: What code are you referring to? Except for the returns being integers instead of constants, the whole `execute()` function is identical to the one in [Contao 5.0](https://github.com/contao/contao/blob/5.0/core-bundle/src/Command/MigrateCommand.php#L67-L110).

core-bundle/src/Command/MigrateCommand.php
m-vo: see https://github.com/contao/contao/pull/5355/files#r994335495

core-bundle/src/Command/MigrateCommand.php
leofeyer: True. It seems I was comparing to Contao 4.13. 🤦‍♂️

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5359, opened 2022-10-12
Change the default value for tl_layout.viewport

### LINKED ISSUE #4436: Set default Viewport-tag
### Description

https://docs.contao.org/manual/en/layout/theme-manager/manage-page-layouts/#expert-settings

The FE output
`<meta name="viewport" content="width=device-width,initial-scale=1.0">`
currently only occurs when

- responsive.css is selected or
- the "Viewport-tag" entry is set manually

It would be easier & clearer if the "viewport-tag" entry is set by default with: 
`width=device-width, initial-scale=1.0`

leofeyer:
Fixes #4436

### DIFF
--- a/core-bundle/contao/dca/tl_layout.php
+++ b/core-bundle/contao/dca/tl_layout.php
@@ -230,7 +230,7 @@ $GLOBALS['TL_DCA']['tl_layout'] = array
 			'search'                  => true,
 			'inputType'               => 'text',
 			'eval'                    => array('decodeEntities'=>true, 'maxlength'=>255, 'tl_class'=>'w50'),
-			'sql'                     => "varchar(255) NOT NULL default ''"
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

ausi (approved):

aschempp:
Wouldn't `shrink-to-fit=no` cause any page that is **not** response to be basically inaccessible on mobile?

fritzmg:
I think you are confusing this with `user-scalable=no`.

aschempp:
no, `user-scalable` means a user cannot zoom in our out of the viewport. If I open a non-responsive website on a mobile, I will get a "zoomed out" view of everything on the page (because it will "shrink to fit" the view). I would assume this is no longer the case?

fritzmg:
> If I open a non-responsive website on a mobile, I will get a "zoomed out" view of everything on the page

If you have a non-responsive website, then you will need to adjust the viewport in any case, because then even `width=device-width,initial-scale=1.0` will be wrong.

PR #5364, opened 2022-10-14
Allow defining a default search field

### LINKED ISSUE #4594: Optimize default value for search filter by pagetree
### Description

It would be helpful if there were new default settings for a value which is used more often, e.g. title as default.
Similar with the articles and content elements

Currently:
- Page = Assets-URL
- Article = Article alias
- Content elements = Alternative text

I would like to see:
- Page = page title
- Article = title
- Content element = text or heading

leofeyer:
Implements #4594

### DIFF
--- a/core-bundle/contao/drivers/DC_Table.php
+++ b/core-bundle/contao/drivers/DC_Table.php
@@ -5418,7 +5418,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 				$option_label = \is_array($GLOBALS['TL_LANG']['MSC'][$field]) ? $GLOBALS['TL_LANG']['MSC'][$field][0] : $GLOBALS['TL_LANG']['MSC'][$field];
 			}
 
-			$options_sorter[$option_label . '_' . $field] = '  <option value="' . StringUtil::specialchars($field) . '"' . ((isset($session['search'][$this->strTable]['field']) && $session['search'][$this->strTable]['field'] == $field) ? ' selected="selected"' : '') . '>' . $option_label . '</option>';
+			$options_sorter[$option_label . '_' . $field] = '  <option value="' . StringUtil::specialchars($field) . '"' . (((isset($session['search'][$this->strTable]['field']) && $session['search'][$this->strTable]['field'] == $field) || (!isset($session['search'][$this->strTable]['field']) && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['defaultSearchField'] ?? null) == $field)) ? ' selected="selected"' : '') . '>' . $option_label . '</option>';
 		}
 
 		// Sort by option values
--- a/core-bundle/contao/drivers/DC_Table.php
+++ b/core-bundle/contao/drivers/DC_Table.php
@@ -5418,7 +5418,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 				$option_label = \is_array($GLOBALS['TL_LANG']['MSC'][$field]) ? $GLOBALS['TL_LANG']['MSC'][$field][0] : $GLOBALS['TL_LANG']['MSC'][$field];
 			}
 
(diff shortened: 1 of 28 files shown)


core-bundle/contao/drivers/DC_Table.php
```diff
@@ -5418,7 +5418,7 @@ protected function searchMenu()
 				$option_label = \is_array($GLOBALS['TL_LANG']['MSC'][$field]) ? $GLOBALS['TL_LANG']['MSC'][$field][0] : $GLOBALS['TL_LANG']['MSC'][$field];
 			}
 
-			$options_sorter[$option_label . '_' . $field] = '  <option value="' . StringUtil::specialchars($field) . '"' . ((isset($session['search'][$this->strTable]['field']) && $session['search'][$this->strTable]['field'] == $field) ? ' selected="selected"' : '') . '>' . $option_label . '</option>';
+			$options_sorter[$option_label . '_' . $field] = '  <option value="' . StringUtil::specialchars($field) . '"' . (((isset($session['search'][$this->strTable]['field']) && $session['search'][$this->strTable]['field'] == $field) || (!isset($session['search'][$this->strTable]['field']) && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['defaultSearchField'] ?? null) == $field)) ? ' selected="selected"' : '') . '>' . $option_label . '</option>';
```
ausi:
```suggestion
			$options_sorter[$option_label . '_' . $field] = '  <option value="' . StringUtil::specialchars($field) . '"' . ((($session['search'][$this->strTable]['field'] ?? $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['defaultSearchField'] ?? null) == $field) ? ' selected="selected"' : '') . '>' . $option_label . '</option>';
```

Simplified?

aschempp:
Contrary to the discussion in https://github.com/contao/contao/issues/4594 this seems not to store the last selected field in the session, which I still think makes a lot of sense...

core-bundle/contao/drivers/DC_Table.php
leofeyer: Did you test this? Because the original logic is

```php
if (isset($session['search'][$this->strTable]['field']) && $session['search'][$this->strTable]['field'] == $field) {
    // case A
} elseif (!isset($session['search'][$this->strTable]['field']) && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['defaultSearchField'] ?? null) == $field) {
    // case B
}
```

and I‘m not sure your code does the same. It might though… 😄 

Here is what you should test:

1. Reset the session data in the back end profile menu.
2. Open the "Pages" module. → The search field should be `pageTitle`.
3. Change the search field to `description`, enter a value and submit. → The search field should be `description`.
4. Remove the value but leave the search field on `description` and submit. → The search field should still be `description`, because the session value overrides the default value.

leofeyer:
You are wrong. The last selected field is always stored in the session ever since Contao 1.

core-bundle/contao/drivers/DC_Table.php
ausi: Yes, works the same way.
_(but I’m also fine if you want to keep the code as it is...)_

ausi (dismissed):

ausi (approved):

aschempp:
> You are wrong. The last selected field is always stored in the session ever since Contao 1.

Please try that in the current demo. If you clear the filters, the first search field will be selected.

leofeyer:
Of course. Clearing the filters resets the search menu to its default state. And that is correct IMHO.

aschempp:
Latest comment in #4594 

> I think we could just keep the last searched field in the session. Instead of dropping the value from session when filters are cleared, we could set an empty value (string or null) so we know what field was searched for.

leofeyer:
I disagree with that. If the search menu is reset, I want it to be reset to the default search field and not the last one stored in the session.

Issue #3821, opened 2021-12-14
Prevent password spraying in ModuleLogin


aschempp:
As discussed in Mumble on June 6th, we need to track IP addresses in a table and block them if there are too many invalid login attempts. The number should not be too low (e.g. 20).

We might also be able to use a rate limit bundle for Symfony, although we have to ensure that it only limits failed login attempts and not a large number of successful login attempts.

_Originally posted by @leofeyer in https://github.com/contao/contao/issues/1770#issuecomment-638876316_

m-vo:
We should keep in mind that IP != user entity, especially for organizations that share a single gateway.

bytehead:
I'll assign this to me and check what Symfony Security or possible bundles do offer here.

SeverinGloeckle:
I had given some thought to this a few weeks ago.

As far as I know, there is no out-of-the-box solution for password spraying. However, Symfony 5.2 has introduced a [rate limiter](https://symfony.com/doc/current/rate_limiter.html) that can be used very well for this task.


From my point of view, the following points are important:
- The rate limiter works on failed authentications with one bucket per invalid password (hashed for security).
- If the authentication is successful(!), the bucket is checked - if it is empty, the authentication is invalid.
- It is not visible externally whether the authentication failed because of an incorrect password or because a limit was exceeded.

Using this method, the IP address is irrelevant.

Optimally, the rate limiter can be easily adjusted (e.g. number of tokens per bucket) and also deactivated. An allowlist would also be a good idea.

If you need support, I'm happy to help and/or provide a draft.

aschempp:
Sorry I don't understand the concept. How does a bucket per invalid **password** help, and why is the authentication invalid if the bucket is empty but the password correct?

SeverinGloeckle:
Let's first distinguish between two possible attacks:

- Password spraying: One (or more) common passwords are spammed across many different account names.
- Credential stuffing: A list of common combinations of usernames and passwords (or combinations that have become known through a data breach) are spammed.

Since in both cases the attacks are spread over many different accounts, the existing protection for accounts is not effective (e.g. blocking one account after 3 unsuccessful login attempts).

In the case of **password spraying**, the password can be used to detect related login attempts. For example, if the password _12345678_ has been used 5 times in a short period of time for invalid login attempts, password spraying can be detected.

[...]

PR #5368, opened 2022-10-17
Enable the login rate limit

bytehead:
Implements https://github.com/contao/contao/issues/3821#issuecomment-1008708462

Enabling the the [DefaultLoginRateLimiter](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php)  using the [Fixed Window Rate Limiter](https://symfony.com/doc/current/rate_limiter.html#fixed-window-rate-limiter) policy.


manager-bundle/skeleton/config/security.yaml
```diff
@@ -19,6 +19,10 @@ security:
             contao_login:
                 remember_me: false
 
+            login_throttling:
+                max_attempts: 5
+                interval: 1 minute
```
ausi:
```suggestion
            login_throttling: ~
```

According to the docs this should use the default of 5 per minute. ☺️

bytehead: Correct, I thought to make it more obvious where to change what.

ausi:
Shouldn’t we remove our own logic then so that these mechanisms don’t interfere with each other?
https://github.com/contao/contao/blob/09696f510b318cd94c594d9aa83a49d040c2a9f6/core-bundle/src/Security/User/UserChecker.php#L50

bytehead:
Hmm, good question. Our mechanism works only for existing accounts so the Symfony rate limiter covers both (existing and non-existing).

leofeyer:
Yes, we should remove our own logic IMHO.

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #5371, opened 2022-10-17
Allow to set the locale in `Template::trans`

### LINKED ISSUE #4014: Add locale parameter for translations in templates
### Description

With #4013 we'll now be able to request translations for different languages to do something like this:

```php
dump( $this->translator->trans('MSC.target.0', [], 'contao_default','de') ); // "In neuem Fenster öffnen"
dump( $this->translator->trans('MSC.target.0', [], 'contao_default','en') ); // "Open in new window"
dump( $this->translator->trans('MSC.target.0', [], 'contao_default','fr') ); // "Ouvrir dans une nouvelle fenêtre"
dump( $this->translator->trans('MSC.target.0', [], 'contao_default','it') ); // "Si apre in una nuova finestra"
```

This is all fine for modules and controllers using the Translator service.

If we need translations of different languages within a template like this...

```php
dump( $this->trans('MSC.target.0', [], 'contao_default','de') );
dump( $this->trans('MSC.target.0', [], 'contao_default','en') );
dump( $this->trans('MSC.target.0', [], 'contao_default','fr') );
dump( $this->trans('MSC.target.0', [], 'contao_default','it') );
```
... it won't work because the trans method of the Template class lacks the `$locale` parameter

https://github.com/contao/contao/blob/9e0dceaa6660feddeb36608035ad79cd6e6ba136/core-bundle/src/Resources/contao/library/Contao/Template.php#L382-L394

With the Twig-Support in the works shouldn't there be something possible like this?

```twig
{{ 'MSC.target.0'|trans([],'contao_default','de') }}
{{ 'MSC.target.0'|trans([],'contao_default','en') }}
{{ 'MSC.target.0'|trans([],'contao_default','fr') }}
{{ 'MSC.target.0'|trans([],'contao_default','it') }}
```
> Dicslaimer: I haven't worked with Twig that much so there is a good chance my syntax is total BS. Maybe @m-vo could give it a look? 👀

Long story short: Having the "core" translation stuff fixed with #4013 I'd appreciate if we could also make sure that different lanuages work for the Twig integration as well as using the trans method of the Template class.

fritzmg:
Implements #4014 

This PR allows you to pass the `$locale` parameter when using `Template::trans` and - through decorating the new [Symfony Locale Switcher](https://github.com/symfony/symfony/pull/45793) - also makes sure that `$GLOBALS['TL_LANG']` contains the correct translations afterwards. So for example:

```php
<!-- templates/foobar.html5 -->
en: <?= $this->trans('ERR.general') ?>
de: <?= $this->trans('ERR.general', [], 'contao_default', 'de') ?>
cs: <?= $this->trans('ERR.general', [], 'contao_default', 'cs') ?>
en: <?= $GLOBALS['TL_LANG']['ERR']['general'] ?>
en: <?= $this->trans('ERR.general') ?>
```

yields

```
en: An error occurred!
de: Ein Fehler ist aufgetreten!
cs: Došlo k chybě!
en: An error occurred!
en: An error occurred!
```

However, performance wise it would still be better to query multiple translation strings at once and then reset the requested domains manually, e.g.

```php
$locale = $this->localeSwitcher->getLocale();
$translations = $this->localeSwicher->runWithLocale('de', function () use ($data, $locale) {
    $translations = [
        'moobar1' => $this->translator->trans('moobar1', [], 'contao_default'),
        'moobar2' => $this->translator->trans('moobar2', [], 'contao_default'),
        'foobar1' => $this->translator->trans('foobar1', [], 'contao_exception'),
        'foobar2' => $this->translator->trans('foobar2', [], 'contao_exception'),
    ];

    System::loadLanguageFile('default', $locale);
    System::loadLanguageFile('exception', $locale);

    return $translations;
});
```


fritzmg:
Not sure why the prefer lowest tests fail with

```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "contao.translation.locale_switcher" has a dependency on a non-existent service "translation.locale_switcher".
```

as I am requiring

```json
"symfony/translation": "^6.1"
```

fritzmg:
In hindsight I think we do not need the `LocaleSwitcher` at all. In Twig you can also force the locale via the `trans` Twig function for example, so I think we should just do the same for our `Template::trans` method.

The issue that remains is that `$GLOBALS['TL_LANG']` will potentially be populated with the wrong translations. Which is true regardless of whether you do this in PHP templates or Twig templates. Thus I think every developer needs to make sure that _if_ they force the locale in their templates for `contao_` translation domains, they need to make sure to repopulate the correct translations via `System::loadLanguageFile` again after rendering the template. 

We could of course also introduce an automatic reset in the `trans` method of our `Translator`, however that would be executed for every `->trans` call with a forced local (that is different from the default one) thus I don't think we should do that.

@ausi wdyt?

ausi:
> The issue that remains is that `$GLOBALS['TL_LANG']` will potentially be populated with the wrong translations.

I think I could live with that if we deprecate the usage of `$GLOBALS['TL_LANG']` in templates.

fritzmg:
> I think I could live with that if we deprecate the usage of `$GLOBALS['TL_LANG']` in templates.

It's not just templates. Imagine a template is rendered where you use a translation in a forced locale. And after that a news _module_ is generated, which references `$GLOBALS['TL_LANG']` directly quite often in PHP (_not_ the template). Then these translations will be wrong.

ausi:
Can we deprecate `$GLOBALS['TL_LANG']` everywhere then?
Or everywhere except translation files themselfes and DCA files?

Toflar:
I think if we can make sure that `$GLOBALS['TL_LANG']` always contains the default/current request language. Our translator could do something like this:

```php
if ($desiredLocale !== $currentLocale) {
    // Backup $GLOBALS
}

System::loadLanguageFile(...);
// Do whatever we do now

if ($desiredLocale !== $currentLocale) {
    // Restore $GLOBALS
}

return $translation;
```

ausi:
Backing up and restoring `$GLOBALS['TL_LANG']` without breaking the references to it is not trivial I think. It would also break the internal cache of `System::loadLanguageFile`. And performance wise probably not that different to just re-loading the original language.

Toflar:
And what's the issue with just reloading it? It only needs to be done in case the locale differs so it shouldn't be a super common case?

ausi:
> And what's the issue with just reloading it? It only needs to be done in case the locale differs so it shouldn't be a super common case?

If I would render a template with serveral translations in a different language, every single `trans()` call would then result in two complete reloads of the affected language files.

fritzmg:
Nevertheless, I have simplified this PR (it now only implements passing the `$locale` variable along to the translator in the `Template::trans` method).

The PRs #5434 and #5435 would then implement the automatic restoration of the `$GLOBALS['TL_LANG']` translations whenever a custom locale was requested.

> If I would render a template with serveral translations in a different language, every single trans() call would then result in two complete reloads of the affected language files.

In Symfony 6.1 you could use the `LocaleSwitcher` then in your template to improve performance.

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5379, opened 2022-10-17
Use sendfile for local files downloads

m-vo:
This PR is a follow-up to https://github.com/contao/contao/pull/4862 that does the following two things:

1) ~~We're now using a real route for downloads (like initial tried). Instead of checking permissions there, we only do it when generating (and signing) the download URL. I added a timestamp and made sure that URLs expire after 15min in case anyone revoked access to a resource in the meantime.~~

1) Local files will now utilize sendfile (via `BinaryFileResponse`) instead of streaming which allows to use large files and handle them by the server instead of the PHP process.

### DIFF
--- a/core-bundle/src/Controller/ContentElement/DownloadsController.php
+++ b/core-bundle/src/Controller/ContentElement/DownloadsController.php
@@ -71,9 +71,9 @@ class DownloadsController extends AbstractContentElementController
             $request,
             $this->filesStorage,
             function (FilesystemItem $item, array $context): Response|null {
-                // Only allow URIs that aren't older than 15 minutes
-                if ($context['time'] - time() > 600) {
-                    return new Response('The resource ');
+                // Only allow URIs that aren't older than 24 hours
+                if ($context['time'] - time() > 86400) {
+                    return new Response('The download URL has expired.');
                 }
 
                 // Make sure the file still exists
--- a/core-bundle/src/Controller/ContentElement/DownloadsController.php
+++ b/core-bundle/src/Controller/ContentElement/DownloadsController.php
@@ -15,7 +15,6 @@ namespace Contao\CoreBundle\Controller\ContentElement;
 use Contao\Config;
 use Contao\ContentModel;
 use Contao\CoreBundle\DependencyInjection\Attribute\AsContentElement;
-use Contao\CoreBundle\Exception\ResponseException;
 use Contao\CoreBundle\File\MetadataBag;
 use Contao\CoreBundle\Filesystem\FileDownloadHelper;
 use Contao\CoreBundle\Filesystem\FilesystemItem;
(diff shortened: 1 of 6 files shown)


core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -36,31 +35,62 @@
...
-        $this->handleDownload($request);
+        $this->initializeContaoFramework();
+
+        return $this->fileDownloadHelper->handle(
+            $request,
+            $this->filesStorage,
+            function (FilesystemItem $item, array $context): Response|null {
+                // Only allow URIs that aren't older than 15 minutes
+                if ($context['time'] - time() > 600) {
+                    return new Response('The resource ');
```
ausi: This line seems to be incomplete ☺️

ausi:
> I added a timestamp and made sure that URLs expire after 15min

Wouldn’t this render download pages uncachable now?

Should we use a more forgiving timeframe like 24 hours or so instead? 
and set cache headers so that the HTML pages with the links expire soon enough?

I’d still like to move from the signed URLs to checking the actual permissions for this in the future (but this future might be far away).

Toflar:
Caching is indeed an issue. The downloads only need to be signed if they are protected, right? In which case we don't have the problem of caching...

m-vo:
Interesting… I didn't think of caching. I'm not sure we can differentiate here. You would normally always get streaming, because the file could be located on a remote filesystem.

Toflar:
The check is here to protect the following case: If you copy the link and you get the permissions removed later on, you still have access to the file. Am I right? What about instead of having the time in the URL, use some member permission hash? Or so? Again, if you're not logged in, the check is irrelevant. Or what did I miss?

m-vo:
Bummer, reverted the first approach again, because we unfortunately won't be able to fix this for now. :-/

m-vo:
Memo to myself: We could probably enable `autoEtag` or set one based on the already known file hash.

ausi (dismissed):

core-bundle/src/Filesystem/FileDownloadHelper.php
```diff
@@ -115,14 +117,22 @@ public function handle(Request $request, VirtualFilesystemInterface $storage, \C
...
-                    $storage->readStream($file->getPath()),
-                    fopen('php://output', 'w')
-                );
-            }
-        );
+        $stream = $storage->readStream($file->getPath());
+        $metadata = stream_get_meta_data($stream);
+
+        // Prefer sendfile for local resources
+        if ('STDIO' === $metadata['stream_type'] && 'plainfile' === $metadata['wrapper_type'] && Path::isAbsolute($localPath = $metadata['uri'])) {
```
Toflar, line 124: Should we maybe extract that in a static util class? Or on the VFS maybe? Afaik it's duplicated here and in the `FigureBuilder`? /cc @ausi

Toflar (dismissed):

core-bundle/src/Filesystem/FileDownloadHelper.php
m-vo: IMHO we are going to drop it in the `FigureBuilder` as soon as `contao/image` is capable of streaming.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5399, opened 2022-10-17
Draft of URL rewrites

aschempp:
This merges the basic features of our [terminal42/contao-url-rewrite](https://github.com/terminal42/contao-url-rewrite) extension into the core. It will allow users to add custom URL rewrite rules, as well as the core possibility to e.g. add an automatic rewrite if a page is deleted or an alias is changed.

We have discussed to not use the `tl_page` here because the page tree (mostly) represents a navigational structure, but url rewrites are not really related to a navigation.

**TODOs**:
- [ ] limit the routes from DB search based on the current URL
- [x] implement the `getRouteByName(s)` methods
- [x] ~add a 410 error page controller that generates a message if a page is gone~ render the 404 page for 410 errors
- [ ] write tests

### DIFF
--- a/core-bundle/config/controller.yaml
+++ b/core-bundle/config/controller.yaml
@@ -176,3 +176,9 @@ services:
             - '@contao.routing.page_registry'
         tags:
             - controller.service_arguments
+
+    Contao\CoreBundle\Controller\UrlRewriteController:
+        arguments:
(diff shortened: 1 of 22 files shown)


asaage:
Nice...
I like contao-url-rewrite but if tracking and adjusting/redirecting alias-changes will not be part of it why merge it with the core?

PR #5404, opened 2022-10-18
Add a token authenticator

bytehead:


core-bundle/src/Security/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
+    {
+        return null;
+    }
+
+    public function onAuthenticationFailure(Request $request, AuthenticationException $exception): ?Response
+    {
+        return new JsonResponse([
```
aschempp: maybe we should check the `Accept` header here

aschempp (commented):
might also need to update the security config in the README

aschempp:
be aware of https://github.com/contao/contao/issues/6220 before we finish this 🙈

core-bundle/src/Command/AccessTokenRemoveExpiredCommand.php
```diff
@@ -0,0 +1,44 @@
...
+        private readonly AccessTokenRepository $accessTokenRepository,
+    ) {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $this->accessTokenRepository->removeExpired();
```
aschempp, line 38: shouldn't this be a cron job instead? I can't imagine anyone ever running that command?

core-bundle/src/Security/AccessTokenHandler.php
```diff
@@ -0,0 +1,187 @@
...
+     * @param non-empty-string $secret
+     */
+    public function __construct(
+        private readonly AccessTokenRepository $accessTokenRepository,
+        private readonly Connection $connection,
+        private readonly HttpClientInterface $httpClient,
+        string $secret,
+        private Configuration|null $configuration = null,
+    ) {
+        $this->configuration = $configuration ?: Configuration::forSymmetricSigner(new Sha256(), InMemory::plainText($secret));
```
aschempp, line 47:
```suggestion
        $this->configuration ??= Configuration::forSymmetricSigner(new Sha256(), InMemory::plainText($secret));
```

core-bundle/src/Security/AccessTokenHandler.php
```diff
@@ -0,0 +1,187 @@
...
+        $this->configuration->setValidationConstraints(new SignedWith($this->configuration->signer(), $this->configuration->signingKey()));
```
aschempp, line 48: do we want to set this if the constructor parameter is already initialized?

core-bundle/src/Security/AccessTokenHandler.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public function createTokenForUser(string $username): string
+    {
+        $this->accessTokenRepository->removeExpired();
```
aschempp, line 53: this seems unrelated to `createTokenForUser`?

core-bundle/src/Security/AccessTokenHandler.php
```diff
@@ -0,0 +1,187 @@
...
+        return $builder
+            ->issuedAt(new \DateTimeImmutable())
+            ->expiresAt(new \DateTimeImmutable('now +10 seconds'))
+            ->getToken($this->configuration->signer(), $this->configuration->signingKey())
+        ;
+    }
+
+    public function parseToken(string $data): UnencryptedToken|null
+    {
+        $parser = new Parser(new JoseEncoder());
```
aschempp, line 81: comparing with https://github.com/contao/contao/blob/5.x/manager-bundle/src/HttpKernel/JwtManager.php#L114, shouldn't this use the parser from the configuration?

core-bundle/src/Security/AccessTokenHandler.php
```diff
@@ -0,0 +1,187 @@
...
+
+        return $username === $claims->get('username');
+    }
+
+    public function getUserBadgeFrom(string $accessToken): UserBadge
+    {
+        try {
+            $token = $this->accessTokenRepository->findByToken($accessToken);
+        } catch (TokenNotFoundException $e) {
+            throw new AuthenticationException('Token not found.');
```
aschempp, line 118:
```suggestion
            throw new AuthenticationException('Token not found.', 0, $e);
```

aschempp (commented):
While reviewing I have found some fundamental questions:
 1. why is there a `tl_access_token` table if we have a (stateless) JWT?
 2. is it intended that I can only authenticate in the back end using the `Authentication` header, but in the front end only using the `access_token` query parameter?
 3. also, why do we need the query parameter at all (since we are doing the HttpClient magic to get a new session cookie)?
 4. how can the JWT/table only contain the username? In the current situation, a token generated for the front end user `foobar` would also be valid for the back end user `foobar`?
 5. can't we disable 2FA for token authentication instead of bypassing the validation?

Toflar:
> 3\. also, why do we need the query parameter at all (since we are doing the HttpClient magic to get a new session cookie)?

This is likely foundational work for the preview switch to a different domain :)

aschempp:
I just re-thought this query parameter afer the JWT session at SymfonyCon, and I think we need to solve this differently. If we really need cross-domain authentication, we should use the URL hash part and probably need an iframe or thelike to execute javascript. But we sure need to put more thoughts into that.

aschempp:
Stubled onto this by accident: https://symfony.com/doc/current/security/login_link.html
Can't we just use login links (similar to what is sent by email) for cross-domain logins?

bytehead:
This is somewhat hard out of sync.

PR #5405, opened 2022-10-18
Introduce background workers

Toflar:
This introduces real background workers based on cronjobs and Symfony Messenger.

## Motivation

Modern applications often require background workers in order to move more heavy processes in the background. We already have things in place that should better be done asynchronously in order to speed up the front end:

- Sending e-mails via Symfony Mailer
- Search indexing

This PR also opens up the opportunity for extension developers to e.g. put other heavy workloads in the background. Here are some ideas:

- Preparing huge ZIP files (combining lots of files in the BE anyone?)
- Creating complex reports
- Sending lots of things to a slow external API
- etc.

It's also relevant for us, if we want to work on a more complex search index because it might (we don't know yet) increase the time it takes to index a page for example.

## Concept

The concept builds on top of Symfony Messenger. This means you can technically route your messages to RabbitMQ or any other transport. Contao ships with the Doctrine based transport by default which means there are 0 additional requirements.
In order to make this work, however, we need real Cronjobs. Which we don't want to make a system requirement because for simple/small sites, we probably don't need real cronjobs. Also, we don't need them for simple development/testing instances etc.

Here's how it works from a developer's perspective (default Contao Managed Edition):

- We have three priorities: High, Medium, Low. As a developer of an extension, you don't need to do **anything at all** in terms of configuration. All you need to do is prepare your message, implement one of the priority interfaces (`LowPriorityMessageInterface`, `NormalPriorityMessageInterface` and `HighPriorityMessageInterface`) and pass it to the message bus (`MessageBusInterface`). Then you're good to go and build your message handler. Contao takes care of the rest.
- Taking care of the rest means it automatically detects if there is a worker running for the specific queue. If so, messages are dispatched to the configured transport (Doctrine by default). If not, messages are processed synchronously immediately.
- In terms of the `SearchIndexListener`, for example, this means that nothing changes at all if you do not have any worker running. Contao detects that there's no worker and indexes synchronously on every request like it does today. If, however, you configured a real worker, things will automatically happen asynchronously in the back end, making your front end considerably faster and freeing up your FPM childs etc.

Here's how it works from a Contao user's perspective (default Contao Managed Edition):

- Running workers and having the benefit of asynchronous processes is **as easy as** configuring, a real, minutely cronjob calling our [cronjob framework (`contao:cron`) as documented](https://docs.contao.org/manual/en/performance/cronjobs/). Done, Contao does the rest - magic! 🥳 
- Even better: Contao even does **autoscaling**! When there are more messages to be processed on the queue, more workers are started automatically. But this too is managed by Contao completely 😎 

Technical details and remarks:

- Workers are `messenger:consume` commands and started via [our new async cronjobs support](https://github.com/contao/contao/pull/5554).
- The implementation is completely independent from our cron job framework. The default configuration for the Managed Edition looks like this and I have some remarks to it:

```yaml
framework:
    messenger:
        transports:
            sync: sync://
            contao_auto_fallback_prio_high: contao_auto_fallback://contao_prio_high?fallback=sync
            contao_auto_fallback_prio_normal: contao_auto_fallback://contao_prio_normal?fallback=sync
            contao_auto_fallback_prio_low: contao_auto_fallback://contao_prio_low?fallback=sync
            contao_prio_high: doctrine://default?table_name=tl_message_queue&queue_name=prio_high&auto_setup=false
            contao_prio_normal: doctrine://default?table_name=tl_message_queue&queue_name=prio_normal&auto_setup=false
            contao_prio_low: doctrine://default?table_name=tl_message_queue&queue_name=prio_low&auto_setup=false
        routing:
            'Symfony\Component\Mailer\Messenger\SendEmailMessage': contao_auto_fallback_prio_low
            'Contao\CoreBundle\Messenger\Message\HighPrioMessageInterface': contao_auto_fallback_prio_high
            'Contao\CoreBundle\Messenger\Message\NormalPrioMessageInterface': contao_auto_fallback_prio_normal
            'Contao\CoreBundle\Messenger\Message\LowPrioMessageInterface': contao_auto_fallback_prio_low

# Contao configuration
contao:
    messenger:
        console_path: '%kernel.project_dir%/vendor/bin/contao-console'
        workers:
            -
                transports:
                    - contao_prio_high
                options:
                    - --time-limit=60
                    - --sleep=5
                autoscale:
                    desired_size: 5
                    max: 10
            -
                transports:
                    - contao_prio_normal
                options:
                    - --time-limit=60
                    - --sleep=10
                autoscale:
                    desired_size: 10
                    max: 10
            -
                transports:
                    - contao_prio_low
                options:
                    - --time-limit=60
                    - --sleep=20
                autoscale:
                    desired_size: 20
                    max: 10
```

As you can see, we configure the routing for the Symfony Messenger in such a way that, our **interfaces** map to one of our three queues (`contao_prio_low`, `contao_prio_normal` and `contao_prio_high`). This has a major advantage: As a developer, you don't need to register any config. You can just implement the desired interface and that's it. So you can have

```php
class FoobarMessage implements NormalPriorityMessageInterface
{
}
```

and

```php
$this->messageBus->dispatch(new FoobarMessage()).
```

and that's it. Contao automatically does the rest for you and ensures that if there is no worker running, the `sync` transport will be used. So your message is processed - either in the optimized way when workers are running - or not.

Using the interface is a very nice defaul. It provides the mentioned DX and **it still allows manual configuration adjustment** if needed. So you could addjust your `framework.messenger.routing` config like so, if you like:

```yaml
framework:
    messenger:
        routing:
            'App\Messenger\FoobarMessage': my_other_queue
```

The concrete class name always wins over the interface implemented.

Even the transport logic is completely independent from Contao. The automatic fallback to sync is implemented by a custom `contao_auto_fallback` which you are free to use. So instead of using the Contao worker framework and auto detection magic, you can disable it all and configure Contao the way you like. Or you can combine as you like:

```yaml
# This would map all messages implementing our default interfaces to the same
# RabbitMQ queue and our workers are disabled. So you are responsible to run
# `messenger:consume rabbit_mq ...` yourself using supervisord or whatever you prefer.

framework:
    messenger:
        transports:
            rabbit_mq: amqp://guest:guest@localhost:5672/%2f/messages
        routing:
            'Contao\CoreBundle\Messenger\Message\HighPriorityMessageInterface': rabbit_mq
            'Contao\CoreBundle\Messenger\Message\NormalPriorityMessageInterface': rabbit_mq
            'Contao\CoreBundle\Messenger\Message\LowPriorityMessageInterface': rabbit_mq

contao:
    messenger:
        workers: []
```

```yaml
# This would map all messages implementing our default interfaces to the same RabbitMQ
# queue if it is running. If Contao does not detect any worker running, it will fallback to 
# the sync transport. Again you are responsible to run `messenger:consume rabbit_mq ...`
# yourself using supervisord or whatever you prefer but in this case, if they do not run,
# Contao falls back to sync.

framework:
    messenger:
        transports:
            sync: sync://
            contao_auto_fallback_rabbit_mq: contao_auto_fallback://rabbit_mq?fallback=sync
            rabbit_mq: amqp://guest:guest@localhost:5672/%2f/messages
        routing:
            'Contao\CoreBundle\Messenger\Message\HighPriorityMessageInterface': contao_auto_fallback_rabbit_mq
            'Contao\CoreBundle\Messenger\Message\NormalPriorityMessageInterface': contao_auto_fallback_rabbit_mq
            'Contao\CoreBundle\Messenger\Message\LowPriorityMessageInterface': contao_auto_fallback_rabbit_mq

contao:
    messenger:
        workers: []
```

Maximum flexibility! 😎 

To finish, let's look at our default workers configuration in the Managed Edition:

```yaml
contao:
    messenger:
        console_path: '%kernel.project_dir%/vendor/bin/contao-console'
        workers:
            -
                transports:
                    - contao_prio_high
                options:
                    - --time-limit=60
                    - --sleep=5
                autoscale:
                    desired_size: 5
                    max: 10
            -
                transports:
                    - contao_prio_normal
                options:
                    - --time-limit=60
                    - --sleep=10
                autoscale:
                    desired_size: 10
                    max: 10
            -
                transports:
                    - contao_prio_low
                options:
                    - --time-limit=60
                    - --sleep=20
                autoscale:
                    desired_size: 20
                    max: 10
```

As you can see, I've configured autoscaling for all 3 queues but I've configured them in a way, that there are never more than `10` workers per queue. Also, you can see that `--sleep` differs. Important note: `sleep` does **not** define the sleep time between **all** messages. Instead, it only sleeps when **idle**. This means it only sleeps e.g. `20` seconds if the last time it asked, there was **no message**. If there was one, it will ask for the next message immediately after it finished processing. Hence, I configured them to `5`, `10` and `20` seconds. In other words: In worst case, it takes 5 seconds for a message to be processed if it was on the "high prio" queue and 20 on the "low prio".

Pretty proud of this implementation, I hope it will serve us well and allow for some cool new use cases for extensions 😎


core-bundle/contao/config/default.php
```diff
@@ -100,7 +100,6 @@
 $GLOBALS['TL_CONFIG']['backendTheme']         = 'flexible';
 $GLOBALS['TL_CONFIG']['doNotCollapse']        = false;
 $GLOBALS['TL_CONFIG']['minPasswordLength']    = 8;
-$GLOBALS['TL_CONFIG']['disableCron']          = false;
```
Toflar: This config is obsolete and not needed anymore. We have a zero config approach. Contao automatically detects if you have a real cron, otherwise you get the poor man's fallback.

core-bundle/src/EventListener/DoctrineSchemaListener.php
```diff
@@ -30,5 +32,19 @@ public function __construct(private DcaSchemaProvider $provider)
     public function postGenerateSchema(GenerateSchemaEventArgs $event): void
     {
         $this->provider->appendToSchema($event->getSchema());
+
+        foreach (['contao_prio_high', 'contao_prio_medium', 'contao_prio_low'] as $transportName) {
```
Toflar: Hardcoded here. I think if you define additional transports yourself, you should also add them to the schema yourself.

core-bundle/src/Search/Document.php
```diff
@@ -244,4 +244,22 @@ private function filterJsonLdContexts(array $data, array $contexts): array
 
         return $found ? $newData : [];
     }
+
+    public function __serialize(): array
```
Toflar: Required because Symfony Messenger serializes objects by default for the transport. Without explicit methods, PHP would try to serialize `DOM...` objects which is not supported.

manager-bundle/skeleton/config/config.yaml
```diff
@@ -31,10 +31,26 @@ framework:
...
+        transports:
+            sync: 'sync://'
+            contao_prio_high: 'contao_cron_fallback://contao_prio_high_doctrine?fallback=sync'
+            contao_prio_normal: 'contao_cron_fallback://contao_prio_normal_doctrine?fallback=sync'
+            contao_prio_low: 'contao_cron_fallback://contao_prio_low_doctrine?fallback=sync'
+            contao_prio_high_doctrine: 'doctrine://default?table_name=contao_queue&queue_name=prio_high&auto_setup=false'
+            contao_prio_normal_doctrine: 'doctrine://default?table_name=contao_queue&queue_name=prio_normal&auto_setup=false'
+            contao_prio_low_doctrine: 'doctrine://default?table_name=contao_queue&queue_name=prio_low&auto_setup=false'
+        routing:
+            'Symfony\Component\Mailer\Messenger\SendEmailMessage': contao_prio_low
```
Toflar, line 44: We can now send e-mails asynchronously for free 🥳

manager-bundle/skeleton/config/config.yaml
```diff
@@ -31,10 +31,26 @@ framework:
                 adapter: cache.app
             doctrine.system_cache_pool:
                 adapter: cache.system
+    messenger:
```
Toflar, line 34: The way to read this configuration:

- 3 interfaces map to 3 transports
- These 3 transports are configured to use our own `contao_auto_fallback` transport. All of them are configured to use the doctrine transports and fallback to `sync`.

core-bundle/src/Cron/AbstractConsoleCron.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Contao\CoreBundle\Cron;
+
+use Symfony\Component\Process\PhpExecutableFinder;
+use Symfony\Component\Process\Process;
+
+abstract class AbstractConsoleCron
```
Toflar: Helpful foundation for other crons that want to trigger a child process.

core-bundle/src/Messenger/Transport/CronFallbackTransport.php
```diff
@@ -0,0 +1,64 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Messenger\Transport;
+
+use Contao\CoreBundle\Cron\Cron;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Transport\TransportInterface;
+
+class CronFallbackTransport implements TransportInterface
```
Toflar: This is the one that simply checks if we have a real cron job for every call. If so, we can use the configured target (Doctrine by default), if not -> fallback (sync).

core-bundle/src/Cron/MessengerCron.php
```diff
@@ -0,0 +1,71 @@
...
+
+            // Start the job asynchronously
+            $process->start();
+            $processes[] = $process;
+        }
+
+        // Now we need to sleep to keep the parent process open. Otherwise, this script will end and thus kill
+        // our child processes.
+        // All jobs run for 60 seconds, so we don't need to check every second yet
+        sleep(55);
```
Toflar: Needs to use the `ClockInterface` to be released with Symfony 6.2 in about a month ;)

Toflar:
Notes for myself:

- The current implementation would block other cronjobs. Async crons should be a top level feature.
- Could make the number of workers dynamic based on `MessageCountAwareInterface` 😎

aschempp (commented):
I don't think I'm able to fully review this without any messenger knowledge, but it looks good in general. One thought though: 

Why do we need the full worker configuration in the Contao bundle? I mean we are setting up three priority queues for devs that "just want it to work". If you have other needs, use your own transport. Why would I ever reconfigure the workers? Also, if I want to run workers differently, because I _might_ have special needs, wouldn't I just want to disable the cron jobs and run the actual `messenger:consume` command with my priority etc?

Toflar:
> Why do we need the full worker configuration in the Contao bundle? I mean we are setting up three priority queues for devs that "just want it to work". If you have other needs, use your own transport. Why would I ever reconfigure the workers? Also, if I want to run workers differently, because I _might_ have special needs, wouldn't I just want to disable the cron jobs and run the actual `messenger:consume` command with my priority etc?

Not sure I understand all the questions here but let me try.

> Why do we need the full worker configuration in the Contao bundle? I mean we are setting up three priority queues for devs that "just want it to work". If you have other needs, use your own transport. Why would I ever reconfigure the workers?

Why do we need it? Because Contao uses it itself as of this PR. The `SearchIndexListener` does no longer index right away but instead sends the new `SearchIndexMessage` to the message bus. Which means that if no workers are running, it will be processed immediately and thus the same as today. But if workers do run, it will happen in the background.

> Why would I ever reconfigure the workers?

You mean reconfigure the `routing`? Because maybe you want the `SearchIndexMessage` to be treated as high priority. Then you can adjust that easily by providing a specific entry for the `SearchIndexMessage`.

> Also, if I want to run workers differently, because I _might_ have special needs, wouldn't I just want to disable the cron jobs and run the actual `messenger:consume` command with my priority etc?

I think you're mixing up things. "Priority" is not a thing that exists in Symfony Messenger. It's just a different transport and we make it a priority thing because `prio_high` is pinged more often and more volatile in terms of autoscaling configuration than `prio_low`.
Also, there is no "disable the cron jobs and run the actual `messenger:consume` command". We do run the actual `messenger:consume` commands as well. We just do it via a minutely cron job which is super handy for all shared hostings etc. that will certainly not allow you to configure your own workers with supervisord or whatever.

aschempp:
I don't think that's what I meant 😂 I meant why do we need the stuff in `contao` bundle configuration. Isn't it "just" used to configure the messenger queues? So if I _really **desperately**_ need to change the default Contao queues, can't I "just override the messenger configuration that is applied by default from Contao?

Toflar:
Sorry but I have absolutely no clue what you're talking about.

core-bundle/src/Cron/MessengerCron.php
```diff
@@ -0,0 +1,130 @@
...
+        $workerPromises = [];
+
+        foreach ($this->workers as $worker) {
+            $this->addWorkerPromises($worker, $workerPromises);
+        }
+
+        return Utils::all($workerPromises);
+    }
+
+    public function setPromiseFactory(callable $callable): self
```
ausi: Is this method only used for testing?
If so, would it make sense to refactor `ProcessUtil` into a service so that it can be mocked instead and this method not needed anymore then?

Toflar: Yes. None of our `Util` classes are currently a service. Do you think I can just make it one anyway?

ausi: I don’t know if we would have to rename it to something not Util then. But a service seams to be a better fit here I think.

Toflar: Converted in https://github.com/contao/contao/pull/5405/commits/6d24eda65108ea84ab72987028a32df7b3889788.

core-bundle/src/Cron/MessengerCron.php
```diff
@@ -0,0 +1,106 @@
...
+    private function addWorkerPromises(array $worker, array &$workerPromises): void
+    {
+        // Always add one worker
+        $workerPromises[] = $this->createProcessPromiseForWorker($worker);
+
+        if ($worker['autoscale']['enabled']) {
+            $totalMessages = $this->collectTotalMessages($worker['transports']);
+            $desiredWorkers = ceil($totalMessages / $worker['autoscale']['desired_size']);
+
+            // Never more than the max (-1 because we've already started one)
```
ausi:
```suggestion
            // Never more than the max
```
this is done two lines below with its own comment.

core-bundle/src/Messenger/Transport/AutoFallbackTransportFactory.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+    }
+
+    public function createTransport(string $dsn, array $options, SerializerInterface $serializer): TransportInterface
+    {
+        if (false === $parsedUrl = parse_url($dsn)) {
+            throw new InvalidArgumentException(sprintf('The given Auto Fallback DSN "%s" is invalid.', $dsn));
+        }
+
+        $target = isset($parsedUrl['path']) ? substr($parsedUrl['path'], \strlen('contao_auto_fallback://')) : '';
```
ausi: The underscore is an invalid scheme character, because of this the `path` does not strip the protocol part here automatically.

I think we have to change the scheme name to `contao-auto-fallback` to make it a valid DSN. _(The Symfony example also uses a dash, like `my-transport://`)_

```suggestion
        $target = $parsedUrl['path'] ?? '';
```

core-bundle/src/Messenger/MessageHandler/SearchIndexMessageHandler.php
```diff
@@ -0,0 +1,40 @@
...
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Messenger\MessageHandler;
+
+use Contao\CoreBundle\Messenger\Message\SearchIndexMessage;
+use Contao\CoreBundle\Search\Indexer\IndexerInterface;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+class SearchIndexMessageHandler implements MessageHandlerInterface
```
ausi:
```suggestion
use Symfony\Component\Messenger\Attribute\AsMessageHandler;

#[AsMessageHandler]
class SearchIndexMessageHandler
```

The interface seems to be deprecated: https://github.com/symfony/symfony/blob/6.2/UPGRADE-6.2.md#messenger

core-bundle/src/Messenger/MessageHandler/SearchIndexMessageHandler.php
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        if ($message->shouldIndex()) {
+            $this->indexer->index($message->getDocument());
+        }
+
+        if ($message->shouldDelete()) {
+            $this->indexer->delete($message->getDocument());
+        }
```
ausi: Is there a try catch missing here?

At least the `Contao\CoreBundle\Search\Indexer\IndexerException`s should be ignored here, otherwise the failure queue gets filled up quickly with any reasons why pages are not indexed (like wrong status code, noSearch enabled,…)

core-bundle/src/Messenger/MessageHandler/SearchIndexMessageHandler.php
Toflar: Yes but I left it intentionally since we do not require 6.2 as minimum version.

Toflar: Ah but the attribute seems to be there since 5.4, so I can actually change it, yes.

core-bundle/src/Messenger/Transport/AutoFallbackTransportFactory.php
Toflar: Ah right, it's `host` though, not `path` then ;) I'll fix it.

ausi (dismissed):
Very nice 😎

ausi (dismissed):

core-bundle/src/Messenger/Message/HighPrioMessageInterface.php
```diff
@@ -0,0 +1,17 @@
...
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Messenger\Message;
+
+interface HighPrioMessageInterface
```
leofeyer: I would prefer `HighPriorityMessageInterface`.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5406, opened 2022-10-18
Add stimulus controllers in the back end

aschempp:
This adds the foundation for Webpack/Symfony Encore and Stimulus controllers.

I have (re)written three controllers to Stimulus. All existing "modern" JS files are now processed with Webpack. Some _might_ be rewritten to Stimulus, but certainly not all – and they still work perfectly atm.

The `core.js` and `mootao.js` are still built using Gulp and are untouched by Webpack.

**TODOs/Thoughts:**
- [x] we might need to think about namespacing controllers to prevent conflicts across bundles (e.g. `data-controller="contao--core-bundle--clipboard"` following [the folder name conventions](https://stimulus.hotwired.dev/reference/controllers#identifiers)).
- [x] see if we can get rid of the `controller.json` files since our bundle will never import third-party controllers
- [x] Requires https://github.com/contao-components/mootools/pull/2


core-bundle/contao/widgets/MetaWizard.php
```diff
@@ -171,7 +171,7 @@ public function generate()
 			// Add the input fields
 			foreach ($this->varValue as $lang=>$meta)
 			{
-				$item = '<li data-language="' . $lang . '"><span class="lang">' . ($languages[$lang] ?? $lang) . ' ' . Image::getHtml('delete.svg', '', 'class="tl_metawizard_img" title="' . $GLOBALS['TL_LANG']['MSC']['delete'] . '" data-delete') . '</span>';
+				$item = '<li data-language="' . $lang . '" data-controller="metawizard"><span class="lang">' . ($languages[$lang] ?? $lang) . ' <button data-action="metawizard#delete:prevent">' . Image::getHtml('delete.svg', '', 'class="tl_metawizard_img" title="' . $GLOBALS['TL_LANG']['MSC']['delete'] . '"') . '</button></span>';
```
leofeyer: Just a note to myself: adding the `button` element needs to be backported to Contao 4 to fix the accessibility of the function.

aschempp: see https://github.com/contao/contao/pull/5411

core-bundle/assets/backend.js
```diff
@@ -0,0 +1,21 @@
+import { Application } from '@hotwired/stimulus'
+import { definitionForModuleAndIdentifier, identifierForContextKey } from '@hotwired/stimulus-webpack-helpers'
+
+import './scripts/autofocus'
```
Toflar, line 4: Just wondering if we shouldn't put these into some `backend` folder as well? Or where would the FE scripts go?

core-bundle/contao/languages/en/tl_preview_link.xlf
```diff
@@ -75,7 +75,7 @@
         <source>Please save this record to generate the preview link.</source>
       </trans-unit>
       <trans-unit id="tl_preview_link.hintEdit">
-        <source>Preview link: &lt;a href="%s" data-to-clipboard="%s"&gt;%s&lt;/a&gt;</source>
+        <source>Preview link</source>
```
Toflar, line 78: This to me is the main advantage of using Stimulus :)

core-bundle/package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "private": true,
+    "scripts": {
+        "dev": "NODE_ENV=development node_modules/.bin/encore dev --watch",
+        "prod": "env NODE_ENV=production node_modules/.bin/encore production"
```
Toflar: Might be a stupid question but why does `prod` need an `env` command and `dev` does not?

Toflar (commented):
Looks good to me, I would just like to know what browser targets we now have with babel-preset-env not being configured explicitly :) Maybe we should?

core-bundle/webpack.config.js
```diff
@@ -0,0 +1,16 @@
...
+Encore
+    .setOutputPath('public/')
+    .setPublicPath('/bundles/contaocore')
+    .setManifestKeyPrefix('')
+    .cleanupOutputBeforeBuild(['**/*', '!core.**', '!mootao.**'])
+    .disableSingleRuntimeChunk()
+
+    .enableSourceMaps()
+    .enableVersioning()
+
```
m-vo: I'm unsure about the babel defaults. Should we create a `babel.config.js` or do we need to enable `babel/preset-env` polyfills here?

```js
    .configureBabelPresetEnv((config) => {
        config.useBuiltIns = 'usage';
        config.corejs = 3;
        // …
    })
```

core-bundle/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "devDependencies": {
+        "@babel/core": "^7.19.3",
+        "@babel/preset-env": "^7.19.4",
+        "@hotwired/stimulus": "^3.1.0",
+        "@hotwired/stimulus-webpack-helpers": "^1.0.1",
+        "@symfony/webpack-encore": "^4.0.0",
+        "autoprefixer": "^10.4.12",
+        "webpack": "^5.74.0",
+        "webpack-cli": "^4.10.0"
+    }
```
m-vo: IMHO we should add a `browserslist` section here (or add an additional `.browserslistrc` file for that):

Something like this (but with our actual requirements that I'm unsure about):

```json
"browserslist": [
    "> 0.5%",
    "not dead",
    "not op_mini all",
    "not chrome < 51",
    "not safari < 10"
  ]
```

core-bundle/contao/languages/en/tl_preview_link.xlf
aschempp: absolutely not related to Stimulus 🤣

core-bundle/assets/backend.js
aschempp: I did think about that. We can always move scripts somewhere else and adjust the build scripts.

core-bundle/webpack.config.js
aschempp: according to https://symfony.com/doc/current/frontend/encore/babel.html that is not necessary

core-bundle/package.json
aschempp: added in 42ad6f0578ce726dbdb405bd36c25c98df993e65 according to https://symfony.com/doc/current/frontend/encore/postcss.html#adding-browserslist-to-package-json

core-bundle/package.json
aschempp: no idea, I copied that from my local recipes. But it seems unnecessary, I've added the defaults in 63b6b625e2be22927f66253bef099762fbd2075b

aschempp:
I would love to get this merged asap so I/we can continue reworking other elements to Stimulus (e.g. using https://www.stimulus-components.com/docs/stimulus-sortable/)

core-bundle/webpack.config.js
m-vo: Ah nice, I didn't know that!

core-bundle/contao/templates/backend/be_main.html5
```diff
@@ -34,17 +34,11 @@
...
-      $objCombiner->add('bundles/contaocore/autofocus.min.js');
-      $objCombiner->add('bundles/contaocore/limit-height.min.js');
-      $objCombiner->add('bundles/contaocore/tips.min.js');
-      $objCombiner->add('bundles/contaocore/sectionwizard.min.js');
-      $objCombiner->add('bundles/contaocore/modulewizard.min.js');
-      $objCombiner->add('bundles/contaocore/toggle-fieldset.min.js');
       $objCombiner->add('system/themes/'.$this->theme.'/hover.min.js');
       echo $objCombiner->getCombinedFile();
     ?>"></script>
+    <script src="<?= $this->asset('backend.js', 'contao_core') ?>"></script>
```
m-vo, line 41: This will only work with having `disableSingleRuntimeChunk()` set and  `splitEntryChunks()` not enabled. Otherwise you would miss the other needed files (like the runtime). 

I'm fine with this for now but I definitively think we should use the `symfony/webpack-encore-bundle` in the future that already handles this automatically. There is a `TagRenderer` that can `renderWebpackScriptTags` and `renderWebpackLinkTags` and uses the provided `EntrypointLookup` to find all files (plus Twig functions using it).

m-vo (dismissed):
I think we should merge this as is. We can add further improvements down the line.

I'm currently trialing ux-turbo for a new file manager and this foundation would also be needed there (and plays nicely together).

core-bundle/contao/templates/backend/be_main.html5
aschempp: as long as we only have one back end script, there is no point in splitting stuff imho. But yes, that's something to decide when time comes.

m-vo: Yes, I also think we should currently produce a single chunk right now (splitting has some overhead as well). :+1: 
My point was making the config independent from the usage.

m-vo (dismissed):

leofeyer (dismissed):

leofeyer (approved):

PR #5407, opened 2022-10-18
Add cache busting for the public core bundle scripts

leofeyer:
After merging this, we could use `$this->asset('ajax-form.min.js', 'contao_core')` in #5307. It might be rendered redundant by #5406 therefore it is a draft PR at the moment.

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/AddAssetsPackagesPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/AddAssetsPackagesPass.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Contao\CoreBundle\DependencyInjection\Compiler;
 
 use Composer\InstalledVersions;
+use Contao\CoreBundle\ContaoCoreBundle;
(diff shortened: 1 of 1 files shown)


core-bundle/src/DependencyInjection/Compiler/AddAssetsPackagesPass.php
```diff
@@ -70,6 +71,13 @@ private function addBundles(ContainerBuilder $container): void
 
                 $container->setDefinition('assets._version_'.$packageName, $def);
                 $packageVersion = new Reference('assets._version_'.$packageName);
+            } elseif ('contao_core' === $packageName) {
```
Toflar, line 74: Could be built as a general solution with a new interface: `if ($bundle instance of VersionProvidingBundleInterface)`

Then everybody could benefit from it by implementing the interface.

Also, this approach would leak the version of Contao again (we did hide it everywhere in the back end etc.).
So I think it should something along the lines of `substr(sha1($secret . $version), 0, 8)`.

leofeyer:
Closing this in favor of #5598.

PR #5339, opened 2022-10-05
Evaluate custom regex in backend pwd validation.

AlexanderWillner:
This commit is based on the initial pull request https://github.com/contao/contao/pull/2190 that addresses https://github.com/terminal42/contao-password-validation/issues/3. In the context of https://github.com/contao/contao/issues/5316 the changes were rebased for Contao 4.13 and also refactored after initial tests.

Fixes #5316

PR #2190, opened 2020-08-24
Call rgxp callback in BackendPassword controller

richardhj:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes terminal42/contao-password-validation#3
| Docs PR or issue | 

The `BackendPassword` controller implements its own form validation.
Usually, we would do a `$objWidget = new Password();` and call `$objWidget->validate()` to have all form validation done, but this is not the case in the `BackendPassword` controller.

In this case, the BackendPassword controller lacks the `rgxp` check, which validates the value if a custom rgxp is set on the password field.
This results in an issue in our "terminal42/contao-password-validation" extension that validates the password.

I don't know, why the `BackendPassword` controller is not using the Widget to build and validate the form, but there might be a reason. I decided to check for registered rgxp hooks.

## Thoughts
- [ ] In the back end scope, you usually can throw \Exceptions in the save_callback to add an error to the widget. This will not work in the current implementation. Do we need to change this as well to be consistent?
- [ ] Do we rather need to change the whole implementation, e.g. populate  a `$objWidget = new Password();`? 

cc @aschempp

PR #5415, opened 2022-10-23
Use password widget for pwChange

fritzmg:
Fixes https://github.com/terminal42/contao-password-validation/issues/3

Alternate version of #2190 / #5339 using the `Password` widget completely, including rendering. Though a custom template is in use to be compatible with the `be_password` view.


core-bundle/src/Resources/contao/templates/backend/be_widget_pwchange.html5
```diff
@@ -0,0 +1,9 @@
+
+<div class="widget">
+  <label for="password"><?= $this->label ?></label>
+  <input type="password" name="password" id="password" class="tl_text" value="" autocomplete="new-password" placeholder="<?= $this->label ?>" required>
+</div>
+<div class="widget">
+  <label for="confirm"><?= Contao\System::getContainer()->get('translator')->trans('MSC.confirm.0', [], 'contao_default') ?></label>
+  <input type="password" name="password_confirm" id="confirm" class="tl_text" value="" autocomplete="new-password" placeholder="<?= Contao\System::getContainer()->get('translator')->trans('MSC.confirm.0', [], 'contao_default') ?>" required>
```
leofeyer: Why not use `$this->trans()` here?

fritzmg: That is a feature of regular templates and does not work for Widget templates, as the `trans` method is only available in `Template` and not `Widget` (or `TemplateInheritance`).

leofeyer: Then we should add it there as well or move it to the trait.

fritzmg: I didn't want to do either in this bugfix PR though as this should be a separate discussion. I can switch back to `<?= $GLOBALS['TL_LANG']… ?>` for this PR, if you prefer.

leofeyer: What do you mean by "switch back to $GLOBALS"? The original code is this:

`<label for="confirm"><?= $this->confirm ?></label>`

It seems we have to add something like this:

```
$widget->password = $GLOBALS['TL_LANG']['MSC']['password'][0];
$widget->confirm = $GLOBALS['TL_LANG']['MSC']['confirm'][0];
```

fritzmg: > What do you mean by "switch back to $GLOBALS"?

I meant using 

```php
<?= $GLOBALS['TL_LANG']… ?>
```

in the template, like we do in other templates.

fritzmg: > It seems we have to add something like this:

I do not think we should set translation strings for a widget's template this way. But if you prefer that over `<?= $GLOBALS['TL_LANG']… ?>` in the template I will change it.

leofeyer: Either we add `$this->trans()` for widgets or we set the translated string as template variable (like we do everywhere else). @ausi What would you prefer?

ausi: I think I’d set the template variables as you outlined in https://github.com/contao/contao/pull/5415/files#r1004633603

Side note: we also use `$GLOBALS['TL_LANG']` directly in some templates (like `be_welcome` for example).

leofeyer: I have fixed this in #5439 now.

fritzmg: Changed in 777a030d6b

fritzmg:
Regarding `Message::reset()`: otherwise the following might occur:



as `Contao\Password` executes `Message::addConfirmation` itself when the password validates.

core-bundle/src/Resources/contao/controllers/BackendPassword.php
```diff
@@ -52,25 +51,37 @@ public function __construct()
...
 
 		$objTemplate = new BackendTemplate('be_password');
+		$objTemplate->widget = $widget->parse();
 
 		if (Input::post('FORM_SUBMIT') == 'tl_password')
 		{
+			$widget->validate();
+
+			// Remove any messages added by the Password widget itself
+			Message::reset();
```
leofeyer: This seems strange to me. Why is there a _Please enter a password_ error message if a password has been provided?

fritzmg: The 

> _Pleaser enter a **new** password._ 

error occurs, if you enter the same password as you previously had again. The `BackendPassword` controller checks for this and displays this error message via `Message::addError`. However, the `Contao\Password` widget does not care about that and thus, since the password validates, wants to add a confirmation message that the password has been updated via `Message::addConfirmation`.

core-bundle/src/Resources/contao/controllers/BackendPassword.php
```diff
@@ -52,25 +51,37 @@ public function __construct()
...
 
 		$objTemplate = new BackendTemplate('be_password');
+		$objTemplate->widget = $widget->parse();
 
 		if (Input::post('FORM_SUBMIT') == 'tl_password')
 		{
+			$widget->validate();
+
+			// Remove the confirmation message added by the widget
+			Message::reset();
```
leofeyer: We should first merge the fix in #5489, then this hack will become obsolete.

leofeyer (dismissed):
It still feels kinda hacky but it sure works.

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

core-bundle/src/Resources/contao/controllers/BackendPassword.php
```diff
@@ -52,25 +51,35 @@ public function __construct()
...
+		$widget->confirm = $GLOBALS['TL_LANG']['MSC']['confirm'][0];
 
 		$objTemplate = new BackendTemplate('be_password');
+		$objTemplate->widget = $widget->parse();
 
 		if (Input::post('FORM_SUBMIT') == 'tl_password')
 		{
+			$widget->validate();
+
 			$pw = $request->request->get('password');
```
leofeyer, line 78: @fritzmg Why are we not using `$widget->value` here instead of reading the POST data?

fritzmg: No particular reason (other than this being untouched from before).

leofeyer: Turns out we cannot use `$widget->value` because it returns the password hash instead of the actual password.

AlexanderWillner:
It seems this was merged into 4.13.13 (correct me if I wrong). However, in my tests, in contrast to #5339, this patch does not invoke the custom regexp, as I can still change the password to something very simple (it seems it just needs to be longer than 8 characters).

fritzmg:
@AlexanderWillner I cannot confirm that.

1. Install `terminal42/contao-password-validation`.
2. Configure `terminal42_password_validation.Contao\BackendUser` in your `config.yaml`.
3. Try to change the password of any back end user - the password validation will be enforced <sup>1</sup>.
4. Try to change your own password in your back end profile - the password validation will be enforced <sup>1</sup>.
5. Enforce a password change on any back end user (or your own).
6. Log in with that back end user and set a new password - the password validation will be enforced <sup>1</sup>.

<sup>1</sup> from `terminal42_password_validation.Contao\BackendUser`

PR #5416, opened 2022-10-24
Add a listener to set an accessible label for node toggles

### LINKED ISSUE #5392: Link „Alle umschalten“  - Status für Screenreader nicht erkennbar
Ähnlich wie in Ticket contao/contao#5393  können assistive Technologien wie Screenreader auch beim Link „Alle umschalten“ nicht erkennen, welcher Zustand (geöffnet/geschlossen) gerade aktiv ist. Die Nutzer müssten also zum nachfolgenden Tree View wechseln, dort nachschauen und dann ggf wieder zurücktabben. Möglich, aber ziemlich nervig.

aschempp:
Fixes https://github.com/contao/contao/issues/5392


core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
```diff
@@ -0,0 +1,66 @@
...
+        $session = $sessionBag->all();
+
+        $node = $table . '_tree';
+
+        if ($GLOBALS['TL_DCA'][$table]['list']['sorting']['mode'] == 6) {
+            $node = $table . '_' . ($GLOBALS['TL_DCA'][$table]['config']['ptable'] ?? '') . '_tree';
+        }
+
+        if (empty($session[$node]) || !\is_array($session[$node]) || current($session[$node]) != 1) {
+            $GLOBALS['TL_DCA'][$table]['list']['global_operations']['toggleNodes']['label'] = &$GLOBALS['TL_LANG']['DCA']['expandNodes'];
```
Toflar, line 56: Translator?

aschempp: can't use that for references

core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function __invoke(string $table): void
+    {
+        if (
+            DC_Table::class !== DataContainer::getDriverForTable($table)
+            || !isset($GLOBALS['TL_DCA'][$table]['list']['global_operations']['toggleNodes'])
+            || isset($GLOBALS['TL_DCA'][$table]['list']['global_operations']['toggleNodes']['label'])
+            || 'ptg=all' !== ($GLOBALS['TL_DCA'][$table]['list']['global_operations']['toggleNodes']['href'] ?? null)
+            || null === ($request = $this->requestStack->getCurrentRequest())
+            || null === ($session = $request->getSession())
```
leofeyer, line 40: The session should be retrieved from the request stack.

core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function __construct(RequestStack $requestStack)
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    public function __invoke(string $table): void
+    {
+        if (
+            DC_Table::class !== DataContainer::getDriverForTable($table)
```
leofeyer: Why are you limiting this to `DC_Table`? There is a "toggle nodes" button in the file manager, too.

core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
aschempp: That is not available in Symfony 4.4 (which we use for Contao 4.9)

core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
aschempp: We could extend this for DC_Folder, but the toggle operation is handled by the DC and the query key is different for each driver 😉

aschempp: added in 9226ea5ffbdc32bafb104d4607c73ff67f6fe3f3

core-bundle/src/EventListener/DataContainer/ToggleNodesLabelListener.php
leofeyer: Then please create a second PR for Contao 5 as the current changes are not upstream compatible. (Or only add this feature in Contao 5.1, which I would prefer.)

aschempp: I can do that as soon as this is merged so we know the final implementation, right?

aschempp:
not sure how to fix the failing tests?

ausi:
> not sure how to fix the failing tests?

Fixed in d7a316f89e3230c03f24b67ae1e515be6d86dd01 (not sure if there is a better way to tell phpstan that the `$GLOBALS` can be mutated externally)

leofeyer:
Actually, this is a pretty good workaround IMHO. We have been ignoring these kind of errors so far:

https://github.com/contao/contao/blob/2b25d7ddd2ead2df62a58ef97c7de15fdb11499a/phpstan.neon#L34-L47

ausi:
> We have been ignoring these kind of errors so far:

I like `/** @var array $GLOBALS */` better than the ignore list I think.
Maybe add a comment to describe why this is there?
Something like `/** @var array $GLOBALS PHPStan cannot infer the array shape here */`?

leofeyer:
See #5542.

aschempp:
anything else to do here for me? 🙃

ausi:
> anything else to do here for me? 🙃

I think https://github.com/contao/contao/pull/5416#discussion_r1006522091 needs to be resolved first, otherwise merging this PR would completely block the upstream merge.

leofeyer:

leofeyer:
I just realized that there is a PR for Contao 4.13 already (see #5674), so I‘m closing this one.

PR #5418, opened 2022-10-24
Load the language file before applying default labels

aschempp:
Fixes https://github.com/contao/contao/issues/5029

I've run into the same issue while implementing a new feature (PR to follow) and this fixed the issue for me. It also makes sense to make sure labels are loaded before checking if they exist 🙃

### DIFF
--- a/core-bundle/contao/library/Contao/DcaLoader.php
+++ b/core-bundle/contao/library/Contao/DcaLoader.php
@@ -149,6 +149,8 @@ class DcaLoader extends Controller
 	 */
 	private function addDefaultLabels()
 	{
+		System::loadLanguageFile($this->strTable);
+
 		// Operations
 		foreach (array('global_operations', 'operations') as $key)
 		{
--- a/core-bundle/contao/library/Contao/DcaLoader.php
+++ b/core-bundle/contao/library/Contao/DcaLoader.php
@@ -149,8 +149,6 @@ class DcaLoader extends Controller
 	 */
 	private function addDefaultLabels()
 	{
-		System::loadLanguageFile($this->strTable);
-
 		// Operations
 		foreach (array('global_operations', 'operations') as $key)
 		{
@@ -166,13 +164,11 @@ class DcaLoader extends Controller
 					continue;
 				}
 
-				if (isset($GLOBALS['TL_LANG'][$this->strTable][$k]) || !isset($GLOBALS['TL_LANG']['DCA'][$k]))
+				$v['label'] = &$GLOBALS['TL_LANG'][$this->strTable][$k];
+
+				if (!isset($GLOBALS['TL_LANG'][$this->strTable][$k]) && isset($GLOBALS['TL_LANG']['DCA'][$k]))
 				{
-					$v['label'] = &$GLOBALS['TL_LANG'][$this->strTable][$k];
-				}
-				elseif (isset($GLOBALS['TL_LANG']['DCA'][$k]))
-				{
-					$v['label'] = &$GLOBALS['TL_LANG']['DCA'][$k];
+					$GLOBALS['TL_LANG'][$this->strTable][$k] = $GLOBALS['TL_LANG']['DCA'][$k];
 				}
 			}
 


leofeyer:
I don‘t think that this is correct, because now you cannot create DCA extracts without also loading the language files anymore. This might be desirable in unit tests or when warming up the cache.

Instead, you should make sure to always call `loadLanguageFiles()` before loading a data container.

aschempp:
I have changed the implementation and I think now it is correct. We must **always** use the translated label to allow overriding if translation files are loaded later. This fixed the problem in my installation.

One thing that will not work now is loading a different language afterwards, but I honestly don't think that's a real use case.

ausi:
I think we should instead try the fix outlined in https://github.com/contao/contao/pull/5349#issuecomment-1277714331

> Maybe a better solution would be to fall back to the live schema from the database if the type cache is missing. This way we wouldn’t have to load the DCAs at all.

I will create a PR for that. (UPDATE: see #5428)

leofeyer:
> One thing that will not work now is loading a different language afterwards, but I honestly don't think that's a real use case.

Haven‘t we been working on this on the last developers meeting? Being able to generate a German PDF on an English page by somehow temporarily overwriting the labels and then restoring them? This is a valid use case IMHO.

fritzmg:
> Haven‘t we been working on this on the last developers meeting? Being able to generate a German PDF on an English page by somehow temporarily overwriting the labels and then restoring them? This is a valid use case IMHO.

This will always work, if you use the translator.

The issue here is that the label's translations won't change, if you manually execute `System::loadLanguageFile` as the labels are not stored by reference anymore.

aschempp:
> Haven‘t we been working on this on the last developers meeting? Being able to generate a German PDF on an English page by somehow temporarily overwriting the labels and then restoring them? This is a valid use case IMHO.

Yes, but I don't think this is related. We're only fixing the default DCA operations label here.

> The issue here is that the label's translations won't change, if you manually execute `System::loadLanguageFile` as the labels are not stored by reference anymore.

On the DCA operations, exactly. And this is what this changes now 🙃

leofeyer:
See #5428.

aschempp:
I don't understand how the two PRs are related? This PR fixes a clear issue of incorrectly assigning a label by reference, whereas the other PR fixes an issue that made this issue show up. Nonetheless we are still incorrectly assigning labels!

core-bundle/contao/library/Contao/DcaLoader.php
```diff
@@ -164,13 +164,11 @@ private function addDefaultLabels()
...
-				{
-					$v['label'] = &$GLOBALS['TL_LANG'][$this->strTable][$k];
-				}
-				elseif (isset($GLOBALS['TL_LANG']['DCA'][$k]))
+				$v['label'] = &$GLOBALS['TL_LANG'][$this->strTable][$k];
+
+				if (!isset($GLOBALS['TL_LANG'][$this->strTable][$k]) && isset($GLOBALS['TL_LANG']['DCA'][$k]))
 				{
-					$v['label'] = &$GLOBALS['TL_LANG']['DCA'][$k];
+					$GLOBALS['TL_LANG'][$this->strTable][$k] = $GLOBALS['TL_LANG']['DCA'][$k];
```
ausi, line 171: I don’t think it is a good idea to set values on `$GLOBALS['TL_LANG']` outside of our language files.

aschempp: what's _your language files_ in the DCA language loader? It is setting the labels to the _language files_ it loads the DCA for, and the if-condition skips if **exactly this** label is already set.

ausi: In a DCA with a reference to `&$GLOBALS['TL_LANG']['tl_mytable']['something']` that is missing from the language file, with this change suddenly the label of `$GLOBALS['TL_LANG']['DCA']['something']` would show up. This would be very confusing and unexpected IMO.

I think changes to `$GLOBALS['TL_LANG']` should only be done in language files to prevent confusion and weird problems with references.

aschempp: It would also appear currently since we assign the label to the field? Also, this is exactly what we are trying to solve! If you did not assign an english label but only a german one, loading the DCA would replace the reference and now allow you to load german.

On another note, maybe that problem could be solved by using `array_key_exists` instead of `isset`?

leofeyer:
As discussed in the Contao call, we cannot merge this PR, because it would change the behavior of the language references. If at all, we could think about automatically loading the language files (as initially requested in https://github.com/contao/contao/pull/5418/commits/cfe693f00fb4f6e061c9f53cf6da3ed5bdb76a92) <del>but this might not work if someone has loaded a custom language beforehand</del>.

PR #5420, opened 2022-10-24
Fix the reverse toggle operation

aschempp:
The toggle operation in `tl_member`, `tl_member_group`, `tl_user` and `tl_user_group` are currently not working 🙈


Toflar (approved):
For anybody asking the question if this does not affect 4.13: No it does not, `reverseToggle` is a new 5.0 feature :)

PR #5421, opened 2022-10-25
Handle errors with important parts

aschempp:
I still occasionally get the _The Y coordinate plus the height must not be greater than 1_ error in my client project, possibly due to old image resize configurations.

This PR fixes two issues:
 1. it adds the file information to the `CoordinatesOutOfBoundsException` so one knows which file is actually affected
 2. it then allows to edit the file in the back end, since it currently would throw the same error again when editing in the file manager. The file manager now looks like this and I can easily fix the errors.
 
Bildschirmfoto 2022-10-25 um 07 33 28

### DIFF
--- a/core-bundle/src/Image/ImageFactory.php
+++ b/core-bundle/src/Image/ImageFactory.php
@@ -16,6 +16,7 @@ use Contao\CoreBundle\Framework\ContaoFramework;
 use Contao\CoreBundle\Monolog\ContaoContext;
 use Contao\FilesModel;
 use Contao\Image\DeferredResizerInterface;
+use Contao\Image\Exception\CoordinatesOutOfBoundsException;
 use Contao\Image\Image;
 use Contao\Image\ImageInterface;
 use Contao\Image\ImportantPart;
@@ -158,7 +159,11 @@ class ImageFactory implements ImageFactoryInterface
 
         if (!\is_object($path) || !$path instanceof ImageInterface) {
             if (null === $importantPart) {
-                $importantPart = $this->createImportantPart($image);
+                try {
+                    $importantPart = $this->createImportantPart($image);
+                } catch (CoordinatesOutOfBoundsException $exception) {
+                    throw new CoordinatesOutOfBoundsException($exception->getMessage()." for file \"$path\"", $exception->getCode(), $exception);
+                }
             }
 
             $image->setImportantPart($importantPart);
--- a/core-bundle/src/Image/ImageFactory.php
+++ b/core-bundle/src/Image/ImageFactory.php
@@ -162,7 +162,7 @@ class ImageFactory implements ImageFactoryInterface
                 try {
                     $importantPart = $this->createImportantPart($image);
                 } catch (CoordinatesOutOfBoundsException $exception) {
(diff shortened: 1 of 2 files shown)


fritzmg:
Since this is a bugfix, it should be based against Contao 4.9.

core-bundle/src/Image/ImageFactory.php
```diff
@@ -121,7 +122,11 @@ public function create($path, $size = null, $options = null): ImageInterface
 
         if (!\is_object($path) || !$path instanceof ImageInterface) {
             if (null === $importantPart) {
-                $importantPart = $this->createImportantPart($image);
+                try {
+                    $importantPart = $this->createImportantPart($image);
+                } catch (CoordinatesOutOfBoundsException $exception) {
+                    throw new CoordinatesOutOfBoundsException($exception->getMessage()." for file \"$path\"", $exception->getCode(), $exception);
```
leofeyer:
```suggestion
                    throw new CoordinatesOutOfBoundsException(sprintf('%s for file "%s"', $exception->getMessage(), $path), $exception->getCode(), $exception);
```

leofeyer:
The error message in the screen could be shortened by omitting the file path, which is already mentioned in the headline and below the "Name" field. Do we really need it a third time?

fritzmg:
But then the full path would be missing when the error is shown in the front end (via debug mode) or via sentry for example.

ausi (dismissed):
Should be backported to 4.9

aschempp:
> Should be backported to 4.9

I don't think Contao 4.9 has an `ImageFactory` right? So it would be a different solution?

fritzmg:
Contao 4.9 has an `ImageFactory` and its implementation is identical to Contao 4.13 (disregarding code style and a few other details). There is no need for a back port I think as this should be easily merged upstream.

ausi:
> I don't think Contao 4.9 has an `ImageFactory` right?

The `ImageFactory` is there since Contao 4.3.0 😅

leofeyer:
> But then the full path would be missing when the error is shown in the front end (via debug mode) or via sentry for example.

The call is wrapped in a try-catch block, so the exception will never be caught by the debugger or by Sentry, will it?

fritzmg:
Which call do you mean? [this](https://github.com/contao/contao/blob/e7d26741707d00e806c6eff2388e8335de1db823/core-bundle/src/Image/ImageFactory.php#L128) isn't wrapped in a try catch block. The error will be shown (and logged by sentry) in the `ImageController` for example.

leofeyer (changes requested):
As stated above, this PR needs to be rebased to the `4.9` branch.

aschempp:
> As stated above, this PR needs to be rebased to the `4.9` branch.

done


> The call is wrapped in a try-catch block, so the exception will never be caught by the debugger or by Sentry, will it?

The try-catch block is only applied in the back end file view! The exception will also be triggered e.g. in the front end if an image with invalid important part is rendered.

ausi (dismissed):

Toflar (dismissed):
LGTM

leofeyer (approved):

PR #5424, opened 2022-10-25
Add a description list content element

### LINKED ISSUE #5374: Neues Contentelement: Definitionsliste
Ich möchte ein neues CE vorschlagen: eine Definitionsliste, also eine Erweiterung des ce_list um jeweils ein Feld (siehe auch: http://de.selfhtml.org/html/text/listen.htm#definition).
Damit wäre es Möglich, schnell ein paar Wertepaar anzugeben ohne extra eine Tabelle formatieren zu müssen.
_Beispiele:_
Max Mustermann | 1. Vorsitzender
Hans Mustermann | 2. Vorsitzender ...
_oder_
Ort | Simcity
Strasse | Neuestrasse ...

MfG, artemis

--- _Originally created by artemis on July 15th, 2009, at 07:07pm (ID 886)_

aschempp:
Fixes https://github.com/contao/contao/issues/5374

This also extends the `KeyValueWizard` to allow a predefined list of keys.


core-bundle/contao/languages/en/default.xlf
```diff
@@ -320,6 +320,12 @@
       <trans-unit id="CTE.list.1">
         <source>Generates an ordered or unordered list.</source>
       </trans-unit>
+      <trans-unit id="CTE.description_list.0">
+        <source>Description list</source>
+      </trans-unit>
+      <trans-unit id="CTE.definition_list.1">
```
ausi:
```suggestion
      <trans-unit id="CTE.description_list.1">
```

typo?

core-bundle/contao/dca/tl_content.php
```diff
@@ -324,6 +325,13 @@
 			),
 			'sql'                     => "blob NULL"
 		),
+		'descriptionlistitems' => array
```
m-vo: I know this tries to be consistent with `listitems` and `tableitems` but its very unreadable. I think I'd rather use camel case like in `unfilteredHtml` or `sortIndex`.

core-bundle/contao/templates/_new/content_element/description_list.html.twig
```diff
@@ -0,0 +1,9 @@
+{% extends "@Contao/content_element/_base.html.twig" %}
+
+{% block content %}
+    <dl>
+        {% for item in items %}
+            <{{ item.key }}>{{ item.value|insert_tag|raw }}</{{ item.key }}>
```
m-vo: I don't understand the markup - the [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) should contain zero or more groups each consisting of a `<dt>`, followed by a `<dd>` element.

So something like this:
```twig
<dl>
  {% for description in descriptions %}
    <dt>{{ description.term|insert_tag_raw }}</dt>
    <dd>{{ description.details|insert_tag_raw }}</dd>
  {% endfor %}
</dl>
```

Also I think, we should not support HTML by default (just in the insert tags) and add HTML attributes to all three element types.

m-vo (changes requested):
I don't understand the part about the predefined keys. Isn't this supposed to be a list of 2-tuples data-wise?

core-bundle/contao/dca/tl_content.php
m-vo: Or - if this is going to be a key/value storage after all - what about reusing `data` (the one the `template` content element is using)?

core-bundle/contao/templates/_new/content_element/description_list.html.twig
ausi: The description list can not only have `<dt>`/`<dd>` pairs but `<dt>` as well as `<dd>` can appear multiple times e.g. the following is also valid:

```html
<dl>
  <dt>Martin Auswöger</dt>
  <dt>Ausi</dt>
  <dd>A web developer</dd>
  <dd>Some guy</dd>
</dl>
```

But I think too that we should use the same data structure as in `data` and skip empty elements to make the example from above possible:

```twig
<dl>
  {% for description in descriptions %}
    {% if description.term is not empty %}
      <dt>{{ description.term|insert_tag_raw }}</dt>
    {% endif %}
    {% if description.details is not empty %}
      <dd>{{ description.details|insert_tag_raw }}</dd>
    {% endif %}
  {% endfor %}
</dl>
```

Additionally I’d want to have every group wrapped inside a `<div>` as this is a best practise for styling these elements.

core-bundle/contao/dca/tl_content.php
aschempp: the one for lists is called `listitems`, thats just where I got this from. My first idea was to call it `dlitems` but not sure that's more descriptive …

core-bundle/contao/templates/_new/content_element/description_list.html.twig
aschempp: > Additionally I’d want to have every group wrapped inside a `<div>` as this is a best practise for styling these elements.

You can always do that by adjusting your template imho.

aschempp: > Also I think, we should not support HTML by default (just in the insert tags) and add HTML attributes to all three element types.

I've applied the same configuration as for `listitems` (ol/ul)

ausi: > You can always do that by adjusting your template imho.

But we should provide a good default, that does it right.

fritzmg: Personally I wouldn't want the `<dt><dd>` pairs to be wrapped in `<div>` as I would want to use CSS grids to style them like this for example: https://codepen.io/fritzmg/pen/oNygPjQ

ausi: > I would want to use CSS grids to style them

Not a problem ☺️: https://codepen.io/ausi/pen/OJEPopy

fritzmg: Yes but still - that additional wrapper ist not really necessary? Wouldn't it be better to output as little HTML as possible by default? But I'm fine either way ;).

However I strongly agree that it should always be `<dt><dd>` pairs (without being user selectable).

core-bundle/contao/dca/tl_content.php
fritzmg: I'd just re-use `tl_content.data`. The key should be `dt` and the value `dd`.

aschempp: as @ausi already pointet out, a description list can contain multiple `<dd>` per `<dt>`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#single_term_multiple_descriptions

m-vo: I honestly don't think this would fit the use case many people have. I'd assume if you need several terms you would want to style them individually as well and then you need another structure again. For the details, I would rather allow a RTE field so that you could have multiple paragraphs to structure things.

ausi: >> I'd just re-use `tl_content.data`. The key should be `dt` and the value `dd`.
>
> as @ausi already pointet out, a description list can contain multiple `<dd>` per `<dt>`

This should be no problem by using the `data` column and a template similar to the one I mentioned in https://github.com/contao/contao/pull/5424#discussion_r1006192829

aschempp:
Two issues with the proposed solution
1. the KeyValueWizard removes rows with empty keys. We obviously could change that, but that would also make invalid options for the template element?
2. The description of the `tl_content.data` input field is really bad for that use case. So either we still need a custom field or another way to customise the field label. An onload_callback would do for regular edits, but that wouldn't work for edit-multiple 🤷 

Bildschirmfoto 2022-12-18 um 07 59 08

leofeyer:
As discussed in the Contao call, we can fix this as soon as #5673 has been merged.

ausi:
> 1\. the KeyValueWizard removes rows with empty keys. We obviously could change that, but that would also make invalid options for the template element?

I’d change that and skip empty keys in the template element.

> 2\. So either we still need a custom field or another way to customise the field label.

We have now since #5673 🎉

core-bundle/src/EventListener/DataContainer/Content/DescriptionListLabelListener.php
```diff
@@ -0,0 +1,39 @@
...
+    public function __invoke(array $attributes, DataContainer $dc = null): array
+    {
+        if (null === $dc || 'description_list' !== ($dc->getCurrentRecord()['type'] ?? null)) {
+            return $attributes;
+        }
+
+        $attributes['label'] = $this->translator->trans('tl_content.dl_label.0', [], 'contao_default');
+        $attributes['description'] = $this->translator->trans('tl_content.dl_label.1', [], 'contao_default');
+        $attributes['keyLabel'] = $this->translator->trans('tl_content.dl_key', [], 'contao_default');
+        $attributes['valueLabel'] = $this->translator->trans('tl_content.dl_value', [], 'contao_default');
```
ausi:
```suggestion
        $attributes['label'] = $this->translator->trans('tl_content.dl_label.0', [], 'contao_tl_content');
        $attributes['description'] = $this->translator->trans('tl_content.dl_label.1', [], 'contao_tl_content');
        $attributes['keyLabel'] = $this->translator->trans('tl_content.dl_key', [], 'contao_tl_content');
        $attributes['valueLabel'] = $this->translator->trans('tl_content.dl_value', [], 'contao_tl_content');
```

ausi (dismissed):
Apart from https://github.com/contao/contao/pull/5424#discussion_r1185977548 this LGTM 🎉

ausi (dismissed):

core-bundle/src/Controller/ContentElement/DescriptionListController.php
```diff
@@ -0,0 +1,43 @@
...
+                $descriptions[] = [
+                    'term' => $row['key'],
+                    'details' => $row['value'],
+                ];
+            }
+        }
+
+        $template->set('descriptions', $descriptions);
+
+        return $template->getResponse();
```
fritzmg:
```suggestion
        $template->set('descriptions', array_map(
            static function ($row): array {
                return [
                    'term' => $row['key'],
                    'details' => $row['value'],
                ];
            },
            StringUtil::deserialize($model->data, true)
        ));

        return $template->getResponse();
```

core-bundle/contao/widgets/KeyValueWizard.php
```diff
@@ -73,7 +73,7 @@ public function validate()
 				$options[$key]['key'] = trim($option['key']);
 				$options[$key]['value'] = trim($option['value']);
 
-				if ($options[$key]['key'])
+				if ($options[$key]['key'] || ($this->allowEmptyKeys && $options[$key]['value']))
```
ausi:
```suggestion
				if ($options[$key]['key'] !== '' || ($this->allowEmptyKeys && $options[$key]['value'] !== ''))
```
Otherwise the string `'0'` would be seen as empty.

fritzmg (dismissed):

ausi (dismissed):

core-bundle/src/Controller/ContentElement/DescriptionListController.php
m-vo: You can shorten this to an arrow function :wink:.

fritzmg: I always rely on the CS fixer or Rector to do it for me and this time it did not 😁

core-bundle/config/listener.yaml
```diff
@@ -69,6 +69,14 @@ services:
             - '@security.helper'
             - '@request_stack'
 
+    contao.listener.data_container.content.description_list_label:
+        class: Contao\CoreBundle\EventListener\DataContainer\Content\DescriptionListLabelListener
+        public: true
+        arguments:
+            - '@translator'
+        tags:
+            - { name: contao.callback, table: tl_content, target: fields.data.attributes }
```
leofeyer: This should be an `#[AsCallback()]` attribute in the class, shouldn‘t it?

aschempp: It could be, yes. This was probably added before we had `autoconfigure: true`. There are other listeners here though with tags. Technically it does not make a difference, I would suggest to review all tags-instead-of-attributes in a separate PR.

core-bundle/contao/templates/twig/content_element/description_list.html.twig
```diff
@@ -0,0 +1,14 @@
+{% extends "@Contao/content_element/_base.html.twig" %}
+
+{% block content %}
+    <dl>
```
leofeyer: @m-vo Do you want to add more blocks to this template, too?

core-bundle/contao/languages/en/tl_content.xlf
```diff
@@ -731,6 +731,18 @@
       <trans-unit id="tl_content.table.1">
         <source>Import table items from a CSV file</source>
       </trans-unit>
+      <trans-unit id="tl_content.dl_label.0">
+        <source>Description list content</source>
+      </trans-unit>
+      <trans-unit id="tl_content.dl_label.1">
+        <source>Enter a list of terms and details for the description list. You can conditionally leave the term field empty to create multiple details for one term.</source>
```
leofeyer: What does **conditionally** mean in this context? Under what conditions can one leave the term field empty?

core-bundle/contao/classes/DataContainer.php
```diff
@@ -785,9 +785,9 @@ protected function row($strPalette=null)
 	 *
 	 * @return string
 	 */
-	public function help($strClass='')
+	public function help($strClass='', $strDescription=null)
 	{
-		$return = $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['label'][1] ?? null;
+		$return = $strDescription ?? $GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['label'][1] ?? null;
```
leofeyer, line 791: This is a new concept that did not exist before. Normally, you would dynamically overwrite `$GLOBALS['TL_DCA'][$this->strTable]['fields'][$this->strField]['label']` in a listener.

@contao/developers Do you agree with this concept?

Toflar: Maybe @aschempp can shed some light on why this is needed and what other use cases this would simplify. There must be a reason for his decision here.

core-bundle/contao/languages/en/tl_content.xlf
aschempp: that's how descriptions lists work. You can have multiple descriptions per term.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#single_term_multiple_descriptions

core-bundle/contao/classes/DataContainer.php
aschempp: The problem is we are dynamically chaning the label of an element in the `DescriptionListLabelListener` using the new `attributes_callback`. Because in an edit-multiple scenario, one field can be a description element, another can be a template using the same widget. Accessing the DCA details would not work here.

core-bundle/contao/languages/en/tl_content.xlf
leofeyer: Then it should just be _You can leave the term field empty to create multiple details for one term_.

core-bundle/contao/templates/twig/content_element/description_list.html.twig
```diff
@@ -0,0 +1,14 @@
...
+        {% for description in descriptions %}
+            {% if description.term is not empty %}
+                <dt>{{ description.term|insert_tag_raw }}</dt>
+            {% endif %}
+            {% if description.details is not empty %}
+                <dd>{{ description.details|insert_tag_raw }}</dd>
+            {% endif %}
+        {% endfor %}
+    </dl>
```
m-vo: I would add attributes to the three main tags as well as a block that allows grouping the description (e.g. in a div) as described here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl?retiredLocale=de#wrapping_name-value_groups_in_div_elements.

```suggestion
    <dl{{attrs(description_list_attributes|default)}}>
        {% for description in descriptions %}
            {% block description %}
                {% if description.term is not empty %}
                    <dt{{attrs(description_term_attributes|default)}}>{{ description.term|insert_tag_raw }}</dt>
                {% endif %}
                {% if description.details is not empty %}
                    <dd{{attrs(description_details_attributes|default)}}>{{ description.details|insert_tag_raw }}</dd>
                {% endif %}
            {% endblock %}
        {% endfor %}
    </dl>
```

leofeyer (approved):

leofeyer:
Thank you @aschempp.

PR #5425, opened 2022-10-25
Allow sorting DCA fields ascending and descending

### LINKED ISSUE #5383: BE-Feature: Reverse and column sorting
At this point, Contao only offers the functionality to sort either ascending or descending as defined in the DCA. It would be a nice feature if the user could reverse the sorting. Especially  for tables sorted by date, clients are often asking whether it's possible to reverse the sorting sometimes.

Furthermore, I think it would be a nice feature and an improvement regarding the user experience if table columns (`showColumns => true`) would be clickable to change the sort by field. It's a feature, the people know from their operating system.

aschempp:
Fixes https://github.com/contao/contao/issues/5383


Toflar (dismissed):

leofeyer (changes requested):
I like the feature a lot, although the PR only enables it for the members listing. But I can change this myself.

However, there is a problem that needs to be fixed: The sorting menu now no longer recognizes the default sorting field.

leofeyer (approved):

PR #5427, opened 2022-10-25
Check the administrator e-mail address

fritzmg:
This PR implements the administrator email address check in `5.x`, as originally present in https://github.com/contao/contao/pull/5330


ausi (dismissed):

aschempp (changes requested):
I usually disable the system settings module for my clients, which means the link will become invalid. This also does not check for user permissions to the back end settings.

core-bundle/src/EventListener/AdministratorEmailListener.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __invoke(): string|null
+    {
+        /** @var Config $config */
+        $config = $this->framework->getAdapter(Config::class);
+
+        if ($config->get('adminEmail')) {
+            return null;
+        }
+
+        if (isset($GLOBALS['BE_MOD']['system']['settings']) && $this->security->isGranted(ContaoCorePermissions::USER_CAN_ACCESS_MODULE, 'settings')) {
```
leofeyer: I would remove this condition entirely, because the admin email address can also be set in the container configuration. Just say _No administrator email address has been set_ if there is none.

fritzmg: If it is set in the container configuration, then the message would not appear in the first place. If the email address is not set, the link would help regular users while advanced users will then may be go ahead and set `contao.localconfig.adminEmail` in their `config.yaml` instead. I would definitely not just say _No administrator email address has been set_ as there is also an administrator email setting in each website root and this message is specifcally about the `adminEmail` in the system config (and thus the link would still be helpful).

leofeyer: True.

ausi (dismissed):

fritzmg:
@leofeyer however in order to not need another translation string we could may be just not display the message at all, if the settings module is unset or the back end user does not have access?

Yes, in the former case you will have no notification in the back end at all anymore. But I would argue that if you completely removed the back end system settings you presumably know what you are doing anyway and you still have the exception telling you what's wrong when trying to send an email for example.

In the latter case the administrator will still see notification in the back end - or anyone who has access to the module.

aschempp:
I would personally prefer to get the message even if the settings module is disabled, because I'm sure I'll constantly forget about it but copy my template that disables the back end module 😅

leofeyer:
@fritzmg Can you resolve the conflicts, please?

fritzmg:
@leofeyer done

ausi (dismissed):

aschempp (dismissed):

core-bundle/contao/languages/en/default.xlf
```diff
@@ -275,6 +275,12 @@
       <trans-unit id="ERR.startStop">
         <source>The end time must be later than the start time.</source>
       </trans-unit>
+      <trans-unit id="ERR.noAdminEmailUrl">
+        <source>No administrator e-mail address has been set in the &lt;a href="%s" target="_top"&gt;system settings&lt;/a&gt;.</source>
+      </trans-unit>
+      <trans-unit id="ERR.noAdminEmail">
+        <source>No administrator e-mail address has been set in the system settings.</source>
```
leofeyer: This is a bit misleading, because the address might as well be defined in the container configuration. Would it be an option to just say _There is no administrator e-mail address set_?

core-bundle/contao/languages/en/default.xlf
```diff
@@ -275,6 +275,12 @@
       <trans-unit id="ERR.startStop">
         <source>The end time must be later than the start time.</source>
       </trans-unit>
+      <trans-unit id="ERR.noAdminEmailUrl">
+        <source>No administrator e-mail address has been set in the &lt;a href="%s" target="_top"&gt;system settings&lt;/a&gt;.</source>
```
leofeyer:
```suggestion
        <source>There is no administrator e-mail address set in the &lt;a href="%s" target="_top"&gt;system settings&lt;/a&gt;.</source>
```

core-bundle/contao/languages/en/default.xlf
fritzmg: In that case the system settings module will show you that it is set via the configuration.

leofeyer: You mean the system settings that the user does not have access to?

fritzmg: You mean in the case where:

* The system settings are not enabled for the user.
* The administrator email address is set via the container configuration.

In that case the user will have to contact the system administrator anyway (regardless of what the error message says). And the system administrator will then either:

* have access to the system settings, where it is explained to them that the setting is set via the container configuration.
* or know where to change the administrator email address, since they are the one who set it in the container configuration in the first place.

leofeyer: No, that is not what I meant. If a use cannot access the system settings, we should not tell them "it has not been set in the system settings", because that will make them search the system settings trying to fix the issue. We should say "has not been set" or "has not been set, please contact the site administrator" instead.

fritzmg: Alright, so

> There is no administrator e-mail address set, please contact the site administrator.

?

leofeyer: Yes, I like this. 👍

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5428, opened 2022-10-25
Read column types from database if cache is missing

### LINKED ISSUE #5029: Custom labels for operations in DCA not respected anymore

ausi:
Alternative to #5349
Fixes https://github.com/contao/contao/issues/5029#issuecomment-1216531922

In debug mode, or when the type cache is missing, we now retrieve the type information directly from the database instead of loading all DCAs.

The cache warmer still uses the information from the DCA, because the database might not be up to date when warming the cache.


leofeyer:
I definitely like this better than #5349. 👍

m-vo (dismissed):

core-bundle/contao/library/Contao/Model.php
```diff
@@ -391,10 +391,18 @@ public function mergeRow(array $arrData)
 	 *
 	 * @return array<string,array<string,string>>
 	 */
-	public static function getColumnCastTypes(): array
+	public static function getColumnCastTypes(bool $fromDatabase): array
```
leofeyer:
```suggestion
	public static function getColumnCastTypes(bool $fromDatabase = false): array
```

Since it is a public static function, we should probably not enforce the argument.

ausi: It is marked as `@internal`, so this should be fine IMO.

leofeyer: Ah, I missed that. I still think a default argument would be nicer, though. 🙈

ausi: Split into two methods in e5104da699d0105ecdb1ba9baaadde5096583b3a

leofeyer (approved):

leofeyer:
Thank you @ausi.

aschempp:
can we use this for Widget::getDefault/empty value as well?

ausi:
> can we use this for Widget::getDefault/empty value as well?

IMO yes, extending this with the required information should work fine. Can you create a new ticket for that?

PR #5434, opened 2022-10-26
Restore previous translations in `$GLOBALS['TL_LANG']`

fritzmg:
See discussion in #5371

Currently when doing

```php
echo $translator->trans('ERR.general', [], 'contao_default');
echo $translator->trans('ERR.general', [], 'contao_default', 'de');
echo $translator->trans('ERR.general', [], 'contao_default', 'cs');
echo $GLOBALS['TL_LANG']['ERR']['general'];
```

you get 

```
An error occurred!
Ein Fehler ist aufgetreten!
Došlo k chybě!
Došlo k chybě!
```

as the last translator call populates `$GLOBALS['TL_LANG']` with the given language. This PR resets the translations accordingly, so the output will be

```
An error occurred!
Ein Fehler ist aufgetreten!
Došlo k chybě!
An error occurred!
```


core-bundle/src/Translation/Translator.php
```diff
@@ -64,6 +65,12 @@ public function trans($id, array $parameters = [], $domain = null, $locale = nul
             $translated = vsprintf($translated, $parameters);
         }
 
+        // Restore previous translations in $GLOBALS['TL_LANG'] (#5371)
+        if (null !== $locale && $locale !== $this->getLocale()) {
+            $system = $this->framework->getAdapter(System::class);
+            $system->loadLanguageFile(substr($domain, 7), $this->getLocale());
```
leofeyer, line 71: Why are you not using `$this->getCatalogue($locale)` to restore the translations?

fritzmg: That won't restore the translations. The catalogue calls `System::loadLanguageFile` only when actually retrieving a translation.

core-bundle/tests/Translation/TranslatorTest.php
```diff
@@ -220,6 +220,53 @@ public function testUsesDecoratedCatalogues(): void
...
+        ;
+
+        $originalTranslator
+            ->method('getCatalogue')
+            ->willReturnCallback(
+                function ($locale) {
+                    $catalogue = $this->createMock(MessageCatalogueInterface::class);
+                    $catalogue
+                        ->method('getLocale')
+                        ->willReturn($locale ?? 'de')
```
ausi:
```suggestion
                        ->willReturn($locale ?? 'en')
```
shouldn’t this fall back to the default?

ausi (dismissed):
Apart from the nitpick in https://github.com/contao/contao/pull/5434#discussion_r1006173424 ✅

Toflar (dismissed):

ausi (dismissed):

core-bundle/src/Translation/Translator.php
leofeyer: But now the translator knows how to load language files which previously only the catalogue had to know. This violates the "separation of concerns" principle. @ausi /cc

ausi: I think our translator and catalogue class are strongly connected, so I don’t see a problem with this here. The `contao_` prefix check for example is also used in both classes.

fritzmg: If we want to separate it, we could introduce a `public function loadLanguageFile` in our `MessageCatalogue`.

ausi: I’d keep it as is.

leofeyer: Why not simply use 

```php
if (null !== $locale && $locale !== $this->getLocale()) {
    $this->getCatalogue($this->getLocale())->get($id, $domain);
}
```

? This should do the job and it will not break the separation of concerns principle.

fritzmg: I opted to not do it this way, because `get` will do more than just `System::loadLanguageFile`. Specifically this would be executed every time, even though it's not necessary:

https://github.com/contao/contao/blob/f81d9e91c36d0d065ae14e0c2703d15c837560fb/core-bundle/src/Translation/MessageCatalogue.php#L164-L186

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5440, opened 2022-10-27
Do not boot Contao framework in DefaultIndexer::delete()

### LINKED ISSUE #5426: SearchIndexListener causing potential `Failed to start the session because headers have already been sent`
### Affected version(s)

all

### Description

The issue can be best reproduced by calling `assets/images/foobar-image-that-does-not-exist.jpg`.

The `ImagesController` is one that does not require the `ContaoFramework` to be initialized. So what happens in this case is that you'll end up in a 404 and thus the `SearchIndexListener` will try to `delete()` on the `IndexerInterface $indexer` in the `kernel.terminate` event. So **after** the response (and thus the headers) have already been sent to the client.

The used indexer, by default, is our `DefaultSearchIndexer` which accesses the "old" `Search` library that supports hooks and thus, has to make sure, the `ContaoFramework` is initialized.
Hence, [it initializes the framework](https://github.com/contao/contao/blob/4.9/core-bundle/src/Search/Indexer/DefaultIndexer.php#L119). Normally, this is no problem because in 99% of all cases, it has already been initialized and so nothing happens. However, for more modern routes like our `ImagesController`, it is not.
So effectively, it will now initialize the `ContaoFramework` in the `kernel.terminate` environment. It will call hooks like the ones listening to `initializeSystem` and they potentially access the session causing `Failed to start the session because headers have already been sent` issues. But essentially, it could be anything.

I'm not quite sure how to fix this so marking `up for discussion`.

Toflar:
Fixes https://github.com/contao/contao/issues/5426

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -652,11 +652,10 @@ class Search
 		/** @var Connection $connection */
 		$connection = $connection ?? System::getContainer()->get('database_connection');
 
-		$result = $connection->prepare('SELECT id FROM tl_search WHERE url=:url')
-			->executeQuery(array('url' => $strUrl))
-		;
+		$stmt = $connection->prepare('SELECT id FROM tl_search WHERE url=:url');
+		$stmt->execute(array('url' => $strUrl));
 
-		foreach ($result->fetchAllAssociativeIndexed() as $id)
+		foreach ($stmt->fetchAllAssociative() as $id)
 		{
 			$connection->delete('tl_search', array('id' => $id));
 			$connection->delete('tl_search_index', array('pid' => $id));
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Contao\Database\Result;
+use Doctrine\DBAL\Driver\Connection;
 use Nyholm\Psr7\Uri;
 
 /**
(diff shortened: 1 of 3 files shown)


core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Contao\Database\Result;
+use Doctrine\DBAL\Driver\Connection;
```
fritzmg, line 14:
```suggestion
use Doctrine\DBAL\Connection;
```

Or is there a specific reason for `Doctrine\DBAL\Driver\Connection` in this case? (We usually use `Doctrine\DBAL\Connection` everywhere else.)

Toflar: We use the interface in the `DefaultIndexer`.

fritzmg: I see, may be that's an oversight there.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -646,20 +647,18 @@ public static function searchFor($strKeywords, $blnOrSearch=false, $arrPid=array
...
+		/** @var Connection $connection */
+		$connection = $connection ?? System::getContainer()->get('database_connection');
 
-		$objResult = $objDatabase->prepare("SELECT id FROM tl_search WHERE url=?")
-								 ->execute($strUrl);
+		$stmt = $connection->prepare('SELECT id FROM tl_search WHERE url=:url');
+		$stmt->execute(array('url' => $strUrl));
 
-		while ($objResult->next())
+		foreach ($stmt->fetchAllAssociative() as $id)
```
ausi, line 658: This causes an error `An exception occurred while executing 'DELETE FROM tl_search WHERE id = ?' with params [{"id":7}]`

```suggestion
		foreach ($stmt->fetchFirstColumn() as $id)
```

Missed that in my first review 🙈

PR #5442, opened 2022-10-27
Do not boot Contao framework in DefaultIndexer::delete()

### LINKED ISSUE #5426: SearchIndexListener causing potential `Failed to start the session because headers have already been sent`

Toflar:
Same as https://github.com/contao/contao/pull/5440 for 4.13

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Search.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Search.php
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Contao\Database\Result;
+use Doctrine\DBAL\Connection;
 use Nyholm\Psr7\Uri;
 
 /**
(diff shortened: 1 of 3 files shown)


Toflar:
@fritzmg I did change the typehint here because `->delete()` shortcuts etc. are not part of the interface and we use it anywhere else. I did not bother changing it in 4.9 though because there we use DBAL 2 and there's no difference.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

ausi:
This breaks in my setup, I get an error `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':url' at line 1`

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -782,33 +783,31 @@ public static function query(string $strKeywords, bool $blnOrSearch=false, array
...
+		$connection = $connection ?? System::getContainer()->get('database_connection');
 
-		$objResult = $objDatabase->prepare("SELECT id FROM tl_search WHERE url=?")
-								 ->execute($strUrl);
+		$result = $connection
+			->prepare('SELECT id FROM tl_search WHERE url=:url')
+			->executeQuery(array('url' => $strUrl));
 
-		while ($objResult->next())
+		foreach ($result->fetchAllAssociativeIndexed() as $id)
```
ausi, line 794: This doesn’t seem to work

```suggestion
		foreach ($result->fetchFirstColumn() as $id)
```

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -782,33 +783,31 @@ public static function query(string $strKeywords, bool $blnOrSearch=false, array
...
 		{
 			// Decrement document frequency counts
-			$objDatabase
+			$connection
 				->prepare("
 					UPDATE tl_search_term
-					INNER JOIN tl_search_index ON tl_search_term.id = tl_search_index.termId AND tl_search_index.pid = ?
+					INNER JOIN tl_search_index ON tl_search_term.id = tl_search_index.termId AND tl_search_index.pid = :id
```
ausi, line 800:
```suggestion
					INNER JOIN tl_search_index ON tl_search_term.id = tl_search_index.termId AND tl_search_index.pid = ?
```

Using `?` instead of the named parameters fixed it in my setup.

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -782,33 +783,31 @@ public static function query(string $strKeywords, bool $blnOrSearch=false, array
...
-	public static function removeEntry($strUrl)
+	public static function removeEntry($strUrl, Connection $connection = null)
 	{
-		$objDatabase = Database::getInstance();
+		$connection = $connection ?? System::getContainer()->get('database_connection');
 
-		$objResult = $objDatabase->prepare("SELECT id FROM tl_search WHERE url=?")
-								 ->execute($strUrl);
+		$result = $connection
+			->prepare('SELECT id FROM tl_search WHERE url=:url')
```
ausi, line 791:
```suggestion
			->prepare('SELECT id FROM tl_search WHERE url=?')
```
same here

core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -782,33 +783,31 @@ public static function query(string $strKeywords, bool $blnOrSearch=false, array
...
+			->executeQuery(array('url' => $strUrl));
```
ausi, line 792:
```suggestion
			->executeQuery('SELECT id FROM tl_search WHERE url=:url', array('url' => $strUrl));
```

This way, named parameters would work. Not sure which style we prefer.

PR #5444, opened 2022-10-28
Fix issues introduced by CSS reset changes in the back end

fritzmg:
Fixes the following issues introduced by #5370


core-bundle/contao/themes/flexible/main.css
```diff
@@ -2581,6 +2582,12 @@ fieldset.collapsed > legend {
 	background-color: var(--table-odd);
 }
 
+#picker-menu > ul {
```
leofeyer: Can you please move this up to where the other `#picker-menu` definitions are (around line 2126)?

leofeyer (changes requested):

core-bundle/contao/themes/flexible/main.css
fritzmg: Done

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5446, opened 2022-10-28
Fix issues in images content element if no images are defined

fritzmg:
This PR fixes two issues. The first issue is that `tl_content.singleSRC` and `tl_content.multiSRC` are nullable fields and thus the following error can occur:

```
TypeError: Contao\CoreBundle\Controller\ContentElement\ImagesController::getSources(): Return value must be of type array|string, null returned
```

To reproduce this create a new content element, set the type to _Gallery_, then save and close.

The second issue is that if no image has been defined - or the defined images have been deleted, the content element currently creates empty HTML output:

```html
<div class="content-image">            
</div>

<div class="content-gallery--cols-4 content-gallery">            
  <ul>
  </ul>
</div>
```

In Contao 4 nothing will be rendered if there are no actual images to be rendered and I think we should do the same in Contao 5.

### DIFF
--- a/core-bundle/src/Controller/ContentElement/ImagesController.php
+++ b/core-bundle/src/Controller/ContentElement/ImagesController.php
@@ -77,6 +77,10 @@ class ImagesController extends AbstractContentElementController
             iterator_to_array($filesystemItems)
         );
 
+        if (empty($imageList)) {
+            return new Response();
+        }
+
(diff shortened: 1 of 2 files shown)


ausi (approved):

m-vo (approved):

leofeyer:
Thank you @fritzmg.

PR #5451, opened 2022-10-28
Fix bug in Search::removeEntry()

ausi:
Fixes https://github.com/contao/contao/pull/5442#discussion_r1007971628


core-bundle/src/Resources/contao/library/Contao/Search.php
```diff
@@ -788,19 +788,17 @@ public static function removeEntry($strUrl, Connection $connection = null)
 		$connection = $connection ?? System::getContainer()->get('database_connection');
 
 		$result = $connection
-			->prepare('SELECT id FROM tl_search WHERE url=:url')
-			->executeQuery(array('url' => $strUrl));
+			->executeQuery('SELECT id FROM tl_search WHERE url=:url', array('url' => $strUrl));
```
leofeyer: I don‘t understand why this does not work. It seems to work fine here:

https://github.com/contao/contao/blob/87319db2093b8caed203820a41aa13fea4103ee3/core-bundle/src/Migration/Version404/Version447Update.php#L95-L109

(The same applies to the changes in line 798 and 802–809 below.)

leofeyer (dismissed):

core-bundle/src/Resources/contao/library/Contao/Search.php
leofeyer: @ausi pointed out to me that the parameters must start with a colon. 🦊

leofeyer (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5453, opened 2022-10-28
Make the version updates and the install tool MySQLi compatible

leofeyer:
If you want to reproduce the initial problem, make sure to set `'DATABASE_URL' => 'mysqli://…'` in your .env file.

### DIFF
--- a/core-bundle/src/Migration/Version408/Version480Update.php
+++ b/core-bundle/src/Migration/Version408/Version480Update.php
@@ -177,16 +177,10 @@ class Version480Update extends AbstractMigration
                 if (\is_array($jquery) && false !== ($i = array_search('j_mediaelement', $jquery, true))) {
                     unset($jquery[$i]);
 
-                    $stmt = $this->connection->prepare('
-                        UPDATE
-                            tl_layout
-                        SET
-                            jquery = :jquery
-                        WHERE
-                            id = :id
-                    ');
-
-                    $stmt->executeStatement([':jquery' => serialize(array_values($jquery)), ':id' => $row['id']]);
+                    $this->connection->executeStatement(
+                        'UPDATE tl_layout SET jquery = :jquery WHERE id = :id',
+                        ['jquery' => serialize(array_values($jquery)), 'id' => $row['id']]
+                    );
                 }
             }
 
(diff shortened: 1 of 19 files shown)


ausi:
Should we duplicate https://github.com/contao/contao/blob/5e17a4acbc0370e8c956f2d3f8db94d4e9648cbb/.github/workflows/ci.yml#L264-L267 to use mysqli too?

```yaml
 - name: Run the functional tests with mysqli
   run: vendor/bin/phpunit --testsuite=functional --colors=always 
   env: 
       DATABASE_URL: mysqli://root:root@127.0.0.1:3306/contao_test 
```
_(all three occurrences)_

core-bundle/src/Migration/Version408/Version480Update.php
```diff
@@ -348,26 +334,21 @@ public function runImportantPart(): void
...
-                        importantPartY = :y,
-                        importantPartWidth = :width,
-                        importantPartHeight = :height
-                    WHERE
-                        id = :id
-                ')
-                ->executeStatement($updateData)
-            ;
+            $this->connection->executeStatement('
+                UPDATE
```
ausi:
```suggestion
            $this->connection->executeStatement(
                'UPDATE
```
Seems to be inconsistent? _(just a nitpick, feel free to ignore)_

installation-bundle/src/InstallTool.php
```diff
@@ -384,14 +384,35 @@ public function persistAdminUser(string $username, string $name, string $email,
             '=' => '&#61;',
         ];
 
-        $statement->executeStatement([
```
ausi, line 387: `$statement = $this->connection->prepare("…` on line 355 needs to be removed I think.

leofeyer: Fixed in e7ef92549dcec88463628f5bd274df6c13879fa9.

core-bundle/src/Migration/Version408/Version480Update.php
leofeyer: Fixed in 1068a12064e06624a77d7660b92bfa3ec5bec387.

ausi (dismissed):

ausi (dismissed):

ausi (approved):

leofeyer:
Thanks a lot for your help @ausi!

Issue #5184, opened 2022-08-17
Allow preview links longer than 30 days


leofeyer:
### Affected version(s)

4.13

### Description

While discussing #5099, we asked us how much of a security layer the 30 days limit really is and whether there might be use cases for long running preview links. The SQL query could then be simplified to `DELETE FROM tl_preview_link WHERE expiresAt>UNIX_TIMESTAMP(NOW())`.

OMOSde:
I would be in favor of removing the limit. We often have projects with customer access to the preview that take much longer than 30 days.

Sending new preview links to the customer is always problematic because they quickly get confused with the links.

Alternatively, extending the preview links might be an option, but then they should not be removed when they expire.

The intervals could be adjusted well (without a limit) via DCA.

Translated with www.DeepL.com/Translator (free version)

ausi:
With #5454 it is now possible to specify an arbitrary “Valid until” date.

PR #5454, opened 2022-10-29
Check the preview link validity on every request

### LINKED ISSUE #4130: Restrict preview links to single page
### Description

There should be a possibility that the preview links are valid only for the page for which they were generated.
All other "unpublished" pages should remain invisible.

However, this use case is very rare for me. If there is no real need for the function, the ticket can also be closed.

see #4109

ausi:
Fixes #4130, #5089 and #5184

The ID of the preview link is now stored in the `FrontendPreviewToken`. This makes it possible to check if the current URL matches the URL of the shared link. But it also ensures that the preview link is still valid, so if you delete a preview link in the backend, open sessions from this link are invalidated immediately and therefore solves the security considerations from #5184 too.


core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -97,14 +101,36 @@ public function getBackendUsername(): ?string
...
+            return true;
+        }
+
+        $token = $this->getToken(self::FRONTEND_FIREWALL);
+
+        if (!$token instanceof FrontendPreviewToken) {
+            return false;
+        }
+
+        if (null === $token->getPreviewLinkId()) {
```
Toflar, line 120: Any reason for this? If preview is never working without that, why is it not a mandatory argument in the `FrontendPreviewToken`?

ausi: `$token->getPreviewLinkId()` is null for backend users and for existing sessions of previous Contao versions.

core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -172,4 +198,39 @@ private function getTokenFromSession(string $sessionKey): ?TokenInterface
 
         return $token;
     }
+
+    private function validatePreviewLink(FrontendPreviewToken $token): bool
```
leofeyer: This method should probably be called `isValidPreviewLink`.

core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -97,14 +101,36 @@ public function getBackendUsername(): ?string
         return $token->getUser()->getUserIdentifier();
     }
 
+    /**
+     * Tells whether the front end preview can be accessed.
+     */
+    public function isPreviewAllowed(): bool
```
leofeyer: I would prefer `$tokenChecker->hasValidPreviewToken()` instead of `isPreviewAllowed`. WDYT?

ausi: I’m not sure. A backend user does not have a valid PreviewToken but is still allowed to view the preview. So correct would be `hasValidPreviewTokenOrIsBackendUser()`

core-bundle/src/Security/Authentication/Token/TokenChecker.php
ausi:
```suggestion
    private function isValidPreviewLink(FrontendPreviewToken $token): bool
```

core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -97,14 +101,36 @@ public function getBackendUsername(): ?string
...
+
+        if (!$token instanceof FrontendPreviewToken) {
+            return false;
+        }
+
+        if (null === $token->getPreviewLinkId()) {
+            return false;
+        }
+
+        return $this->validatePreviewLink($token);
```
ausi:
```suggestion
        return $this->isValidPreviewLink($token);
```

core-bundle/src/Security/Authentication/Token/TokenChecker.php
ausi: Maybe naming it similar to `hasBackendUser()` and `hasFrontendUser()`?
Like `hasPreview()`?

leofeyer: Or `canAccessPreview`? It should be "speaking" when used together with the `$tokenChecker` variable, so maybe `$tokenChecker->allowsPreview()`?

ausi: Done in 11a9101049a335767f3775f26544c23a2d5f025a

core-bundle/src/Security/Authentication/Token/TokenChecker.php
```diff
@@ -172,4 +198,39 @@ private function getTokenFromSession(string $sessionKey): ?TokenInterface
...
+            return false;
+        }
+
+        if (!$previewLink['restrictToUrl']) {
+            return true;
+        }
+
+        $request = $this->requestStack->getMainRequest();
+
+        if ($request && strtok($request->getUri(), '?') === strtok($previewLink['url'], '?')) {
```
ausi: What is the correct way to compare two URLs, I should probably use something like `Request::create($previewLink['url'])` here and compare the two objects?

leofeyer: I think your solution is fine.

core-bundle/src/Resources/contao/dca/tl_preview_link.php
```diff
@@ -136,7 +134,7 @@
 			'flag'                    => DataContainer::SORT_DAY_DESC,
 			'sorting'                 => true,
 			'inputType'               => 'text',
-			'eval'                    => array('rgxp'=>'datim', 'readonly'=>true, 'doNotCopy'=>true, 'tl_class'=>'w50'),
+			'eval'                    => array('rgxp'=>'date', 'mandatory'=>true, 'doNotCopy'=>true, 'datepicker'=>true, 'tl_class'=>'w50 wizard'),
```
leofeyer: Why not `datim`?

core-bundle/src/Resources/contao/dca/tl_preview_link.php
```diff
@@ -106,21 +106,19 @@
...
 		(
 			'filter'                  => true,
-			'inputType'               => 'select',
-			'options'                 => array('1', '7', '30'),
-			'reference'               => &$GLOBALS['TL_LANG']['tl_preview_link']['expire_options'],
-			'eval'                    => array('mandatory'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50'),
-			'sql'                     => "int(10) unsigned NOT NULL default 1"
+			'inputType'               => 'checkbox',
+			'eval'                    => array('tl_class'=>'w50'),
+			'sql'                     => "char(1) NOT NULL default ''"
```
leofeyer, line 114:
```suggestion
			'sql'                     => "char(1) NOT NULL default '1'"
```

Shouldn‘t the "show unpublished" option be checked by default, too? This is the primary use case for preview links IMHO.

leofeyer (dismissed):
Works as expected. 👍

core-bundle/src/Resources/contao/dca/tl_preview_link.php
ausi: I think I’d keep it that way to make sure people explicitly allow that if they want to. 
We also shouldn’t change this in a bugfix release I think.

core-bundle/src/Resources/contao/dca/tl_preview_link.php
ausi: I didn’t know that our datepicker can pick the time too ☺️

```suggestion
			'eval'                    => array('rgxp'=>'datim', 'mandatory'=>true, 'doNotCopy'=>true, 'datepicker'=>true, 'tl_class'=>'w50 wizard'),
```

core-bundle/src/Security/Authentication/Token/TokenChecker.php
ausi: Improved in 1d57cdcac1f0e31980100a235af96610b00a3c3e

leofeyer (approved):

leofeyer:
Thank you @ausi.

Issue #5100, opened 2022-08-08
Add ArrayUtil::resolveReferences() method


ausi:
Should we add a helper method to `ArrayUtil` that removes references?

```php
public static function resolveReferences(array $array): array 
{
    return array_map(fn ($value) => \is_array($value) ? self::resolveReferences($value) : $value, $array);
}
```

See https://3v4l.org/niF5u#v8.1.8

_originally posted in https://github.com/contao/contao/pull/5060#issuecomment-1204884218_

We should then check the whole codebase where we manually use `unset()` to remove references and use this new method there.

ausi:
Closed in favor of #5457

PR #5457, opened 2022-10-30
Add a helper method to dereference arrays

ausi:
Implements #5100

> We should then check the whole codebase where we manually use `unset()` to remove references and use this new method there.

I didn’t find where we had problems with references in the past, looks like we did not use the `unset()` fix there 😕
So maybe we should move this to 5.0 and use it only in https://github.com/contao/contao/blob/f79bca7f82f5afd339ea70de6492724955e07b7c/core-bundle/src/DataContainer/DataContainerOperation.php#L34-L36


m-vo (dismissed):
The changes look good to me. I think I'd prefer to only merge it into 5.0 for now and backport it if we find any more occurrences in code originating <5.0. But I'm fine with merging it into 4.9 as well. :relaxed: .

ausi:
> So maybe we should move this to 5.0 and use it only in …

Done

m-vo (dismissed):
LGTM :+1: 

I wonder if we should handle the self-reference case?
```php
$array = [&$array];

StringUtil::resolveReferences($array); // results in an endless loop
```

ausi:
> I wonder if we should handle the self-reference case?

And throw an exception in that case?
How can we detect them?

core-bundle/contao/library/Contao/StringUtil.php
```diff
@@ -1129,4 +1129,12 @@ public static function numberToString(float|int $number, int|null $precision = n
...
 		return rtrim("$signPart$wholePart.$decimalPart", '.');
 	}
+
+	public static function resolveReferences(array $array): array
+	{
+		return array_map(
+			static fn ($value) => \is_array($value) ? self::resolveReferences($value) : $value,
+			$array,
+		);
+	}
```
m-vo, line 1139: Something like this should probably work to detect self references:

```suggestion
    public static function resolveReferences(array $array): array
    {
        return self::doResolveReferences($array);
    }

    private static function doResolveReferences(array $array, array $traversed = []): array {

        return array_map(
            static function ($value) use ($array, $traversed) {
                if (in_array($array, $traversed, true)) {
                    throw new \InvalidArgumentException('Cannot dereference an array that contains a reference to itself.');
                }

                $traversed[] = $array;

                return \is_array($value) ? self::doResolveReferences($value, $traversed) : $value;
            },
            $array,
        );
    }
```

But I'm not sure if we should consider this edge case or keep it simple.

ausi: The `in_array()` check would be pretty expensive I think. I’d probably be in favor of keeping it simple and fast ☺️

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5458, opened 2022-10-30
Update wrong config file path

CMSworker:
`contao/config/config.php` didn't work for me, had to add it in `system/config/localconfig.php` to see the old templates in the template-selection dropdown list of an element


fritzmg (changes requested):
`system/config/localconfig.php` is the wrong location. `contao/config/config.php` is the correct one.

May be you forgot to rebuild the application cache?

CMSworker:
Uh, nevermind, your right. This PR can be deleted/closed.

PR #5459, opened 2022-10-31
Warning: Undefined array key "PATH_TRANSLATED"

dennisbohn:
Got the following warning in debug mode when I tried to use Environment::get('documentRoot') under PHP 8.1.

image

Added "?? null" in the condition to prevent this.

```php
($_SERVER['ORIG_PATH_TRANSLATED'] ?? $_SERVER['PATH_TRANSLATED'] ?? null)
```

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Environment.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Environment.php
@@ -95,7 +95,7 @@ class Environment
 	 */
 	protected static function scriptFilename()
 	{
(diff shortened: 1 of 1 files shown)


ausi:
Is Contao 4.9 not affected by this?

dennisbohn:
Lines are equal in both versions so I think, Contao 4.9 will be affected too. Is it possible to consider both versions in one pull request? Or will changes in 4.9 be merged to 4.13? In this case I would create a new pull request for 4.9.

dennisbohn:
After looking into the commits I guess it will be merged to 4.13, so I created the pull request for 4.9.

#5460

PR #5461, opened 2022-11-01
Disable the request token check for idempotent actions

aschempp:
FINALLY removes the request token check for idempotent actions in the back end.

### DIFF
--- a/core-bundle/contao/drivers/DC_Table.php
+++ b/core-bundle/contao/drivers/DC_Table.php
@@ -116,13 +116,13 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 		$objSession = $container->get('request_stack')->getSession();
 
 		// Check the request token (see #4007)
-		if (Input::get('act') !== null)
-		{
-			if (Input::get('rt') === null || !$container->get('contao.csrf.token_manager')->isTokenValid(new CsrfToken($container->getParameter('contao.csrf_token_name'), Input::get('rt'))))
-			{
-				$objSession->set('INVALID_TOKEN_URL', Environment::get('requestUri'));
-				$this->redirect($container->get('router')->generate('contao_backend_confirm'));
-			}
+		if (
+			Input::get('act') !== null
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):
❤️

leofeyer:
This reenables the CSRF vulnerability that we had closed in https://github.com/contao/core/issues/4007, doesn‘t it? @ausi /cc

fritzmg:
The original issue was about editing and deleting. The request token check is still in place for `delete` and `toggle` actions and for POST requests.

Toflar:
Anything we know is idempotent, does not need a token.

ausi:
Are we sure, that `act=edit`, `act=show` and `act=select` are always idempotent?

Toflar:
Yes, that's the whole purpose of this PR :D

aschempp:
> Are we sure, that `act=edit`, `act=show` and `act=select` are always idempotent?

Technically, someone _could_ manipulate data in the `onload` callbacks. But then again you could also manipulate data in the list/parent/tree view where there is no request token check.

core-bundle/contao/drivers/DC_Table.php
```diff
@@ -116,13 +116,13 @@ public function __construct($strTable, $arrModule=array())
...
-		if (Input::get('act') !== null)
-		{
-			if (Input::get('rt') === null || !$container->get('contao.csrf.token_manager')->isTokenValid(new CsrfToken($container->getParameter('contao.csrf_token_name'), Input::get('rt'))))
-			{
-				$objSession->set('INVALID_TOKEN_URL', Environment::get('requestUri'));
-				$this->redirect($container->get('router')->generate('contao_backend_confirm'));
-			}
+		if (
+			Input::get('act') !== null
+			&& !\in_array(Input::get('act'), array('edit', 'show', 'select'), true)
```
ausi:
```suggestion
			!\in_array(Input::get('act'), array(null, 'edit', 'show', 'select'), true)
```

ausi (approved):

Toflar (approved):

Issue #5456, opened 2022-10-29
Error when editing multiple files


CMSworker:
### Affected version(s)

5.0.4

### Description

I'm currently testing 5.0.4 (local installation in XAMPP, only Core bundle and News bundle installed)

I added some images in the **files** folder. When I try to **edit multiple** I get a error message.

Debug Mode - Stack Trace:

```
TypeError:
Contao\DataContainer::isFieldExcluded(): Argument #2 ($field) must be of type string,
null given, called in \vendor\contao\core-bundle\contao\drivers\DC_Folder.php on line 1730

  at \vendor\contao\core-bundle\contao\classes\DataContainer.php:1532
  at Contao\DataContainer::isFieldExcluded('tl_files', null)
     (\vendor\contao\core-bundle\contao\drivers\DC_Folder.php:1730)
  at Contao\DC_Folder->editAll()
     (\vendor\contao\core-bundle\contao\classes\Backend.php:547)
  at Contao\Backend->getBackendModule('files', null)
     (\vendor\contao\core-bundle\contao\controllers\BackendMain.php:149)
  at Contao\BackendMain->run()
     (\vendor\contao\core-bundle\src\Controller\BackendController.php:44)
  at Contao\CoreBundle\Controller\BackendController->mainAction()
     (\vendor\symfony\http-kernel\HttpKernel.php:163)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (\vendor\symfony\http-kernel\HttpKernel.php:75)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (\vendor\symfony\http-kernel\Kernel.php:202)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (\web\index.php:44)
```

Test 1: Checkboxes "name" and "metadata" both activated
Test 2: Only checkbox "name" activated
Test 3: Only checkbox "metadata" activated

I cleared ProdCache in between, same results.

stefansl:
Same here

aschempp:
see https://github.com/contao/contao/pull/5463 (and https://github.com/contao/contao/pull/5462 for Contao 4)

PR #5462, opened 2022-11-01
Correctly check for excluded field in DC_Folder

aschempp:
See https://github.com/contao/contao/issues/5456

This appears to have been wrong since like forever… 🙈


Toflar (approved):
Further down we then use `$this->strField` because it's set `$this->strField = $v;` but here it's not yet. Probably a copy & paste issue ;)

PR #5463, opened 2022-11-01
Correctly check for excluded field in DC_Folder

aschempp:
Fixes https://github.com/contao/contao/issues/5456

same as https://github.com/contao/contao/pull/5462 but for Contao 5


Toflar (changes requested):
This change is actually correct but while we're on it, don`you think we should move the

```
if (!\in_array($v, $fields))
{
    continue;
}
```

check before the excluded and permissions check? No need to vote anything if it's not part of `$fields` anyway, is there?

aschempp:
changed in e80f9c230628f58fcc60cc3227ddd3ea0f2ad7d6

Toflar (approved):
Thanks!

ausi (approved):

PR #5465, opened 2022-11-02
Add end-to-end tests with Cypress

### LINKED ISSUE #5273: Discuss integration of Cypress tests
### Description

Following the comment in https://github.com/contao/contao/pull/5267#issuecomment-1245680941, we should discuss if we want to integrate Cypress tests in our suite and possibly the CI and how we should approach this. 

Cypress tests are (headless) browser based end-to-end tests, that work similar to manual testing by clicking stuff in the browser. IMHO they would be a great addition to our testing suite, especially because they would also cover things in the legacy codebase. Here is an example how they are written:

```js
describe('Users - back end modules', () => {
  // Visit back end
  before(() => { cy.visit(backend.routes.backend); })

  it('Load users', () => {
    // Go to Users
    cy.get('#tl_navigation a').contains('Users').click({force: true})
    // Make sure the first row has a user with username "admin"
    cy.get('table.tl_listing tr').first().get('td.col_username').contains('admin')
  })

	// …
})
```

If a test fails, you also get a screenshot or video. If running locally, there is also the option to run in "headed" mode, so that you can actually see what's happening in a real browser window.

@MDevster has created a prototype, that you can try right now: https://github.com/pdir/contao-cypress-testing. I also stole the above example from there. Big thanks for all that work so far (and to come :stuck_out_tongue:) :heart:. 

I got some ideas in my head how we could integrate this but added the `up for discussion` flag first to get some more input and opinions.

MDevster:
Fixes #5273

The test requires an empty Contao installation as described in the [development section](https://github.com/contao/contao/blob/5.x/README.md#development).

For the backend login we run contao:migrate and create a user. Then the tests are executed. After the tests, the database is reset and ready for new tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,28 @@ Then run the functional tests via the `run` command:
 composer run functional-tests
 ```
 
+## End to end tests with Cypress
+
+Install Cypress and copy `cypress.config.dev.js` from `vendor/contao/contao/cypress.config.dev.js` to root and rename to ``cypress.config.js`.
+
+
+```bash
+npm install cypress platform --save-dev
+cp vendor/contao/contao/cypress.config.dev.js ./cypress.config.js
+```
+
+Run e2e tests without ui.
+
+```bash
+npx cypress run
+```
+
+Run e2e tests with Cypress test suite.
+
+```bash
+npx cypress open
+```
+
 ## License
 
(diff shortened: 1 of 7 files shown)


ausi (approved):

leofeyer:
Thank you @MDevster.

PR #5466, opened 2022-11-02
add cypress to ci

MDevster:
If the PR https://github.com/contao/contao/pull/5465 is merged, cypress can be integrated into the ci.
We need to use the pr branch to run Cypress tests in ci.
The videos of tests will be in the summary section.

grafik


MDevster:
I need to do some tweaking here since it doesn't find the remote repo via the variables. I have to get the repo with `gh pr checkout`.

ausi:
I think this pull request can be closed in favor of #5465 and #6196

PR #5467, opened 2022-11-02
Correctly handle special characters when encoding domain names

leofeyer:


core-bundle/src/Resources/contao/library/Contao/Idna.php
```diff
@@ -38,7 +38,7 @@ public static function encode($strDomain)
 			return '';
 		}
 
-		if (($encoded = idn_to_ascii($strDomain)) === false)
+		if (($encoded = idn_to_ascii($strDomain, IDNA_NONTRANSITIONAL_TO_ASCII)) === false)
```
ausi, line 41: Some issues from other projects indicate that this constant might not always exist, should we a `defined()` check?

leofeyer: Yes. Or can we use the value of the constant instead? It should be `16`, right?

ausi:
```suggestion
		if (($encoded = idn_to_ascii($strDomain, defined(IDNA_NONTRANSITIONAL_TO_ASCII) ? IDNA_NONTRANSITIONAL_TO_ASCII : 16)) === false)
```

leofeyer: I just realized that the Symfony Intl polyfill will always define the constant, so I guess we do not need the `defined()` check after all: https://github.com/symfony/polyfill-intl-idn/blob/639084e360537a19f9ee352433b84ce831f3d2da/bootstrap.php#L79-L81

ausi: Reading more detailed it seems that the missing constants problem only exsisted in older PHP versions, so we should be fine here I think.

ausi (approved):

PR #5470, opened 2022-11-02
Use executeStatement() instead of query() in search

ausi:
Should fix the deprecations in #5468

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Database.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database.php
@@ -193,6 +193,18 @@ class Database
 	}
 
 	/**
+	 * Execute a statement and return the number of affected rows
+	 *
+	 * @param string $strQuery The query string
+	 *
+	 * @return int The number of affected rows
+	 */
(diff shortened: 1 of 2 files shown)


leofeyer (approved):

leofeyer:
Thank you @ausi.

core-bundle/src/Resources/contao/library/Contao/Database.php
```diff
@@ -192,6 +192,18 @@ public function execute($strQuery)
...
 	}
 
+	/**
+	 * Execute a statement and return the number of affected rows
+	 *
+	 * @param string $strQuery The query string
+	 *
+	 * @return int The number of affected rows
+	 */
+	public function executeStatement(string $strQuery): int
```
aschempp, line 202: shouldn't we also allow for parameters and types like doctrine does?

ausi: Don’t think so, this is merely a replacement for `->query()` that works with non-SELECT statements.
If you need parameters and types you should use the `database_connection` services instead anyways.

PR #4291, opened 2022-03-16
Remove the first bunch of BC layers

leofeyer:
These are just the low hanging fruits. There will be a lot more. 🙈

PR #5472, opened 2022-11-02
Remove dead code from the Database class

ausi:
…and deprecate magic `__get()`

We forgot to remove the `$arrCustomConfig` from the `getInstance()` method in #4291 I think, it does not work anymore in Contao 5 because the constructor argument was removed.


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #5474, opened 2022-11-02
Fix bug with cache in Database::listTables()

ausi:
`Database::listFields()` and `Database::listTables()` currently use the same array to cache its values. This results in incorrect data if you have a table with the same name as your database.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Database.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database.php
@@ -54,10 +54,16 @@ class Database
 	protected $blnDisableAutocommit = false;
 
 	/**
-	 * Cache
+	 * listFields Cache
 	 * @var array
 	 */
 	protected $arrCache = array();
+
+	/**
+	 * listTables Cache
+	 * @var array
+	 */
+	protected $arrTablesCache = array();
 
 	/**
 	 * Establish the database connection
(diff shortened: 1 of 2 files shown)


Toflar (approved):

aschempp (changes requested):
Since this is an "instance" cache, I don't think we need a database name cache key at all? Each different instance to a different database would get its own cache?

m-vo:
> Since this is an "instance" cache, I don't think we need a database name cache key at all? 

You can probably not even select a different database because it gets read by `Config::get('dbDatabase')`. :wink:. 

I think using two caches is still correct but the table cache could be of `list<string>` instead of `array<string, list<string>>`.

fritzmg:
> You can probably not even select a different database because it gets read by `Config::get('dbDatabase')`. 😉.

You can override that with

```php
$db = Database::getInstance(['host' => '…', 'database' => '…', …]);
```

m-vo:
> You can override that with

Sure you could, but it would still only be a single db per instance.

ausi:
> Since this is an "instance" cache, I don't think we need a database name cache key at all?

But the results are different, so we have to have a cache array per database, not per instance IMO. Calling `->listTables('foo')` and `->listTables('bar')` returns different results that still need to be cached.

> Each different instance to a different database would get its own cache?

Yes, of course. One `Database` instance represents one connection, and connections to different servers should definitely not share a cache.

***

This PR only splits up `$arrCache` into `$arrCache` and `$arrTablesCache` to fix the naming collision bug, everything else is the same as it always was.

aschempp (approved):
ah I didn't see there's a database argument on the `listTables` method 👍

leofeyer:
Thank you @ausi.

PR #5475, opened 2022-11-03
Handle cross-website-root redirects in the Frontend::jumpToOrReload method

qzminski:
For example, if the target page of a form is set to another website root, the resulting URL will be rendered incorrectly.

Example:

```
https://sub.domain.tld/domain.tld/en/confirmation.html
```

Whereas it should be:

```
https://domain.tld/en/confirmation.html
```


Toflar (approved):
LGTM

leofeyer:
Thank you @qzminski.

PR #5478, opened 2022-11-04
Add the `#[\SensitiveParameter]` attribute

ausi:


ausi:
~~Only one dependency seems to be blocking PHP 8.2 support, `scheb/2fa-backup-code`. It seems that they are waiting until the release of 8.2.0, see https://github.com/scheb/2fa/issues/162#issuecomment-1236371291~~

ausi:
Should I move the `#[\SensitiveParameter]` changes to the 5.1 milestone?
Symfony also categorized this as a feature: https://github.com/symfony/symfony/pull/46183

m-vo:
> Should I move the #[\SensitiveParameter] changes to the 5.1 milestone?

IMHO we can also merge this in 5.0. It's really just an attribute for quality, so no change in code/logic.

ausi:
The following dependencies are currently not compatible with PHP 8.2:

- [x] `laminas/laminas-code`  
  https://github.com/laminas/laminas-code/issues/142
- [x] `nette/schema`  
  ~~already compatible but not yet released~~  
  https://github.com/nette/schema/releases
- [x] `scheb/2fa-backup-code`, `scheb/2fa-bundle` and `scheb/2fa-trusted-device`  
  plans to release once PHP is released  
  https://github.com/scheb/2fa/issues/162#issuecomment-1236371291

ausi:
Contao can already be installed on PHP 8.2 by requiring `"scheb/2fa": "dev-php8.2"` in the root composer.json. All other packages are already compatible and everything seems to work fine 🚀

ausi:
This PR is RTM now, everything is compatible with PHP 8.2 🎉

Toflar (dismissed):

core-bundle/src/Image/Preview/PreviewFactory.php
```diff
@@ -43,12 +44,13 @@ public function __construct(
         private PictureFactoryInterface $pictureFactory,
         private Studio $imageStudio,
         private ContaoFramework $framework,
-        private string $secret,
+        #[\SensitiveParameter] string $secret,
```
leofeyer: This looks like a bug in PHP 8.2, doesn‘t it? Have you tried both `#[\SensitiveParameter] private string $secret` and `private #[\SensitiveParameter] string $secret`?

m-vo: Using `#[\SensitiveParameter] private string $secret` works, see https://3v4l.org/EVulG#v8.2.0.

There is a caveat, though: Currently only the call stack won't leak information, properties are not supported. If you dump the object, the string will be visible. So, adding the attribute here will only affect the code inside the constructor, which is… empty :smile:. Or am I missing something?

ausi: > Have you tried both `#[\SensitiveParameter] private string $secret` and `private #[\SensitiveParameter] string $secret`?

I thought I did, but as @m-vo demonstrated `#[\SensitiveParameter] private string $secret` seems to work. Not sure what I did wrong the first time... will change...

m-vo: Should we maybe wrap the value in a `\SensitiveParameterValue`? Like so: https://3v4l.org/hP5M8#v8.2.0

ausi: Done in 53b90bffa890a34e952600db63c699f5960d2d80

ausi:
Failing CI seems unrelated

Toflar (dismissed):

m-vo (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5485, opened 2022-11-08
Fix and improve mime type handling in the VFS

m-vo:
This PR…

* …makes the `MountManager` lazy-fetch file metadata (like the mime type) if it was not defined while listing contents of an adapter. We need to do this, because adapters aren't guaranteed to set this data.

* …allows suppressing errors from the mime type guesser when calling `FilesystemItem#getMimeType()`. This comes handy when outputting mime types from a `FilesystemItem` in a template. Mime type guessing can for example fail if a file has no content.

* …fixes an edge case where querying the `MountManager` for the root path `''` with no root adapter mounted lead to an endless loop instead of an exception.


core-bundle/tests/Filesystem/FilesystemItemTest.php
```diff
@@ -196,6 +197,24 @@ static function () use (&$invocationCounts): string {
...
+            true,
+            'some/file.txt',
+            null,
+            null,
+            static function (): void {
+                throw VirtualFilesystemException::unableToRetrieveMetadata('some/file.txt');
+            }
+        );
+
+        $this->expectException(VirtualFilesystemException::class);
```
Toflar, line 212: Afaik we usually put the expected exception calls first in the tests :)

Toflar (dismissed):

core-bundle/tests/Filesystem/FilesystemItemTest.php
leofeyer: No, you are confusing this with `$this->expectDeprecation()`. The position of the `$this->expectException()` call is not arbitrary and should be as close to the call triggering the exception as possible.

core-bundle/src/Filesystem/MountManager.php
```diff
@@ -402,12 +402,14 @@ private function getAdapterAndPath(string $path): array
     {
         $prefix = $path;
 
-        // Find adapter with the longest (= most specific) matching prefix
-        do {
-            if (null !== ($adapter = $this->mounts[$prefix] ?? null)) {
-                return [$adapter, Path::makeRelative($path, $prefix), $prefix];
-            }
-        } while ('.' !== ($prefix = \dirname($prefix)));
+        if ('' !== $path) {
```
ausi: Should `$prefix = $path;` be moved inside the `if` block?

ausi (dismissed):

core-bundle/src/Filesystem/MountManager.php
m-vo: See 1ffee24e16e05bbf9bbfa30bb01c2ba9c01e4338.

ausi (dismissed):

core-bundle/src/Filesystem/FilesystemItem.php
```diff
@@ -156,10 +156,21 @@ public function getFileSize(): int
         return $this->fileSize ?? 0;
     }
 
-    public function getMimeType(): string
+    public function getMimeType(bool $throwOnError = true): string
```
leofeyer: Isn‘t this considered an anti-pattern? You should always throw the exception and wrap the call in a try-catch block instead.

m-vo: Well, by default this is the case: if you use items in your code and you rely on having a mime type, you can use `getMimeType()` and the exception will bubble up. But if the value is optional, `getMimeType(false)` is much more convenient. 

The latter is also crucial when outputting items in templates, e.g. to add a custom icon class depending on the type:
```twig
<span class="icon icon--{{ item.mimeType(false) }}">{{ item.name }}</span>
```

ausi: How about a `$default` parameter, and the method only throws if no default is provided? I think this concept is pretty common?

```suggestion
    public function getMimeType(string|null $default = null): string
```

m-vo: Good idea, implemented in 780b0f34d3bbf7403b525e80201997da53c2f2d7.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5486, opened 2022-11-08
Correctly copy templates from symlinked sources

m-vo:
This PR build on top of #5270 and also fixes the creation of templates from symlinked sources.

This includes two things: 
* I removed an unnecessary call for `Validator::isInsecurePath()`. The post data (source) is already checked against the array of options a few lines down.
* The file now gets copied by the `Filesystem` class instead of the `Files` class, because the latter does not allow anything outside the project dir.

### DIFF
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -312,6 +312,7 @@ class tl_templates extends Contao\Backend
 							Path::makeAbsolute($strOriginal, $projectDir),
 							Path::makeAbsolute($strTarget, $projectDir)
 						);
+
 						$this->redirect($this->getReferer());
 					}
 				}
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -266,7 +266,6 @@ class tl_templates extends Contao\Backend
 		// Copy an existing template
 		if (Contao\Input::post('FORM_SUBMIT') == 'tl_create_template')
 		{
-			$strOriginal = Contao\Input::post('original', true);
 			$strTarget = Contao\Input::post('target', true);
 
 			if (Contao\Validator::isInsecurePath($strTarget))
@@ -282,6 +281,7 @@ class tl_templates extends Contao\Backend
 			else
 			{
 				$blnFound = false;
+				$strOriginal = Contao\Input::post('original', true);
 
 				// Validate the source path
 				foreach ($arrAllTemplates as $arrTemplates)
--- a/core-bundle/src/Resources/contao/dca/tl_templates.php
+++ b/core-bundle/src/Resources/contao/dca/tl_templates.php
@@ -336,7 +336,7 @@ class tl_templates extends Contao\Backend
 		// Show form
 		return ($strError ? '
 <div class="tl_message">
-<p class="tl_error">' . $strError . '</p>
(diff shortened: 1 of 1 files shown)


ausi (dismissed):

leofeyer:
> * I removed an unnecessary call for `Validator::isInsecurePath()`. The post data (source) is already checked against the array of options a few lines down.

Where exactly is that?

m-vo:
> Where exactly is that?

Here: https://github.com/contao/contao/blob/e6d74d62337e9dd44859c7201718d99da5657983/core-bundle/src/Resources/contao/dca/tl_templates.php#L289-L299

If `$blnFound` is not set to `true`, an error will be shown and no files copied.

leofeyer:
But now the insecure path is written to `$strError` and output in line 342 instead of triggering an exception. Not sure if that is better?

m-vo:
> But now the insecure path is written to `$strError` and output in line 342 

This is a good point, we should definitively encode characters in there. Alternatively, we could probably either remove the placeholder from the `tl_templates.invalid` translation or the whole translation key altogether (because it should never happen without messing with the form itself). Wdyt?

leofeyer:
I think that this is getting unnecessarily complicated. Just revert all changes but the `(new Filesystem())->copy()` change and we are good to merge this.

m-vo:
But all changes are needed to solve the issue? Otherwise an exception is thrown when selecting a file from symlinked resources and submitting.

leofeyer:
Why is that? Is the symlinked path treated as insecure?

m-vo:
Yes, because it begins with `../`.

leofeyer:
I see. Then we should probably keep the check but allow `../`:

```php
// Allow the source path to begin with ../ in case it is a symlinked resource
if (strncmp($strOriginal, '../', 3) !== 0 && Contao\Validator::isInsecurePath($strOriginal))
{
	throw new RuntimeException('Invalid path ' . $strOriginal);
}
```

WDYT?

m-vo:
Hmm, I don't think the `isInsecurePath()` check should have been in there in the first place. If the value is in the options (that were built without any user input), it's never a problem. If we keep the check, you would either get an exception or an error message if you are messing with the values. To me we should either always output the error message (as it is now) or remove the error message completely and always throw an exception.

ausi:
> I don't think the `isInsecurePath()` check should have been in there in the first place.

I agree. `$strOriginal` is only accepted if it is in the list of possible values, so checking if this path is secure at this point doesn’t make any sense.

> But now the insecure path is written to `$strError` and output in line 342 instead of triggering an exception. Not sure if that is better?

“insecure” means in that case that it has something like `../` in the path, not that it is insecure in terms of HTML special characters. But adding the `specialchars()` call is correct regardless of that, because even “secure” paths can contain special HTML characters.

ausi (dismissed):

core-bundle/src/Resources/contao/dca/tl_templates.php
```diff
@@ -266,12 +267,6 @@ public function addNewTemplate()
 		if (Contao\Input::post('FORM_SUBMIT') == 'tl_create_template')
 		{
 			$strOriginal = Contao\Input::post('original', true);
```
leofeyer: I guess we can move this line to line 285 now, can‘t we? Right below `$blnFound = false;`.

ausi: Agree.

leofeyer (approved):

leofeyer:
Thank you @m-vo.

PR #5487, opened 2022-11-08
Render variant templates in the back end

m-vo:
If you select a template variant, it was not rendered in the back end. This PR removes this limitation and only falls back to the default template if the template does not exist (as a Twig template).


ausi (dismissed):

ausi:
> … only falls back to the default template if the template does not exist …

But it does this now also in the frontend, right?

m-vo:
> But it does this now also in the frontend, right?

Yes. But on a second thought, I might have to slightly adjust the logic to make it still work with legacy variant templates. 🤔

m-vo:
I adjusted the logic again - this way legacy variant templates keep working as they did before.

Now: if the selected variant template is a legacy one, we simply keep the old logic, i.e. not rendering in the back end, always rendering (and potentially failing) in the front end. For modern templates, we always fall back to the default template if not existing no matter if front or back end.

ausi (approved):

leofeyer:
Thank you @m-vo.

PR #5489, opened 2022-11-10
Do not confirm password changes in the widget

leofeyer:
The reason why we have to use the ugly `Messages::reset()` hack in #5415 is that the password widgets wrongly assumes that a password has been changed if it **validates**. However, only the controller that actually **saves** the password can know this, so only the controller should add the confirmation message.

The `BackendPassword` controller already does this correctly. This PR adds the same functionality to the `DataContainer` class.


fritzmg (approved):

ausi (approved):

aschempp:
Does this still work for the frontend password change module (and probably `ModulePersonalData`) now?

fritzmg:
`ModulePersonalData` and `ModulePassword` do not collect messages from `Message`, or do you mean something else?

PR #5490, opened 2022-11-10
Pass the unmodified row to the onshow callback

ausi:
In Contao 4 the `onshow_callback` received the row as retrieved from the database without modifications.
This was changed in #4770, by accident I believe.

Contao 4: `[… "type" => "headline" …]`
Contao 5 before this PR: `[… "type" => "Überschrift" …]`


leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5496, opened 2022-11-11
Fix and improve mime type handling in the VFS

ausi:
Same as #5485 for Contao 5

### DIFF
--- a/core-bundle/contao/templates/_new/component/_download.html.twig
+++ b/core-bundle/contao/templates/_new/component/_download.html.twig
@@ -41,7 +41,7 @@
         {% set dowload_link_attributes = attrs(dowload.link_attributes|default)
             .set('href', download.href)
             .set('title','MSC.download'|trans([title]))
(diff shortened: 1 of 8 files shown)


core-bundle/contao/templates/_new/component/_download.html.twig
```diff
@@ -41,7 +41,7 @@
         {% set dowload_link_attributes = attrs(dowload.link_attributes|default)
             .set('href', download.href)
             .set('title','MSC.download'|trans([title]))
-            .setIfExists('type', download.file.mimeType|default) %}
+            .setIfExists('type', download.file.mimeType('')) %}
```
ausi, line 44: @m-vo is this correct here?

m-vo: Yes, perfect. :+1:

m-vo (approved):

leofeyer:
Thank you @ausi.

PR #5498, opened 2022-11-12
Handle broken images in FigureBuilder#buildIfResourceExists()

m-vo:
When the resizer tries to open an invalid image file, it will throw an exception. This is all fine except for the case that the `FigureBuilder` was used with `buildIfResourceExists()`. Due to the lazy nature, we do get a `Figure` object (instead of `null`) and the exception will only appear once accessing the image/picture data.

To solve this issue, we are now accessing `$figure->getImage()->getOriginalDimensions();` before returning the object, which makes the underlying system open the image (no resize yet) and then catch the exception/return null in case of an error. Props to @ausi for this idea.

### DIFF
--- a/core-bundle/src/Image/Studio/FigureBuilder.php
+++ b/core-bundle/src/Image/Studio/FigureBuilder.php
@@ -512,7 +512,18 @@ class FigureBuilder
             return null;
         }
 
-        return $this->doBuild();
+        $figure = $this->doBuild();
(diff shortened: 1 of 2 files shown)


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #5499, opened 2022-11-12
Fix automatic DBAFS sync for root resources

m-vo:
There was a bug with automatic DBAFS syncing if the updated resource was situated in the search root: When the `Dbafs` service was analyzing at the filesystem, resources didn't get yielded as "file" then but as "non-existing". This meant files would not be tracked when calling `$filesystem->move('a', 'b');`. 

I also fixed a related bug for DBAFSs without a prefix: There, when searching for an empty prefix (`''`), the logic would still add a leading `/` making the query not return any results.

The added functional test case now covers both cases.

### DIFF
--- a/core-bundle/src/Filesystem/Dbafs/Dbafs.php
+++ b/core-bundle/src/Filesystem/Dbafs/Dbafs.php
@@ -790,16 +790,16 @@ class Dbafs implements DbafsInterface, ResetInterface
             }
 
             if (null === ($isDir = $analyzedPaths[$searchPath] ?? null)) {
-                if (!$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
-                    // Yield file
-                    yield $searchPath => self::RESOURCE_FILE;
-
-                    continue;
-                }
-
                 // Analyze parent path
                 $analyzedPaths = array_merge($analyzedPaths, $analyzeDirectory(Path::getDirectory($searchPath)));
                 $isDir = $analyzedPaths[$searchPath] ??= false;
+            }
+
+            if (!$isDir && !$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
+                // Yield existing file
+                yield $searchPath => self::RESOURCE_FILE;
+
+                continue;
             }
 
             if ($isDir) {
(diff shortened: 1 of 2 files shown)


core-bundle/src/Filesystem/Dbafs/Dbafs.php
```diff
@@ -788,18 +790,18 @@ private function getFilesystemPaths(array $searchPaths, array $parentDirectories
...
 
+            if (!$isDir && !$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
+                // Yield existing file
+                yield $searchPath => self::RESOURCE_FILE;
+
+                continue;
+            }
+
             if ($isDir) {
                 yield from $traverseRecursively($searchPath, $shallow);
```
ausi:
```suggestion
            if ($isDir) {
                yield from $traverseRecursively($searchPath, $shallow);

                continue;
            }
            
            if (!$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
                // Yield existing file
                yield $searchPath => self::RESOURCE_FILE;

```

The `!$isDir` check would not be needed if we move the code after the `if ($isDir)` early return I think?

m-vo: True, that. Thanks for the suggestion. :heart:

core-bundle/src/Filesystem/Dbafs/Dbafs.php
```diff
@@ -805,6 +800,13 @@ private function getFilesystemPaths(array $searchPaths, array $parentDirectories
 
                 continue;
             }
+
```
m-vo:
```suggestion

```

core-bundle/src/Filesystem/Dbafs/Dbafs.php
```diff
@@ -805,6 +800,12 @@ private function getFilesystemPaths(array $searchPaths, array $parentDirectories
 
                 continue;
             }
+            if (!$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
```
m-vo, line 804:
```suggestion

            if (!$shallow && $this->filesystem->fileExists($searchPath, VirtualFilesystemInterface::BYPASS_DBAFS)) {
```

ausi (approved):

Toflar (approved):

leofeyer:
Thank you @m-vo.

PR #5500, opened 2022-11-12
Display results when reloading finished crawler

### LINKED ISSUE #4157: Update search index - download debug log and search index log again
### Description

Affects system maintenance @ Contao 4.9 and 4.13.

Is there a way to download the two logs again in the "Update search index" feature? Even if you left the page in the backend in the meantime or after you get logged out?

The problem is that in a project, e.g. 26,000 pages are crawled and that takes over 1 hour. In the meantime, of course, I don't want to wait. 
After a while I go back to the browser tab of this website again. It often happens that you are already logged out or want to look at something else in the backend. And then you get really annoyed that you can't see the result of the search index crawler tool and additionally you can't download the logs.
You have to check and wait again over 1 hour... 

So could you please create a function to cache the last result or save the log files on the server? There's a nice folder /var/logs - something like that could go in there, couldn't it?

ausi:
Fixes #4157

### DIFF
--- a/core-bundle/src/Resources/contao/classes/Crawl.php
+++ b/core-bundle/src/Resources/contao/classes/Crawl.php
@@ -174,6 +174,37 @@ class Crawl extends Backend implements \executable
 		}
 		catch (InvalidJobIdException $e)
 		{
+			if (file_exists($resultCache))
+			{
+				$results = json_decode(file_get_contents($resultCache), true);
+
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5511, opened 2022-11-16
Fix undefined array key access

### LINKED ISSUE #5509: Frontend User groups check first if array is empty.
### Affected version(s)

4.13.12

### Description

Contao 4.13.12
PHP: 8.1

Please check if the member groups array is empty before accessing its 0th element.

Screenshot 2022-11-15 at 17 47 45

bytehead:
Fixes #5509

### DIFF
--- a/core-bundle/src/Resources/contao/classes/FrontendUser.php
+++ b/core-bundle/src/Resources/contao/classes/FrontendUser.php
@@ -237,7 +237,7 @@ class FrontendUser extends User
 		}
 
 		// Get the group login page
-		if ($this->groups[0] > 0)
+		if (\count($this->groups) && $this->groups[0] > 0)
 		{
 			$objGroup = MemberGroupModel::findPublishedById($this->groups[0]);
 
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/FrontendUser.php
```diff
@@ -237,7 +237,7 @@ protected function setUserFromDb()
 		}
 
 		// Get the group login page
-		if ($this->groups[0] > 0)
+		if (\count($this->groups) && $this->groups[0] > 0)
```
ausi: Is `$this->groups` guaranteed to be indexed starting with zero? Otherwise I’d just use:
```suggestion
		if (($this->groups[0] ?? 0) > 0)
```

bytehead: I've never seen any different - but who knows what's out there?

bytehead: Adjusted in b367929334.

ausi (approved):

leofeyer:
Thank you @bytehead.

PR #5520, opened 2022-11-20
Improve the stability of the JSON export for the "user list" command

### LINKED ISSUE #4937: contao:user:list bugs with JSON format
### Affected version(s)

4.13.5

### Description

Running `contao:user:list --format=json` results in an error: 
`Malformed UTF-8 characters, possibly incorrectly encoded` 
(but this error is swallowed because we’re not using `JSON_THROW_ON_ERROR` and the output is just empty)

Running `contao:user:list --format=json --column=username --admins` results in the following output:

```
 ! [NOTE] Only showing admin accounts

[{"username":"admin"}]
```

which is invalid JSON.

IMO we should:
- Use `JSON_THROW_ON_ERROR`
- Base64 encode columns containing binary data, or skip them
- Make sure all output is JSON if `--format=JSON` is used (including errors, warnings and notices)

/cc @richardhj

richardhj:
This PR improves stability for the JSON output of the `contao:user:list --format=json` command.

In particular:

- Use JSON_THROW_ON_ERROR to not silence unexpected errors
- Use the same default columns as for text output (fixes errors when trying to export columns with binary data, i.e., `secret`).
- Silence informational output, so the output is JSON only

This PR closes #4937.

### DIFF
--- a/core-bundle/src/Command/UserListCommand.php
+++ b/core-bundle/src/Command/UserListCommand.php
@@ -53,7 +53,7 @@ class UserListCommand extends Command
     {
         $io = new SymfonyStyle($input, $output);
 
-        if ($input->getOption('admins')) {
+        if ($input->getOption('admins') && 'json' !== $input->getOption('format')) {
             $io->note('Only showing admin accounts');
         }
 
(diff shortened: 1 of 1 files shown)


ausi (approved):

Toflar (approved):

leofeyer:
Thank you @richardhj.

PR #5526, opened 2022-11-23
Correctly sort tree items when PID can be null

aschempp:
In https://github.com/contao/contao/pull/5142 we added support for nullable PIDs in tree views, because that is required when using foreign keys (e.g. with Doctrine relations). Unfortunately, it is currently not possible to re-sort items in the tree because of incorrect PID checks.

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -1189,7 +1189,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 
 					if ($limit > 0)
 					{
-						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid == 0 ? '(pid=? OR pid IS NULL)' : 'pid=?') . " ORDER BY sorting, id")
+						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid ? 'pid=?' : '(pid=? OR pid IS NULL)') . " ORDER BY sorting, id")
 														 ->limit(1, $limit - 1)
 														 ->execute($pid);
 
@@ -1206,7 +1206,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 				{
 					$newPID = $pid;
 
-					$objSorting = $this->Database->prepare("SELECT MIN(sorting) AS sorting FROM " . $this->strTable . " WHERE " . ($pid == 0 ? '(pid=? OR pid IS NULL)' : 'pid=?'))
+					$objSorting = $this->Database->prepare("SELECT MIN(sorting) AS sorting FROM " . $this->strTable . " WHERE " . ($pid ? 'pid=?' : '(pid=? OR pid IS NULL)'))
 												 ->execute($pid);
 
 					// Select sorting value of the first record
@@ -1217,7 +1217,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 						// Resort if the new sorting value is not an integer or smaller than 1
 						if (($curSorting % 2) != 0 || $curSorting < 1)
 						{
-							$objNewSorting = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid == 0 ? '(pid=? OR pid IS NULL)' : 'pid=?') . " ORDER BY sorting, id")
+							$objNewSorting = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid ? 'pid=?' : '(pid=? OR pid IS NULL)') . " ORDER BY sorting, id")
 															->execute($pid);
 
 							$count = 2;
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1189,7 +1189,7 @@ protected function getNewPosition($mode, $pid=null, $insertInto=false)
 
 					if ($limit > 0)
 					{
-						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE pid=? ORDER BY sorting, id")
+						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid == 0 ? '(pid=? OR pid IS NULL)' : 'pid=?') . " ORDER BY sorting, id")
```
```suggestion
						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE COALESCE(pid, 0)=? ORDER BY sorting, id")
```
maybe this could be a simpler version? (untested)

aschempp: feels less readable and not sure about performance. If there is no technical advantage I don't think we should use that.

richardhj (dismissed):

core-bundle/src/Resources/contao/drivers/DC_Table.php
ausi: We could also use `IFNULL()` if `COALESCE()` is less readable.

```suggestion
						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE IFNULL(pid, 0)=? ORDER BY sorting, id")
```

But `IFNULL(pid, 0)=?` is way more readable then `" . ($pid == 0 ? '(pid=? OR pid IS NULL)' : 'pid=?') . "` don’t you think?

aschempp: do you think that's more/equally performant as a simple OR statement?

ausi: Yes. (But I also think that the performance for that part is not relevant here)
But I just saw that we’d need to change `->execute($pid)` then to `->execute((int) $pid)` too, so maybe that wouldn’t be better then.

How about at least removing the scary `== 0` part?

```suggestion
						$objInsertAfter = $this->Database->prepare("SELECT id FROM " . $this->strTable . " WHERE " . ($pid ? 'pid=?' : '(pid=? OR pid IS NULL)') . " ORDER BY sorting, id")
```

aschempp: > How about at least removing the scary == 0 part?

`$pid` in this case refers to a string `'0'` which would not equal to boolean false, so that won't work.

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1312,8 +1308,13 @@ protected function getNewPosition($mode, $pid=null, $insertInto=false)
 					}
 				}
 
+				if (!$newPID)
+				{
+					$newPID = Widget::getEmptyValueByFieldType($GLOBALS['TL_DCA'][$this->strTable]['fields']['pid']['sql'] ?? array()) === null ? null : 0;
```
leofeyer, line 1317: I don‘t understand the ` === null ? null : 0` part. What else except `null` and `0` could possibly be returned for an INT() field?

core-bundle/src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1258,50 +1258,46 @@ protected function getNewPosition($mode, $pid=null, $insertInto=false)
 						$newPID = $objSorting->pid;
 						$curSorting = $objSorting->sorting;
 
-						// Do not proceed without a parent ID
-						if (is_numeric($newPID))
```
leofeyer, line 1262: As @ausi pointed out, wouldn't it be much easier to use `is_numeric($newPID) || $newPID === null` here? This would solve the problem in a less invasive way.

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: the widget class returns an empty string by default if there is no definition. This makes sure the value is either NULL or 0

core-bundle/src/Resources/contao/drivers/DC_Table.php
aschempp: you mean instead of removing the if-condition? I don't think this was ever intended to catch a string-value

leofeyer: Yes, keeping the if condition and allowing `null` in addition to numeric values.

aschempp: even though the comment explicitly said to check for "without a parent ID"??

ausi: `0` and `null` are valid parent IDs, but the empty string is not I think, so the comment would be somewhat correct.
If the `if` is needed at all is hard to tell IMO, therefore I’d prefer the less invasive way with the minimal change.

ausi (approved):

PR #5528, opened 2022-11-24
Increase padding on selects in Firefox

fritzmg:
Previously `<select>`s in Firefox always had a hidden padding that you could not remove. Thus in our `main.css` we have added

```css
@-moz-document url-prefix() {
	.tl_select,.tl_mselect,.tl_select_column,.tl_select_unit,.tl_select_interval {
		padding-left:2px;
	}
}
```

to reduce the CSS padding to account for the non-removable padding in Firefox.

However, since the recent Firefox versions this is not the case anymore:

This PR fixes that by removing the Firefox specific CSS that reduces the left padding on `<select>`s in the back end, so that it will look like this again:


Toflar (approved):

aschempp (approved):

ausi (approved):

leofeyer:
Thank you @fritzmg.

PR #5532, opened 2022-11-25
Make sure text content is always a string

aschempp:


Toflar (dismissed):

bytehead (dismissed):

leofeyer (changes requested):
Please also adjust the `@property` annotation to `string|null`.

https://github.com/contao/contao/blob/f12775f117d6e770005dba54010de84ee1f29481/core-bundle/src/Resources/contao/elements/ContentElement.php#L25

leofeyer (approved):

PR #5534, opened 2022-11-25
Fix javascript error in popups

### LINKED ISSUE #5508: Selection of important part of the image is incorrect
### Affected version(s)

4.13.13

### Description

After updating to Contao 4.13.13, I now have the error that the important part of the image can only be defined via the menu item "System > File management (in german: Dateiverwaltung)".

For example, if I select an image in the Contao content element "Gallery" and want to define or change the important part of it, the mouse selection is not available. The cursor does not turn into a crosshair (under Windows).

More information:
https://github.com/contao/contao/pull/5437

image

ausi:
Fixes #5508


bytehead (approved):

fritzmg (approved):

leofeyer:
Thank you @ausi.

PR #5537, opened 2022-11-28
Fix a PHP 8 issue in the PageTree widget

Tastaturberuf:
Fix `Warning: Trying to access array offset on value of type null` if there is no label set.

If you have no label on a picker field (mostly in development) you became this warning.


bytehead (approved):

leofeyer:
Thank you @Tastaturberuf.

PR #5542, opened 2022-11-30
Make $GLOBALS more understandable for PHPStan

leofeyer:
See https://github.com/contao/contao/pull/5416#issuecomment-1329607513

### DIFF
--- a/core-bundle/tests/EventListener/DataContainer/DefaultOperationsListenerTest.php
+++ b/core-bundle/tests/EventListener/DataContainer/DefaultOperationsListenerTest.php
@@ -44,7 +44,7 @@ class DefaultOperationsListenerTest extends TestCase
 
     public function testAddsDefaultOperations(): void
     {
-        /** @var array $GLOBALS (signals PHPStan that the value is not immutable) */
+        /** @var array $GLOBALS (signals PHPStan that the array shape may change) */
         $GLOBALS['TL_DCA']['tl_foo'] = [
             'list' => [
                 'sorting' => [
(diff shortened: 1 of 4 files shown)


aschempp (dismissed):

core-bundle/tests/EventListener/DataContainer/DefaultOperationsListenerTest.php
```diff
@@ -44,6 +44,7 @@ protected function tearDown(): void
 
     public function testAddsDefaultOperations(): void
     {
+        /** @var array $GLOBALS (signals PHPStan that the value is not immutable) */
```
ausi: I think this is technically not 100% accurate.
It is more about that the inferred type (the [array shape](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) of `$GLOBALS`) changes outside of this scope, not the value itself.

```suggestion
        /** @var array $GLOBALS (signals PHPStan that the array shape may change) */
```

```suggestion
        /** @var array $GLOBALS (signals PHPStan that the type can not be inferred) */
```

Not sure. I’m fine either way or also keeping it as is.

ausi (dismissed):

bytehead (dismissed):

core-bundle/tests/EventListener/DataContainer/DefaultOperationsListenerTest.php
leofeyer: Adjusted in 2e1dc26ad18c73089824c6ae9525b7579919b19c.

PR #5544, opened 2022-11-30
Merge CSS classes in the "root page dependent modules" module

### LINKED ISSUE #5535: Insert module "root page dependent modules" via content element does not merge css classes
### Affected version(s)

4.13

### Description

When the Root Page Dependent Module is inserted via a content element, css classes are not merged.

### Reproducible by creating the following modules
- Module A: css class `module-a`
- Module B: css class `module-b`
- Module C (Root page dependent modules): with A and B
- Content Element (Module): Inserts module C and assigns css class `test`

Only the classes `module-a` or `module-b` are output, `test` is ignored

Is this the expected behavior? Then we should consider removing the ID/CSS class selection for this content element.

bytehead:
Fixes #5535

### DIFF
--- a/core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
+++ b/core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
@@ -46,7 +46,7 @@ class RootPageDependentModulesController extends AbstractFrontendModuleControlle
             $cssID[0] = substr($idAttribute, 5, -1);
         }
 
-        $cssID[1] = trim(sprintf('%s %s', $cssID[1], implode(' ', $model->classes)));
+        $cssID[1] = trim(sprintf('%s %s', $cssID[1] ?? '', implode(' ', $model->classes)));
         $module->cssID = $cssID;
 
         $controller = $this->container->get('contao.framework')->getAdapter(Controller::class);
--- a/core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
+++ b/core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
@@ -37,8 +37,15 @@ class RootPageDependentModulesController extends AbstractFrontendModuleControlle
             return new Response('');
         }
 
+        $moduleModel = $this->container->get('contao.framework')->getAdapter(ModuleModel::class);
+        $module = $moduleModel->findByPk($modules[$pageModel->rootId]);
+
+        $cssID = StringUtil::deserialize($module->cssID, true);
+        $cssID[1] = trim(sprintf('%s %s', $cssID[1], implode(' ', $model->classes)));
+        $module->cssID = $cssID;
+
         $controller = $this->container->get('contao.framework')->getAdapter(Controller::class);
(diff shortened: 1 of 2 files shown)


core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -37,8 +37,15 @@ public function __invoke(Request $request, ModuleModel $model, string $section,
             return new Response('');
         }
 
+        $moduleModel = $this->container->get('contao.framework')->getAdapter(ModuleModel::class);
+        $module = $moduleModel->findByPk($modules[$pageModel->rootId]);
+
+        $cssID = StringUtil::deserialize($module->cssID, true);
```
ausi, line 46: The ID should get overwritten if set I think.

```suggestion
        $cssID = StringUtil::deserialize($module->cssID, true);
        if ($idAttribute = $request->attributes->get('templateProperties', [])['cssID'] ?? null) {
            $cssID[0] = substr($idAttribute, 5, -1);
        }
```

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -37,8 +37,15 @@ public function __invoke(Request $request, ModuleModel $model, string $section,
...
+        $cssID[1] = trim(sprintf('%s %s', $cssID[1], implode(' ', $model->classes)));
```
ausi: Possibly undefined array key.

```suggestion
        $cssID[1] = trim(sprintf('%s %s', $cssID[1] ?? '', implode(' ', $model->classes)));
```

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
```diff
@@ -37,8 +37,15 @@ public function __invoke(Request $request, ModuleModel $model, string $section,
...
+        $module->cssID = $cssID;
```
ausi, line 54: `cssID` in the model is usually serialized AFAIK.

```suggestion
        $module->cssID = serialize($cssID);
```

bytehead: It is, but not needed, see https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/elements/ContentModule.php#L78.

I tried both, makes no difference 😅

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
bytehead: Adjusted in 35fa17cb4f.

core-bundle/src/Controller/FrontendModule/RootPageDependentModulesController.php
bytehead: Fixed in 8d4482a4a5.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @bytehead.

PR #5548, opened 2022-12-01
Fix the missing module references in the newsletter-bundle hooks

Toflar:
Both, the `activateRecipient` as well as the `removeRecipient` hooks lack the module instance making it impossible to apply logic based on module settings.

This is possible in basically all our other module hooks so I'd consider this a bug :)


ausi (approved):

bytehead (approved):

fritzmg (approved):

aschempp (approved):

leofeyer:
Thank you @Toflar.

PR #5554, opened 2022-12-07
Add support for async CLI cron jobs

Toflar:
This implements support for asynchronous cron jobs for the CLI scope.

Currently, all cron jobs are run synchronously and sequentially which means that if you e.g. have 10 cron jobs and they all take 20 seconds to run, it will take 200 seconds to complete.
For most cronjobs, this is not a problem because they don't usually run 20 seconds.

However, if you have cronjobs that trigger child processes or are asnyc in any other way, you would want them to start immediately in parallel.
What we can do now is this:

```php
#[AsCronJob('minutely')]
class MessengerCron
{
    public function __invoke(string $scope): ?PromiseInterface
    {
        if ($scope !== Cron::SCOPE_CLI) {
            return null;
        }

        $promise = new Promise(static function () use (&$promise): void {$promise->resolve('promise'); });

        return $promise;
    }
}
```

This will allow you to e.g. start a `Process` immediately and then return a promise that waits for it to finish.

I've implemented a helper class for Symfony `Process` instances as well. I put it into `ProcessUtil` because this is a place where I will add more methods (e.g. helpers to run a process with the current PHP binary).

What you can do with it: 

```php
#[AsCronJob('minutely')]
class MessengerCron
{
    public function __invoke(string $scope): ?PromiseInterface
    {
        if ($scope !== Cron::SCOPE_CLI) {
            return null;
        }

        return ProcessUtil::createPromise(new Process(['ls']));
    }
}
```

This will automatically start the process and create a promise for it.

This is foundational work for https://github.com/contao/contao/pull/5405.


ausi:
> What happens now is that the `Process` instances will be started (`$process->start()`) first, then all synchronous cronjobs are being worked on as it happens already today, then it will wait for the async processes to finish.

How about starting the async jobs last? This way we could just check the return type of the cron and would not need the `isAsync` method at all.

This would also make it potentially more stable against race conditions if we consider synchronous cronjobs as “dangerous” if they clear or build caches.

Toflar:
> How about starting the async jobs last? This way we could just check the return type of the cron and would not need the `isAsync` method at all.

The whole goal of this PR is to make sure that they are **NOT** started at last? :D You want to start them immediately. Otherwise if you have a minutely job, it could take 10 minutes until you actually start running which is exactly what I want to omit.

ausi:
But that would mean that e.g. the `PurgeTempFolderCron` runs at the same time as all asynchronous jobs do, which potentially use the tmp folder themself.

I’d prefer to run the sync jobs alone first and use an async job for everything that requires more time. This way the _starting 10 minutes too late_ issue should never occur.

But I can see that this is a trade-off as you loose the guaranteed start time.

Toflar:
> But I can see that this is a trade-off as you loose the guaranteed start time.

Which is exaclty what we need for the messenger.

Toflar:
> But that would mean that e.g. the `PurgeTempFolderCron` runs at the same time as all asynchronous jobs do, which potentially use the tmp folder themself.

Can also happen in other areas. You can also have the web process that runs a cron and you trigger the same cron on CLI at the same time. This is not an issue of this PR, it can happen anywhere and always. If we ever encounter a real issue with that, we should use locking for the specific components.

core-bundle/src/Cron/CronJob.php
```diff
@@ -29,13 +29,29 @@ public function __construct(private object $service, private string $interval, p
...
+        }
+
+        $ref = new \ReflectionMethod($this->getService(), $this->getMethod());
+        $returnType = $ref->getReturnType();
+
+        if (!$returnType instanceof \ReflectionNamedType) {
+            return false;
+        }
+
+        return ProcessCollection::class === $returnType->getName();
```
fritzmg: Rather than using reflection - should we may be introduce an interface instead? That way `CronJob::isAsync` could simply return `$this->service instanceof AsyncCronJobInterface` - or the interface itself could mandate an `isAsync` method that the service has to implement. Just a thought.

ausi: How about adding `isAsync` to the `AsCronJob` attribute and the associated tag?

Toflar: I would probably indeed add an interface. You need to ensure the correct return type. Just adding the attribute doesn't fulfil any contract.

Toflar: I'll give the interface variant a try - let's see how we like that :)

fritzmg: Though I get why the reflection is used - since it ultimately depends on what the CronJob's method is returning.

ausi: But the method name can be chosen freely so an enforcing interface is not possible I think.

But the `AsCronJob` attribute can verify the method signature in the `AddCronJobsPass` (this is also done for some other tags AFAIK)

Toflar: I think we can just ignore the method configuration if you implement the interface ;)

ausi: I think controlling this via an interface would be a confusing mixture of configuration. Everything else is configured by the `contao.cronjob` tag, so why not configure `isAsync` there too?

fritzmg: Agreed. It would be weird if you put your `AsCronJob` attribute on a method - but then were forced to use a different method.

Toflar:
There we go, the PR is updated to use promises now which allows you to return anything you want to to asynchronously, not just `Process` instances. I've introduced a helper for them, though as I think this will be a very typical use case.

Also, we now do not check the order at runtime anymore and we also do not force asynchronous cron jobs to always get executed before the others. They will just get added to `Cron` before the others in the `AddCronJobsPass` now. Meaning that if you - for some unknown reason - reeeeally need to adjust this, you can with your own compiler pass.

core-bundle/src/Cron/Cron.php
```diff
@@ -135,11 +137,39 @@ private function doRun(array $cronJobs, string $scope, bool $force = false): voi
...
+
+                $promises[$cron->getName()] = $return;
+            }
+        }
+
+        if (0 === \count($promises)) {
+            return;
+        }
+
+        foreach (Utils::settle($promises)->wait() as $cronjobName => $result) {
```
m-vo: This would be executed when all promises are fulfilled, no? Wouldn't we want to log immediately after resolving each promise, i.e. by chaining one to `$return` a few lines up?

m-vo:
Looks promising (pun intended), looking forward to this! :heart:

core-bundle/src/Cron/Cron.php
Toflar: Right, adjusted in https://github.com/contao/contao/pull/5554/commits/bb16ced847221d1a9b1e53f85eba002d9a261ceb.

m-vo (dismissed):

core-bundle/src/Cron/Cron.php
```diff
@@ -135,11 +137,40 @@ private function doRun(array $cronJobs, string $scope, bool $force = false): voi
...
-            $cron($scope);
+            $promise = $cron($scope);
+
+            if ($promise instanceof PromiseInterface) {
+                $promise->then(
+                    function () use ($cron): void {
+                        $this->logger?->debug(sprintf('Asynchronous cron job "%s" finished successfully', $cron->getName()));
+                    },
+                    function ($reason) use ($cron): void {
+                        $this->logger?->debug(sprintf('Asynchronous cron job "%s" failed: %s', $cron->getName(), $reason));
```
ausi:
```suggestion
                        $this->logger?->debug(sprintf('Asynchronous cron job "%s" failed: %s', $cron->getName(), (string) $reason));
```
AFAIK the reason is of type `mixed` an can be an exception object for example.

core-bundle/src/DependencyInjection/Compiler/AddCronJobsPass.php
```diff
@@ -53,14 +59,28 @@ public function process(ContainerBuilder $container): void
...
-                    'addCronJob',
-                    [
-                        new Definition(CronJob::class, [new Reference($serviceId), $interval, $method]),
-                    ]
-                );
+                $newDefinition = new Definition(CronJob::class, [new Reference($serviceId), $interval, $method]);
+
+                $reflector = new \ReflectionMethod($jobDefinition->getClass(), $method ?? '__invoke');
+                $returnType = $reflector->getReturnType();
+                $returnsPromise = $returnType instanceof \ReflectionNamedType && PromiseInterface::class === $returnType->getName();
```
ausi:
```suggestion
                $returnsPromise = $returnType instanceof \ReflectionNamedType && is_a($returnType->getName(), PromiseInterface::class, true);
```
This would allow for covariance i.e. a more specific return type.

core-bundle/src/Util/ProcessUtil.php
```diff
@@ -0,0 +1,46 @@
...
+    public static function createPromise(Process $process, bool $start = true): PromiseInterface
+    {
+        $promise = new Promise(
+            static function () use (&$promise, $process): void {
+                $process->wait();
+
+                if ($process->isSuccessful()) {
+                    $promise->resolve($process->getOutput());
+                } else {
+                    $promise->reject($process->getErrorOutput());
```
ausi:
```suggestion
                    $promise->reject($process->getErrorOutput() ?: $process->getOutput());
```
Should we return the output if the error output is empty?

ausi (dismissed):
Looks awesome! 🚀

_just some minor comments_

core-bundle/src/Cron/Cron.php
Toflar: I actually did add a `(string)` cast but phpstorm tells me it's obsolete due to `sprintf()` in conjuction with `%s` which I guess does that anyway? ;)

ausi (dismissed):

m-vo (dismissed):

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5556, opened 2022-12-07
Fix SitemapController not working for protected pages

Toflar:
`$pageModel->groups` is a serialized string of groups so it never returned `true`. Or in other words, the `sitemap.xml` did not return protected pages even when logged in for 1.5 years without anybody noticing...

The tests were pretty nonsense because they tested only that a page is not added when protected which of course was always green since the check never worked. The inverse (check whether a protected is added if access is given) was not checked. This is fixed now.

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -18,6 +18,7 @@ use Contao\CoreBundle\Event\SitemapEvent;
 use Contao\CoreBundle\Routing\Page\PageRegistry;
 use Contao\CoreBundle\Security\ContaoCorePermissions;
 use Contao\PageModel;
+use Contao\StringUtil;
 use Contao\System;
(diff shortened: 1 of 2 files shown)


aschempp (changes requested):
This can't really be true, because PageModel::loadDetails() does deserialize the data already: https://github.com/contao/contao/blob/5.x/core-bundle/contao/models/PageModel.php#L881

Maybe the controller does not load page details?

Toflar:
Yes, it doesn't that's why it needs a fix.

aschempp:
the fix should be to call `PageModel::loadDetails()` in the loop, your code would not apply inherited access permissions.

Toflar:
This is ready. Failing tests unrelated.

fritzmg (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5558, opened 2022-12-08
Replace "Contao 4" with "Contao" in readmes

ausi:
Or should we replace “Contao 4” just with “Contao” without a major version number?


leofeyer:
> Or should we replace “Contao 4” just with “Contao” without a major version number?

I would prefer this. 👍

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #5562, opened 2022-12-08
Update TL_LANGUAGE deprecation

ausi:
We did not remove it in Contao 5.0 (not sure if on purpose?), but I think we should remove it in Contao 6.


Toflar (approved):

leofeyer:
Thank you @ausi.

PR #5566, opened 2022-12-09
Remove basic entities from SerpPreview

### LINKED ISSUE #5565: Remove Basic Entities from SerpPreview
### Description

Sometimes it makes sense to use [Basic Entities](https://docs.contao.org/manual/de/artikelverwaltung/insert-tags/#basic-entities) in the title field.

Think of a news article to announce the a new "Donaudampfschifffahrtsgesellschaftsschiff". In another case you want to avoid a line break: "Boarding tickets are 10[nbsp]€"

It is very simple to build this feature: In [SerpPreview.php#L44](https://github.com/contao/contao/blob/5.x/core-bundle/contao/widgets/SerpPreview.php#L44) replace
```php
$title = StringUtil::substr($this->getTitle($model), 64);
```
with
```php
$title = StringUtil::restoreBasicEntities(StringUtil::substr($this->getTitle($model), 64));
```

In [mootao.js](https://github.com/contao/contao/blob/5.x/core-bundle/public/mootao.js#L359) just add `.replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' ')` in line 359, 361 & 369.

Browsers just ignore `$nbsp;` oder `&shy;` in the title bar and so does Google – at least this is commonly assumed (https://www.google.com/search?q=html+title+tag+%22%26nbsp%3B%22).

I'd be happy to make a pull request!

a-v-l:
Remove [Basic Entities](https://docs.contao.org/manual/de/artikelverwaltung/insert-tags/#basic-entities) [-] `&shy;` and [nbsp] `&nbsp;` in title or pageTitle fields for the serp preview.

#5565

### DIFF
--- a/core-bundle/contao/widgets/SerpPreview.php
+++ b/core-bundle/contao/widgets/SerpPreview.php
@@ -41,7 +41,7 @@ class SerpPreview extends Widget
 		}
 
 		$id = $model->id;
-		$title = StringUtil::substr($this->getTitle($model), 64);
+		$title = StringUtil::substr(str_replace(array('&nbsp;', '&shy;'), array(' ', ''), $this->getTitle($model)), 64);
 		$description = StringUtil::substr($this->getDescription($model), 160);
 		$alias = $this->getAlias($model);
 
--- a/core-bundle/assets/scripts/mootao.js
+++ b/core-bundle/assets/scripts/mootao.js
@@ -348,9 +348,9 @@ Contao.SerpPreview = new Class(
 
 		titleField && titleField.addEvent('input', function() {
 			if (titleField.value) {
-				serpTitle.set('text', this.shorten(titleTag.replace(/%s/, titleField.value).replace(/%%/g, '%'), 64));
+				serpTitle.set('text', this.shorten(titleTag.replace(/%s/, titleField.value).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
 			} else if (titleFallbackField && titleFallbackField.value) {
-				serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%'), 64));
+				serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
 			} else {
 				serpTitle.set('text', '');
 			}
@@ -358,7 +358,7 @@ Contao.SerpPreview = new Class(
 
 		titleFallbackField && titleFallbackField.addEvent('input', function() {
 			if (titleField && titleField.value) return;
-			serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%'), 64));
+			serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
 		}.bind(this));
 
 		aliasField && aliasField.addEvent('input', function() {


core-bundle/contao/widgets/SerpPreview.php
```diff
@@ -41,7 +41,7 @@ public function generate()
 		}
 
 		$id = $model->id;
-		$title = StringUtil::substr($this->getTitle($model), 64);
+		$title = StringUtil::restoreBasicEntities(StringUtil::substr($this->getTitle($model), 64));
```
leofeyer: Shouldn‘t the basic entities be replaced before the title is shortened? At least the JavaScript behaves this way and I think it would make more sense, too.

a-v-l: Absolutely right!

Also, it doesn't make sense to use `StringUtil::restoreBasicEntities()` in this case: We might end up with strings like `…Dampf&sh` if `substr` splits within an HTML entity.

Fixed in [e3002e0](https://github.com/contao/contao/pull/5566/commits/e3002e0ea90303c2ab24d50c43fd338c43861563)

core-bundle/contao/widgets/SerpPreview.php
```diff
@@ -41,7 +41,7 @@ public function generate()
 		}
 
 		$id = $model->id;
-		$title = StringUtil::substr($this->getTitle($model), 64);
+		$title = StringUtil::substr(str_replace(array('[nbsp]', '[-]'), array(' ', ''), $this->getTitle($model)), 64);
```
ausi: This should not be needed as we do not store basic entities in the database anymore in Contao 5, see [#4453](https://github.com/contao/contao/pull/4453#issuecomment-1094112590)

a-v-l: So, how do you enter `&nbsp;` & `&shy;` then?

ausi: If `['eval' => ['basicEntities' => true]]` is enabled for that field, you can still enter `[nbsp]` and `[-]`. But the value that is stored in the database (and in `$model` in this case) is the actual entity, so `&nbsp;` and `&shy;`.

a-v-l: But this also could lead to strings like `…Dampf&sh`. Therefore I would suggest striping any `&shy;` and `&nbsp;` from `$this->getTitle($model)` before shortening it:

```php
$title = StringUtil::substr(str_replace(array('&nbsp;', '&shy;'), array(' ', ''), $this->getTitle($model)), 64);
```

In mootao.js this would make be `title.replace(/\&shy;/g, '').replace(/\&nbsp;/g, ' ')` instead of `title.replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' ')`

m-vo:
From your original issue:

> Browsers just ignore $nbsp; oder &shy; in the title bar

I don't think this is true. If I hover over a tab in Chrome or Firefox for example, the tooltip windows *do* respect things like `&nbsp;` and arrange their line breaks accordingly.

a-v-l:
> > Browsers just ignore `&nbsp;` or `&shy;`­ in the title bar
> 
> I don't think this is true. If I hover over a tab in Chrome or Firefox for example, the tooltip windows _do_ respect things like `&nbsp;` and arrange their line breaks accordingly.

You are right! But it doesn't do any harm – does it?
In the end the user oder editor should decide on wether to use or not to use `[-]` or `[nbsp]` in the page title. In my case it makes a lot of sense to use those entities. Also in the news section the news title appears as a header in the news list but also as the title meta tag in the news reader.

core-bundle/contao/widgets/SerpPreview.php
a-v-l: > In mootao.js this would make be title.replace(/\&shy;/g, '').replace(/\&nbsp;/g, ' ') instead of title.replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' ')

Wrong: in mootao.js `title.replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' ')` is the right way since the _Basic Entities_ are still in use…

core-bundle/public/mootao.js
```diff
@@ -356,17 +356,17 @@ Contao.SerpPreview = new Class(
 
 		titleField && titleField.addEvent('input', function() {
 			if (titleField.value) {
-				serpTitle.set('text', this.shorten(titleTag.replace(/%s/, titleField.value).replace(/%%/g, '%'), 64));
+				serpTitle.set('text', this.shorten(titleTag.replace(/%s/, titleField.value).replace(/%%/g, '%').replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' '), 64));
```
ausi:
```suggestion
				serpTitle.set('text', this.shorten(titleTag.replace(/%s/, titleField.value).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
```

core-bundle/public/mootao.js
```diff
@@ -356,17 +356,17 @@ Contao.SerpPreview = new Class(
...
 			} else if (titleFallbackField && titleFallbackField.value) {
-				serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%'), 64));
+				serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' '), 64));
```
ausi:
```suggestion
				serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
```

core-bundle/public/mootao.js
```diff
@@ -356,17 +356,17 @@ Contao.SerpPreview = new Class(
...
 			} else {
 				serpTitle.set('text', '');
 			}
 		}.bind(this));
 
 		titleFallbackField && titleFallbackField.addEvent('input', function() {
 			if (titleField && titleField.value) return;
-			serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%'), 64));
+			serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '').replace(/\[nbsp\]/g, ' '), 64));
```
ausi:
```suggestion
			serpTitle.set('text', this.shorten(this.html2string(titleTag.replace(/%s/, titleFallbackField.value)).replace(/%%/g, '%').replace(/\[-\]/g, '\xAD').replace(/\[nbsp\]/g, '\xA0'), 64));
```

core-bundle/contao/widgets/SerpPreview.php
ausi: > But this also could lead to strings like `…Dampf&sh`.

This would happen with any HTML entities, so it is not related to this PR IMO.
To fix this the right way we would have to use a `substr` method that can handle entities I think.

ausi (changes requested):

core-bundle/contao/widgets/SerpPreview.php
a-v-l: Actually [`StringUtil::substr()`](https://github.com/contao/contao/blob/5.x/core-bundle/contao/library/Contao/StringUtil.php#L29) does handle entities:

> The function preserves words, so the result might be a bit shorter or longer than the number of characters given.

So `Dampf&shy;schiff` or `1000&nbsp;€` would be preserved. I am just not sure how Google shortens the title on the result page. Therefore I think we can just replace `&nbsp;` with ` ` [space] and remove `&shy;` (or keep it – makes no difference).

core-bundle/public/mootao.js
a-v-l: Thank you! Makes sense!

core-bundle/contao/widgets/SerpPreview.php
ausi: I think we should just keep all entities here, nothing special about `nbsp` and `shy` IMO.

core-bundle/contao/widgets/SerpPreview.php
```diff
@@ -41,7 +41,7 @@ public function generate()
 		}
 
 		$id = $model->id;
-		$title = StringUtil::substr($this->getTitle($model), 64);
+		$title = StringUtil::substr(str_replace(array('&nbsp;', '&shy;'), array(' ', ''), $this->getTitle($model)), 64);
```
ausi, line 44: Or maybe also just replace it with the unicode chars here too:
```suggestion
		$title = StringUtil::substr(str_replace(array('&nbsp;', '&shy;'), array("\u{A0}", "\u{AD}"), $this->getTitle($model)), 64);
```

a-v-l: I am pretty sure Google removes any given `&nbsp;` or `&shy;`.

Running `for (const txt of document.querySelectorAll('h3')) {console.log(txt)}` in the js-console on any google search result page never returned any `&nbsp;` or `&shy;`. Even searching with a long word didn't return word breaks:
https://www.google.com/search?q=donaudampfschifffahrtselektrizit%C3%A4tenhauptbetriebswerkbauunterbeamtengesellschaft
They also strip these entities from the page description – just compare the SERP source code of above search with the entry in wikipeadia (with a lot of `&shy;`s!

So I would just strip `&shy;` and replace `&nbsp;` with a space character.

ausi: I don’t see any entities in the title there on wikipedia: 
Bildschirmfoto 2022-12-13 um 17 02 35

a-v-l: > I don’t see any entities in the title there on wikipedia

I was referring to the description. This is a screenshot from the search result:
image

And this is the text on the wikipedia page used by google (wikipedia doesn't use the description meta tag):
image
As one can see, google also strips the `<b>` tag…

ausi (dismissed):

leofeyer:
This PR is based on the `5.x` branch but it has been assigned to the `5.0` milestone. Rebase or change milestone?

ausi:
IMO rebase. Maybe even a backport to 4.13 is needed here?

aschempp (commented):
I would suggest to add this to the `html2string` method to automatically clean all such places (there are more uses of `html2string` than replace methods here. Also, the replace could be adjusted to a list of replacements

```
'...'.replace(/\[(-|nbsp)]/g, (v) => ({
    '[nbsp]': '\xA0', 
    '[-]':'\xAD'
})[v] || v)
```

leofeyer:
@aschempp What exactly do you mean? There is no `html2string` method.

@a-v-l Can you please rebase the PR and resolve the conflicts?

ausi (approved):

leofeyer:
Thank you @a-v-l.

aschempp:
> @aschempp What exactly do you mean? There is no `html2string` method.

there is in line 325, and it's used in all the modified lines ;-)

leofeyer:
See #5897

PR #5571, opened 2022-12-13
Ignore unrelated files in the downloads controller

leofeyer:
Fixes #5568

PR #5573, opened 2022-12-13
Do not handle downloads for other elements on the same page

### LINKED ISSUE #5568: Downloads schützen funktioniert nicht
### Affected version(s)

5.0.7

### Description

Ordner angelegt -> nicht öffentlich
Datei in diesen Ordner abgelegt
Datei über downloads/download zur Verfügung gestellt und mit Schutz auf Mitglieder versehen
Im Frontend wird nun folgender Link erzeugt:
https://xxxx.xxxxx.xx/informationen/xxxxinformationen.html?_hash=RxQlTd8yrbRvUXr7vd7sEmh1%2BYs%2FEsUPAdQz0UC4%2Bn4%3D&ctx=a%3A1%3A%7Bs%3A2%3A%22id%22%3Bi%3A2176%3B%7D&d=attachment&f=Ingwersaft.pdf&p=documents+internal%2FIngwersaft.pdf

Wird dieser Link in einem neuen Browser ohne Anmeldung an die Seite aufgerufen, wird die Datei angezeigt.
Sollte so ja nicht sein.

m-vo:
Fixes #5568, alternative approach to #5571.

### DIFF
--- a/core-bundle/src/Controller/ContentElement/DownloadsController.php
+++ b/core-bundle/src/Controller/ContentElement/DownloadsController.php
@@ -234,9 +234,7 @@ class DownloadsController extends AbstractContentElementController
                     return new Response('', Response::HTTP_NO_CONTENT);
                 }
 
-                if (
-                    !$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())
-                ) {
+                if (!$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())) {
                     return new Response('The resource can not be accessed anymore.', Response::HTTP_GONE);
                 }
 
--- a/core-bundle/src/Controller/ContentElement/DownloadsController.php
+++ b/core-bundle/src/Controller/ContentElement/DownloadsController.php
@@ -228,16 +228,13 @@ class DownloadsController extends AbstractContentElementController
             $request,
             $this->filesStorage,
             function (FilesystemItem $item, array $context) use ($model): Response|null {
-                $id = $context['id'] ?? null;
-
                 // Do not handle downloads from other DownloadController
                 // elements on the same page (see #5568)
-                if ($id !== $model->id) {
+                if ($model->id !== ($context['id'] ?? null)) {
                     return new Response('', Response::HTTP_NO_CONTENT);
                 }
 
                 if (
-                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($id)) ||
                     !$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())
                 ) {
                     return new Response('The resource can not be accessed anymore.', Response::HTTP_GONE);
--- a/core-bundle/src/Controller/ContentElement/DownloadsController.php
+++ b/core-bundle/src/Controller/ContentElement/DownloadsController.php
@@ -59,7 +59,7 @@ class DownloadsController extends AbstractContentElementController
         // TODO: Remove method and move logic into its own action, once we have
         // a strategy how to handle permissions for downloads via a real route.
         // See #4862 for more details.
-        $this->handleDownload($request);
+        $this->handleDownload($request, $model);
 
         $filesystemItems = $this->getFilesystemItems($model);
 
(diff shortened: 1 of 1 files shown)


m-vo:
By the way: this issue will go away completely when we migrate to a proper controller for the actual download. This way it won't happen that another element on the same page handles the request first…

aschempp:
Just FYI this was previously already implemented using a `cid` parameters. Not sure we might want to keep that?
https://github.com/contao/contao/blob/5.x/core-bundle/contao/elements/ContentDownload.php#L128

m-vo:
> Just FYI this was previously already implemented using a `cid` parameters. Not sure we might want to keep that?

The `$context` of the current download URL basically contains the `cid` parameter (`id`).

aschempp:
where does `$context` come from?

m-vo:
> where does `$context` come from?

https://github.com/contao/contao/blob/a9e1f8c76106bdd269ee4100bf583376647f86c3/core-bundle/src/Controller/ContentElement/DownloadsController.php#L151-L174

aschempp:
so instead of a `cid` we now have a serialized array of data in the URL?

ausi (dismissed):

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -222,14 +222,22 @@ private function getPreviews(FilesystemItem $filesystemItem, ContentModel $model
...
+                // Do not handle downloads from other DownloadController
+                // elements on the same page (see #5568)
+                if ($id !== $model->id) {
+                    return new Response('', Response::HTTP_NO_CONTENT);
+                }
+
                 if (
-                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($context['id'] ?? null)) ||
+                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($id)) ||
                     !$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())
```
leofeyer: How is this better than directly passing the `$filesystemItems` as done in #5571?

leofeyer: NVM, I have just read your https://github.com/contao/contao/pull/5571#issuecomment-1348183251. 👍

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -222,14 +222,22 @@ private function getPreviews(FilesystemItem $filesystemItem, ContentModel $model
...
+
+                // Do not handle downloads from other DownloadController
+                // elements on the same page (see #5568)
+                if ($id !== $model->id) {
+                    return new Response('', Response::HTTP_NO_CONTENT);
+                }
+
                 if (
-                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($context['id'] ?? null)) ||
+                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($id)) ||
```
leofeyer: We should rename this parameter to `$contentModel` or `$contextModel` or something else that does not interfere with the given `use ($model)` now. WDYT?

m-vo: Agree, that would be cleaner if the variable was not redeclared. In fact as the retrieved model is identical to the now passed argument, the logic can be further simplified. 

Changed in a50d049f9eac6f4d45892306aac888b88d7afe33.

ausi (dismissed):

core-bundle/src/Controller/ContentElement/DownloadsController.php
leofeyer: Nice. 👍

leofeyer (approved):

leofeyer:
Thank you @m-vo.

core-bundle/src/Controller/ContentElement/DownloadsController.php
```diff
@@ -222,16 +222,19 @@ private function getPreviews(FilesystemItem $filesystemItem, ContentModel $model
...
-            function (FilesystemItem $item, array $context): Response|null {
-                if (
-                    null === ($model = $this->getContaoAdapter(ContentModel::class)->findById($context['id'] ?? null)) ||
-                    !$this->getFilesystemItems($model)->any(static fn (FilesystemItem $listItem) => $listItem->getPath() === $item->getPath())
-                ) {
+            function (FilesystemItem $item, array $context) use ($model): Response|null {
+                // Do not handle downloads from other DownloadController
+                // elements on the same page (see #5568)
+                if ($model->id !== ($context['id'] ?? null)) {
+                    return new Response('', Response::HTTP_NO_CONTENT);
```
Toflar, line 234: @m-vo shouldn't that rather end in a 404? Sorry for being late to the party 🙈

ausi: AFAIK the empty response is ignored later in the code and never sent to the client.

leofeyer: Correct. And additionally, 404 would be wrong, because this is not a "file not found" case but a "not responsible for this URL" case.

m-vo: Yes, we are only going to return those responses once switching to a real controller. (If we manage to solve permissions…)

PR #5584, opened 2022-12-15
Add the missing files to the "prepareFormData" hook

Toflar:
It has always been missing but I guess nobody really cared because up to version 5, we did store all uploads in `$_SESSION['FILES']` which we thankfully do not do anymore.

So we need a way to get to the formatted `$arrFiles` which is coming from the upload widgets which are not exactly the same as `$_FILES`.

### DIFF
--- a/core-bundle/contao/forms/Form.php
+++ b/core-bundle/contao/forms/Form.php
@@ -319,7 +319,7 @@ class Form extends Hybrid
 			foreach ($GLOBALS['TL_HOOKS']['prepareFormData'] as $callback)
 			{
 				$this->import($callback[0]);
-				$this->{$callback[0]}->{$callback[1]}($arrSubmitted, $arrLabels, $arrFields, $this);
+				$this->{$callback[0]}->{$callback[1]}($arrSubmitted, $arrLabels, $arrFields, $this, $arrFiles);
 			}
 		}
 


fritzmg (approved):

core-bundle/contao/forms/Form.php
```diff
@@ -319,7 +319,7 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields, $arrFi
 			foreach ($GLOBALS['TL_HOOKS']['prepareFormData'] as $callback)
 			{
 				$this->import($callback[0]);
-				$this->{$callback[0]}->{$callback[1]}($arrSubmitted, $arrLabels, $arrFields, $this);
+				$this->{$callback[0]}->{$callback[1]}($arrSubmitted, $arrLabels, $arrFields, $this, $arrFiles);
```
leofeyer, line 322:
```suggestion
				$this->{$callback[0]}->{$callback[1]}($arrSubmitted, $arrLabels, $arrFields, $arrFiles, $this);
```

I know this is technically a BC break, however, we have changed all callbacks to pass `$this` as last argument in Contao 5 and therefore this could be seen as a bugfix.

fritzmg: > I know this is technically a BC break, however, we have changed all callbacks to pass `$this` as last argument in Contao 5 and therefore this could be seen as a bugfix.

But in all these cases we only _added_ it at the end. I don't think we should break BC here as this provides no gain.

Toflar: I wouldn't do that either TBH. It's probably one of the most frequently used hooks so I'd just add it at the end.

leofeyer:
Thank you @Toflar.

Toflar:
Huh, was that targetet at the wrong branch?🙈

leofeyer:
I‘ll try to cherry pick it.

leofeyer:
See 96def4a6cbb6d566d69cee5a05e9c644974c3060.

PR #5585, opened 2022-12-15
Fix `PageModel::getPreviewUrl` on Windows

### LINKED ISSUE #5582: Backend preview URL does not work
### Affected version(s)

5.0.7, 5.0.x-dev

### Description

If you open the preview from the back end, i.e. via an URL like

```
https://example.com/contao/preview?page=8
```

then the resulting redirect response URL will be

```
https://example.com\/preview.php/page-alias
```

which will lead to a 404 response.

For some reason this call in `PageModel::getPreviewUrl`

https://github.com/contao/contao/blob/0941be9e3562ad09f44665723146627da7a8894c/core-bundle/contao/models/PageModel.php#L1180

results in such an aforementioned URL (with the first slash "escaped").

/cc @aschempp

fritzmg:
Fixes #5582

`\dirname('/preview.php')` and `\dirname('/')` return `\` under Windows (`\dirname('/c50/public/preview.php)` will return `/c50/public` btw., that's why I did not notice back then). This PR fixes that by instead checking whether the preview script is already contained in the base URL and only then adds it to the base URL.

Btw. instead of 

```php
rtrim($baseUrl, '/') . $previewScript
```

we could also use

```php
Path::join($baseUrl, $previewScript)
```

However, I don't know how "bad" it is to use `Symfony\Component\Filesystem\Path::join` for URL path concatenation, rather than file path concatenation?


core-bundle/contao/models/PageModel.php
```diff
@@ -1171,7 +1171,10 @@ public function getPreviewUrl($strParams=null)
 		$baseUrl = $context->getBaseUrl();
 
 		// Add the preview script
-		$context->setBaseUrl(rtrim(\dirname($baseUrl), '/') . $previewScript);
+		if (false === strpos($baseUrl, $previewScript))
+		{
+			$context->setBaseUrl(rtrim($baseUrl, '/') . $previewScript);
+		}
```
ausi:
```suggestion
		$context->setBaseUrl(preg_replace('(/[^/]*$)', '', $baseUrl) . $previewScript);
```
Simpler alternative?

fritzmg: 🦊

ausi (approved):

m-vo (approved):

leofeyer:
> we could also use
>
> ```php
> Path::join($baseUrl, $previewScript)
> ```

What about this?

ausi:
AFAIK `Path` is not meant for URLs? Not sure if it is a good idea to use it here.

leofeyer:
Thank you @fritzmg.

PR #5587, opened 2022-12-16
Also display the visible root trail when searching/filtering

Toflar:
Currently, when filtering or searching for elements in the tree view, you do not get any visible parent trail. It looks like this with the Contao demo data:

Bildschirmfoto 2022-12-16 um 09 20 15

This is especially painful if you have multiple pages that are named the same (from a customer project):

Bildschirmfoto 2022-12-16 um 09 57 45

This fixes this behaviour, two fixes were necessary:

1. The first problem is that `$this->root` is updated again after `initRoots()` is called but it does not update`$this->visibleRootTrails` and `$this->rootChildren`. Basically the order of method calls is like this:

* `__construct()`
* `initRoots()` - sets `$this->root`, `$this->visibleRootTrails` as well as `$this->rootChildren`.
* `panel()` - renders the filters and search bars, and also adds filters to the session.
* `treeView()` - applies the filters by updating `$this->root`. This is the problem: It only updates `$this->root` but lacks updating the `$this->visibleRootTrails` and `$this->rootChildren`. This has to be done always, when you update `$this->root`. That's why I've moved it to a separate `$this->updateRoot()` method.

2. The second problem is that if the `treeView()` is filtered, `$arrFound` must also contain the visible root trails, otherwise they will never be rendered.

This is the result then:

Bildschirmfoto 2022-12-16 um 09 54 10

### DIFF
--- a/core-bundle/src/Resources/contao/drivers/DC_Table.php
+++ b/core-bundle/src/Resources/contao/drivers/DC_Table.php
@@ -3699,7 +3699,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 
 			if ($objRoot->numRows < 1)
 			{
-				$this->root = array();
+				$this->updateRoot(array());
 			}
 			// Respect existing limitations (root IDs)
 			elseif (!empty($this->root))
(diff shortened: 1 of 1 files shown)


ausi (approved):

asaage:
Wasn't this issue already solved in https://github.com/contao/contao/pull/3391 ?
Is there a way to mark the trail-related items greyed out / disabled or something like that?
If you serach for something that is not displayed in the treeview it might be way easier to find it in the searchresult.

Toflar:
This PR enables https://github.com/contao/contao/pull/3391 also when filtered. The rest would be a feature and not part of a bugfix for sure.

leofeyer:
Thank you @Toflar.

PR #5590, opened 2022-12-18
Fix the `tl_page.accesskey` database type

### LINKED ISSUE #5586: Empty accesskeys in navigation
### Affected version(s)

5.0.7

### Description

In a local test installation of Contao 5.0.7 in XAMPP (PHP 8.1.6) I noticed multiple empty `accesskey=""` in the HTML source code of navigation blocks. I did not set any keyboard shortcuts in the page settings in the backend. The accesskey fields in the backend show `0` as default value.

To investigate it, I made a copy of the template `nav_default.html5` and added a variable dump with
```php
echo \Contao\Template::dumpTemplateVars();
```
In the array lists `accesskey => false` is defined for all navigation elements.

The accesskey is generated in this line in the template file:

https://github.com/contao/contao/blob/015787fc6498cf9ecc68215d2d18e04978a3540a/core-bundle/contao/templates/navigation/nav_default.html5#L7

To avoid the generation of the empty accesskeys I modified this snippet:
`<?php if ('' !== $item['accesskey'] ): ?>`
to this:
`<?php if ('' !== $item['accesskey'] && $item['accesskey'] != false ): ?>`

If someone can confirm the generation of empty accesskeys, it could be fixed in the core template.

fritzmg:
Unfortunately in https://github.com/contao/contao/pull/4797 I accidentally also changed  `tl_page.accesskey` to a boolean field, leading to this issue: https://github.com/contao/contao/issues/5586. ~And unfortunately anyone updating to Contao `>=5.0.0,<=5.0.7` will lose their access keys.~ // edit: actually that is not exactly true as you would get an error during the database update.

This PR restores the original field type of `tl_page.accesskey` and adds a migration to reset `tl_page.accesskey` so that it won't erroneously show in the front end at least.


ausi (approved):

ausi (changes requested):
We should probably fix this in tl_form_field too https://github.com/contao/contao/blob/ec62d9636b50448aa046cadb342935fe1ebbac1d/core-bundle/contao/dca/tl_form_field.php#L351-L357

fritzmg:
Yes 🙈

core-bundle/config/migrations.yaml
```diff
@@ -31,6 +31,11 @@ services:
         arguments:
             - '@database_connection'
 
+    contao.migration.version_500.fix_accesskey:
+        class: Contao\CoreBundle\Migration\Version500\AccesskeyMigration
```
ausi:
```suggestion
    contao.migration.version_500.accesskey:
        class: Contao\CoreBundle\Migration\Version500\AccesskeyMigration
```

ausi (dismissed):

ausi (approved):

core-bundle/src/Migration/Version500/AccesskeyMigration.php
```diff
@@ -0,0 +1,70 @@
...
+
+        return false;
+    }
+
+    public function run(): MigrationResult
+    {
+        foreach (self::$affectedTables as $table) {
+            if ($this->needsMigration($table)) {
+                $this->connection->executeQuery("ALTER TABLE $table CHANGE accesskey accesskey CHAR(1) DEFAULT '' NOT NULL");
+                $this->connection->executeQuery("UPDATE $table SET accesskey = '' WHERE accesskey = '0'");
```
leofeyer, line 47: This would reset my access key `0` that I am using in my navigation menu.

ausi: In Contao 5.0, until this fix, all access keys are broken anyways.

If updating from an older database this migration would not be executed, because it only runs if the column type is boolean.

leofeyer: I would prefer not to add the migration. Only 1% of the users are already using Contao 5 and potentially have the issue, but the migration will break the access key `0` for the remaining 99% once they upgrade to Contao 5.

ausi: > but the migration will break the access key `0` for the remaining 99% once they upgrade to Contao 5.

No, this migration will never be executed for them.

> Only 1% of the users are already using Contao 5 and potentially have the issue

All Contao 5 users have an accesskey of `0` for every single page, they would all have to fix this manually anyways.

leofeyer:
Thank you @fritzmg.

PR #5591, opened 2022-12-19
Remove localconfig.disableCron in favor of a new zero config approach

Toflar:
While trying to deprecate `disableCron` and move it to the bundle config I found that actually I do not want to configure that on every setup. If I have configured a minutely, real `contao:cron` CLI job, we never need the front end scheduler.

So instead of making it a bundle config and documenting that you need to do both, configure a real cronjob and disable the scheduler listener, I went for a new zero config approach.
Very simple: We just register a minutely cron job that updates a cache file every minute.

Writing a file every minute should not be a problem at all. Servers write thousands of log entries to files in that same minute ;-)

Also, this approach prevents from errors: If your CLI cron is misconfigured (PHP version changed, you forgot to update the CLI path), it would automatically fall back to the web crons, ensuring GDPR compliance with cleaned up tokens etc. which currently wouldn't happen.

The behavior of the the listener can be configured as follows:

```yaml
contao:
     cron:
        web_listener: 'auto' # Default, enables the new zero config approach. "false" and "true" explicitly enable or disable the listener
```

PS: If you wonder why we do not just re-use the DB entry: I thought about that but I think there are use cases where one would call `$cron->hasMinutelyCliCron()` from places where there was no DB connection initiated yet and it would be the only one just to answer that question. So I went for the possible zero connection approach there too.


core-bundle/src/Cron/Cron.php
```diff
@@ -175,4 +203,17 @@ function ($reason) use ($cron): void {
...
         Utils::settle($promises)->wait();
     }
+
+    /**
+     * Checks if a database connection can be established and the table exist.
+     */
+    private function canRunDbQuery(Connection $connection): bool
+    {
+        try {
+            return $connection->isConnected()
```
Toflar: @fritzmg why exactly do we check for `isConnected()`? This would mean that if there wasn't any previous query, crons would never be run? 🤔

ausi:
Would that mean, that every time the cache gets cleared the frontend cron is enabled for about one minute?

Toflar:
> Would that mean, that every time the cache gets cleared the frontend cron is enabled for about one minute?

In the current implementation, yes.

Toflar:
On the other hand it would also prevent errors: If your CLI cron is misconfigured (PHP version changed, you forgot to update the CLI path), it would automatically fall back to the web crons, ensuring GDPR compliance with cleaned up tokens etc. which currently wouldn't happen.

core-bundle/src/Cron/Cron.php
fritzmg: Why did you move this to the Cron service? This is only necessary for the `CommandSchedulerListener` as it is a `kernel.terminate` event and it can be executed even when no DB credentials are set up yet (iirc).

/cc @ausi

Toflar: It's always needed when `$this->cron->run()` is executed. No matter where it's executed from. The responsibility should be in the cron service.

fritzmg: But the purpose was to prevent an exception from happening in regular requests and thus it is only relevant in the `kernel.terminate` listener. If you run the web or CLI cron an exception is fine.

Toflar: I've pushed the previous state again, as you can see, tests are somehow failing. The cron scheduler seems to be executed still.

fritzmg: Does it work locally? i.e., if you remove the database credentials (or make them purposefully wrong) and you access the front end - will you see a separate exception from the `kernel.terminate` event regarding the database?

Toflar: I found the problem: The issue is that our functional tests do not in fact run on a 100% complete setup. They only insert `tl_page` entries etc. So the cron jobs are now triggered and thus fail at e.g. the `PurgePreviewLinksCron`. We previously disabled running crons during the tests which is now not possible anymore.
Seems like we still need the setting for our own tests? Any other ideas?

fritzmg: May be we should still introduce a bundle config, with which you can forcibly disable the `CommandSchedulerListener`? Then we could disable it for our tests and in the real world you can also disable it, if you want to (but you don't have to).

Toflar: Alternative proposal: We move the setting to `contao.cron.web_listener` that allows to take on the current values `true` and `false`. We can force disabling it in our tests as we do now then and someone can also force enable it, if they want to have the `web` scope for some reason. New would be an `auto` configuration which implements the logic of this PR and becomes the new default value.
That way, we can support it all.

fritzmg: Sounds good to me :)

ausi: +1 for a true/false/auto setting ❤️

core-bundle/src/Cron/Cron.php
```diff
@@ -34,8 +36,27 @@ class Cron
...
+    }
+
+    public function updateMinutelyCliCron(string $scope): void
+    {
+        if (self::SCOPE_CLI !== $scope) {
+            return;
+        }
+
+        $cacheItem = $this->cachePool->getItem(self::MINUTELY_CACHE_KEY);
+        $cacheItem->expiresAfter(70); // 70 instead of 60 seconds to give some time for stale caches
```
ausi, line 60: I’d even increase this to something like 300 (or 600), so that long running crons or other issues don’t accidentally enable the frontend cron.

fritzmg: This would mean though that a cron like this:

```
*/5 * * * * vendor/bin/contao-console cron:run
```
or
```
*/10 * * * * vendor/bin/contao-console cron:run
```
respectively is still "allowed".

ausi: Yes, in both cases no frontend cron should be needed, right?

fritzmg: Well iirc @Toflar wanted to make a _minutely_ cron execution mandatory in order to ensure that the message queue gets processed properly.

Toflar: Not just that. It just makes no sense to configure anything different than `* * * * *` on CLI. Our cron framework delegates so one needs to have the lowest configuration possible in order to have the other ones get their desired execution interval.

Toflar:
@ausi @fritzmg new concept is implemented. I like it a lot 😊

core-bundle/src/EventListener/CommandSchedulerListener.php
```diff
@@ -52,9 +50,16 @@ private function canRunCron(Request $request): bool
             return false;
         }
 
-        $config = $this->framework->getAdapter(Config::class);
+        // Without the db table, the cron framework cannot work
+        if (!$this->canRunDbQuery()) {
+            return false;
+        }
+
+        if ($this->autoMode && !$this->cron->hasMinutelyCliCron()) {
```
ausi:
```suggestion
        if ($this->autoMode && $this->cron->hasMinutelyCliCron()) {
```
If we are in `auto` mode and we _have_ a minutely cli cron we want to skip the frontend cron, right?

Toflar: Lol, wrote the tests incorrectly. Of course. Fixed in https://github.com/contao/contao/pull/5591/commits/e2999b5ef709db746ddbddebd8b259b05940bf12.

core-bundle/src/Cron/Cron.php
Toflar: BTW: Long running crons wouldn't be an issue, they would run in the process of the previous minute so I think we should keep it as is. Imho this is no blocker, we can still increase or lower the time later on if there are valid reasons to do so.

fritzmg: Yeah but within the process the minutely cronjob will run last and it is currently synchronous, which means it could be executed much later. May be also return a promise for that minutely cron?

Toflar: Can do that, if desired. No problem :)

Toflar: Implemented in https://github.com/contao/contao/pull/5591/commits/145edfa4504e7dd0651854fd76573413ce518dac.

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

core-bundle/src/Cron/Cron.php
```diff
@@ -34,8 +37,38 @@ class Cron
...
+    public function hasMinutelyCliCron(): bool
+    {
+        return $this->cachePool->getItem(self::MINUTELY_CACHE_KEY)->isHit();
+    }
+
+    public function updateMinutelyCliCron(string $scope): PromiseInterface|null
     {
+        if (self::SCOPE_CLI !== $scope) {
+            return null;
+        }
```
ausi, line 57: This does not (yet) work as expected, I created an issue for that: https://github.com/contao/contao/issues/5645

Issue #5401, opened 2022-10-18
Favorites and third menu level


leofeyer:
As discussed on the developers meeting, we want to improve the back end navigation.

### Favorites

The user shall be able to add any back end URL to a favorites list:

<img width="376" alt="" src="https://user-images.githubusercontent.com/1192057/196389116-404a2bbf-253d-4305-b761-f3c72827c87f.png">

The favorites shall then be displayed on top of the existing back end navigation:

<img width="210" alt="" src="https://user-images.githubusercontent.com/1192057/196389737-79ac299d-62e0-4135-8c98-8253a1e23a51.png">

The favorites shall be organized in a tree structure, so the favorites menu can be nested, too.

### Third menu level

Where applicable, the third menu level shall show the child elements:

<img width="210" alt="" src="https://user-images.githubusercontent.com/1192057/196384986-f8be4aee-c5bb-4303-9067-5e1e8e63378e.png">

Clicking on "Contao Official Demo" will open the child elements view:

<img width="1230" alt="" src="https://user-images.githubusercontent.com/1192057/196385857-683e18de-5fd6-4727-b286-9a9d492bb21a.png">

Maybe we can add a special handling for themes, so the fourth level items (modules, layouts and image sizes) are also directly accessible in the navigation.

asaage:
I like the idea of Favorites.
But a third menu-level in the sidebar not so much.
That would lead to inconsistencies (why can't you navigate the whole site-/ or file-tree from the Sidebar then?)
I figure that would also increase scrolling. You never know how much calendars, newsarchives etc. one might have and the name of them would (even without any indention) most likely not fit in one line of sidebar-space.

contaoacademy:
Perhaps it would be helpful to add custom links to the favorites. Example: A Link to Matomo-Login-URL.

contaoacademy:
It would also be interesting if the data could be exported. So that I can transfer my favorites to another installation.

doishub:
Are the favorites bound to a user or are they defined globally in the project?
If they are bound to a user (which I would prefer), it should also be possible to assign favorites to users as an administrator.

The submenus in the backend could quickly become confusing, which is why I am not completely convinced of the idea.

fritzmg:
> Are the favorites bound to a user or are they defined globally in the project?

Per user.

asaage:
> transfer my favorites to another installation

[...]

PR #5592, opened 2022-12-19
Add a favorites menu in the back end

leofeyer:
Implements #5401

### TODO

- [x] Unit tests
- [x] Check permissions


core-bundle/src/EventListener/DataContainer/BackendFavoritesListener.php
```diff
@@ -0,0 +1,88 @@
...
+            case 'toggle':
+            case 'delete':
+            default:
+                $createdBy = (int) $this->connection->fetchOne(
+                    'SELECT user FROM tl_favorites WHERE id = :id',
+                    ['id' => $dc->id]
+                );
+
+                if ($createdBy !== $userId) {
+                    throw new AccessDeniedException(sprintf('Favorite ID %s does not belong to user ID %s', $dc->id, $userId));
```
Toflar: This whole permission check logic seems wrong. We've worked so hard to finally have security voters! 😊 Why is this an onload_callback again now?

core-bundle/src/EventListener/Menu/BackendFavoritesListener.php
```diff
@@ -0,0 +1,160 @@
...
+            return;
+        }
+
+        $this->buildFavoritesMenu($event, $user);
+    }
+
+    private function buildFavoritesMenu(MenuEvent $event, BackendUser $user): void
+    {
+        if (!$request = $this->requestStack->getCurrentRequest()) {
+            throw new \RuntimeException('The request stack did not contain a request');
```
Toflar:
```suggestion
            return;
```

Why throwing an exception? You just cannot contribute to the menu but 🤷‍♂️

Toflar (changes requested):
Wohey, apart from the code reviews, I've checked out the branch and tested it. I've noticed that when I am on a page that is already a favorite, I can still add it as a favorite again - that doesn't make sense, don't you think?

Looks pretty good already, though! 🥳

core-bundle/src/EventListener/Menu/BackendFavoritesListener.php
leofeyer: Changed in c4ea87858276c033101994efaf2c0bca9999ed97.

leofeyer:
> I've noticed that when I am on a page that is already a favorite, I can still add it as a favorite again - that doesn't make sense, don't you think?

Changed in b3cc8a9a192d720708e7d87d76ca707546ca6815.

core-bundle/src/EventListener/DataContainer/BackendFavoritesListener.php
leofeyer: Because we haven‘t converted any onload_callback permission check to a security voter yet and I didn‘t want to do it without having a concept that we all agree on. Otherwise there will be follow-up discussions and the PR will not be merged in time.

Toflar: We have. We just haven't transformed any other than the news which we should. Please, let's not add even more stuff we ned to work on. This here is going to be more or less a copy paste of https://github.com/contao/contao/blob/5.x/news-bundle/src/Security/Voter/NewsArchiveAccessVoter.php.

leofeyer: Changed in ae6459ca9d4c0616fb4b7cb373a1e1327ebe42df.

core-bundle/config/services.yaml
```diff
@@ -744,6 +744,12 @@ services:
         tags:
             - { name: kernel.reset, method: reset }
 
+    contao.security.backend_favorites_voter:
+        class: Contao\CoreBundle\Security\Voter\DataContainer\BackendFavoritesVoter
```
leofeyer, line 748: @Toflar Not sure if `Voter\DataContainer` is correct? It could also be just `Voter`.

leofeyer:
This one is ready for review again. 😉

Toflar (dismissed):
Look how beautiful this code has turned out with the new permission voter ❤️ 

I think I can approve this. I have just two UX remarks which I noticed using the feature that are debatable and you might want to look into or also not:

1. If somehow possible it would be great if the favorite had a default title when I add a new one.
2. Not sure if it's good to remove the icon from the navigation when already added. It makes the navigation change all the time when I navigate. Maybe disabling it with a different icon indicating that you cannot click would be better.

Again not sure for both. Both can also be changed or added later on. I just wanted to leave the remarks here.

leofeyer:
I have implemented no. 2 in 7a2f03d6f2e2711e2df3ad1a5e4ace4bd9fc2f2b.



Now if a URL is a favorite already, the symbol changes and clicking it takes you to the favorites module:



No. 1 is unfortunately much harder to implement. The best predefined value would be the page title, however, since we do not have a request context in the back end, I cannot retrieve it in the listener. We may have to open a separate issue and discuss this in the next call.

Toflar (approved):
The navigation feels way better (to me) now! Thanks!

Regarding the tilte: Sure, that's something we can still improve.

contaoacademy:
Awesome, thanks for the implementation

PR #5593, opened 2022-12-19
Fix the `aria-hidden` attribute in the tips.js file

leofeyer:
The attribute cannot be a boolean value.


Toflar (dismissed):

aschempp (changes requested):
According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

> `aria-hidden="true"` should not be added when:
>
> - The HTML [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute is present
> - The element or the element's ancestor is hidden with [`display: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
> - The element or the element's ancestor is hidden with [`visibility: hidden`](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)

According to our code, the elements `display` state is always in sync with the aria attribute, so we can as well remove it?

leofeyer:
Yes, I think so, too. Removed in e86a22ac278161fe44f212abc2e2c653b4e8c31e.

aschempp (approved):

PR #5597, opened 2022-12-20
Correctly open and close tree nodes if there is a filter

leofeyer:
How to reproduce:

1. Log into the back end of the official demo and go to pages.

2. Search for all pages with an `o` in their name.

3. The tree will look like this with **all nodes open**:

4. Try closing a node or toggling all nodes. It does not work.

5. Checkout this PR and try again. Now it works. 😄 

Hint: The diff is best viewed with whitespaces ignored.

### DIFF
--- a/core-bundle/contao/drivers/DC_Table.php
+++ b/core-bundle/contao/drivers/DC_Table.php
@@ -3676,7 +3676,7 @@ class DC_Table extends DataContainer implements ListableDataContainerInterface, 
 			$node = ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] ?? null) == self::MODE_TREE_EXTENDED ? $this->strTable . '_' . $table . '_tree' : $this->strTable . '_tree';
 
 			// Expand tree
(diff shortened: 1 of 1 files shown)


ausi:
If you change the filter now, do all nodes get opened again? 
_(IMO it should)_

leofeyer:
I see what you mean. So opening all nodes is intended when searching but it is unwanted in the new favorites menu (see #5592). I guess I have to come up with another solution then.

leofeyer:
No, actually I think we should merge the PR.

Always expanding all nodes if someone searches for specific pages might return a large number of results and we surely don‘t want to expand all tree nodes in this case.

This PR does not change the node status at all, which means if a node was open before searching, it will remain open (if there are results) and vice versa. See the following examples:

https://user-images.githubusercontent.com/1192057/208906853-4d0be039-1690-4798-8d3b-90e6981cd457.mov

https://user-images.githubusercontent.com/1192057/208906964-164031d3-df96-4e22-8833-939e4b346e91.mov

PR #5598, opened 2022-12-20
Also minify the core.js and mootao.js scripts with Webpack

leofeyer:
Follow-up on #5406


core-bundle/contao/templates/backend/be_main.html5
```diff
@@ -33,8 +33,6 @@
       $objCombiner->add('assets/chosen/js/chosen.min.js');
       $objCombiner->add('assets/simplemodal/js/simplemodal.min.js');
```
Toflar, line 34: What about the other ones? Can't we add them via `asset()` too?

Toflar (commented):
LGTM, just wondering why we still have the Combiner in use. Not gonna approve here as imho @aschempp had a plan regarding migration to Encore. He should approve then :)

core-bundle/contao/templates/backend/be_main.html5
leofeyer: Yes, but then we would end up with a lot of single requests instead of just one for the combined file.

Toflar: So? It's even better with single requests. Why would I download a file that contains the scripts for chosen, simplemodal, datepicker etc. combined? Chances that only one of them changes during a Contao update are pretty high. The way it is now it would generate a new combined file forcing me to download it all again.
Yes, I agree it's still better to combine the stuff that belongs together in one file, even with http2/3. But these files are completely independent from each other. Chosen is not related to simplemodal whatsoever. Imho, we should not combine here.

leofeyer: Agreed. But let‘s change this in a separate pull request, because it is not really related to this one.

Toflar: Okay 👍

m-vo:
LGTM :+1:, though I didn't have time to manually test, yet.

aschempp (commented):
are you sure this still makes the mootao stuff globally available and not just within the Webpack runtime?

leofeyer:
Yes, I have tested it. Using `window.Backend = {` instead of `var Backend = {` does the trick.

core-bundle/contao/templates/backend/be_main.html5
leofeyer: See #5626.

PR #5606, opened 2022-12-23
Use the request language to match `iflng` tags

aschempp:
If insert tags are replaced in the back end, the language cannot be matched on the page model. In my case this happens in the Isotope order details template, but it could also happen for our new Twig templates if I'm right.

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -1527,14 +1527,14 @@ class InsertTags extends Controller
 	 */
 	private function languageMatches($language)
 	{
-		$request = $container->get('request_stack')->getCurrentRequest();
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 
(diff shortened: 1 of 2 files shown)


Toflar (dismissed):

Toflar (approved):

ausi (commented):
Does `$request->getLocale()` always match the `$objPage->language`? 
Or should we rather use the request as a fallback if `objPage` is not available?

aschempp:
> Does `$request->getLocale()` always match the `$objPage->language`?

it does because we set it in https://github.com/contao/contao/blob/4.13/core-bundle/src/Resources/contao/pages/PageRegular.php#L88

ausi (dismissed):

Toflar (dismissed):

leofeyer (changes requested):
Please also adjust the tests.

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @aschempp.

Issue #391, opened 2019-03-08
Nested content elements


bezin:
Hello there,

I struggle with the possibilities to structure content in Contao once in a while. Pages, articles an content elements are already great and give us many possibilities to structure our content. However, many times I am missing another level of detail, namely grouping or nesting content elements. 

This is partly implemented with the wrapper content elements of course, but they feel more like a workaround rather than a decent solution (for some cases). So I am looking for a more versatile and editor-friendly way to have nested content elements. I fiddled with the online demo and dev tools to illustrate my ideas: 

<img width="600" alt="sketch_dc_parent_view" src="https://user-images.githubusercontent.com/4400435/54029671-7afa1900-41a9-11e9-8ad2-12942c119e2c.png">

<img width="599" alt="sketch_dc_edit_view" src="https://user-images.githubusercontent.com/4400435/54029670-7afa1900-41a9-11e9-908b-92c7f2c16b8a.png">

Editing nested content elements (2nd screenshot) could then open in a modal window (issues ahead, see below).

[As I read in a fairly old issue](https://github.com/contao/core/issues/1247#issuecomment-374533143) the core team seems to have thought about this as well. I would like to hear what your ideas are?

I guess, wrapper elements were kind of a compromise to achieve element grouping within the constraints of the given Contao framework back when they where implemented. I also know that my sketched idea fails right away, when you would like to nest nested content elements, which is a Must requirement for a future solution in my opinion. 

 I am really looking forward to comments in any direction :-) Please treat the screenshots as what they were intended to be: As basic sketches that are totally not thought through ;-)

Cheerio

planepix:
Hi bezin,
have a look at this bundle ( https://gitlab.com/tricksiebzehn/trk17columns ) for have a better visualisation of grids in backend.
![backend-view-bootstrap](https://user-images.githubusercontent.com/1823889/54051084-37230600-41e1-11e9-93b1-e4e71658a18d.png)

![backend-view-bootstrap2](https://user-images.githubusercontent.com/1823889/54051145-646fb400-41e1-11e9-9e80-a86d381eda90.png)
You can switch in backend to see the result for different widths

asaage:
I think Wrapper-elements like *Accordion (Wrapper start/stop)* & *Content slider (Wrapper start/stop)*
are a pretty fine solution - i'm just **missing a generic one in the core** to just wrap elements and attach a class/id.
An Improvement would definitely be:

- auto-creation of the stop-Element
[...]

PR #5609, opened 2022-12-23
Improve the content elements view

leofeyer:
I‘ve been working on an improved content elements view that has many advantages:

- More accurate element preview
- Option to align the items to a grid in the future (see images below)
- Pseudo-nested elements are slightly easier to comprehend
- Will work nicely with nested content elements


leofeyer:
Forgot to mention that the new view can easily be adapted for other resources, too:

leofeyer:
This is how nested content elements (see #391) could look like:

fkaminski:
Maybe you could fill the corresponding nested area with a background color? 
Then one would not have to indent the nested elements so strongly.

image

fritzmg (dismissed):

leofeyer:
> Maybe you could fill the corresponding nested area with a background color?

Changed in d773551aad752fe7f9074ace1157016f614d4acd.

core-bundle/contao/classes/Backend.php
```diff
@@ -544,6 +544,11 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 				}
 			}
 
+			if (($GLOBALS['TL_DCA'][$strTable]['list']['sorting']['useNewDesign'] ?? false) && \in_array(Input::get('act'), array('paste', 'select', null), true))
```
Toflar: Do you think that's necessary? How many "new designs" will we get? I mean, why is it configurable? We don't have any BC on how the back end looks like, templates, styles etc.

Toflar:
Looks nice! I'm playing around with it a little. One thing I immediately noticed: The contrast on hover for the parent view is not really good now:

Bildschirmfoto 2022-12-28 um 18 09 50

core-bundle/contao/classes/Backend.php
leofeyer: Of course it is necessary, otherwise it wouldn‘t be there. 😄 

However, it is not there for user configuration. The new design does not make sense for all parent views, so I needed a way to specify where it shall be applied. We can change the name of the config option if you prefer.

leofeyer: This is how the modules view would look like with the new design:



That‘s why we need the config option.

Toflar: Ah I see. That was actually the question ;) I'd prefer a different name, yes. Because what's after "new design"? "new design 2", "new design 3"? 😊 I don't know what name comes to your mind? "Boxed layout"? Anyway, would also be good to read in the CHANGELOG as such: "Introduced new xy layout". Wdyt?

leofeyer: I have renamed the option to `renderAsGrid` in e609c6d20d4624c733653e6fc1d5affa9716414f because that is what we actually need to distinguish (render as table vs render as grid). To remain backwards compatible, "render as table" is the default.

Toflar: Very good, I like that :)

Toflar:
I still think my [comment](https://github.com/contao/contao/pull/5609#issuecomment-1366801854) should be addressed, other than that I quite like the grid view :)

leofeyer:
I disagree with your comment. The image is showing the hover background color and the contrast ratio is 15.35, which is just fine according to https://contrast-ratio.com/#%23222-on-%23fffce1.

Toflar:
I mean the contrast vs. the background which is white: https://contrast-ratio.com/#%23fff-on-%23fffce1

leofeyer:
Contrast ratio between two background colors is not a thing AFAIK. 😄 

The background color and the hover background color in the particular view are fine. I don‘t want to change them.

PR #5610, opened 2022-12-26
Add an "edit image size" popup link

### LINKED ISSUE #5390: Feature: Edit image-size in content element via lightbox
It would be nice if we can edit the predefined image-sizes via lightbox from a content element. So we dont have to switch for a quick change.
artikel iso 9001 inhaltselement id 2092 bearbeiten contao open source cms 3 5

Maybe a small icon, after the select fields.

aschempp:
Fixes https://github.com/contao/contao/issues/5390


core-bundle/contao/widgets/ImageSize.php
```diff
@@ -233,11 +238,44 @@ public function generate()
...
+
+		$config = array(
+			'ids' => $ids,
+			'href' => System::getContainer()->get('router')->generate('contao_backend', array('do'=>'themes', 'table'=>'tl_image_size_item', 'popup'=>'1', 'nb'=>'1')),
+			'title' => $GLOBALS['TL_LANG']['MSC']['editImageSize'],
+			'icon' => Image::getUrl('edit.svg'),
+			'iconDisabled' => Image::getUrl('edit_.svg'),
+		);
+
+		return ' data-controller="contao--image-size" data-contao--image-size-config-value="' . StringUtil::specialchars(json_encode($config)) . '"';
```
Toflar, line 281: Wouldn't it be better to have a general attribute logic? No matter whether it's Stimulus or not?🤔

aschempp: Whats an "attribute logic"?

Toflar: Giving the widget the ability to get arbitrary data attributes sounds like a concept any widget could use and not limited to Stimulus :)

m-vo: This is a good use case for the `HtmlAttributes` class - it also already handles encoding for you. I would like to add some stimulus shortcuts to it in a later PR and then we can use it here as well (will bring code completion instead of having to memorize the rather complex naming).

```php
return (new HtmlAttributes())
  ->set('data-controller', 'contao--image-size')
  ->set('data-contao--image-size-config-value', json_encode($config))
  ->toString()
;

fritzmg (dismissed):

core-bundle/contao/widgets/ImageSize.php
aschempp: @Toflar no idea what you're referring to, this is a private method and each widget is generating HTML itself so a "generic" attribute feature does not make much sense to me. Also, any attribute of the widget is already added, but these attributes are coming from the widget rendering not the widget configuration.

leofeyer (changes requested):
Great work. Just one little issue: the button title is not converted to a tooltip.





That is probably because the `tips.js` logic runs before the `image-size-controller.js`.

aschempp:
also because tips.js is not watching the DOM using a MutationObserver. I don't think we should fix that here though, and it isn't really problem, is it?

leofeyer:
Why would we not fix this?

aschempp:
> Why would we not fix this?

we should fix this, but it's not related to this PR, since the tips.js is a separate behaviour.

leofeyer:
This still does not work, I‘m afraid.

https://github.com/contao/contao/assets/1192057/904e54e8-4439-443b-847b-3db2ea0ecad4

The button does not have a tooltip and it does not function, either.

aschempp:
both fixed now 😊

leofeyer (approved):

leofeyer:
Thank you @aschempp.

PR #5612, opened 2022-12-27
Improve "copy to clipboard" stimulus controller

### LINKED ISSUE #4147: Improve the "copy to clipboard" feature

m-vo:
Closes https://github.com/contao/contao/issues/4147

This PR improves the "copy to clipboard" functionality by adding visual feedback. I installed a new postcss toolchain top compile the CSS. In a later PR we can also move the core's other CSS into there. 

[clipboard.webm](https://user-images.githubusercontent.com/5305677/209670884-cd95fad3-8fd3-44bb-b1e6-90da94b5ef54.webm)

The button now also uses a HTML button element and inline svgs via data-urls (so that they get loaded without a second request). IMHO correct, but maybe too much difference to the other things?


core-bundle/contao/dca/tl_preview_link.php
```diff
@@ -59,10 +59,7 @@
 			'delete',
 			'toggle',
 			'show',
-			'share' => array
-			(
-				'icon'                => 'share.svg'
-			)
+			'share' => array(),
```
leofeyer: We should keep the existing icon and add a second variant with a checkmark instead of using two completely new icons. I will take care of creating the variant.

core-bundle/assets/controllers/clipboard-controller.js
```diff
@@ -1,25 +1,43 @@
...
+    write () {
+        navigator.clipboard
+            .writeText(this.contentValue)
+            .then(() => {
+                // (Re-) add "written" class to give a visual feedback
+                this.reset();
+
+                requestAnimationFrame(() => {
+                    this.element.classList.add(this.writtenClass);
+                });
```
aschempp, line 29: I don't think we should `requestAnimationFrame` to set a css class

core-bundle/assets/controllers/clipboard-controller.js
```diff
@@ -1,25 +1,43 @@
...
+
+                // Notify others
+                ClipboardController.instances
+                    .filter(i => i !== this)
+                    .forEach(i => i.reset())
+                ;
```
aschempp, line 35: can you explain why it is necessary to notify other controllers? That feels contrary to how Stimulus works using cross-controller events. Also, shouldn't we unset the CSS class automatically after a timeout and not just when another button is clicked?

core-bundle/assets/controllers/clipboard-controller.js
```diff
@@ -1,25 +1,43 @@
...
     }
 
-    clipboardFallback  () {
-        const input = document.createElement('input');
-        input.value = this.contentValue;
-        document.body.appendChild(input);
-        input.select();
-        input.setSelectionRange(0, 99999);
-        document.execCommand('copy');
-        document.body.removeChild(input);
```
aschempp, line 23: @ausi intentionally added clipboard fallback, why was this removed?

core-bundle/assets/styles/clipboard.pcss
```diff
@@ -0,0 +1,69 @@
...
+    &--expired {
+        opacity: .5;
+    }
+
+    &:before {
+        flex-shrink: 0;
+        content: '';
+        width: 16px;
+        height: 16px;
+        background: url("data:image/svg+xml,%3Csvg fill='none' stroke='%23666' viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath stroke-linecap='round' stroke-linejoin='round' stroke-width='2' d='M9 5H7a2 2 0 00-2 2v12a2 2 0 002 2h10a2 2 0 002-2V7a2 2 0 00-2-2h-2M9 5a2 2 0 002 2h2a2 2 0 002-2M9 5a2 2 0 012-2h2a2 2 0 012 2'%3E%3C/path%3E%3C/svg%3E") no-repeat center center;
```
aschempp: I would prefer to have icon files since they are easier to reuse if we need a clipboard icon somewhere else. Also would be consistent with all the other CSS (apart from this being a separate file, not sure we want to go this route).

aschempp (changes requested):

core-bundle/assets/controllers/clipboard-controller.js
ausi: AFAIK this is to trigger the CSS `animation` if the element already has the written class.

core-bundle/contao/dca/tl_preview_link.php
leofeyer: I have added the icon in a5eaf733d5854dcb250a6301e955cfa0e5936a30.

leofeyer (changes requested):
Please use the existing `share.svg` and `shared.svg` icons and add the CSS to the existing theme style sheets instead of using Webpack. We will develop a concept for back end assets on the next developers meeting.

core-bundle/assets/controllers/clipboard-controller.js
m-vo: > AFAIK this is to trigger the CSS animation if the element already has the written class.

Exactly.

core-bundle/assets/controllers/clipboard-controller.js
m-vo: It's needed so the copied state/icon gets removed from the other controllers.

core-bundle/assets/controllers/clipboard-controller.js
m-vo: Judging from the browserslist we do not support such old browsers in Contao 5, do we?

core-bundle/assets/styles/clipboard.pcss
m-vo: Using files has some drawbacks: If the file wasn't requested before (e.g. regular page load), it would only be loaded once the button was clicked resulting in a flaky animation. How would you solve this?

fritzmg: I would solve it with

```html
<link rel="preload" href="foobar.svg" as="image" type="image/svg+xml">
```

zoglo:
This is or is not ready to be merged. It's Moritz' work :)

aschempp:
I would prefer a notification similar to what the Contao Manager (and afaik the Template Studio) uses for the confirmation. The "glowing icon" is very subtle and might not be noticed by visually impaired users. It also would not allow a screen reader to tell any confirmation.

m-vo:
> I would prefer a notification similar to what the Contao Manager (and afaik the Template Studio) uses for the confirmation.

I agree. With #8064 we can easily do this.

m-vo:
See #8204

PR #5615, opened 2022-12-30
Fix the foreignKey handling when showColumns is enabled

Toflar:
Everywhere else, `foreignKey` is always translated but in the list view when `showColumns` is enabled, it just outputs the ID which I consider a bug as it's very counter-intuitive to the rest of the DCA behavior.


aschempp (dismissed):
❤️ so much! Just added a listener yesterday to manually fix the column view for a client system 😂

ausi (dismissed):

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5617, opened 2022-12-30
Introduce OAuth login functionality

fritzmg:
## Description

This introduces OAuth login functionality for Contao, as a separate bundle.

This bundle adds a new _OAuth clients_ section to the back end

Screen Shot 2022-12-30 at 21 10 12

where you can create/edit your OAuth clients.

Screen Shot 2022-12-30 at 21 13 17

For all (supported) OAuth providers you can currently define the client ID and client secret as well as the authorization scope. For Facebook you also need to define the Graph API version.

When creating your OAuth App at the respective provider you need to define the following as the valid callback URL:

```
https://example.com/_oauth/check
```

Once configured you can create an _OAuth connect_ front end module with which visitors will be able to register/log in to your site.

Screen Shot 2022-12-30 at 21 18 03

In the settings of this module you can select one ore more OAuth clients (and sort them). The module also provides the same settings as the login and registration module regarding autologin, redirection, member groups and home directory creation.

The module renders a form with different submit buttons (and optionally the checkbox to enable autologin) in the front end.

Screen Shot 2022-12-30 at 21 20 51

```html
<div class="module-oauth_connect">
    <h2>OAuth connect</h2>
    <form method="post" id="module-oauth_connect-32">
        <input type="hidden" name="FORM_SUBMIT" value="module-oauth_connect-32">
        <input type="hidden" name="REQUEST_TOKEN" value="…">
        <input type="hidden" name="_target_path" value="https://example.com/protected-page">
        <div class="widget widget-submit client-amazon">
            <button type="submit" name="client" value="5" title="Connect via Amazon">Amazon</button>
        </div>
        <div class="widget widget-submit client-facebook">
            <button type="submit" name="client" value="1" title="Connect via Facebook">Facebook</button>
        </div>
        <div class="widget widget-submit client-github">
            <button type="submit" name="client" value="3" title="Connect via GitHub">GitHub</button>
        </div>
        <div class="widget widget-submit client-google">
            <button type="submit" name="client" value="4" title="Connect via Google">Google</button>
        </div>
        <div class="widget widget-submit client-slack">
            <button type="submit" name="client" value="2" title="Connect via Slack">Slack</button>
        </div>
        <div class="widget widget-checkbox">
            <fieldset class="checkbox_container">
                <span><input type="checkbox" name="autologin" id="autologin" value="1" class="checkbox"> <label for="autologin">Remember me</label></span>
            </fieldset>
        </div>
    </form>
</div>
```

Clicking on one of these buttons will then start the OAuth2 authorization process and when successful the authenticator will log the user in (and create a new member on the fly if applicable).

_Note:_ this module can also be used by members that are already logged in. They will then be connected with the respective OAuth2 provider.

In the _Members_ section there will be a new operation with which you can view the OAuth connections for each member and delete them if you want.

Screen Shot 2022-12-30 at 21 23 41

(the last operation - icon is a placeholder)

Screen Shot 2022-12-30 at 21 25 36

Currently only Amazon, Facebook, GitHub, Google and Slack are available. However, via the `contao_oauth.enabled_providers` configuration you can unlock any (or all) of the **58** providers supported by the [`knpuniversity/oauth2-client-bundle`](https://github.com/knpuniversity/oauth2-client-bundle).

## Implementation Details

Rather than having custom implementations of each individual OAuth2 provider I set out to unify this as much as possible. I am not too happy though with the current implementation as it feels a bit clunky overall.

The reason why only the 5 providers mentioned above are unlocked is because in theory each provider needs some adjustment in its implementation. This could be either some default scopes that are required or some configuration settings for the provider or client. Therefore each implementation needs to be actually tested whether anything is required additionally for it to work (in its most basic form at least) - and I did not do this for all 58 possible options 😉. The ones that are unlocked I tested.

For example:

* Amazon requires scopes to be defined.
* Facebook requires a `graph_api_version` for the client configuration, but scopes are optional.
* Slack requires scopes to be defined.
* For each client, only the OAuth ID is guaranteed to be delivered. Most will deliver also an email address, some might deliver a firstname and a lastname.

If you unlock more providers than Contao unlocks by default, you can use the `OAuthConnectEvent` to add additional data to the member and you can decorate the `contao_oauth.client_generator` service so that you can provide additional client and provider configurations and scopes by default (though scopes can also be defined in the back end as previously mentioned).

The `tl_oauth_client` table uses snake case for its fields because its values are directly used for the OAuth2Client configuration (and its parameter names are snake case).

The OAuth2Clients are created on the fly as defined in the database. Thus this completely bypasses the `KnpUOAuth2ClientBundle` functionality regarding dependency injection.

## Discussion Questions

1. I used _OAuth_ for all names, but technically this is all based on _OAuth2_ implementations. Should I use _OAuth2_ everywhere instead?
2. The authenticator will connect to an _existing_ Contao user, if an email address is provided and that email address is found in Contao. Not sure if this is sensible or not, but I have seen this behavior on other sites.

## Todo

- [ ] Tests
- [ ] Permissions for `tl_member_oauth`
- [ ] 2FA
- [ ] Actually create https://github.com/contao/oauth-bundle/
- [ ] Error handling


fritzmg:
I added some basic error handling via the Symfony `AuthenticationUtils`. However there is one caveat at the moment: if you have a login module and an OAuth connect module on the same page (which is likely) then the login module will capture the last authentication error instead.

oauth-bundle/src/ContaoOAuthBundle.php
```diff
@@ -0,0 +1,34 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\OAuthBundle;
```
ausi, line 13:
```suggestion
namespace Contao\OauthBundle;
```

In camel-case names it has to be written as `Oauth` IMO, otherwise snake-casing it would result in `o_auth_bundle` instead of `oauth_bundle`.

ausi:
Should this work for backend users too? Or would this be a separate pull request?

ausi:
> 2\. The authenticator will connect to an _existing_ Contao user, if an email address is provided and that email address is found in Contao.

This would mean I could access someones existing account on a Contao website if I have the password to an account on any of the activated OAuth providers with the same email address? I would definitely be against that as I’d not expect that sharing my Amazon account with someone would also mean sharing many other accounts 😳

> 1\. I used _OAuth_ for all names, but technically this is all based on _OAuth2_ implementations. Should I use _OAuth2_ everywhere instead?

IMO it should be `OauthBundle`, `oauth-bundle` and `oauth_bundle`, no need for the `2`.

Toflar:
> if you have a login module and an OAuth connect module on the same page (which is likely) then the login module will capture the last authentication error instead.

Same issue would currently happen if you placed the login module twice on the website. Like e.g. in the header and the content. We could fix both of those issues by having a special `FrontendModuleAuthenticationException` that takes the inner exception maybe? Then we can check for that in our modules.

```php
class FrontendModuleAuthenticationException extends AuthenticationException
{
    public function __construct(public int $moduleId, public AuthenticationException $exception) {}
}
```

Just as an idea.

oauth-bundle/src/ContaoOAuthBundle.php
fritzmg: That's already taken care of. The name of the bundle is `ContaoOAuthBundle`, but its alias will be `contao_oauth`, not `contao_o_auth` (which it would be by default). See 

https://github.com/contao/contao/blob/d21c3f9805eed522d152d169a351baf49d23765c/oauth-bundle/src/DependencyInjection/ContaoOAuthExtension.php#L33-L36

`knpuniversity/oauth2-client-bundle` handles it the same way. Its bundle's name is `KnpUOAuth2ClientBundle` but its alias is `knpu_oauth2_client` and not `knp_u_o_auth2_client`.

Personally I prefer to keep `OAuth` in the FQCN instead of `Oauth`. But that's of course highly subjective.

fritzmg:
> Should this work for backend users too? Or would this be a separate pull request?

It would not be much trouble. However, the question is where you would define your OAuth clients then.

oauth-bundle/src/ContaoOAuthBundle.php
ausi: Personally I (strongly 😅) prefer to have strict naming rules for camel case names, and that is IMO that the casing of the base words is strictly ignored. We also lowercase abbreviations like XML or HTTP.

But I was not able to find such a rule in the symfony coding style, so it is subjective, and I’m fine either way ☺️

Ahmadre:
@Toflar @ausi what about this feature? it's still in review since 2023! 

If this is not tracked anymore, I will create another PR.

PR #5619, opened 2023-01-01
Allow using both modern fragments and Twig templates in extensions

m-vo:
This backports #5315 for Contao 4.13.

With this PR, it should be possible to use modern fragments and Twig templates in extensions and be compatible with Contao 4.13 and Contao 5.

Aside from the syntax changes to be compatible with PHP 7.4, the only difference to #5315 now basically is the default template name, that gets set in the `RegisterFragmentsPass` (see fe115410e3e2a4b19f8fe78e618b11fa228275ca).

I went through a lot of manual testing but please also test this PR, preferably in a real project.  :relaxed: There a so many setup variants, so it's easy to miss one.


Toflar:
Looks like phpstan found an unused variable :) For me, the stuff works though but last time, @fritzmg found like 5 other cases. So it looks like I'm not a reference here :D

fritzmg (approved):
lgtm 👍 

I tested with different combinations of templates - including variant templates, similar to https://github.com/contao/contao/pull/5315#pullrequestreview-1152703117 - but with PHP and Twig templates. All the templates were found as expected and also were rendered in the front end accordingly.

Toflar (approved):
I did not test the application from a User's perspective, like @fritzmg did but I can confirm this allows finding and displaying the correct template options in both, 4.13 and 5.x which is a real must! Thanks everybody <3

leofeyer:
Am I correct in assuming that we must not merge these changes upstream?

leofeyer:
Thank you @m-vo.

PR #5620, opened 2023-01-02
Fix `{{date::Y}}` caching

fritzmg:
In #4782 we introduced special handling for the `{{date::Y}}` insert tag, so that it can be cached until the end of the current year, rather than always being evaluated on every request.

However, the response was never actually cached as we add the `no-store` directive to any insert tag response (if no client cache setting is involved):

https://github.com/contao/contao/blob/accec603257a4fcc6c560b273623c5d265ca243d/core-bundle/src/Controller/InsertTagsController.php#L43

This PR fixes that by removing `no-store` for the special `{{date::Y}}` case.

### DIFF
--- a/core-bundle/src/Controller/InsertTagsController.php
+++ b/core-bundle/src/Controller/InsertTagsController.php
@@ -48,6 +48,7 @@ class InsertTagsController
         if ('{{date::Y}}' === $insertTag) {
             $response->setPublic();
             $response->setExpires(new \DateTimeImmutable(date('Y').'-12-31 23:59:59'));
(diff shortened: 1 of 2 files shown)


Toflar (approved):
We might also remove potentially set `E-Tag` and `Last-Modified` headers.
See https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/Response.php#L532 and https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/Response.php#L557

leofeyer:
Thank you @fritzmg.

PR #5626, opened 2023-01-04
Do not combine different asset .js files

leofeyer:
Follow-up on https://github.com/contao/contao/pull/5598#discussion_r1053627360

> So? It's even better with single requests. Why would I download a file that contains the scripts for chosen, simplemodal, datepicker etc. combined? Chances that only one of them changes during a Contao update are pretty high. The way it is now it would generate a new combined file forcing me to download it all again.
Yes, I agree it's still better to combine the stuff that belongs together in one file, even with http2/3. But these files are completely independent from each other. Chosen is not related to simplemodal whatsoever. Imho, we should not combine here.

@contao/developers Do we want to apply the same to the asset .css files?

### DIFF
--- a/core-bundle/contao/templates/backend/be_main.html5
+++ b/core-bundle/contao/templates/backend/be_main.html5
@@ -26,16 +26,12 @@
     <?= $this->stylesheets ?>
 
     <script><?= $this->getLocaleString() ?></script>
-    <script src="<?php
-      $objCombiner = new Contao\Combiner();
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):
Imho, yes :)

leofeyer:
We have to find a solution for bundles to load their asset files. E. g. now the `news.min.css` file is always loaded, because it is part of the minified back end CSS file. This approach is controversial and needs to be discussed.

leofeyer:
See #5796

PR #5628, opened 2023-01-04
Move the back end assets to core-bundle/assets

leofeyer:
This PR moves the back end components (chosen, datepicker and simplemodal) into `core-bundle/assets`.


leofeyer:
Closing this, because @aschempp correctly noted that the moved assets can also be used in the front end and thus have to remain available as components. We will discuss the whole components concept in one of the next developers meeting.

PR #5630, opened 2023-01-05
Run the tools

leofeyer:


core-bundle/src/Twig/Runtime/HighlightResult.php
```diff
@@ -15,7 +15,7 @@
 use Highlight\HighlightResult as BaseHighlightResult;
 
 /**
- * This class is a thin wrapper around @see \Highlight\HighlightResult that
+ * This class is a thin wrapper around @see HighlightResult that
```
ausi, line 18: @leo Doesn’t this change the meaning and should be replaced with this instead?

```suggestion
 * This class is a thin wrapper around @see BaseHighlightResult that
```

leofeyer: Of course. See #5633.

PR #5631, opened 2023-01-05
Implement a dark scheme toggle

aschempp:
as mentioned in https://github.com/contao/contao/pull/5031#issuecomment-1369084846, here's the follow-up to implement a dark mode toggle 😎 

Bildschirmfoto 2023-01-05 um 22 26 18
Bildschirmfoto 2023-01-05 um 22 26 26

It's functionality is derived from what I implemented at https://app.trakked.io. It works like this:
By default, the back end is rendered by the browser/system preference. She user can then toggle to **the opposite state** through the profile menu. This selection is then remembered in the local storage. If a custom selection is switched back to the system default, the locale storage state is dropped and the back end behaves like the system again.

Storing the value in local storage imho makes a lot more sense than the user profile (or the session, which is stored in the user profile as well), simply because I would select a dark mode depending on my device, not my user account.

PS: I also refined the CSS for the drop down to include the small arrow border (that's the changes in `#tmenu ul.menu_level_1:after`) 😉


core-bundle/assets/controllers/colorscheme-controller.js
```diff
@@ -0,0 +1,61 @@
...
+    static targets = ['label'];
+    static values = {
+        i18n: {
+            type: Object,
+            default: { light: 'Disable dark mode', dark: 'Enable dark mode'}
+        }
+    };
+
+    static afterLoad (_identifier, _application) {
+        document.firstElementChild.dataset.colorScheme = prefersDark() ? 'dark' : 'light';
```
ausi: Any reason you are not using the usual `document.documentElement` here?

```suggestion
        document.documentElement.dataset.colorScheme = prefersDark() ? 'dark' : 'light';
```

core-bundle/contao/themes/flexible/basic.css
```diff
@@ -77,61 +77,67 @@
...
-		--active-bg: #1b325f;
-		--active-border: #264787;
-		--drag-bg: #1b325f;
-		--legend: #747b8b;
-		--serp-preview: #bdc1c6;
-		--serp-preview-title: #8ab4f8;
-		--nested-bg: #1e2024;
-		--datepicker-hover: #393e46;
-	}
+html[data-color-scheme="dark"] {
```
ausi, line 80:
```suggestion
html[data-color-scheme="dark"] {
	color-scheme: dark;
```

This is needed so that the UI elements of the browser (e.g. checkboxes) also switch to dark mode.

core-bundle/assets/controllers/colorscheme-controller.js
aschempp: nope 😂  changed in e5ceb2fb7689609127c7e7f38d549432b2bca5ff

core-bundle/contao/themes/flexible/basic.css
aschempp: added in e5ceb2fb7689609127c7e7f38d549432b2bca5ff

leofeyer (changes requested):
There are three major drawbacks with this approach:

1. The light theme icons are always loaded, even in dark mode. This generates unnecessary requests.

2. TinyMCE cannot access `html[data-color-scheme="dark"]` and thus always renders the light color scheme.

3. The icons are no longer aligned correctly in dark mode, which is probably caused by the double images. The issue does not occur in light mode.

leofeyer:
I was able to fix 1 and 3 in ea8c136ca16bed6d1f5e51f62ecbd2457f8eedca. 🎉

leofeyer (dismissed):
Fixed the last issue in 6e2850fd2d0daa18f2ea4cb4e7a2758dc9633ecb.

leofeyer (approved):

Issue #5616, opened 2022-12-30
listWizard missing some CSS reset


fritzmg:
### Affected version(s)

5.0.7, 5.0.x-dev

### Description

There is still at least one other issue caused by #5370 - the `listWizard` is missing some CSS reset:

<img src="https://user-images.githubusercontent.com/4970961/210111351-fed182ca-6c1f-424c-ba00-ec23451a6f70.png" width="540">

aschempp:
see https://github.com/contao/contao/pull/5632

PR #5632, opened 2023-01-06
Reset the list wizard list-style

aschempp:
Fixes https://github.com/contao/contao/issues/5616


Toflar (dismissed):

fritzmg (dismissed):

core-bundle/contao/themes/flexible/main.css
```diff
@@ -1867,6 +1867,12 @@ ul.sgallery li {
 }
 
 /* List wizard */
+.tl_listwizard, .tl_listwizard li {
```
leofeyer:
```suggestion
.tl_listwizard {
```

No need to apply this to the `li` elements, too, is there?

fritzmg: The margin and padding needs to be reset on the `li` elements.

leofeyer: Not on my system (Safari 16.2 on macOS Ventura).



In which browser on which OS does the issue occur for you?

fritzmg: You are right, I must have remembered that wrong.

leofeyer (approved):

PR #5633, opened 2023-01-06
Fix the @see annotations

leofeyer:


ausi (dismissed):

core-bundle/src/Slug/Slug.php
```diff
@@ -27,7 +27,9 @@ public function __construct(private SlugGeneratorInterface $slugGenerator, priva
     }
 
     /**
-     * @param int|iterable $options A page ID, object or options array {@see SlugGeneratorInterface::generate()}
+     * @param int|iterable $options A page ID, object or options array
+     *
+     * @see SlugGeneratorInterface::generate()
```
leofeyer, line 32: We must not use the inline `{@see …}` annotation, because it is apparently not supported by the phpdoc-parser and thus not associated with the `use` statements (see https://github.com/slevomat/coding-standard/issues/770).

core-bundle/src/Repository/RememberMeRepository.php
```diff
@@ -18,8 +18,6 @@
 use Symfony\Component\Security\Core\Exception\TokenNotFoundException;
 
 /**
- * @template-extends ServiceEntityRepository<RememberMe>
```
ausi: This does not seam related to this PR?

leofeyer: The CI will fail without removing this:

```
PHPDoc tag @extends contains generic type Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository<Contao\CoreBundle\Entity\RememberMe> but class  
         Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository is not generic.              
```

ausi: Interesting, removing this line breaks phpstan on my system with the following error: `Class Contao\CoreBundle\Repository\RememberMeRepository extends generic class Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository but does not specify its types: T`

ausi: I think I know why. This was fixed in the doctrine bundle yesterday, see https://github.com/doctrine/DoctrineBundle/commit/251cd5aaea32bb92cdad4204840786b317dcdd4c

So this has to be reverted, otherwise the CI will fail on the next run (with doctrine-bundle version 2.8.2).

core-bundle/src/Repository/CronJobRepository.php
```diff
@@ -18,8 +18,6 @@
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 
 /**
- * @template-extends ServiceEntityRepository<CronJob>
- *
```
ausi: Same here.

core-bundle/src/Repository/RememberMeRepository.php
leofeyer: Reverted in a3377eb096d354b0e242c51ee00b0816653ccffa.

ausi (approved):

PR #5646, opened 2023-01-09
Use enumNode instead of custom validation

Toflar:


core-bundle/tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -252,9 +252,6 @@ public function testValidCronConfiguration(array $params, bool|string $expected)
         $this->assertSame($expected, $configuration['cron']['web_listener']);
     }
 
-    /**
-     * @dataProvider cronConfigurationProvider
-     */
```
Toflar, line 257: This was wrong anyway. It runs the same test 3 times for nothing. This provider belongs to the test above. Copy & paste issue.

fritzmg (approved):

leofeyer:
Thank you @Toflar.

PR #5652, opened 2023-01-09
Only hide the dark/light theme icons via CSS

leofeyer:
I have noticed that the dark theme icons were not hidden in the site structure:

Turned out that `$attributes` has the value `style="margin-right:2px"` in the site structure, which overrides the `style="display:none"` that is added in the `Image::getHtml()` method.

@aschempp What do you think about adding the required CSS format definitions to automatically hide the dark icons to [contao-components/contao](https://github.com/contao-components/contao), too, in case people use the method in the front end?


core-bundle/contao/library/Contao/Image.php
```diff
@@ -148,7 +148,7 @@ public static function getHtml($src, $alt='', $attributes='')
 
 			if (file_exists(Path::join($projectDir, $darkSrc)))
 			{
-				return '<img class="color-scheme--dark" src="' . self::getUrl($darkSrc) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy"' . ($attributes ? ' ' . $attributes : '') . ' style="display:none"><img class="color-scheme--light" src="' . self::getUrl($src) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy"' . ($attributes ? ' ' . $attributes : '') . '>';
+				return '<img class="color-scheme--dark" src="' . self::getUrl($darkSrc) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy"' . ($attributes ? ' ' . $attributes : '') . '><img class="color-scheme--light" src="' . self::getUrl($src) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy"' . ($attributes ? ' ' . $attributes : '') . '>';
```
ausi, line 151:
```suggestion
				return '<img class="color-scheme--dark" src="' . self::getUrl($darkSrc) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy" style="display:none"' . ($attributes ? ' ' . $attributes : '') . '><img class="color-scheme--light" src="' . self::getUrl($src) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . StringUtil::specialchars($alt) . '" loading="lazy"' . ($attributes ? ' ' . $attributes : '') . '>';
```

Changing the order would also fix the issue, right?

ausi (approved):

PR #5654, opened 2023-01-09
Allow cronjobs to be skipped

### LINKED ISSUE #5645: Running a cron job in a specific scope only is not reliable
### Affected version(s)

4.9

### Description

The documentation suggests to use something like the following code if you want to limit a cron job to a specific scope:

```php
// Do not execute this cron job in the web scope
if (Cron::SCOPE_WEB === $scope) {
    return;
}
```

However, this is not reliable because the cron system sees this execution as successful and updates the `lastRun` column accordingly. This means, if you have very regular requests to the frontend the actual code of such a cron job will never be executed.

A possible solution would be to use something like the following:

```php
// Do not execute this cron job in the web scope
if (Cron::SCOPE_WEB === $scope) {
    throw new CronExecutionSkipped();
}
```

and handling this exception in the cron system accodingly by not updating the `lastRun` column in this case.

fritzmg:
Fixes #5645

This introduces a `CronExecutionSkippedException` allowing you to do the following in your cronjob:

```php
public function __invoke(string $scope): void
{
    if (Cron::SCOPE_CLI !== $scope) {
        throw new CronExecutionSkippedException();
    }
}
```

which will ensure that the cronjob will still be executed on the next _CLI_ run - which it otherwise might not if you just return normally (see the description of #5654).

Furthermore this PR also allows all cron jobs to be executed, even when one cronjob happens to throw another exception. The first occurring exception will then be rethrown.

I'll provide a separate PR for `5.0` if we agree on this concept.

### DIFF
--- a/core-bundle/src/Cron/Cron.php
+++ b/core-bundle/src/Cron/Cron.php
@@ -91,8 +91,6 @@ class Cron
 
                 // Determine the last run date
                 $lastRunDate = null;
-
-                /** @var CronJobEntity $lastRunEntity */
                 $lastRunEntity = $repository->findOneByName($name);
 
                 if (null !== $lastRunEntity) {
@@ -136,7 +134,6 @@ class Cron
                 $cron($scope);
             } catch (CronExecutionSkippedException $e) {
                 // Restore previous run date in case cronjob skips itself
-                /** @var CronJobEntity $lastRunEntity */
                 $lastRunEntity = $repository->findOneByName($cron->getName());
                 $lastRunEntity->setLastRun($cron->getPreviousRun());
                 $entityManager->flush();
--- a/core-bundle/src/Cron/Cron.php
+++ b/core-bundle/src/Cron/Cron.php
@@ -150,7 +150,7 @@ class Cron
 
         // Throw the first exception
         if (null !== $exception) {
-            throw $e;
+            throw $exception;
         }
     }
 }
--- a/core-bundle/src/Cron/Cron.php
+++ b/core-bundle/src/Cron/Cron.php
@@ -140,7 +140,7 @@ class Cron
                 $lastRunEntity = $repository->findOneByName($cron->getName());
                 $lastRunEntity->setLastRun($cron->getPreviousRun());
                 $entityManager->flush();
-            } catch (\Exception $e) {
+            } catch (\Throwable $e) {
                 // Catch any exceptions so that other cronjobs are still executed
                 if (null === $exception) {
                     $exception = $e;
--- a/core-bundle/src/Cron/Cron.php
+++ b/core-bundle/src/Cron/Cron.php
@@ -125,7 +125,7 @@ class Cron
         }
 
         $exception = null;
-        
+
         // Execute all cron jobs to be run
         foreach ($cronJobsToBeRun as $cron) {
             try {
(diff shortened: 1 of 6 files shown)


core-bundle/src/Cron/Cron.php
```diff
@@ -120,13 +124,33 @@ public function run(string $scope): void
...
+                    $this->logger->debug(sprintf('Executing cron job "%s"', $cron->getName()));
+                }
+
+                $cron($scope);
+            } catch (CronExecutionSkippedException $e) {
+                // Restore previous run date in case cronjob skips itself
+                /** @var CronJobEntity $lastRunEntity */
+                $lastRunEntity = $repository->findOneByName($cron->getName());
+                $lastRunEntity->setLastRun($cron->getPreviousRun());
+                $entityManager->flush();
```
Toflar, line 139: Isn't there a `persist` call missing here?

fritzmg: You only have to persist new entities, not existing ones. At this stage the entity was either already fetched from the DB or persisted in line 102.

core-bundle/src/Cron/Cron.php
```diff
@@ -120,13 +124,33 @@ public function run(string $scope): void
...
+            } catch (\Exception $e) {
```
Toflar:
```suggestion
            } catch (\Throwable $e) {
```

core-bundle/src/Cron/Cron.php
```diff
@@ -120,13 +124,33 @@ public function run(string $scope): void
...
-                $this->logger->debug(sprintf('Executing cron job "%s"', $cron->getName()));
+            try {
+                if (null !== $this->logger) {
+                    $this->logger->debug(sprintf('Executing cron job "%s"', $cron->getName()));
+                }
+
+                $cron($scope);
+            } catch (CronExecutionSkippedException $e) {
+                // Restore previous run date in case cronjob skips itself
+                /** @var CronJobEntity $lastRunEntity */
```
Toflar: Should we also debug log here?

fritzmg: I thought about that but decided not to since it's the cronjob itself that decides to skip, not the cron service and thus if logging is needed, they would add it themselves there and provide info why it is skipped, like

```php
$this->logger->debug(sprintf('Skipping %s because it is not executed in the CLI scope.', __METHOD__));
```

Toflar (dismissed):
LGTM but I'd prefer having a ✅  by @ausi as well :)

core-bundle/src/Cron/Cron.php
```diff
@@ -120,13 +124,33 @@ public function run(string $scope): void
...
+                $lastRunEntity = $repository->findOneByName($cron->getName());
+                $lastRunEntity->setLastRun($cron->getPreviousRun());
+                $entityManager->flush();
+            } catch (\Throwable $e) {
+                // Catch any exceptions so that other cronjobs are still executed
```
ausi, line 141: Should we add something like this here?
```php
if (null !== $this->logger) {
    $this->logger->error((string) $e);
}
```

ausi (dismissed):

core-bundle/src/Cron/Cron.php
fritzmg: Yes, but not for the first one, since it will be thrown (and thus logged), right?

ausi: I’d probably do it for all of them to make the output more consistent and less confusing

Toflar (dismissed):

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5660, opened 2023-01-10
Improve working with custom back end controllers

m-vo:
This is a very early draft that should help the discussion how custom back end controllers could be improved. Here is the  strategy I'm currently after:

**Primary goals:**

* Have an abstract base class, that adds a thin layer over the current `Contao\CoreBundle\Controller\AbstractController`. This class should be able to generate everything needed to render the chrome. Could alternatively be a trait or a service.
* Allow to opt out from creating the chrome data. This is for example needed if a controller contains multiple actions and some of them return AJAX data (no full page).
* Have a Twig base template `@Contao/backend/_chrome.html.twig`, that renders the familiar Contao back end and has all the things, that needs to be adjusted (blocks, `HtmlAttributes` and so on).

**Secondary goals:**

* Consider a class level attribute (e.g. `#[AsBackendModule('foo', category: 'bar')]` that can be placed on the controller and adds a respective entry in the main menu.

**For Contao 6:**

* Use the controller and template to render all of the back end. This would for instance allow extensions to adjust general things like they can now with the content element base template.

So all in all this is quite similar to the current way of doing things except that we've got a Twig base template and therefore get the need and opportunity to change things up in the abstract controller as well.

<sup>I'm using the word <i>chrome</i> in the code/description meaning [a GUI wrapper](https://stackoverflow.com/questions/5071905/what-does-chrome-mean) (not referring to the browser).</sup>

### DIFF
--- a/core-bundle/config/controller.yaml
+++ b/core-bundle/config/controller.yaml
@@ -21,6 +21,8 @@ services:
                 - { name: container.service_subscriber, id: contao.csrf.token_manager }
 
     # We explicitly allow autowiring and FQCN service IDs in controllers
+    Contao\CoreBundle\Controller\Backend\FooController: ~
+
     Contao\CoreBundle\Controller\BackendController: ~
 
     Contao\CoreBundle\Controller\BackendCsvImportController:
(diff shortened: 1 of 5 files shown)


rabauss:
> I'm using the word chrome in the code/description meaning [a GUI wrapper]

Isn't there a better naming? First I was confused without your hint 😕

m-vo:
> Isn't there a better naming?

That's what browsers/the spec are using, but open for input of course.

m-vo:
Superseded by #3180 and #8828.

PR #5662, opened 2023-01-10
Handle symlinked upload directories outside the Contao root

### LINKED ISSUE #5661: Symlinking files outside Contao root dir
### Affected version(s)

5.1.x

### Description

Our system is deployed using the [Deployer](https://deployer.org/), so some folders are shared across the releases, one of them being the `/files/` folder. The basic structure looks like this:

```
- current –> points to releases/10
- releases
  - 10
    - files –> points to ../../shared/files
- shared
  - files
``` 

This setup has been battle-tested on Contao 4 and everything worked fine. After the recent update to Contao 5.1, we noticed that some images were throwing an error at https://github.com/contao/image/blob/1.x/src/Image.php#L94:

```
Path "/httpdocs/shared/files/media/image/brochure6.png" is not inside root directory "/httpdocs/releases/10"
```

After some debugging I realized that the issue has its roots here: https://github.com/contao/contao/blob/5.x/core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php#L66

The `$item->getPathname()` equals the correct `/httpdocs/releases/10/files` but this is a symlink, so `$item->getLinkTarget()` returns `../../shared/files` which is relative, and ends up being the `/httpdocs/shared/files` root path (or prefix, dunno) for the filesystem adapter. Thus the error above – `/httpdocs/shared/files` is not inside the `/httpdocs/releases/10`.

It looks like the `files` folder can never be symlinked outside the Contao root/project directory. Is there any way this can be fixed?

/cc @Toflar @m-vo @ausi

qzminski:
Closes #5661.

### DIFF
--- a/core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php
@@ -55,7 +55,7 @@ class ConfigureFilesystemPass implements CompilerPassInterface
         $projectDir = $parameterBag->resolveValue($parameterBag->get('kernel.project_dir'));
         $uploadDir = $parameterBag->resolveValue($parameterBag->get('contao.upload_path'));
 
-        $finder = (new Finder())->in($projectDir)->followLinks()->directories()->path('/^'.preg_quote($uploadDir, '/').'/');
+        $finder = (new Finder())->in($projectDir)->directories()->path('/^'.preg_quote($uploadDir, '/').'\//');
 
         foreach ($finder as $item) {
             if (!$item->isLink()) {
--- a/core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php
+++ b/core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php
@@ -55,7 +55,7 @@ class ConfigureFilesystemPass implements CompilerPassInterface
         $projectDir = $parameterBag->resolveValue($parameterBag->get('kernel.project_dir'));
         $uploadDir = $parameterBag->resolveValue($parameterBag->get('contao.upload_path'));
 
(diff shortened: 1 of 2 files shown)


Toflar:
Can we add a test case to cover what this is actually closing? Otherwise we cannot prevent regression bugs.

core-bundle/src/DependencyInjection/Compiler/ConfigureFilesystemPass.php
```diff
@@ -55,7 +55,7 @@ private function mountAdaptersForSymlinks(ContainerBuilder $container, Filesyste
         $projectDir = $parameterBag->resolveValue($parameterBag->get('kernel.project_dir'));
         $uploadDir = $parameterBag->resolveValue($parameterBag->get('contao.upload_path'));
 
-        $finder = (new Finder())->in($projectDir)->directories()->path('/^'.preg_quote($uploadDir, '/').'/');
+        $finder = (new Finder())->in($projectDir)->followLinks()->directories()->path('/^'.preg_quote($uploadDir, '/').'/');
```
ausi: But wouldn’t this find symlinks inside other symlinks and the code below would break then?
What was the issue with the first solution?

qzminski: You are right, I incorrectly thought that following symlinks was the issue, whereas the problem is actually that the `/files/` folder is a symlink that is being resolved. I am still working on this pull request, will keep you updated.

qzminski: I have just pushed the new commit and a unit test, please have a look.

ausi (approved):

Toflar (approved):

leofeyer:
Thank you @qzminski.

m-vo (commented):
LGTM :+1: 

Thanks for fixing @qzminski + @ausi :heart:

PR #5669, opened 2023-01-11
Skip pages in sitemap.xml if URL cannot be generated

aschempp:
I'm using a custom page controller in my project that requires additional route parameters. The `/sitemap.xml` route generates a 500 error because one page URL could not be generated.

### DIFF
--- a/core-bundle/src/Controller/SitemapController.php
+++ b/core-bundle/src/Controller/SitemapController.php
@@ -152,18 +152,18 @@ class SitemapController extends AbstractController
             ) {
                 try {
                     $urls = [$pageModel->getAbsoluteUrl()];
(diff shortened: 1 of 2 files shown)


Toflar:
Why is requireItem not true for this route then?

aschempp:
because only regular Contao pages have `requireItem`. My controller does not have a `parameters` parameter.

Toflar (changes requested):
Then this is fine for me but we need a test to prevent regression bugs.

fritzmg (dismissed):
lgtm, apart from missing test

Toflar (approved):

leofeyer:
Thank you @aschempp.

PR #5672, opened 2023-01-12
Add more back end grid classes

### LINKED ISSUE #1119: Neue CSS-Klassen für "full screen mode"
mit dem neuen "full screen mode" ist die kleinste Aufteilung der Eingabemasken vom 50% immer noch recht breit - die Idee von https://github.com/OMOSde/contao-om-backend/blob/master/FEATURES.md#17-backend-css finden wir vom "MM-Team" sehr gut...

Bitte kurze Rückmeldung, ob sowas in Planung ist oder aus OMO übernommen wird - ansonsten würden wir das für MM in unser CSS + Stylepicker einbauen.

leofeyer:
Implements #1119


core-bundle/contao/dca/tl_member.php
```diff
@@ -206,13 +206,6 @@
 			'eval'                    => array('maxlength'=>64, 'rgxp'=>'phone', 'decodeEntities'=>true, 'feEditable'=>true, 'feViewable'=>true, 'feGroup'=>'contact', 'tl_class'=>'w50'),
 			'sql'                     => "varchar(64) NOT NULL default ''"
 		),
-		'fax' => array
```
Toflar: Why did you remove this field? That's an unnessecary BC break. Data would get lost.

Toflar (commented):
I could approve if it weren't for an (unrelated) field removal.

core-bundle/contao/dca/tl_member.php
leofeyer: I must have committed this unintentionally. Will revert.

leofeyer: Reverted in c9ad1f13d3a82d668e890a2ddb2018f473ea8161.

core-bundle/contao/themes/flexible/basic.css
```diff
@@ -813,12 +813,16 @@ a.tl_submit {
 	display: table;
 }
 
-.w50 {
+.w25, .w33, .w50, .w66, .w75 {
```
ausi, line 816: I think `.w66` and `.w75` should fall back to 100% width in non-large screens.

```suggestion
.w25, .w33, .w50 {
```

leofeyer: No, that will not work (see the first form example). The fields have to behave backwards compatible.

leofeyer: With 100% width, it would fall-back to this:



However, this is how it looks in Contao < 5.1:

ausi (approved):

PR #5673, opened 2023-01-12
Add an attributes_callback for DCA fields

aschempp:


ausi (approved):

ausi:
Example _(do not use, use a listener instead!)_:

```php
'attributes_callback' => array(
	static function (array $attributes, DataContainer $dc = null)
	{
		if ($dc !== null && \in_array($dc->getCurrentRecord()['type'] ?? null, ['download', 'downloads']))
		{
			$attributes['eval']['mandatory'] = true;
		}

		return $attributes;
	}
),
```

PR #5679, opened 2023-01-13
Add a default value for the worker "min" configuration

Toflar:
Fixes the current `Warning: Undefined array key "min"` on `vendor/contao/core-bundle/src/Cron/MessengerCron.php(64): Contao\CoreBundle\Cron\MessengerCron::addWorkerPromises`.


leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5680, opened 2023-01-13
Add isRequired() for desired_size and max settings on workers

Toflar:
Otherwise one can ommit `max` and `desired_size` when configuring autoscaling which makes no sense and will lead to errors.


leofeyer:
Can‘t we set sane defaults?

Toflar:
There are no sane defaults. Every worker works on 1 or more transports/queues. The desired size and maximum number of workers depend on what this queue is handling and what your server can take.

ausi (approved):

leofeyer:
Thank you @Toflar.

PR #5687, opened 2023-01-16
Move the dark mode toggle to the header bar

### LINKED ISSUE #5685: Position of "Dark mode/Light mode" in BE User menu
### Affected version(s)

5.1.0-RC1

### Description

The new option "Dark mode/Light mode" (which I love btw) in the backend user menu (upper right corner) is placed at the very bottom.
This feels somehow misplaced.

On almost every application I know the "logout" option is at the very bottom of the menu.
In my opition the "Dark mode/Light mode" option should be moved further up.
I would even prefer to have no menu entry but a symbol in the top bar.

This is not a bug, maybe this topic can simply be discussed.

leofeyer:
Fixes #5685


Toflar (approved):
Tested and works nicely!

PR #5689, opened 2023-01-16
Make console_path a general Contao configuration

Toflar:
I just had the first use case already where I wanted to re-use `ProcessUtil` which forces me to pass the console path myself at the moment. I think we shouldn't release it like this but instead have a general option. There can only ever be one console path so I think our `ProcessUtil` deserves to know that.


leofeyer:
This makes sense to me. 👍 Will it also work with `vendor/bin/contao-console` in the Managed Edition?

Toflar:
Sure, see https://github.com/contao/contao/pull/5689/files#diff-d5f51e6817cc4908e469dbf9799bebbe87bd74ebd9aab93ea770550f475083cfR59-L60

core-bundle/tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -180,6 +180,24 @@ public function getReservedImageSizeNames(): \Generator
         yield ['right_bottom'];
     }
 
+    public function testSetsTheConsolePath(): void
```
leofeyer, line 183: This method can be removed again IMHO. It basically only tests the Symfony config definition processor, which is already tested in their test suite.

Toflar: Yeah I had a typo initially, hence I added a test case :D Sometimes covering the simplest things make sense. But you can also remove it if otherwise you'd merge :)

leofeyer: No, let‘s keep it then.

leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5690, opened 2023-01-16
Fix the icons when toggling structures in dark mode

leofeyer:
Affects the site structure, the file manager and the checkbox group widget.


Toflar (approved):
I can confirm this fixes the issue :)

fritzmg (approved):

Issue #5096, opened 2022-08-07
Editing templates with the source editor changes line endings to Windows format


ausi:
### Affected version(s)

4.9

### Description

After editing a template file via the source editor in the backend, the line endings get changed from `\n` to `\r\n`. Not sure since when this issue occurs, but AFAIK this wasn’t always the case.

fritzmg:
As far as I tested this solely depends on what the browser (or OS) inputs into the `<textarea>`. However, we could of course always normalize to `\n`.

ameotoko:
I can confirm the problem, see the Slack thread: https://contao.slack.com/archives/CK9GW3ARL/p1657717360616099

ausi:
> As far as I tested this solely depends on what the browser (or OS) inputs into the `<textarea>`.

You sure? The HTML spec seems to define that browsers always have to send `\r\n` for line breaks, see https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart-form-data

fritzmg:
Ah I see. And yeah, seems to apply to [`application/x-www-form-urlencoded`](https://url.spec.whatwg.org/#concept-urlencoded) as well.

ausi:
But why was this not a problem earlier, I’m pretty sure I’d have noticed that.

fritzmg:
Would you though? 🤔  I wouldn't if I don't check the line ending indicator in the lower right within VSCode for example - and when committing such a file to a repo it would likely be normalized by Git (depending on your settings).

ausi:
> Would you though?

I think so, my GIT settings do not normalize this. (but this might not always have been the case 🤔)

How should we solve this? Always normalize to `PHP_EOL` or detect the line endings of the file use the detected ones when saving?

fritzmg:
I think we should detect the line endings before saving and then normalize accordingly.

ausi:
As discussed in the call today we want detect and normalize accordingly.

PR #5692, opened 2023-01-17
Unify newlines in textarea widgets

aschempp:
As recently discussed (and mentioned by @ausi afair), the browser default/HTML standard says textarea newlines should always be submitted in Windows-style (`\r\n`). However, this causes multiple issues. We originally discussed that template files should not be saved/converted to Windows style newlines. However, I also noticed our `maxlength` property counts `\r\n` as two characters, even though for a user it only is one.

I'm not exactly sure this is the right place to fix it, and I don't know if @ausi remembers the RFC about newlines in textarea?

### DIFF
--- a/core-bundle/src/Resources/contao/forms/FormTextArea.php
+++ b/core-bundle/src/Resources/contao/forms/FormTextArea.php
@@ -159,6 +159,15 @@ class FormTextArea extends Widget
 	}
 
 	/**
+	 * Unify newline characters to UNIX style.
+	 * This will also prevent our maxlength from counting newlines as two characters.
+	 */
+	protected function getPost($strKey)
+	{
+		return str_replace("\r\n", "\n", parent::getPost($strKey));
+	}
+
+	/**
 	 * Generate the widget and return it as string
 	 *
 	 * @return string The widget markup
--- a/core-bundle/src/Resources/contao/widgets/TextArea.php
+++ b/core-bundle/src/Resources/contao/widgets/TextArea.php
@@ -99,6 +99,15 @@ class TextArea extends Widget
 	}
 
 	/**
+	 * Unify newline characters to UNIX style.
+	 * This will also prevent our maxlength from counting newlines as two characters.
+	 */
+	protected function getPost($strKey)
+	{
+		return str_replace("\r\n", "\n", parent::getPost($strKey));
+	}
+
+	/**
 	 * Generate the widget and return it as string
 	 *
 	 * @return string


fritzmg:
The [specification](https://html.spec.whatwg.org/multipage/form-elements.html) actually says the following:

> When a [textarea](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element) is [mutable](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-mutable), its [raw value](https://html.spec.whatwg.org/multipage/form-elements.html#concept-textarea-raw-value) should be editable by the user: the user agent should allow the user to edit, insert, and remove text, and to insert and remove line breaks in the form of **U+000A LINE FEED (LF) characters**.

> For historical reasons, the element's value is normalized in three different ways for three different purposes. The [raw value](https://html.spec.whatwg.org/multipage/form-elements.html#concept-textarea-raw-value) is the value as it was originally set. It is not normalized. The [API value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-api-value) is the [value](https://html.spec.whatwg.org/multipage/form-elements.html#dom-textarea-value) used in the value IDL attribute, [textLength](https://html.spec.whatwg.org/multipage/form-elements.html#dom-textarea-textlength) IDL attribute, and by the [maxlength](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-maxlength) and [minlength](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-minlength) content attributes. It is normalized so that line breaks use U+000A LINE FEED (LF) characters. Finally, there is the [value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-value), as used in form submission and other processing models in this specification. It is normalized as for the [API value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-api-value), and in addition, if necessary given the element's [wrap](https://html.spec.whatwg.org/multipage/form-elements.html#attr-textarea-wrap) attribute, additional line breaks are inserted to wrap the text at the given width.
>
> The algorithm for obtaining the element's [API value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-api-value) is to return the element's [raw value](https://html.spec.whatwg.org/multipage/form-elements.html#concept-textarea-raw-value), with [newlines normalized](https://infra.spec.whatwg.org/#normalize-newlines).

> To normalize newlines
in a [string](https://infra.spec.whatwg.org/#string), replace every U+000D CR U+000A LF [code point](https://infra.spec.whatwg.org/#code-point) pair with a single U+000A LF [code point](https://infra.spec.whatwg.org/#code-point), and then replace every remaining U+000D CR [code point](https://infra.spec.whatwg.org/#code-point) with a U+000A LF [code point](https://infra.spec.whatwg.org/#code-point).

ausi:
> The [specification](https://html.spec.whatwg.org/multipage/form-elements.html) actually says the following:

but [it also says](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#converting-an-entry-list-to-a-list-of-name-value-pairs):

> 3\. Replace every occurrence of U+000D (CR) not followed by U+000A (LF), and every occurrence of U+000A (LF) not preceded by U+000D (CR), in value, by a string consisting of U+000D (CR) and U+000A (LF).

ausi:
This PR does not implement what we discussed in https://github.com/contao/contao/issues/5096#issuecomment-1353189856

But maybe we should discuss this again in the context of all usages of `textarea`.

fritzmg:
So according to the spec, if I understand that correctly, a `<textarea>`'s content should have only have LF - but when the browser submits its content all line endings will be CRLF. Therefore this PR would technically be correct and according to the spec, however it would indeed result in changing the line endings of your existing template files when you submit it.

ausi (approved):

leofeyer:
As discussed in the Contao call, we want to merge the PR as is.

leofeyer:
Thank you @aschempp.

PR #5694, opened 2023-01-17
Parse back end URLs with Symfony assets

aschempp:
My application uses Symfony Webpack Encore to compile assets, and therefore has `framework.assets.json_manifest_path` configured. However, since Webpack Encore generates hashed file names, I am unable to use them in the Contao `backend.custom_css` etc. options.

This PR fixes that by always running backend assets through Symfony Assets. If a file does not exist in the manifest.json (or no such file is configured), Symfony Assets will return the original path. It _might_ add a slash to make the URL absolute (since we're calling `getUrl`), but imho that should not be a problem.


fritzmg:
The [specification](https://html.spec.whatwg.org/multipage/form-elements.html) actually says the following:

> When a [textarea](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element) is [mutable](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-mutable), its [raw value](https://html.spec.whatwg.org/multipage/form-elements.html#concept-textarea-raw-value) should be editable by the user: the user agent should allow the user to edit, insert, and remove text, and to insert and remove line breaks in the form of **U+000A LINE FEED (LF) characters**.

leofeyer:
How can I setup a test case to verify the changes?

aschempp:
Add Symfony Encore to your app/project root (see https://github.com/contao-association/website for an example) including the mentioned manifest configuration. Then add a backend CSS (or JS) to the Contao bundle config.

leofeyer (approved):

leofeyer:
Thank you @aschempp.

PR #5695, opened 2023-01-17
Use two different icons for light and dark mode

leofeyer:
Because the current one apparently means "high contrast mode", which is not what the button does.


Toflar (approved):

bytehead (approved):

m-vo:
Noice! :heart:

PR #5696, opened 2023-01-17
Fix installer issue if SQL field has no precision

aschempp:
fixes https://github.com/isotope/core/pull/2351#issuecomment-1378849855

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
@@ -369,7 +369,7 @@ class Installer extends Controller
 					// Field type
 					if ($field['length'])
 					{
-						$field['type'] .= '(' . $field['length'] . ($field['precision'] ? ',' . $field['precision'] : '') . ')';
+						$field['type'] .= '(' . $field['length'] . (isset($field['precision']) ? ',' . $field['precision'] : '') . ')';
 
 						unset($field['length'], $field['precision']);
 					}
--- a/core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
+++ b/core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
@@ -367,7 +367,7 @@ class Installer extends Controller
 					unset($field['index'], $field['origtype']);
 
 					// Field type
-					if ($field['length'])
+					if ($field['length'] ?? null)
 					{
 						$field['type'] .= '(' . $field['length'] . (isset($field['precision']) ? ',' . $field['precision'] : '') . ')';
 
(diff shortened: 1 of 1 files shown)


Toflar (dismissed):

core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -369,7 +369,7 @@ public function getFromDb()
 					// Field type
 					if ($field['length'])
 					{
-						$field['type'] .= '(' . $field['length'] . ($field['precision'] ? ',' . $field['precision'] : '') . ')';
+						$field['type'] .= '(' . $field['length'] . (isset($field['precision']) ? ',' . $field['precision'] : '') . ')';
```
fritzmg, line 372:
```suggestion
						$field['type'] .= '(' . $field['length'] . (!empty($field['precision']) ? ',' . $field['precision'] : '') . ')';
```

Might be more safe?

aschempp: I thought about this as well, but 
 1. we use `isset` for other such cases
 2. `empty` would return false for `0`. Even though that's probably an invalid value, we don't really want to validate the value itself but if something is set we want to use it?

fritzmg (dismissed):

ameotoko:
Doesn't help: https://github.com/isotope/core/pull/2351#issuecomment-1385179734

core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -369,7 +369,7 @@ public function getFromDb()
 					// Field type
 					if ($field['length'])
```
ausi:
```suggestion
					if ($field['length'] ?? null)
```

ameotoko:
There's a lot of things to be fixed here, and they emerge one by one. Perhaps I should have been even more specific in https://github.com/isotope/core/issues/2343 from the very start. When I said `After some digging` in https://github.com/isotope/core/issues/2343#issuecomment-1209367458 – I meant that I already tried exactly the fixes proposed in this PR. There will be a lot of them!

When you suddenly face uncomparable amount of code you need to fix, it might be a sign that you're fixing it in the wrong place. So I went one level up. I am still convinced, that what I said in https://github.com/isotope/core/issues/2343#issuecomment-1209367458 makes sense:

> the actual failing method `Contao\Database\Installer->getFromDb()` is only used by Contao in `Contao\Database\Installer->generateSqlForm()`, which is deprecated since `4.0` and currently not used by Contao itself in any branch.

leofeyer (dismissed):

core-bundle/src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -367,9 +367,9 @@ public function getFromDb()
 					unset($field['index'], $field['origtype']);
 
 					// Field type
-					if ($field['length'])
+					if ($field['length'] ?? null)
```
leofeyer:
```suggestion
					if (isset($field['length']))
```

leofeyer (approved):

leofeyer:
Thank you @aschempp.

PR #5697, opened 2023-01-17
Fix the module wizard

### LINKED ISSUE #5684: Front end modules in page layout broken
### Affected version(s)

5.1.0-RC1

### Description

I'm working on a fresh Contao 5.1.0-RC1 installation.

Problem:
1. create a new theme
2. create a new page layout (e.g. with 3 rows - header, main row and footer)
3. editing/hiding the "Included module" in the section "Front end modules" is not working, creating a new one also does not work

leofeyer:
Fixes #5684


fritzmg:
Why the revert to the state of Contao 4? The modulewizard does not work because it is currently not executed. Unfortunately I do not know how the new `backend.js` entry point works.

leofeyer:
> Why the revert to the state of Contao 4?

The new implementation has other issues besides the one reported in #5684 (e.g. it duplicates multiple rows instead of one). And since it was never a real vanilla-JS rewrite, because it still uses the MooTools `Sortables` class, I decided that it is more efficient to go back to the existing script for now instead of spending time and energy on fixing the new script.

The changes remain in the Git history, though, so we can add them again once we are working on the "getting rid of MooTools" project again.

fritzmg:
> (e.g. it duplicates multiple rows instead of one)

I wasn't aware of that. Though I am also unable to reproduce that. Are there specific circumstances with which this can be reproduced?

core-bundle/assets/scripts/core.js
```diff
@@ -1675,6 +1799,26 @@ window.Backend =
...
+
+		a.href = a.href.replace(/id=[0-9]+/, 'id=' + el.value);
+
+		if (el.value > 0) {
+			td.getElement('a.module_link').setStyle('display', null);
+			td.getElement('img.module_image').setStyle('display', 'none');
+		} else {
+			td.getElement('a.module_link').setStyle('display', 'none');
+			td.getElement('img.module_image').setStyle('display', null);
+		}
```
fritzmg: This will not work, as the `module_image` class does not exist anymore. See https://github.com/contao/contao/issues/5684#issuecomment-1386728717

fritzmg (changes requested):

leofeyer:
> Are there specific circumstances with which this can be reproduced?

It probably was a follow-up issue after I had fixed the initial one.

fritzmg:
I will provide an alternate PR fixing the original issue.

core-bundle/assets/scripts/core.js
leofeyer: Fixed in 30f07c23763068bd1227c730a3e61d176fd547bb.

leofeyer:
Feel free to do so, however, it is a waste of time because it will not be merged anyway; unless your PR also fixes the other issues and replaces the Mootools `Sorting` class with a vanilla-JS solution.

fritzmg:
Originally we decided to replace all _Sortable_ instances separately, once we decided which JavaScript library to choose: https://github.com/contao/contao/pull/4624#issuecomment-1131514181

We haven't really decided yet.

> also fixes the other issues

There are no other issues as far as I can tell.

leofeyer:
Of course there are. All the ones I have fixed in this PR would have to be fixed in your PR as well.

fritzmg:
> All the ones I have fixed in this PR would have to be fixed in your PR as well.

And which issues are those? Otherwise it will not be clear what I have to do in the new PR.

fritzmg (approved):
Tested and did not find any new issues 👍

PR #5698, opened 2023-01-17
Fix wrong page alias when changing the title of a page

Toflar:
Steps to reproduce:

1. Duplicate a page (or empty the alias of an existing page which results in the same initial situation)
2. Change the title of the page
3. Save the changes

Result:

Alias is based on the previous title.

Expected result:

Alias is based on the new title.

The listener is actually wrong already in 4.13 but because it relies on the now deprecated active record there, the output is still correct. Hence, I figured I only adjust and fix this in 5.0 as otherwise we'll have upstream merging issues anyway.
Generating the alias on `getCurrentRecord()` can of course never work, it always represents the current state in the database.

### DIFF
--- a/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
+++ b/core-bundle/src/EventListener/DataContainer/PageUrlListener.php
@@ -80,12 +80,10 @@ class PageUrlListener
             return $value;
         }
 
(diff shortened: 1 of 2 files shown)


leofeyer (approved):

leofeyer:
Thank you @Toflar.

PR #5701, opened 2023-01-18
Revert "Make console_path a general Contao configuration"

leofeyer:
Follow-up on #5689 

This reverts commit 56aaab7ba22bce46d024f47143e8ce397d971c85, because it changed the signature of the `ProcessUtil::createSymfonyConsoleProcess()` method, which can only be done in Contao 6.0.


Toflar:
> This reverts commit [56aaab7](https://github.com/contao/contao/commit/56aaab7ba22bce46d024f47143e8ce397d971c85), because it changed the signature of the `ProcessUtil::createSymfonyConsoleProcess()` method, which can only be done in Contao 6.0.

It's a completely new class that was only added in 5.1. Please revert again, there's no reason to introduce this issue in core.

Toflar (changes requested):
I disagree with this.

leofeyer:
You are right, the class has been added in #5554. Sorry.

PR #5704, opened 2023-01-18
Do not require overwriting the console path

leofeyer:
Follow-up on #5689

With this PR, we no longer have to overwrite the console path because we can finally use `bin/console`.


fritzmg (commented):
May be trigger a deprecation within `manager-bundle/bin/contao-console` saying that `bin/console` should be used instead and `vendor/bin/contao-console` will be removed in Contao 6? (Unless there is a different plan.)

leofeyer:
If we remove the entry point, we have to adjust `contao-setup` accordingly because it runs before the skeleton file has been copied to `bin/console`. We should postpone this until we have gained some experience IMHO.

manager-bundle/src/Command/InstallBinDirCommand.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $fs = new Filesystem();
+        $io = new SymfonyStyle($input, $output);
+
+        $sourcePath = __DIR__.'/../../skeleton/bin/console';
```
m-vo:
```suggestion
        $sourcePath = Path::canonicalize(__DIR__.'/../../skeleton/bin/console');
```

manager-bundle/src/Command/InstallBinDirCommand.php
```diff
@@ -0,0 +1,47 @@
...
+        $targetPath = Path::join($this->projectDir, 'bin/console');
+
+        $fs->copy($sourcePath, $targetPath, true);
```
m-vo: Wouldn't this be a problem in a regular Symfony application, because it would overwrite the console entry point? Or aren't you supposed to run the `ContaoSetupCommand`/`contao-setup` binary in a non-managed edition?

fritzmg: The ManagerBundle is only for the Managed Edition.

m-vo: Ah yes, and the binary and command live there as well. :+1:

manager-bundle/src/Command/InstallBinDirCommand.php
leofeyer: I have copied this from the `InstallWebDirCommand` where it seems to work without `Path::canonicalize()`. So either we change it there as well or we leave it as it is.

aschempp:
Be aware that the Contao Manager also uses the bundle path and a deprecation would unnecessarily confuse people. The Contao Manager can **never** use a user configurable path, because it needs to know what is always available. The Contao Manager can never use `/bin/console`.

fritzmg:
The Contao Manager can still use the bin/console within the bundle and thus we could add a deprecation in contao-console.

manager-bundle/src/Command/InstallBinDirCommand.php
m-vo: I would vote for using it in both places, but feel free to resolve this discussion anyways. :upside_down_face:

leofeyer:
No, that‘s incorrect. The Manager has to have its own entry point just like @aschempp has explained above. We must not use the skeleton file in its original location because the include path would be wrong.

manager-bundle/src/Command/InstallBinDirCommand.php
leofeyer: Changed in b5eb0f0c988bdc69e7120e50861cb2bd98dcba8c.

aschempp:
Should we maybe just have on `install-skeleton` command instead of adding a new command for every new skeleton? 😅
Afair the command could be aliased to continue to work as `install-web-dir`

leofeyer:
IMHO we should have an event that is triggered by the `contao:install` command, so we do not need separate `install-web-dir` and `install-bin-dir` commands at all. But that is outside the scope of this PR.

leofeyer:
Merged and renamed in 10ec7c32939ddffa5a20ef2dfcbec86a742088f7.

bezin:
Just as a note to anybody using the **Symfony dev server** and having a symlink from `bin/console` to `vendor/bin/contao-console` in place in order to run `symfony console`: The changes here will break your console command, because the new `SkeletonInstallCommand` copies the console script to `bin/console`. However, since it is a symlink pointing to `vendor/bin/contao-console`, the script will be copied there and the `autoload.php` path will be wrong, thus breaking the CLI. 

**Solution** Delete the symlink first before updating – or delete your symlink and run `composer install` again.

PR #5706, opened 2023-01-19
Do not add `target="_blank"` to the `{{link}}` insert tag for non-redirect pages

fritzmg:
Fixes the following issue:

1. Create a new page of type `redirect` and enter a title and alias.
2. Enable the **Open in new window** setting and save.
3. Change the page type to `regular` and save.
4. Insert the `{{link::*}}` insert tag (with the ID of the created page) somewhere.

The output of the insert tag will be

```html
<a href="…" title="…" target="_blank" rel="noreferrer noopener">…</a>
```

instead of

```html
<a href="…" title="…">…</a>
```

### DIFF
--- a/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
+++ b/core-bundle/src/Resources/contao/library/Contao/InsertTags.php
@@ -476,7 +476,7 @@ class InsertTags extends Controller
 						}
 
 						$strName = $objNextPage->title;
-						$strTarget = $objNextPage->target ? ' target="_blank" rel="noreferrer noopener"' : '';
+						$strTarget = ($objNextPage->target && 'redirect' === $objNextPage->type) ? ' target="_blank" rel="noreferrer noopener"' : '';
 						$strClass = $objNextPage->cssClass ? sprintf(' class="%s"', $objNextPage->cssClass) : '';
 						$strTitle = $objNextPage->pageTitle ?: $objNextPage->title;
 					}


aschempp (commented):
This also means any custom page type that uses this option won't work anymore. But I can't think of another solution 😕

leofeyer (changes requested):
This is not an acceptable solution because it makes the option unusable for all other page types. That is an unnecessary behavior change to me.

fritzmg:
We could instead implement an onsubmit_callback that disables that setting, if not present in the POST request.

aschempp:
Then again, isn't this a generic issue we have in a lot of places in Contao? Maybe we can come up with a generic solution? Like a list of fields to reset if not present in the current palette 🧐

m-vo:
> Maybe we can come up with a generic solution? Like a list of fields to reset if not present in the current palette 

If we want to discuss this, there are two (linked) challenges, IMHO:

1) Which data/columns belongs to an element? Our models do not know anything about the structure, so, the problem shows, because we use a relational database with shared fields. We could either improve the models (unlikely to work, because of the dynamic nature of the palettes) or we could try to store the data "sparsely", i.e. remove or backup every unrelated piece of information (next question would be: what means "remove"?). 

1) Which data/columns should be shared with a previous selection? This is the (implicit) feature we got by reusing database fields or palette field names respectively. I.e.: If you select another element, you keep the old headline. The question is: Should this selection be preserved if you select element A, then B, then C and only A and C share fields? If yes: How long should the data be kept?

To me, it looks like the only real solution would be to have something like a JSON field `previous_input`, which gets used for the edit mask and populated/merged/reset when writing to the row. Probably nothing that is easily done without rethinking the storage back end. 

So, long story short:  I'd vote to merge the PR as is for now :+1:

asaage:
> Maybe we can come up with a generic solution?

Sometimes it's realy cumbersome to delete field content especialy if the field is mandatory in one type of page/content but not visible in an other type. I don't know if a list of fields to reset if not present in the current palette would do the trick if that happens automatically onsubmit.

leofeyer:
As discussed in the Contao call, we want to merge the PR as it is, since the insert tag should behave like the navigation module and the same check takes place there as well:

https://github.com/contao/contao/blob/b320f38f2809d5d358bcc23be432fdf9b90aef3a/core-bundle/src/Resources/contao/modules/Module.php#L438-L445

leofeyer (approved):

leofeyer:
Thank you @fritzmg.

PR #5708, opened 2023-01-23
Add the missing null check for button_callback

aschempp:
This was already added for callables in 6b47b8ec5acb18fc09389325e2ad0ad3f80b501b but I forgot it for array callbacks.

### DIFF
--- a/core-bundle/src/Resources/contao/classes/DataContainer.php
+++ b/core-bundle/src/Resources/contao/classes/DataContainer.php
@@ -1091,7 +1091,7 @@ abstract class DataContainer extends Backend
 			if (\is_array($v['button_callback'] ?? null))
 			{
 				$this->import($v['button_callback'][0]);
-				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'] ?? '', $label, $title, $v['class'], $attributes, $this->strTable, $this->root);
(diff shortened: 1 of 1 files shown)


core-bundle/src/Resources/contao/classes/DataContainer.php
```diff
@@ -1091,7 +1091,7 @@ protected function generateGlobalButtons()
 			if (\is_array($v['button_callback'] ?? null))
 			{
 				$this->import($v['button_callback'][0]);
-				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'] ?? '', $label, $title, $v['class'], $attributes, $this->strTable, $this->root);
+				$return .= $this->{$v['button_callback'][0]}->{$v['button_callback'][1]}($v['href'] ?? null, $label, $title, $v['class'] ?? null, $attributes, $this->strTable, $this->root);
```
Toflar, line 1094: Why would `href` suddenly return null rather than `''`?  That's a behavior change?

aschempp: Its now the same as in the line below. It is also better imho because it allows the callback to distinguish whether the option was set or not.

leofeyer (approved):

core-bundle/src/Resources/contao/classes/DataContainer.php
leofeyer: We also use `?? null` in lines 929, 935 and 1100, so this is the only place where it is `?? ''` and thus the change is correct IMHO.

Toflar (approved):

leofeyer:
Thank you @aschempp.

PR #5711, opened 2023-01-23
Add the missing "header_icon" CSS classes

leofeyer:


aschempp:
Can't we automatically add that class to the configuration array? Otherwise every extension has to be adjusted as well (we all use at least `header_edit_all`)

leofeyer:
I see the problem. Let‘s just close the PR then. It is not worth spending time on this.

PR #5713, opened 2023-01-23
Fix vertical align of drag handle

fritzmg:
The drag handle in the file manager is currently slightly offset (at least under Firefox and Chrome):

This is fixed by setting `vertical-align: baseline` on the `<button>`.


leofeyer:
All other elements have `vertical-align: middle`. Are you sure that `baseline` is correct?

fritzmg:
Only the ``s have `vertical-align: middle`. Their parents have nothing set. The browser default is `vertical-align: baseline` for `<a>` and `vertical-align: middle` for `<button>`, hence the discrepancy.

leofeyer:
I can reproduce the issue in Firefox on macOS 13.2. It seems that the `vertical-align: middle` of the `<button>` element causes the issue. Can you please check if removing it fixes the issue, too?

fritzmg:
I can also reproduce the problem in Chrome 109.0.5414.75 under Windows 10:



Removing `vertical-align: middle` would fix the issue (this would move all the buttons a few pixels up though, if we remove that generally).

leofeyer:
Please check if #5732 fixes the issue for you.

fritzmg:
> Please check if #5732 fixes the issue for you.

👍

PR #5714, opened 2023-01-23
Use the HTML sanitizer component as Twig filter

ausi:


composer.json
```diff
@@ -121,6 +121,7 @@
         "symfony/filesystem": "^5.4 || ^6.0",
         "symfony/finder": "^5.4 || ^6.0",
         "symfony/framework-bundle": "^5.4 || ^6.0",
+        "symfony/html-sanitizer": "^5.4 || ^6.0",
```
m-vo:
```suggestion
        "symfony/html-sanitizer": "^6.1",
```

The component was first released in version `6.1`.

Can we maybe also bump the "symfony/twig-bridge" to ^6.1? Then we would also automatically get the [`HtmlSanitizerExtension`](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Bridge/Twig/Extension/HtmlSanitizerExtension.php) that does the setup for the `sanitize_html` filter for free.

ausi: Done in 55f82e25a04a4de064ad3c328a4ce9cd9861f921.

core-bundle/src/Twig/Runtime/SanitizerRuntime.php
```diff
@@ -12,66 +12,25 @@
...
-    public function __construct(private readonly ContaoFramework $framework)
+    public function __construct(private readonly HtmlSanitizerExtension $sanitizerExtension)
     {
     }
 
-    public function sanitizeHtml(string $html): string
+    public function sanitizeHtml(string $html, string $sanitizer = null): string
     {
-        $this->framework->initialize();
+        $html = $this->sanitizerExtension->sanitize($html, $sanitizer);
```
m-vo: This looks good to me, just a thought:

In the `ContaoExtension`, for referencing the `EscaperExtension` we inject `@twig` and use `$escaperExtension = $environment->getExtension(EscaperExtension::class);`. Should we do the same here or also inject the explicit service at the other place?

A point for using `@twig` is, that it's independent of service name changes or refactoring of the Twig/bridge setup (happened before). But this version maybe looks a bit cleaner.

m-vo:
@ausi I raised the "symfony/service-contracts" requirements as well in 3505db0, this seems to have made the CI green again. The issue originates from here: https://github.com/symfony/symfony/pull/45680.

IMHO this is a Symfony bug, so maybe we should add a conflict via `contao/conflicts` instead and wait for it to be fixed.

m-vo:
> IMHO this is a Symfony bug, so maybe we should add a conflict via contao/conflicts instead and wait for it to be fixed.

Ok, after some more investigation: Yes it is a bug, no Symfony won't fix this anymore. It's a dependency problem baked in forever. So we need to resolve it ourselves anyway.

core-bundle/src/Twig/Runtime/SanitizerRuntime.php
ausi: Done in 55242b72cab09a5a7a5d42f8e27f3c1d8e0385c0

core-bundle/src/Twig/Runtime/SanitizerRuntime.php
```diff
@@ -12,66 +12,26 @@
...
-        ['key' => 'style', 'value' => 'media'],
-        ['key' => 'time', 'value' => 'datetime'],
-        ['key' => 'details', 'value' => 'open'],
-    ];
-
     /**
      * @internal
      */
-    public function __construct(private readonly ContaoFramework $framework)
+    public function __construct(readonly private Environment $twig)
```
fritzmg:
```suggestion
    public function __construct(private readonly Environment $twig)
```

m-vo (dismissed):
Big thanks! :heart:

composer.json
```diff
@@ -200,6 +201,8 @@
         "phpunit/phpunit": "<8.0",
         "roave/better-reflection": "<4.12.2 || >=6.0",
         "symfony/dependency-injection": "5.4.16 || 6.0.16 || 6.1.8 || 6.2.0 || 6.2.1",
+        "symfony/framework-bundle": "6.1.0 || 6.1.1",
+        "symfony/twig-bundle": "6.1.0",
```
leofeyer: These conflicts are unnecessary. We can require `^6.1.2` and `^6.1.1` above instead.

leofeyer (approved):

leofeyer:
Thank you @ausi.

PR #3276, opened 2021-08-07
Check catalogue for country or locale translations

ausi:
| Q                | A
| -----------------| ---
| Fixed issues     | Fixes #3266

PR #5717, opened 2023-01-24
Make the template selection translatable in the back end

m-vo:
This is how the template variant selection (dropdown) currently looks:

image

With this PR, you are now able to provide translations for your template files, so that they become more readable to back end users:

```yaml
# translations/templates.en.yaml
content_element/text: 'Default text'
content_element/text/special: 'Special text'
```

image

I made sure the template names were still included in brackets behind the translation, because you need them to easily find/create the respective file for adjustments.

/cc @frontendschlampe


core-bundle/src/Twig/Finder/Finder.php
```diff
@@ -123,14 +122,34 @@ public function asTemplateOptions(): array
             return $this->translator->trans('MSC.global', [], 'contao_default');
         };
 
-        foreach (array_keys(iterator_to_array($this->getIterator())) as $identifier) {
-            $label = sprintf(
-                '%s [%s]',
+        $getCustomLabel = function (string $identifier, array $sourceLabels): string|null {
+            $translationId = "TEMPLATE.$identifier";
+
+            if (!$this->translator->getCatalogue()->has($translationId, 'contao_default')) {
```
ausi: I’m not sure if we should introduce a new `TEMPLATE` category here that is maped to the `default` language file. We would have to add it here for example: https://github.com/contao/contao/blob/3df28e79dfaddcdf403023e12b033d7a27a68cd4/core-bundle/contao/library/Contao/InsertTags.php#L360-L363

But if we would use lowercase `template` the `template.php` translation file would be loaded automatically.

```suggestion
            $translationId = "template.$identifier";

            if (!$this->translator->getCatalogue()->has($translationId, 'contao_template')) {
```

fritzmg (commented):
We already use regular Symfony translations for the back end labels of image sizes (`image_sizes`) and mailer transports (`mailer_transports`). I think it would make sense to also have a `templates` translation domain accordingly - instead of creating a new legacy Contao translation category.

m-vo:
>  I think it would make sense to also have a `templates` translation domain accordingly

I agree. Implemented in da977df4462ee66dc8deeba66fcdbf32ba807b2b.

core-bundle/src/Twig/Finder/Finder.php
m-vo: I went with a new translation domain instead of a prefix in https://github.com/contao/contao/pull/5717#issuecomment-1403472654.

core-bundle/src/Translation/Translator.php
```diff
@@ -76,7 +76,7 @@ public function getLocale(): string
         return $this->translator->getLocale();
     }
 
-    public function getCatalogue($locale = null): MessageCatalogue
+    public function getCatalogue($locale = null): MessageCatalogueInterface
```
fritzmg: That change is unrelated, I think?

m-vo: No, that's intentional and correct this way in terms of LSP (IMHO). Otherwise mocking is impossible.

fritzmg: The return type was changed from `MessageCatalogueInterface` to `MessageCatalogue` in https://github.com/contao/contao/pull/3276 
But yes, the return type as defined in the `TranslatorBagInterface` is indeed supposed to be `MessageCatalogueInterface`.

/cc @ausi

ausi: We always return an instance of `Contao\CoreBundle\Translation\MessageCatalogue` so IMO this was correct.
I’d mock the translations by setting `$GLOBALS['TL_LANG'][…]`. 
But we can also change the `MessageCatalogue` from `final` to `@final`.
But I’m also fine with the change to `MessageCatalogueInterface`, so it’s your choice ☺️

m-vo: > We always return an instance of Contao\CoreBundle\Translation\MessageCatalogue so IMO this was correct.

Yes, it is correct from what the current implementation is in the method, but, I think, when I want to work with a instance of `TranslatorBagInterface`, I only ever want/need to get a `MessageCatalogueInterface` (that's what the contract defined by the interface says). Our `MessageCatalogue` does not add anything that is not covered by the interface. So I only see downsides in not using the interface as the return type. :smile: 

Does that make sense? If so, I'd go with the current version. :relaxed:

fritzmg: Also if you decorate `Contao\CoreBundle\Translation\MessageCatalogue` this would also fail, no?

fritzmg (approved):

ausi (approved):

leofeyer:
Thank you @m-vo.

aschempp:
Can a bundle provide such a translation file to translate its own templates? As far as I know, translations are not merged from different bundles?

m-vo:
I don't know - how do we do it for image sizes (or mailer transports) that are already around for a while?

aschempp:
Image sizes and mailer transports are a contao bundle / application configuration. Having templates is not ... I'm pretty sure extensions - including complex themes - would love to include template labels, but afaik they currently cannot because Symfony does not merge translation files.

m-vo:
I also think this would be nice to have. Any ideas how to solve this?

fritzmg:
> Can a bundle provide such a translation file to translate its own templates?

Yes they can.

```
bundles/test/
├── composer.json
├── config
│   └── services.yaml
├── contao
│   ├── dca
│   │   └── tl_content.php
│   └── templates
│       └── content_element
│           └── test.html.twig
├── src
│   ├── ContaoManager
│   │   └── Plugin.php
│   ├── Controller
│   │   └── ContentElement
│   │       └── TestController.php
│   ├── DependencyInjection
│   │   └── TestExtension.php
│   └── TestBundle.php
└── translations
    └── templates.en.yaml
```
```yaml
# bundles/test/translations/templates.en.yaml
content_element/test: 'Custom translation for default template'

```

aschempp:
> I also think this would be nice to have. Any ideas how to solve this?

by using regular Contao translations

> > Can a bundle provide such a translation file to translate its own templates?
> 
> Yes they can.
> 
> ```
> bundles/test/
> ├── composer.json
> ├── config
> │   └── services.yaml
> ├── contao
> │   ├── dca
> │   │   └── tl_content.php
> │   └── templates
> │       └── content_element
> │           └── test.html.twig
> ├── src
> │   ├── ContaoManager
> │   │   └── Plugin.php
> │   ├── Controller
> │   │   └── ContentElement
> │   │       └── TestController.php
> │   ├── DependencyInjection
> │   │   └── TestExtension.php
> │   └── TestBundle.php
> └── translations
>     └── templates.en.yaml
> ```
> 
>  ```yaml
> # bundles/test/translations/templates.en.yaml
> content_element/test: 'Custom translation for default template'
> ```
> 
> 

as far as I now the Symfony translator, multiple bundles cannot provide the same translation file. They would not be merged. Or did you try this with multiple bundles, maybe the behaviour has changed?

fritzmg:
It works:

```yaml
# bundles/lorem/translations/templates.en.yaml
content_element/lorem: 'Lorem translation for default template'
```



```yaml
# bundles/ipsum/translations/templates.en.yaml
content_element/ipsum: 'Ipsum translation for default template'
```

aschempp:
Great! Maybe they improved that in a recent version 😎 

> # bundles/lorem/translations/templates.en.yaml
> # bundles/ipsum/translations/templates.en.yaml

I assume that means two `translations` folders in different `vendor` bundles?

fritzmg:
It's two different bundles, yes (local bundles in this case).

It also works in Symfony 4.4 btw. - so doesn't seem to be that recent ;)

PR #5720, opened 2023-01-24
List the default template first in the template selection drop-down

### LINKED ISSUE #5719: Core twig templates aren't selected by default if a custom twig template exists
### Affected version(s)

5.0

### Description

In Contao 4, you can add a custom content element template like `ce_text_custom.html5` and Contao will continue using `ce_text.html5` by default for new text elements. In order to use `ce_text_custom.html5` you have to specify this on the element settings by changing the "Content element template" select field.

In Contao 5, if you add a custom twig template for a content element then it's possible that this becomes the default instead of the core template (I believe the sorting is alphabetical, so it depends what you name the custom template as to whether or not it becomes the default).

The expected behaviour is that Contao prefers default core templates and only uses custom templates when they are explicty set like in Contao 4.

This can be recreated by creating a custom Twig template for a text element:
`/templates/content_element/text/custom.html.twig`

This now becomes the default for all text elements, instead of `content_element/text` from ContaoCore which should still be the default.

templates

m-vo:
Closes #5719 

@jamesdevine Can you check if this solves your issue?

### DIFF
--- a/core-bundle/src/Twig/Finder/Finder.php
+++ b/core-bundle/src/Twig/Finder/Finder.php
@@ -132,6 +132,9 @@ final class Finder implements \IteratorAggregate, \Countable
             $options[$identifier !== $this->identifier ? $identifier : ''] = $label;
         }
 
+        // Make sure the default option is the first one (see #5719)
+        ksort($options);
+
         return $options;
     }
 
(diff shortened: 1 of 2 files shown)


m-vo:
Note to myself: when this fix is accepted, we need to introduce it in #5619 as well.

Toflar (approved):

fritzmg (approved):

leofeyer:
Thank you @m-vo.

